PR #4896, opened 2024-02-01
Implements wrangler deploy/dev support for Python.

dom96:

This PR implements `wrangler deploy index.py` and `wrangler deploy` (with a wrangler.toml that defines `main = "index.py"`). This means Python Workers can be deployed using Wrangler.

Tested manually via:

```
$ cat $PWD/index.py
from js import Response
def fetch(request):
  return Response.new('Updated Python Worker value')
$ pnpm run --filter wrangler start dev $PWD/index.py --compatibility-flag experimental
```

Also using tests:

```
$ cd packages/wrangler
$ pnpm test # added --testNamePattern=\"can run and modify python worker during dev session (local)\" --silent=false in packages.json
$ pnpm test:e2e
```

I included an e2e test for `dev` but I couldn't run it locally. Other e2e tests fail locally for me too so I guess I just have something misconfigured. Help appreciated in getting it working.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because:

**Note for PR author:**

### DIFF
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -240,6 +240,77 @@ describe("basic dev tests", () => {
 	});
 });
 
+describe("basic dev python tests", () => {
+	let worker: DevWorker;
+
+	beforeEach(async () => {
+		worker = await makeWorker();
+		await worker.seed((workerName) => ({
+			"wrangler.toml": dedent`
+					name = "${workerName}"
+					main = "index.py"
+					compatibility_date = "2023-01-01"
+			`,
+			"index.py": dedent`
+				from js import Response
+				def fetch(request):
+					return Response.new('py hello world')`,
+			"package.json": dedent`
+					{
+						"name": "${workerName}",
+						"version": "0.0.0",
+						"private": true
+					}
+					`,
+		}));
+	});
+
+	it.only("can run and modify python worker during dev session (local)", async () => {
+		await worker.runDevSession("", async (session) => {
+			const { text } = await retry(
+				(s) => s.status !== 200,
+				async () => {
+					const r = await fetch(`http://127.0.0.1:${session.port}`);
+					return { text: await r.text(), status: r.status };
+				}
+			);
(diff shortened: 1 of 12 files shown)


packages/wrangler/e2e/dev.test.ts
```diff
@@ -240,6 +240,77 @@ describe("basic dev tests", () => {
...
+					{
+						"name": "${workerName}",
+						"version": "0.0.0",
+						"private": true
+					}
+					`,
+		}));
+	});
+
+	it.only("can run and modify python worker during dev session (local)", async () => {
```
penalosa, line 268:
```suggestion
	it("can run and modify python worker during dev session (local)", async () => {
```

packages/wrangler/src/deployment-bundle/worker.ts
```diff
@@ -4,7 +4,7 @@ import type { Json } from "miniflare";
 /**
  * The type of Worker
  */
-export type CfScriptFormat = "modules" | "service-worker";
+export type CfScriptFormat = "modules" | "service-worker" | "python";
```
penalosa, line 7: I'm not sure this is quite right—a python worker should still be a modules worker, I think?

penalosa:
This should probably include a default module rule mapping `*.py` to `python`

packages/wrangler/src/deployment-bundle/worker.ts
dom96: I guess it really depends what the meaning for this type is. I think differentiating `module` (JS) vs. `modules` (Python) makes sense since the code which reads these ultimately does a completely different thing depending on its value. Maybe renaming "python" to "modules-python" would be best?

dom96:
@penalosa can you elaborate on this? Where do I add it and what effect will it have?

dom96:
Closing in favour of #4901

PR #4899, opened 2024-02-01
chore: bump vite and vitest to latest versions

petebacondarwin:
Closes #4798 

In attempting to update Vitest in a dependabot PR, we noticed that there were a mix of versions of Vitest and that there would be some tricky parts to bumping the version. Moreover updating to the latest Vitest requires updating Vite as a transitive dependency, and we have a direct dependency on this too.

This work here brings everything to the latest versions and aligns it across the repository.

The update to Vitest had some breaking changes to a number of tests and configuration files have been fixed accordingly.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [x] Not necessary because: it is an internal chore.
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: it is an internal chore.

### DIFF
--- a/fixtures/additional-modules/test/index.test.ts
+++ b/fixtures/additional-modules/test/index.test.ts
@@ -167,33 +167,33 @@ describe("find_additional_modules deploy", () => {
 		const bundledEntry = await fs.readFile(bundledEntryPath, "utf8");
 		expect(bundledEntry).toMatchInlineSnapshot(`
 			"// src/index.ts
-			import common from \\"./common.cjs\\";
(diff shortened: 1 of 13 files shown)


petebacondarwin:
I believe the only blocker for this is now fixing the e2e tests: https://github.com/cloudflare/workers-sdk/pull/4907

mrbbot (approved):

dario-piotrowicz (approved):

CarmenPopoviciu (approved):

Issue #4578, opened 2023-12-08
🚀 Feature Request:


PANstudio:
### Describe the solution

Wrangler should not demand that local simulation of Hyperdrive is achieved by putting the Postgres connection string in the TOML file. The TOML file gets committed, and if the developer's dev DB lives remotely (as opposed to on their local machine), this will mean committing sensitive information. Instead, it should be possible to put the connection string in the `.dev.vars` file, or to load one TOML file in dev and another in production. Right now, omitting the connection string from the TOML raises an error.

mrbbot:
Hey! 👋 Thanks for raising this. For now, you should be able to get around this by pointing Wrangler to a different file that's generated from your `wrangler.toml` file, then adding the generated file to your `.gitignore`. /cc @OilyLime

Cali93:
+1 for this feature. I'm also in the same situation.

gp27:
+1 for this feature.
Before PR #4222 was merged it was possible to setup a variable inside `.dev.vars`:
```
DEV_CONNECTION_STRING=postgresql://[...]
```
It would then be possible to connect without issues, both in dev and production, with code similiar to this.
```js
const client = new pg.Client({
  connectionString: env.HYPERDRIVE_BINDING?.connectionString || env.DEV_CONNECTION_STRING
});```

After the merge this workaround doesn't work anyore. The new `localConnectionString` hyperdrive parameter is required to be set in wrangler.toml, which is less than ideal since it forces to commit the development conenction string.

There is also another issue with the use of `localConnectionString`: all the url parameters are ignored and replaced with `?ssl_mode=ignore` (this actually happens inside workerd and not wrangler / miniflare)
https://github.com/cloudflare/workerd/blob/2c0cd6821ad016f9b14d533c62a12c524dea7aac/src/workerd/api/hyperdrive.c%2B%2B#L77
If can try to set `localConnectionString = "postgresql://user:pass@localhost:5432/db?ssl_mode=require"`but then at runtime`env.HYPERDRIVE_BINDING.connectionString` will be equel to `postgresql://user:pass@localhost:5432/db?ssl_mode=ignore` which breaks SSL only databases

PANstudio:
Thanks all. @mrbbot yes that's what I ended up doing after kind help from CF Discord. @gp27 yeah I too discovered that you apparently can't specify HD bindings from `.dev.vars` - thanks for pointing out the merge that put an end to this (not sure why, or whether this was intended or a side effect of something else.)

[...]

PR #4900, opened 2024-02-01
Add feature to use environment variable for Hyperdrive local dev connection string

OilyLime:
Fixes # [4578](https://github.com/cloudflare/workers-sdk/issues/4578#issuecomment-1855419281).

This PR allows users to use an environment variable to configure their local dev connection string. This means they do not have to commit that connection string to source control.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because:


packages/wrangler/src/dev.tsx
```diff
@@ -949,11 +950,26 @@ export function getBindings(
 		vectorize: configParam.vectorize,
 		constellation: configParam.constellation,
 		hyperdrive: configParam.hyperdrive.map((hyperdrive) => {
-			if (!hyperdrive.localConnectionString) {
+			const connectionStringFromEnv =
+				getHyperdriveLocalConnectionStringFromEnv();
+			if (!connectionStringFromEnv || !hyperdrive.localConnectionString) {
 				throw new UserError(
-					`In development, you should use a local postgres connection string to emulate hyperdrive functionality. Please setup postgres locally and set the value of "${hyperdrive.binding}"'s "localConnectionString" to the postgres connection string in your wrangler.toml`
+					`When developing locally, you should use a local Postgres connection string to emulate Hyperdrive functionality. Please setup Postgres locally and set the value of the 'HYPERDRIVE_LOCAL_CONNECTION_STRING' variable or "${hyperdrive.binding}"'s "localConnectionString" to the Postgres connection string.`
```
rozenmd, line 957: But it doesn't *have* to be a postgres instance running locally, right? I could spin up a dev db and use its connection string here?

OilyLime: Correct. It's the connection string to the database the local runtime will use, but it doesn't have to be local to that machine.

OilyLime:
> I'm a little concerned about this feature. What happens if the user has multiple hyperdrive bindings? Maybe `wrangler.toml` should support something like `local_connection_string = "$env:MY_LOCAL_ENV_VAR"`, which would load the value from `MY_LOCAL_ENV_VAR`? Or instead of just `HYPERDRIVE_LOCAL_CONNECTION_STRING`, you could prefer `HYPERDRIVE_LOCAL_CONNECTION_STRING_<hyperdrive_id>` variables if specified? I'd also suggest prefixing the variable with `WRANGLER_` to match our other environment variable configurations:
> 
> https://github.com/cloudflare/workers-sdk/blob/e1b777193e478af3dd83f148f38ce5a41656259c/packages/wrangler/src/environment-variables/factory.ts#L3
> 
> 
> What do you think? Open to other ideas too. 🙂

Thanks @mrbbot! Opted for using the name of the binding in the environment variable, like `WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_FOO` to allow users to play around with local dev without having created a Hyperdrive config. It also makes the strings less unwieldy.

packages/wrangler/src/dev.tsx
```diff
@@ -949,11 +950,27 @@ export function getBindings(
 		vectorize: configParam.vectorize,
 		constellation: configParam.constellation,
 		hyperdrive: configParam.hyperdrive.map((hyperdrive) => {
-			if (!hyperdrive.localConnectionString) {
+			const connectionStringFromEnv = getHyperdriveLocalConnectionStringFromEnv(
+				hyperdrive.binding
+			);
```
mrbbot, line 955: Rather than using multiple functions here, you could probably simplify this to...
```suggestion
			const connectionStringFromEnv = process.env[`WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_${hyperdrive.binding}`];
```

`getEnvironmentVariableFactory()` provides additional features such as default values and deprecated aliases that we don't need here.

packages/wrangler/src/dev.tsx
```diff
@@ -949,11 +950,27 @@ export function getBindings(
...
+			if (!connectionStringFromEnv || !hyperdrive.localConnectionString) {
 				throw new UserError(
-					`In development, you should use a local postgres connection string to emulate hyperdrive functionality. Please setup postgres locally and set the value of "${hyperdrive.binding}"'s "localConnectionString" to the postgres connection string in your wrangler.toml`
+					`When developing locally, you should use a local Postgres connection string to emulate Hyperdrive functionality. Please setup Postgres locally and set the value of the 'WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_{binding_name}' variable or "${hyperdrive.binding}"'s "localConnectionString" to the Postgres connection string.`
```
mrbbot, line 958: Could you include the actual binding name in the error message?
```suggestion
					`When developing locally, you should use a local Postgres connection string to emulate Hyperdrive functionality. Please setup Postgres locally and set the value of the 'WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_${hyperdrive.binding}' variable or "${hyperdrive.binding}"'s "localConnectionString" to the Postgres connection string.`
```

.changeset/khaki-feet-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+wrangler dev now supports the HYPERDRIVE_LOCAL_CONNECTION_STRING environmental variable for connecting to a local database instance when testing Hyperdrive in local development. This environmental variable takes precedence over the localConnectionString set in wrangler.toml.
```
mrbbot, line 5: Could you update this to follow the [`Changeset message format`](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changeset-message-format) guidelines?

mrbbot (commented):
Nice! Added a few comments. 🙂

mrbbot (approved):

rozenmd:
Merging as requested.

packages/wrangler/src/dev.tsx
```diff
@@ -955,11 +955,28 @@ export function getBindings(
 		vectorize: configParam.vectorize,
 		constellation: configParam.constellation,
 		hyperdrive: configParam.hyperdrive.map((hyperdrive) => {
-			if (!hyperdrive.localConnectionString) {
+			const connectionStringFromEnv =
+				process.env[
+					`WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_${hyperdrive.binding}`
+				];
+			if (!connectionStringFromEnv || !hyperdrive.localConnectionString) {
```
petebacondarwin, line 962: Oops!

Should be && here not ||

gp27:
Thanks for implementing this feature.

I just got a question. Is there a specific reason why these specific variables are read only from `.env` instead of  `.dev.vars`?

PR #4901, opened 2024-02-02
Python support

penalosa:
An alternative approach to https://github.com/cloudflare/workers-sdk/pull/4896

Compared to #4896, this leans more on Wrangler's setup for `--no-bundle` JS workers, and treats Python workers as a special case of `--no-bundle` workers. It:
- Ensures that python modules are properly supported throughout the `--no-bundle` pipeline
- Turns on `--no-bundle` for Python workers, as well as adding a default Python module rule

This has a couple of benefits—notably saving python source files will trigger a `workerd` restart, and multiple python files which import each other are supported (there appears to be a `workerd` bug with imported nested python modules, so that isn't supported)

This PR also adds basic support for a `requirements.txt` file for python workers

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs 
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because:


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -1142,6 +1146,15 @@ function updateQueueConsumers(config: Config): Promise<string[]>[] {
 	});
 }
 
+// TODO(soon): workerd requires python modules to be named without a file extension
+// We should remove this restriction
```
dom96, line 1150: Already fixed in https://github.com/cloudflare/workerd/pull/1597, we'll just need to make a new release and update to it.

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -49,16 +50,41 @@ export async function findAdditionalModules(
...
+			for (const requirement of pythonRequirements.split("\n")) {
+				modules.push({
+					type: "python-requirement",
+					name: path.resolve(entry.moduleRoot, requirement),
+					content: "",
+					filePath: undefined,
+				});
+			}
+			// We don't care if a requirements.txt isn't found
+		} catch (e) {}
```
dom96, line 74: Should we log the error as a warning? or alternatively be more explicit about which exceptions we are ignoring?

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -49,16 +50,41 @@ export async function findAdditionalModules(
...
+			const pythonRequirements = await readFile(
+				path.resolve(entry.directory, "requirements.txt"),
+				"utf-8"
+			);
+
+			// This is incredibly naive. However, it supports common syntax for requirements.txt
+			for (const requirement of pythonRequirements.split("\n")) {
+				modules.push({
+					type: "python-requirement",
+					name: path.resolve(entry.moduleRoot, requirement),
```
dom96, line 68: I guess this results in `/path/to/module/root/bcrypt==4.0.1`. Can you strip the operator and version from the requirements? The absolute path shouldn't be necessary either.

We probably just want to parse as long as we see a valid identifier (something like `^[^\d\W]\w*` should work). This should grab the package name even if it contains dashes and avoid reading in any operators and version numbers.

dom96 (commented):
Nice! Thank you for helping with this.

I re-tested with my test plan and ran into issues:

```
$ cat $PWD/index.py
from js import Response
def fetch(request):
  return Response.new('hello world')
$ pnpm run --filter wrangler start dev $PWD/index.py --compatibility-flag experimental
...
⎔ Starting local server...
✘ [ERROR] Uncaught (in promise) ErrnoError


✘ [ERROR] Uncaught (in response) Error: ErrnoError: FS error


[wrangler:inf] GET / 500 Internal Server Error (3108ms)
✘ [ERROR] Uncaught (in promise) ErrnoError


✘ [ERROR] Uncaught (in response) Error: ErrnoError: FS error


[wrangler:inf] GET /favicon.ico 500 Internal Server Error (111ms)
```

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -49,16 +50,41 @@ export async function findAdditionalModules(
...
 		logger.table(
-			modules.map(({ name, type, content }) => {
-				return {
-					Name: name,
-					Type: type ?? "",
-					Size: `${(content.length / 1024).toFixed(2)} KiB`,
-				};
-			})
+			modules
+				.filter((m) => m.type !== "python-requirement")
```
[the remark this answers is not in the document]
penalosa: In this case it's meant to log out the additional modules attached to your worker:
Screenshot 2024-02-02 at 13 45 28
I'm not sure it makes sense to list python requirements here, but maybe it does?

penalosa:
> pnpm run --filter wrangler start dev $PWD/index.py --compatibility-flag experimental

@dom96 where are you running this? Is `index.py` in the root of the Wrangler repo?

dom96:
@penalosa it's in the root of the `workers-sdk` repo

penalosa:
> @penalosa it's in the root of the `workers-sdk` repo

Ah, that'll probably be it then. What happens if you try to run the fixture in `fixtures/python-worker` or add an `index.py` file to a new folder?

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
penalosa: Do we want to strip version specifiers? Pyodide seems to throw a relatively helpful error if you specify a version that isn't supported. e.g. `bcrypt==4.0.0` gives `ValueError: Can't find a pure Python 3 wheel for 'bcrypt==4.0.0'.`

dom96:
@penalosa ahh, that works. Thanks!

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
penalosa: Ahhh, this doesn't wokr in production, so I'll strip these

dom96 (approved):
Accepting to unblock, but it seems tests are still failing in CI

fixtures/python-worker/src/arith.py
mrbbot, line 1: There are some missing newlines on files in the `src` directory.

packages/wrangler/e2e/dev.test.ts
```diff
@@ -240,6 +240,77 @@ describe("basic dev tests", () => {
...
+describe("basic dev python tests", () => {
+	let worker: DevWorker;
+
+	beforeEach(async () => {
+		worker = await makeWorker();
+		await worker.seed((workerName) => ({
+			"wrangler.toml": dedent`
+					name = "${workerName}"
+					main = "index.py"
+					compatibility_date = "2023-01-01"
```
mrbbot, line 252: Does this not need `compatibility_flags = ["experimental"]`?

packages/wrangler/e2e/dev.test.ts
```diff
@@ -240,6 +240,77 @@ describe("basic dev tests", () => {
...
+				`,
+			}));
+			const { text: text3 } = await retry(
+				(s) => s.status !== 200 || s.text === "Updated Python Worker value",
+				async () => {
+					const r = await fetch(`http://127.0.0.1:${session.port}`);
+					return { text: await r.text(), status: r.status };
+				}
+			);
+			expect(text3).toMatchInlineSnapshot('"Updated Python Worker value"');
```
mrbbot, line 309: Is this part of the test needed? It seems to be using the same `wrangler.toml` file? Did you mean to change some other binding settings, or is this verifying that restarting with the same scripts also work?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -8775,6 +8775,72 @@ export default{
...
+			);
+			mockSubDomainRequest();
+			mockUploadWorkerRequest({
+				expectedMainModule: "index.py",
+			});
+
+			await runWrangler("deploy");
+			expect(
+				std.out.replace(
+					/.wrangler\/tmp\/deploy-(.+)\/index.py/,
```
mrbbot, line 8795: Would this test work if run from Windows? I'm assuming yes, since these are the module names, not the on disk paths?

packages/wrangler/src/config/index.ts
```diff
@@ -57,6 +57,17 @@ export function readConfig<CommandArgs>(
 		throw new UserError(diagnostics.renderErrors());
 	}
 
+	const mainModule = ("script" in args ? args.script : config.main) as string;
+	if (typeof mainModule === "string" && mainModule.endsWith(".py")) {
+		// Workers with a python entrypoint should have bundling turned off, since all of Wrangler's bundling is JS/TS specific
+		config.no_bundle = true;
+
+		// Workers with a python entrypoint need module rules for "*.py". Add one automatically as a DX nicety
+		if (!config.rules.find((rule) => rule.type === "PythonModule")) {
```
mrbbot, line 66: Nit:
```suggestion
		if (!config.rules.some((rule) => rule.type === "PythonModule")) {
```

packages/wrangler/src/config/index.ts
```diff
@@ -57,6 +57,17 @@ export function readConfig<CommandArgs>(
 		throw new UserError(diagnostics.renderErrors());
 	}
 
+	const mainModule = ("script" in args ? args.script : config.main) as string;
```
mrbbot, line 60: You shouldn't need the `as string` here, as you're doing a runtime `typeof` check below before using it.

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -49,14 +50,47 @@ export async function findAdditionalModules(
...
+	if (isPythonEntrypoint) {
+		try {
+			const pythonRequirements = await readFile(
+				path.resolve(entry.directory, "requirements.txt"),
+				"utf-8"
+			);
+
+			// This is incredibly naive. However, it supports common syntax for requirements.txt
+			for (const requirement of pythonRequirements.split("\n")) {
+				const packageName = requirement.match(/^[^\d\W]\w*/);
```
mrbbot, line 66: What would be the difference between this RegExp and `/^\w+/` here?

packages/wrangler/src/deployment-bundle/guess-worker-format.ts
```diff
@@ -20,6 +20,17 @@ export default async function guessWorkerFormat(
 	hint: CfScriptFormat | undefined,
 	tsconfig?: string | undefined
 ): Promise<CfScriptFormat> {
+	const parsedEntryPath = path.parse(entryFile);
+	if (parsedEntryPath.ext == ".py") {
```
mrbbot, line 24: Would `entryFile.endsWith(".py")` not be sufficient here?

mrbbot (approved):
Nice! 🐍 Added some comments, but LGTM! ✅

packages/wrangler/src/config/index.ts
```diff
@@ -57,6 +57,17 @@ export function readConfig<CommandArgs>(
...
+	if (typeof mainModule === "string" && mainModule.endsWith(".py")) {
+		// Workers with a python entrypoint should have bundling turned off, since all of Wrangler's bundling is JS/TS specific
+		config.no_bundle = true;
+
+		// Workers with a python entrypoint need module rules for "*.py". Add one automatically as a DX nicety
+		if (!config.rules.find((rule) => rule.type === "PythonModule")) {
+			config.rules.push({ type: "PythonModule", globs: ["**/*.py"] });
+		}
```
RamIdeas, line 68: Do we need to add a rule for requirements.txt or is that only used locally?

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
RamIdeas: It won't allow the requirement to begin with a number whereas `/^\w+/` would

packages/wrangler/e2e/dev.test.ts
RamIdeas: Is this a copy paste error? If not, worth to add some comments to clarify what exactly you're testing

RamIdeas (approved):

packages/wrangler/src/config/index.ts
penalosa: It can't really leverage the rules system in the same way as .py modules (since it's one file that has to be parsed and then turned into multiple modules)

packages/wrangler/src/deployment-bundle/guess-worker-format.ts
dom96: Unlikely to matter here, but `path.parse` is more portable and future proof

PR #4902, opened 2024-02-02
[wrangler] Use `esbuild-plugins-node-modules-polyfill` for Node.js polyfills instead of `@esbuild-plugins`

rnmeow:

Removes dependency `@esbuild-plugins/node-globals-polyfill` and `@esbuild-plugins/node-modules-polyfill` since they're both out-of-date.

Adds dependency `esbuild-plugin-polyfill-node` for support of Node.js polyfills.

This is probably not needed to test, without breaking changes, I think.

- Tests
  - [ ] Included
  - [x] Not necessary because: Unrelated.
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: Unrelated.

### DIFF
--- a/.changeset/rare-bananas-add.md
+++ b/.changeset/rare-bananas-add.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Use `esbuild-plugin-polyfill-node` for Node.js polyfill instead of `@esbuild-plugins` pack.
--- a/packages/wrangler/package.json
+++ b/packages/wrangler/package.json
@@ -102,11 +102,10 @@
 	},
 	"dependencies": {
 		"@cloudflare/kv-asset-handler": "^0.2.0",
-		"@esbuild-plugins/node-globals-polyfill": "^0.2.3",
-		"@esbuild-plugins/node-modules-polyfill": "^0.2.2",
 		"blake3-wasm": "^2.1.5",
 		"chokidar": "^3.5.3",
 		"esbuild": "0.17.19",
+		"esbuild-plugins-node-modules-polyfill": "^1.6.2",
 		"miniflare": "workspace:*",
 		"nanoid": "^3.3.3",
 		"path-to-regexp": "^6.2.0",
--- a/packages/wrangler/scripts/deps.ts
+++ b/packages/wrangler/scripts/deps.ts
@@ -12,8 +12,7 @@ export const EXTERNAL_DEPENDENCIES = [
 	// todo - bundle miniflare too
 	"selfsigned",
 	"source-map",
-	"@esbuild-plugins/node-globals-polyfill",
-	"@esbuild-plugins/node-modules-polyfill",
+	"esbuild-plugins-node-modules-polyfill",
 	"chokidar",
 ];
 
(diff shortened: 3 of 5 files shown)


.changeset/rare-bananas-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
petebacondarwin, line 2: Is there any reason why this should not just be a patch version bump?

```suggestion
"wrangler": patch
```

.changeset/rare-bananas-add.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Use `esbuild-plugin-polyfill-node` for Node.js polyfill instead of `@esbuild-plugins` pack.
```
petebacondarwin, line 5: Feels like we are fixing the broken aspects of the unsupported package?

```suggestion
fix: Use `esbuild-plugin-polyfill-node` for Node.js polyfill instead of `@esbuild-plugins` pack.
```

petebacondarwin (changes requested):
Let me see if I can get the CI checks running on this...

petebacondarwin:
I can't push a fixup commit to your PR as you did not allow contributors to update your branch.
But you need to fix the formatting with this diff:

```diff
diff --git a/packages/wrangler/src/deployment-bundle/bundle.ts b/packages/wrangler/src/deployment-bundle/bundle.ts
index dab86a06..cd7b0db9 100644
--- a/packages/wrangler/src/deployment-bundle/bundle.ts
+++ b/packages/wrangler/src/deployment-bundle/bundle.ts
@@ -325,12 +325,14 @@ export async function bundleWorker(
                plugins: [
                        moduleCollector.plugin,
                        ...(legacyNodeCompat
-                               ? [nodeModulesPolyfillPlugin({
-                                               globals: {
-                                                       process: true,
-                                                       Buffer: true,
-                                               },
-                                       })]
+                               ? [
+                                               nodeModulesPolyfillPlugin({
+                                                       globals: {
+                                                               process: true,
+                                                               Buffer: true,
+                                                       },
+                                               }),
+                                 ]
                                : []),
                        ...(nodejsCompat ? [nodejsCompatPlugin] : []),
                        cloudflareInternalPlugin,
```

You can fix this via the command line by running `pnpm fix` in the root of the monorepo.

.changeset/rare-bananas-add.md
rnmeow: The packages work, for now, actually. But they are out of date.

petebacondarwin:
I'm a bit concerned about these issues with the proposed package:

- https://github.com/cyco130/esbuild-plugin-polyfill-node/issues/8
- https://github.com/cyco130/esbuild-plugin-polyfill-node/issues/13
- https://github.com/cyco130/esbuild-plugin-polyfill-node/issues/11

I think we need to check that these do not impair Wrangler before we can merge this.

rnmeow:
> I'm a bit concerned about these issues with the proposed package:
> ... 
> I think we need to check that these do not impair Wrangler before we can merge this.

At a glance, the problems seem to appear at Vite-side and old Node.js runtimes. However, to examine a little bit more will be safer (especially for the users).

PR #4904, opened 2024-02-02
fix: return 4xx error from playground preview worker on invalid tokens

mrbbot:

Previously, we were passing user-controllable token inputs (`X-CF-Token` header, `user` search param, `token` cookie) directly to the `DurableObjectNamespace#idFromString()` function. This validated the token to be a valid Durable Object ID, but threw a reportable `TypeError`, generating a 500 response and increasing the errored request count. This is a user error, so shouldn't be reported. This change wraps calls to `idFromString()` inside `try`/`catch`es, and re-throws non-reportable user errors that generate 4xx responses. We could just validate that the tokens were 64 hex digits before calling `idFromString()`, but this approach ensures the IDs are for the `UserSession` object by validating the signature encoded in the ID too.

I've added some new tests for this behaviour. They should be run as this PR has the required label.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [x] Not necessary because: updating internal worker
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: updating internal worker

**Note for PR author:**

### DIFF
--- a/packages/playground-preview-worker/src/index.ts
+++ b/packages/playground-preview-worker/src/index.ts
@@ -98,21 +98,21 @@ export class PreviewError extends HttpError {
 class TokenUpdateFailed extends HttpError {
 	name = "TokenUpdateFailed";
 	constructor() {
-		super("Provide token", 400, false);
+		super("Provide valid token", 400, false);
 	}
 }
 
 class RawHttpFailed extends HttpError {
 	name = "RawHttpFailed";
 	constructor() {
-		super("Provide token", 400, false);
+		super("Provide valid token", 400, false);
 	}
 }
 
 class PreviewRequestFailed extends HttpError {
 	name = "PreviewRequestFailed";
 	constructor(private tokenId: string | undefined, reportable: boolean) {
-		super("Token not found", 400, reportable);
+		super("Valid token not found", 400, reportable);
 	}
 	get data(): { tokenId: string | undefined } {
 		return { tokenId: this.tokenId };
@@ -122,7 +122,7 @@ class PreviewRequestFailed extends HttpError {
 class UploadFailed extends HttpError {
 	name = "UploadFailed";
 	constructor() {
-		super("Token not provided", 401, false);
+		super("Valid token not provided", 401, false);
 	}
 }
 
@@ -144,8 +144,13 @@ async function handleRawHttp(request: Request, url: URL, env: Env) {
 	if (!token) {
 		throw new RawHttpFailed();
 	}
-
-	const userObject = env.UserSession.get(env.UserSession.idFromString(token));
+	let userObjectId: DurableObjectId;
+	try {
+		userObjectId = env.UserSession.idFromString(token);
+	} catch {
+		throw new RawHttpFailed();
+	}
+	const userObject = env.UserSession.get(userObjectId);
 
 	// Delete these consumed headers so as not to bloat the request.
 	// Some tokens can be quite large and may cause nginx to reject the
@@ -250,14 +255,16 @@ app.get(`${rootDomain}/`, async (c) => {
 
 app.post(`${rootDomain}/api/worker`, async (c) => {
 	let userId = getCookie(c, "user");
-
 	if (!userId) {
 		throw new UploadFailed();
 	}
-
-	const userObject = c.env.UserSession.get(
-		c.env.UserSession.idFromString(userId)
-	);
+	let userObjectId: DurableObjectId;
+	try {
+		userObjectId = c.env.UserSession.idFromString(userId);
+	} catch {
+		throw new UploadFailed();
+	}
+	const userObject = c.env.UserSession.get(userObjectId);
 
 	return userObject.fetch("https://example.com", {
 		body: c.req.body,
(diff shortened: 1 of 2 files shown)


packages/playground-preview-worker/src/index.ts
```diff
@@ -98,21 +98,21 @@ export class PreviewError extends HttpError {
...
-		super("Provide token", 400, false);
+		super("Provide valid token", 400, false);
 	}
 }
 
 class RawHttpFailed extends HttpError {
 	name = "RawHttpFailed";
 	constructor() {
-		super("Provide token", 400, false);
+		super("Provide valid token", 400, false);
```
petebacondarwin, line 108: Is it worth specifying the header name that is expected? E.g.

```suggestion
		super("Provide valid `X-CF-Token` header", 400, false);
```

packages/playground-preview-worker/src/index.ts
```diff
@@ -122,7 +122,7 @@ class PreviewRequestFailed extends HttpError {
 class UploadFailed extends HttpError {
 	name = "UploadFailed";
 	constructor() {
-		super("Token not provided", 401, false);
+		super("Valid token not provided", 401, false);
```
petebacondarwin, line 125: Similarly...

```suggestion
		super("Valid `token` cookie not provided", 401, false);
```

packages/playground-preview-worker/tests/index.test.ts
```diff
@@ -45,7 +45,27 @@ describe("Preview Worker", () => {
 			'"/hello?world"'
 		);
 		expect(resp.headers.get("set-cookie") ?? "").toMatchInlineSnapshot(
-			'"token=TEST_TOKEN; Domain=random-data.playground-testing.devprod.cloudflare.dev; HttpOnly; Secure; SameSite=None"'
+			`"token=${defaultUserToken}; Domain=random-data.playground-testing.devprod.cloudflare.dev; HttpOnly; Secure; SameSite=None"`
+		);
+	});
+	it("shouldn't be redirected with invalid token", async () => {
```
petebacondarwin, line 51: NIT - sometimes there are empty lines between `it()` clauses and sometimes not. I feel we should be consistent.

packages/playground-preview-worker/tests/index.test.ts
```diff
@@ -110,6 +130,18 @@ describe("Preview Worker", () => {
 
 		expect(await resp.text()).toMatchInlineSnapshot('"407"');
 	});
+	it("should reject invalid token", async () => {
```
petebacondarwin, line 133: Perhaps also test for no token at all?

petebacondarwin (changes requested):
I don't see any tests for `X-CF-Token` header being invalid. Am I missing something?

packages/playground-preview-worker/src/index.ts
mrbbot: Maybe, these errors aren't meant to be user facing. And really they should never be shown unless a user is doing something they shouldn't.

packages/playground-preview-worker/tests/index.test.ts
mrbbot: I think the empty lines are ok here, they split out tests for different routes of the worker.

packages/playground-preview-worker/tests/index.test.ts
mrbbot: Added some test tests for this 👍

packages/playground-preview-worker/tests/index.test.ts
```diff
@@ -22,13 +22,50 @@ describe("Preview Worker", () => {
...
+		})
+	}
+}
+
+------WebKitFormBoundaryqJEYLXuUiiZQHgvf
+Content-Disposition: form-data; name="metadata"; filename="blob"
+Content-Type: application/json
+
+{"compatibility_date":"2023-05-04","main_module":"index.js"}
+------WebKitFormBoundaryqJEYLXuUiiZQHgvf--`,
```
petebacondarwin, line 62: 😍

packages/playground-preview-worker/tests/index.test.ts
```diff
@@ -153,6 +288,35 @@ describe("Upload Worker", () => {
...
+				"Content-Type":
+					"multipart/form-data; boundary=----WebKitFormBoundaryqJEYLXuUiiZQHgvf",
+			},
+			body: '------WebKitFormBoundaryqJEYLXuUiiZQHgvf\nContent-Disposition: form-data; name="index.js"; filename="index.js"\nContent-Type: application/javascript+module\n\nexport default {\n	fetch(request) {\n		const url = new URL(request.url)\n		if(url.pathname === "/exchange") {\n			return Response.json({\n				token: "TEST_TOKEN",\n				prewarm: "TEST_PREWARM"\n			})\n		}\n		if(url.pathname === "/redirect") {\n			return Response.redirect("https://example.com", 302)\n		}\n		if(url.pathname === "/method") {\n			return new Response(request.method)\n		}\n		if(url.pathname === "/status") {\n			return new Response(407)\n		}\n		if(url.pathname === "/header") {\n			return new Response(request.headers.get("X-Custom-Header"))\n		}\n		return Response.json({\n			url: request.url,\n			headers: [...request.headers.entries()]\n		})\n	}\n}\n\n------WebKitFormBoundaryqJEYLXuUiiZQHgvf\nContent-Disposition: form-data; name="metadata"; filename="blob"\nContent-Type: application/json\n\n{"compatibility_date":"2023-05-04","main_module":"index.js"}\n------WebKitFormBoundaryqJEYLXuUiiZQHgvf--',
```
petebacondarwin, line 298: What no lovely formatted code like above? 😄

petebacondarwin (approved):

packages/playground-preview-worker/tests/index.test.ts
mrbbot: Fixed 😄

Issue #4875, opened 2024-01-30
🚀 Feature Request [getBindingsProxy]: `getBindingsProxy` should expose a `cf` object


dario-piotrowicz:
In the `getBindingsProxy` return object we should include a [`cf` object](https://developers.cloudflare.com/workers/runtime-apis/request/#incomingrequestcfproperties) that developers can use to mock the real production thing.

It should reuse the same functionality Miniflare already implements: https://github.com/cloudflare/workers-sdk/blob/6a4cb8c6456f1dba95cae2b8bbe658f7227349f8/packages/miniflare/src/cf.ts

For that we either:
- need to expose the `setupCf` function: https://github.com/cloudflare/workers-sdk/blob/6a4cb8c6456f1dba95cae2b8bbe658f7227349f8/packages/miniflare/src/cf.ts#L70
So that we can all it from wrangler/`getBindingsProxy`
- or add a `cf` or `getCf` to the Miniflare class

I did think that the former could be better (since it doesn't force the caller to be tied to the Miniflare instance), but I changed my mind and actually think that the latter could be better (since we already do have a Miniflare instance and get the `cf` from it... also other Miniflare users could make use of such additional field/method)

(In any case either should do)

houmark:
Hey guys! Any timeline on this one?

dario-piotrowicz:
Hi @houmark, this should make it into one of the stable releases next week (I would be very very surprised if it doesn't) 🙂

For it we just need to merge: https://github.com/cloudflare/workers-sdk/pull/4905 and then simply expose the `cf` in `getBindingsProxy` 🙂

houmark:
Brilliant! Thanks so much!

dario-piotrowicz:
No problem 🙂👍

PR #4905, opened 2024-02-02
expose cf object from Miniflare instances

dario-piotrowicz:
prerequisite for #4875

### DIFF
--- a/.changeset/gentle-chairs-pull.md
+++ b/.changeset/gentle-chairs-pull.md
@@ -8,7 +8,7 @@ add a new `getCf` method attached to instances of `Miniflare`, this `getCf` retu
 the `cf` object that the Miniflare instance provides to the actual workers and it
 depends of the core option of the same name
 
-Basic example:
+Example:
 
 ```ts
 import { Miniflare } from "miniflare";
@@ -18,24 +18,4 @@ const mf = new Miniflare({ ... });
 const cf = await mf.getCf();
 
 console.log(`country = ${cf.country} ; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
-```
-
-Example with a custom `cf` option:
-
-```ts
-import { Miniflare } from "miniflare";
-
-const mf = new Miniflare({
-	script: "",
-	modules: true,
-	cf: {
-		helloW: "Hello World",
-	},
-});
-
-const cf = await mf.getCf();
-
-console.log(cf.helloW); // logs 'Hello World'
-
-await mf.dispose();
 ```
--- a/.changeset/gentle-chairs-pull.md
+++ b/.changeset/gentle-chairs-pull.md
@@ -1,0 +1,46 @@
+---
+"miniflare": minor
+---
+
+feature: add a `getCf` method to Miniflare instances
+
+add a new `getCf` method attached to instances of `Miniflare`, this `getCf` returns
+the `cf` object that the Miniflare instance provides to the actual workers and it
+depends of the core option of the same name
+
+Basic example:
+
+```ts
+import { Miniflare } from "miniflare";
+
+const mf = new Miniflare({
+	script: "",
+	modules: true,
+});
+
+const cf = await mf.getCf();
+
+console.log(`country = ${cf.country} ; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
+
+await mf.dispose();
+```
+
+Example with a custom `cf` option:
+
+```ts
+import { Miniflare } from "miniflare";
+
+const mf = new Miniflare({
+	script: "",
+	modules: true,
+	cf: {
+		helloW: "Hello World",
+	},
+});
+
+const cf = await mf.getCf();
+
+console.log(cf.helloW); // logs 'Hello World'
+
+await mf.dispose();
+```
--- a/.changeset/gentle-chairs-pull.md
+++ b/.changeset/gentle-chairs-pull.md
@@ -13,16 +13,11 @@ Basic example:
 ```ts
 import { Miniflare } from "miniflare";
 
-const mf = new Miniflare({
-	script: "",
-	modules: true,
-});
+const mf = new Miniflare({ ... });
 
 const cf = await mf.getCf();
 
 console.log(`country = ${cf.country} ; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
-
-await mf.dispose();
 ```
 
 Example with a custom `cf` option:
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -771,8 +771,6 @@ defined at the top-level.
   function will invalidate any values returned by the `Miniflare#get*()`
   methods, preventing them from being used.
 
-- `getCf: Promise<Record<string, any>>`
+- `getCf(): Promise<Record<string, any>>`
 
-  Returns the same object returned from incoming `Request`'s `cf` property.
-
-  This object depends on the core options property of the same name.
+  Returns the same object returned from incoming `Request`'s `cf` property. This object depends on the `cf` property from `SharedOptions`.
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -406,7 +406,7 @@ parameter in module format Workers.
             await this.STORE.fetch(this.baseURL + key, { method: "DELETE" });
           }
         }
-    
+
         // env has the type { STORE: Fetcher, NAMESPACE?: string }
         export default function (env) {
           return new MiniKV(env);
@@ -770,3 +770,9 @@ defined at the top-level.
   `Miniflare#dispatchFetch()` cannot be called. Additionally, calling this
   function will invalidate any values returned by the `Miniflare#get*()`
   methods, preventing them from being used.
+
+- `getCf: Promise<Record<string, any>>`
+
+  Returns the same object returned from incoming `Request`'s `cf` property.
+
+  This object depends on the core options property of the same name.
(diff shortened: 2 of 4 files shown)


.changeset/gentle-chairs-pull.md
```diff
@@ -0,0 +1,46 @@
...
+		helloW: "Hello World",
+	},
+});
+
+const cf = await mf.getCf();
+
+console.log(cf.helloW); // logs 'Hello World'
+
+await mf.dispose();
+```
```
mrbbot, line 46: I think we can remove this second example. The first demonstrates the feature nicely.

.changeset/gentle-chairs-pull.md
```diff
@@ -0,0 +1,46 @@
...
+
+Basic example:
+
+```ts
+import { Miniflare } from "miniflare";
+
+const mf = new Miniflare({
+	script: "",
+	modules: true,
+});
```
mrbbot, line 19: These examples needn't be copy-pastable. To avoid making the changelog too verbose, I'd simplify things a little.
```suggestion
const mf = new Miniflare({ ... });
```

.changeset/gentle-chairs-pull.md
```diff
@@ -0,0 +1,46 @@
...
+
+const cf = await mf.getCf();
+
+console.log(`country = ${cf.country} ; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
+
+await mf.dispose();
```
mrbbot, line 25:
```suggestion
```

packages/miniflare/README.md
```diff
@@ -770,3 +770,9 @@ defined at the top-level.
   `Miniflare#dispatchFetch()` cannot be called. Additionally, calling this
   function will invalidate any values returned by the `Miniflare#get*()`
   methods, preventing them from being used.
+
+- `getCf: Promise<Record<string, any>>`
+
+  Returns the same object returned from incoming `Request`'s `cf` property.
+
+  This object depends on the core options property of the same name.
```
mrbbot, line 778: Maybe reference the type where this is defined instead? I'd also avoid splitting this onto multiple lines to match the style of the other methods.
```suggestion
- `getCf(): Promise<Record<string, any>>`

  Returns the same object returned from incoming `Request`'s `cf` property. This object depends on the `cf` property from `SharedOptions`.
```

packages/miniflare/test/index.spec.ts
```diff
@@ -1736,3 +1737,21 @@ test("Miniflare: prohibits invalid wrapped bindings", async (t) => {
 		}
 	);
 });
+
+test("Miniflare: getCf() returns a standard cf object", async (t) => {
+	const mf = new Miniflare({ script: "", modules: true });
+	await mf.ready;
+	const cf = await mf.getCf();
+	t.snapshot(cf);
+	t.deepEqual(cf, JSON.parse(JSON.stringify(fallbackCf)));
```
mrbbot: I've tried not to import things from `src` files directly in tests like this. I'd probably use `t.like()` instead and select a subset of `cf` options to check. We don't need to compare the whole object here, just that it has some of the fallback values.
```suggestion
	t.like(cf, {
	  colo: "DFW",
	  city: "Austin",
	  regionCode: "TX",
	});
```

packages/miniflare/test/index.spec.ts
```diff
@@ -1736,3 +1737,21 @@ test("Miniflare: prohibits invalid wrapped bindings", async (t) => {
...
+});
+
+test("Miniflare: getCf() returns a user provided cf object", async (t) => {
+	const mf = new Miniflare({ script: "", modules: true, cf: {
+		myFakeField: 'test'
+	} });
+	await mf.ready;
+	const cf = await mf.getCf();
+ t.snapshot(cf);
```
mrbbot:
```suggestion
```

packages/miniflare/test/index.spec.ts
```diff
@@ -1736,3 +1737,21 @@ test("Miniflare: prohibits invalid wrapped bindings", async (t) => {
...
+test("Miniflare: getCf() returns a standard cf object", async (t) => {
+	const mf = new Miniflare({ script: "", modules: true });
+	await mf.ready;
+	const cf = await mf.getCf();
+	t.snapshot(cf);
+	t.deepEqual(cf, JSON.parse(JSON.stringify(fallbackCf)));
+});
+
+test("Miniflare: getCf() returns a user provided cf object", async (t) => {
+	const mf = new Miniflare({ script: "", modules: true, cf: {
```
mrbbot, line 1749: It looks like you might need to format this file. You should be able to run `pnpm lint:fix` in `packages/miniflare` to fix this.

packages/miniflare/test/index.spec.ts
```diff
@@ -1736,3 +1737,21 @@ test("Miniflare: prohibits invalid wrapped bindings", async (t) => {
 		}
 	);
 });
+
+test("Miniflare: getCf() returns a standard cf object", async (t) => {
+	const mf = new Miniflare({ script: "", modules: true });
+	await mf.ready;
```
mrbbot, line 1743: `getCf()` should implicitly `await mf.ready`.
```suggestion
```

packages/miniflare/test/index.spec.ts
```diff
@@ -1736,3 +1737,21 @@ test("Miniflare: prohibits invalid wrapped bindings", async (t) => {
...
+	const cf = await mf.getCf();
+	t.snapshot(cf);
+	t.deepEqual(cf, JSON.parse(JSON.stringify(fallbackCf)));
+});
+
+test("Miniflare: getCf() returns a user provided cf object", async (t) => {
+	const mf = new Miniflare({ script: "", modules: true, cf: {
+		myFakeField: 'test'
+	} });
+	await mf.ready;
```
mrbbot, line 1752: `getCf()` should implicitly `await mf.ready`.
```suggestion
```

mrbbot (commented):
Nice! 🎉 Added some comments, thanks for updating the docs too! 👍

.changeset/gentle-chairs-pull.md
dario-piotrowicz: ok 🙂👍

.changeset/gentle-chairs-pull.md
dario-piotrowicz: Sure, I agree that it doesn't really add much 🙂👍

packages/miniflare/README.md
dario-piotrowicz: 🙏

packages/miniflare/test/index.spec.ts
dario-piotrowicz: of course, sorry my bad, left over after refactoring 👍

packages/miniflare/test/index.spec.ts
dario-piotrowicz: Are you sure? 😕 `pnpm lint:fix` doesn't seem to do anything here

Screenshot 2024-02-02 at 18 34 24

packages/miniflare/test/index.spec.ts
dario-piotrowicz: sure sounds good to me 🙂👍

packages/miniflare/test/index.spec.ts
dario-piotrowicz: thanks, I committed that by mistake 😅

packages/miniflare/test/index.spec.ts
dario-piotrowicz: I feel like something is off with the formatting, I could not get prettier to apply to this file

As I mentioned `pnpm lint:fix` didn't seem to do anything

 I ended up running (in the `miniflare` directory):
```sh
pnpm prettier ./test/index.spec.ts --write
```

(PS: `pnpm prettier .` would change lots of files 🤷)

Hopefully this is fine, please let me know 🙂

packages/miniflare/test/index.spec.ts
```diff
@@ -1736,3 +1735,27 @@ test("Miniflare: prohibits invalid wrapped bindings", async (t) => {
 		}
 	);
 });
+
+test("Miniflare: getCf() returns a standard cf object", async (t) => {
+	const mf = new Miniflare({ script: "", modules: true });
```
mrbbot, line 1740: Rather than calling `dispose()` at the end of the test, I'd use `t.teardown()` to register a teardown hook that gets executed regardless of the test result. This will ensure if any assertions fail, the `Miniflare` instance is still disposed.

```suggestion
	const mf = new Miniflare({ script: "", modules: true });
	t.teardown(() => mf.dispose());
```

Same below. You should then be able to remove the `await mf.dispose()` calls.

packages/miniflare/test/index.spec.ts
mrbbot: Sounds good! Thanks! 👍

packages/miniflare/test/index.spec.ts
dario-piotrowicz: ah! for sure! 😮 

Thanks! 😄

dario-piotrowicz: Sorry I missed the pattern, I can see that's used in all the other tests in this file 😓

dario-piotrowicz: should be all good now 🙂👍

mrbbot (approved):
Awesome, thank you! ✅

PR #4458, opened 2023-11-16
Pack miniflare for e2e

penalosa:
Fixes e2e tests on Changesets PR

**What this PR solves / how to test:**

Previously, e2e tests on the changesets PR were trying to fetch an unreleased version of Miniflare from the npm registry. This ensure that they fetch the correct packed version of Miniflare.

**Author has addressed the following:**

- Tests
  - [ ] Included
  - [x] Not necessary because: This is a fix for existing tests
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] Included
  - [x] Not necessary because: Not user facing
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: Not user facing

**Note for PR author:**

We want to celebrate and highlight awesome PR review! If you think this PR received a particularly high-caliber review, please assign it the label `highlight pr review` so future reviewers can take inspiration and learn from it.

PR #4907, opened 2024-02-02
[wrangler] test: fix E2E tests

mrbbot:

#4458 accidentally left a `continue-on-error: true` in our E2E testing GitHub actions workflow. This meant E2E test failures weren't reported as check failures. This PR fixes that, and updates the E2E tests so they pass with the changes that have been added since then.

It includes a couple of user-facing changes for issues that these tests would've caught. See commit descriptions for more details on the changes.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: fixing regressions

**Note for PR author:**


RamIdeas (approved):

packages/wrangler/e2e/deploy.test.ts
petebacondarwin, line 1: Why delete this whole file?

packages/wrangler/e2e/dev.test.ts
```diff
@@ -46,6 +50,20 @@ interface SessionData {
 	stderr: string;
 }
 
+function getPort() {
```
petebacondarwin, line 53: Why not use the `get-port` package for this?

petebacondarwin (approved):

penalosa (approved):

petebacondarwin:
Screenshot 2024-02-02 at 15 23 58

petebacondarwin:
Seems like the E2E tests are not actually fixed?

packages/wrangler/src/dev.tsx
petebacondarwin, line 1: cc @Lekensteyn 

We had to fix this because checking via `localhost` for port 9229 could say that it was available twice - once for `127.0.0.1` and again for `::1`, which mean that we ended up with two Wrangler sessions that both tried to start workerd on `127.0.0.1:9229`; the second of which would fail.

PR #4909, opened 2024-02-02
[D1] teach wrangler how to fetch insights about D1's queries

rozenmd:
Experimental feature, WIP.

### DIFF
--- a/.changeset/red-icons-flow.md
+++ b/.changeset/red-icons-flow.md
@@ -11,11 +11,11 @@ This command defaults to fetching the top 5 queries that took the longest to run
 You can also fetch the top 5 queries that consumed the most rows read over the last week, for example:
 
 ```bash
-npx wrangler d1 insights northwind --sortBy reads --last 7
+npx wrangler d1 insights northwind --sortBy reads --timePeriod 7
 ```
 
 Or the top 5 queries that consumed the most rows written over the last month, for example:
 
 ```bash
-npx wrangler d1 insights northwind --sortBy writes --last 31
+npx wrangler d1 insights northwind --sortBy writes --timePeriod 31
 ```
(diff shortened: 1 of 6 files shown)


elithrar (approved):

packages/wrangler/src/d1/insights.ts
```diff
@@ -0,0 +1,175 @@
...
+							},
+						},
+					}),
+					headers: {
+						"Content-Type": "application/json",
+					},
+				});
+
+			if (graphqlQueriesResult) {
+				graphqlQueriesResult.data?.viewer?.accounts[0]?.d1QueriesAdaptiveGroups?.forEach(
```
petebacondarwin, line 147:
```suggestion
				graphqlQueriesResult?.data?.viewer?.accounts[0]?.d1QueriesAdaptiveGroups?.forEach(
```

packages/wrangler/src/d1/insights.ts
```diff
@@ -0,0 +1,175 @@
...
+						});
+					}
+				);
+			}
+		}
+
+		if (json) {
+			logger.log(JSON.stringify(output, null, 2));
+		} else {
+			await printWranglerBanner();
```
petebacondarwin, line 168: Should we also be calling `d1BetaWarning()` here too?

(Actually looking at the rest of the code, `d1BetaWarning()` is currently only used in the epilogue... But I wonder if it would be worth reiterating that warning on each function?)

.changeset/red-icons-flow.md
```diff
@@ -0,0 +1,27 @@
+---
+"wrangler": patch
+---
+
+experimental-feat: add an `insights` command to `wrangler d1`
```
petebacondarwin, line 5: `experimental-feat` is not strictly one of the types for our changeset format.

```suggestion
feat: add an experimental `insights` command to `wrangler d1`
```

petebacondarwin (approved):

PR #4911, opened 2024-02-02
fix: don't report invalid `format-errors` input

mrbbot:

Previously, invalid input to the `format-errors` worker was reported as an internal server error. As with #4904, this is a user error that the service can't do anything about. This PR 

- Tests
  - [ ] Included
  - [x] Not necessary because: `format-errors` currently has no tests, we can add some once we've landed the new Vitest pool work.
  - [ ] Included
  - [x] Not necessary because: change to an internal worker
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: change to an internal worker

**Note for PR author:**

### DIFF
--- a/packages/format-errors/src/index.ts
+++ b/packages/format-errors/src/index.ts
@@ -155,8 +155,16 @@ export default {
 				},
 			},
 		});
+
+		// Validate payload outside of Sentry/metrics reporting
+		let payload: Payload;
 		try {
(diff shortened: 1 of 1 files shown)


petebacondarwin (approved):
I feel that this Worker is getting complex enough that it deserves some tests, soonish.

PR #4916, opened 2024-02-03
In pages functions avoid cloning the request when not necessary

### LINKED ISSUE #3259: 🐛 BUG: Your worker called response.clone(), but did not read the body of both clones

Wrangler

3.0.0

Windows

### Describe the Bug

After upgrading to 3.0.0 new message started appearing in the console:
```
Your worker called response.clone(), but did not read the body of both clones. This is wasteful, as it forces the system to buffer the entire response body in memory, rather than streaming it through. This may cause your worker to be unexpectedly terminated for going over the memory limit. If you only meant to copy the response headers and metadata (e.g. in order to be able to modify them), use `new Response(response.body, response)` instead.  
```
However, it does not say which exactly part of code is to blame.
I scanned my code and I do not call `response.clone()` directly anywhere, also, I have multiple levels of functions, making it hard to pin-point where the issue is and whether the issue is there at all.

dario-piotrowicz:
Partially fixes #3259 (the full fix can only be applied as a breaking change so it'll have to wait for a wrangler major release unfortunately)

Supersedes #4861 

___

Example for the changes: https://github.com/dario-piotrowicz/cf-pages-functions-request-clone-repro

___

TODO:
 - [ ] understand why this only occurred in POST requests
 - [ ] investigate alternative suggested in [comment](https://github.com/cloudflare/workers-sdk/pull/4916#issuecomment-1925948244)
 - [ ] add tests for plugins

### DIFF
--- a/.changeset/lucky-fans-repeat.md
+++ b/.changeset/lucky-fans-repeat.md
@@ -1,0 +1,15 @@
+---
+"wrangler": patch
+---
+
+fix: Remove `request.clone()` from Pages functions when possible
+
+When using the `/functions` directory requests get cloned unconditionally, such request bodies
+can not be read causing workerd errors regarding unused streams and wasted memory.
+
+These changes prevent `request.clone()` for routes containing a single handler, resolving the
+above issue for such simple type of routes.
+
+Routes with multiple handlers (with middlewares for example) sill generate clones unconditionally,
+such routes cannot be amended in a non breaking way, so they will be amended at a later time before
+a new major release of wrangler.
--- a/fixtures/pages-functions-request-clone/README.md
+++ b/fixtures/pages-functions-request-clone/README.md
@@ -1,0 +1,35 @@
+# Request Cloning
+
+This route + tests is for testing the `request.clone()` call that is currently causing workerd
+runtime errors: https://github.com/cloudflare/workers-sdk/issues/3259
+
+The cloning should be in the next wrangler major release, optimally alongside with this route + tests
+
+## Why can't we fix it in a patch release
+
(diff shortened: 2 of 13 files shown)


Skye-31:
I wonder if we could use [Request.bodyUsed](https://developer.mozilla.org/en-US/docs/Web/API/Request/bodyUsed) alongside [Request.body.cancel()](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/cancel) to remove the warning in all cases, by cancelling the body streams of any requests where its not been read?

russelldavis:
@Skye-31 that's an interesting idea. I think it would make the warnings go away (because the clones would be cleaned up right before the detection logic sees them), but unfortunately I think the underlying issue that is being warned about would remain. You'd still be calling `request.clone()` before each call to `next()`, and you'd only be able to safely call `request.body.cancel()` after the call to `next()` returns, which means the entire request would still get processed while clones of the stream exist, causing the problems mentioned in the warning message.

GregBrimble:
There's two things in one here:

1. We can minimize the number of `clone`s in a Pages Functions Worker. That can happen today. There shouldn't be any impact to users. We simply check the number of handlers. If `< 2`, then we don't need to clone. At a quick skim, that seems to be what the code of this PR does, and like I say, that could ship today because it's not a breaking change unless I'm misunderstanding something.

2. Foregoing automatic `clone`ing entirely. We can't do this, even with a major version bump, because Pages Functions is versioned independently (it currently is not versioned at all). This will need to wait until we figure out what we're doing with Pages Functions ownership long-term.

PR #4922, opened 2024-02-05
Add a `ctx` field to the `getBindingsProxy` result

### LINKED ISSUE #4876: 🚀 Feature Request [getBindingsProxy]: `getBindingsProxy` should expose a `ctx`/`context` object
### Describe the solution

In the `getBindingsProxy` return object we should include a `ctx`/`context` object with two methods:
[`waitUntil`](https://developers.cloudflare.com/workers/runtime-apis/handlers/fetch/#contextwaituntil) and [`passThroughOnException`](https://developers.cloudflare.com/workers/runtime-apis/handlers/fetch/#contextpassthroughonexception)

They should be both just simple no-op functions.

dario-piotrowicz:
Fixes #4876

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/issues/12840
  - [ ] Not necessary because:

**Note for PR author:**

### DIFF
--- a/fixtures/get-bindings-proxy/tests/get-bindings-proxy.ctx.test.ts
+++ b/fixtures/get-bindings-proxy/tests/get-bindings-proxy.ctx.test.ts
@@ -1,0 +1,55 @@
+import { describe, expect, it } from "vitest";
+import { getBindingsProxy } from "./shared";
+
+describe("getBindingsProxy - ctx", () => {
+	it("should provide a no-op waitUntil method", async () => {
+		const { ctx, dispose } = await getBindingsProxy();
+		try {
+			let value = 4;
+			ctx.waitUntil(
+				new Promise((resolve) => {
+					value++;
+					resolve(value);
+				})
+			);
+			expect(value).toBe(5);
+		} finally {
+			await dispose();
+		}
+	});
+
+	it("should provide a no-op passThroughOnException method", async () => {
+		const { ctx, dispose } = await getBindingsProxy();
+		try {
+			expect(ctx.passThroughOnException()).toBe(undefined);
+		} finally {
+			await dispose();
+		}
+	});
+
+	it("should match the production runtime ctx object", async () => {
+		const { ctx, dispose } = await getBindingsProxy();
+		try {
+			expect(ctx.constructor.name).toBe("ExecutionContext");
+			expect(typeof ctx.waitUntil).toBe("function");
+			expect(typeof ctx.passThroughOnException).toBe("function");
+
+			ctx.waitUntil = ((str: string) => `- ${str} -`) as any;
+			expect(ctx.waitUntil("waitUntil can be overridden" as any)).toBe(
+				"- waitUntil can be overridden -"
+			);
+
+			ctx.passThroughOnException = ((str: string) => `_ ${str} _`) as any;
+			expect(
+				(ctx.passThroughOnException as any)(
+					"passThroughOnException can be overridden"
+				)
+			).toBe("_ passThroughOnException can be overridden _");
+
+			(ctx as any).text = "the ExecutionContext can be extended";
+			expect((ctx as any).text).toBe("the ExecutionContext can be extended");
+		} finally {
+			await dispose();
+		}
+	});
+});
--- a/packages/wrangler/src/api/integrations/bindings/executionContext.ts
+++ b/packages/wrangler/src/api/integrations/bindings/executionContext.ts
@@ -1,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// waitUntil: (promise: Promise<any>) => void;
+// passThroughOnException: () => void;
+// };
+
+export class ExecutionContext {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any, unused-imports/no-unused-vars
+	waitUntil(promise: Promise<any>): void {}
+	passThroughOnException(): void {}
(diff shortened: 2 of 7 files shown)


packages/wrangler/src/api/integrations/bindings/executionContext.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// waitUntil: (promise: Promise<any>) => void;
+// passThroughOnException: () => void;
+// };
+
+export class ExecutionContext {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any, unused-imports/no-unused-vars
+	waitUntil(promise: Promise<any>): void {}
```
mrbbot, line 8: Switching to `unknown` here should avoid the need for `eslint` disables.
```suggestion
	waitUntil(promise: Promise<unknown>): void {}
```

packages/wrangler/src/api/integrations/bindings/executionContext.ts
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+// waitUntil: (promise: Promise<any>) => void;
+// passThroughOnException: () => void;
+// };
```
mrbbot, line 4:
```suggestion
```

packages/wrangler/src/api/integrations/bindings/index.ts
```diff
@@ -88,6 +93,7 @@ export async function getBindingsProxy<Bindings = Record<string, unknown>>(
 			...vars,
 			...bindings,
 		},
+		ctx: new ExecutionContext(),
```
mrbbot, line 96: `ctx` is slightly different than `caches` and `cf` in that in real workers, a new instance of `ExecutionContext` is created for each request. This doesn't really matter at the moment, since the implementation is just a no-op. I'm guessing there are no plans to implement a proper `ExecutionContext` here? If there are, you might want to consider an API that allows consumers to create `ExecutionContext`s on each request, but that's probably too much effort considering this is a stop-gap solution. 👍

fixtures/get-bindings-proxy/tests/get-bindings-proxy.ctx.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+			ctx.passThroughOnException = ((str: string) => `_ ${str} _`) as any;
+			expect(
+				(ctx.passThroughOnException as any)(
+					"passThroughOnException can be overridden"
+				)
+			).toBe("_ passThroughOnException can be overridden _");
+
+			(ctx as any).text = "the ExecutionContext can be extended";
+			expect((ctx as any).text).toBe("the ExecutionContext can be extended");
```
mrbbot, line 50: Just curious, is this something frameworks are doing? Or are you just checking this in case they decide to?

This is potentially one reason you might want to consider an `ExecutionContext` per request (see my other comment). In the real runtime, additions made to `ctx` won't be persisted across requests, but if you're reusing the same instance they would be.

packages/wrangler/src/api/integrations/bindings/executionContext.ts
dario-piotrowicz: 😓👍

packages/wrangler/src/api/integrations/bindings/executionContext.ts
dario-piotrowicz: yes, but I wanted to keep the type the exact same as the `@cloudflare/workers-types`....

not that it makes a huge difference either way though 😕 

what do you think is it better to have `unknown` and slightly diverge from workers-types?

packages/wrangler/src/api/integrations/bindings/index.ts
dario-piotrowicz: ah wow, ok that's a really interesting suggestion 🤔

For now I don't think there's much value in implementing the above or having the `ctx` non no-op tough

We can always add an API for that later on as an additive change 🙂👍

__

By the way, what do you mean by "a proper ExecutionContext here"? what should it do in a node dev server?

packages/wrangler/src/api/integrations/bindings/executionContext.ts
mrbbot: Ah ok, didn't realise `@cloudflare/workers-types` was `any`, I thought we fixed that. 😅 Ah well, let's leave this as `any` then. 👍

packages/wrangler/src/api/integrations/bindings/index.ts
mrbbot: > what should it do in a node dev server?

That's a really good question. `waitUntil()` doesn't really matter in Node. `passThroughOnException()` could arguably do something if the user had an upstream configured, but that seems unlikely. So I guess no-ops are the correct implementation here. 😃

fixtures/get-bindings-proxy/tests/get-bindings-proxy.ctx.test.ts
mrbbot: Yeah, no harm in leaving them, was just interested. 👍

mrbbot (approved):

Issue #4792, opened 2024-01-19
🚀 Feature Request [getBindingsProxy]: add a convenient way to get hold of bindings of a specific types


dario-piotrowicz:
The `getBindingsProxy` utility (introduced in https://github.com/cloudflare/workers-sdk/pull/4523) should include as part of its return object some utility methods that allow users to get hold of specific type of binding proxies

Something along the line of
```js
import { getBindingsProxy } from "wrangler";

const { getBindingsOfType } = await getBindingsProxy();

const kvs = getBindingsOfType('kv');
const d1s = getBindingsOfType('d1');

// ...
```

Or maybe the bindings could be already organized per-type as in:
```js
import { getBindingsProxy } from "wrangler";

const { groupedBindings } = await getBindingsProxy();

const kvs = groupedBindings.kv;
const d1s = groupedBindings.d1;

// ...
```

dario-piotrowicz:
cc. @alexanderniebuhr

admah:
@dario-piotrowicz can this be closed as complete?

penalosa:
This is a very old issue so I'm going to close it for now. If you're still running into problems feel free to comment with more details and we can investigate further.

PR #4926, opened 2024-02-05
add a `cf` field to the `getBindingsProxy` result

### LINKED ISSUE #4875: 🚀 Feature Request [getBindingsProxy]: `getBindingsProxy` should expose a `cf` object

dario-piotrowicz:
Fixes #4875

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/issues/12840
  - [ ] Not necessary because:

**Note for PR author:**

### DIFF
--- a/.changeset/strong-otters-hope.md
+++ b/.changeset/strong-otters-hope.md
@@ -1,0 +1,16 @@
+---
+"wrangler": minor
+---
+
+feature: add a `cf` field to the `getBindingsProxy` result
+
+Add a new `cf` filed to the `getBindingsProxy` result that people can use to mock the production
+`cf` (`IncomingRequestCfProperties`) object.
+
+Example:
+
+```ts
+const { cf } = await getBindingsProxy();
+
+console.log(`country = ${cf.country} ; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
+```
--- a/.changeset/strong-otters-hope.md
+++ b/.changeset/strong-otters-hope.md
@@ -4,7 +4,7 @@
 
 feature: add a `cf` field to the `getBindingsProxy` result
 
-Add a new `cf` filed to the `getBindingsProxy` result that people can use to mock the production
+Add a new `cf` field to the `getBindingsProxy` result that people can use to mock the production
 `cf` (`IncomingRequestCfProperties`) object.
 
 Example:
@@ -12,5 +12,5 @@ Example:
 ```ts
 const { cf } = await getBindingsProxy();
 
-console.log(`country = ${cf.country} ; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
+console.log(`country = ${cf.country}; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
 ```
--- a/packages/wrangler/src/api/integrations/bindings/index.ts
+++ b/packages/wrangler/src/api/integrations/bindings/index.ts
@@ -93,12 +93,15 @@ export async function getBindingsProxy<Bindings = Record<string, unknown>>(
 
 	const vars = getVarsForDev(rawConfig, env);
 
+	const cf = await mf.getCf();
+	deepFreeze(cf);
+
 	return {
 		bindings: {
 			...vars,
 			...bindings,
 		},
-		cf: Object.freeze(await mf.getCf()),
+		cf,
 		ctx: new ExecutionContext(),
 		caches: new CacheStorage(),
 		dispose: () => mf.dispose(),
@@ -182,4 +185,15 @@ function getMiniflarePersistOptions(
 		r2Persist: `${persistPath}/r2`,
 		d1Persist: `${persistPath}/d1`,
 	};
+}
+
+function deepFreeze<T extends Record<string | number | symbol, unknown>>(
+	obj: T
+): void {
+	Object.freeze(obj);
+	Object.entries(obj).forEach(([, prop]) => {
+		if (prop !== null && typeof prop === "object" && !Object.isFrozen(prop)) {
+			deepFreeze(prop as Record<string | number | symbol, unknown>);
+		}
+	});
 }
--- a/packages/wrangler/src/api/integrations/bindings/index.ts
+++ b/packages/wrangler/src/api/integrations/bindings/index.ts
@@ -37,7 +37,10 @@ export type GetBindingsProxyOptions = {
 /**
  * Result of the `getBindingsProxy` utility
  */
-export type BindingsProxy<Bindings = Record<string, unknown>> = {
+export type BindingsProxy<
+	Bindings = Record<string, unknown>,
+	CfProperties extends Record<string, unknown> = IncomingRequestCfProperties
+> = {
 	/**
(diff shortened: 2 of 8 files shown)


packages/wrangler/src/api/integrations/bindings/index.ts
```diff
@@ -93,6 +98,7 @@ export async function getBindingsProxy<Bindings = Record<string, unknown>>(
 			...vars,
 			...bindings,
 		},
+		cf: Object.freeze(await mf.getCf()),
```
dario-piotrowicz, line 101: Arguably we should move the `freeze` to miniflare 🤔 

what do you think @mrbbot?

packages/wrangler/src/api/integrations/bindings/index.ts
```diff
@@ -41,6 +41,11 @@ export type BindingsProxy<Bindings = Record<string, unknown>> = {
 	 * Object containing the various proxies
 	 */
 	bindings: Bindings;
+	/**
+	 * Mock of the context object that Workers received in their request handler, all the object's methods are no-op
+	 */
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	cf: Record<string, any>;
```
dario-piotrowicz, line 48: PS: I wanted to use `IncomingRequestCfProperties` here but I've been having issues with the `@microsoft/api-extractor` package (it seems not to cope with `workers-types` for some reason 😕)

I would go with this simple type for now and look into improving the types later on (also as I want to get proper types for #4792 anyways)

But I don't mind spending more time investigating the `api-extractor` problem for this PR if that's preferred 🙂

packages/wrangler/src/api/integrations/bindings/index.ts
mrbbot: Hmmm, I'd probably remove the `freeze` entirely. `Object.freeze()` doesn't apply to nested objects, so the user could still edit some of the `cf` object. I believe this isn't frozen in the real runtime either. We are also cloning this before returning from `getCf()` so changes won't apply to requests to this instance.

mrbbot (approved):
Looks good! One minor comment, but other than that... ✅

.changeset/strong-otters-hope.md
```diff
@@ -0,0 +1,16 @@
+---
+"wrangler": minor
+---
+
+feature: add a `cf` field to the `getBindingsProxy` result
+
+Add a new `cf` filed to the `getBindingsProxy` result that people can use to mock the production
```
jculvey, line 7:
```suggestion
Add a new `cf` field to the `getBindingsProxy` result that people can use to mock the production
```

.changeset/strong-otters-hope.md
```diff
@@ -0,0 +1,16 @@
...
+`cf` (`IncomingRequestCfProperties`) object.
+
+Example:
+
+```ts
+const { cf } = await getBindingsProxy();
+
+console.log(`country = ${cf.country} ; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
```
jculvey, line 15:
```suggestion
console.log(`country = ${cf.country}; colo = ${cf.colo}`); // logs 'country = GB ; colo = LHR'
```

jculvey (approved):

.changeset/strong-otters-hope.md
dario-piotrowicz: 🤦 thanks for catching that! 🙏

packages/wrangler/src/api/integrations/bindings/index.ts
dario-piotrowicz: As discussed we've kept the `freeze` and actually implemented it recursively as that matches more closely workerd 🙂

packages/wrangler/src/api/integrations/bindings/index.ts
IgorMinar: how about we make this `Record<string, string>` initially and later replace it with `IncomingRequestCfProperties`? That way you'll be widening the type which is a non-breaking change.

If you go with `any`, you'll be forced to narrow the type in the future which is breaking.

IgorMinar (commented):
left one suggestion, but the rest looks great. thanks @dario-piotrowicz

dario-piotrowicz:
@IgorMinar if we were to go with `Record<string, string>` we'd be providing a type which is not compatible with `IncomingRequestCfProperties` since the latter can contain nested objects.

If we're afraid of `any` being to lenient, what about `unknown`? or alternatively I can define a `JSON` type here (or expose Miniflare's)?

dario-piotrowicz:
Converted to draft as I'll try to improve the `cf` type, ideally it should come from `workers-types`

dario-piotrowicz:
I made `@cloudflare/workers-types` an optional peer dependency of wrangler and I'm excluding its types from the `d.ts` file we generate this produces the following result:

| project without `@cloudflare/workers-types` installed | project with `@cloudflare/workers-types` installed |
|--------|--------|
| Screenshot 2024-02-07 at 12 41 04| Screenshot 2024-02-07 at 12 41 42 |

meaning that types don't get recognized (and typescript falls back to `any`s) when the package is not installed, and gets the workers-types types otherwise, this is good as:
 - it is consistent with Miniflare's behavior
 - allows the user to have control of that `workers-types` version they want to use (not too important here for `IncomingRequestCfProperties` but potentially important when/if we want to use other `workers-types` types in wrangler)
 
Alternatively if preferred I can make the types getting bundled in the `d.ts` file (so that those become just part of wrangler and completely separated from `workers-types`)

> [!Note]
> One last thing to mention, although having `@cloudflare/workers-types` an optional peer dependency allows
> users freedom to specify what version of the package they want, in our code we do need to import from a specific
> entrypoint, in this case (and in Miniflare's code too) we import from `/experimental`, that's unfortunately not something
> that the user can have control over

IgorMinar (approved):
lgtm! thanks for updating the types

petebacondarwin (approved):

PR #4930, opened 2024-02-06
[D1] default `wrangler d1 execute` to local mode first.

rozenmd:
This PR defaults both `wrangler d1 execute` and `wrangler d1 migrations apply` to use the local development environment provided by wrangler to match the default behaviour in `wrangler dev`.

BREAKING CHANGE: `wrangler d1 execute` and `wrangler d1 migrations apply` now default `--local` to `true`. When running `wrangler d1 execute` or `wrangler d1 migrations apply` against a remote D1 database, you will need to provide the `--remote` flag.

---

## Why?

`wrangler dev` used to default to remote preview Workers, and `wrangler d1 execute`/`wrangler d1 migrations apply` followed this practice. With wrangler v3, `wrangler dev` was updated without updating `wrangler d1 execute`/`wrangler d1 migrations apply`. This PR fixes this, and ensures `wrangler dev` and `wrangler d1 execute`/`wrangler d1 migrations apply` match again.

Closes STOR-2892 internally.

---

## Preview

You can try this out via `npx wrangler@d1`. For example:

```
npx wrangler@d1 d1 execute <DBNAME> --command ="..."
```

### Local
```
npx wrangler@d1 d1 execute northwind-test --command="SELECT 1;"         
 ⛅️ wrangler 0.0.0-e4c703ba
---------------------------
🌀 Mapping SQL input into an array of statements
🌀 Executing on local database northwind-test (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) from .wrangler/state/v3/d1:
🌀 To execute on your remote database, add a --remote flag to your wrangler command.
┌───┐
│ 1 │
├───┤
│ 1 │
└───┘
```

### Remote
```
 npx wrangler@d1 d1 execute northwind-test --command="SELECT 1;" --remote
 ⛅️ wrangler 0.0.0-e4c703ba
---------------------------
🌀 Mapping SQL input into an array of statements
🌀 Parsing 1 statements
🌀 Executing on remote database northwind-test (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx):
🌀 To execute on your local development database, remove the --remote flag from your wrangler command.
🚣 Executed 1 commands in 0.1596ms
┌───┐
│ 1 │
├───┤
│ 1 │
└───┘
```

### DIFF
--- a/.changeset/hungry-cougars-tie.md
+++ b/.changeset/hungry-cougars-tie.md
@@ -2,8 +2,8 @@
 "wrangler": minor
 ---
 
-feat: default `wrangler d1 execute` to local mode first, to match `wrangler dev`
+feat: default `wrangler d1 execute` and `wrangler d1 migrations` commands to local mode first, to match `wrangler dev`
 
-This PR defaults `wrangler d1 execute` to use the local development environment provided by wrangler to match the default behaviour in `wrangler dev`.
+This PR defaults `wrangler d1 execute` and `wrangler d1 migrations` commands to use the local development environment provided by wrangler to match the default behaviour in `wrangler dev`.
 
-BREAKING CHANGE: `wrangler d1 execute` now defaults `--local` to `true`. When running `wrangler d1 execute` against a remote D1 database, you will need to provide the `--remote` flag.
+BREAKING CHANGE: `wrangler d1 execute` and `wrangler d1 migrations` commands now default `--local` to `true`. When running `wrangler d1 execute` against a remote D1 database, you will need to provide the `--remote` flag.
--- a/.changeset/hungry-cougars-tie.md
+++ b/.changeset/hungry-cougars-tie.md
@@ -2,8 +2,8 @@
 "wrangler": minor
 ---
 
-feat: default `wrangler d1 execute` and `wrangler d1 migrations` commands to local mode first, to match `wrangler dev`
+refactor: default `wrangler d1 execute` and `wrangler d1 migrations` commands to local mode first, to match `wrangler dev`
 
 This PR defaults `wrangler d1 execute` and `wrangler d1 migrations` commands to use the local development environment provided by wrangler to match the default behaviour in `wrangler dev`.
 
-BREAKING CHANGE: `wrangler d1 execute` and `wrangler d1 migrations` commands now default `--local` to `true`. When running `wrangler d1 execute` against a remote D1 database, you will need to provide the `--remote` flag.
+BREAKING CHANGE (for a beta feature): `wrangler d1 execute` and `wrangler d1 migrations` commands now default `--local` to `true`. When running `wrangler d1 execute` against a remote D1 database, you will need to provide the `--remote` flag.
--- a/.changeset/hungry-cougars-tie.md
+++ b/.changeset/hungry-cougars-tie.md
@@ -1,0 +1,9 @@
+---
+"wrangler": minor
+---
+
+feat: default `wrangler d1 execute` to local mode first, to match `wrangler dev`
+
+This PR defaults `wrangler d1 execute` to use the local development environment provided by wrangler to match the default behaviour in `wrangler dev`.
+
+BREAKING CHANGE: `wrangler d1 execute` now defaults `--local` to `true`. When running `wrangler d1 execute` against a remote D1 database, you will need to provide the `--remote` flag.
--- a/packages/wrangler/src/__tests__/d1/execute.test.ts
+++ b/packages/wrangler/src/__tests__/d1/execute.test.ts
@@ -18,7 +18,7 @@ describe("execute", () => {
 		});
 
 		await expect(
-			runWrangler("d1 execute db --command 'select 1;'")
+			runWrangler("d1 execute db --command 'select 1;' --remote")
 		).rejects.toThrowError(
 			`In a non-interactive environment, it's necessary to set a CLOUDFLARE_API_TOKEN environment variable for wrangler to work. Please go to https://developers.cloudflare.com/fundamentals/api/get-started/create-token/ for instructions on how to create an api token, and assign its value to CLOUDFLARE_API_TOKEN.`
 		);
@@ -50,6 +50,19 @@ describe("execute", () => {
 		).rejects.toThrowError(`Error: can't use --preview with --local`);
 	});
 
+	it("should reject the use of --remote with --local", async () => {
+		setIsTTY(false);
+		writeWranglerToml({
+			d1_databases: [
+				{ binding: "DATABASE", database_name: "db", database_id: "xxxx" },
+			],
+		});
+
+		await expect(
+			runWrangler(`d1 execute db --command "select;" --local --remote`)
+		).rejects.toThrowError(`Error: can't use --remote with --local`);
+	});
+
 	it("should reject the use of --preview with --local with --json", async () => {
 		setIsTTY(false);
 		writeWranglerToml({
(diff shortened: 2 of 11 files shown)


.changeset/hungry-cougars-tie.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": minor
+---
+
+feat: default `wrangler d1 execute` and `wrangler d1 migrations` commands to local mode first, to match `wrangler dev`
```
petebacondarwin, line 5: Let's call this a `refactor`, since it is not really a new feature and having `feat` in a minor release might confuse people.

.changeset/hungry-cougars-tie.md
```diff
@@ -0,0 +1,9 @@
...
+
+This PR defaults `wrangler d1 execute` and `wrangler d1 migrations` commands to use the local development environment provided by wrangler to match the default behaviour in `wrangler dev`.
+
+BREAKING CHANGE: `wrangler d1 execute` and `wrangler d1 migrations` commands now default `--local` to `true`. When running `wrangler d1 execute` against a remote D1 database, you will need to provide the `--remote` flag.
```
petebacondarwin, line 9: Can we clarify that this is a "beta feature breaking change" so that people don't get confused when it is not in a major version bump.

packages/wrangler/src/__tests__/d1/execute.test.ts
```diff
@@ -50,6 +50,19 @@ describe("execute", () => {
...
+				{ binding: "DATABASE", database_name: "db", database_id: "xxxx" },
+			],
+		});
+
+		await expect(
+			runWrangler(`d1 execute db --command "select;" --local --remote`)
+		).rejects.toThrowError(`Error: can't use --remote with --local`);
+	});
+
 	it("should reject the use of --preview with --local with --json", async () => {
```
petebacondarwin, line 66: I note that since `--local` is now the default then this test should fail as long as `--remote` is not provided, right?
There is a similar non-json test above.
Should the messaging be changed to something like "must use --remote with --preview"?

packages/wrangler/src/d1/execute.tsx
```diff
@@ -48,6 +48,13 @@ export function Options(yargs: CommonYargsArgv) {
 			describe:
 				"Execute commands/files against a local DB for use with wrangler dev",
 			type: "boolean",
+			deprecated: true,
```
petebacondarwin, line 51: Why deprecate this rather than removing it? Or at least hard failing if this appears?

Given that we are saying this is a beta command that can have breaking changes at any point, it seems unnecessary to have a deprecation period.

Better to just hard fail now with a decent error message and get people to update their stuff.

petebacondarwin: Conversely if we are deprecating this with a warning, then how about a warning for people who are not providing any `--local` nor `--remote` and tell them "did you know that the behaviour has changed - you are now actually running locally here"?

packages/wrangler/src/d1/execute.tsx
```diff
@@ -192,12 +203,21 @@ export async function executeSql({
...
 		logger.loggerLevel = "error";
 	}
+	if (local) {
+		logger.log(
+			"You no longer need to provide --local with `wrangler d1 execute`, local will be used by default."
+		);
+	}
 	const sql = file ? readFileSync(file) : command;
 	if (!sql) throw new UserError(`Error: must provide --command or --file.`);
 	if (preview && local)
```
petebacondarwin, line 213: Should this be?

```suggestion
	if (preview && !remote)
```

DaniFoldi:
Hi :wave:,

While I do agree with the local-first approach this PR proposes for D1, merging it without other products following will add additional complexity for users - which commands need --local, and which ones do --remote? Until now, it used to be
dev is local-first (since v3, gives a nice warning and usually not destructive anyways, so easy to fix), and storage products are remote first. I feel like one command is trying to accomplish too much, could it be moved to `wrangler local d1` or `wrangler prod d1`?

cc @admah

elithrar:
> could it be moved to wrangler local d1 or wrangler prod d1?

We should definitely avoid subcommand groups that further nest products. `prod` also has very different interpretations.

rozenmd:
`wrangler d1 execute` is used by folks to load and interact with their databases during dev. 

An extremely common point of confusion is that `wrangler dev` and `wrangler d1 execute` do not talk to the same environments any more (they used to before wrangler v3).

DaniFoldi:
> An extremely common point of confusion is that wrangler dev and wrangler d1 execute do not talk to the same environments any more
The same could be said about KV, for example:
- KV docs > reference > KV commands doesn't mention remote/local at all https://developers.cloudflare.com/kv/reference/kv-commands/
- Workers docs > wrangler > commands lists `--local`, but is inconsistent with dev the same way

> We should definitely avoid subcommand groups that further nest products.
I agree that it wouldn't be the best solution, just an idea. Could also be `wrangler-local`, or a "switch" from remote to local mode that persists across commands, or something else. I'm just saying that `wrangler d1` isn't the only one which is remote-first while `dev` isn't.

joshthoward:
Hey @DaniFoldi. 

This is a good callout, but right now we have a decision to make before we go GA on whether we want to be inconsistent with KV or inconsistent with dev and the latter has led to substantial confusion among beta customers. Defaulting to local is also a much safer alternative because it reduces the likelihood that someone would inadvertently touch to production.

vy-ton:
@joshthoward @rozenmd If `wrangler d1 execute` defaults to local first, so should `wrangler d1 migrations apply`?

joshthoward:
@vy-ton is this not what happens? It's what Max included in the description.

vy-ton:
@joshthoward My migration executed remotely (I see the tables in dash)

```quick-test % npx wrangler d1 migrations apply duck
 ⛅️ wrangler 3.31.0
-------------------
✔ Select an account › D1 Sandbox
Migrations to be applied:
┌───────────────────────┐
│ name                  │
├───────────────────────┤
│ 0000_create-table.sql │
└───────────────────────┘
✔ About to apply 1 migration(s)
Your database may not be available to serve requests during the migration, continue? … yes
🌀 Mapping SQL input into an array of statements
🌀 Parsing 30 statements
🌀 Executing on remote database duck (7b0fab2c-9ab5-4325-adb9-c3850548ebbc):
🌀 To execute on your local development database, pass the --local flag to 'wrangler d1 execute'
🚣 Executed 30 commands in 2.7105ms
┌───────────────────────┬────────┐
│ name                  │ status │
├───────────────────────┼────────┤
│ 0000_create-table.sql │ ✅       │
└───────────────────────┴────────┘```

**Comment by rozenmd** [comment]
@vy-ton @joshthoward this PR is not ready for review

File: packages/wrangler/src/d1/execute.tsx [hunk 6]
```diff
@@ -100,6 +109,11 @@ export const Handler = async (args: HandlerOptions): Promise<void> => {
 		logger.loggerLevel = "error";
 	}
 	await printWranglerBanner();
+	if (local) {
+		throw new UserError(
+			"`wrangler d1 execute` now defaults to local mode. Remove --local from your command to continue."
```
nora-soderlund, line 114: Some developer experience insight here... even if the commands defaults to either local or remote, I would prefer if both `--remote` and `--local` is supported.

I'm always being explicit right now and typing `--local` and `--remote` for all of my commands because I can't honestly remember what is the default, but I'm being 'punished' by wrangler for being explicit.

rozenmd: Good point.

petebacondarwin (approved):

PR #4931, opened 2024-02-06
make the entrypoint optional for the `types` command

dario-piotrowicz:
Run `wrangler types` against a `wrangler.toml` file not including a `main` field.

- Tests
  - [ ] Included
  - [x] Not necessary because: the `types` command doesn't seem to have tests associated with it (I can look into adding them here or in a followup PR if we want)
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because:
  - [x] I can add the info here: https://developers.cloudflare.com/workers/wrangler/commands/#types if desired

**Note for PR author:**

### DIFF
--- a/.changeset/breezy-worms-pull.md
+++ b/.changeset/breezy-worms-pull.md
@@ -1,0 +1,15 @@
+---
+"wrangler": patch
+---
+
+fix: make the entrypoint optional for the `types` command
+
+Currently running `wrangler types` against a `wrangler.toml` file without a defined entrypoint (`main` value)
+causes the command to error with the following message:
+
+```
+✘ [ERROR] Missing entry-point: The entry-point should be specified via the command line (e.g. `wrangler types path/to/script`) or the `main` config field.
+```
+
+However developers could want to generate types without the entrypoint being defined (for example when using `getBindingsProxy`), so these changes
+make the entrypoint no longer mandatory for the `types` command, defaulting on assuming the use of the modules syntax if no entrypoint is specified.
--- a/.changeset/breezy-worms-pull.md
+++ b/.changeset/breezy-worms-pull.md
@@ -12,4 +12,4 @@ causes the command to error with the following message:
 ```
 
 However developers could want to generate types without the entrypoint being defined (for example when using `getBindingsProxy`), so these changes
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/type-generation.ts
```diff
@@ -4,14 +4,19 @@ import { getEntry } from "./deployment-bundle/entry";
...
 export async function generateTypes(
 	configToDTS: Partial<Config>,
 	config: Config
 ) {
-	const entry = await getEntry({}, config, "types");
+	let entrypointFormat: CfScriptFormat = "modules";
+	try {
+		const entry = await getEntry({}, config, "types");
+		entrypointFormat = entry.format;
+	} catch {}
```
mrbbot, line 19: Is there any way we can check whether `getEntry()` is likely to fail before calling it rather than catching the error? Maybe check for `main` in the `config`? This would avoid the swallowing `... } catch {}` which can make debugging issues tricky.

.changeset/breezy-worms-pull.md
```diff
@@ -0,0 +1,15 @@
...
+
+Currently running `wrangler types` against a `wrangler.toml` file without a defined entrypoint (`main` value)
+causes the command to error with the following message:
+
+```
+✘ [ERROR] Missing entry-point: The entry-point should be specified via the command line (e.g. `wrangler types path/to/script`) or the `main` config field.
+```
+
+However developers could want to generate types without the entrypoint being defined (for example when using `getBindingsProxy`), so these changes
+make the entrypoint no longer mandatory for the `types` command, defaulting on assuming the use of the modules syntax if no entrypoint is specified.
```
mrbbot, line 15:
```suggestion
make the entrypoint optional for the `types` command, assuming modules syntax if none is specified.
```

mrbbot (commented):
Looks good, a couple minor comments. 🙂

packages/wrangler/src/type-generation.ts
dario-piotrowicz: good point, I've updated the logic to call `getEntry()` only when there actually is an entrypoint 🙂

mrbbot (approved):

PR #4937, opened 2024-02-07
[playground-preview-worker] fix: don't report invalid upload input

mrbbot:

Similar to #4904, the PR validates returns 4xx errors for invalid user input, as opposed to 500's reported to Sentry. In particular, this should prevent the `Parsing a Body as FormData requires a Content-Type header.` errors we've been seeing. There's nothing we can do about these errors as they're caused by the client. This PR also bumps the `hono` version to address the `Cannot read properties of undefined (reading 'split')` errors.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [x] Not necessary because: internal worker
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: internal worker

**Note for PR author:**

### DIFF
--- a/packages/playground-preview-worker/src/sentry.ts
+++ b/packages/playground-preview-worker/src/sentry.ts
@@ -1,5 +1,5 @@
 import { Toucan } from "toucan-js";
-import { z, ZodError } from "zod";
+import { ZodError } from "zod";
 import { HttpError, ZodSchemaError } from ".";
 
 export function handleException(e: unknown, sentry: Toucan): Response {
--- a/packages/playground-preview-worker/src/sentry.ts
+++ b/packages/playground-preview-worker/src/sentry.ts
@@ -1,6 +1,6 @@
 import { Toucan } from "toucan-js";
 import { ZodError } from "zod";
-import { HttpError, ZodSchemaError } from ".";
+import { HttpError, ZodSchemaError } from "./errors";
 
 export function handleException(e: unknown, sentry: Toucan): Response {
 	console.error(e);
--- a/packages/playground-preview-worker/src/user.do.ts
+++ b/packages/playground-preview-worker/src/user.do.ts
@@ -1,5 +1,6 @@
 import assert from "node:assert";
 import { Buffer } from "node:buffer";
+import z from "zod";
 import { constructMiddleware } from "./inject-middleware";
 import {
 	doUpload,
@@ -8,7 +9,7 @@ import {
 	UploadResult,
 } from "./realish";
 import { handleException, setupSentry } from "./sentry";
-import { ServiceWorkerNotSupported, WorkerTimeout } from ".";
+import { BadUpload, ServiceWorkerNotSupported, WorkerTimeout } from ".";
 
 const encoder = new TextEncoder();
 
@@ -25,6 +26,15 @@ function switchRemote(url: URL, remote: string) {
 	workerUrl.port = remoteUrl.port;
 	return workerUrl;
 }
+
+const UploadedMetadata = z.object({
+	body_part: z.ostring(),
+	main_module: z.ostring(),
+	compatibility_date: z.ostring(),
+	compatibility_flags: z.array(z.string()).optional(),
+});
+
+type UploadedMetadata = z.infer<typeof UploadedMetadata>;
 
 /**
  * This Durable object coordinates operations for a specific user session. It's purpose is to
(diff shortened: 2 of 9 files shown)


packages/playground-preview-worker/src/sentry.ts
```diff
@@ -1,5 +1,5 @@
 import { Toucan } from "toucan-js";
-import { z, ZodError } from "zod";
+import { ZodError } from "zod";
```
petebacondarwin, line 2: This was a redundant import? Do we not have linting rules to catch this on this package?

packages/playground-preview-worker/src/user.do.ts
```diff
@@ -158,17 +160,33 @@ export class UserSession {
...
+
 		const userSession = this.state.id.toString();
-		const worker = await request.formData();
 
-		const m = worker.get("metadata");
+		let worker: FormData;
+		try {
+			worker = await request.formData();
+		} catch (e) {
+			throw new BadUpload(`Expected valid form data`);
```
petebacondarwin, line 170: Is it worth appending the info from the original `e` to this new exception?
In case that would be helpful to the user debugging their problem?

packages/playground-preview-worker/src/user.do.ts
```diff
@@ -158,17 +160,33 @@ export class UserSession {
...
-		assert(m instanceof File);
+		const m = worker.get("metadata");
+		if (!(m instanceof File)) {
+			throw new BadUpload("Expected metadata file to be defined");
+		}
 
-		const uploadedMetadata = JSON.parse(await m.text());
+		let uploadedMetadata: UploadedMetadata;
+		try {
+			uploadedMetadata = UploadedMetadata.parse(JSON.parse(await m.text()));
```
petebacondarwin, line 180: Out of curiosity, I seem to remember having to do this before but can't remember why this works when `UploadedMetadata.parse(await m.json())` doesn't...

packages/playground-preview-worker/src/user.do.ts
```diff
@@ -158,17 +160,33 @@ export class UserSession {
...
+		} catch {
+			throw new BadUpload("Expected metadata file to be valid");
```
petebacondarwin, line 182:
```suggestion
			throw new BadUpload("Expected metadata file to be valid JSON");
```

petebacondarwin (approved):
By the way, not related to this PR but I note that `index.ts` and `sentry.ts` have a cyclic import dependency that would be nice to avoid.

packages/playground-preview-worker/src/sentry.ts
mrbbot: Looks like we didn't, added in bcfb5ff

packages/playground-preview-worker/src/user.do.ts
mrbbot: d07ec1b

packages/playground-preview-worker/src/user.do.ts
mrbbot: Ah, I think I was going to include the body in the error message, but then decided not to? Do you think it should be included there?

packages/playground-preview-worker/src/user.do.ts
mrbbot: Since the error message includes both the case when the metadata isn't valid JSON, and that's it doesn't match the schema, I think this error message is ok.

packages/playground-preview-worker/tests/index.test.ts
```diff
@@ -85,7 +85,7 @@ describe("Preview Worker", () => {
 			'"/hello?world"'
 		);
 		expect(resp.headers.get("set-cookie") ?? "").toMatchInlineSnapshot(
-			`"token=${defaultUserToken}; Domain=random-data.playground-testing.devprod.cloudflare.dev; HttpOnly; Secure; SameSite=None"`
+			`"token=${defaultUserToken}; Domain=random-data.playground-testing.devprod.cloudflare.dev; Path=/; HttpOnly; Secure; SameSite=None"`
```
mrbbot, line 88: I think this change in `Set-Cookie` header is ok, as cookie `Path`s apply to subpaths to?

packages/playground-preview-worker/src/user.do.ts
petebacondarwin: Ah right. I didn't realise that `UploadedMetadata.parse()` might throw.

packages/playground-preview-worker/src/user.do.ts
petebacondarwin: Right that's it. If you use `.json()` then you don't get access to the bad content to use in the error message.
Given the brevity of our other errors I think adding the body to the error is not necessary.

packages/playground-preview-worker/tests/index.test.ts
petebacondarwin: Yes that is my understanding too.

petebacondarwin (approved):
Still LGTM

PR #4945, opened 2024-02-07
feat: Add feature metrics to asset-server

jahands:
PAGES-3495

- Tests
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because:

**Note for PR author:**

### DIFF
--- a/packages/pages-shared/asset-server/handler.ts
+++ b/packages/pages-shared/asset-server/handler.ts
@@ -70,12 +70,20 @@ type ServeAsset<AssetEntry> = (
 	options?: { preserve: boolean }
 ) => Promise<Response>;
 
+type CacheStatus = "hit" | "miss";
+type CacheResult<A extends string> = `${A}-${CacheStatus}`;
+export type HandlerMetrics = {
+	preservationCacheResult?: CacheResult<"checked">;
+	earlyHintsResult?: CacheResult<"used" | "notused">;
+};
+
 type FullHandlerContext<AssetEntry, ContentNegotiation, Asset> = {
 	request: Request;
 	metadata: Metadata;
 	xServerEnvHeader?: string;
 	xDeploymentIdHeader?: boolean;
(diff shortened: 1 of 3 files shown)


packages/pages-shared/asset-server/handler.ts
```diff
@@ -70,12 +70,20 @@ type ServeAsset<AssetEntry> = (
...
+	earlyHintsResult?: CacheResult<"used" | "notused">;
+};
+
 type FullHandlerContext<AssetEntry, ContentNegotiation, Asset> = {
 	request: Request;
 	metadata: Metadata;
 	xServerEnvHeader?: string;
 	xDeploymentIdHeader?: boolean;
 	logError: (err: Error) => void;
+	setMetrics: (metrics: HandlerMetrics) => void;
```
GregBrimble, line 86: We should probably make this optional, so we can just avoid passing it in with `wrangler pages dev`'s usage.

jahands: Ah good call, will update

packages/pages-shared/asset-server/handler.ts
```diff
@@ -70,12 +70,20 @@ type ServeAsset<AssetEntry> = (
 	options?: { preserve: boolean }
 ) => Promise<Response>;
 
+type CacheStatus = "hit" | "miss";
+type CacheResult<A extends string> = `${A}-${CacheStatus}`;
+export type HandlerMetrics = {
+	preservationCacheResult?: CacheResult<"checked">;
```
jrf0110, line 76: Is this one used at all?

jahands: Oops, it is now!

jrf0110 (approved):

PR #4812, opened 2024-01-22
fix: ensure that Origin header is rewritten as necessary

petebacondarwin:
The `wrangler dev` command puts the Worker under test behind a proxy server. This proxy server should be transparent to the client and the Worker, which means that the `Request` arriving at the Worker with the correct `url` property, and `Host` and `Origin` headers.

Previously we fixed the `Host` header but missed the `Origin` header which is only added to a request under certain circumstances, such as cross-origin requests.

**What this PR solves / how to test:**

This change fixes the `Origin` header as well, so that it is rewritten, when it exists, to use the `origin` of the `url` property.

A nice way to test this is to simply clone the reproduction from #4761 - https://github.com/Le0Developer/wrangler-host-header-repro, then replace the local Miniflare with the pre-release one from this PR.

Fixes #4761 

**Author has addressed the following:**

- Tests
  - [x] Included (updated worker-app fixture tests)
  - [ ] Not necessary because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: no user facing change

PR #4950, opened 2024-02-07
fix: ensure we do not rewrite external Origin headers in wrangler dev

petebacondarwin:
In https://github.com/cloudflare/workers-sdk/pull/4812 we tried to fix the Origin headers to match the Host header but were overzealous and rewrote Origin headers for external origins (outside of the proxy server's origin).

This is now fixed, and moreover we rewrite any headers that refer to the proxy server on the request with the configured host and vice versa on the response.

This should ensure that CORS is not broken in browsers when a different host is being simulated based on routes in the Wrangler configuration.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):


packages/wrangler/src/dev/remote.tsx
```diff
@@ -323,10 +323,13 @@ export function useWorker(
...
 				},
-				userWorkerInnerUrlOverrides: {}, // there is no analagous prop for this option because we did not permit overriding request.url in remote mode
+				userWorkerInnerUrlOverrides: {
+					hostname: props.host,
+					port: props.port.toString(),
+				},
 				headers: {
 					"cf-workers-preview-token": workerPreviewToken.value,
-					Cookie: accessToken && `CF_Authorization=${accessToken}`,
+					...(accessToken ? { Cookie: `CF_Authorization=${accessToken}` } : {}),
```
petebacondarwin, line 332: I noted that we had to make the `headers` record allow `undefined` values just to accommodate this `Cookie` assignment. But in reality we don't actually want to assign `undefined` to Cookie, since the code later one would result in the Cookie looking like: `previous-cookies;undefined`.

So I fixed this up here, removed the `undefined` from the type and was able to avoid a `!` operator later on.

RamIdeas: This is fair enough. And is probably the "correct" solution. But this type will eventually be exposed to users to write themselves, and I believe allowing `undefined` will be more ergonomic and subtly improve DX

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
```diff
@@ -250,3 +255,20 @@ function insertLiveReloadScript(
 
 	return htmlRewriter.transform(response);
 }
+
+/**
+ * Rewrite references to URLs in request headers.
+ *
+ * This function is used to map the URLs in headers like Origin and Access-Control-Allow-Origin
```
RamIdeas, line 262: Access-Control-Allow-Origin is a response header so I assume you expect this function to be used in both directions?

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
```diff
@@ -127,6 +130,8 @@ export class ProxyWorker implements DurableObject {
 			// if we decide to await, we should include a timeout (~100ms) in case the user worker has long-running/parellel requests
 			void fetch(userWorkerUrl, new Request(request, { headers }))
 				.then((res) => {
+					rewriteUrlRelatedHeaders(headers, innerUrl, outerUrl);
```
RamIdeas, line 133: Ah yes both directions! This is clever. Now the browser will have no idea that the URL was faked inside the worker either 😄 

Small bug/typo:
```suggestion
					rewriteUrlRelatedHeaders(res.headers, innerUrl, outerUrl);
```

But you might not be able to mutate response headers without:
```suggestion
					res = new Response(res);
					rewriteUrlRelatedHeaders(res.headers, innerUrl, outerUrl);
```

RamIdeas (changes requested):

packages/wrangler/src/dev/remote.tsx
petebacondarwin: I'm not convinced that allowing `undefined` here adds to user DX. If anything it creates a mental overhead because the user has to think about what setting to `undefined` actually means (when it actually means ignore this header and not remove the original header).

packages/wrangler/src/dev/remote.tsx
```diff
@@ -436,10 +439,13 @@ export async function startRemoteServer(props: RemoteProps) {
 					port: previewToken.inspectorUrl.port.toString(),
 					pathname: previewToken.inspectorUrl.pathname,
 				},
-				userWorkerInnerUrlOverrides: {}, // there is no analagous prop for this option because we did not permit overriding request.url in remote mode
+				userWorkerInnerUrlOverrides: {
+					hostname: props.host,
+					port: props.port.toString(),
```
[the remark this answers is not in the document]
petebacondarwin: Yeah I was a bit worried about that. Should we create new override properties that are specifically for header rewrites?

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
petebacondarwin: That's not a typo! It's a bona fide mistake! Thanks for spotting that.

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
petebacondarwin: I'll tweak the description.

petebacondarwin (commented):
Thanks for the review

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
RamIdeas: I also think this is a "new"-ish feature that should get its own changeset and documentation. It's really cool that a dev could write ``Response.redirect(`${request.origin}/redirect/path`)`` and that would "just work"™️ in `wrangler dev [--remote]` – we should shout from the rooftops!

petebacondarwin: Strictly this was already part of the old proxy: https://github.com/cloudflare/workers-sdk/blob/d637bd59a8ea6612d59ed4b73e115287615e617d/packages/wrangler/src/dev/proxy.ts#L70

packages/wrangler/src/dev/remote.tsx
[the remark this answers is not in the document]
petebacondarwin: Actually, as far as I can tell `MF-Original-URL` and `MF-Disable-Pretty-Error` are always set on every proxy worker request whether in local or remote mode. So I don't think that this change actually introduces any new side effect.

Whether those headers should be added in remote mode is a question for another PR I think.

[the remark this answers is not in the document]
RamIdeas: Yeah let's follow up in another PR if needed.

I think in remote mode, the inner url overrides aren't set so the header wouldn't have been set?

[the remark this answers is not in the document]
petebacondarwin: https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/templates/startDevWorker/ProxyWorker.ts#L125-L126

RamIdeas (approved):
Happy to merge. Might want to hold off merging since I believe we want to get a hot fix out today.

One remaining point (you decide whether to address or not): should we print something to the console to let the user know the origin/host is being rewritten across all request/response headers inc Set-Cookie (and not the body) and let them know they can opt-out by...?

petebacondarwin:
> Happy to merge. Might want to hold off merging since I believe we want to get a hot fix out today.
> 
> One remaining point (you decide whether to address or not): should we print something to the console to let the user know the origin/host is being rewritten across all request/response headers inc Set-Cookie (and not the body) and let them know they can opt-out by...?

Yeah good question... we have not previously had a message for rewriting the `request.url`, and prior to 3.18, I believe that we would have been rewriting response headers in this way anyway.

Perhaps we could put some debug logging messages in, so at least if the user is getting confused it could be analysed via increasing the logging level?

penalosa (approved):
The `edge-preview-authenticated-proxy` stuff here looks good!

petebacondarwin:
> In remote mode, yeah no, but locally workerd logs now reach the log file (not the terminal but that's what you were going for anyway)

Hmm I tried adding `console.debug()` logs to the ProxyWorker but these always appeared in the output of `wrangler dev` (whatever the logging level). So I think this is too much noise. Perhaps we need to tweak the Proxy Worker's Miniflare configuration?

I think we can leave this to another time, or if someone gets confused by this rewriting.

jfnault-seedbox:
@petebacondarwin Can you please take a look at 
https://github.com/cloudflare/workers-sdk/issues/5221

I think this PR is related, I had couple of notes:
https://github.com/cloudflare/workers-sdk/issues/5221#issuecomment-1996147061

PR #4795, opened 2024-01-19
Implement Vitest testing environment for Miniflare 3

mrbbot:
Closes all tickets in RM-17030 and RM-17070. Ref #4490.

**What this PR solves / how to test:**

---

This PR reintroduces Miniflare 2's Vitest unit testing environment to Miniflare 3! 🎉

Miniflare 3 runs your code inside `workerd`, rather than Node. This means we can no longer access Workers runtime APIs directly inside Node, and instead need to run tests inside a `workerd` process.
We do this by implementing a Vitest [Custom Pool](https://vitest.dev/advanced/pool.html), rather than a [Test Environment](https://vitest.dev/guide/environment.html). Pools can completely customise how test are run, whereas environments can only customise the global scope. Many thanks to [@sheremet-va](https://github.com/sheremet-va) for [implementing custom pools](https://github.com/vitest-dev/vitest/pull/4417), and their continued support throughout this project.
Our pool starts `Miniflare` instances running the regular Vitest worker script, and uses WebSockets rather than inter-thread/process communication for message passing.

This worker script expects to run inside a Node-like environment. To support this, we've implemented a "module fallback" service that provides Node-like disk-backed module resolution for `require()`s and `import` statements in `workerd`. Many thanks to [@jasnell](https://github.com/jasnell) for [implementing the fallback service API](https://github.com/cloudflare/workerd/pull/1423) and [unsafe-eval bindings](https://github.com/cloudflare/workerd/pull/1338) this project also depends on.

Right now, the pool only supports `vitest@1.1.3`. I'll upgrade to the latest Vitest version soon.

### Features

🟢 Implemented, 🟡 Planned, 🔴 Not Planned

- 🟢 Unit testing any function in a Worker
- 🟢 Calling `fetch()`, `scheduled()` and `queue()` exported handlers directly
- 🟢 Waiting until `waitUntil()`ed `Promise`s have settled
- 🟢 Integration testing of Workers via service bindings
- 🟢 Multi-worker support
- 🟢 Accessing bindings, with support for seeding data
- 🟢 Isolated per-test storage
- 🟢 Directly calling Durable Object instance methods
- 🟢 Directly accessing Durable Object state
- 🟢 Immediately executing Durable Object alarms
- 🟢 Listing Durable Objects
- 🟢 Mocking outbound `fetch()` requests
- 🟡 Testing Pages functions
- 🟡 Easily apply D1 migrations
- Provided by Vitest:
  - 🟢 Fast re-runs with HMR
[...]

PR #4954, opened 2024-02-08
[miniflare] fix: allow relative modules `scriptPath` outside of CWD

### LINKED ISSUE #4721: 🐛 BUG: "can't use ".." to break out of starting directory" when using Service Bindings in monorepo

Miniflare

3.20231218.1 [miniflare]

20

Mac

### Describe the Bug

I'm attempting to use Miniflare to run one app (app-2) with a second app (app-1) as a Service Binding. The docs are a little thin here, so I'm not sure if I'm defining the workers in the correct way.

```js
return new Miniflare({
  port: 4000,

  workers: [
    {
      name: 'musicfetch-api',
      scriptPath: 'dist/index.js',
      modules: true,

      serviceBindings: {
        APP_2: 'app-2',
      },
    },
    {
      name: 'app-2',
      scriptPath: resolvePath(__dirname, '../app-2/dist/index.js'),
      modules: true,

      durableObjects: {
        MyDurableObject: 'MyClass',
      },
    },
  ],
});
```

In doing this I'm getting the error:

```
workerd/util/symbolizer.c++:98: warning: Not symbolizing stack traces because $LLVM_SYMBOLIZER is not set. To symbolize stack traces, set $LLVM_SYMBOLIZER to the location of the llvm-symbolizer binary. When running tests under bazel, use `--test_env=LLVM_SYMBOLIZER=<path>`.
workerd/jsg/util.c++:281: error: e = kj/filesystem.c++:315: failed: expected parts.size() > 0 [0 > 0]; can't use ".." to break out of starting directory
stack: 1028e4c07 1028e3963 1028e383f 1008c776f 100b7ece3 100da687f 100db41ef 100db4083 100877e67 10087ddc7 1008815c3 10087ff13 100871b03 100871733 100871653 10087163b 1028fb9ff 1028fbd47 1028fa4a7 1028fa277 1008641af 182aed0df; sentryErrorContext = jsgInternalError
service core:user:rate-limiter: Uncaught Error: internal error
```

Notably: `can't use ".." to break out of starting directory`

I've tried various other approaches (eg. importing a Miniflare instance from `app-2` and calling it programatcally), but it results in the same error. It feels like I'm battling against the framework here for what I imagine is a common project structure.

- Is there a way to define the Miniflare CWD? If so I could have it execute in the parent directory.
- For now I will probably hack around this, but just wanted to leave an issue here to confirm there wasn't a better way.

Cheers 🍺 

_No response_

_No response_

mrbbot:
Fixes #4721.

Previously, Miniflare would resolve relative `scriptPath`s against `moduleRoot` multiple times resulting in incorrect paths and module names. This would lead to `can't use ".." to break out of starting directory` `workerd` errors. This change ensures Miniflare uses `scriptPath` as is, and only resolves it relative to `modulesRoot` when computing module names. Note this bug didn't affect service workers. This allows you to reference a modules `scriptPath` outside the working directory with something like:

```js
const mf = new Miniflare({
	modules: true,
	modulesRoot: "..",
	scriptPath: "../worker.mjs",
});
```

Notably, this wasn't a problem if `modulesRoot` and `scriptPath` were absolute paths.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: fixing a bug

**Note for PR author:**

### DIFF
--- a/packages/miniflare/src/plugins/core/modules.ts
+++ b/packages/miniflare/src/plugins/core/modules.ts
@@ -174,8 +174,6 @@ export class ModuleLocator {
 	}
 
 	visitEntrypoint(code: string, modulePath: string) {
-		modulePath = path.resolve(this.modulesRoot, modulePath);
-
 		// If we've already visited this path, return
 		if (this.#visitedPaths.has(modulePath)) return;
 		this.#visitedPaths.add(modulePath);
--- a/packages/miniflare/test/plugins/core/modules.spec.ts
+++ b/packages/miniflare/test/plugins/core/modules.spec.ts
@@ -1,8 +1,9 @@
 import assert from "assert";
+import fs from "fs/promises";
 import path from "path";
 import test from "ava";
 import { Miniflare, MiniflareCoreError, stripAnsi } from "miniflare";
-import { utf8Encode } from "../../test-shared";
+import { useCwd, useTmp, utf8Encode } from "../../test-shared";
 
 const ROOT = path.resolve(
 	__dirname,
@@ -278,6 +279,30 @@ You must manually define your modules when constructing Miniflare:
     at ${depPath}:2:8`
 	);
 });
+test.serial(
+	"Miniflare: collects modules outside of working directory",
+	async (t) => {
+		// https://github.com/cloudflare/workers-sdk/issues/4721
+		const tmp = await useTmp(t);
+		const child = path.join(tmp, "child");
+		await fs.mkdir(child);
+		await fs.writeFile(
+			path.join(tmp, "worker.mjs"),
+			'export default { fetch() { return new Response("body"); } }'
+		);
+		useCwd(t, child);
+
+		const mf = new Miniflare({
+			modules: true,
+			modulesRoot: "..",
+			scriptPath: "../worker.mjs",
+		});
+		t.teardown(() => mf.dispose());
+
+		const res = await mf.dispatchFetch("http://localhost");
+		t.is(await res.text(), "body");
+	}
+);
 test("Miniflare: suggests bundling on unknown module", async (t) => {
 	// Try with npm-package-like import
 	let mf = new Miniflare({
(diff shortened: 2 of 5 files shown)


packages/miniflare/src/plugins/core/modules.ts
```diff
@@ -174,8 +174,6 @@ export class ModuleLocator {
 	}
 
 	visitEntrypoint(code: string, modulePath: string) {
-		modulePath = path.resolve(this.modulesRoot, modulePath);
```
mrbbot, line 177: I think this change is safe to make. Resolving to the `modulesRoot` root isn't correct here if `modulePath` is relative. If `modulePath` were absolute, this call would do nothing anyways.

packages/miniflare/test/plugins/core/modules.spec.ts
```diff
@@ -278,6 +279,30 @@ You must manually define your modules when constructing Miniflare:
...
+		await fs.writeFile(
+			path.join(tmp, "worker.mjs"),
+			'export default { fetch() { return new Response("body"); } }'
+		);
+		useCwd(t, child);
+
+		const mf = new Miniflare({
+			modules: true,
+			modulesRoot: "..",
+			scriptPath: "../worker.mjs",
```
RamIdeas, line 298: could we avoid the cwd change+resetting by choosing another tmp dir for the worker and using a relative path from the cwd to the other tmp dir?

packages/miniflare/src/plugins/core/modules.ts
mrbbot: Yep, the recursive function is `#visitJavaScriptModule()` below. 👍

packages/miniflare/test/plugins/core/modules.spec.ts
mrbbot: Yeah, we could. Was keen to this behaviour using the CWD specifically since it's very simple, and is what users are most likely to hit. The `useCwd()` implementation is actually from #4795, which uses it for some other tests too, so we're not just adding it for this. 👍

PR #4955, opened 2024-02-08
Update prefixes for e2e tests to cleanup all e2e workers

penalosa:

In a number of places, we run e2e tests against Cloudflare APIs that result in Workers and/or Pages projects being created. This included:
- C3 E2E tests
- Wrangler E2E tests
- Internal E2E tests

These tests all include cleanup as part of their logic (in an `afterEach` or similar), but sometimes fail to properly clean up the created projects (whether through tests crashing, CI timeouts, or random chance). Because of that, we have a CI job which cleans up created e2e projects after some period of time. Previously, it only cleanup up projects created by the C3 E2E tests—this PR ensures that the cleanup action cleans up projects created from the Wrangler, C3, or internal E2E tests. It also updates the Wrangler tests to include a clearer well-known prefix to support this. 

- Tests
  - [ ] Included
  - [x] Not necessary because: This is a modification to how our existing tests run
  - [ ] Included
  - [x] Not necessary because: This isn't a user-facing change
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: This isn't a user-facing change

### DIFF
--- a/packages/create-cloudflare/scripts/common.ts
+++ b/packages/create-cloudflare/scripts/common.ts
@@ -18,6 +18,8 @@ export type Worker = {
 	created_on: string;
 };
 
+const tmpPrefixes = ["tmp-stratus-e2e-", "tmp-wrangler-e2e-", "tmp-c3-e2e-"];
+
 const apiFetch = async (
 	path: string,
 	init = { method: "GET" },
@@ -78,7 +80,7 @@ export const listC3Projects = async () => {
 
 	return projects.filter(
 		(p) =>
-			p.name.startsWith("c3-e2e-") &&
+			tmpPrefixes.some((prefix) => p.name.startsWith(prefix)) &&
 			// Projects are more than an hour old
 			Date.now() - new Date(p.created_on).valueOf() > 1000 * 60 * 60
 	);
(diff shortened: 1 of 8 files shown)


packages/create-cloudflare/scripts/common.ts
```diff
@@ -18,6 +18,8 @@ export type Worker = {
 	created_on: string;
 };
 
+const tmpPrefixes = ["tmp-stratus-e2e-", "tmp-wrangler-e2e-", "tmp-c3-e2e-"];
```
petebacondarwin, line 21: Alternatively, one could name these: "tmp-e2e-stratus-...", "tmp-e2e-wrangler-...", etc and then we only need to check for "tmp-e2e-" as a prefix.

packages/create-cloudflare/scripts/common.ts
```diff
@@ -78,7 +78,7 @@ export const listC3Projects = async () => {
 
 	return projects.filter(
 		(p) =>
-			p.name.startsWith("c3-e2e-") &&
+			p.name.startsWith("tmp-e2e-") &&
```
RamIdeas, line 81: This function is called listC3Projects

If this job now deletes all "tmp-e2e-" projects/workers, we update the name

PR #4957, opened 2024-02-08
Remove stripping .py extensions and bump workerd version

garrettgu10:
The main branch of `workerd` no longer requires stripping the `.py` extension from Python modules (and in fact, crashes if Python modules do not end in `.py`). 

This PR makes it so that Wrangler no longer strips the `.py` extension from Python modules.


penalosa:
This will likely need a `workerd` bump in Miniflare before we can land it. Are the `workerd` changes in a release yet?

dom96:
This one contains the changes necessary: https://github.com/cloudflare/workerd/releases/tag/v1.20240208.0.

@garrettgu10  We should update to it in this PR.

garrettgu10:
@penalosa I bumped the workerd version :-) but it looks like I'm failing a bunch of random test cases? Not sure what's going on... :-/

garrettgu10:
@mrbbot I'm not sure what you mean by that. I thought updating the PythonModule names to include .py would fix things, but it looks like a bunch of unrelated tests are failing?

garrettgu10:
@mrbbot does this work now? It seems to still fail CI but the D1 test does pass now

mrbbot (approved):

PR #4499, opened 2023-11-23
Improve DX with `node:*` modules

penalosa:
Fixes  #4050

**What this PR solves / how to test:**

Previously, importing `node:*` modules would fail at build time if neither nodejs compat mode was enabled. This PR changes the build time error into a build time warning, which will support writing runtime-agnostic code that conditionally imports NodeJS APIs only in a NodeJS environment.

Additionally, it adds the `navigator.userAgent` constant to esbuild's bundle settings. This means that as long as users are feature detecting with `navigator.userAgent`, `esbuild` will tree-shake away the Node.JS code, and they won't see warnings.

**Author has addressed the following:**

- Tests
  - [x] Included
  - [ ] Not necessary because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): In progress, coming with a larger update to the bundling docs
  - [ ] Not necessary because:

**Note for PR author:**

We want to celebrate and highlight awesome PR review! If you think this PR received a particularly high-caliber review, please assign it the label `highlight pr review` so future reviewers can take inspiration and learn from it.

PR #4962, opened 2024-02-09
fix: ensure worker reloads when importing `node:*` modules

### LINKED ISSUE #4959: 🐛 BUG: Crashes on reload if using `node:xxx` imports

Wrangler core

3.28.0

20.11.0

Windows 11

### Describe the Bug

In the latest version of wrangler, when you make a change in a worker that imports anything from `node:`, it seems to crash and be unable to rebuild.

`wrangler.toml`:
```toml
name = "wandering-butterfly-5921"
main = "src/index.ts"
compatibility_date = "2024-02-08"
compatibility_flags = ["nodejs_compat"]

```

`src/index.ts`:
```typescript
import { Buffer } from 'node:buffer';

export interface Env {}
export default {
	async fetch(request: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
		const str = Buffer.from('Hello World!').toString('base64');
		return new Response(str)
	},
};
```

Steps to reproduce:

- Use worker config and worker above
- npx wrangler@3.28.0 dev
- Make a change to `src/index.ts`
- Observe wrangler crashing with the following:

```
✘ [ERROR] Could not resolve "node:buffer"

    src/index.ts:1:23:
      1 │ import { Buffer } from 'node:buffer';
        ╵                        ~~~~~~~~~~~~~

  The package "node:buffer" wasn't found on the file system but is built into node.
  Add "node_compat = true" to your wrangler.toml file and make sure to prefix the module name with "node:" to enable Node.js compatibility.
```

mrbbot:
Fixes #4959.

#4499 introduced a bug which prevented reloads when importing `node:*` modules. The internal per-build `seen` cache was not cleared between builds, resulting in this branch being hit...

https://github.com/cloudflare/workers-sdk/blob/1c4f1635165d98abc0558dcf148fe1d3ce9a2ca1/packages/wrangler/src/deployment-bundle/esbuild-plugins/nodejs-compat.ts#L28-L30

...and the resolution failing. This change wraps that in an `onStart()`, so it's called at the start of each build, and also moves the caches inside the plugin, so multiple instances of the plugin can be created.

To test this, follow the reproduction instructions in #4959, no errors should be reported.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: fixing a regression

**Note for PR author:**

### DIFF
--- a/.changeset/cuddly-apples-divide.md
+++ b/.changeset/cuddly-apples-divide.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: ensure `wrangler dev` can reload without crashing when importing `node:*` modules
+
+The previous Wrangler release introduced a regression that caused reloads to fail when importing `node:*` modules. This change fixes that, and ensures these modules can always be resolved.
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -162,6 +162,7 @@ describe("basic dev tests", () => {
 					name = "${workerName}"
 					main = "src/index.ts"
 					compatibility_date = "2023-01-01"
+					compatibility_flags = ["nodejs_compat"]
 
 					[vars]
 					KEY = "value"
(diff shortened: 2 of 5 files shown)


penalosa (approved):
Thanks for this!

PR #4999, opened 2024-02-12
C3: next templates refactoring

dario-piotrowicz:
This PR aligns the Next.js template code with the rest of the C3 templates by removing its `templates.ts` file and
including template files directly in the next template directory

In order to properly implement the above also it was needed to extend the C3 config logic to support custom template variants (not only `js` and `ts`) and also allow the possibility to specify a custom destination directory for the template files to copy


packages/create-cloudflare/templates/next/c3.ts
```diff
@@ -181,6 +128,37 @@ export default {
...
+				`${ctx.project.path}/src/app`,
+				`${ctx.project.path}/app`,
+			]);
+
+			const isTypescript = usesTypescript(ctx);
+
+			const dir = isApp ? "app" : "pages";
+			return `${dir}/${isTypescript ? "ts" : "js"}`;
+		},
+		async destinationDir(ctx) {
```
jculvey, line 143: Does this need to be async?

packages/create-cloudflare/src/templates.ts
```diff
@@ -412,3 +436,25 @@ export const getTemplatePath = (ctx: C3Context) => {
...
+
+	if (!copyFiles?.destinationDir) {
+		return undefined;
+	}
+
+	if (typeof copyFiles.destinationDir === "string") {
+		return copyFiles.destinationDir;
+	}
+
+	return copyFiles.destinationDir(ctx);
```
jculvey, line 459: Same question here, does this need to be async?

packages/create-cloudflare/templates/next/c3.ts
dario-piotrowicz: nope, removed 🙂👍

packages/create-cloudflare/src/templates.ts
dario-piotrowicz: removed 🙂👍

packages/create-cloudflare/templates/next/next.config.mjs
dario-piotrowicz, line 1: @jculvey in this PR I have this file and simply copying it other the default one, I'll move to code modding in my followup PR when I do update the Next.js template if that's ok by you 🙏

packages/create-cloudflare/src/templates.ts
```diff
@@ -80,12 +88,24 @@ export type TemplateConfig = {
...
 };
 
-type StaticFileMap = Record<string, VariantInfo>;
+type StaticFileMap = {
+	selectVariant?: (ctx: C3Context) => Promise<string>;
+	variants: Record<string, VariantInfo>;
+};
+
+const defaultSelectVariant = async (ctx: C3Context) => {
+	const typescript = await shouldUseTs(ctx);
```
jculvey, line 106: There's a pretty subtle existing bug here that I didn't notice until I was reviewing your PR. It's not introduced by your change, but I'm wondering if you can fix it since you're touching this code path. 

If you run this code with a pages template that has a `copyFiles` definition and select `js` during the process, you'll be prompted if you want to use typescript by c3. 

You can repro this by running the solid template and selecting not to use `ts`. Here's what happens:
image.

I think to fix this, we need to adjust `shouldUseTs` so that it returns the result of `usesTypescript` if `ctx.platform === 'pages'`.

packages/create-cloudflare/templates/next/c3.ts
```diff
@@ -181,6 +133,37 @@ export default {
 	previewScript: "pages:preview",
 	generate,
 	configure,
+	copyFiles: {
+		async selectVariant(ctx) {
```
jculvey, line 137: I really like how this turned out 👍

jculvey (approved):

packages/create-cloudflare/src/templates.ts
dario-piotrowicz: done 🙂

lrapoport-cf:
for posterity: there are existing tests that cover this code that did not need modification because this is only refactoring 👍

PR #5007, opened 2024-02-13
[KV] Remove `worker-` prefix from kv:namespace create when using wrangler in a pages project or without a `name` property

### LINKED ISSUE #4248: `wrangler kv:namespace create <namespace>` treats `<namespace>` as a binding name rather than the resource name
### Describe the solution

Currently, when creating a new KV namespace via Wrangler, it tries to be helpful and print out a snippet for the user to paste into their `wrangler.toml` file. 
```shell
wrangler kv:namespace create "MY_KV"
🌀 Creating namespace with title "worker-MY_KV"
✨ Success!
Add the following to your configuration file in your kv_namespaces array:
kv_namespaces = [
  { binding = "MY_KV", id = "e29b263ab50e42ce9b637fa8370175e8" }
]
```

But notice that we're treating what the user expects to be their _resource name_ as the name of the binding, with the actual name given to the resource being some decorated variation of that. As a result, we also artificially limit what a user can name their resource via this command by what constitutes a valid JS identifier (e.g. no dashes are allowed here, though KV namespaces can indeed contain dashes in their name).

We should fix this so that the resource is created with the name the user intended. We should output the resource name along with its ID. We can still output a helpful snippet for the user to paste into `wrangler.toml`, albeit leaving it up to them to replace the binding name with what they want.

```shell
wrangler kv:namespace create "my-kv-namespace"
🌀 Creating KV namespace...
✨ Success!
Created the following KV namespace:
    my-kv-namespace: e29b263ab50e42ce9b637fa8370175e8

To start using this resource in your worker, add an entry to `kv_namespaces` in your wrangler.toml file:

kv_namespaces = [
  { binding = "<VARIABLE_NAME>" id = "e29b263ab50e42ce9b637fa8370175e8" }
]
```

nora-soderlund:
Fixes #

**What this PR solves / how to test:**  Currently the kv:namespace create command differs from other `create` commands where it prefixes the input with `worker-` when using the command in e.g. a pages project without a `wrangler.toml` or with a `wrangler.toml` but without a `name` property. This is a regression to how the dashboard and other products works and can cause confusion for developers.

**Author has addressed the following:** This has been broken into two separate PRs to focus on two different topics, see #5050 for the original change.

- Tests
  - [X] Included
  - [ ] Not necessary because:
  - [X] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s): 
  - [x] Not necessary because: the docs do not mention the overly zealous validation of the title passed to the command.

### DIFF
--- a/packages/wrangler/src/kv/index.ts
+++ b/packages/wrangler/src/kv/index.ts
@@ -22,6 +22,7 @@ import {
 	getKVKeyValue,
 	getKVNamespaceId,
 	isKVKeyValue,
+	isValidKVNamespaceBinding,
 	listKVNamespaceKeys,
 	listKVNamespaces,
 	putKVBulkKeyValue,
@@ -60,10 +61,18 @@ export function kvNamespace(kvYargs: CommonYargsArgv) {
 					);
 				}
 
-				const name = config.name || "worker";
-				const environment = args.env ? `-${args.env}` : "";
+				if (!isValidKVNamespaceBinding(args.namespace)) {
+					throw new CommandLineArgsError(
+						`The namespace binding name "${args.namespace}" is invalid. It can only have alphanumeric and _ characters, and cannot begin with a number.`
+					);
+				}
+
+				const name = config.name || "";
+				const environment = args.env ? `${name && "-"}${args.env}` : "";
 				const preview = args.preview ? "_preview" : "";
-				const title = `${name}${environment}-${args.namespace}${preview}`;
+				const title = `${name}${environment}${(name || environment) && "-"}${
+					args.namespace
+				}${preview}`;
 
 				const accountId = await requireAuth(config);
 
--- a/.changeset/tall-planets-pull.md
+++ b/.changeset/tall-planets-pull.md
@@ -1,0 +1,12 @@
+---
+"wrangler": major
+---
+
+feat: remove the default 'worker' prefix from the `kv:namespace create` command
+
+Previously, if there was no Worker defined by a local `wrangler.toml` or that Worker had no `name` property
+then any KV namespace that was created by Wrangler would use `worker_` as a prefix, rather than the Worker's
+name.
+
+This is a minor breaking change to the name given to a newly created KV namespace, which is unlikely to affect
+many, if any, developers.
--- a/packages/wrangler/src/__tests__/kv.test.ts
+++ b/packages/wrangler/src/__tests__/kv.test.ts
@@ -113,10 +113,10 @@ describe("wrangler", () => {
 			});
 
 			it("should create a namespace", async () => {
-				mockCreateRequest("worker-UnitTestNamespace");
+				mockCreateRequest("UnitTestNamespace");
 				await runWrangler("kv:namespace create UnitTestNamespace");
 				expect(std.out).toMatchInlineSnapshot(`
-			          "🌀 Creating namespace with title \\"worker-UnitTestNamespace\\"
+			          "🌀 Creating namespace with title \\"UnitTestNamespace\\"
 			          ✨ Success!
 			          Add the following to your configuration file in your kv_namespaces array:
 			          { binding = \\"UnitTestNamespace\\", id = \\"some-namespace-id\\" }"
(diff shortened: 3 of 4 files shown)


packages/wrangler/src/kv/index.ts
```diff
@@ -87,9 +82,7 @@ export function kvNamespace(kvYargs: CommonYargsArgv) {
 				logger.log(
 					`Add the following to your configuration file in your kv_namespaces array${envString}:`
 				);
-				logger.log(
-					`{ binding = "${args.namespace}", ${previewString}id = "${namespaceId}" }`
-				);
+				logger.log(`{ binding = "KV", ${previewString}id = "${namespaceId}" }`);
```
petebacondarwin, line 85: I think it would be good to use the `args.namespace` if possible. Perhaps removing all invalid characters?

packages/wrangler/src/kv/index.ts
```diff
@@ -53,23 +52,19 @@ export function kvNamespace(kvYargs: CommonYargsArgv) {
...
 				const config = readConfig(args.config, args);
 				if (!config.name) {
 					logger.warn(
 						"No configured name present, using `worker` as a prefix for the title"
 					);
 				}
 
-				const name = config.name || "worker";
-				const environment = args.env ? `-${args.env}` : "";
+				const name = config.name || "";
```
petebacondarwin, line 62: I think this is arguably a breaking change, since users might be expecting their KVs to have a particular shape for their title.

petebacondarwin (changes requested):

packages/wrangler/src/kv/index.ts
nora-soderlund: I understand what you're saying but I wouldn't classify it as breaking since it won't actually break anything. I think I would arguably say that this a regression in comparison to how the dashboard works and how other products work, e.g. creating a D1 database doesn't prepend it with `worker-`.

Do you want to discuss this change internally before we move on? :)

packages/wrangler/src/kv/index.ts
nora-soderlund: I can do that. Would it be ideal to maybe have a shared function for the other products to do the same thing with the example binding output?

packages/wrangler/src/kv/index.ts
Cherry: I'd definitely recommend a change like this land in 4.x, just in case anyone is relying on this behaviour. The `worker-` prefix expectation is even explicitly logged and called out [a few lines before](https://github.com/cloudflare/workers-sdk/pull/5007/files/0ba7867ed98d521fd222893dde3cf4aff7f5a5b7#diff-e7b74168762ef6aa3ecc8734c44ae185480aa7a3cfb58d92f2b66a2fa7151760R58): `"No configured name present, using `worker` as a prefix for the title"` when there's no `name` config set. The basic "create" example on the docs page also shows it working this way, though doesn't call it out explicitly: https://developers.cloudflare.com/kv/reference/kv-commands/#create-command-to-create-a-kv-namespace-called-my_kv

```
 ⛅️ wrangler 3.28.3
-------------------
▲ [WARNING] No configured name present, using `worker` as a prefix for the title


🌀 Creating namespace with title "worker-test1"
✨ Success!
```

nora-soderlund: After some discussions, I can understand this being a breaking change :) I wasn't considering the case of this being used in any CI/CD flows.

Considering the discussions in this pull request, I will withdraw the original fix (namespace name restrictions) and create another pull request for that, and we can change this PR to propose a major change for removing the `worker-` prefix.

packages/wrangler/src/kv/index.ts
nora-soderlund: I've split this PR into two to focus on two separate changes, the original fix is now in #5050.

petebacondarwin:
Rebased and added to the V4 milestone.

petebacondarwin (approved):

PR #5009, opened 2024-02-13
C3: Update Next.js template (to use the latest tooling from next-on-pages)

dario-piotrowicz:

This PR updates the C3 Next.js template so that it uses the latest `@cloudflare/next-on-pages` tooling (i.e. `setupDevPlatform` and `getRequestContext`)

- Tests
  - [ ] Included
  - [x] Not necessary because: the Next.js scaffolding is already being tested
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/issues/12972
  - [ ] Not necessary because:

**Note for PR author:**


packages/create-cloudflare/src/cli.ts
```diff
@@ -138,13 +138,13 @@ const configure = async (ctx: C3Context) => {
...
 	await installWorkersTypes(ctx);
 
-	await updateWranglerToml(ctx);
-
 	const { template } = ctx;
 	if (template.configure) {
 		await template.configure({ ...ctx });
 	}
 
+	await updateWranglerToml(ctx);
```
dario-piotrowicz, line 146: @jculvey I'm moving this because we're creating the toml file in the `configure` phase, what do you think?

Alternatively if it's better I can put the toml file creation into `generate` and leave this as is

what do you prefer?

jculvey: I don't think we can move it, because the `pre-existing` template needs to run wrangler commands to download the existing script file in the configure phase, and this will fail if the name is still `<TBD>`. Can you try running this with `--existing-script` and see if it fails?

`copyFiles` can normally happen because it happens in the generation step. I guess we have to manually copy over the wrangler file because of the `destinationDir` being either `.` or `src/`. I think moving this code to `generate` is the better option.

jculvey: Now that I think about it, we should definitely leave a comment here on why this function call needs to be in this particular place. Silly me for not leaving one last time 😓

dario-piotrowicz: Don't we have any tests that would fail? 😓 

I guess we don't have e2es for existing scripts uh?

packages/create-cloudflare/templates/next/env.d.ts
```diff
@@ -1,13 +1,7 @@
...
-  namespace NodeJS {
-    interface ProcessEnv {
-      // Add here the Cloudflare Bindings you want to have available in your application
-      // (for more details on Bindings see: https://developers.cloudflare.com/pages/functions/bindings/)
-      //
-      // KV Example:
-      // MY_KV: KVNamespace
-    }
-  }
+interface CloudflareEnv extends Env {
```
IgorMinar, line 1: isn't this file going to conflict with the output from `wrangler types`?

packages/create-cloudflare/templates/next/wrangler.toml
```diff
@@ -0,0 +1,55 @@
+name = "<TBD>"
+compatibility_date = "<TBD>"
```
IgorMinar, line 2: don't we also need to enable the nodejs_compat flag by default to make ALS work in Next which natively depends on it?

IgorMinar (approved):
this looks very good!

packages/create-cloudflare/templates/next/env.d.ts
dario-piotrowicz: yes good point.... let me remove it for now....

then later we can re-enable it when `wrangler types` allows us to directly set the interface name `CloudflareEnv` and the file path `env.d.ts`

packages/create-cloudflare/templates/next/wrangler.toml
dario-piotrowicz: yeah, I just left it in the `pages:preview` script, let me move it here, thanks for noticing it 🙂

dario-piotrowicz: compat flag added 🙂👍

jculvey (approved):

PR #5015, opened 2024-02-14
[do not merge] Don't strip version numbers from requirements.txt

garrettgu10:
Currently we strip version numbers from requirements.txt when translating to a worker config (and that's probably what we want). However, for now we need to specify the version number to make some packages work within pyodide and workerd.


petebacondarwin (approved):

petebacondarwin:
@garrettgu10 - what do you want to do with this PR? It needs a rebase and is currently set to "is draft".

PR #5016, opened 2024-02-14
[quick-edit] Partitioned cookies

penalosa:
Closes [DEVX-1156](https://jira.cfdata.org/browse/DEVX-1156). 

This PR ensures that the Workers Quick Edit uses partitioned cookies for it's preview system, to address the [upcoming deprecation of third party cookies in Chrome](https://developers.google.com/privacy-sandbox/3pcd/chips).

_Testing instructions TBD_

- Tests
  - [x] Included, but manual testing in Chrome TBD
  - [ ] Not necessary because:
  - [ ] Included
  - [x] Not necessary because: The package doesn't follow changesets
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: This isn't a change to documented behaviour

### DIFF
--- a/packages/wrangler/templates/startDevWorker/ProxyWorker.ts
+++ b/packages/wrangler/templates/startDevWorker/ProxyWorker.ts
@@ -305,10 +305,12 @@ function insertLiveReloadScript(
  */
 function rewriteUrlRelatedHeaders(headers: Headers, from: URL, to: URL) {
 	headers.forEach((value, key) => {
-		if (typeof value === "string" && value.includes(from.host)) {
+		if (typeof value === "string" && value.includes(from.hostname)) {
 			headers.set(
 				key,
-				value.replaceAll(from.origin, to.origin).replaceAll(from.host, to.host)
+				value
+					.replaceAll(from.origin, to.origin)
+					.replaceAll(from.hostname, to.hostname)
 			);
 		}
 	});
--- a/packages/wrangler/templates/startDevWorker/ProxyWorker.ts
+++ b/packages/wrangler/templates/startDevWorker/ProxyWorker.ts
@@ -304,13 +304,22 @@ function insertLiveReloadScript(
  * so that this proxy is transparent to the Client Browser and User Worker.
  */
 function rewriteUrlRelatedHeaders(headers: Headers, from: URL, to: URL) {
+	const setCookie = headers.getAll("Set-Cookie");
+	headers.delete("Set-Cookie");
+	for (const cookie of setCookie) {
+		headers.append(
+			"Set-Cookie",
+			cookie.replace(
+				new RegExp(`Domain=${from.hostname}($|;|,)`),
+				`Domain=${to.hostname}$1`
+			)
+		);
+	}
 	headers.forEach((value, key) => {
-		if (typeof value === "string" && value.includes(from.hostname)) {
+		if (typeof value === "string" && value.includes(from.host)) {
 			headers.set(
 				key,
-				value
-					.replaceAll(from.origin, to.origin)
-					.replaceAll(from.hostname, to.hostname)
+				value.replaceAll(from.origin, to.origin).replaceAll(from.host, to.host)
 			);
 		}
 	});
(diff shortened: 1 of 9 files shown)


penalosa:
@petebacondarwin @RamIdeas, this contains a change to https://github.com/cloudflare/workers-sdk/pull/4950, which was breaking `set-cookie` headers (which require a `hostname`, not `host`). I'm not entirely sure about the change though, so would appreciate a review!

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
```diff
@@ -305,10 +305,12 @@ function insertLiveReloadScript(
...
 function rewriteUrlRelatedHeaders(headers: Headers, from: URL, to: URL) {
 	headers.forEach((value, key) => {
-		if (typeof value === "string" && value.includes(from.host)) {
+		if (typeof value === "string" && value.includes(from.hostname)) {
 			headers.set(
 				key,
-				value.replaceAll(from.origin, to.origin).replaceAll(from.host, to.host)
+				value
+					.replaceAll(from.origin, to.origin)
+					.replaceAll(from.hostname, to.hostname)
```
petebacondarwin, line 313: I think we also want to replace the host as well, in case the port has changed...

```suggestion
					.replaceAll(from.origin, to.origin)
					.replaceAll(from.host, to.host)
					.replaceAll(from.hostname, to.hostname)
```

petebacondarwin: I am a little worried though that we would, for example, now replace all `localhost` references in all headers whether or not they are actually referencing the correct host (e.g. you might have a header that refers to some other port on localhost and this would get swapped.

I wonder if we need to special case the "cookie" header in this case?

penalosa: I think we might need to, yes. Browsers don't seem to accept `Set-Cookie` on localhost with a port specified, and in any case cookies are shared across ports in general.

1000hz:
I tested these changes locally and verified they are working.

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
```diff
@@ -304,8 +304,23 @@ function insertLiveReloadScript(
  * so that this proxy is transparent to the Client Browser and User Worker.
  */
 function rewriteUrlRelatedHeaders(headers: Headers, from: URL, to: URL) {
+	const setCookie = headers.getAll("Set-Cookie");
+	headers.delete("Set-Cookie");
+	for (const cookie of setCookie) {
```
petebacondarwin, line 309: NIT - if you put this for loop after the headers.forEach below then you wouldn't have to check that the key is not `set-cookie`.

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
```diff
@@ -304,8 +304,23 @@ function insertLiveReloadScript(
...
+				`Domain=${to.hostname}$1`
+			)
+		);
+	}
 	headers.forEach((value, key) => {
-		if (typeof value === "string" && value.includes(from.host)) {
+		if (
+			typeof value === "string" &&
+			value.includes(from.host) &&
+			key !== "set-cookie"
```
petebacondarwin, line 322: I think you need to lowercase `key` here since it wouldn't match `Set-Cookie`.

petebacondarwin (changes requested):

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
penalosa: Happy to move it to after the `.forEach()`, but I think the check will still be required. The `.forEach()` calls `headers.set()` which AFAIK will clear all but the first `Set-Cookie` (while this section of code calls `headers.append()`)

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
penalosa: Headers are case-insensitive, and judging by the docs `headers.forEach()` will lowercase header name: https://developer.mozilla.org/en-US/docs/Web/API/Headers/forEach

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
penalosa: It shouldn't—this code should delete the original header, but then loop over the existing set-cookies to rewrite the domain

petebacondarwin (approved):

petebacondarwin:
Love it! But sadly legit failures in CI to deal with before landing.

Issue #4411, opened 2023-11-09
🐛 BUG: `wrangler pages dev` cannot detect `compatibility_date` from `wrangler.toml`


rschristian:
### Which Cloudflare product(s) does this pertain to?

Wrangler core

### What version(s) of the tool(s) are you using?

3.15.0

### What version of Node are you using?

18.18.0

### What operating system are you using?

Linux

### Describe the Bug

Upon running my dev command (`wrangler pages dev src`), every single time, I'm met with the same "No compatibility_date was specified", despite having it listed in my `wrangler.toml` as the docs link to. Setting the flag (`--compatability-date=...`) does silence that pesky warning but it's quite verbose and should be unnecessary.

### Please provide a link to a minimal reproduction

https://github.com/rschristian/bug__wrangler-compatibility-date

### Please provide any relevant error logs

```
$  yarn start

yarn run v1.22.19
$ wrangler pages dev src
▲ [WARNING] No compatibility_date was specified. Using today's date: 2023-11-09.

  Pass it in your terminal:
  .```
  --compatibility-date=2023-11-09
  .```
  See https://developers.cloudflare.com/workers/platform/compatibility-dates/ for more
  information.


Compiling worker to "/home/ryun/Projects/tmp/.wrangler/tmp/pages-Jxzdeo/functionsWorker-0.007770138516417946.mjs"...
✨ Compiled Worker successfully
 ⛅️ wrangler 3.15.0
-------------------
▲ [WARNING] --local is no longer required and will be removed in a future version.

  `wrangler dev` now uses the local Cloudflare Workers runtime by default. 🎉


⎔ Starting local server...
[mf:inf] Ready on http://0.0.0.0:8788
[mf:inf] - http://127.0.0.1:8788
[mf:inf] - http://192.168.100.128:8788
```

Binero:
For what it is worth, I was able to reproduce this.

xdivby0:
I have similar issues and I think in wrangler 2 this wasn't the case, just happened when I upgraded to 3

penalosa:
`pages dev` doesn't currently fully support `wrangler.toml`, although we're currently working on it (see https://github.com/cloudflare/workers-sdk/pull/5284).

PR #5027, opened 2024-02-15
[C3] Improve bindings support in Svelte template 

jculvey:

- Updates the Svelte template add a hook to populate the `platform` object with `getPlatformProxy` in dev.
- Adds a default `wrangler.toml` file
- Adds a `build-cf-types` script to generate bindings types via `wrangler types`. 
- Updates the `app.d.ts` to include types for `ctx` and `cf` in addition to the `env` type described above
- Improves the reliability of parsing a project's `tsconfig.json` when adding the `workers-types` entrypoint by switching the implementation to `jsonc-parser`

- Tests
  - [ x ] Included
  - [ ] Not necessary because:
  - [ x ] Included
  - [ ] Not necessary because:
- Associated docs
  - [ x ] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/issues/12897
  - [ ] Not necessary because:

**Note for PR author:**


packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -468,6 +478,9 @@ const verifyBuildScript = async (
 		[pm, "run", script],
 		{
 			cwd: projectPath,
+			env: {
+				NODE_ENV: "production",
```
dario-piotrowicz, line 482: ooc: why is this needed? 😮

packages/create-cloudflare/templates/svelte/c3.ts
```diff
@@ -39,39 +45,83 @@ const configure = async (ctx: C3Context) => {
...
 	configure,
-	transformPackageJson: async () => ({
-		scripts: {
-			"pages:preview": `${npm} run build && wrangler pages dev ${await compatDateFlag()} .svelte-kit/cloudflare`,
-			"pages:deploy": `${npm} run build && wrangler pages deploy .svelte-kit/cloudflare`,
-		},
-	}),
+	transformPackageJson: async (original: PackageJson, ctx: C3Context) => {
+		let scripts: Record<string, string> = {
+			preview: `${npm} run build && wrangler pages dev ${await compatDateFlag()} .svelte-kit/cloudflare`,
```
dario-piotrowicz, line 110: is the compat date necessary when we have it in the toml file?

(I sure hope not because I did remove it in the Next.js template 😅)

packages/create-cloudflare/src/templates.ts
```diff
@@ -11,7 +11,12 @@ import degit from "degit";
 import { C3_DEFAULTS } from "helpers/cli";
 import { readJSON, usesTypescript, writeJSON } from "helpers/files";
 import { validateTemplateUrl } from "./validators";
-import type { C3Args, C3Context } from "types";
+import type { C3Args, C3Context, PackageJson } from "types";
+
+export type PackageTransformer = (
+	pkgJson: PackageJson,
+	ctx: C3Context
+) => Promise<Record<string, string | object>>;
```
dario-piotrowicz, line 19: is this used anywhere?

Screenshot 2024-02-15 at 23 02 29

packages/create-cloudflare/templates/svelte/c3.ts
```diff
@@ -39,39 +45,83 @@ const configure = async (ctx: C3Context) => {
...
+			deploy: `${npm} run build && wrangler pages deploy .svelte-kit/cloudflare`,
+		};
+
+		if (usesTypescript(ctx)) {
+			scripts = {
+				...scripts,
+				"build-cf-types": "wrangler types && mv worker-configuration.d.ts src/",
```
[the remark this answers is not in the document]
jculvey: It is, unfortunately adding this file to `include` in `tsconfig.json` doesn't work since the config inherits, and adding `include` overwrites rather than merges with inherited settings. I can't think of another way around this until `wrangler types` gets the enhancement we talked about.

packages/create-cloudflare/e2e-tests/frameworks.test.ts
jculvey: This is needed because without it the `NODE_ENV` is set to `test`, which it inherits from the `vitest` call. When `svelte` is built with `NODE_ENV=test` it fails.

packages/create-cloudflare/templates/svelte/c3.ts
jculvey: That's a great point 🙈

packages/create-cloudflare/src/templates.ts
jculvey: Good catch. It was gonna be, but it ended up being worse than what we have.

packages/create-cloudflare/templates/svelte/c3.ts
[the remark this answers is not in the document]
dario-piotrowicz: isn't this going to be a problem on non unix machines?

packages/create-cloudflare/e2e-tests/frameworks.test.ts
dario-piotrowicz: ah ok, thanks for the clarification 🙂

packages/create-cloudflare/templates/svelte/c3.ts
dario-piotrowicz: PS: it might not be that simple thought 😓 

the compat date might not actually properly work with Pages: https://github.com/cloudflare/workers-sdk/issues/4411

I haven't had a chance I should investigate that issue 😓

packages/create-cloudflare/src/workers.ts
```diff
@@ -105,32 +106,46 @@ export async function addWorkersTypesToTsConfig(ctx: C3Context) {
...
-				...(tsConfigJson.compilerOptions.types ?? []),
-				typesEntrypoint,
-			];
-			updated = JSON.stringify(tsConfigJson, null, 2);
-		} catch {
-			warn("Could not parse tsconfig.json file");
-			updated = tsconfig;
-		}
+		const explicitEntrypoint = (currentTypes as string[]).some((t) =>
+			t.match(/@cloudflare\/workers-types\/\d{4}-\d{2}-\d{2}/)
```
dario-piotrowicz, line 114: Nit: should we also consider `@cloudflare/workers-types/experimental`?

dario-piotrowicz (approved):
Looks good to me 🙂 

Besides the failing test and `mv`

PR #5030, opened 2024-02-16
[wrangler] fix: don't suggest reporting `UserError`s to GitHub

### LINKED ISSUE #3867: 🐛 BUG: `whoami` results in an fatal error when the session is expired

Wrangler core

3.6.0

18

Mac

### Describe the Bug

When the current refresh token is expired and the user invokes `wrangler whoami`, the command results in an error:

(I only have a c3 version of the error, but the same error appears also when wrangler is invoked directly with an expired token)

Screenshot 2023-08-24 at 18 16 25

It is unsurprising that invoking the command results in an error if the current session is invalid. What is surprising is that we are displaying this error as if it was an internal error or an unexpected crash. Notice the following part:

```
 If you think this is a bug, please open an issue at:
  https://github.com/cloudflare/workers-sdk/issues/new/choose
```

This seems to indicate to the user that this error is in some way unexpected, but that's not the case at all. 

We know up front that calling `whoami` can result in three possible outcomes:
1. success: the user is logged in, we list user info
2. fail: the user is not logged in, we should simply notify the user and suggest that they use `wrangler login` to log in
3. crash: some internal error, in which case we should print any kind of contextual info about the error and suggest that the user files a bug

Note that we have two types of failure modes, one that is expected, and one that is unexpected.

Could we change the output of `whoami` (and elsewhere if there are other places like this) to make the expected failure mode look distinct (and more helpful, e.g. by suggesting `wrangler login`) from crashes due to an internal error?

_No response_

_No response_

mrbbot:
Fixes #3867.

I tried to reproduce #3867 by updating my `.wrangler/config/default.toml` with an access token expiry in the past, and an invalid refresh token, but `wrangler whaomi` just opened a browser and prompted me to login. It's possible this specific issue was fixed in another PR. 🤔 Either way, given the recent work to split out `UserError`s, of which the `Not logged in.` error seen here is one, it seemed like a good idea to hide the bug reporting message in these cases.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: changing error messaging

**Note for PR author:**

### DIFF
--- a/.changeset/four-comics-battle.md
+++ b/.changeset/four-comics-battle.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: don't suggest reporting user errors to GitHub
+
+Wrangler has two different types of errors: internal errors caused by something going wrong, and user errors caused by an invalid configuration. Previously, we would encourage users to submit bug reports for user errors, even though there's nothing we can do to fix them. This change ensures we only suggest this for internal errors.
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -1609,10 +1609,7 @@ addEventListener('fetch', event => {});`
 			                - \\"site.bucket\\" is a required field."
 		            `);
 
-			expect(std.out).toMatchInlineSnapshot(`
-			        "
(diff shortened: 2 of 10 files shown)


petebacondarwin (approved):
Love it!

Issue #3907, opened 2023-09-07
🐛 BUG: 'open a browser' still opens custom routes/domains in wrangler dev --remote


KianNH:
### Which Cloudflare product(s) does this pertain to?

Wrangler core

### What version(s) of the tool(s) are you using?

3.7.0

### What version of Node are you using?

18.17.0

### What operating system are you using?

Windows

### Describe the Bug

Duplicate of https://github.com/cloudflare/workers-sdk/issues/2691, which is a duplicate of https://github.com/cloudflare/workers-sdk/issues/2610

The 'open a browser' hotkey still opens the route/custom domain, which takes the user away from their local proxy & usually doesn't load because the default port of `8787` isn't something that Cloudflare listens on.

![image](https://github.com/cloudflare/workers-sdk/assets/94662631/5417c7fb-933f-422f-95cd-339214605c3a)

KianNH:
The `port = 0` behaviour still happens too.

![image](https://github.com/cloudflare/workers-sdk/assets/94662631/186cea1e-531d-4697-a5e1-1e6975ac2697)

KianNH:
It looks like this should have been fixed by https://github.com/cloudflare/workers-sdk/pull/3337 but it seems like there has been a regression (or that fix didn't work), as none of the reported behaviour in the historic issues has been fixed.

mrbbot:
Hey! 👋 Thanks for reporting this. I was able to reproduce the `--port 0` issue with `--remote` mode only. Will get that fixed. 👍 

---

I wasn't able to reproduce the `route` issue though with the following `wrangler.toml` and `--remote`:

```toml
main = "index.mjs"
compatibility_date = "2024-01-01"

[route]
pattern = "miniflare.dev"
custom_domain = true
```

Could you try this again on the latest version of `wrangler`, and if this is still a problem, share a minimal `wrangler.toml` reproduction?

lrapoport-cf:
hey @KianNH , did @mrbbot 's PR (https://github.com/cloudflare/workers-sdk/pull/5033) address the issue for you? if so we will close this out, else can you share a minimal wrangler.toml reproduction?

lrapoport-cf:
hey @KianNH :) we’re going to close this issue for now, but if this is still a problem, please feel free to open a new issue referencing this one. please be sure to include an updated summary and any additional requested details, followup, or context and we will revisit and investigate further. thanks!

PR #5033, opened 2024-02-16
fix: wait for actual port before opening browser with `--port=0`

mrbbot:
Fixes part of #3907.

When running `wrangler dev --port=0`, Wrangler doesn't know the port the dev server started up on until the `onReady` callback is called. In `--remote` mode, this can take a while meaning there's a significant period of time where pressing <kbd>b</kbd> opens `localhost:0`. This change ensures we wait until the first `onReady` call before opening the browser with the correct port.

- Tests
  - [ ] Included
  - [x] Not necessary because: should ideally have some, but the hotkeys are only testable when stdin is a TTY, meaning we'd need to write these in `interactive-dev-tests`. But the behaviour is only observable when using `--remote` mode which requires Cloudflare account credentials. We could write a unit test for this, but again this is tricky because `useHotkeys()` relies on the `useInput()` hook from Ink, which we'd need to mock. This behaviour also relies on module-level variables. 😬 
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: fixing a regression

**Note for PR author:**

### DIFF
--- a/.changeset/khaki-wasps-glow.md
+++ b/.changeset/khaki-wasps-glow.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: wait for actual port before opening browser with `--port=0`
+
+Previously, running `wrangler dev --remote --port=0` and then immediately pressing `b` would open `localhost:0` in your default browser. This change queues up opening the browser until Wrangler knows the port the dev server was started on.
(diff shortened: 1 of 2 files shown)


petebacondarwin (approved):
Pulled this down and tested locally, given that we don't have any tests for user interactions in `wrangler dev`.

petebacondarwin:
The only thing I would note is that we are not told that Wrangler is waiting for the port to become available.
So I ended up pressing `b` multiple times, thinking it had not worked. And so subsequently got about 5 browsers all pop up at once!

PR #5042, opened 2024-02-16
Add `wrangler types` customization and improvements

dario-piotrowicz:

This PR:
 - adds a positional path argument to `wrangler types` alongside an optional `--env-interface` flag

   The former allows you to specify where to save the file and the latter allows you to customize the name of the generated interface

   Example:
   ```sh
   wrangler types --env-interface CloudflareEnv ./src/env.d.ts
   ```
 - makes the command code more consistent with (most of) the rest of the other commands
 - makes the command honor the `-c`/`--config` flag (currently ignored)
 - makes the command present a warning if no config file is detected
 - makes the command pick up secrets from `.dev.vars`
 
- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/pull/13040
  - [ ] Not necessary because:

**Note for PR author:**


packages/wrangler/src/type-generation.ts
```diff
@@ -187,7 +313,7 @@ function writeDTSFile({
 
 	if (envTypeStructure.length || modulesTypeStructure.length) {
 		fs.writeFileSync(
-			"worker-configuration.d.ts",
+			path,
 			`// Generated by Wrangler on ${new Date()}` + "\n" + combinedTypeStrings
```
mrbbot, line 317: Could this comment include the full command used to generate the file, so the user can easily regenerate it again in the future?

dario-piotrowicz: Yes, let me quickly do that 🙂 :+1:

.changeset/curly-masks-raise.md
```diff
@@ -0,0 +1,17 @@
+---
+"wrangler": minor
+---
+
+feat: add new `path` positional argument to `wrangler types`
+
+Add the possibility for users to specify the d.ts file they want
```
petebacondarwin, line 7:
```suggestion
Allow users to specify the path to the typings (.d.ts) file they want
```

.changeset/itchy-dragons-burn.md
```diff
@@ -0,0 +1,10 @@
+---
+"wrangler": patch
+---
+
+fix: make `wrangler types` honor top level config argument
+
+the `wrangler types` command currently ignores the `-c|--config` argument
```
petebacondarwin, line 7:
```suggestion
The `wrangler types` command currently ignores the `-c|--config` argument
```

.changeset/itchy-dragons-burn.md
```diff
@@ -0,0 +1,10 @@
...
+(although it is still getting shown in the command's help message), make
+sure that the command honors the flag, also if no config file is detected
```
petebacondarwin, line 9:
```suggestion
sure that the command honors the flag.
Also, if no config file is detected
```

.changeset/serious-socks-tan.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+fix: make `wrangler types` pick up secrets from `.dev.vars`
```
petebacondarwin, line 5:
```suggestion
fix: make the `wrangler types` command pick up local secret keys from `.dev.vars`
```

.changeset/serious-socks-tan.md
```diff
@@ -0,0 +1,9 @@
...
+
+Make sure that the `wrangler types` command correctly picks up
+secrets defined in `.dev.vars` and includes them in the generated
```
petebacondarwin, line 8:
```suggestion
secret keys defined in `.dev.vars` and includes them in the generated
```

packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -255,4 +325,220 @@ describe("generateTypes()", () => {
 		"
 	`);
 	});
+
+	it("should include secrets from .dev.vars", async () => {
```
petebacondarwin, line 329:
```suggestion
	it("should include secret keys from .dev.vars", async () => {
```

packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -255,4 +325,220 @@ describe("generateTypes()", () => {
...
+		"
+	`);
+	});
+
+	it("should override vars with secrets", async () => {
+		fs.writeFileSync(
+			"./wrangler.toml",
+			TOML.stringify({
+				vars: {
+					MY_VARIABLE: "my variable",
```
petebacondarwin, line 369: Perhaps make this var a complex object instead of a string, so that we can see that it really does get overridden?

```suggestion
					MY_VARIABLE: { NAME = "foo" },
```

packages/wrangler/src/type-generation.ts
```diff
@@ -1,16 +1,115 @@
...
+	) {
+		logger.warn(
+			`No config file detected${
+				args.config ? ` (at ${args.config})` : ""
+			}, aborting`
+		);
+		return;
+	}
+
+	const config = readConfig(args.config, args);
```
petebacondarwin, line 69: Given that we have already found the configPath above then there is no point in asking `readConfig()` to do it again internally. Just pass the `configPath` in here.

packages/wrangler/src/type-generation.ts
```diff
@@ -1,16 +1,115 @@
...
+
+	const configBindings: Partial<Config> = {
+		kv_namespaces: config.kv_namespaces ?? [],
+		vars: { ...config.vars },
+		wasm_modules: config.wasm_modules,
+		text_blobs: {
+			...config.text_blobs,
```
petebacondarwin, line 76: Out of interest, do you know why we are cloning this object?

packages/wrangler/src/type-generation.ts
```diff
@@ -1,16 +1,115 @@
...
+		rules: config.rules,
+		queues: config.queues,
+		constellation: config.constellation,
+	};
+
+	const secrets = getVarsForDev(
+		{ configPath, vars: {} },
+		args.env,
+		true
+	) as Record<string, string>;
```
petebacondarwin, line 96: If you did this bit before defining the object above, you could just include it directly as a property rather than having to spread it below in the call to `generateType()`.

packages/wrangler/src/type-generation.ts
```diff
@@ -19,6 +118,20 @@ export async function generateTypes(
...
+	// Note: we infer whether the user has provided an envInterface by checking
+	//       if it is different from the default `Env` value, this works well
+	//       besides the fact that the user itself can actually provided `Env` as
+	//       an argument... we either need to do this or removing the yargs
+	//       default value for envInterface and do `envInterface ?? "Env"`,
+	//       for a better UX we chose to go with the yargs default value
+	const userProvidedEnvInterface = envInterface !== "Env";
+
+	if (userProvidedEnvInterface && entrypointFormat === "service-worker") {
+		logger.warn(
```
petebacondarwin, line 130: Actually I think this should be a fatal error for this command.
It would be confusing for a developer to provide a the `--env-interface` argument but for it to still output `interface Env`.

packages/wrangler/src/type-generation.ts
```diff
@@ -28,8 +141,11 @@ export async function generateTypes(
 	}
 
 	if (configToDTS.vars) {
-		for (const varName in configToDTS.vars) {
-			const varValue = configToDTS.vars[varName];
+		// Note: vars get overridden by secrets, so should their types
+		const vars = Object.entries(configToDTS.vars).filter(
+			([key]) => !(key in configToDTS.secrets)
+		);
```
petebacondarwin, line 147: Could we simplify this by doing something like:

```
const vars = Object.assign({}, configToDTS.vars, configToDTS.secrets);
```

or 

```
const vars = {...configToDTS.vars, ...configToDTS.secrets};
```


Then you don't need to add the secrets below either, right?

packages/wrangler/src/type-generation.ts
```diff
@@ -165,7 +291,7 @@ function writeDTSFile({
 				.includes("Generated by Wrangler")
 		) {
 			throw new UserError(
-				"A non-wrangler worker-configuration.d.ts already exists, please rename and try again."
+				`A non-wrangler ${path} already exists, please rename and try again.`
```
petebacondarwin, line 294: ASIDE: it feels to me like a common usage would be to run this `wrangler types` command quite often, particularly every time a change is made to `wrangler.toml` or `.dev.vars`. This error requires them to delete the old file first, right?
I wonder if we could offer some `--force` option that will always overwrite the file without an error.

DaniFoldi: Fortunately the implementation already thought of this use case, and there is a condition above this `throw` that checks if the dts file was `Generated by Wrangler` (meaning it contains this string). Can confirm this works nicely, we regenerate the file regularly in CI to check if it's consistent with wrangler config.

petebacondarwin: Oh sweet! Thanks for confirming.

packages/wrangler/src/type-generation.ts
dario-piotrowicz: Done, please have another look 🙂

fixtures/type-generation/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+	"include": [
+		"workers/hello-worker-a",
+		"module-worker-b",
+		"service-worker-a",
+		"module-worker-c",
+		"module-worker-d",
+		"pages-functions-app"
+	],
```
mrbbot, line 9: Should these `include`s be here?

dario-piotrowicz: nope! copy paste! sorry 🥲

packages/wrangler/src/__tests__/type-generation.test.ts
dario-piotrowicz: mh.... but instead of changing it I'd add this test if it's ok by you 🙂

I was testing the overriding because `wrangler types`, when it encounters a var string binding, it types is as a const string (e.g. `MY_VARIABLE: "my variable"` instead of `MY_VARIABLE: string`)

and what I wanted to test here is that when `wrangler types` encounters the same string var binding in both the toml and `.dev.vars` it goes with the generic `string` alternative instead of picking the const one

petebacondarwin: Ah I see. I didn't spot that.

packages/wrangler/src/type-generation.ts
dario-piotrowicz: > It would be confusing for a developer to provide a the `--env-interface` argument but for it to still output `interface Env`.

The command would not output a `interface Env`, it would simply not output an interface at all.

e.g. the output would look something like this:
```ts
export {};
declare global {
	const MY_KV: KVNamespace;
	const MY_VARIABLE: string;
}
```

Making this a fatal error feels a bit extreme to me... but I'm not completely against it either
(my reasoning here was that when you provide the env interface argument you are sort of asking the command to use that _if_ it does generate an env interface (which it does not if you're using the service worker syntax))

given my comment above, are you still convinced that the command should generate a fatal error?

packages/wrangler/src/type-generation.ts
dario-piotrowicz: good point 😅👍

packages/wrangler/src/type-generation.ts
dario-piotrowicz: now that you mention it, we have:
```ts
vars: { ...config.vars },
```
as well as:
```ts
kv_namespaces: config.kv_namespaces
```
and
```ts
durable_objects: config.durable_objects
```

etc...

a bit inconsistent 😕 (and also unnecessary I feel? 🤷)

packages/wrangler/src/type-generation.ts
dario-piotrowicz: yes, I preferred it this way as I find it more readable (keeping the "config" clearly separate from the "secrets"), but I don't mind changing this 🙂👍

packages/wrangler/src/type-generation.ts
dario-piotrowicz: As I mentioned in my comment above: https://github.com/cloudflare/workers-sdk/pull/5042/files#r1497428862

vars get written as const (`e.g. MY_VAR: "my variable"`) while secrets should definitely be kept generic (`e.g. MY_SECRET: string`)

That's why here I am filtering out the secrets and not combining them with the vars

packages/wrangler/src/type-generation.ts
petebacondarwin: Yeah I still think this should be an error. Think about it this way: `--env-interface` is not a valid option when it cannot be used.
Or to put it the other way, why would someone add this option if it is going to have no effect?

packages/wrangler/src/type-generation.ts
petebacondarwin: Feel free to leave it as is

mrbbot (approved):

packages/wrangler/src/type-generation.ts
dario-piotrowicz: I've changed this to error 🙂👍 

Please have another look 🙏 

(and please let me know if the error message is ok 🙏)

petebacondarwin (approved):

.changeset/beige-pants-promise.md
```diff
@@ -0,0 +1,8 @@
+---
+"wrangler": patch
+---
+
+feat: include command run in `wrangler types` comment
```
dario-piotrowicz, line 5:
```suggestion
feat: include command run in the `wrangler types` comment
```

shaunco:
@dario-piotrowicz way late with this comment, but why are secrets picked up from `.dev.vars` but not the `[secrets]` section of wrangler.toml?

PR #5043, opened 2024-02-16
Python playground support

penalosa:

Go to https://8f2e7cec.workers-playground.pages.dev/playground/python and you should be able to try it out! (the first time you visit you might get a "Valid token must be provided" error—just reload the page. Cookies are a bit funny in the preview deployments)

- Tests
  - [ ] Included
  - [x] Not necessary because: No tests have been added for the python functionality, since it's not yet stable enought to test realiably. There should be no affect on the JS functionality
  - [ ] Included
  - [x] Not necessary because: Not relevant
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: Not at a stage for docs

**Note for PR author:**


irvinebroque:
🙌 🔥

irvinebroque (approved):
🙌 👏 🚀

hoodmane (approved):
I didn't do a super detailed review but everything looks reasonable to me.

packages/workers-playground/functions/playground/[[route]].ts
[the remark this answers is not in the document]
Comment (penalosa) [reply packages/workers-playground/functions/playground/[[route]].ts:1]: I'll add comments to explain that!

dom96 (approved):
lgtm

packages/playground-preview-worker/src/user.do.ts
```diff
@@ -200,11 +200,24 @@ export class UserSession {
 			compatibility_flags: uploadedMetadata?.compatibility_flags ?? [
 				"nodejs_compat",
 			],
+			// TODO(soon): Add a CPU time limiter once standard pricing is out
+			usage_model: "unbound",
```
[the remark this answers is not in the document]
penalosa: The playground is unauthenticated, so this should work for all users

packages/playground-preview-worker/src/index.ts
```diff
@@ -201,15 +201,20 @@ app.get(`${rootDomain}/api/inspector`, async (c) => {
...
-	) {
+		if (
+			c.req.header("Sec-Fetch-Dest") !== "iframe" ||
+			!(
+				referer.hostname === "workers.cloudflare.com" ||
+				referer.hostname === "localhost" ||
+				referer.hostname.endsWith("workers-playground.pages.dev")
+			)
+		) {
+			throw new PreviewRequestForbidden();
```
mrbbot, line 215: I assume this double `throw` is to catch URL parsing failures? Maybe you want a `maybeParse()` function that returns `undefined` on failures, then you could include a check for this in the same `if` statement?

packages/workers-playground/functions/playground/[[route]].ts
```diff
@@ -1,3 +1,11 @@
+/**
+ * These functions are purely for local development, and when running a Pages preview. In production, requests go through the following pipeline:
+
+ * eyeball -> playground-preview-worker -> workers-playground
+ * However, locally, and in a Pages preview, requests go through this pipeline:
+
+ * eyeball ->  workers-playground (local or preview) -> playground-preview-worker (staging)
```
Comment (mrbbot) [inline packages/workers-playground/functions/playground/[[route]].ts:7]: Is the idea here that these functions make the local pipeline behave more like the remote? I guess I'm still a bit confused why these pipelines are different. This comment doesn't really explain _why_ we need these functions. It's possible I'm just being silly though.

Comment (penalosa) [reply packages/workers-playground/functions/playground/[[route]].ts:7]: These effectively allow you to run Wrangler pages dev in the local `workers-playground` repo and have it work and make sense. It's effectively the same as we have in the Vite config

mrbbot (approved):

packages/workers-playground/generate-default-hashes.ts
```diff
@@ -0,0 +1,131 @@
...
+	return `export default ${JSON.stringify(workerHashes)};\n`;
+}
+
+const pythonWorker = async () => {
+	const worker = new FormData();
+
+	const metadata = {
+		main_module: "index.py",
+		compatibility_date: `${year}-${month}-${date}`,
+		compatibility_flags: ["experimental"],
```
dom96, line 25: Need to change this to `python_workers`

irvinebroque:
Should we merge so that can start playing with this in production, etc? And then layer on other options as discussed in designs in separate PR?

PR #5050, opened 2024-02-17
[KV] Fix kv:namespace create name validation

### LINKED ISSUE #4248: `wrangler kv:namespace create <namespace>` treats `<namespace>` as a binding name rather than the resource name

nora-soderlund:
Fixes #4248

**What this PR solves / how to test:** I came across this today which felt very odd. I couldn't name my namespace `myproduct-prod` because `wrangler kv:namespace create <namespace>` was expecting `<namespace>` to be a binding name.

image

Yet it created the namespace with a `worker-` prefix when I changed my input to `myproduct_prod`. 

image

Today, it doesn't do anything with the binding name, although in the code it's indicated that it will automatically add itself to wrangler.toml in the future. I believe the better approach for differencing between the namespace name and binding name is to opt for another input for the binding name (or as suggested in the code, create a stripped version of the input).

In the meantime, here's a PR which removes the _binding_ name check and replaces it instead with nothing, since it seems as if anything is acceptable as a namespace name in the dashboard.

image

**Author has addressed the following:** This PR is a breakout from #5007 to leave that PR open for an unrelated breaking change.

- Tests
  - [X] Included
  - [ ] Not necessary because:
  - [X] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s): 
  - [x] Not necessary because: the docs do not mention the overly zealous validation of the title passed to the command.


.changeset/lovely-dolls-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
petebacondarwin, line 2: Not really a new feature, it's a bug fix, so I think we can go with a patch bump here.

.changeset/lovely-dolls-taste.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+fix: kv:namespace create to refer to the namespace name instead of the binding name
```
petebacondarwin, line 5: Let's expand this description a bit. Without any context I think the current text is a bit cryptic. Perhaps something like:

```
fix: allow `kv:namespace create` to accept a namespace name that contains characters not allowed in a binding name

This command tries to use the namespace name as the binding. Previously, we would unnecessarily error if this namespace name did not fit the binding name constraints. Now we accept such names and then remove invalid characters when generating the binding name.
```

petebacondarwin (approved):
Magic! Thanks for adding the extra unit tests for the new function.
Can we consider updating the changeset before landing?
Otherwise LGTM!

.changeset/lovely-dolls-taste.md
nora-soderlund: For sure. Thanks. :)

PR #5052, opened 2024-02-18
[C3] Improve type support in Qwik template

jculvey:

Adds a `build-cf-types` script to the Qwik template and modifies the entrypoint to contain a definition for the `env` object composed of all bound resources defined in `wrangler.toml`. 

Also adds a helper function that simplifies handling snippets to be use with codemods. These snippets can now be kept in their own files and benefit from syntax highlighting instead of being embedded as strings in template code.

There's also a new helper script for simplifying the development of codemods in isolation.

- Tests
  - [ ] Included
  - [ x ] Not necessary because: Already has adequate coverage.
  - [ x ] Included
  - [ ] Not necessary because:
- Associated docs
  - [ x ] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/pulls/jculvey
  - [ ] Not necessary because:

**Note for PR author:**


packages/create-cloudflare/src/helpers/codemod.ts
```diff
@@ -76,3 +77,31 @@ export const transformFile = (
...
+			.filter((fileName) => [".js", ".ts"].includes(extname(fileName)))
+			.reduce((acc, snippetPath) => {
+				const [file, ext] = snippetPath.split(".");
+				const key = `${file}${ext === "js" ? "Js" : "Ts"}`;
+				return {
+					...acc,
+					[key]: parseFile(join(snippetsPath, snippetPath))?.body,
+				};
+			}, {}) as Record<string, recast.types.ASTNode[]>
+	);
```
petebacondarwin, line 110: NIT - personally I find use of filters and reduce actually makes the code less readable and in fact less performant due to the recreation of arrays and objects on each iteration.

How about:

```ts
	const result: Record<string, Program["body"]|undefined> = {}
	for(const file of files) {
		if (!lstatSync(join(snippetsPath, file)).isFile) {
		  continue;
		}
		const ext = extname(file);
		if (!['.js', '.ts'].includes(ext)) {
			continue;
		}
		const baseName = basename(ext);
		const key = `${baseName}${ext === "js" ? "Js" : "Ts"}`;
		result[key] = parseFile(join(snippetsPath, file))?.body;
	}

	return result;
```

packages/create-cloudflare/.gitignore
```diff
@@ -12,3 +12,6 @@ templates/**/pnpm-lock.yaml
 
 # the build step renames .gitignore files to __dot__gitignore
 templates/**/__dot__gitignore
+
+scripts/snippets/**
+!scripts/snippets/.gitkeep
```
dario-piotrowicz, line 17:
```suggestion
scripts/snippets/**
!scripts/snippets/.gitkeep

```

dario-piotrowicz: PS: I find it personally not necessary to ignore everything in the `snippets` directory but include the directory itself, for me just ignoring `scripts/snippets` should do (If I understood correctly and this is only used for local C3 development)

packages/create-cloudflare/templates/qwik/c3.ts
```diff
@@ -35,43 +36,67 @@ const addBindingsProxy = (ctx: C3Context) => {
...
+			return this.traverse(n);
+		},
+		// Pass the `platform` object from the declaration to the `qwikCity` plugin
+		visitCallExpression: function (n) {
+			const callee = n.node.callee as recast.types.namedTypes.Identifier;
+			if (callee.name !== "qwikCity") {
+				return this.traverse(n);
+			}
+
+			n.node.arguments = [parseTs("{ platform }")];
```
petebacondarwin, line 59: Do we need to be concerned that this plugin can take other properties on its options parameter? See https://github.com/BuilderIO/qwik/blob/main/packages/qwik-city/buildtime/vite/types.ts#L23

Here we are effectively blowing away any other properties.

packages/create-cloudflare/templates/qwik/c3.ts
```diff
@@ -35,43 +36,67 @@ const addBindingsProxy = (ctx: C3Context) => {
...
+			const id = n.node.id as recast.types.namedTypes.Identifier;
+			if (id.name !== "QwikCityPlatform") {
+				this.traverse(n);
 			}
 
-			this.traverse(n);
+			const newBody = [
+				["env", "Env"],
+				// Qwik doesn't supply `cf` to the platform object. Should they do so, uncomment this
+				// ["cf", "CfProperties"],
```
petebacondarwin, line 85: Perhaps add it as an optional property right now?

packages/create-cloudflare/scripts/codemod.ts
dario-piotrowicz, line 1: nit: maybe I'd use a more descriptive name for this file, as `scripts/codemod.ts` gives me the impression that this can be a utility/script that can generally help with code modding

maybe something like `dev-codemod.ts` could be better?

packages/create-cloudflare/templates/qwik/templates/worker-configuration.d.ts
dario-piotrowicz, line 1: is this file needed? are there type errors if the file is not present?

packages/create-cloudflare/templates/nuxt/c3.ts
```diff
@@ -93,6 +93,7 @@ const config: TemplateConfig = {
 		scripts: {
 			deploy: `${npm} run build && wrangler pages deploy ./dist`,
 			preview: `${npm} run build && wrangler pages dev ./dist`,
+			"build-cf-types": `wrangler types`,
```
petebacondarwin, line 96: Feels a bit incongruous to this PR?

petebacondarwin (approved):
A few nits and unblocking comments.

packages/create-cloudflare/src/helpers/codemod.ts
```diff
@@ -55,7 +56,7 @@ export const parseFile = (filePath: string) => {
 		const fileContents = readFile(path.resolve(filePath));
 
 		if (fileContents) {
-			return recast.parse(fileContents, { parser }) as Program;
+			return recast.parse(fileContents, { parser }).program as Program;
```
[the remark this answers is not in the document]
jculvey: Nope, not really a bug fix just a dx improvement, that we returned was technically incorrect.

The only consumer of `parseFile` was  previously `transformFile`, and because of the way it interacts with `recast`, passing it an ast with the entire program vs just the program body are functionally equivalent. Since we didn't ever consume the result of `parseFile` directly, the annoying dx of it's former implementation or the type error was never apparent.

The introduction of `loadSnippets` made the issue apparent since it works more directly with the AST.

As for test coverage: E2E already gives us a pretty good signal here since a codemod failure would likely result in a un-deployable project. That being said, there's probably a case for more direct unit test coverage now that we're using it more broadly.

packages/create-cloudflare/scripts/codemod.ts
jculvey: I wonder if they'd still be thrown off after seeing the gigantic preamble comment 😝 :
image

Fair point though, and I think another pain point of the current name is that I keep pulling up the wrong one when I ctrl + P.

packages/create-cloudflare/templates/qwik/templates/worker-configuration.d.ts
jculvey: Yeah unfortunately it is. 

Once we've modded `src/entry.cloudflare-pages.tsx` to expect an `env` type on the platform object, it expects it to be defined. This isn't an ambient declaration like all frameworks use for this purpose (which default `env` to any), so it will fail to build without an explicit definition.

packages/create-cloudflare/templates/qwik/c3.ts
jculvey: I'm not sure how much value that would really add, since `cf` would exist in dev but not production 🤷 .

petebacondarwin: True but if they add it then it will be available in production right? And then we would immediately support it without having to change the template.

packages/create-cloudflare/scripts/codemod.ts
dario-piotrowicz: > I wonder if they'd still be thrown off after seeing the gigantic preamble comment 😝 

fair 😅 😆

dario-piotrowicz (approved):

packages/create-cloudflare/templates/qwik/templates/worker-configuration.d.ts
dario-piotrowicz: I've got an idea around this (basically add an extra auxiliary d.ts file that guarantees the existence of the interface)

If you're ok with it I'll try my hands on it later on 🙂 (I am not sure if it'd look too too awkward but we can discuss that then 😅)

Also anyways I'd first wait for `wrangler types` customizablility (https://github.com/cloudflare/workers-sdk/pull/5042) + the related C3 changes 🙂

PR #5053, opened 2024-02-18
Update Rust example

aquarhead:
Mostly added a minimum gitignore, and updated various versions. Can confirm it runs fine on my local.

Generate a new project and run `npx wrangler dev`

- Tests
  - [ ] Included
  - [x] Not necessary because: not sure if this even has test? this is experimental template
  - [ ] Included
  - [x] Not necessary because: not sure if this needs to be recorded as change? also I'm very unfamiliar with the tool used but please let me know if I need to do it
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: same as above

### DIFF
--- a/templates/experimental/worker-rust/.gitignore
+++ b/templates/experimental/worker-rust/.gitignore
@@ -1,0 +1,2 @@
+/target
+/node_modules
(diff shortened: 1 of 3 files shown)


penalosa (approved):
For posterity, I ran this by some internal users of Rust on Workers

PR #5054, opened 2024-02-19
Update VSCode to 1.85.2, including updates to TypeScript

penalosa:

This updates the set of VSCode patches for the quick editor to be based of 1.85.2, rather than 1.76.0.

This can be tested by modifying the Quick Edit URL in a local copy of the Workers Playground or the dashboard to point to the prerelease build generated by this PR (https://cdf6467f.quick-edit-cny.pages.dev/). Particular attention should be paid to TypeScript (through JSDoc) support.

- Tests
  - [ ] Included
  - [x] Not included because: Must be manually tested
  - [ ] Included
  - [x] Not necessary because: Not relevant
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: Not relevant


packages/quick-edit/package.json
```diff
@@ -25,6 +25,6 @@
 		"wrangler": "workspace:*"
 	},
 	"engines": {
-		"node": "<17"
+		"node": ">=18.15"
```
CarmenPopoviciu, line 28: not `18.18`? What does this version number depend on?

penalosa: Thanks for catching this! This should probably be 18.18 too

packages/quick-edit/functions/_middleware.ts
```diff
@@ -5,6 +5,8 @@ export const onRequest = async ({
 }: Parameters<PagesFunction>[0]) => {
 	const url = new URL(request.url);
 
+	const isLocalDev = url.hostname === "localhost";
```
petebacondarwin, line 8: Is this a reliable enough way to compute local dev mode?
What if it was 127.0.0.1 or similar?
Do we care? Perhaps not.

packages/quick-edit/package.json
```diff
@@ -11,7 +11,7 @@
 	"author": "workers-devprod@cloudflare.com",
 	"scripts": {
 		"custom:build": "./build.sh",
-		"dev": "concurrently '../../vendor/vscode/scripts/code-web.sh --extensionDevelopmentPath ../../packages/quick-edit/web/quick-edit-extension' 'npm --prefix web/quick-edit-extension run watch-web' 'yarn --cwd ../../vendor/vscode watch' 'yarn --cwd ../../vendor/vscode watch-web'",
+		"dev": "concurrently 'pnpm exec wrangler pages dev ./web' 'npm --prefix web/quick-edit-extension run watch-web' 'yarn --cwd ../../vendor/vscode watch' 'yarn --cwd ../../vendor/vscode watch-web'",
```
petebacondarwin, line 14: So we have swapped out a built in web server for a local pages server?
Anything more to it than that?

petebacondarwin (approved):
I have no idea how to go about reviewing all those patch changes!
But otherwise this seems like a good change.
How do we test this stuff?

petebacondarwin:
Without a changeset we will no longer deploy this to production, I believe.

packages/quick-edit/functions/_middleware.ts
penalosa: We don't really care—this is just for local development and it'll probably always be `localhost`

packages/quick-edit/package.json
penalosa: Yep, that's all that changed here

penalosa:
> I have no idea how to go about reviewing all those patch changes! But otherwise this seems like a good change. How do we test this stuff?

The best way to test this is to run a local copy of the workers playground, and modify the editor URL to point to the prerelease from this branch, and then play around and try and break it

lrapoport-cf:
@petebacondarwin will you be able to test this per @penalosa 's instructions?

petebacondarwin:
I tried this out locally. It seems to be working although I couldn't identify any of the new VS code features in our usage. For example, interactive inlay hints or new refactorings. But these might just not be available in our current setup.

petebacondarwin:
Needs a rebase then it is good to land.

PR #5059, opened 2024-02-20
perf(worker-rust): Add `dev` env for faster local build times

futile:

Add a `dev`-environment to `wrangler.toml` which builds with `worker-build --dev` instead of `worker-build --release`.
This makes build times for local development much faster.

Add `opt-level = "s"` to `Cargo.toml` so release builds are optimized for size.

Remove the deprecated `--local` parameter from the `wrangler` invocation.

Not sure which of these I should do (and how).

- Tests
  - [ ] Included
  - [x] Not necessary because: basic configuration change in a template
  - [ ] Included
  - [x] Not necessary because: we don't have changesets for templates??
- Associated docs
  - [ ] Issue(s)/PR(s): **do we need to update some docs??**
  - [ ] Not necessary because:

### DIFF
--- a/templates/experimental/worker-rust/Cargo.toml
+++ b/templates/experimental/worker-rust/Cargo.toml
@@ -14,6 +14,7 @@ crate-type = ["cdylib"]
 worker = "0.0.17"
 
 [profile.release]
+opt-level = "s" # optimize for size in release builds
 lto = true
 strip = true
 codegen-units = 1
--- a/templates/experimental/worker-rust/package.json
+++ b/templates/experimental/worker-rust/package.json
@@ -4,7 +4,7 @@
 	"private": true,
 	"scripts": {
 		"deploy": "wrangler deploy",
-		"dev": "wrangler dev"
+		"dev": "wrangler dev --env dev"
 	},
 	"devDependencies": {
 		"wrangler": "^3.1.2"
--- a/templates/experimental/worker-rust/package.json
+++ b/templates/experimental/worker-rust/package.json
@@ -4,7 +4,7 @@
 	"private": true,
 	"scripts": {
 		"deploy": "wrangler deploy",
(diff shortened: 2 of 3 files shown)


petebacondarwin (approved):
LGTM

petebacondarwin (approved):
LGTM

futile:
Thanks for the review! One thing I'm not sure about: Removing the `--local` flag. This might fail with older wrangler-versions now, I think? Should I remove it again from this PR?

But then again, this template will probably be used with more recent versions of wrangler (since a template is usually used for _new_ projects), so it might just be fine in this case.

Skye-31:
Wrangler 3, when the change to default to local mode occurred, has been out for ~9 months, so personally I wouldn't be too concerned about the change

futile:
Awesome, thanks! :)

futile:
One other point I realized: Is `opt-level = "s"` for release-builds desired by default? Since it won't be running in the frontend/not downloaded by browsers, maybe this doesn't actually make that much sense?

I added it originally because it's a very common optimization for wasm use-cases, but maybe that doesn't really count for workers.

petebacondarwin:
I'm no Rust specialist so I would bow to your (or anyone else's views) but I would note that Workers still have a size limit on what can be deployed. So it might be worth keeping a size optimisation turned on?

futile:
> I'm no Rust specialist so I would bow to your (or anyone else's views) but I would note that Workers still have a size limit on what can be deployed. So it might be worth keeping a size optimisation turned on?

That's a good point. I think optimizing for size by default should be sensible for wasm-builds (like in this case), and users can still change/adapt this later on anyways.

So if it's up to me I would keep the `opt-level = "s"`, but feel free to let me know if I should change it :)

Skye-31:
That seems reasonable to keep IMO

PR #5060, opened 2024-02-20
[vitest-pool-workers] Bump to `vitest@1.3.0` and add support for `wrangler.{toml,json}` configs

mrbbot:

This PR includes a few assorted improvements to the Workers Vitest pool:

- Bumps the pinned supported version to `1.3.0`
- Ensures we always treat `@cloudflare/vitest-pool-workers` as an external module during development (see comments added by commit for more details)
- Adds support for loading `main`, compatibility settings, and bindings from Wrangler configuration files
- Removes the `getScheduledResult()` function, as it would have no use

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [x] Not necessary because: still not ready to release the Vitest pool, and the APIs added to `miniflare`/`wrangler` are intended to be internal for the time being.
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: the APIs added to `miniflare`/`wrangler` are intended to be internal for the time being.

**Note for PR author:**

### DIFF
--- a/.github/prereleases/1-versions.mjs
+++ b/.github/prereleases/1-versions.mjs
@@ -1,9 +1,5 @@
 import { execSync } from "node:child_process";
-import {
-	getPackagesForPrerelease,
-	getPrereleaseArtifactUrl,
-	setPackage,
-} from "./0-packages.mjs";
+import { getPackagesForPrerelease, setPackage } from "./0-packages.mjs";
 
 function getPrereleaseVersion() {
 	const sha = execSync("git rev-parse --short HEAD", { encoding: "utf8" });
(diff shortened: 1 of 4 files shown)


packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -479,8 +479,17 @@ async function runTests(
...
+	let workerPath = path.join(ctx.distPath, "worker.js");
+	let threadsWorkerPath = path.join(ctx.distPath, "workers", "threads.js");
+	if (process.platform === "win32") {
+		workerPath = `/${ensurePosixLikePath(workerPath)}`;
+		threadsWorkerPath = `/${ensurePosixLikePath(threadsWorkerPath)}`;
+	}
+
 	ctx.state.clearFiles(project.project, files);
 	const data: WorkerContext = {
+		pool: "threads",
```
penalosa, line 548: What is this for?

penalosa (approved):
Looks good!

packages/vitest-pool-workers/src/pool/index.ts
mrbbot: Not too sure, but we're using the `threads` worker script, and we're essentially pretending to be the `threads` pool, so setting this to the same value the `threads` pool does seemed appropriate. I can't seem to find any uses for this internally though, so it's possible it's unused at the moment?

penalosa: That sounds reasonable!

penalosa:
Unfortunately the CI failures seem genuine 😢

PR #5070, opened 2024-02-21
[vitest-pool-workers] Add suite of examples for testing and other assorted fixups

### LINKED ISSUE #5127: 🐛 BUG: vitest-pool-workers node compatibility with the jose and prisma libraries

Wrangler core

3.30.1

v21.6.1

Windows 11

### Observed behavior

When I run the tests in [my repro](https://github.com/vladinator1000/miniflare-node-compat-bug) 
```
npm run test
```

I observe this error

```
 src/jwt.test.ts [ src/jwt.test.ts ]
TypeError: deprecate is not a function
 ❯ node_modules/jose/dist/node/esm/runtime/rsaes.js:23:38
```

The [jose library](https://github.com/panva/jose) actually works in the production runtime of Cloudflare workers, without the `nodejs_compat` flag, but `vitest-pool-workers` paradoxically requires the flag.

### Expected behavior
The tests work.

### Steps to reproduce

Clone https://github.com/vladinator1000/miniflare-node-compat-bug

Run
```
npm i
```
```
npm run test
```

https://github.com/vladinator1000/miniflare-node-compat-bug

```
 src/jwt.test.ts [ src/jwt.test.ts ]
TypeError: deprecate is not a function
 ❯ node_modules/jose/dist/node/esm/runtime/rsaes.js:23:38
 ```
 
I also found a similar issue when using Prisma: https://github.com/prisma/prisma/issues/23193#issuecomment-1971658732

### PR DESCRIPTION
**Pull request by mrbbot**

Fixes #5127 and closes DEVX-1188.

This issue includes a few changes to the `@cloudflare/vitest-pool-workers` package. It also includes a suite of examples we can use as "recipes" for testing common Workers building blocks.

- Allow cross request context `console.log()`s/`import()`s
- Remove internal bindings from user `export default {}` handlers and Durable Objects
- Use Vite's module resolution algorithm to support things like the `package.json` `browser` field, and generally use the same resolution as user source code/test files for the fallback service. Note this would require users to add some `resolve`/`ssr` config to their `vitest.config.ts` files so we...
- Replaced `defineWorkersPoolOptions()` with `defineWorkers{Config,Project{()` functions that replace the usual `define{Config,Project}()` functions. These functions set all the required `resolve`/`ssr` config, and also default the `pool` to `@cloudflare/vitest-pool-workers` so this no longer needs to be specified. ⚠️  **This replacement is a breaking change for anyone who's tried out the pool so far.**
- Wait for storage reset before disposing and all `ctx.waitUntil()`s after each test. Not doing this caused issues with isolated storage.
- Make sure isolated storage errors are understandable by users, and provide suggested fixes.
- Chunk WebSocket RPC messages so they don't exceed `workerd`'s 1MiB limit
- Support loading `define`s from Wrangler configuration files
- Add support for module rules (e.g. WebAssembly imports)
- A few other minor fixups

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: not ready to release this package yet
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/13092
  - [ ] Not necessary because:

### CODE REVIEW

File: fixtures/vitest-pool-workers-examples/tsc-all.mjs [hunk 1]
```diff
@@ -0,0 +1 @@
+// TODO(now): implement
```
KaiSpencer, line 1: Vitest has some nice built in helpers and assertions for testing types, allowing you to scope errors and not have to test using @ts-expect-error in typescript files. Not sure what you had planned for this, but i've had some good success with the built ins from Vitest

mrbbot: Ah, this was just going to be a script for running `tsc` on all `tsconfig.json`s in subdirectories, to check types across all the different TypeScript projects.

KaiSpencer: Gotcha, from vitest 1.0, you can run `vitest --typecheck` which does this under the hood. You can then use `*.test-d.ts` files specifically to write type tests with assertions. 

https://vitest.dev/guide/testing-types
https://vitest.dev/guide/testing-types#run-typechecking

fixtures/vitest-pool-workers-examples/d1/vitest.config.ts
```diff
@@ -0,0 +1,44 @@
+import fs from "node:fs";
+import path from "node:path";
+import { defineWorkersProject } from "@cloudflare/vitest-pool-workers/config";
+import { splitSqlQuery } from "wrangler";
```
mrbbot, line 4: I'm not too sure about exposing this. It's a bit of a pain to implement manually. Maybe it would be better to export a function like `readMigrationQueries()` below?

admah: If it would make it simpler for users to setup, then I say yes. Make this a reusable utility function.

admah:
@mrbbot do you also want to add isolatedStorage default to true to this?

packages/vitest-pool-workers/src/worker/durable-objects.ts
```diff
@@ -88,23 +88,10 @@ function assertSameIsolate(stub: DurableObjectStub) {
...
+async function runInStub<O extends DurableObject, R>(
+	stub: Fetcher,
 	callback: (instance: O, state: DurableObjectState) => R | Promise<R>
 ): Promise<R> {
-	if (!isDurableObjectStub(stub)) {
-		throw new TypeError(
-			"Failed to execute 'runInDurableObject': parameter 1 is not of type 'DurableObjectStub'."
-		);
-	}
-	if (typeof callback !== "function") {
```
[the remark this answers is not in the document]
mrbbot: This function is no longer exported, so it can't be called directly by JavaScript users, meaning we can rely on TypeScript type checking for validating types. This validation is still in the `runInDurableObject()` function below. 👍

packages/vitest-pool-workers/package.json
```diff
@@ -42,7 +42,6 @@
 		"cjs-module-lexer": "^1.2.3",
 		"devalue": "^4.3.0",
 		"esbuild": "0.17.19",
-		"import-meta-resolve": "^4.0.0",
```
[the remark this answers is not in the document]
mrbbot: It did initially, but to aid with rebasing, I've squashed all lockfile changes into one of the later commits. 👍

packages/vitest-pool-workers/src/pool/module-fallback.ts
```diff
@@ -208,30 +211,41 @@ function getApproximateSpecifier(target: string, referrerDir: string): string {
...
-		requires.set(normalisedReferrer, require);
+async function viteResolve(
+	vite: ViteDevServer,
+	specifier: string,
+	referrer: string,
+	isRequire: boolean
+): Promise<string> {
+	const resolved = await vite.pluginContainer.resolveId(specifier, referrer, {
+		ssr: true,
+		// https://github.com/vitejs/vite/blob/v5.1.4/packages/vite/src/node/plugins/resolve.ts#L178-L179
```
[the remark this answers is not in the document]
mrbbot: It is, _but_ this is far from the only private API we're relying on, 😃 which is why we pin the `vitest` version. If this breaks in the future, our tests should catch this before we publish a new version.

packages/vitest-pool-workers/src/worker/events.ts
```diff
@@ -9,6 +9,27 @@ const kConstructFlag = Symbol("kConstructFlag");
...
+// If isolated storage is enabled, we ensure all `waitUntil()`s are `await`ed at
+// the end of each test, as these may contain storage calls (e.g. caching
+// responses). Note we can't wait at the end of `.concurrent` tests, as we can't
+// track which `waitUntil()`s belong to which tests.
+//
+// If isolated storage is disabled, we ensure all `waitUntil()`s are `await`ed
+// at the end of each test *file*. This ensures we don't try to dispose the
+// runtime until all `waitUntil()`s complete.
+const globalWaitUntil: unknown[] = [];
+export function maybeRegisterGlobalWaitUntil(promise: unknown) {
```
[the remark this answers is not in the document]
mrbbot: Ah good catch! In an earlier version, `globalWaitUntil` could've be `undefined`, but this was changed. Will update this. 👍

packages/vitest-pool-workers/src/shared/chunking-socket.ts
```diff
@@ -0,0 +1,58 @@
...
+	on(listener: (message: Message) => void): void;
+}
+
+/**
+ * Wraps a binary/string socket to produce a chunked-string socket. If a string
+ * exceeds the maximum size for a message, it is split into binary chunks
+ * followed by an empty string. If not, it is sent directly as a string.
+ * This assumes in-order delivery of chunks.
+ */
+export function createChunkingSocket(
```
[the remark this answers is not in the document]
mrbbot: Haha, where?

packages/vitest-pool-workers/src/worker/index.ts
```diff
@@ -93,6 +93,8 @@ class WebSocketMessagePort extends events.EventEmitter {
 	}
 
 	postMessage(data: unknown) {
+		if (this.socket.readyState !== WebSocket.READY_STATE_OPEN) return;
```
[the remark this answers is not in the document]
mrbbot: The WebSocket is in an open state when we construct this class, and we never re-open the socket if it gets closed. So buffering isn't really needed here. Will add a comment to clarify this though. 👍

packages/vitest-pool-workers/test/stacked/vitest.config.ts
```diff
@@ -1,18 +1,19 @@
-import { defineWorkersPoolOptions } from "@cloudflare/vitest-pool-workers/config";
-import { defineConfig } from "vitest/config";
+import { defineWorkersProject } from "@cloudflare/vitest-pool-workers/config";
 
-export default defineConfig({
+export default defineWorkersProject({
```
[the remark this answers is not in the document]
mrbbot: Correct, `defineWorkersConfig()` should be used in non-workspace scenarios.

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -373,7 +383,19 @@ function buildProjectWorkerOptions(
...
+	// define script similar to Vite's. When defines change, Miniflare will be
+	// restarted as the input options will be different.
+	const defines = `export default {
+		${Object.entries(project.options.defines ?? {})
+			.map(([key, value]) => `${JSON.stringify(key)}: ${value}`)
+			.join(",\n")}
+	};
+	`;
+
+	// Make sure we define the runner script, including object wrappers & defines
```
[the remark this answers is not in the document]
mrbbot: Yeah, could move it further down maybe?

packages/miniflare/src/plugins/core/modules.ts
```diff
@@ -109,15 +109,15 @@ const DEFAULT_MODULE_RULES: ModuleRule[] = [
 	{ type: "CommonJS", include: ["**/*.js", "**/*.cjs"] },
 ];
 
-interface CompiledModuleRule {
+export interface CompiledModuleRule {
```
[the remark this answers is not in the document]
mrbbot: Because of the newly exported `miniflare` APIs? Miniflare already exports quite a few internal APIs that aren't documented, so I'm not too worried about adding some more. 😅 Essentially, Miniflare's public API surface is whatever's included in the `README`.

fixtures/vitest-pool-workers-examples/d1/test/apply-migrations.ts
```diff
@@ -0,0 +1,6 @@
+import { applyD1Migrations, env } from "cloudflare:test";
+
+// Setup files run outside isolated storage, and may be run multiple times.
+// `applyD1Migrations()` only applies migrations that haven't already been
+// applied, therefore it is safe to call this function here.
+await applyD1Migrations(env.DATABASE, env.TEST_MIGRATIONS);
```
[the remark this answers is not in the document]
mrbbot: You could definitely do it in a `beforeEach()`/`beforeAll()` block. I guess if you had multiple test files, you'd probably want to apply migrations in each file, so including it in a setup file that gets run before each file makes sense.

packages/vitest-pool-workers/src/worker/d1.ts
```diff
@@ -0,0 +1,81 @@
...
+		typeof v === "object" &&
+		v !== null &&
+		"name" in v &&
+		typeof v.name === "string" &&
+		"queries" in v &&
+		Array.isArray(v.queries) &&
+		v.queries.every((query) => typeof query === "string")
+	);
+}
+function isD1Migrations(v: unknown): v is D1Migration[] {
```
petebacondarwin, line 28: `areD1Migrations()`? 😃 (joke)

.prettierignore
```diff
@@ -47,3 +47,8 @@ packages/miniflare
 # so let's ignore all the c3 template files, exlcuding the c3.ts ones
 packages/create-cloudflare/templates/**/*.*
 !packages/create-cloudflare/templates/**/c3.ts
+
+# Some of the `@cloudflare/vitest-pool-workers` examples use `satisfies` as it
```
petebacondarwin, line 51: I guess that as long as they are only in tests this is OK.
But I don't think our version of esbuild supports it, right?
So we need to be careful we don't encourage people to expect that it can be used in the Workers source.

fixtures/vitest-pool-workers-examples/d1/vitest.config.ts
```diff
@@ -0,0 +1,30 @@
...
+	// Read all migrations in the `migrations` directory
+	const migrationsPath = path.join(__dirname, "migrations");
+	const migrations = await readD1Migrations(migrationsPath);
+
+	return {
+		test: {
+			setupFiles: ["./test/apply-migrations.ts"],
+			poolOptions: {
+				workers: {
+					singleWorker: true,
```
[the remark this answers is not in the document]
mrbbot: Yes, only one Worker is instantiated, then all tests run in series in that worker. Isolated storage is enabled-by-default, so yep, writes performed in an `it` are undone after that `it`.

See https://admah-vitest-pool-docs.cloudflare-docs-7ou.pages.dev/workers/testing/vitest-integration/internal-details/ _(modulo the `(Default)` which is now `isolatedStorage: true, singleWorker: false`)_.

packages/vitest-pool-workers/src/worker/d1.ts
mrbbot: 🙃

fixtures/vitest-pool-workers-examples/basics-integration-auxiliary/vitest.config.ts
```diff
@@ -0,0 +1,39 @@
...
+						WORKER: "worker",
+					},
+
+					workers: [
+						// Configuration for the "auxiliary" Worker under test.
+						// Unfortunately, auxiliary Workers cannot load their configuration
+						// from `wrangler.toml` files, and must be configured with Miniflare
+						// `WorkerOptions`.
+						{
+							name: "worker",
```
petebacondarwin, line 28: Might be helpful to give this a slightly less generic name so that it is obvious that it is being used in the service binding.

fixtures/vitest-pool-workers-examples/vitest.workers.config.ts
petebacondarwin, line 1: Is this file the sort of thing that developers would need to write in their own projects?
It feels a bit tricky for most people to work out.

mrbbot: No, this is just for us. As long of `@cloudflare/vitest-pool-workers` is being loaded from `node_modules`, you should be good.

fixtures/vitest-pool-workers-examples/hyperdrive/src/env.d.ts
```diff
@@ -0,0 +1,3 @@
+interface Env {
+	ECHO_SERVER: Hyperdrive;
```
petebacondarwin, line 2: I would be tempted to call this something like `FAKE_HYPERDRIVE`.

petebacondarwin (approved):
Awesome work! Love it.

fixtures/vitest-pool-workers-examples/hyperdrive/src/env.d.ts
mrbbot: Renamed to `ECHO_SERVER_HYPERDRIVE` 👍

fixtures/vitest-pool-workers-examples/basics-integration-auxiliary/vitest.config.ts
mrbbot: Renamed to `worker-under-test` 👍

rubenbernar (commented):
Robo

PR #5083, opened 2024-02-24
WAFFLE-538 Add Miniflare support/tests for Ratelimits emulation

oliy:
Fixes # WAFFLE-538.

Add emulated Ratelimit bindings to Miniflare to allow internal development.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): WAFFLE-538 (not for public consumption yet)
  - [ ] Not necessary because:

**Note for PR author:**

### DIFF
--- a/packages/miniflare/src/plugins/index.ts
+++ b/packages/miniflare/src/plugins/index.ts
@@ -8,6 +8,7 @@ import { HYPERDRIVE_PLUGIN, HYPERDRIVE_PLUGIN_NAME } from "./hyperdrive";
 import { KV_PLUGIN, KV_PLUGIN_NAME } from "./kv";
 import { QUEUES_PLUGIN, QUEUES_PLUGIN_NAME } from "./queues";
 import { R2_PLUGIN, R2_PLUGIN_NAME } from "./r2";
+import { RATELIMIT_PLUGIN, RATELIMIT_PLUGIN_NAME } from "./ratelimit";
 
 export const PLUGINS = {
 	[CORE_PLUGIN_NAME]: CORE_PLUGIN,
@@ -18,6 +19,7 @@ export const PLUGINS = {
 	[QUEUES_PLUGIN_NAME]: QUEUES_PLUGIN,
 	[R2_PLUGIN_NAME]: R2_PLUGIN,
 	[HYPERDRIVE_PLUGIN_NAME]: HYPERDRIVE_PLUGIN,
+	[RATELIMIT_PLUGIN_NAME]: RATELIMIT_PLUGIN,
 };
 export type Plugins = typeof PLUGINS;
 
@@ -63,7 +65,8 @@ export type WorkerOptions = z.input<typeof CORE_PLUGIN.options> &
 	z.input<typeof KV_PLUGIN.options> &
 	z.input<typeof QUEUES_PLUGIN.options> &
 	z.input<typeof R2_PLUGIN.options> &
-	z.input<typeof HYPERDRIVE_PLUGIN.options>;
+	z.input<typeof HYPERDRIVE_PLUGIN.options> &
+	z.input<typeof RATELIMIT_PLUGIN.options>;
 export type SharedOptions = z.input<typeof CORE_PLUGIN.sharedOptions> &
 	z.input<typeof CACHE_PLUGIN.sharedOptions> &
 	z.input<typeof D1_PLUGIN.sharedOptions> &
@@ -110,3 +113,4 @@ export * from "./kv";
 export * from "./queues";
 export * from "./r2";
 export * from "./hyperdrive";
+export * from "./ratelimit";
--- a/packages/miniflare/src/plugins/ratelimit/index.ts
+++ b/packages/miniflare/src/plugins/ratelimit/index.ts
@@ -1,0 +1,82 @@
+import SCRIPT_RATELIMIT_OBJECT from "worker:ratelimit/ratelimit";
+import { z } from "zod";
+import {
+	Worker_Binding,
+} from "../../runtime";
+import {
+	Plugin,
+	kProxyNodeBinding,
+} from "../shared";
+
+export enum PeriodType {
+	TENSECONDS = 10,
+	MINUTE = 60
+}
+
+export const RatelimitConfigSchema = z.object({
+	simple: z.object({
+		limit: z.number().gt(0),
+		
+		// may relax this to be any number in the future
+		period: z.nativeEnum(PeriodType).optional(),
+	})
+});
+export const RatelimitOptionsSchema = z.object({
+	ratelimits: z.record(RatelimitConfigSchema).optional(),
+});
+
+export const RATELIMIT_PLUGIN_NAME = "ratelimit";
+const SERVICE_RATELIMIT_PREFIX = `${RATELIMIT_PLUGIN_NAME}`;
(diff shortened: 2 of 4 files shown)


packages/miniflare/src/plugins/index.ts
```diff
@@ -69,7 +71,8 @@ export type SharedOptions = z.input<typeof CORE_PLUGIN.sharedOptions> &
 	z.input<typeof D1_PLUGIN.sharedOptions> &
 	z.input<typeof DURABLE_OBJECTS_PLUGIN.sharedOptions> &
 	z.input<typeof KV_PLUGIN.sharedOptions> &
-	z.input<typeof R2_PLUGIN.sharedOptions>;
+	z.input<typeof R2_PLUGIN.sharedOptions> &
+	z.input<typeof RATELIMIT_PLUGIN.options>;
```
mrbbot, line 75: This should be added to `WorkerOptions` instead of `SharedOptions`.

packages/miniflare/src/plugins/ratelimit/index.ts
```diff
@@ -0,0 +1,85 @@
...
+function buildJsonBindings(bindings: Record<string, any>): Worker_Binding[] {
+	return Object.entries(bindings).map(([name, value]) => ({
+		name,
+		json: JSON.stringify(value),
+	}));
+}
+
+export const RATELIMIT_PLUGIN: Plugin<typeof RatelimitOptionsSchema> = {
+	options: RatelimitOptionsSchema,
+	getBindings(options: z.infer<typeof RatelimitOptionsSchema>) {
```
mrbbot, line 41: TypeScript should be able to infer this.
```suggestion
	getBindings(options) {
```

packages/miniflare/src/plugins/ratelimit/index.ts
```diff
@@ -0,0 +1,85 @@
...
+		if (!options.ratelimits) {
+			return [];
+		}
+		const bindings = Object.entries(options.ratelimits).map<Worker_Binding>(([name, config]:[string,z.infer<typeof RatelimitConfigSchema>]) => ({
```
mrbbot, line 45: TypeScript should be able to infer some of the types here:
```suggestion
		const bindings = Object.entries(options.ratelimits).map<Worker_Binding>(([name, config]) => ({
```

packages/miniflare/src/plugins/ratelimit/index.ts
```diff
@@ -0,0 +1,85 @@
...
+		}
+		const ratelimits = Object.entries(options.ratelimits);
+		
+		const extensions: Extension[] = ratelimits.map<Extension>(([name, _]) => ({
+			modules: [{
+				name: `${SERVICE_RATELIMIT_PREFIX}:${name}`,
+				esModule: SCRIPT_RATELIMIT_OBJECT(),
+				internal: true,
+			}],
+		}));
```
mrbbot, line 78: It looks like you're using the same script for each ratelimit binding. You should only need to include a single copy of this extension.

packages/miniflare/src/plugins/ratelimit/index.ts
```diff
@@ -0,0 +1,85 @@
...
+		options,
+	}) {
+		if (!options.ratelimits) {
+			return [];
+		}
+		const ratelimits = Object.entries(options.ratelimits);
+		
+		const extensions: Extension[] = ratelimits.map<Extension>(([name, _]) => ({
+			modules: [{
+				name: `${SERVICE_RATELIMIT_PREFIX}:${name}`,
```
mrbbot, line 74: Extension modules will be accessible to any other built-in module. I'd make sure this `name` starts with `cloudflare-internal:`, so we don't accidentally import it. 👍

packages/miniflare/src/workers/ratelimit/ratelimit.worker.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.limitVal = config.limit
+    this.period = config.period
+
+    this.buckets = new Map<string,number>()
+    this.epoch = 0
+  }
+
+  // method that counts and checks against the limit in in-memory buckets
+  async limit({key, limit, period}: RatelimitOptions): Promise<RatelimitResult> {
+    period = period || this.period
```
mrbbot, line 42: As this function is exposed to users, could you perform type validation on the arguments? The `zod` package should be available under the `miniflare:zod` module inside `.worker.ts` files.

```ts
import { z } from "miniflare:zod";

const RatelimitOptionsSchema = z.object({ ... });

async limit(options: unknown) {
  const { key, limit, period } = RatelimitOptionsSchema.parse(options);
}
```

packages/miniflare/test/plugins/ratelimit/index.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+		ratelimits: {
+			TESTRATE: {
+				simple: {
+					limit: 2,
+					period: 60,
+				}
+			}
+		},
+		
+		compatibilityFlags: ["nodejs_compat"],
```
mrbbot, line 20: Is doesn't look like this compatibility flag is required?

packages/miniflare/test/plugins/ratelimit/index.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+		script: `
+		export default {
+			async fetch(request, env, ctx) {
+				const { success } = await env.TESTRATE.limit({
+					key: "test",
+				});
+				if (!success) {
+					return new Response("rate limited", { status: 429 });
+				}
+				return success ? new Response("success", { status: 200 }) : new Response("rate limited", { status: 429 });
```
mrbbot, line 31: `success` is always `true` here, as you've just returned if it was `false`.

mrbbot (commented):
Thanks for submitting this PR! 🙂 I've added some comments.

packages/miniflare/test/plugins/ratelimit/index.spec.ts
oliy: good catch!  fixed.

packages/miniflare/src/plugins/index.ts
oliy: Fixed.

packages/miniflare/src/workers/ratelimit/ratelimit.worker.ts
oliy: having problem now of finding zod:
```
service core:user:: Uncaught Error: No such module "miniflare:zod".
  imported from "cloudflare-internal:ratelimit:module"
```

oliy: ok.  fixed, with reverting back to non-module, but added explicit validations.

mrbbot (approved):
Formatted the code in this PR according to `miniflare`'s rules, and tightened up some of the TypeScript types. Given this isn't for public consumption yet, I think we're ok to merge this without a changeset.

irvinebroque:
@oliy — I took your demo and tried it in local dev:

https://github.com/irvinebroque/rate-limiting-demo

If you clone this and run it in local dev, latest version or Wrangler — binding isn't defined:

```
Cannot read properties of undefined (reading 'limit')
```

Sounds like more changes needed in order to get this working with Wrangler — need to go beyond just Miniflare?

PR #5087, opened 2024-02-24
fix: make `wrangler types` always generate a `d.ts` file for module worders

dario-piotrowicz:

Currently if a config file doesn't define any binding nor module, running
`wrangler types` against such file would not produce a `d.ts` file.

Producing a `d.ts` file can however still be beneficial as it would define a correct
env interface (even if empty) that can be expanded/referenced by user code (this can
be particularly convenient for scaffolding tools that may want to always generate an
env interface).

Example:
Before `wrangler types --env-interface MyEnv` run with an empty `wrangler.toml` file
would not generate any file, after these change it would instead generate a file with
the following content:

```
interface MyEnv {
}
```

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: this is a quality of life improvement that doesn't require documentation

**Note for PR author:**

### DIFF
--- a/packages/wrangler/src/type-generation.ts
+++ b/packages/wrangler/src/type-generation.ts
@@ -306,8 +306,8 @@ function writeDTSFile({
 	let combinedTypeStrings = "";
 	if (formatType === "modules") {
 		combinedTypeStrings += `interface ${envInterface} {${envTypeStructure
-			.map((value, i) => `${i === 0 ? "\n" : ""}\t${value}`)
-			.join("\n")}\n}\n${modulesTypeStructure.join("\n")}`;
+			.map((value) => `\n\t${value}`)
+			.join("")}\n}\n${modulesTypeStructure.join("\n")}`;
 	} else {
 		combinedTypeStrings += `export {};\ndeclare global {\n${envTypeStructure
 			.map((value) => `\tconst ${value}`)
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/type-generation.ts
```diff
@@ -305,8 +305,8 @@ function writeDTSFile({
 
 	let combinedTypeStrings = "";
 	if (formatType === "modules") {
-		combinedTypeStrings += `interface ${envInterface} {\n${envTypeStructure
-			.map((value) => `\t${value}`)
+		combinedTypeStrings += `interface ${envInterface} {${envTypeStructure
+			.map((value, i) => `${i === 0 ? "\n" : ""}\t${value}`)
 			.join("\n")}\n}\n${modulesTypeStructure.join("\n")}`;
```
petebacondarwin, line 310: NIT: I think this could be simpler. You are just trying to ensure that every item has a `\n` at the start, no?

```suggestion
			.map((value) => `\n\t${value}`)
			.join("")}\n}\n${modulesTypeStructure.join("\n")}`;
```

petebacondarwin (approved):

packages/wrangler/src/type-generation.ts
dario-piotrowicz: good point! 😅👍

PR #5089, opened 2024-02-25
Add support for new bindings to wrangler types

DaniFoldi:

This PR adds support for Email Send, Vectorize, Hyperdrive, mTLS, Browser Rendering and Workers AI bindings to `wrangler types`.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because: 

**Note for PR author:**


packages/wrangler/src/type-generation.ts
```diff
@@ -240,6 +246,43 @@ async function generateTypes(
...
+		// The BrowserWorker type in @cloudflare/puppeteer is of type
+		// { fetch: typeof fetch }, but workers-types doesn't include it
+		// and Fetcher is valid for the purposes of handing it to puppeteer
+		envTypeStructure.push(`${configToDTS.browser.binding}: Fetcher;`);
+	}
+
+	if (configToDTS.ai) {
+		// The @cloudflare/ai package types specify any, but from its usage
+		// we can infer it's a Fetcher type (await this.binding.fetch)
+		envTypeStructure.push(`${configToDTS.ai.binding}: Fetcher;`);
```
dario-piotrowicz, line 283: I'd be tempted to keep it as `any` if it is in `@cloudflare/ai` :thinking:

dario-piotrowicz (approved):
LGTM :smiley: :rocket: :heart:

packages/wrangler/src/type-generation.ts
```diff
@@ -95,6 +95,12 @@ export async function typesHandler(
 		rules: config.rules,
 		queues: config.queues,
 		constellation: config.constellation,
+		send_email: config.send_email,
+		vectorize: config.vectorize,
+		hyperdrive: config.hyperdrive,
+		mtls_certificates: config.mtls_certificates,
+		browser: config.browser,
+		ai: config.ai,
```
penalosa, line 103: Can we change the type of this object to make sure we catch issues like this in the future? It should be something like `Omit<EnvironmentNonInheritable, "define" | "unsafe" | "tail_consumers" | "constellation" | "cloudchamber">`

packages/wrangler/src/type-generation.ts
penalosa: I think this should probably be `any`/`unknown` since I'm not sure it's guaranteed to always be a `Fetcher`

DaniFoldi: I'd like to avoid generating `any` into user projects - lots of people use the recommended preset from typescript-eslint which includes [no-explicit-any](https://typescript-eslint.io/rules/no-explicit-any/). `unknown` is somewhat better, I haven't thought of that myself. Let me push a fix

dario-piotrowicz: yes I would also go with `unknown`, I didn't suggest it as I don't know if `@cloudflare/ai` could complain about that?

DaniFoldi: Fortunately _anything_ can be passed to `any`, including unknown: [ts playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygCgIYCcDmAXItmAJ4CUiA3gFCIOK0C+ttAJihADZ4qIQEGRADc8MbACNuKYuADWYOAHcwbVBkxjcE6Sgq0gA)

Changed to unknown in [b69e965](https://github.com/cloudflare/workers-sdk/pull/5089/commits/b69e9650bfee4804e4b82e9674afbda5d0d5d0ca)

mrbbot (approved):

DaniFoldi:
Of course, I'll remove `unsafe` from the omit since the test actually has unsafe bindings defined.
How should I go about fixing the `dispatch_namespaces` error? It's defined on `EnvironmentInheritable`, so I can either move it to non-inheritable (which shouldn't be breaking as only their union is exported, and if anything this is a `fix`, or I can manually patch the type to also include dispatch namespaces from the inheritable config (smaller diff, but worse solution imo).

DaniFoldi:
Rebased on main, test failures seemed unrelated. Could you please take another look, @mrbbot?

PR #5091, opened 2024-02-25
fix(wrangler): prevent invalid JS identifiers in types generation

### LINKED ISSUE #5090: 🐛 BUG: `wrangler types` generates wrong types with dashes in vars, bindings, etc.

Wrangler core

3.29.0

20.11.1

Windows

### Observed behavior
With the following config:
```toml
[vars]
some-var = "foobar"
```

The generated types file includes:
```typescript
interface Env {
	some-var: "foobar";
}
```

This is syntactically invalid.

### Expected behavior

This should wrap all keys in quotes:
```typescript
interface Env {
	"some-var": "foobar";
}
```

_No response_

_No response_

Cherry:
Fixes #5090 

Previously, dashes in bindings or vars would result in an invalid output with `wrangler types`. With the following:
```toml
[vars]
some-var = "foobar"
```
You would get:
```typescript
interface Env {
	some-var: "foobar";
}
```
which is syntactically invalid. This PR fixes that by wrapping necessary keys in quotes. So now you get:

```typescript
interface Env {
	"some-var": "foobar";
}
```

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because:

**Note for PR author:**


Cherry:
Thanks @dario-piotrowicz. I agree it would be good to only quote when needed. I originally opted for consistency, since that's what I personally use in my ESLint config (https://eslint.style/rules/default/quote-props `consistent`), but that's definitely opinionated.

On further thought and discussion though, I think there's significantly more edge-cases here that end up syntactically invalid. While not all valid binding names, these are all valid vars:

```toml
[vars]
dashed-var="foo"
"single-quote'"="bar"
'"double-quote"'="baz"
"escaped\"quote"="qux"
"escaped\\backslash"="quux"
123num="quuz'"
123numquote="'''"
```

Which today results in the following:
```ts
interface Env {
	dashed-var: "foo";
	single-quote': "bar";
	"double-quote": "baz";
	escaped"quote: "qux";
	escaped\backslash: "quux";
	123num: "quuz'";
	123numquote: "'''";
}
```

With my PR here, it's slightly better, but still not great:
```ts
interface Env {
	"dashed-var": "foo";
	"single-quote'": "bar";
	""double-quote"": "baz";
	"escaped"quote": "qux";
	"escaped\backslash": "quux";
	"123num": "quuz'";
	"123numquote": "'''";
}
```
My PR handles the vars starting with numbers, and the dashes, but falls over once you introduce quotes and escaping other chars. This is likely going to need some better thought and handling. I'm happy to take another look later this week, but if you have an idea for how you want to best support this, please feel free.

dario-piotrowicz:
@Cherry Regarding the optional inclusion of quotes, I think that it could be great to add the quotes only when needed and maybe have an option like: `--with-consistency` or something along those lines if you always want quotes...?

Although maybe I would not even add the flag and just have the quotes only added when needed. Users can very easily then just use their formatter to adjust the file after the fact anyways right? they'll get their style as soon as they run prettier or whatever they use anyways no?

Mh... based on my argument above you could actually argue that you could always add quotes for simplicity and developers can remove them with their formatter right? 😅 🤷  So I am not too too sure 😅  (but I am opinionated on having it only when needed 😛) 

___


Regarding edge cases and whatnot, when I saw your PR I did suspect that there would be more cases in which this breaks, but it didn't feel like a too big of a deal to me as I think that as long as we improve things/go in the right direction it makes sense to fix the small bits we find without getting bugged into solving everything...
(but yeah it's probably better to at least start with a nice thought out robust solution instead of throwing things at the wall and see what sticks)

>  I'm happy to take another look later this week, but if you have an idea for how you want to best support this, please feel free.

Please do continue! I personally don't have an idea on how to tackle this on the top of my head 🙂
but anyways if you won't find the time or drop this I am always happy to step in and help 🙂

> The current implementation also seems to assume that variables are always strings, but when you do foo=1 in vars, it's available under env as a number, and not a string.

I think you're right! good catch!
https://github.com/cloudflare/workers-sdk/blob/cab7e1c7f24ff0097e15d90030c805fe2785d173/packages/wrangler/src/type-generation.ts#L152-L158

but that's a separate bug, if you want please open a new issue for that and we can tackle that separately 🙂  (I don't mind doing it if you want)
(it should also be pretty straightforward to handle)

> You can do fun stuff like "breaking /*" = "testing" with the current implementation and break the rest of the generation too (very edge-casey).

😨 😬

DaniFoldi:
I agree with @dario-piotrowicz that preferably wrangler shouldn't start quoting all props, since that would suddenly change all previously generated d.ts files just by updating wrangler and running the command.

If we go further in preferences, should it use single `'` or double `"` quotes for keys? What about string literals?

dario-piotrowicz:
> If we go further in preferences, should it use single `'` or double `"` quotes for keys? What about string literals?

I also do prefer single quotes in js/ts code 😄, but in this case, changing that would again suddenly change all previously generated d.ts files, wouldn't it? so I guess that given that double quotes have been used thus far I'd just stick with those 😕

Cherry:
With the latest commits, this should be good for review. It'll (naively) check if an identifier is valid in JS or not, and only if it's not, will it then quote it.

It doesn't cover _every single edge-case_, especially when it comes to weird UTF8 characters, but it handles what I'd consider the vast vast majority in the real-world. I've added tests for the functions I've added, and abstracted them so they can be improved in future as needed.

packages/wrangler/src/type-generation.ts
```diff
@@ -112,6 +112,54 @@ export async function typesHandler(
 	);
 }
 
+/**
+ * Check if a string is a valid TypeScript identifier. This is a naive check and doesn't cover all cases
+ * If not, it should be wrapped in quotes
+ */
+export function isValidIdentifier(key: string) {
+	return /^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(key);
```
dario-piotrowicz, line 120: `\w` is equivalent to `[a-zA-Z0-9_$]` no?
```suggestion
	return /^[a-zA-Z_$]\w*$/.test(key);
```

Cherry: `\w` doesn't include `$`, but I could do `\w$`, yep!

packages/wrangler/src/type-generation.ts
```diff
@@ -112,6 +112,54 @@ export async function typesHandler(
...
+ * If not, it should be wrapped in quotes
+ */
+export function isValidIdentifier(key: string) {
+	return /^[a-zA-Z_$][a-zA-Z0-9_]*$*$/.test(key);
+}
+
+/**
+ * Escape a string value for use in a string in a TypeScript file
+ */
+export function escapeStringValue(value: string) {
```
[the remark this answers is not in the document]
Cherry: Agreed, I previously using this for something else, but now it's online a single place so will just inline it and update my tests for `constructType`.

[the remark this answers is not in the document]
Cherry: Done!

packages/wrangler/src/type-generation.ts
Cherry: Done!

.changeset/late-eggs-enjoy.md
```diff
@@ -0,0 +1,20 @@
+---
+"wrangler": patch
+---
+
+fix: better handle dashes and other invalid JS identifier characters in `wrangler types` generation for vars, bindings, etc.
+
+Previously, with the follwing in your `wrangler.toml`, an invalid types file would be generated:
```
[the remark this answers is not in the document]
Cherry: Done!

packages/wrangler/src/type-generation.ts
dario-piotrowicz: ah yeah sorry I missed the `$` 😅 (which I included in the comment as I just copy-pasted the thing 🙈)

packages/wrangler/src/type-generation.ts
```diff
@@ -112,6 +112,47 @@ export async function typesHandler(
...
+
+/**
+ *
+ * Construct a type key, if it's not a valid identifier, wrap it in quotes
+ * If rawVal is true, we'll use the value as the type, otherwise we'll format it for a string definition
+ */
+export function constructType(
+	key: string,
+	value: string | number | boolean,
+	rawVal = true
```
dario-piotrowicz, line 141: `rawVal` seems like the function is receiving a raw value to me, I'd go with something like `useRawString`, `useVerbatinString`, etc... instead

packages/wrangler/src/type-generation.ts
```diff
@@ -112,6 +112,47 @@ export async function typesHandler(
...
+	if (typeof value === "string") {
+		if (rawVal) {
+			return `${typeKey}: ${value};`;
+		}
+		return `${typeKey}: "${value.replace(/"/g, '\\"')}";`;
+	}
+	if (typeof value === "number" || typeof value === "boolean") {
+		return `${typeKey}: ${value};`;
+	}
+	return `${typeKey}: unknown;`;
```
dario-piotrowicz, line 153: I _think_ this function could be simplified:
```suggestion
	const typeKey = constructTypeKey(key);

	if(rawVal && typeof value === 'string') {
		return `${typeKey}: ${value};`;
	}

	return `${typeKey}: "${`${value}`.replace(/"/g, '\\"')}";`;
```

dario-piotrowicz (approved):
Looks good to me 🙂 

Except a few nits that I hope get addressed

Also I personally dislike tsdoc comments in imperative form (e.g. `check ...`) and prefer them in third person (e.g. `checks ...`), I'd push back on them but I've checked and in the whole codebase it seems we have a mix of both (and imperative even looks like the more common 🥲), so I'm fine with them.

packages/wrangler/src/type-generation.ts
Cherry: If I'm understanding correctly, that's very different logically, and doesn't handle numbers/booleans like the previous implementation did.

`constructType('valid', 2)`:

Previous: `'valid: 2;'`
New: `'valid: "2";'`

packages/wrangler/src/type-generation.ts
```diff
@@ -112,6 +112,47 @@ export async function typesHandler(
...
+export function constructTypeKey(key: string) {
+	if (isValidIdentifier(key)) {
+		return `${key}`;
+	}
+	return `"${key}"`;
+}
+
+/**
+ *
+ * Construct a type key, if it's not a valid identifier, wrap it in quotes
```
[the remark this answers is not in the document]
Cherry: Good catch, fixed.

DaniFoldi:
I found an edge case which breaks the literal type escaping introduced in this PR:
```toml
[vars]
VARIABLE = 'some-"string\\'
```

Cherry:
> I found an edge case which breaks the literal type escaping introduced in this PR:
> 
> ```toml
> [vars]
> VARIABLE = 'some-"string\\'
> ```

Yeah I would definitely consider that one of the edge-cases here that's not worth fixing.

---

My main goal in this PR was to handle the common things like dashes, numbers at start, etc. Given how flexible toml is, there's likely going to be lots of ways to break this, but with the changes here, it's a lot less fragile than the current release.

packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -1,11 +1,76 @@
...
+		expect(constructType("123invalid", "string")).toBe('"123invalid": string;');
+		expect(constructType("invalid-123", "string")).toBe(
+			'"invalid-123": string;'
+		);
+		expect(constructType("invalid 123", "string")).toBe(
+			'"invalid 123": string;'
+		);
+
+		expect(constructType("valid", 'a"', false)).toBe('valid: "a\\"";');
+		expect(constructType("valid", "a\\", false)).toBe('valid: "a\\"";');
```
dario-piotrowicz, line 68: I think there an extra `"` here
```suggestion
		expect(constructType("valid", "a\\", false)).toBe('valid: "a\\";');
```

Cherry: Good catch!, thanks Not sure how I missed this locally 🤔

packages/wrangler/src/type-generation.ts
dario-piotrowicz: yes, you're right! sorry I must have missed something in the code 🙇 

it looks good to me 🙂 

but you could please then add other unit tests for the non-string scenarios too? 🙏 

(such as `expect(constructType("valid", 2)).toBe("valid: 2;");` for example 🙂 )

Cherry:
Tests all passing now, woot! 🥳 Should be ready for a final review and then merge @dario-piotrowicz, thanks!

dario-piotrowicz (approved):

dario-piotrowicz:
@Cherry looks good to me! thanks a bunch for the improvement! 😄 

Unfortunately an approval from `wrangler` is still needed 😓, I'll post this PR in the internal channel to get someone to review this 🙂👍

Cherry:
With `cf-typegen` being a growing thing in many templates, it would be good to see this merged in soon if possible. 👀

petebacondarwin (approved):

PR #2746, opened 2023-02-17
feat(wrangler): able to deploy complex worker to Workers for Platforms

andrasbacsai:
What this PR solves / how to test:
We (https://webstudio.is) would like to deploy complex ([Remix-based](https://remix.run/)) customer applications to Workers for Platforms, which is not possible currently using wrangler and requires a custom script.

Adding an option `platformNamespace` to `wrangler` makes it possible to:
1. Upload all generated static assets to the KV store.
2. Upload the worker to the namespace instead of the normal Workers' namespace.

Associated docs issues/PR:

- [insert associated docs issue(s)/PR(s)]

Author has included the following, where applicable:

- [ ] Tests
- [x] Changeset

Reviewer has performed the following, where applicable:

- [x] Checked for inclusion of relevant tests
- [x] Checked for inclusion of a relevant changeset
- [ ] Checked for creation of associated docs updates
- [ ] Manually pulled down the changes and spot-tested

Fixes # [insert issue number].
No issue opened.

PR #5093, opened 2024-02-25
Feature: `dispatch-namespace` option in `wrangler deploy`

benycodes:
Ability to deploy workers to dispatch namespaces (Workers for Platforms product).

Usage: `wrangler deploy --dispatch-namespace <namespace>`

This is an improvement to PR #2746 because:
- Correct terminology of 'dispatch namespace' used instead of 'platform namespace'.
- Excludes tasks related to routes. custom domains. cron/schedules, and queue consumers when `dispatch-namespace` option used (these features aren't supported in Workers for Platforms)
- Allows for any name to be used for the script when `dispatch-namespace` option used, because Workers for Platforms doesn't check script names

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/13442
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/configuration.test.ts
@@ -1190,6 +1190,19 @@ describe("normalizeAndValidateConfig()", () => {
 			            - Expected \\"name\\" to be of type string, alphanumeric and lowercase with dashes only but got \\"!@#$%^&*(()\\"."
 		        `);
 			});
+
+			it("should accept any Worker name if the dispatch-namespace flag is used", () => {
+				const { diagnostics } = normalizeAndValidateConfig(
+					{
+						name: "example.com",
+						main: "index.js",
+					},
+					undefined,
+					{ env: undefined, "dispatch-namespace": "test-namespace" }
+				);
+				expect(diagnostics.hasWarnings()).toBe(false);
+				expect(diagnostics.hasErrors()).toBe(false);
+			});
 		});
 
 		describe("[build]", () => {
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -9015,6 +9015,34 @@ export default{
 			expect(std.err).toMatchInlineSnapshot(`""`);
 		});
 	});
+
+	describe("--dispatch-namespace", () => {
+		it("should upload to dispatch namespace", async () => {
+			writeWranglerToml();
+			const scriptContent = `
+      export default {
+				fetch() {
+					return new Response("Hello, World!");
+				}
+			}
+    `;
+			fs.writeFileSync("index.js", scriptContent);
+			mockUploadWorkerRequest({
+				expectedMainModule: "index.js",
+				expectedDispatchNamespace: "test-dispatch-namespace",
+			});
+
+			await runWrangler(
+				"deploy --dispatch-namespace test-dispatch-namespace index.js"
+			);
+			expect(std.out).toMatchInlineSnapshot(`
+			"Total Upload: xx KiB / gzip: xx KiB
+			Uploaded test-name (TIMINGS)
+			  Dispatch Namespace: test-dispatch-namespace
+			Current Deployment ID: Galaxy-Class"
+		`);
+		});
+	});
 });
 
 /** Write mock assets to the file system so they can be uploaded. */
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -136,10 +136,6 @@ export function normalizeAndValidateConfig(
 		diagnostics,
 		configPath,
 		rawConfig,
-		undefined,
-		undefined,
-		undefined,
-		undefined,
 		isDispatchNamespace
 	);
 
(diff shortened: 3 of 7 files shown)


WalshyDev (approved):

WalshyDev:
Thanks for the PR @benycodes 
I've pushed some changes to simplify along with adding a changeset and tests. From my side, this is good to go now :)

Appreciate it!

packages/wrangler/src/config/validation.ts
```diff
@@ -128,10 +128,17 @@ export function normalizeAndValidateConfig(
 		);
 	}
 
+	const isDispatchNamespace = typeof args["dispatch-namespace"] === "string";
```
CarmenPopoviciu, line 131: should we check for an empty string here? so smth like 
```
const isDispatchNamespace = typeof args["dispatch-namespace"] === "string" && args["dispatch-namespace"].trim() !== ""

File: packages/wrangler/src/config/validation.ts [hunk 2]
```diff
@@ -128,10 +128,17 @@ export function normalizeAndValidateConfig(
...
 	const topLevelEnv = normalizeAndValidateEnvironment(
 		diagnostics,
 		configPath,
-		rawConfig
+		rawConfig,
+		undefined,
+		undefined,
+		undefined,
+		undefined,
+		isDispatchNamespace
```
CarmenPopoviciu, line 141: I know technically it's the same, but should we add support for `isDispatchNamespace` in the first fn signature, instead of passing all these `undefined` params? Would make things much cleaner IMHO, which I assume was the reason behind having this particular fn signature to begin with

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -947,6 +956,15 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 	logger.log("Current Deployment ID:", deploymentId);
 }
 
+function deployWfpUserWorker(
+	dispatchNamespace: string,
+	deploymentId: string | null
+) {
+	// Will go under the "Uploaded" text
+	logger.log("  Dispatch Namespace:", dispatchNamespace);
```
CarmenPopoviciu, line 964: `logger.log("Dispatch Namespace:", dispatchNamespace);`

packages/wrangler/src/config/validation.ts
```diff
@@ -1130,7 +1153,7 @@ function normalizeAndValidateEnvironment(
 			topLevelEnv,
 			rawEnv,
 			"name",
-			isValidName,
+			isDispatchNamespace ? isString : isValidName,
```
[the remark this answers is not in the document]
WalshyDev: Yeah so the reason for this change is to change how the worker name is validated (see last point in original PR description), it seems we don't validate the script name in wfp like we do other parts. We probably should though so I'll revert this change.

packages/wrangler/src/config/validation.ts
WalshyDev: Yeah we could move it lower down

packages/wrangler/src/config/validation.ts
WalshyDev: Yeah good shout

packages/wrangler/src/config/validation.ts
WalshyDev: This actually isn't needed if we do the normal script name validation

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -9015,6 +9015,34 @@ export default{
 			expect(std.err).toMatchInlineSnapshot(`""`);
 		});
 	});
+
+	describe("--dispatch-namespace", () => {
```
CarmenPopoviciu, line 9019: can we please add one extra test in https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/__tests__/configuration.test.ts to verify we perform the correct `name` validation if arg.dispatchNamespace is present?

CarmenPopoviciu: we could test this by passing in a `name` value that is a `string` but **not** a valid Worker name. If `args["dispatch-namespace"]` is not undefined and a string, `name` validation should pass

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: > but there's a sig with fully required args which we couldn't put it in without it going to the end

packages/wrangler/src/__tests__/deploy.test.ts
WalshyDev: Added

packages/wrangler/src/config/validation.ts
WalshyDev: Haha yes very good point, moved.

packages/wrangler/src/__tests__/configuration.test.ts
```diff
@@ -1190,6 +1190,19 @@ describe("normalizeAndValidateConfig()", () => {
...
+			it("should accept any Worker name if the dispatch-namespace flag is used", () => {
+				const { diagnostics } = normalizeAndValidateConfig(
+					{
+						name: "example.com",
+						main: "index.js",
+					},
+					undefined,
+					{ env: undefined, "dispatch-namespace": "test-naemspace" }
+				);
+
```
CarmenPopoviciu, line 1203: can we pls check for warnings as well? 🙏 
`expect(diagnostics.hasWarnings()).toBe(false);`

CarmenPopoviciu:
thank you so much to both of you for taking this PR across the finish line <3

PR #5099, opened 2024-02-26
feat: cli flag to disable interactive dev session

### LINKED ISSUE #5063: 🚀 Feature Request: Add flag to hide wrangler CLI UI box
### Describe the solution

I'd like to request a cli flag or non-interactive option to be added that hides the CLI UI box that is created by wrangler that gives the following options

> [b] open a browser, [d] open Devtools, [l] turn off local mode, [c] clear console, [x] to exit

 I get this when running `wrangler dev` but this would be applicable to any other wrangler command that spawns that box.

Screenshot 2024-02-20 at 3 07 00 PM

My use case for this is I'm using the wrangler CLI in the context of a monorepo ([turborepo](https://turbo.build/)) which has it's own CLI UI to combine multiple process logs one view and it seems to be pretty incompatible with the wrangler UI as it repeats that box outline for each line logged. I still would like to see the logs that are produced by application just not the "UI" that surrounds it as It's quite hard to view logs this way for me.

Screenshot 2024-02-20 at 3 25 13 PM

I could see how this could also be useful in other non-interative CLI environments as well.

KaiSpencer:
Closes #5063 

For monorepos using tools like `turborepo` or `nx` it may be desirable to disable to interactive feature of the `wrangler dev` command.

The flag to disable this exists internally already in package.

This PR exposes this internal flag (that can be passed to `unstable_dev`) as a CLI arg, allowing the interactive session to be disabled.

I maintained consistency with the current variable naming.

- Tests
  - [x] Included (and fixture tests cover the defaults further)
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because: TODO
- Associated docs
  - [x] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/pull/13188
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/tender-buckets-confess.md
+++ b/.changeset/tender-buckets-confess.md
@@ -4,5 +4,5 @@
 
 feat: expose `--show-interactive-dev-session` flag
 
-This flag disables the interactive mode of the dev session, a feature that already exists internally but was not exposed to the user.
+This flag controls the interactive mode of the dev session, a feature that already exists internally but was not exposed to the user.
 This is useful for CI/CD environments where the interactive mode is not desired, or running in tools like `turbo` and `nx`.
--- a/.changeset/tender-buckets-confess.md
+++ b/.changeset/tender-buckets-confess.md
@@ -1,0 +1,8 @@
+---
+"wrangler": patch
+---
+
+feat: expose `--show-interactive-dev-session` flag
+
+This flag disables the interactive mode of the dev session, a feature that already exists internally but was not exposed to the user.
+This is useful for CI/CD environments where the interactive mode is not desired, or running in tools like `turbo` and `nx`.
--- a/packages/wrangler/src/__tests__/dev.test.tsx
+++ b/packages/wrangler/src/__tests__/dev.test.tsx
@@ -1258,7 +1258,8 @@ describe("wrangler dev", () => {
 			      --persist-to                                 Specify directory to use for local persistence (defaults to .wrangler/state)  [string]
 			      --live-reload                                Auto reload HTML pages when change is detected in local mode  [boolean]
 			      --test-scheduled                             Test scheduled events by visiting /__scheduled in browser  [boolean] [default: false]
-			      --log-level                                  Specify logging level  [choices: \\"debug\\", \\"info\\", \\"log\\", \\"warn\\", \\"error\\", \\"none\\"] [default: \\"log\\"]",
+			      --log-level                                  Specify logging level  [choices: \\"debug\\", \\"info\\", \\"log\\", \\"warn\\", \\"error\\", \\"none\\"] [default: \\"log\\"]
+			      --show-interactive-dev-session               Show interactive dev session  [boolean] [default: true]",
 			  "warn": "",
 			}
 		`);
@@ -1476,6 +1477,30 @@ describe("wrangler dev", () => {
 			",
 			}
 		`);
+		});
+	});
+
+	describe("--show-interactive-dev-session", () => {
+		it("should show interactive dev session by default", async () => {
+			fs.writeFileSync("index.js", `export default { }`);
+			await runWrangler("dev index.js");
+			expect(
+				(Dev as jest.Mock).mock.calls[0][0].showInteractiveDevSession
+			).toBeTruthy();
+		});
+		it("should show interactive dev session with --show-interactive-dev-session", async () => {
+			fs.writeFileSync("index.js", `export default { }`);
+			await runWrangler("dev index.js --show-interactive-dev-session");
+			expect(
+				(Dev as jest.Mock).mock.calls[0][0].showInteractiveDevSession
+			).toBeTruthy();
+		});
+		it("should not show interactive dev session with --show-interactive-dev-session=false", async () => {
+			fs.writeFileSync("index.js", `export default { }`);
+			await runWrangler("dev index.js --show-interactive-dev-session=false");
+			expect(
+				(Dev as jest.Mock).mock.calls[0][0].showInteractiveDevSession
+			).toBeFalsy();
 		});
 	});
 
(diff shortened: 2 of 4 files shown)


KaiSpencer:
Cheers @admah 
Some test failures `Timed out starting long-lived Wrangler` in the `pages-workersjs-directory`.
Expected flake to retry? or did I break something :D

petebacondarwin:
Probably flakes. Let me retry them.

KaiSpencer:
I can get most test to pass every time individually locally. 
The fixture `pages-workerjs-app` seems to be about 50/50 failing with the error on CI and passing the other half of the runs.

I am yet to get `pnpm test:ci` to pass every test at all locally in one go 🫣

petebacondarwin:
I think the problem is that internally `showInteractiveDevSession` defaults to `true` and for `wrangler pages dev` it was passing it through as `undefined`.
Inside the `DevImplementation` React component it was doing a further initialization default to `isRawModeSupported` which is now never being used because by the time it reaches there it is either `true` or `false` and not `undefined`.
I think you need to allow `undefined` to be a value for the CLI arg (i.e. not default it to `true` and then rely upon the internal defaulting to happen as it does now.

KaiSpencer:
> I think the problem is that internally `showInteractiveDevSession` defaults to `true` and for `wrangler pages dev` it was passing it through as `undefined`. Inside the `DevImplementation` React component it was doing a further initialization default to `isRawModeSupported` which is now never being used because by the time it reaches there it is either `true` or `false` and not `undefined`. I think you need to allow `undefined` to be a value for the CLI arg (i.e. not default it to `true` and then rely upon the internal defaulting to happen as it does now.
> 
> See
> 
> https://github.com/cloudflare/workers-sdk/blob/0c0949da60e3287c05a5884bb9f869ce5609a9a1/packages/wrangler/src/dev/dev.tsx#L190

Cheers @petebacondarwin 

Have removed the defaulting in yargs definitions. 

And subsequently removed the manual default to undefined, which will now be undefined with no arg, true with true arg and false with false arg

.changeset/tender-buckets-confess.md
```diff
@@ -0,0 +1,8 @@
+---
+"wrangler": patch
+---
+
+feat: expose `--show-interactive-dev-session` flag
+
+This flag disables the interactive mode of the dev session, a feature that already exists internally but was not exposed to the user.
```
petebacondarwin, line 7: Strictly it allows you to disable interactive mode by setting `--show-interactive-dev-session=false`.
Perhaps it can just say:
```suggestion
This flag controls the interactive mode of the dev session, a feature that already exists internally but was not exposed to the user.
```

packages/wrangler/src/dev.tsx
```diff
@@ -277,6 +277,10 @@ export function devOptions(yargs: CommonYargsArgv) {
 				// Yargs requires this to type log-level properly
 				default: "log" as LoggerLevel,
 			})
+			.option("show-interactive-dev-session", {
+				describe: "Show interactive dev session",
```
petebacondarwin, line 281: It might be worth putting the default into the `describe` since we don't want to tell yargs about it directly...

```suggestion
				describe: "Show interactive dev session (defaults to true)",
```

petebacondarwin (approved):
A couple of nits. But otherwise seems to work well. Thanks.

petebacondarwin:
Last of all we need to create a PR in https://github.com/cloudflare/cloudflare-docs to add this feature to the docs.

.changeset/tender-buckets-confess.md
KaiSpencer: Sure, have updated in [4d374f9](https://github.com/cloudflare/workers-sdk/pull/5099/commits/4d374f91b220510df6fe77c1900ea446879a5cb8)

packages/wrangler/src/dev.tsx
KaiSpencer: Yeah, makes sense. Have added in [4d374f9](https://github.com/cloudflare/workers-sdk/pull/5099/commits/4d374f91b220510df6fe77c1900ea446879a5cb8)

KaiSpencer:
> Last of all we need to create a PR in https://github.com/cloudflare/cloudflare-docs to add this feature to the docs.

PR for docs https://github.com/cloudflare/cloudflare-docs/pull/13188

packages/wrangler/src/__tests__/dev.test.tsx
```diff
@@ -1268,7 +1268,8 @@ describe("wrangler dev", () => {
 			      --persist-to                                 Specify directory to use for local persistence (defaults to .wrangler/state)  [string]
 			      --live-reload                                Auto reload HTML pages when change is detected in local mode  [boolean]
 			      --test-scheduled                             Test scheduled events by visiting /__scheduled in browser  [boolean] [default: false]
-			      --log-level                                  Specify logging level  [choices: \\"debug\\", \\"info\\", \\"log\\", \\"warn\\", \\"error\\", \\"none\\"] [default: \\"log\\"]",
+			      --log-level                                  Specify logging level  [choices: \\"debug\\", \\"info\\", \\"log\\", \\"warn\\", \\"error\\", \\"none\\"] [default: \\"log\\"]
+			      --show-interactive-dev-session               Show interactive dev session  [boolean]",
```
petebacondarwin, line 1272: I bet this needs updating now

packages/wrangler/src/pages/dev.ts
```diff
@@ -217,6 +217,10 @@ export function Options(yargs: CommonYargsArgv) {
 				choices: ["debug", "info", "log", "warn", "error", "none"] as const,
 				describe: "Specify logging level",
 			},
+			"show-interactive-dev-session": {
+				describe: "Show interactive dev session",
```
petebacondarwin, line 221:
```suggestion
				describe: "Show interactive dev session (defaults to true)",
```

petebacondarwin (approved):

PR #5102, opened 2024-02-26
Add DeliveryDelay and RetryDelay to create Queue and Add Worker Consumer commands.

pmiguel:
Fixes # MQ-499 internally 

CR for merge under LDW: # CR-850147

Adds support for message delivery delay on create new Queue, and retry delay when create new consumer.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): https://github.com/cloudflare/cloudflare-docs/issues/13370 / https://github.com/cloudflare/cloudflare-docs/pull/13545/commits/ab005895ea739fdbc781f428ad60c7b602e4e46c
  - [ ] Not necessary because:


packages/wrangler/src/queues/cli/commands/create.ts
```diff
@@ -1,25 +1,65 @@
...
+				type: "string",
+				demandOption: true,
+				description: "The name of the queue",
+			})
+		.options({
+			"delivery-delay": {
+				type: "number",
+				describe: "How long a published messages should be delayed for, in seconds. Must be a positive integer",
+			},
+			"no-delivery-delay": {
```
jbwcloudflare, line 24: I'm actually not sure why we need this option (for `create` Queue, at least).

Couldn't the user just omit the `delivery-delay` option?

(It might be needed for the `update` Queue command, when we add that)

packages/wrangler/src/queues/cli/commands/create.ts
```diff
@@ -1,25 +1,65 @@
...
+			}
+		});
+}
+
+function createBody(args: StrictYargsOptionsToInterface<typeof options>): CreateQueueBody {
+	const body: CreateQueueBody = {
+		queue_name: args.name
+	}
+
+	// Workaround, Yargs does not play nicely with both --parameter and --no-parameter set.
```
jbwcloudflare, line 37: I just did some googling and found: https://yargs.js.org/docs/#api-reference-conflictsx-y

Have you tried using `conflicts`? (if we actually keep both options...)

packages/wrangler/src/queues/cli/commands/create.ts
```diff
@@ -1,25 +1,65 @@
...
-	await createQueue(config, { queue_name: args.name });
-	logger.log(`Created queue ${args.name}.`);
+	const body = createBody(args);
+	try {
+		logger.log(`Creating queue ${args.name}.`);
+		await createQueue(config, body);
+		logger.log(`Created queue ${args.name}.`);
+	} catch(e) {
+		handleFetchError(e as {code?: number})
+		throw e;
```
jbwcloudflare, line 63: It looks like `handleFetchError` will also `throw` the error, can we get rid of one?

packages/wrangler/src/queues/cli/commands/consumer/add.ts
```diff
@@ -42,15 +41,21 @@ export function options(yargs: CommonYargsArgv) {
 				describe:
 					"The maximum number of concurrent consumer Worker invocations. Must be a positive integer",
 			},
+			"retry-delay": {
+				type: "number",
+				describe: "How long a retried messages should be delayed for, in seconds. Must be a positive integer",
+				number: true,
+			},
+			"no-retry-delay": {
```
jbwcloudflare, line 49: Similar comment here, is this really needed for the `add` command?

packages/wrangler/src/queues/cli/commands/consumer/add.ts
```diff
@@ -63,9 +68,33 @@ export async function handler(
 			max_concurrency: args.maxConcurrency,
 		},
 		dead_letter_queue: args.deadLetterQueue,
-	};
+	} as PostConsumerBody;
```
jbwcloudflare, line 71: Avoid using `as`....

If you want to give the `body` a type, then do: `const body : PostConsumerBody = {...}`

packages/wrangler/src/__tests__/queues.test.ts
```diff
@@ -247,6 +259,40 @@ describe("wrangler", () => {
...
+				const requests = mockCreateRequest("testQueue", { delivery_delay: 10 });
+				await runWrangler("queues create testQueue --delivery-delay=10");
+				expect(std.out).toMatchInlineSnapshot(`
+					"Creating queue testQueue.
+					Created queue testQueue."
+			  `);
+				expect(requests.count).toEqual(1);
+			});
+
+			it("should send send queue settings with 0 delivery delay when noDeliveryDelay is set", async () => {
```
jbwcloudflare, line 273: Ideally the request would just omit the delivery_delay (or set it to `undefined`) instead of `0` IMO

jbwcloudflare (commented):
Looking good, I think there are some test failures that need to be fixed though.

packages/wrangler/src/queues/cli/commands/create.ts
pmiguel: Didn't use the right keywords to find that one. Will take a look.

pmiguel: Didn't work. Apparently "no-x" has a special meaning, meaning negation of x. It's not viewed as a separate parameter, it's viewed as x.

jbwcloudflare: OK no worries, we don't need to use it

packages/wrangler/src/queues/cli/commands/consumer/add.ts
```diff
@@ -65,7 +65,31 @@ export async function handler(
 		dead_letter_queue: args.deadLetterQueue,
 	};
 
-	logger.log(`Adding consumer to queue ${args.queueName}.`);
-	await postConsumer(config, args.queueName, body);
-	logger.log(`Added consumer to queue ${args.queueName}.`);
+
+	// Workaround, Yargs does not play nicely with both --parameter and --no-parameter set.
```
jbwcloudflare, line 69: This comment is stale now right? (now that the --no-delivery-delay is gone)

packages/wrangler/src/__tests__/queues.test.ts
```diff
@@ -12,6 +14,14 @@ describe("wrangler", () => {
 	runInTempDir();
 	const std = mockConsoleMethods();
 
+	const { setIsTTY } = useMockIsTTY();
```
jbwcloudflare, line 17: Are these changes related to getting the error handling to work properly in your tests? Mind adding a comment to explain what these are for?

jbwcloudflare:
This looks solid to me, I think you are good to open it up for review by the wrangler team.

packages/wrangler/src/queues/cli/commands/create.ts
pmiguel: As discussed, at least for these, removed. Closes back a can of worms with the negated parameters at least for now.

packages/wrangler/src/__tests__/queues.test.ts
pmiguel: Was due to trial an error. Had no effect. Will remove it.

packages/wrangler/src/queues/cli/commands/create.ts
```diff
@@ -1,25 +1,62 @@
...
+function createBody(
+	args: StrictYargsOptionsToInterface<typeof options>
+): CreateQueueBody {
+	const body: CreateQueueBody = {
+		queue_name: args.name,
+	};
+
+	if (Array.isArray(args.deliveryDelay)) {
+		throw new CommandLineArgsError(
+			`Error: can't use more than a delay setting.`
```
mrbbot, line 37:
```suggestion
			"Cannot specify --delivery-delay multiple times"
```

mrbbot (approved):

petebacondarwin:
What does this mean?

```
Associated docs
 Not necessary because: full range of options not yet documented.
```

Are you saying that you don't need to add docs for this because there are none at all yet?
At the least I think it would be good to create a placeholder issue in the docs repo, to track that this needs to be done at some point (soonish?).

pmiguel:
> What does this mean?
> 
> ```
> Associated docs
>  Not necessary because: full range of options not yet documented.
> ```
> 
> Are you saying that you don't need to add docs for this because there are none at all yet? At the least I think it would be good to create a placeholder issue in the docs repo, to track that this needs to be done at some point (soonish?).

There is documentation (https://github.com/cloudflare/cloudflare-docs/blob/production/content/workers/wrangler/commands.md#queues), it's just the `[OPTIONS]` that are lacking. I'll check if we already have an issue for that.

petebacondarwin:
I've created an issue in the docs repo to track the addition of the OPTIONS descriptions: https://github.com/cloudflare/cloudflare-docs/issues/13370.

I see you have changed the title to `[do not merge]`. So I will also change this to "draft". Please mark it as "ready to review" when you are ready.

pmiguel:
Approved under LDW with CR-850147

mrbbot (approved):

PR #5106, opened 2024-02-26
Workaround "Error: Network connection lost." in POSTs from workerd to workerd

RamIdeas:
Potentially fixes #5905

This PR tries to fix POST requests to the UserWorker via the ProxyWorker with at least 2 bytes in the body failing at a rate of roughly 1 in 3 to 1 in 5.

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] Included
  - [ ] Not necessary because: TODO
- Associated docs
  - [x] Issue(s)/PR(s): #5905
  - [ ] Not necessary because:

**Note for PR author:**


fixtures/dev-env/tests/index.test.ts
```diff
@@ -649,3 +649,42 @@ describe("startDevWorker: ProxyController", () => {
 		await expect(res.text()).resolves.toBe("UserWorker:3");
 	});
 });
+
+describe("regressions", () => {
+	// https://github.com/cloudflare/workers-sdk/issues/5095
+	test("#5095 requests where body (of >= 2 bytes) is not drained sometimes fail", async () => {
```
penalosa, line 655:
```suggestion
	test("requests where body (of >= 2 bytes) is not drained sometimes fail", async () => {
```
Could this description describe what the test is testing? Is it checking that the regression is fixed, or demonstrating a code path that fails?

fixtures/dev-env/tests/index.test.ts
```diff
@@ -649,3 +649,42 @@ describe("startDevWorker: ProxyController", () => {
...
+					.fetch(run.proxyWorkerUrl, {
+						method: "POST",
+						body: new Uint8Array(2), // 2 bytes
+					})
+					.then((res) => {
+						if (!res.ok) errorCount++;
+						requestCount++;
+					});
+			})
+		);
```
penalosa, line 685: Is there a reason this can't be a for loop?

fixtures/dev-env/tests/index.test.ts
```diff
@@ -649,3 +649,42 @@ describe("startDevWorker: ProxyController", () => {
...
+		let errorCount = 0;
+
+		await Promise.allSettled(
+			Array.from({ length: COUNT }).map(async (_, i) => {
+				await new Promise((r) => setTimeout(r, i * STEP_DELAY));
+
+				return undici
+					.fetch(run.proxyWorkerUrl, {
+						method: "POST",
+						body: new Uint8Array(2), // 2 bytes
```
penalosa, line 678: Why 2 bytes?

fixtures/dev-env/tests/index.test.ts
```diff
@@ -649,3 +649,42 @@ describe("startDevWorker: ProxyController", () => {
...
+			script: `
+				export default {
+					async fetch(request) {
+						return new Response("UserWorker:1");
+					}
+				}
+			`,
+		});
+
+		const COUNT = 30;
```
penalosa, line 666: This feels like it should be higher—are we guaranteed to see the failure case within 30 iterations?

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
```diff
@@ -0,0 +1,14 @@
...
+		return await middlewareCtx.next(request, env);
+	} finally {
+		if (!request.bodyUsed) {
+			_ctx.waitUntil(request.arrayBuffer());
+		}
+	}
+};
+
+export default drainBody;
+export const wrap = undefined;
```
penalosa, line 14:
```suggestion

```

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
```diff
@@ -0,0 +1,14 @@
+import type { Middleware } from "./common";
+
+const drainBody: Middleware = async (request, env, _ctx, middlewareCtx) => {
```
penalosa, line 3:
```suggestion
const drainBody: Middleware = async (request, env, ctx, middlewareCtx) => {
```

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
```diff
@@ -0,0 +1,14 @@
...
+	try {
+		return await middlewareCtx.next(request, env);
+	} finally {
+		if (!request.bodyUsed) {
+			_ctx.waitUntil(request.arrayBuffer());
```
penalosa, line 8:
```suggestion
			ctx.waitUntil(request.arrayBuffer());
```

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
```diff
@@ -0,0 +1,14 @@
+import type { Middleware } from "./common";
+
+const drainBody: Middleware = async (request, env, _ctx, middlewareCtx) => {
+	try {
+		return await middlewareCtx.next(request, env);
+	} finally {
+		if (!request.bodyUsed) {
```
penalosa, line 7: Clever!

fixtures/dev-env/tests/index.test.ts
petebacondarwin: Any smaller and the bug does not appear.

fixtures/dev-env/tests/index.test.ts
petebacondarwin: Yeah this can be simpler and still trigger the regression:

```ts
		for (let i = 0; i < COUNT; i++) {
			const response = await undici.fetch(run.proxyWorkerUrl, {
				method: "POST",
				body: new Uint8Array(2), // 2 bytes
			});
			requestCount++;
			if (!response.ok) errorCount++;
		}
```

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
Skye-31: I wonder if `request.body.cancel()` would work better here?

petebacondarwin: Interesting!!

petebacondarwin: Nope. That didn't help. Still fails the test.

fixtures/dev-env/tests/index.test.ts
RamIdeas: This is what I wrote originally. But I wanted to see if parallel requests would cause the problem to occur more often or less often. My hypothesis was the former.

In fact, if you send them within quick succession (all at once, or less than 4ms between each), the problem doesn't even manifest.

In conclusion, I don't think the problem is as simple as "request bodies need to be drained". The workaround in this PR helps. But either it delays a race condition or it releases pressure on reaching a memory limit.

fixtures/dev-env/tests/index.test.ts
RamIdeas: I was seeing something like 30% failures so 30 felt like enough. But I wasn't done tuning the value anyway. I'm still investigating (should've opened this PR as a draft 🤦)

fixtures/dev-env/tests/index.test.ts
RamIdeas: This bit was actually in the test name 😄

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
RamIdeas: I copied another middleware file. I assume it was required for the module interface of a "middleware". Not necessary then?

penalosa: It shouldn't be, unless something has changed since I added the `wrap` export. It _should_ be optional

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
petebacondarwin: This is no longer used.

petebacondarwin (approved):
As long as CI is green

.changeset/empty-zoos-beam.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: automatically drain incoming request bodies
+
+Previously, requests sent to `wrangler dev` with unconsumed bodies could result in `Network connection lost` errors. This change attempts to work around the issue by ensuring incoming request bodies are drained if they're not used. This is a temporary fix whilst we try to address the underlying issue. Whilst we don't think this change will introduce any other issues, it can be disabled by setting the `WRANGLER_DISABLE_REQUEST_BODY_DRAINING=true` environment variable. Note this fix is only applied if you've enabled Wrangler's bundling—`--no-bundle` mode continues to have the previous behaviour.
```
petebacondarwin, line 7: I wonder if we should suggest that not draining bodies is bad practice generally? And that as a first call devs should attempt to do so rather than relying upon our workaround?

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
```diff
@@ -0,0 +1,14 @@
+import type { Middleware } from "./common";
+
+const drainBody: Middleware = async (request, env, _ctx, middlewareCtx) => {
+	try {
+		return await middlewareCtx.next(request, env);
+	} finally {
+		if (request.body !== null && !request.bodyUsed) {
+			const reader = request.body.getReader();
+			while (!(await reader.read()).done) {}
```
penalosa, line 9: Why was the change from cancelling the stream required?

.changeset/empty-zoos-beam.md
RamIdeas: It's not bad practice at all. If you have a POST endpoint which required auth, you will want to be able to return a 403 without reading the body at all

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
petebacondarwin: Cancelling the stream, via `request.body.cancel()` did not resolve the problem.
The original PR drained the stream by dumping it into a buffer (e.g. `request.arrayBuffer()` but this creates unnecessary memory churn.
So the current approach literally just iterates over the chunks and throws them away.

RamIdeas, line 9: Do we need try..catch around this await?

.changeset/empty-zoos-beam.md
petebacondarwin: Even in that case it would be better to cancel the body (if that worked, which it doesn't appear to)?
I think one reason why this is so tricky for the library/runtime author to resolve is that it is not always clear whether the user actually wants the body stream to be left open?

packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
petebacondarwin: That sounds like a good idea.
Out of curiosity, in what way might this fail? Like if the client disconnects during the drain?

RamIdeas: I actually don't know. But an async loop reading chunks of a stream seems like something that _could_ fail. My original ctx.waitUntil handled this by allowing the arrayBuffer promise to reject without affecting the response promise. We should handle it to be safe.

.changeset/empty-zoos-beam.md
RamIdeas: This is a workaround for a workerd bug. The runtime team should fix the underlying bug. Do we even know what it is yet? Is it a memory leak of request body streams left open which eventually reach a limit?

If this is something that should be handled by users, the runtime team are gonna have to tell us that and really justify it.

PR #5115, opened 2024-02-28
Implement `versions deploy` command

RamIdeas:

This PR implements the `versions deploy` command as a sibling command to the, already merged, `versions upload` command. Both are still behind the `--experimental-gradual-rollouts` flag.

Note: in the commands below, it is assumed you have followed the install instructions of the prerelease build (from the bot comment in this PR). You can, instead, replace "wrangler" with the prerelease URL from the bot comment.

The most common usage will be without args where the prompts will walk you through selecting Versions and percentages for each and adding a deployment message. A sane default is provided for each percentage, evenly distributing the remaining percentage across the remaining versions.
```sh
$ npx wrangler versions deploy    --experimental-gradual-rollouts
```

You can also specify positional args in the form of `<version-id>` or `<version-id>@<percentage>` which will set the default selected Versions and the default associated percentage for each. The prompts are still presented but with prefilled defaults that the user can press ENTER to confirm.
```sh
# 1 version preselected by default
$ npx wrangler versions deploy 095f00a7-23a7-43b7-a227-e4c97cab5f22    --experimental-gradual-rollouts

# 2 versions preselected by default (the default percentage for each will be 50%)
$ npx wrangler versions deploy 095f00a7-23a7-43b7-a227-e4c97cab5f22  1a88955c-2fbd-4a72-9d9b-3ba1e59842f2    -experimental-gradual-rollouts

# 2 versions preselected by default – one with explicit 10% traffic, the other with implicit 90%
$ npx wrangler versions deploy 095f00a7-23a7-43b7-a227-e4c97cab5f22@10%   1a88955c-2fbd-4a72-9d9b-3ba1e59842f2    --experimental-gradual-rollouts
# ^ the 2nd version will have 90% by default (evenly distributed after the first version was assigned 10%)

# 2 versions preselected by default – both with explicit traffic 10% and 90%
$ npx wrangler versions deploy 095f00a7-23a7-43b7-a227-e4c97cab5f22@10%   1a88955c-2fbd-4a72-9d9b-3ba1e59842f2@90%    --experimental-gradual-rollouts
```

You can also specify named args `--version-id` and `--percentage` repeatedly. All `--version-id` args are read in order into an array. All `--percentage` args are read in order into an array. The indexes of each array are then used to correlate. By specifying less `--percentage` args than `--version-id` args, the percentages for the excess versions are deemed unspecified, like with positional args without `@<percentage>`, and will have the remaining percentage distributed evenly as with positional/prompted percentages.

```sh
$ npx wrangler versions deploy    --experimental-gradual-rollouts \
    --version-id 095f00a7-23a7-43b7-a227-e4c97cab5f22 \
    --percentage 10% \
    --version-id 1a88955c-2fbd-4a72-9d9b-3ba1e59842f2 \
    --percentage 90% 
# is equivalent to
$ npx wrangler versions deploy    --experimental-gradual-rollouts \
    --version-id 095f00a7-23a7-43b7-a227-e4c97cab5f22 \
    --version-id 1a88955c-2fbd-4a72-9d9b-3ba1e59842f2 \
    --percentage 10% \
    --percentage 90% 
```

A mixture of positional and named args will work and should be relatively predictable, with positional args taking precedence.

By using the `--yes` flag, the prompt defaults will be automatically accepted. This will allow the following: 
```sh
# deploy 2 versions with (implicit) 50% split – with positional args
$ npx wrangler versions deploy 095f00a7-23a7-43b7-a227-e4c97cab5f22  1a88955c-2fbd-4a72-9d9b-3ba1e59842f2  --yes    --experimental-gradual-rollouts

# deploy 2 versions with (explicit) 10% and (implicit) 90% split – with positional args
$ npx wrangler versions deploy 095f00a7-23a7-43b7-a227-e4c97cab5f22@10%   1a88955c-2fbd-4a72-9d9b-3ba1e59842f2  --yes    --experimental-gradual-rollouts

# deploy 2 versions with (explicit) 10% and (explicit) 90% split – with positional args
$ npx wrangler versions deploy 095f00a7-23a7-43b7-a227-e4c97cab5f22@10%   1a88955c-2fbd-4a72-9d9b-3ba1e59842f2@90%  --yes    --experimental-gradual-rollouts

# deploy 2 versions with (explicit) 10% and (implicit) 90% split – with named args
$ npx wrangler versions deploy  --yes    --experimental-gradual-rollouts \
    --version-id 095f00a7-23a7-43b7-a227-e4c97cab5f22 \
    --percentage 10% \
    --version-id 1a88955c-2fbd-4a72-9d9b-3ba1e59842f2

# deploy 2 versions with (explicit) 10% and (explicit) 90% split – with named args
$ npx wrangler versions deploy  --yes    --experimental-gradual-rollouts \
    --version-id 095f00a7-23a7-43b7-a227-e4c97cab5f22 \
    --percentage 10% \
    --version-id 1a88955c-2fbd-4a72-9d9b-3ba1e59842f2 \
    --percentage 90% 
```

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO before merging
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO before merging
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: internal command for now, documentation coming before open beta.

**Note for PR author:**

### DIFF
--- a/packages/cli/interactive.ts
+++ b/packages/cli/interactive.ts
@@ -15,13 +15,14 @@ import type { Prompt } from "@clack/core";
 const logUpdate = createLogUpdate(process.stdout);
 
 export type Arg = string | boolean | string[] | undefined | number;
-const grayBar = gray(shapes.bar);
-const blCorner = gray(shapes.corners.bl);
-const leftT = gray(shapes.leftT);
+export const grayBar = gray(shapes.bar);
+export const blCorner = gray(shapes.corners.bl);
+export const leftT = gray(shapes.leftT);
 
 export type Option = {
-	label: string;
-	value: string;
+	label: string; // user-visible string
+	sublabel?: string; // user-visible string
+	value: string; // underlying key
 	hidden?: boolean;
 };
 
@@ -38,6 +39,8 @@ export type BasePromptConfig = {
 	format?: (value: Arg) => string;
 	// Returns a user displayed error if the value is invalid
 	validate?: (value: Arg) => string | void;
+	// override some/all renderers (can be used for custom renderers before hoisting back into shared code)
+	renderers?: Partial<ReturnType<typeof getRenderers>>;
 };
 
 export type TextPromptConfig = BasePromptConfig & {
@@ -87,7 +90,10 @@ type RenderProps =
 	| Omit<SelectRefreshablePrompt, "prompt">;
 
 export const inputPrompt = async <T = string>(promptConfig: PromptConfig) => {
-	const renderers = getRenderers(promptConfig);
+	const renderers = {
+		...getRenderers(promptConfig),
+		...promptConfig.renderers,
+	};
 
 	let prompt:
 		| SelectPrompt<Option>
@@ -150,7 +156,7 @@ export const inputPrompt = async <T = string>(promptConfig: PromptConfig) => {
 		prompt = new TextPrompt({
 			...promptConfig,
 			initialValue: promptConfig.initialValue,
-			defaultValue: String(promptConfig.defaultValue),
+			defaultValue: String(promptConfig.defaultValue ?? ""),
 			render() {
 				return dispatchRender(this, prompt);
 			},
@@ -213,14 +219,10 @@ export const getRenderers = (config: PromptConfig) => {
 };
 
 const getTextRenderers = (config: TextPromptConfig) => {
-	const {
-		defaultValue,
-		question,
-		helpText: _helpText,
-		format: _format,
-	} = config;
-	const helpText = _helpText ?? "";
-	const format = _format ?? ((val: Arg) => String(val));
+	const { question } = config;
+	const helpText = config.helpText ?? "";
+	const format = config.format ?? ((val: Arg) => String(val));
+	const defaultValue = config.defaultValue?.toString() ?? "";
 
 	return {
 		initial: () => [
@@ -230,9 +232,7 @@ const getTextRenderers = (config: TextPromptConfig) => {
 		],
 		active: ({ value }: { value: Arg }) => [
 			`${blCorner} ${bold(question)} ${dim(helpText)}`,
-			`${space(2)}${format(
-				value || dim(typeof defaultValue === "string" ? defaultValue : ``)
-			)}`,
+			`${space(2)}${format(value || dim(defaultValue))}`,
 			``, // extra line for readability
 		],
 		error: ({ value, error }: { value: Arg; error: string }) => [
@@ -242,7 +242,8 @@ const getTextRenderers = (config: TextPromptConfig) => {
 			`${space(2)}${format(value)}`,
 			``, // extra line for readability
 		],
-		submit: ({ value }: { value: Arg }) => renderSubmit(config, format(value)),
+		submit: ({ value }: { value: Arg }) =>
+			renderSubmit(config, format(value ?? "")),
 		cancel: handleCancel,
 	};
 };
@@ -261,15 +262,16 @@ const getSelectRenderers = (
 			const active = i === cursor;
 			const isInListOfValues =
 				Array.isArray(value) && value.includes(optionValue);
-			const color = isInListOfValues || active ? blue : dim;
+			const color = isInListOfValues || active ? blue : white;
 			const text = active ? color.underline(optionLabel) : color(optionLabel);
+			const sublabel = opt.sublabel ? color.grey(opt.sublabel) : "";
 
 			const indicator =
 				isInListOfValues || (active && !Array.isArray(value))
 					? color(shapes.radioActive)
 					: color(shapes.radioInactive);
 
-			return `${space(2)}${indicator} ${text}`;
+			return `${space(2)}${indicator} ${text} ${sublabel}`;
 		};
 
 		const renderOptionCondition = (_: unknown, i: number): boolean => {
--- a/packages/cli/interactive.ts
+++ b/packages/cli/interactive.ts
@@ -8,7 +8,7 @@ import {
 import { createLogUpdate } from "log-update";
 import { blue, bold, brandColor, dim, gray, white } from "./colors";
 import SelectRefreshablePrompt from "./select-list";
-import { cancel, newline, shapes, space, status } from "./index";
+import { cancel, crash, logRaw, newline, shapes, space, status } from "./index";
 import type { OptionWithDetails } from "./select-list";
 import type { Prompt } from "@clack/core";
 
@@ -33,13 +33,15 @@ export type BasePromptConfig = {
 	helpText?: string;
 	// The value to use by default
 	defaultValue?: Arg;
+	// Accept the initialValue/defaultValue as if the user pressed ENTER when prompted
+	acceptDefault?: boolean;
 	// The status label to be shown after submitting
 	label: string;
 	// Pretty-prints the value in the interactive prompt
 	format?: (value: Arg) => string;
 	// Returns a user displayed error if the value is invalid
 	validate?: (value: Arg) => string | void;
-	// override some/all renderers (can be used for custom renderers before hoisting back into shared code)
+	// Override some/all renderers (can be used for custom renderers before hoisting back into shared code)
 	renderers?: Partial<ReturnType<typeof getRenderers>>;
 };
 
@@ -89,7 +91,25 @@ type RenderProps =
 	| Omit<ConfirmPrompt, "prompt">
 	| Omit<SelectRefreshablePrompt, "prompt">;
 
-export const inputPrompt = async <T = string>(promptConfig: PromptConfig) => {
+function acceptDefault<T>(
+	promptConfig: PromptConfig,
+	renderers: Pick<ReturnType<typeof getRenderers>, "submit">,
+	initialValue: T
+): T {
+	const error = promptConfig.validate?.(initialValue as Arg);
+	if (error) {
+		crash(error);
+	}
+
+	const lines = renderers.submit({ value: initialValue as Arg });
+	logRaw(lines.join("\n"));
+
+	return initialValue as T;
+}
+
+export const inputPrompt = async <T = string>(
+	promptConfig: PromptConfig
+): Promise<T> => {
 	const renderers = {
 		...getRenderers(promptConfig),
 		...promptConfig.renderers,
@@ -109,18 +129,30 @@ export const inputPrompt = async <T = string>(promptConfig: PromptConfig) => {
 	};
 
 	if (promptConfig.type === "select") {
+		const initialValue = String(promptConfig.defaultValue);
+
+		if (promptConfig.acceptDefault) {
+			return acceptDefault<T>(promptConfig, renderers, initialValue as T);
+		}
+
 		prompt = new SelectPrompt({
 			...promptConfig,
 			options: promptConfig.options.filter((o) => !o.hidden),
-			initialValue: String(promptConfig.defaultValue),
+			initialValue,
 			render() {
 				return dispatchRender(this, prompt);
 			},
 		});
 	} else if (promptConfig.type === "confirm") {
+		const initialValue = Boolean(promptConfig.defaultValue);
+
+		if (promptConfig.acceptDefault) {
+			return acceptDefault<T>(promptConfig, renderers, initialValue as T);
+		}
+
 		prompt = new ConfirmPrompt({
 			...promptConfig,
-			initialValue: Boolean(promptConfig.defaultValue),
+			initialValue,
 			active: promptConfig.activeText || "",
 			inactive: promptConfig.inactiveText || "",
 			render() {
@@ -134,6 +166,11 @@ export const inputPrompt = async <T = string>(promptConfig: PromptConfig) => {
 		} else if (promptConfig.defaultValue !== undefined) {
 			initialValues = [String(promptConfig.defaultValue)];
 		}
+
+		if (promptConfig.acceptDefault) {
+			return acceptDefault<T>(promptConfig, renderers, initialValues as T);
+		}
+
 		prompt = new MultiSelectPrompt({
 			...promptConfig,
 			options: promptConfig.options,
@@ -143,16 +180,29 @@ export const inputPrompt = async <T = string>(promptConfig: PromptConfig) => {
 			},
 		});
 	} else if (promptConfig.type === "list") {
+		const initialValue = String(promptConfig.defaultValue);
+
+		if (promptConfig.acceptDefault) {
+			return acceptDefault<T>(promptConfig, renderers, initialValue as T);
+		}
+
 		prompt = new SelectRefreshablePrompt({
 			...promptConfig,
 			onRefresh:
 				promptConfig.onRefresh ?? (() => Promise.resolve(promptConfig.options)),
-			initialValue: String(promptConfig.defaultValue),
+			initialValue,
 			render() {
 				return dispatchRender(this, prompt);
 			},
 		});
 	} else {
+		const initialValue =
+			promptConfig.initialValue ?? String(promptConfig.defaultValue ?? "");
+
+		if (promptConfig.acceptDefault) {
+			return acceptDefault<T>(promptConfig, renderers, initialValue as T);
+		}
+
 		prompt = new TextPrompt({
 			...promptConfig,
 			initialValue: promptConfig.initialValue,
--- a/packages/cli/interactive.ts
+++ b/packages/cli/interactive.ts
@@ -567,6 +567,30 @@ export const spinner = (
 	};
 };
 
+type FactoryOrValue<T> = T | (() => T);
+const unwrapFactory = <T>(input: FactoryOrValue<T>): T => {
+	const output = typeof input === "function" ? (input as () => T)() : input;
+	return output;
+};
+export const spinnerWhile = async <T>(opts: {
+	promise: FactoryOrValue<Promise<T>>;
+	startMessage: FactoryOrValue<string>;
+	endMessage?: FactoryOrValue<string>;
+	spinner?: ReturnType<typeof spinner>;
+}): Promise<T> => {
+	const s = opts.spinner ?? spinner();
+
+	s.start(unwrapFactory(opts.startMessage));
+
+	try {
+		const result = await unwrapFactory(opts.promise);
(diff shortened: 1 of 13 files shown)


petebacondarwin (commented):
There's a lot of new code in here with no tests...

packages/wrangler/src/index.ts
```diff
@@ -707,12 +712,19 @@ export function createCLIParser(argv: string[]) {
...
+			return versionYargs
+				.command(
+					"upload",
+					"Uploads your Worker code and config as a new version [beta]",
+					versionsUploadOptions,
+					versionsUploadHandler
+				)
+				.command(
+					"deploy",
+					"Safely roll out new versions of your Worker by splitting traffic between multiple versions",
```
mrbbot, line 724:
```suggestion
					"Safely roll out new versions of your Worker by splitting traffic between multiple versions [beta]",
```

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,652 @@
...
+import { printWranglerBanner } from "../update-check";
+import { requireAuth } from "../user";
+import type {
+	CommonYargsArgv,
+	StrictYargsOptionsToInterface,
+} from "../yargs-types";
+
+const EPSILON = 0.001; // used to avoid floating-point errors. Comparions to a value +/- EPSILON will mean "roughly equals the value".
+const BLANK_INPUT = "-"; // To be used where optional user-input is displayed and the value is nullish
+const ZERO_WIDTH_SPACE = "​"; // Some log lines get trimmed and so, to indent, the line is prefixed with a zero-width space
```
mrbbot, line 23: Could you use `"\u200B"` instead here, so it's clear this string actually has something in it? 😅

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,652 @@
...
+		source: "api" | string;
+		author_id: string;
+		author_email: string;
+	};
+	annotations: Record<string, string> & {
+		"workers/triggered_by"?: "upload" | string;
+		"workers/message"?: string;
+		"workers/tag"?: string;
+	};
+	// resources: { script: [Object]; script_runtime: [Object]; bindings: [] };
```
mrbbot, line 65:
```suggestion
```

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,652 @@
...
+			describe: "Don't actually deploy",
+			type: "boolean",
+			default: false,
+		})
+		.option("max-versions", {
+			hidden: true, // experimental, not supported long-term
+			describe: "Maximum allowed versions to select",
+			type: "number",
+			default: 2,
+		});
```
mrbbot, line 120: Is there server-side validation for this too? Will any user be able to deploy more than 2 versions with this flag even though it's hidden?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,652 @@
...
+	// 	label: "Deployment tag",
+	// 	defaultValue: args.tag,
+	// 	question: "Add a deployment tag",
+	// 	helpText: "(optional)",
+	// });
+
+	// if (args.dryRun) {
+	// 	cli.cancel("--dry-run: exiting");
+	// 	return;
+	// }
```
mrbbot, line 187: Should this code be removed?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,652 @@
...
+		message
+		// , tag
+	);
+
+	const elapsedMilliseconds = Date.now() - start;
+	const elapsedSeconds = elapsedMilliseconds / 1000;
+	const elapsedString = `${elapsedSeconds.toFixed(2)} sec`;
+	const trafficSummaryString = Array.from(confirmedVersionTraffic)
+		.map(([versionId, percentage]) => `version ${versionId} at ${percentage}%`)
+		.join(" and ");
```
mrbbot, line 204: Could you use `Intl.ListFormat` here, for a comma separated list if more than 2 versions are selected?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,652 @@
...
+		args.config || (args.name && findWranglerToml(path.dirname(args.name)));
+	const config = readConfig(configPath, args);
+	const accountId = await requireAuth(config);
+
+	return accountId;
+}
+
+async function printLatestDeployment(accountId: string, workerName: string) {
+	const s = spinner();
+	s.start("Fetching latest deployment");
```
mrbbot, line 224: How do spinners work if `s.stop()` isn't called because an exception is thrown? Presumably the unhandled exception just aborts the process and stops the spinner, but should we be cleaning it up in a `finally`? Would probably be a good candidate for explicit-resource-management/`using` once that lands.

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,652 @@
...
+${trafficString} ${versionIdString}
+      Created:  ${version.created.toLocaleString()}
+          Tag:  ${version.tag ?? BLANK_INPUT}
+      Message:  ${version.message ?? BLANK_INPUT}`)
+		);
+	}
+
+	cli.newline();
+
+	VERSION_CACHE; // store in
```
mrbbot, line 250: This line doesn't appear to be doing anything?

mrbbot (commented):
Halfway through my review, but left comments so far.

packages/wrangler/src/versions/deploy.ts
RamIdeas: Yeah the server-side only supports 2 right now. But that limitation will be lifted later and all we'll need to do in wrangler is change this default value (or remove any usage of this arg entirely)

packages/wrangler/src/versions/deploy.ts
RamIdeas: https://github.com/cloudflare/workers-sdk/pull/5115/commits/b3cb614192f6980c7eeb8f60a4910898b777bade

packages/wrangler/src/versions/deploy.ts
RamIdeas: https://github.com/cloudflare/workers-sdk/pull/5115/commits/9490cfeceecf3970a638ac5553206e9216e9689f

packages/wrangler/src/versions/deploy.ts
RamIdeas: https://github.com/cloudflare/workers-sdk/pull/5115/commits/da5d88d6d4c6b821cab041bf72ce7fa2979e20aa

packages/wrangler/src/versions/deploy.ts
```diff
@@ -523,14 +523,19 @@ async function createDeployment(
...
 	const versionIds: string[] = [];
 	const percentages: OptionalPercentage[] = [];
 
-	for (const spec of args.versionSpecs ?? []) {
+	// args.versionSpecs weren't being parsed in tests, so fallback to args._ if missing
+	const versionSpecs =
+		args.versionSpecs ??
+		args._.filter((arg): arg is string => typeof arg === "string");
+
+	for (const spec of versionSpecs) {
```
RamIdeas, line 538: Had to fallback to `args._` because the way I tried unit-testing this function didn't seem to parse the positional args into `args.versionSpecs` – maybe someone knows a better way?

RamIdeas: cc @petebacondarwin

RamIdeas, line 538: Had to fallback to args._ because the way I tried unit-testing this function didn't seem to parse the positional args into args.versionSpecs – maybe someone knows a better way?

cc @petebacondarwin

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,657 @@
...
+					grayBar,
+				];
+			},
+		},
+	});
+
+	return result;
+}
+
+async function promptPercentages(
```
mrbbot, line 326: Some comments explaining what each of these parameters are could be helpful.

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,657 @@
...
+}
+function castAndCacheWorkerVersion(apiVersion: ApiVersion) {
+	const version: WorkerVersion = {
+		id: apiVersion.id,
+		created: new Date(apiVersion.metadata.created_on),
+		message: apiVersion.annotations["workers/message"],
+		tag: apiVersion.annotations["workers/tag"],
+	};
+
+	console.log(apiVersion, "\n\n\n");
```
mrbbot, line 484:
```suggestion
```

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,657 @@
...
+	const versionIds: string[] = [];
+	const percentages: OptionalPercentage[] = [];
+
+	// args.versionSpecs weren't being parsed in tests, so fallback to args._ if missing
+	const versionSpecs =
+		args.versionSpecs ??
+		args._.filter((arg): arg is string => typeof arg === "string");
+
+	for (const spec of versionSpecs) {
+		const [versionId, percentageString] = spec.split("@");
```
mrbbot, line 537: Should we check whether `split()` returns an array with `<= 2` elements?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,657 @@
...
+
+		const percentage =
+			percentageString === "" ? null : parseFloat(percentageString);
```
mrbbot, line 540: Won't `percentageString` be `undefined` if the user just specifies a version ID as the spec?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -0,0 +1,657 @@
...
+	versionIds: VersionId[],
+	optionalVersionTraffic: Map<VersionId, OptionalPercentage>
+): Map<VersionId, Percentage> {
+	const { subtotal, unspecifiedCount } = summariseVersionTraffic(
+		optionalVersionTraffic,
+		versionIds
+	);
+
+	const unspecifiedPercentageReplacement =
+		unspecifiedCount === 0 || subtotal > 100
```
mrbbot, line 600: Should we be throwing if `subtotal > 100`, or is this to account for floating point errors?

RamIdeas: This occurs during user input. If the subtotal is not 100 (+/- EPSILON) we take the user back through the prompts for the values (where the default values this time are what they entered previously)

packages/wrangler/src/versions/deploy.ts
RamIdeas: Ah yes good catch! I'll change to:
```suggestion
			percentageString === "" || percentageString == null ? null : parseFloat(percentageString);
```

irvinebroque:
If I create a new Worker, publish a new version, and then run this, in interactive mode I'm shown:

```
➜  gradual-test git:(main) ✗ npx https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/8195782752/npm-package-wrangler-5115 versions deploy --experimental-gradual-rollouts
 ⛅️ wrangler 0.0.0-fb147c36
---------------------------
╭ Deploy Worker Versions by splitting traffic between multiple versions
│ 
├ Your current deployment has 1 version(s):
│ 
│ (100%) 19c005af-5788-4991-a2a0-00a080f2b650
│       Created:  3/7/2024, 4:20:39 PM
│           Tag:  -
│       Message:  -
│ 
╰ Which version(s) do you want to deploy? Use SPACE to select/unselect version(s) and ENTER to submit.
  ○ 7d80087b-422a-41f5-b4bc-c6bc38deeeb9
​       Created:  3/7/2024, 4:24:35 PM
​           Tag:  -
​       Message:  -
```

...and then no matter which version I select, I get the following error:

```
├ Which version(s) do you want to deploy?
├ 0 Worker Version(s) selected
│

✘ [ERROR] You must select at least 1 version to deploy.
```

cc @tanushreesharma-cf

RamIdeas:
@irvinebroque that's odd – have you had similar problems with C3's interactive prompts?

I don't have any issues inside VSCode's terminal or in iTerm2:

https://github.com/cloudflare/workers-sdk/assets/5822355/a2189b2f-edfe-4370-a06e-0c55239237be

packages/cli/interactive.ts
```diff
@@ -86,8 +92,29 @@ type RenderProps =
 	| Omit<ConfirmPrompt, "prompt">
 	| Omit<SelectRefreshablePrompt, "prompt">;
 
-export const inputPrompt = async <T = string>(promptConfig: PromptConfig) => {
-	const renderers = getRenderers(promptConfig);
+function acceptDefault<T>(
```
mrbbot, line 95: Should `T extends Arg` to avoid `as Arg` casting below?

packages/cli/interactive.ts
```diff
@@ -86,8 +92,29 @@ type RenderProps =
...
+): T {
+	const error = promptConfig.validate?.(initialValue as Arg);
+	if (error) {
+		crash(error);
+	}
+
+	const lines = renderers.submit({ value: initialValue as Arg });
+	logRaw(lines.join("\n"));
+
+	return initialValue as T;
```
mrbbot, line 108: `initialValue` is already typed `T` so I assume this case isn't necessary?

packages/cli/interactive.ts
```diff
@@ -86,8 +92,29 @@ type RenderProps =
...
+}
+
+export const inputPrompt = async <T = string>(
```
mrbbot, line 111: Should this `T extends string` too? It looks like later you cast a `string` to `T`.

packages/wrangler/e2e/helpers/normalize.ts
```diff
@@ -167,3 +168,7 @@ function removeStandardPricingWarning(stdout: string): string {
 		""
 	);
 }
+
+function removeZeroWidthSpaces(stdout: string) {
+	return stdout.replaceAll(/\u200a|\u200b/g, " ");
```
mrbbot, line 173: Would it be worth replacing these with a more recognisable sequence to distinguish these in tests? Maybe `<ZWSP>`?

packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
```diff
@@ -0,0 +1,900 @@
...
+	describe("with wrangler.toml", () => {
+		beforeEach(writeWranglerToml);
+
+		test("no args", async () => {
+			const result = runWrangler(
+				"versions deploy --yes --experimental-gradual-rollouts"
+			);
+
+			await expect(result).rejects.toMatchInlineSnapshot(
+				`[Error: You must select at least 1 version to deploy.]`
```
mrbbot, line 123: Presumably this is only failing because the `--yes` flag is set? Without that, you'd get an interactive prompt?

packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
```diff
@@ -0,0 +1,900 @@
...
+			├ 1 Worker Version(s) selected
+			│
+			├     Worker Version 1:  00000000-0000-0000-0000-000000000000
+			│              Created:  1/4/2021, TIMESTAMP AM
+			│                  Tag:  -
+			│              Message:  -
+			│
+			├ What percentage of traffic should Worker Version 1 receive?
+			├ 100% of traffic
+			├
```
mrbbot, line 183: Should this be a `│` not a `├`?

mrbbot (approved):

packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
RamIdeas: Yes and if you didn't select anything from the interactive prompt – same flow.

packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
RamIdeas: I didn't want to mess too much with the CLI package because of existing usage – I'm not doing anything special with it here so it can and should be fixed there but can be done as a follow-up when thought through properly.

For now, this is a minor design glitch I think we're fine with going ahead with

packages/wrangler/e2e/helpers/normalize.ts
RamIdeas: It feels like an implementation detail. The zero-width space is for clack to not remove the indentation spacing. Doesn't need to be in the snapshots – since the absence of the zwsp will mean no indentation in the snapshot which will be caught and fail the test

packages/cli/interactive.ts
RamIdeas: Unfortunately T can be anything – for example, select prompts can be anything and will be whatever the type of "value" is (it's an underlying key so doesn't need to be stringifiable) and multiselect will be an array of anything

packages/cli/interactive.ts
RamIdeas: I think the reason was not all possible values of T are Arg...

RamIdeas:
Had to rebase because of conflict in main

packages/cli/interactive.ts
```diff
@@ -261,15 +313,16 @@ const getSelectRenderers = (
 			const active = i === cursor;
 			const isInListOfValues =
 				Array.isArray(value) && value.includes(optionValue);
-			const color = isInListOfValues || active ? blue : dim;
```
jculvey, line 264: @RamIdeas was this change intentional? This is causing select values to render differently than they have in previous versions:

new:
image

old:
image

We should change this back or make it overridable somehow

RamIdeas: Ah yeah my bad I made it match the design doc but I think this is where single select va multi select may differ. We should double check with Meaghan. If we want different styles for single vs multi select we can make it overridable in a base implementation and have higher level getSelectRenderers and getMultiSelectRenderers

RamIdeas: I also wanted to cc you on the acceptDefault option I added to inputPrompt

PR #5118, opened 2024-02-28
More strict handling of requirement specifiers in requirements.txt.

garrettgu10:
After some discussion, the runtime team has decided that we should prevent users from specifying version identifiers altogether within requirements.txt, rather than silently dropping them. This would prevent users from being deluded into thinking that they could specify versions of Python packages in Python workers, then being confused because their code mysteriously breaks. 

Fixes EW-8189.

- Tests
  - [ ] Included
  - [x] Not necessary because: We're moving fast on this experimental functionality and I will be testing the change manually with the prerelease build. The original code also didn't include tests. 
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: documented inline

**Note for PR author:**

### DIFF
--- a/.changeset/swift-peaches-battle.md
+++ b/.changeset/swift-peaches-battle.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Including version identifiers in Python requirements.txt will now throw an error
--- a/.changeset/swift-peaches-battle.md
+++ b/.changeset/swift-peaches-battle.md
@@ -2,4 +2,4 @@
 "wrangler": minor
 ---
 
-Including version identifiers in Python requirements.txt will now throw an error
+fix: Including version identifiers in Python requirements.txt will now throw an error
--- a/packages/wrangler/src/deployment-bundle/find-additional-modules.ts
+++ b/packages/wrangler/src/deployment-bundle/find-additional-modules.ts
@@ -74,9 +74,9 @@ export async function findAdditionalModules(
 			for (const requirement of pythonRequirements.split("\n")) {
 				if (requirement === "") continue;
 				if (!isValidPythonPackageName(requirement)) {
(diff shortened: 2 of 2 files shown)


.changeset/swift-peaches-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Including version identifiers in Python requirements.txt will now throw an error
```
mrbbot, line 5: Could you prefix this with either `feature:` or `fix:`?

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -61,20 +71,25 @@ export async function findAdditionalModules(
...
+					name: requirement,
+					content: "",
+					filePath: undefined,
+				});
 			}
-			// We don't care if a requirements.txt isn't found
 		} catch (e) {
+			if (e instanceof UserError) throw e;
+
+			// We don't care if a requirements.txt isn't found
```
mrbbot, line 92: It seems like this `catch` is for catching the `readFile()` file-not-found error? Could you update this code so only that `readFile()` is covered by the `try`/`catch`? Then there would be no need to rethrow the user error. Something like...
```ts
let pythonRequirements: string | undefined;
try {
  pythonRequirements = await readFile(...);
} catch {
  // We don't care if a requirements.txt isn't found
  logger.debug(
		"Python entrypoint detected, but no requirements.txt file found."
	);
}
if (pythonRequirements !== undefined) {
  for (const requirement of pythonRequirements.split("\n") {
    ...
  }
}
```

mrbbot (approved):

PR #5119, opened 2024-02-28
Add python support for remote dev

garrettgu10:
This PR makes sure that the bundle type is set correctly in remote dev mode when the main module is a Python module. It also ensures that the source URL annotation is written in Python style when the entry point is a Python module. 

- Tests
  - [ ] Included
  - [x] Not necessary because: one-line change on experimental feature
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: one-line change on experimental feature; we will also be writing docs for the feature in the near future. 

**Note for PR author:**

### DIFF
--- a/.changeset/kind-ducks-pay.md
+++ b/.changeset/kind-ducks-pay.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Python support for remote dev
--- a/.changeset/kind-ducks-pay.md
+++ b/.changeset/kind-ducks-pay.md
@@ -2,4 +2,4 @@
 "wrangler": minor
 ---
 
-Python support for remote dev
+feature: Python support for remote dev
--- a/packages/wrangler/src/deployment-bundle/source-url.ts
+++ b/packages/wrangler/src/deployment-bundle/source-url.ts
@@ -3,7 +3,7 @@ import { pathToFileURL } from "url";
 import type { CfModule } from "./worker";
 
 function withSourceURL(source: string, sourcePath: string, isPython = false) {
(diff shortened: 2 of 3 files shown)


.changeset/kind-ducks-pay.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Python support for remote dev
```
mrbbot, line 5: Could you prefix this with `feature:`?

packages/wrangler/src/deployment-bundle/source-url.ts
```diff
@@ -2,7 +2,10 @@ import fs from "node:fs";
 import { pathToFileURL } from "url";
 import type { CfModule } from "./worker";
 
-function withSourceURL(source: string, sourcePath: string) {
+function withSourceURL(source: string, sourcePath: string, isPython = false) {
+	if (isPython) {
+		return `${source}\n# sourceURL=${pathToFileURL(sourcePath)}`;
```
mrbbot, line 7: In JavaScript, these `sourceURL` comments are interpreted by V8 and are used to ensure correct stack traces. Do you know if they do anything in Python? I'm guessing not, and if that's the case, could we omit them entirely if this is a Python module?

garrettgu10: ah, I wasn't sure why they were there; I'll just remove them then :-)

mrbbot (approved):

PR #5123, opened 2024-02-29
chore(wrangler): Configuration keys clean-up

CarmenPopoviciu:
As part of an ongoing "configuration keys in wrangler" audit, this commit cleans up some of the configuration related comments, annotations, aso, in an attempt to keep things consistent and updated across the board.

- Tests
  - [ ] Included
  - [x] Not necessary because: it's mostly code comments changes
  - [ ] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [x] Not necessary because: cleaning up docs around configuration will be a separate effort we're tracking internally

**Note for PR author:**

### DIFF
--- a/packages/wrangler/src/config/environment.ts
+++ b/packages/wrangler/src/config/environment.ts
@@ -42,7 +42,7 @@ export type CloudchamberConfig = {
  */
 interface EnvironmentInheritable {
 	/**
-	 * The name of your worker. Alphanumeric + dashes only.
+	 * The name of your Worker. Alphanumeric + dashes only.
 	 *
 	 * @inheritable
 	 */
@@ -75,18 +75,21 @@ interface EnvironmentInheritable {
 	 *
 	 * More details at https://developers.cloudflare.com/workers/platform/compatibility-dates
 	 *
+	 * @default `[]`
 	 * @inheritable
 	 */
 	compatibility_flags: string[];
 
 	/**
 	 * The entrypoint/path to the JavaScript file that will be executed.
+	 *
+	 * @inheritable
 	 */
 	main: string | undefined;
 
 	/**
 	 * If true then Wrangler will traverse the file tree below `base_dir`;
-	 * Any files that match `rules` will be included in the deployed worker.
+	 * Any files that match `rules` will be included in the deployed Worker.
 	 * Defaults to true if `no_bundle` is true, otherwise false.
 	 *
 	 * @inheritable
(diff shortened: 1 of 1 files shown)


packages/wrangler/src/config/environment.ts
```diff
@@ -98,50 +101,55 @@ interface EnvironmentInheritable {
...
 	 * @inheritable
 	 */
 	base_dir: string | undefined;
 
 	/**
 	 * Whether we use <name>.<subdomain>.workers.dev to
-	 * test and deploy your worker.
+	 * test and deploy your Worker.
 	 *
+	 * // Carmen according to our tests the default is undefined
```
penalosa, line 121: The default isn't quite undefined: it's a bit weird. It's documented https://developers.cloudflare.com/workers/configuration/routing/routes/#routes-with-workersdev, but essentially the default is `true` unless you have `route` or `routes` defined, in which case the default is `false`.

packages/wrangler/src/config/environment.ts
```diff
@@ -210,7 +218,7 @@ interface EnvironmentInheritable {
 	 * Refer to the [custom builds documentation](https://developers.cloudflare.com/workers/cli-wrangler/configuration#build)
 	 * for more details.
 	 *
-	 * @default {}
+	 * @default {watch_dir:"./src"}
```
penalosa, line 221: I don't think this is accurate? AFAIK the default for `build.watch_dir` is `undefined` (or the current working directory if `build.command` is defined)

packages/wrangler/src/config/environment.ts
```diff
@@ -423,13 +417,13 @@ interface EnvironmentNonInheritable {
 	 * NOTE: This field is not automatically inherited from the top level environment,
 	 * and so must be specified in every named environment.
 	 *
-	 * @default `{}`
+	 * @default `{consumers:[],producers:[]}`
```
CarmenPopoviciu, line 420: as per our tests, this seems to be the default

packages/wrangler/src/config/environment.ts
```diff
@@ -652,14 +660,50 @@ interface EnvironmentNonInheritable {
...
+	/**
+	 * Specifies namespace bindings that are bound to this Worker environment.
+	 *
+	 * NOTE: This field is not automatically inherited from the top level environment,
+	 * and so must be specified in every named environment.
+	 *
+	 * @default `[]`
+	 * @nonInheritable
+	 */
+	dispatch_namespaces: {
```
CarmenPopoviciu, line 699: can we pls confirm this is indeed nonInheritable? The code comment seems to point to that, but the key was listed in the inheritable list

packages/wrangler/src/config/environment.ts
CarmenPopoviciu: is [this test](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/__tests__/configuration.test.ts#L23) incorrect then?

penalosa: Oh sorry, yeah, looks like I was reading the code wrong—apologies!

packages/wrangler/src/config/environment.ts
mrbbot: Based on this line...
https://github.com/cloudflare/workers-sdk/blob/046930eb898db6d45a6b26751dede07793435d28/packages/wrangler/src/config/validation.ts#L1341
...it seems like this is indeed non inheritable.

mrbbot (approved):

packages/wrangler/src/config/environment.ts
petebacondarwin: In which case it is in the incorrect interface?

CarmenPopoviciu: indeed. will fix that!

petebacondarwin (approved):

PR #5128, opened 2024-02-29
pass legacy-env through versions upload

taylorlee:

- Tests

  - [x] Not necessary because: versions upload is experimental, doesn't have test suite yet. This brings it into conformance with standard deploy by not hardcoding is_legacy prop.
  - [x] Included

### DIFF
--- a/.changeset/hungry-cooks-brake.md
+++ b/.changeset/hungry-cooks-brake.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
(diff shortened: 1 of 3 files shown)


petebacondarwin (commented):
LGTM - but are there no tests for this command that we can augment?

petebacondarwin (approved):
It seems there are none, so it is probably not on you @taylorlee to add them now.

PR #5138, opened 2024-03-01
Fix Workers AI local mode fetcher not returning headers to client worker

G4brym:

Fixes Workers AI local mode fetcher not returning headers to client worker

- Tests
  - [ ] Included
  - [x] Not necessary because:
  - [ ] Included
  - [x] Not necessary because:
- Public documentation
  - [x] Not necessary because: this is a small fix

### DIFF
--- a/.changeset/yellow-phones-buy.md
+++ b/.changeset/yellow-phones-buy.md
@@ -2,4 +2,4 @@
 "wrangler": patch
 ---
 
-Fix Workers-AI local mode fetcher not returning headers to client worker
+fix: ensure Workers-AI local mode fetcher returns headers to client worker
--- a/.changeset/yellow-phones-buy.md
+++ b/.changeset/yellow-phones-buy.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Fix Workers-AI local mode fetcher not returning headers to client worker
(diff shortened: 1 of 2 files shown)


packages/wrangler/src/ai/fetcher.ts
```diff
@@ -16,5 +16,12 @@ export async function AIFetcher(request: Request) {
 		duplex: "half",
 	});
 
-	return new Response(res.body, { status: res.status });
+	const respHeaders: Record<string, string> = {};
+	for (const [name, value] of res.headers) {
+		respHeaders[name] = value;
+	}
+	respHeaders.headers.delete("Host");
+	respHeaders.headers.delete("Content-Length");
```
mrbbot, line 24: `respHeaders` is a `Record<string, string>` typed value, not a `Response`, so probably has no `headers` property. Probably not an issue here but in general, I'd recommend using the `Headers` class when mutating headers, as it correctly handles things like `Set-Cookie` which can have multiple values.
```suggestion	
	const respHeaders = new Headers(res.headers);
	respHeaders.delete("Host");
	respHeaders.delete("Content-Length");
```
Be sure to `import { Headers } from "miniflare"` too.

.changeset/yellow-phones-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Fix Workers-AI local mode fetcher not returning headers to client worker
```
mrbbot, line 5: All changesets should start with `fix:`, `feature:`, `chore:`, etc.
```suggestion
fix: ensure Workers-AI local mode fetcher returns headers to client worker
```

mrbbot (approved):
We don't currently test Workers AI in `workers-sdk`, as it requires real account credentials and we're concerned about hitting rate limits, so approving this without tests. 👍

PR #5143, opened 2024-03-01
wrangler: Try to URL decode hyperdrive password and database name before sending them to API

OilyLime:

Fixes #SQC-153

I'm building on top of the changes in https://github.com/cloudflare/workers-sdk/pull/5064

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] Not necessary because: This does not change the documented behavior of the hyperdrive command.


packages/wrangler/src/hyperdrive/create.ts
```diff
@@ -78,15 +88,19 @@ export async function handler(
 				host: url.hostname,
 				port: parseInt(url.port),
 				scheme: url.protocol.replace(":", ""),
-				database: url.pathname.replace("/", ""),
+				database: decodeURIComponent(url.pathname.replace("/", "")),
 				user: url.username,
```
mrbbot, line 92: Again, should we be URL decoding the username here too?

packages/wrangler/src/hyperdrive/create.ts
```diff
@@ -78,15 +88,19 @@ export async function handler(
...
 			},
-			caching: { disabled: args.cachingDisabled ?? false },
 		});
 		logger.log(
 			`✅ Created new Hyperdrive config\n`,
 			JSON.stringify(database, null, 2)
 		);
 	}
-}
+}
```
mrbbot, line 106:
```suggestion
}

```

packages/wrangler/src/hyperdrive/update.ts
```diff
@@ -39,52 +40,92 @@ export function options(yargs: CommonYargsArgv) {
...
 		.epilogue(hyperdriveBetaWarning);
 }
 
+const requiredOriginOptions = [
+	"origin-host",
+	"origin-port",
+	"database",
+	"origin-user",
+	"origin-password",
+];
```
mrbbot, line 70: If you as `as const` to the end here, I think you might be able to remove the `as keyof typeof options` below.
```suggestion
] as const;
```

packages/wrangler/src/hyperdrive/update.ts
```diff
@@ -39,52 +40,92 @@ export function options(yargs: CommonYargsArgv) {
...
+		if (args.originScheme !== undefined) {
+			database.origin.scheme = args.originScheme;
+		} else {
+			database.origin.scheme = "postgresql"; // setting default if not passed
+		}
 	}
-	if (args.cachingDisabled !== undefined) {
-		database.caching.disabled = args.cachingDisabled;
+
+	if (args.cachingDisabled || args.maxAge || args.swr) {
```
mrbbot, line 120: If the user wants to re-enable caching, could they do this with `--caching-disabled=false`? If so, this wouldn't set `database.caching`, as `args.cachingDisabled` would be `false`.

packages/wrangler/src/hyperdrive/update.ts
```diff
@@ -39,52 +40,92 @@ export function options(yargs: CommonYargsArgv) {
...
+			port: args.originPort,
+			database: args.database,
+			user: args.originUser,
+			password: args.originPassword,
+		};
+		if (args.originScheme !== undefined) {
+			database.origin.scheme = args.originScheme;
+		} else {
+			database.origin.scheme = "postgresql"; // setting default if not passed
+		}
```
mrbbot, line 117:
```suggestion
		database.origin = {
			scheme: args.originScheme ?? "postgresql",
			host: args.originHost,
			port: args.originPort,
			database: args.database,
			user: args.originUser,
			password: args.originPassword,
		};
```

.changeset/twenty-laws-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+"miniflare": minor
+"wrangler": minor
+---
+
+URL decode components of the Hyperdrive config connection string
```
mrbbot, line 6: Could you prefix this with `feature:` or `fix:`? Presumably `feature:` if this is a `minor` change?

PR #5151, opened 2024-03-04
[C3] chore: fix dependabot so that it used the correct package version

dario-piotrowicz:

Currently our dependabot workflow is flawed and uses outdated version of framework
CLIs as you can see from this PR's description: https://github.com/cloudflare/workers-sdk/pull/5149

This is due to the wrong checkout performed in the `e2e-only-dependabot-bumped-framework` step which uses the version of the package.json file from `main` instead of getting the updated one.

This PR fixes such behavior.

- Tests
  - [ ] Included
  - [x] Not necessary because: this is only a gh actions change
  - [ ] Included
  - [x] Not necessary because: this is only a gh actions change
- Public documentation
  - [x] Not necessary because: this is only a gh actions change

### DIFF
--- a/.github/workflows/c3-e2e-dependabot.yml
+++ b/.github/workflows/c3-e2e-dependabot.yml
@@ -10,11 +10,6 @@ on:
 env:
   node-version: 18.17.1
   bun-version: 1.0.3
-  # we need to fetch 3 commits because dependabot generates one, our changeset
-  # generation workflow generates another and we need the parent of both commits
-  # so that we can diff it (`git diff HEAD~n`) and see what package changed in
-  # the `package.json` file
-  COMMITS_TO_FETCH: 3
 
 jobs:
   get-dependabot-bumped-framework:
@@ -28,7 +23,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          fetch-depth: ${{ env.COMMITS_TO_FETCH }}
+          fetch-depth: 0
 
       - name: detect-bumped-framework
         id: detect
@@ -60,7 +55,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
-          fetch-depth: ${{ env.COMMITS_TO_FETCH }}
+          fetch-depth: 0
 
       - name: Install Dependencies
         uses: ./.github/actions/install-dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/c3-e2e-dependabot.yml
```diff
@@ -55,7 +55,8 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0
+          fetch-depth: 3
+          ref: ${{ github.head_ref }}
```
petebacondarwin, line 59: I think the correct fix is to change the trigger to `pull_request` rather than `pull_request_target`, which according to the docs runs the job in the context of the base of the PR rather than the HEAD.

petebacondarwin (changes requested):

.github/workflows/c3-e2e-dependabot.yml
dario-piotrowicz: sounds reasonable to me, thanks 🙂 

I've update the trigger, please have another look 🙏

.github/workflows/c3-e2e-dependabot.yml
```diff
@@ -55,7 +60,7 @@ jobs:
       - name: Checkout Repo
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0
```
dario-piotrowicz, line 58: Note: `0` would also work but it seems quite unnecessary in my opinion since it fetches _everything_ (https://github.com/actions/checkout)

Screenshot 2024-03-05 at 13 29 51

I'm happy to use 0 if we find it more clear/robust and don't care about the extra fetching, please let me know what you think 🙂

petebacondarwin: We've been going with 0 because it guarantees that all the tags are also there, and generally no one is complaining about performance. So I would err on that side. It is very confusing if something is missing from the cloned repo because the errors are a bit vague.

petebacondarwin (changes requested):

.github/workflows/c3-e2e-dependabot.yml
dario-piotrowicz: I've changed it to `0` 🙂👍

PR #5152, opened 2024-03-04
Add dropdown item for kv asset handler to issue template

lrapoport-cf:

Fixes N/A.
Adds dropdown item to issue template for kv asset handler.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: just a change to the issue template
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: just a change to the issue template
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: just a change to the issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-template.yaml
+++ b/.github/ISSUE_TEMPLATE/bug-template.yaml
@@ -27,6 +27,7 @@ body:
         - Workers Runtime
         - Wrangler core
         - Miniflare
+        - KV Asset Handler
         - Other
     validations:
       required: true


mrbbot (approved):

.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -27,6 +27,7 @@ body:
         - Workers Runtime
         - Wrangler core
         - Miniflare
+        - KV Asset Handler
```
penalosa, line 30: Super nitpicky, but we should standardise on title case or sentence case (looking at Wrangler core above)

penalosa (approved):

.github/ISSUE_TEMPLATE/bug-template.yaml
lrapoport-cf: @penalosa yeah i thought about that -- my thinking was that kv asset handler is the actual proper name of a package so it should be all capitalized, but that wrangler core isn't really an officially recognized entity on its own -- maybe we just change it to say `Wrangler` or `Wrangler (core)`?

PR #5157, opened 2024-03-05
C3: Add `env.d.ts` to Nuxt template

jculvey:

Updates the Nuxt template by adding a `env.d.ts` file which updates the type difinition for `H3EventContext` to include the `cf` property from the request and an `env` type generated from the recently added `build-cf-types` script.

- Tests
  - [ ] Included
  - [x] Not necessary because: covered by existing e2e coverage
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/13236
  - [ ] Not necessary because:


packages/create-cloudflare/templates/nuxt/c3.ts
```diff
@@ -27,13 +28,44 @@ const generate = async (ctx: C3Context) => {
...
 
-const configure = async () => {
-	await installPackages(["nitro-cloudflare-dev"], {
+const configure = async (ctx: C3Context) => {
+	const packages = ["nitro-cloudflare-dev"];
+
+	// When using pnpm, explicitly add h3 package so the H3Event type declaration can be updated.
+	// Package managers other than pnpm will hoist the dependency, as will pnpm with `--shamefully-hoist`
+	if (pm === "pnpm") {
+		packages.push("h3");
```
dario-piotrowicz, line 37: instead of installing `h3` could we just make sure that it gets hoisted?

we could just create an `.npmrc` file with the following content:
```
public-hoist-pattern[]=*h3*
```

and that would do the trick, wouldn't that be a (slightly) cleaner solution?

dario-piotrowicz (approved):
Looks good to me 🙂 , with just one small suggestion

packages/create-cloudflare/templates/nuxt/c3.ts
jculvey: I considered the `.npmrc` approach, but I there's an issue: `pnpm install` happens as part of `nitro init`. There's no files on disk before we run nitro, and files don't get copied over from template directories until immediately after running `nitro`.

So if we want to use an `.npmrc`, we'd need to check if we're using pnpm, `rm -rf node_modules pnpm-lock.yaml`, then re-install after the `.npmrc` is in place. This isn't great perf wise.

I'm kind of curious what the `pnpm` philosophy would consider clean here. We have this issue with `pnpm` because by default, it has semistrict `node_modules`, and undeclared dependencies can't be referenced outside of `node_modules`. 

It seems like what they're trying to encourage people to do is actually all declare dependencies they use (such as `h3` in our case)? If so, declaring and installing the dependency seems like the correct thing to do. Do you think having this as a declared dependency is less clean than the hoisting approach?

dario-piotrowicz: > Do you think having this as a declared dependency is less clean than the hoisting approach?

I'm not really sure... I was just wondering if having it as a specific dependency could have negative effects, such as multiple `h3` versions getting installed or version conflicts, probably not... I am not really sure 🤷 

Anyways your explanation about about the cumbersomeness that `.npmrc` would introduce and `pnpm`'s philosophy make sense to me, so yeah if you prefer I think it makes sense to proceed with adding `h3` as a direct dependency 🙂👍

PR #5158, opened 2024-03-05
Export wrap from middleware-ensure-req-body-drained

dpeek:
This one liner fixes a warning when using `unstable_dev` with `experimental.testMode: false` (needed because I want to customise the dev server and still have watch mode).

```
▲ [WARNING] Import "wrap" will always be undefined because there is no matching export in "../../node_modules/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts" [import-is-undefined]

    .wrangler/tmp/bundle-e0BN78/middleware-insertion-facade.js:4:42:
      4 │ ...vWrappers = [__MIDDLEWARE_0__.wrap,__MIDDLEWARE_1__.wrap].filter...
        ╵                                  ~~~~
```


packages/wrangler/templates/middleware/middleware-ensure-req-body-drained.ts
```diff
@@ -16,3 +16,4 @@ const drainBody: Middleware = async (request, env, _ctx, middlewareCtx) => {
 };
 
 export default drainBody;
+export const wrap = undefined;
```
mrbbot, line 19:
```suggestion
export const wrap = undefined;

```

mrbbot (approved):

PR #5159, opened 2024-03-05
ci: disable file parallelism for fixture tests

mrbbot:

Some of the `get-{bindings,platform}-proxy` tests have been flaking in CI recently. I think this is because they're trying to connect to the dev registry, but the dev registry was started by an earlier test that just finished. This leads to socket errors like `other side closed`. The proper solution to this problem is to start a dev registry for each test (file). We could do this by assigning a different dev registry port for each test (file/project). In the meantime, this change disables parallelism for the fixture tests, ensuring if a dev registry is started, it is definitely shut down before another test begins.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: changing CI configuration
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: changing CI configuration
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: changing CI configuration

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
 		"fix": "pnpm run prettify && dotenv -- turbo check:lint -- --fix",
 		"prettify": "prettier . --write --ignore-unknown",
 		"test": "dotenv -- turbo test",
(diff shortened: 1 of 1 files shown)


petebacondarwin:
Blocked on https://github.com/cloudflare/workers-sdk/pull/5161

penalosa (approved):

PR #5166, opened 2024-03-05
feat(wrangler): Add config validation for Pages

CarmenPopoviciu:

Wrangler proper has a mechanism in place through which it validates a `wrangler.toml` file for Workers projects. As part of adding `wrangler.toml` support for Pages, we need to put a similar mechanism in place, to validate a configuration file against Pages specific requirements.

This PR implements the Pages specific validation logic

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not needed at this stage of the work

### DIFF
--- a/packages/wrangler/src/__tests__/config-validation-pages.test.ts
+++ b/packages/wrangler/src/__tests__/config-validation-pages.test.ts
@@ -1,0 +1,224 @@
+import { defaultWranglerConfig } from "../config/config";
+import { validatePagesConfig } from "../config/validation-pages";
+import type { Config } from "../config";
+
+describe("validatePagesConfig()", () => {
+	describe("`main` field validation", () => {
+		it("should error if configuration contains both `pages_build_output_dir` and `main` config fields", () => {
+			const config = generateConfigurationWithDefaults();
+			config.pages_build_output_dir = "./public";
+			config.main = "./src/index.js";
+
+			const diagnostics = validatePagesConfig(config, []);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeTruthy();
+			expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Running configuration file validation for Pages:
+			  - Configuration file cannot contain both both \\"main\\" and \\"pages_build_output_dir\\" configuration keys.
+			    			Please use \\"main\\" if you are deploying a Worker, or \\"pages_build_output_dir\\" if you are deploying a Pages project.
+			  - Configuration file for Pages projects does not support \\"main\\""
+		`);
+		});
+	});
+
+	describe("named environments validation", () => {
+		it("should pass if no named environments are defined", () => {
+			const config = generateConfigurationWithDefaults();
+			config.pages_build_output_dir = "./public";
+
+			const diagnostics = validatePagesConfig(config, []);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeFalsy();
+		});
+
+		it("should pass for environments named 'preview' and/or 'production'", () => {
+			const config = generateConfigurationWithDefaults();
+			config.pages_build_output_dir = "./public";
+
+			let diagnostics = validatePagesConfig(config, ["preview"]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeFalsy();
+
+			diagnostics = validatePagesConfig(config, ["production"]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeFalsy();
+
+			diagnostics = validatePagesConfig(config, ["preview", "production"]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeFalsy();
+		});
+
+		it("should error for any other named environments", () => {
+			const config = generateConfigurationWithDefaults();
+			config.pages_build_output_dir = "./assets";
+
+			let diagnostics = validatePagesConfig(config, [
+				"unsupported-env-name-1",
+				"unsupported-env-name-2",
+			]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeTruthy();
+			expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Running configuration file validation for Pages:
+			  - Configuration file contains environment names that are not supported by Pages projects:
+			    			unsupported-env-name-1,unsupported-env-name-2.
+			    			The supported named-environments for Pages are \\"preview\\" and \\"production\\"."
+		`);
+
+			diagnostics = validatePagesConfig(config, [
+				"production",
+				"unsupported-env-name",
+			]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeTruthy();
+			expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Running configuration file validation for Pages:
+			  - Configuration file contains environment names that are not supported by Pages projects:
+			    			unsupported-env-name.
+			    			The supported named-environments for Pages are \\"preview\\" and \\"production\\"."
+		`);
+		});
+	});
+
+	describe("unsupported fields validation", () => {
+		it("should pass if configuration contains only Pages-supported configuration fields", () => {
+			let config = generateConfigurationWithDefaults();
+			config.pages_build_output_dir = "./dist";
+
+			let diagnostics = validatePagesConfig(config, ["preview"]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeFalsy();
+
+			config = {
+				...config,
+				...{
+					name: "test-pages-project",
+					compatibility_date: "2024-01-01",
+					compatibility_flags: ["FLAG1", "FLAG2"],
+					send_metrics: true,
+					limits: { cpu_ms: 100 },
+					placement: { mode: "smart" },
+					vars: { FOO: "foo" },
+					durable_objects: {
+						bindings: [
+							{ name: "TEST_DO_BINDING", class_name: "TEST_DO_CLASS" },
+						],
+					},
+					kv_namespaces: [{ binding: "TEST_KV_BINDING", id: "1" }],
+					queues: {
+						producers: [{ binding: "TEST_QUEUE_BINDING", queue: "test-queue" }],
+					},
+					r2_buckets: [
+						{ binding: "TEST_R2_BINDING", bucket_name: "test-bucket" },
+					],
+					d1_databases: [
+						{
+							binding: "TEST_D1_BINDING",
+							database_id: "111",
+							database_name: "test-db",
+						},
+					],
+					vectorize: [
+						{ binding: "VECTORIZE_TEST_BINDING", index_name: "test-index" },
+					],
+					hyperdrive: [{ binding: "HYPERDRIVE_TEST_BINDING", id: "222" }],
+					services: [
+						{ binding: "TEST_SERVICE_BINDING", service: "test-worker" },
+					],
+					analytics_engine_datasets: [
+						{ binding: "TEST_AED_BINDING", dataset: "test-dataset" },
+					],
+					ai: { binding: "TEST_AI_BINDING" },
+					dev: {
+						ip: "127.0.0.0",
+						port: 1234,
+						inspector_port: 5678,
+						local_protocol: "https",
+						upstream_protocol: "https",
+						host: "test-host",
+					},
+				},
+			};
+
+			diagnostics = validatePagesConfig(config, ["production"]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeFalsy();
+		});
+
+		it("should fail if configuration contains any fields that are not supported by Pages projects", () => {
+			const defaultConfig = generateConfigurationWithDefaults();
+			defaultConfig.pages_build_output_dir = "./public";
+
+			// test with top-level only config fields
+			let config: Config = {
+				...defaultConfig,
+				...{
+					wasm_modules: {
+						MODULE_1: "testModule.mjs",
+					},
+					text_blobs: {
+						BLOB_2: "readme.md",
+					},
+				},
+			};
+			let diagnostics = validatePagesConfig(config, ["preview"]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeTruthy();
+			expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Running configuration file validation for Pages:
+			  - Configuration file for Pages projects does not support \\"wasm_modules\\"
+			  - Configuration file for Pages projects does not support \\"text_blobs\\""
+		`);
+
+			// test with inheritable environment config fields
+			config = {
+				...defaultConfig,
+				...{
+					triggers: { crons: ["cron1", "cron2"] },
+					usage_model: "bundled",
+					build: {
+						command: "npm run build",
+					},
+					node_compat: true,
+				},
+			};
+			diagnostics = validatePagesConfig(config, ["production"]);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeTruthy();
+			expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Running configuration file validation for Pages:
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/config/config.ts
```diff
@@ -183,6 +185,18 @@ export interface ConfigFields<Dev extends RawDevConfig> {
 	keep_vars?: boolean;
 }
 
+// Pages-specific configuration fields
```
CarmenPopoviciu, line 188: keeping this separate so that when things converge it'll be easy to remove these code paths :)

packages/wrangler/src/config/validation-pages.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+
+import { Diagnostics } from "./diagnostics";
+import { hasProperty } from "./validation-helpers";
+import type {
+	Config,
+	RawConfig,
+} from "./config";
+
+const supportedPagesConfigFields = [
```
CarmenPopoviciu, line 17: I am really hoping there's a better way and I'm too dumb to see it right now :/.

My original hope was we could use a dedicated type for this
```
export type PagesSupportedConfig = PagesConfigFields &
	DevConfig &
	Pick<Environment,
		"name" |
		"compatibility_date" |
		"compatibility_flags" |
		"placement" |
		"vars" |
		"kv_namespaces" |
		"queues" |
		"r2_buckets" |
		"d1_databases" |
		"vectorize" |
		"services" |
		"analytics_engine_datasets" |
		"ai"
	>;
	```
and use that to extract the keys. But extracting those as an array of strings is not supported in TS, so that was that. I eventually resorted to the good ol `Array`, but all years if TS gives us a more elegant way to achieve this

File: packages/wrangler/src/config/validation-pages.ts [hunk 3]
```diff
@@ -0,0 +1,126 @@
...
+			Please use "main" if you are deploying a Worker, or "pages_build_output_dir" if you are deploying a Pages project.`
+		);
+	}
+}
+
+/**
+ * Validate that no named-environments other than "preview" and "production"
+ * were declared in the configuration file for Pages
+ */
+function validatePagesEnvironmentNames(
```
CarmenPopoviciu, line 73: Pete suggested this validation should be performed inside `normalizeAndValidateConfig()`, which would remove the need to pass in `rawConfig` here. I tend to agree, so will address that

packages/wrangler/src/config/validation.ts
```diff
@@ -194,6 +202,10 @@ export function normalizeAndValidateConfig(
 	// Process the top-level default environment configuration.
 	const config: Config = {
 		configPath,
+		pages_build_output_dir: normalizeAndValidatePagesBuildOutputDir(
```
CarmenPopoviciu, line 205: I tried to keep all Pages things as contained as possible, with the the idea of the converged future in mind. However, this bit IMO, needs to happen here

packages/wrangler/src/config/validation-pages.ts
penalosa: I'm not sure about that—this feels like the right place for it given it's doing Pages specific validation?

packages/wrangler/src/config/index.ts
```diff
@@ -32,17 +33,31 @@ export function readConfig<CommandArgs>(
...
+		/**
+		 * Check if configuration file belongs to a Pages project.
+		 *
+		 * The `pages_build_output_dir` config key is used to determine if the
+		 * configuration file belongs to a Workers or Pages project. This key
+		 * should always be set for Pages but never for Workers. Furthermore,
+		 * Pages projects currently have support only for `wrangler.toml`,
+		 * which is why we only perform the check in this code path.
+		 */
+		isPagesConfig = rawConfig.pages_build_output_dir?.trim() !== "";
```
penalosa, line 54: Is there a reason we can't just check for `rawConfig.pages_build_output_dir !== undefined`

CarmenPopoviciu: I went and thought abt this some more and I _think_ there actually isn't. 

It seems that both `pages dev` and `pages deploy` are fine if we pass in `" "` as the directory arg, which is the equivalent of `pages_build_output_dir` being an empty string in this validation logic. What in effect this means, is that the build output directory will be resolved to the (most prolly root) directory that contains the `wrangler.toml` file. Since the build output directory contains static assets + the built functions + all other metadata files, like `_routes.json`, _headers`, etc, this'll mean that all these files + `wrangler.toml` + potentially the `functions` directory, will all live under this (root) dir (unless I have an error in judgement here). This would be a pretty, IMHO, funky setup, given that on `pages deploy` we deploy the assets inside the build output dir, which in this particular case would mean everything I mentioned earlier would get deployed?

If we were to decide against supporting a use case such as 👆, then we should perform that check against an empty string. However, seeing as we currently have nothing in place to prevent that, we can let it slide

petebacondarwin: Is the reason this is in here because we are explicitly not supporting wrangler.json for Pages projects?

packages/wrangler/src/config/validation-pages.ts
petebacondarwin: If it is Pages specific I agree with you @penalosa.

But doing it this way has a different downside, which is that this code needs to understand how RawConfig works and parse out the `env` property from there.

My suggestion was that it would be nice if these functions only cared about the processed `Config`. As such we could augment `normalizeAndValidateConfig()` with an extra parameter called `allowedEnvironments` (or something like that) which is not Pages specific per se but means that we don't need to pass the `rawConfig` into this function.

Moreover, this could feature of `normalizeAndValidateConfig()` could be extended to handle the environment fallback logic that we need for Pages - e.g. something like:

```
const allowedEnvs = [
 { name: "production", fallbackToTopLevel: true },
 { name: "preview", fallbackToTopLevel: true },
]
normalizeAndValidateConfig(rawConfig, configPath, args, allowedEnvs);
```

That kind of thing... open to other ideas but it seems awkward for us to have to have special logic that messes around with `RawConfig.env` outside of that function.

packages/wrangler/src/config/index.ts
CarmenPopoviciu: that's correct yes. I mention that in the comment as well

packages/wrangler/src/config/validation-pages.ts
```diff
@@ -0,0 +1,126 @@
...
+
+	if (unsupportedFields.size > 0) {
+		const fields = Array.from(unsupportedFields.keys()).map((field) => `"${field}"`);
+
+		fields.forEach((field) => {
+			if(field === "queues" && !!config.queues?.consumers?.length) {
+				diagnostics.errors.push(
+					`Pages configuration file does not support "queues.consumers"`
+				);
+			} else {
```
petebacondarwin, line 119: Why not just do this in the for loop above to avoid having to check for `field == "queues"` twice?

packages/wrangler/src/config/index.ts
petebacondarwin: Doh! I didn't read that comment fully.
Should we error if there is a wrangler.json with this field in it?

packages/wrangler/src/config/validation-pages.ts
CarmenPopoviciu: because that would break the logical grouping of the keys in `supportedPagesConfigFields` when errors would be printed to stdout, assuming multiple key-related errs. 

So for example if we had errors for each of `compatibility_flags`, `placement`, `kv_namespaces`, `queues` and `d1_databases`, the `queues` related error would always be printed first, which felt off to me. While not a big deal, and I understand not everybody is as OCD as I am 😅 , I figured the user experience would be much nicer, if we can maintain some logical grouping of the errors, based on the config grouping we have in our docs.

packages/wrangler/src/config/index.ts
CarmenPopoviciu: oh, that's a very good callout! we totes should! I'll fix that!

packages/wrangler/src/config/validation-pages.ts
petebacondarwin: Fair enough. Should have guessed you had thought about it

CarmenPopoviciu: totes valid callout!

CarmenPopoviciu: I'll add a comment to clarify the "why"

CarmenPopoviciu: ✅

packages/wrangler/src/config/config.ts
```diff
@@ -183,6 +185,18 @@ export interface ConfigFields<Dev extends RawDevConfig> {
...
+export interface PagesConfigFields {
+	/**
+	 * The directory of static assets to serve.
+	 *
+	 * The presence of this field in `wrangler.toml` indicates a Pages project,
+	 * and will prompt the handling of the configuration file according to the
+	 * Pages-specific validation rules.
+	 */
+	pages_build_output_dir: string | undefined;
```
petebacondarwin, line 197:
```suggestion
	pages_build_output_dir?: string;
```

to avoid people having to specify it when creating an object of this type in programmatic scenarios.

packages/wrangler/src/config/validation.ts
```diff
@@ -384,6 +396,26 @@ function normalizeAndValidateBaseDirField(
...
+	configPath: string | undefined,
+	rawPagesDir: string | undefined
+): string | undefined {
+	const configDir = path.dirname(configPath ?? "wrangler.toml");
+	if (rawPagesDir !== undefined) {
+		if (typeof rawPagesDir === "string") {
+			const directory = path.resolve(configDir);
+			return path.resolve(directory, rawPagesDir);
+		} else {
+			return rawPagesDir;
```
petebacondarwin, line 412: When could this branch occur?

petebacondarwin: Ah I see you copied this from my code for `main` 😄 
I think that code is also overzealous here.

packages/wrangler/src/config/validation-pages.ts
CarmenPopoviciu: removed the dependency on `RawConfig` for now

packages/wrangler/src/__tests__/config-validation-pages.test.ts
```diff
@@ -0,0 +1,219 @@
...
+			const config = generateConfigurationWithDefaults();
+			config.pages_build_output_dir = "./public";
+			config.main = "./src/index.js";
+
+			const diagnostics = validatePagesConfig(config, []);
+			expect(diagnostics.hasWarnings()).toBeFalsy();
+			expect(diagnostics.hasErrors()).toBeTruthy();
+			expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Running configuration file validation for Pages:
+			  - Configuration file cannot contain both both \\"main\\" and \\"pages_build_output_dir\\" configuration keys.
```
CarmenPopoviciu, line 17: are we OK that the presence of `main` will trigger these two errors?

packages/wrangler/src/config/validation-pages.ts
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	for (const field of unsupportedFields) {
+		// check for unsupported fields with default values and exclude if found.
+		// These were most likely set as part of `normalizeAndValidateConfig()`
+		// processing, and not via the config file.
+		if (
+			config[field] === undefined ||
+			JSON.stringify(config[field]) ===
+				JSON.stringify(defaultWranglerConfig[field])
```
CarmenPopoviciu, line 119: if there is a better way to test for defaults I'm all 👀

packages/wrangler/src/__tests__/config-validation-pages.test.ts
```diff
@@ -0,0 +1,219 @@
...
+			    			Please use \\"main\\" if you are deploying a Worker, or \\"pages_build_output_dir\\" if you are deploying a Pages project.
+			  - Configuration file for Pages projects does not support \\"main\\""
+		`);
+		});
+	});
+
+	describe("named environments validation", () => {
+		it("should pass if no named environments are defined", () => {
```
CarmenPopoviciu, line 25: is this actually true? should indeed validation pass if no envs were defined?

packages/wrangler/src/config/validation-pages.ts
CarmenPopoviciu: actually...I want to come back this. so as I am looking at implementing env inheritance for Pages, I better understand now where you're coming from @petebacondarwin. told you my brain works in baby avocado increments 😅 🥑 

so assuming we would change the `normalizeAndValidateConfig()` fn signature to accept that extra `allowedEnvs` param, are you saying that the env name validation for Pages should take place inside that fn as well? If we don't, we basically risk creating that normalized config for an invalid Pages env name, only to then fail in `validatePagesConfig()`, which feels wasteful to me (tho IMHO we are already potentially doing wasteful config processing as we validate for Pages *after* we normalize, so maybe not a biggie). However, if we do, we leak Pages-specific validation things into `normalizeAndValidateConfig`, which is something I was hoping to prevent.

It seems to me, that either way we go abt it, there's no one way forward that feels 100% right and clean, at least to me not sure how y'all feel abt it, and there's some compromise that needs to happen between:
- keeping all Pages-specific validation code in one place
- avoid polluting `normalizeAndValidateConfig` with Pages things
- validating Pages against a normalized config

penalosa: I think my thinking around this was that it's unavoidable that we make some changes to `normalizeAndValidateConfig()`, but by splitting out the validation into Pages specific and workers specific we can minimise the changes to `normalizeAndValidateConfig()` to just the one line change outlined in the spec

CarmenPopoviciu: are you referring to the one liner env inheritance change here?

packages/wrangler/src/config/validation-pages.ts
petebacondarwin, line 119: What about the case where the user has defined a value for a field but it happens to match the default?

penalosa: Carmen and I discussed this, and while it's not perfect, this behaviour should be fine. The defaults for the fields we care about don't do anything significant (they're undefined or empty arrays), and so wouldn't have any application-facing behaviour semantics.

packages/wrangler/src/config/validation-pages.ts
penalosa: @CarmenPopoviciu yes

CarmenPopoviciu: will leave this conversation unresolved for posterity, but AFAICT, we seem to be OK with the current implementation, so leaving as is. Appreciate the very constructive conversation around this <3

packages/wrangler/src/config/validation-pages.ts
CarmenPopoviciu: what Samuel said :)

CarmenPopoviciu:
@penalosa @petebacondarwin all your comments are ✅ ed. Pls let me know if there is anything else you'd like me address <3. Thank you so much for your thorough reviews!

penalosa (approved):
LGTM other than durable objects 🎉

packages/wrangler/src/config/validation-pages.ts
```diff
@@ -0,0 +1,140 @@
...
+	"queues", // `producers` ONLY
+	"r2_buckets",
+	"d1_databases",
+	"vectorize",
+	"hyperdrive",
+	"services",
+	"analytics_engine_datasets",
+	"ai",
+	"dev",
+] as const;
```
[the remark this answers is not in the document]
CarmenPopoviciu: OH! nice catch! I was 100% convinced we won't support DOs, which is why they were never in here. My audit doc is also wrong, so will update as well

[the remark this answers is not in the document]
CarmenPopoviciu: fixed!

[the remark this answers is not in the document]
CarmenPopoviciu: I just realized that if we support DOs, we need to support `migrations` as well. Will need to add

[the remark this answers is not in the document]
penalosa: No, because we'll only be supporting DOs defined in a different worker (AFAIK this is the current Pages status quo?), and I'm pretty sure you can't define cross worker DO migrations.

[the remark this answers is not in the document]
CarmenPopoviciu: we're gonna run into some warnings then like: https://github.com/cloudflare/workers-sdk/pull/5234/files#diff-91e21fab7c621dcd160ff3a7731adefac5ef37bd372fce4b7c51f5f4393c313cR38-R46. Those will be very noisy AFAICT. Are we OK with that?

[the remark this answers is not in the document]
CarmenPopoviciu: great, then no other necessary changes need wrt DOs

PR #5171, opened 2024-03-05
Update Workers hello-world template to contain D1 comments

joshthoward:

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Comments added
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because:


admah (approved):
These should definitely go in these templates. Thanks for adding 🚀 

@joshthoward it does look like the C3 tests are failing though. Could you check those?

joshthoward:
@petebacondarwin Can you let me know if we can merge this? Not sure what test failures are expected.

packages/create-cloudflare/templates/hello-world/js/wrangler.toml
```diff
@@ -96,3 +96,10 @@ compatibility_flags = ["nodejs_compat"]
 # [[vectorize]]
 # binding = "MY_INDEX"
 # index_name = "my-index"
+
```
jculvey, line 99: There's already an entry for D1 on line 27. If we want to update the comment around it that's fine, but this file is copy pasted between all workers templates, and it would be nice to do it for all of them uniformally if we want to make this change

packages/create-cloudflare/templates/hello-world/ts/src/index.ts
```diff
@@ -23,6 +23,9 @@ export interface Env {
 	//
 	// Example binding to a Queue. Learn more at https://developers.cloudflare.com/queues/javascript-apis/
 	// MY_QUEUE: Queue;
+	//
```
jculvey, line 26: This addition seems fine since there's other examples of it here, but now that I'm looking at it I think it would be better if we convince users to run `wrangler types` to generate this instead of updating the type manually.

packages/create-cloudflare/templates/hello-world/js/wrangler.toml
joshthoward: Oh cool. I didn't realize that this was included in the long string of comments.

jculvey: I was hiding it from you 😄

PR #5172, opened 2024-03-05
Allow marking externals in Pages Functions

GregBrimble:

This allows Pages Plugins authors to mark dependencies as external when bundling their plugin. This is useful if you want to declare a module as a peer dependency.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/13397
  - [ ] Not necessary because

### DIFF
--- a/packages/wrangler/src/pages/build.ts
+++ b/packages/wrangler/src/pages/build.ts
@@ -123,6 +123,11 @@ export function Options(yargs: CommonYargsArgv) {
 				deprecated: true,
 				hidden: true,
 			},
+			external: {
+				describe: "A list of module imports to exclude from bundling",
+				type: "string",
+				array: true,
+			},
 		});
 }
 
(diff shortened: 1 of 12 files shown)


packages/wrangler/src/pages/build.ts
```diff
@@ -105,6 +105,12 @@ export function Options(yargs: CommonYargsArgv) {
 				deprecated: true,
 				hidden: true,
 			},
+			external: {
+				describe: "A list of imports to exclude from bundling",
+				type: "string",
+				array: true,
+				hidden: true,
```
GregBrimble, line 112: Should we keep this hidden?

jahands: Seems reasonable to expose it, unless we think we'll take it away.

jahands (approved):
LGTM

PR #5178, opened 2024-03-06
remove create-cloudflare noisy test logs (by mocking `spinner`)

dario-piotrowicz:

Removes noisy logs from create-cloudflare tests such as these: https://github.com/cloudflare/workers-sdk/actions/runs/8164300583/job/22332302711?pr=5173#step:8:33

- Tests
  - [ ] TODO (before merge)
  - [x] Included (this is updating tests)
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: not needed for tests update
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not needed for tests update

### DIFF
--- a/packages/create-cloudflare/src/__tests__/common.test.ts
+++ b/packages/create-cloudflare/src/__tests__/common.test.ts
@@ -1,3 +1,4 @@
+import { spinner } from "@cloudflare/cli/interactive";
 import * as command from "helpers/command";
 import { SemVer } from "semver";
 import { getGlobalDispatcher, MockAgent, setGlobalDispatcher } from "undici";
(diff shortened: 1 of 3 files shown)


packages/create-cloudflare/src/__tests__/common.test.ts
```diff
@@ -12,11 +13,21 @@ import {
...
 }
 
+beforeEach(() => {
+	// we mock `spinner` to remove noisy logs from the test runs
+	vi.mocked(spinner).mockImplementation(() => ({
+		start() {},
+		update() {},
+		stop() {},
+	}));
+});
```
petebacondarwin, line 29: This works but I note that in the Wrangler tests we mock out `logUpdate` instead:

https://github.com/cloudflare/workers-sdk/blob/dae30015c646502819d79bf8b8ae032c4aa0669d/packages/wrangler/src/__tests__/jest.setup.ts#L53-L59

petebacondarwin (approved):

packages/create-cloudflare/src/__tests__/common.test.ts
dario-piotrowicz: yes, I tried doing that too, but `log-update` is a dependency of the [cli package](https://github.com/cloudflare/workers-sdk/tree/d70f44a1716f0a4a87a7a0732bb61079da036dcc/packages/cli) not `create-cloudflare` itself, so vitest was complaining that it could not find the dependency, I didn't try but I imagine I could add that as a `devDependency` of `create-cloudflare`, but it felt simpler, cleaner and more correct to mock the `cli` package instead.

I am totally fine revisiting this and mock `log-update` instead if that's preferred 🙂

PR #5184, opened 2024-03-06
fix: d1 migrations create to use the highest migration number plus one

nora-soderlund:
Previously, if you wanted to remove a migration (e.g. number 5 out of 10), you had to:
- Comment everything in the migration file and add `SELECT ...` just to have a query in it; or
- Delete it and lower the number of all preceding migrations.

Previously:
```cmd
// 0001_my-cool-migration.sql
// 0003_my-cool-migration.sql

wrangler d1 migrations create my-cool-database my-cool-migration // 0003_my-cool-migration.sql
```

After:
```cmd
// 0001_my-cool-migration.sql
// 0003_my-cool-migration.sql

wrangler d1 migrations create my-cool-database my-cool-migration // 0004_my-cool-migration.sql
```

This PR changes the logic of `getNextMigrationNumber` to actually get the _highest_ migration number plus one, rather than the first missing migration number plus one, which can be an existing migration number. 

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [X] Not necessary because: no existing test cases for the migration helpers
  - [ ] TODO (before merge)
  - [X] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [X] Not necessary because: nothing is explicitly stated about the next migration number in the docs

### DIFF
--- a/packages/wrangler/src/d1/migrations/helpers.ts
+++ b/packages/wrangler/src/d1/migrations/helpers.ts
@@ -131,7 +131,7 @@ export function getNextMigrationNumber(migrationsPath: string): number {
 	const migrationNumbers = getMigrationNames(migrationsPath).map((migration) =>
 		parseInt(migration.split("_")[0])
 	);
-	const highestMigrationNumber = Math.max(...migrationNumbers);
+	const highestMigrationNumber = Math.max(...migrationNumbers, 0);
 
 	return highestMigrationNumber + 1;
 }
--- a/packages/wrangler/src/d1/migrations/helpers.ts
+++ b/packages/wrangler/src/d1/migrations/helpers.ts
@@ -124,16 +124,12 @@ function getMigrationNames(migrationsPath: string): Array<string> {
 	return migrations;
 }
 
+/**
+ * Returns the highest current migration number plus one, ignoring any missing numbers.
+ */
 export function getNextMigrationNumber(migrationsPath: string): number {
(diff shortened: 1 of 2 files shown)


packages/wrangler/src/d1/migrations/helpers.ts
```diff
@@ -124,16 +124,14 @@ function getMigrationNames(migrationsPath: string): Array<string> {
...
-		const migrationNumber = parseInt(migration.split("_")[0]);
-
-		if (migrationNumber > highestMigrationNumber) {
-			highestMigrationNumber = migrationNumber;
-		}
-	}
+	const migrationNumbers = getMigrationNames(migrationsPath).map((migration) =>
+		parseInt(migration.split("_")[0])
+	);
+	const highestMigrationNumber = Math.max(...migrationNumbers, 0);
```
nora-soderlund, line 134: This explicitly makes the first number 0 + 1, which is a regression to how it works currently where the first number is 0.

If regression is unwanted, putting it to `-1` will make it -1 + 1 = 0.
```suggestion
	const highestMigrationNumber = Math.max(...migrationNumbers, -1);
```

rozenmd (approved):

rozenmd:
Just holding off merging this until https://github.com/cloudflare/workers-sdk/pull/4930 is in

Issue #3263, opened 2023-05-17
🐛 BUG: C3's Svelte option is really SvelteKit - confusing?


IgorMinar:
### Which Cloudflare product(s) does this pertain to?

C3

### What version of `Wrangler` are you using?

2.0.2

### What operating system are you using?

all

### Describe the Bug

I'm filing this mostly for posterity and record keeping.

In https://twitter.com/mr_wormhole/status/1658940824468480003 Talha pointed out that we are missing SvelteKit support.

Initially I thought this was a bug, but it turns out that what we call "Svelte" translates to `create-svelte` npm package which scaffolds a SvelteKit app.

My read on this is that we did the right thing: our guiding principle is that we enable full fidelity of UX/DX control for the framework authors. That includes branding. In this situation that means that we should honor that the Svelte/SvelteKit team decided to call their DX entry-point to scaffolding applications `create-svelte` which represents Svelte and not SvelteKit.

If the SvelteKit team would like to be represented differently then we'll honor that. We just need an authoritative Svelte representative to comment here while this issue is open or file a new issue on this repo after this issue is resolved.

dario-piotrowicz:
I just wanted to add that this doesn't only apply to Svelte/SvelteKit but to other frameworks/meta-frameworks as well such as qwik/qwikCity and solid/solidStart

Regardless on whether this is right or potentially confusing and whether it should be changed or not, one thing that I think important here is to keep this consistent with our docs, those currently use the same pattern as C3 does: https://developers.cloudflare.com/pages/framework-guides/ (saying `Deploy a Svelte site` instead of `Deploy a SvelteKit site`)

benmccann:
> If the SvelteKit team would like to be represented differently then we'll honor that. We just need an authoritative Svelte representative to comment here while this issue is open or file a new issue on this repo after this issue is resolved.

[...]

PR #5188, opened 2024-03-06
C3: Update SolidStart template to work with create-solid@0.5.3

jculvey:

Updates the SolidStart template to be compatible with the latest version of `create-solid`

- Tests
  - [ ] TODO (before merge)
  - [ x ] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ x ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ x ] Not necessary because: No user facing changes

### DIFF
--- a/packages/create-cloudflare/templates/solid/c3.ts
+++ b/packages/create-cloudflare/templates/solid/c3.ts
@@ -63,7 +63,7 @@ const configure = async (ctx: C3Context) => {
 const config: TemplateConfig = {
 	configVersion: 1,
 	id: "solid",
-	displayName: "SolidStart",
+	displayName: "Solid",
 	platform: "pages",
 	generate,
 	configure,
--- a/packages/create-cloudflare/templates/solid/c3.ts
+++ b/packages/create-cloudflare/templates/solid/c3.ts
@@ -12,7 +12,8 @@ const { npm } = detectPackageManager();
 
 const generate = async (ctx: C3Context) => {
 	// Run the create-solid command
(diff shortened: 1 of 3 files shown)


petebacondarwin (approved):

packages/create-cloudflare/templates/solid/c3.ts
```diff
@@ -62,7 +63,7 @@ const configure = async (ctx: C3Context) => {
 const config: TemplateConfig = {
 	configVersion: 1,
 	id: "solid",
-	displayName: "Solid",
+	displayName: "SolidStart",
```
dario-piotrowicz, line 66: I personally like this change, but I'd like to point out:

- @IgorMinar seemed interested in this: https://github.com/cloudflare/workers-sdk/issues/3263 so I'd probably check what he thinks about the renaming

- if we rename `Solid` to `SolidStart` I feel we should give the same treatment to all other frameworks or would you like to have solid be the only one? is the rationale here that `create-solid` can create non-solid-start apps? (it is the only CLI that allows such a thing I think, but that's mostly obscured by the user in C3, no?)

- shouldn't we update our docs accordingly? (as I mentioned in #3263)

dario-piotrowicz (approved):
LGTM 🙂

packages/create-cloudflare/templates/solid/c3.ts
jculvey: Good point, I'll revert this change

PR #5200, opened 2024-03-07
Respect --no-bundle when deploying a _worker.js/ directory in a Pages project

### LINKED ISSUE #5199: 🐛 BUG: Support `--no-bundle` when deploying an `_worker.js/` directory

Pages

3.32.0 [wrangler]

_No response_

macOS

### Describe the Bug

https://github.com/cloudflare/workers-sdk/blob/96e4e1654ce4cf48712c5545ace1a90e08afaa2b/packages/wrangler/src/api/pages/deploy.tsx#L258-L293

We're only avoiding bundling if it's a file rather than a dir right now.

_No response_

_No response_

GregBrimble:

Fixes #5199

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)

### DIFF
--- a/packages/wrangler/src/__tests__/pages/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/pages/deploy.test.ts
@@ -1534,6 +1534,304 @@ describe("deployment create", () => {
 		expect(std.err).toMatchInlineSnapshot('""');
 	});
 
+	it("should error with --no-bundle and a single _worker.js file", async () => {
+		// set up the directory of static files to upload.
+		mkdirSync("public");
+		writeFileSync("public/README.md", "This is a readme");
+
+		// set up _worker.js
+		writeFileSync(
+			"public/_worker.js",
+			`
+			import { thing } from "some-module";
+
+      export default {
+        async fetch(request, env) {
+					console.log(thing);
+          const url = new URL(request.url);
+          return url.pathname.startsWith('/api/') ? new Response('Ok') : env.ASSETS.fetch(request);
+        }
+      };
+    `
+		);
+
+		mockGetUploadTokenRequest(
+			"<<funfetti-auth-jwt>>",
+			"some-account-id",
+			"foo"
+		);
+
+		msw.use(
+			rest.post("*/pages/assets/check-missing", async (req, res, ctx) => {
+				const body = (await req.json()) as {
+					hashes: string[];
+				};
+
+				expect(req.headers.get("Authorization")).toBe(
+					"Bearer <<funfetti-auth-jwt>>"
+				);
+				expect(body).toMatchObject({
+					hashes: ["13a03eaf24ae98378acd36ea00f77f2f"],
+				});
+
+				return res.once(
+					ctx.status(200),
+					ctx.json({
+						success: true,
+						errors: [],
+						messages: [],
+						result: body.hashes,
+					})
+				);
+			}),
+			rest.post("*/pages/assets/upload", async (req, res, ctx) => {
+				expect(req.headers.get("Authorization")).toBe(
+					"Bearer <<funfetti-auth-jwt>>"
+				);
+
+				expect(await req.json()).toMatchObject([
+					{
+						key: "13a03eaf24ae98378acd36ea00f77f2f",
+						value: Buffer.from("This is a readme").toString("base64"),
+						metadata: {
+							contentType: "text/markdown",
+						},
+						base64: true,
+					},
+				]);
+				return res.once(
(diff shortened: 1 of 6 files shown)


cmsparks (approved):
lgtm! thanks for making the changes

packages/wrangler/src/__tests__/pages/deploy.test.ts
```diff
@@ -1534,6 +1534,160 @@ describe("deployment create", () => {
...
+		writeFileSync(
+			"public/_worker.js",
+			`
+			import { thing } from "node:some-module";
+			import { thing2 } from "cloudflare:some-module";
+
+      export default {
+        async fetch(request, env) {
+					console.log(thing);
+					console.log(thing2);
```
GregBrimble, line 1552: I think this should become two tests:

1. As it was, ensuring we do in fact get that error message this is failing for

```suggestion
		writeFileSync(
			"public/_worker.js",
			`
			import { thing } from "some-module";

      export default {
        async fetch(request, env) {
					console.log(thing);
```

This check was introduced to ensure `_worker.js` files aren't referencing external modules that won't be available at runtime.

2. A second test which produces an `_worker.js/index.js` with `--no-bundle` which _doesn't_ error. Wrangler should make no effort to help the user if they are providing a `_worker.js/index.js` file because that really does indicate they know what they're doing (they must provide all modules in that directory). Don't try and emulate workerd's module resolution here to test to see if they're present — just blindly upload them and trust the user did it correctly.

cmsparks: Just updated the tests to include a case with a `_worker.js/` directory with the no bundle flag, which should be sufficient. Looks like everything is passing, let me know if there's anything I'm missing.

packages/wrangler/src/__tests__/pages/deploy.test.ts
```diff
@@ -1534,6 +1534,304 @@ describe("deployment create", () => {
...
+						["manifest", "_worker.bundle"].sort()
+					);
+
+					expect(manifest).toMatchInlineSnapshot(`
+				                                      Object {
+				                                        "/README.md": "13a03eaf24ae98378acd36ea00f77f2f",
+				                                      }
+			                                `);
+
+					expect(workerHasD1Shim(workerBundle as string)).toBeFalsy();
```
IgorMinar, line 1631: Is this check still valid? I thought D1 shims were no longer needed by D1. Does wrangler still inject them with --no-bundle=false option?

packages/wrangler/src/__tests__/pages/deploy.test.ts
```diff
@@ -1534,6 +1534,304 @@ describe("deployment create", () => {
...
+					expect(workerBundle).toContain(`some-module`);
```
IgorMinar, line 1781: since this module is unknown to the runtime, in practice a deployment of this bundle would fail. Could you make the test a bit more realistic by using a dynamic import to a second es module stored within this bundle?

IgorMinar (commented):
I fear that this change will cause a regression for next-on-pages and other frameworks that rely on the _worker.js/ directory not to be bundled by default.

I asked @dario-piotrowicz to take a look at this PR and confirm if changing this is a problem or not.

I wonder if it was an oversight that we enabled bundling for _workers.js file as that will break source maps. Was bundling initially needed only to support the older version of D1 that required shims? I thought we removed those shims from D1/wrangler a while back.

GregBrimble (commented):
@IgorMinar , by default, we do bundle. And this doesn't change that behavior. All this PR does is now ensure we _don't_ bundle if the `--no-bundle` arg is provided. next-on-pages shouldn't be affected, but I of course welcome any testing @dario-piotrowicz does before we merge.

@cmsparks , LGTM, thanks for getting this over the line!

dario-piotrowicz:
@IgorMinar I had a look and everything actually seems fine to me, as far as next-on-pages is concerned, as long as the additional modules are not bundled in into the index.js file it's all good and that does seem to be the case:
Screenshot 2024-03-21 at 18 40 56

Indeed they are never part of the bundle anyways: https://github.com/cloudflare/workers-sdk/blob/f8aa8bfb419102f18efc76ea7f72a477cfdc0746/packages/wrangler/src/pages/functions/buildWorker.ts#L312

So I do think it's all good 🙂

PR #5207, opened 2024-03-08
[D1] fix `wrangler d1 migration apply` to for queries not ending in a semi-colon

rozenmd:
Prior to this PR, migrations such as:
```
SELECT 1
```

would fail with a nonsensical error about INSERTS: `✘ [ERROR] near "INSERT": syntax error at offset 199 [code: 7500]`

What was happening was that wrangler was injecting a statement at the end of the migration, to update the migrations table saying the query successfully executed.

To avoid this, this PR moves the statement to the start of the migration.

### DIFF
--- a/.changeset/gorgeous-taxis-argue.md
+++ b/.changeset/gorgeous-taxis-argue.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: fix migration apply to allow migrations that dont end in a semicolon
--- a/.changeset/gorgeous-taxis-argue.md
+++ b/.changeset/gorgeous-taxis-argue.md
@@ -2,4 +2,10 @@
 "wrangler": patch
 ---
 
-fix: fix migration apply to allow migrations that dont end in a semicolon
+fix: fix `wrangler d1 migrations apply` to allow migrations that dont end in a semicolon
+
+Prior to this PR, migrations such as: `SELECT 1` would fail with a nonsensical error about INSERTS: `✘ [ERROR] near "INSERT": syntax error at offset 199 [code: 7500]`
+
+What was happening was that wrangler was injecting a statement at the end of the migration, to update the migrations table saying the query successfully executed.
+
+To avoid this, this PR moves the statement to the start of the migration.
--- a/.changeset/gorgeous-taxis-argue.md
+++ b/.changeset/gorgeous-taxis-argue.md
@@ -2,7 +2,7 @@
 "wrangler": patch
 ---
 
-fix: fix `wrangler d1 migrations apply` to allow migrations that dont end in a semicolon
+fix: fix `wrangler d1 migrations apply` to allow migrations that don't end in a semicolon
 
 Prior to this PR, migrations such as: `SELECT 1` would fail with a nonsensical error about INSERTS: `✘ [ERROR] near "INSERT": syntax error at offset 199 [code: 7500]`
 
--- a/packages/wrangler/src/d1/migrations/apply.tsx
+++ b/packages/wrangler/src/d1/migrations/apply.tsx
@@ -146,15 +146,10 @@ Your database may not be available to serve requests during the migration, conti
 		}
 
 		for (const migration of unappliedMigrations) {
-			let query = fs.readFileSync(
-				`${migrationsPath}/${migration.name}`,
-				"utf8"
-			);
-			query += `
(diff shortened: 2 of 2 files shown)


.changeset/gorgeous-taxis-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: fix migration apply to allow migrations that dont end in a semicolon
```
rozenmd, line 5:
```suggestion
fix: fix `wrangler d1 migrations apply` to allow migrations that dont end in a semicolon

Prior to this PR, migrations such as: `SELECT 1` would fail with a nonsensical error about INSERTS: `✘ [ERROR] near "INSERT": syntax error at offset 199 [code: 7500]`

What was happening was that wrangler was injecting a statement at the end of the migration, to update the migrations table saying the query successfully executed.

To avoid this, this PR moves the statement to the start of the migration.
```

.changeset/gorgeous-taxis-argue.md
```diff
@@ -0,0 +1,11 @@
+---
+"wrangler": patch
+---
+
+fix: fix `wrangler d1 migrations apply` to allow migrations that dont end in a semicolon
```
justin-mp, line 5:
```suggestion
fix: fix `wrangler d1 migrations apply` to allow migrations that don't end in a semicolon
```

packages/wrangler/src/d1/migrations/apply.tsx
```diff
@@ -146,15 +146,10 @@ Your database may not be available to serve requests during the migration, conti
...
 
 		for (const migration of unappliedMigrations) {
-			let query = fs.readFileSync(
-				`${migrationsPath}/${migration.name}`,
-				"utf8"
-			);
-			query += `
-								INSERT INTO ${migrationsTableName} (name)
-								values ('${migration.name}');
+			let query = `
```
justin-mp, line 149: Why is it safe to update the migrations table to say we've completed the migration *before* the migration happens?  Are we relying on transactions somehow?

justin-mp (commented):
Out of curiosity, why do we want to allow folks to have migrations with incomplete statements?  That seems buggy to me.

packages/wrangler/src/d1/migrations/apply.tsx
rozenmd: My assumption is that a query string is executed as a single query, so if one of the queries fail, the DO throws... right @geelen ?

rozenmd:
> Out of curiosity, why do we want to allow folks to have migrations with incomplete statements? That seems buggy to me.
@justin-mp SQL statements that don't terminate in a semi-colon are still valid queries: 
image

justin-mp:
> > Out of curiosity, why do we want to allow folks to have migrations with incomplete statements? That seems buggy to me.
> 
> @justin-mp SQL statements that don't terminate in a semi-colon are still valid queries

Interesting.  According to https://sqlite.org/syntax/sql-stmt-list.html, it looks like the semicolon is indeed optional. 

However, the `sqlite3` command seems to require it

```
$ sqlite3 
SQLite version 3.37.2 2022-01-06 13:25:41
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> select 1
   ...>
```
(It won't accept the command until I pass in a `;`.)

rozenmd:
Worth noting that the sqlite CLI and sqlite core have slightly different implementations

petebacondarwin (approved):

petebacondarwin:
Just as a retro to this change... @rozenmd - can you write down why there were no test additions for this change?

rozenmd:
No reason @petebacondarwin - I was waiting for a reply from @geelen before merging: https://github.com/cloudflare/workers-sdk/pull/5207#discussion_r1517827599

petebacondarwin:
Oh! So should we actually revert this merge until the questions have been resolved?

rozenmd:
I think that'd be best @petebacondarwin - then I can try write a test for it too

packages/wrangler/src/d1/migrations/apply.tsx
geelen: Wow I guess I missed this at the time (I am cleaning up some old tabs), but yes. Anything submitted in a single query to a D1 is executed in a transaction.

PR #5208, opened 2024-03-08
Implement `versions list` + `versions view` commands

RamIdeas:

This PR implements the `versions list` and `versions view <version-id>` commands. They are very similar so a single PR seemed fine.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: currently internal and experimental

### DIFF
--- a/packages/wrangler/src/__tests__/versions/versions.view.test.ts
+++ b/packages/wrangler/src/__tests__/versions/versions.view.test.ts
@@ -26,17 +26,22 @@ describe("versions view", () => {
 				`[Error: Not enough non-option arguments: got 0, need at least 1]`
 			);
 
-			expect(normalizeOutput(std.out)).toMatchInlineSnapshot(`
-			"wrangler versions view <version-id>
+			expect(std.out).toMatchInlineSnapshot(`
+			"
+			wrangler versions view <version-id>
+
 			View the details of a specific version of your Worker [beta]
+
 			Positionals:
 			  version-id  The Worker Version ID to view  [string] [required]
+
 			Flags:
 			  -j, --experimental-json-config  Experimental: Support wrangler.json  [boolean]
 			  -c, --config                    Path to .toml configuration file  [string]
 			  -e, --env                       Environment to use for operations and .env files  [string]
 			  -h, --help                      Show help  [boolean]
 			  -v, --version                   Show version number  [boolean]
+
 			Options:
 			      --name  Name of the worker  [string]"
 		`);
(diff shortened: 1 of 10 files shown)


packages/wrangler/src/__tests__/versions/versions.view.test.ts
```diff
@@ -0,0 +1,200 @@
...
+		test("fails with non-existent version-id", async () => {
+			const result = runWrangler(
+				"versions view ffffffff-ffff-ffff-ffff-ffffffffffff  --experimental-gradual-rollouts"
+			);
+
+			await expect(result).rejects.toMatchInlineSnapshot(
+				`[APIError: A request to the Cloudflare API (/accounts/some-account-id/workers/scripts/test-name/versions/ffffffff-ffff-ffff-ffff-ffffffffffff) failed.]`
+			);
+
+			expect(normalizeOutput(std.out)).toMatchInlineSnapshot(`
```
mrbbot, line 191: Should this not be logged to stderr?

packages/wrangler/src/utils/render-labelled-values.ts
```diff
@@ -0,0 +1,22 @@
+import { logger } from "../logger";
+
+export default function renderLabelledValues(
```
mrbbot, line 3: A comment with example output from this function might be nice.

packages/wrangler/src/versions/list.ts
```diff
@@ -0,0 +1,134 @@
...
+		author_id: string;
+		author_email: string;
+	};
+	annotations?: {
+		"workers/triggered_by"?: "upload" | string;
+		"workers/message"?: string;
+		"workers/tag"?: string;
+	};
+	// other properties not typed as not used
+};
```
mrbbot, line 38: Do these types get extracted into some common file in a later PR? They seem to be duplicated.

packages/wrangler/src/versions/list.ts
```diff
@@ -0,0 +1,134 @@
...
+		case "api":
+			return "API 📡";
+		case "dash":
+			return "Dashboard 🖥️";
+		case "wrangler":
+			return "Wrangler 🤠";
+		case "terraform":
+			return "Terraform 🏗️";
+		default:
+			return "Other";
```
mrbbot, line 118: Should we include the actual value for `Other` and `Unknown` below?
```suggestion
						return `Other (${source})`;
```

packages/wrangler/src/versions/view.ts
```diff
@@ -0,0 +1,87 @@
...
+	number: number;
+	metadata: {
+		created_on: string;
+		modified_on: string;
+		source: "api" | string;
+		author_id: string;
+		author_email: string;
+	};
+	annotations?: Record<string, string> & {
+		"workers/triggered_by"?: "upload" | string;
```
mrbbot, line 32: Should this have the other values, or just be `string`?

mrbbot (approved):
Added some comments, but all non-blocking. Looks good! ✅ Assuming no changeset as this is still experimental, or are you planning to add one before merging?

packages/wrangler/src/__tests__/versions/versions.view.test.ts
RamIdeas: Interesting. This is just the fetchResult call rejecting. Which should bubble up to the main() try..catch error handling. Does wrangler log errors to stdout??

packages/wrangler/src/versions/list.ts
RamIdeas: I was thinking that too – will add

packages/wrangler/src/versions/list.ts
RamIdeas: Yep

packages/wrangler/src/versions/view.ts
RamIdeas: I extract it into another file in the next PR so will leave this for now

packages/wrangler/src/versions/list.ts
RamIdeas: Did this in the next PR instead https://github.com/cloudflare/workers-sdk/pull/5224/commits/abc57e3e32e6050164561ebd4bbf9ed6b1e39ec1

PR #5214, opened 2024-03-08
File based dev registry

penalosa:

Fixes https://github.com/cloudflare/workers-sdk/issues/6037

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not public facing—internal stability refactor gated behind a feature flag

### DIFF
--- a/packages/wrangler/src/dev.tsx
+++ b/packages/wrangler/src/dev.tsx
@@ -12,6 +12,7 @@ import { getVarsForDev } from "./dev/dev-vars";
 import { getLocalPersistencePath } from "./dev/get-local-persistence-path";
 import { startDevServer } from "./dev/start-server";
 import { UserError } from "./errors";
+import { run } from "./experimental-flags";
 import { logger } from "./logger";
 import * as metrics from "./metrics";
 import { getAssetPaths, getSiteAssetPaths } from "./sites";
@@ -295,6 +296,13 @@ export function devOptions(yargs: CommonYargsArgv) {
 					"Use the experimental DevEnv instantiation (unified across wrangler dev and unstable_dev)",
 				default: false,
 			})
+			.option("experimental-registry", {
+				alias: ["x-registry"],
+				type: "boolean",
+				describe:
+					"Use the experimental file based dev registry for multi-worker development",
+				default: false,
+			})
 	);
 }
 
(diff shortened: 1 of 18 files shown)


penalosa:
- Add file heartbeat to detect crashed workers
- Add Wrangler version to files to detect back compat
- Add cleanup mechanism when Wrangler tries to fetch a worker that no longer exists

packages/wrangler/src/dev.tsx
```diff
@@ -295,6 +296,13 @@ export function devOptions(yargs: CommonYargsArgv) {
 					"Use the experimental DevEnv instantiation (unified across wrangler dev and unstable_dev)",
 				default: false,
 			})
+			.option("experimental-registry", {
+				alias: ["x-registry"],
```
RamIdeas, line 300: 👏

packages/wrangler/src/experimental-flags.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const flags = new AsyncLocalStorage<ExperimentalFlags>();
+
+export const run = <V>(flagValues: ExperimentalFlags, cb: () => V) =>
+	flags.run(flagValues, cb);
+
+export const FILE_BASED_REGISTRY = () => flags.getStore()?.FILE_BASED_REGISTRY;
+
+// TODO: actually use this
+export const DEV_ENV = () => flags.getStore()?.DEV_ENV;
```
RamIdeas, line 16: Can we discuss alternatives to this interface please?

RamIdeas: Minor point about interface:

I'd like to make adding a feature flag as easy as adding it to a type with no repetition. Having an exported function per flag requires multiple changes to this file to add each flag. While this allows us to find all usages of _each_ feature flag with Find All References, they won't be easy to find usages of _any_ flag. Can we try something like `export const XFLAG = <F extends keyof ExperimentalFlags>(flag: F) => ExperimentalFlags[F];`?

More for discussion:

This is a great use of AsyncLocalStorage. Can we generalise this use for other singletons like Logger etc? Are there any parts of wrangler this wouldn't work for?

RamIdeas (approved):

petebacondarwin (commented):
Didn't have time to review it. But I see that @RamIdeas has approved.

PR #5215, opened 2024-03-09
Add support for named entrypoints

GregBrimble:
LDW exemption: CR-849919

---

(updated by @mrbbot)

This PR adds support for named entrypoints to `wrangler deploy` and `wrangler (pages) dev`. This also enables support for JS RPC over multiple `wrangler dev` sessions.

Service bindings to other `wrangler dev` sessions are now implemented with `workerd`-native `external` service bindings instead of middleware. This is required to use Cap'n Proto over HTTP for JSRPC. This middleware was the last middleware that wrapped the `env` object. Code for wrapping `env` has been removed from the middleware loader. Implementing this for `WorkerEntrypoint`s would've required us to look for named exports and apply `env` wrapping to all of these.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Not necessary because:


packages/wrangler/templates/facade.d.ts
```diff
@@ -1,9 +1,24 @@
+// TODO(soon): remove once https://github.com/cloudflare/workerd/pull/1870 lands
```
GregBrimble, line 1: Eagle has landed 🦅

Just need to publish to npm, and then we should be good to go. Looks like there was a problem with that latest run though: https://github.com/cloudflare/workerd/actions/runs/8438068332

packages/miniflare/src/index.ts
```diff
@@ -1173,26 +1176,32 @@ export class Miniflare {
...
-			const { unsafeDirectHost, unsafeDirectPort } = workerOpts.core;
-			if (unsafeDirectHost !== undefined || unsafeDirectPort !== undefined) {
-				const name = getDirectSocketName(i);
+			const previousDirectSockets =
+				previousWorkerOpts?.core.unsafeDirectSockets ?? [];
+			const directSockets = workerOpts.core.unsafeDirectSockets ?? [];
+			for (let j = 0; j < directSockets.length; j++) {
+				const previousDirectSocket = previousDirectSockets[j];
+				const directSocket = directSockets[j];
+				const entrypoint = directSocket.entrypoint ?? "default";
```
GregBrimble, line 1185: Entirely not an issue — just a general question: I wonder if, instead of using `"default"` as a semantic value here, it would have been potentially a bit easier to allow `entrypoint` to be `undefined` and use that to mean the default export? I know `default` is a protected word, so no risk of a collision, but it just seems like you're needing to flip back and forth a bit here and in other files from that `"default"` value to the special case.

I also had a bit of an issue with this myself in the UI — React doesn't like undefined/null input values, so I went with an empty string, but same diff.

packages/wrangler/src/cli.ts
```diff
@@ -29,3 +29,4 @@ export type { UnstableDevWorker, UnstableDevOptions };
 
 export * from "./api/integrations";
 export { default as unstable_splitSqlQuery } from "./d1/splitter";
+export { startWorkerRegistryServer as unstable_startWorkerRegistryServer } from "./dev-registry";
```
GregBrimble, line 32: Getting close to pulling this out entirely! :)

packages/wrangler/src/dev-registry.ts
```diff
@@ -9,20 +10,29 @@ import type { Config } from "./config";
 import type { HttpTerminator } from "http-terminator";
 import type { Server } from "node:http";
 
-const DEV_REGISTRY_PORT = 6284;
+// Safety of `!`: `parseInt(undefined)` is NaN
+// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+let DEV_REGISTRY_PORT = parseInt(process.env.WRANGLER_WORKER_REGISTRY_PORT!);
```
GregBrimble, line 15: Do we document these anywhere? Worth adding to the list?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -61,6 +76,7 @@ function createClass({ className, proxyUrl }) {
 			proxyRequest.headers.set(HEADER_URL, request.url);
 			proxyRequest.headers.set(HEADER_NAME, className);
 			proxyRequest.headers.set(HEADER_ID, this.id);
+			proxyRequest.headers.set(HEADER_CF_BLOB, JSON.stringify(request.cf));
```
GregBrimble, line 79: Is everything in `request.cf` JSON-stringify-able? I'm guessing yes, but just double-checking.

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -175,21 +194,34 @@ function buildLog(): Log {
 	return new WranglerLog(level, { prefix: "wrangler-UserWorker" });
 }
 
+async function getEntrypointNames(entrypointSource: string) {
+	await esmLexer.init;
+	const [_imports, exports] = esmLexer.parse(entrypointSource);
+	// TODO(soon): support `export * from "...";` with `--no-bundle`, `esbuild`
+	//  will bundle these so we don't need to worry about this there
```
GregBrimble, line 201:
```suggestion
	// TODO(soon): support `export * from "...";` with `--no-bundle`. Without
	// `--no-bundle`,  `esbuild` will bundle these, so they'll be picked up here.
```

GregBrimble: Assuming I understood that correctly.

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -322,6 +356,7 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 		// script. They don't need to persist anything, and would end up using the
 		// incorrect unsafe unique key.
 		unsafeEphemeralDurableObjects: true,
+		compatibilityDate: "2024-01-01",
```
GregBrimble, line 359: Been a hot minute since I was in here. This is a system Worker we spin up to proxy the DOs, right? So just double-checking that we're good to set our own compat date here? We don't need to use a user-provided value of either the caller or callee, right?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -355,8 +394,88 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
...
+			// If this is a service binding to the current worker, don't bother using
+			// the dev registry to look up the address, just bind to it directly.
+			serviceBindings[service.binding] = getName(config);
+			continue;
+		}
+
+		const target = config.workerDefinitions?.[service.service];
+		if (target?.host === undefined || target.port === undefined) {
+			// If the target isn't in the registry, always return an error response
+			// TODO(someday): could we do some sort of `Proxy`-prototype thing here
```
GregBrimble, line 412: Yes, that sounds nice :) I'll file a fast-follow ticket for that.

GregBrimble: WC-2058

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -355,8 +394,88 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
...
+			//  for a nice error message if a user tried to call an RPC method?
+			serviceBindings[service.binding] = () => {
+				const message = `[wrangler] Couldn't find \`wrangler dev\` session for service "${service.service}" to proxy to`;
```
GregBrimble, line 415: What makes `Response` here make more sense than blowing up with a `UserError`? Is one more graceful than the other, and/or is can we just not recover from a certain types of these failures?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -355,8 +394,88 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
...
+					throw new UserError(
+						`The \`wrangler dev\` session for service "${service.service}" does not support proxying entrypoints. Please upgrade "${service.service}"'s \`wrangler\` version.`
+					);
+				}
+				const entrypointAddress =
+					target.entrypointAddresses[service.entrypoint];
+				if (entrypointAddress === undefined) {
+					// ...but the named entrypoint doesn't exist, throw
+					throw new UserError(
+						`The \`wrangler dev\` session for service "${service.service}" does export an entrypoint named "${service.entrypoint}"`
```
GregBrimble, line 437:
```suggestion
						`The \`wrangler dev\` session for service "${service.service}" does not export an entrypoint named "${service.entrypoint}"`
```

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -355,8 +394,88 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
...
+			} else {
+				// Otherwise, if the user hasn't specified a named entrypoint, assume
+				// they meant to bind to the `default` entrypoint.
+				const defaultEntrypointAddress =
+					target.entrypointAddresses?.["default"];
+				if (defaultEntrypointAddress === undefined) {
+					// If the "server" `wrangler` is too old to provide direct entrypoint
+					// addresses, fallback to sending requests directly to the target...
+					if (target.protocol === "https") {
+						// ...unless the target is listening on HTTPS, in which case throw.
```
GregBrimble, line 450: Is this a limitation of local dev service bindings today? Can I not bind to a service if that server is set up to use HTTPS?

packages/wrangler/src/dev-registry.ts
```diff
@@ -205,7 +229,8 @@ export async function getBoundRegisteredWorkers({
 	const filteredWorkers = Object.fromEntries(
 		Object.entries(workerDefinitions || {}).filter(
 			([key, _value]) =>
-				serviceNames.includes(key) || durableObjectServices.includes(key)
+				key !== name && // Always exclude current worker to avoid infinite loops
```
GregBrimble, line 232: Mind explaining this to me? What loop is possible here? Was this a problem before this PR?

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+	// Fixture for creating a temporary directory
+	async tmpPath({}, use) {
+		const tmpPath = await fs.realpath(await fs.mkdtemp(tmpPathBase));
+		await use(tmpPath);
+		await fs.rm(tmpPath, { recursive: true, maxRetries: 10 });
+	},
+	// Fixture for starting an isolated dev registry server on a random port
+	async isolatedDevRegistryPort({}, use) {
+		// Start a standalone dev registry server for each test
+		const result = await unstable_startWorkerRegistryServer(0);
```
GregBrimble, line 58: We should have done this ages ago :) Thanks!

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+		const response = await fetch(url);
+		const text = await response.text();
+		expect(text).toBe("b:a:pong");
+	});
+});
+
+test("should support binding to Durable Object in another worker", async ({
+	dev,
+}) => {
+	// RPC isn't supported in this case yet :(
```
GregBrimble, line 479: Is that a runtime limitation? Or a local dev thing?

GregBrimble: Do we want to add an expect-error test? (would be good to assert we have a good error message in this case if possible)

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+			};
+			export default {}; // Required to treat as modules format worker
+		`,
+	});
+
+	// Wait for error to be thrown
+	await waitFor(() => {
+		const output = session.getOutput();
+		expect(output).toMatch(
+			'The `wrangler dev` session for service "bound" does export an entrypoint named "ThingEntrypoint"'
```
GregBrimble, line 714:
```suggestion
			'The `wrangler dev` session for service "bound" does not export an entrypoint named "ThingEntrypoint"'
```

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+		}
+	);
+	expect(response.status).toBe(200);
+	expect(await response.text()).toBe("null");
+
+	// Wait for error to be thrown
+	await waitFor(() => {
+		const output = session.getOutput();
+		expect(output).toMatch(
+			'Cannot proxy to `wrangler dev` session for service "bound" because it uses HTTPS. Please remove the `--local-protocol`/`dev.local_protocol` option.'
```
GregBrimble, line 822: If you _can_ bind to a HTTPS in new versions of Wrangler, should this error message be prompting you to upgrade rather than suggesting removing `--local-protocol=https`?

GregBrimble (commented):
Couple of questions, at least one blocking fix to an error message, but all-in-all, fantastic and comprehensive work!

.changeset/tidy-fans-speak.md
```diff
@@ -0,0 +1,48 @@
...
+```ts
+import { kCurrentWorker, Miniflare } from "miniflare";
+
+const mf = new Miniflare({
+	workers: [
+		{
+			name: "a",
+			serviceBindings: {
+				A_RPC_SERVICE: { name: kCurrentWorker, entrypoint: "RpcEntrypoint" },
+				A_NAMED_SERVICE: { name: "a", entrypoint: "namedEntrypoint" },
```
RamIdeas, line 18: Are these service-bindings to entrypoints in the same worker itself?

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -0,0 +1,825 @@
...
+		// Start a standalone dev registry server for each test
+		const result = await unstable_startWorkerRegistryServer(0);
+		const address = result.server.address();
+		assert(typeof address === "object" && address !== null);
+		await use(address.port);
+		await result.terminator.terminate();
+	},
+	// Fixture for starting a worker in a temporary directory, using the test's
+	// isolated dev registry
+	async dev({ tmpPath, isolatedDevRegistryPort }, use) {
```
RamIdeas, line 66: This doesn't seem to match the signature of StartDevSession

Am I missing something?

packages/miniflare/src/index.ts
mrbbot: In `workerd` at least, you can omit the entrypoint and it will assume `default`. That's actually what we do later on. This is just to simplify some of the logic to allow users to use `default` as a named entrypoint themselves, and have that work.

packages/miniflare/.prettierrc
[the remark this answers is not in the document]
mrbbot: Oh yeah, we've been putting off doing this for a while. Just added this now because I got annoyed with all my imports being re-ordered... 😅

packages/wrangler/src/cli.ts
mrbbot: Have you seen https://github.com/cloudflare/workers-sdk/pull/5214? This should improve stability here too. I think Samuel was going to revisit this in a few weeks time.

packages/wrangler/src/dev-registry.ts
mrbbot: We could, but given https://github.com/cloudflare/workers-sdk/pull/5214 I'm not sure how long this incarnation of the dev registry will be hanging around. I can see this variable being useful for multi-worker `unstable_dev()` tests, but ideally people should just use the Vitest integration for testing! 😃

packages/wrangler/src/dev/miniflare.ts
mrbbot: Yep, https://github.com/cloudflare/workerd/blob/a14b5b55617e7209ee7d3753bcb8a69608d0cff8/src/workerd/server/workerd.capnp#L803-L805

packages/wrangler/src/dev/miniflare.ts
mrbbot: Correct, we have full-control of these worker so don't need to respect user compatibility settings here.

packages/wrangler/src/dev/miniflare.ts
mrbbot: In this case where we're returning a response, we couldn't find a session to proxy to. This is probably because the user hasn't started one up. We defer the error to when they actually try to call it with the intention that they'll run `wrangler dev` for the other session before they send a request.

With the other `UserError` cases, we found a session, it's just not configured correctly.

I also wanted to avoid a breaking change here. Currently, when Wrangler doesn't find a target, it doesn't error and will only return an error response when trying to send a request.

packages/wrangler/src/dev/miniflare.ts
mrbbot: Yep, this is a limitation today. It fails with some cryptic internal error. 😃 This is just making that clearer. 👍

packages/wrangler/src/cli.ts
GregBrimble: Briefly, yes. Will be interesting to see if that improves things!

packages/wrangler/src/dev-registry.ts
mrbbot: It's possible this isn't required anymore actually. This was the issue I was trying to avoid: say we have Workers A and B, both with named entrypoints. Worker A starts up first and listens on a random port for each entrypoint. Worker B starts up and again listens on a random port for each entrypoint. Worker A sees Worker B startup, and reloads with new configuration, _and new random ports for each entrypoint_. These new random ports are stored in the registry, which Worker B sees, triggering it to reload with new configuration and generate more random ports. This causes A to update again and so on.

I ended up fixing this by trying to reuse the same random ports for each entrypoint when reloading, but this was my first attempt at fixing it. I've kept it in though since it doesn't make sense for a Worker to bind to itself through the registry. It should aways be able to use a more efficient same-process service binding here.

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: Yeah, probably worth going through the other multi-worker tests and updating them to use this.

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: Local dev thing. Cap'n Proto over HTTP is only supported for service bindings, not Durable Objects unfortunately. 🙁 Will add an expect-error test though. 👍

packages/wrangler/src/dev-registry.ts
GregBrimble: Yeah, I mean, there are a bunch of them now, right? e.g. `WRANGLER_LOG_SANTIZE`, `WRANGLER_SEND_METRICS`, etc. etc. If you're not putting those anywhere (yet?), then definitely no rush on this one either.

packages/wrangler/src/dev/miniflare.ts
GregBrimble: Thoughtful and justified. Nice!

.changeset/tidy-fans-speak.md
GregBrimble: Yup. That is supported 👍

mrbbot: Yep, `kCurrentWorker` and `a` are equivalent here. 👍

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: `StartDevSession` is the type of the value this `dev` fixture provides. The type of `use` is `(fn: StartDevSession) => Promise<void>`. `{ tmpPath, isolatedDevRegistryPort }` are the values provided by the other fixtures for the specific test. See https://vitest.dev/guide/test-context#extend-test-context for more details.

packages/wrangler/src/dev/miniflare.ts
mrbbot: 2d6098ea6a20825e39d3d255730a2a1aa0fe036e

packages/wrangler/src/dev/miniflare.ts
mrbbot: 2d6098ea6a20825e39d3d255730a2a1aa0fe036e

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: 2d6098ea6a20825e39d3d255730a2a1aa0fe036e

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: 2d6098ea6a20825e39d3d255730a2a1aa0fe036e

packages/wrangler/src/dev/miniflare.ts
mrbbot: I nerd sniped myself and did it now: a9f693e9826421e0e896118d376d5decf4c7e0e6 😃

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: Added in 400bd760ea85f4f26dcae8a041f63c12e29f1528, after doing some proxy magic to make the error message nice. 👍

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -529,6 +545,15 @@ test("should support binding to Durable Object in another worker", async ({
 		// Check protocol, host, and cf preserved
 		expect(text).toBe('POST https://placeholder:9999/ {"thing":true}');
 	});
+
+	const rpcResponse = await fetch(new URL("/rpc", url));
+	const errors = await rpcResponse.json();
+	expect(errors).toMatchInlineSnapshot(`
+		[
+		  "Error: Cannot access \`ThingObject#property\` as Durable Object RPC is not yet supported between multiple \`wrangler dev\` sessions. We recommend you only bind to Durable Objects defined in the same Worker. You can define an entrypoint sub-classing \`WorkerEntrypoint\` to expose an RPC interface between Workers.",
+		  "Error: Cannot access \`ThingObject#method\` as Durable Object RPC is not yet supported between multiple \`wrangler dev\` sessions. We recommend you only bind to Durable Objects defined in the same Worker. You can define an entrypoint sub-classing \`WorkerEntrypoint\` to expose an RPC interface between Workers.",
```
GregBrimble, line 554: I find these other points either unnecessary.

1. Asking them to move the DO to a different Worker is probably untenable — that would be the default setup, so they're only doing cross-Worker because they have to (e.g. Pages -> DOs).
2. Offering `WorkerEntrypoint`s as an alternative to DOs is only giving them RPC. I think it's very unlikely that the thing they _need_ is RPC — rather, the thing they need is probably DO state or alarms or whatever. RPC is just a nice interface layer that happens to be available in both entrypoints and DOs.

```suggestion
		  "Error: Cannot access \`ThingObject#property\` as Durable Object RPC is not yet supported between multiple \`wrangler dev\` sessions.",
		  "Error: Cannot access \`ThingObject#method\` as Durable Object RPC is not yet supported between multiple \`wrangler dev\` sessions.",
```

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -53,33 +53,88 @@ import type { Readable } from "node:stream";
...
+		
+	Reflect.setPrototypeOf(klass.prototype, handlerSuperKlass.prototype);
+	Reflect.setPrototypeOf(klass, handlerSuperKlass);
+
+	return klass;
+}
+
+function createDurableObjectClass({ className, proxyUrl }) {
+	const klass = createProxyPrototypeClass(DurableObject, (key) => {
+		throw new Error(\`Cannot access \\\`\${className}#\${key}\\\` as Durable Object RPC is not yet supported between multiple \\\`wrangler dev\\\` sessions. We recommend you only bind to Durable Objects defined in the same Worker. You can define an entrypoint sub-classing \\\`WorkerEntrypoint\\\` to expose an RPC interface between Workers.\`);
```
GregBrimble, line 107:
```suggestion
		throw new Error(\`Cannot access \\\`\${className}#\${key}\\\` as Durable Object RPC is not yet supported between multiple \\\`wrangler dev\\\` sessions.\`);
```

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: Hmmm, the aim here is to say that as an alternative to defining just the Durable Object in a separate worker, you can bind to another worker as a service, then have it do all of the Durable Object accesses. I think this is now the recommended approach? Rather than binding to Durable Objects in another worker, you bind to the service instead.

GregBrimble: Oh, gotcha. That wasn't clear to me on reading the code, but perhaps with the context of a proper project in front of me, I would have understood.

Binding to a different Worker's service, rather than directly to a Durable Object namespace owned by a different Worker is, I'd say, "soft recommended". We don't actually say that anywhere, to my knowledge, and we certainly don't warn about it anywhere. So it definitely is possible to, in a first-class way, bind to those other Durable Object namespaces (without it, Pages couldn't use Durable Objects at all).

Is it likely we'll eventually be able to support Durable Object RPC across multiple sessions? I'd imagine, yes? In which case, I'd still personally suggest trimming off those final two sentences from this error message for now.

.changeset/eight-seahorses-serve.md
```diff
@@ -0,0 +1,44 @@
...
+binding = "SERVICE"
+service = "bound"
+entrypoint = "EntrypointA"
+```
+
+To bind to named entrypoints with `wrangler pages dev`, use the `#` character:
+
+```shell
+$ wrangler pages dev --service=SERVICE=bound#EntrypointA
+```
```
RamIdeas, line 44: Should we encourage users to use the config file now supported for pages commands?

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
RamIdeas: Can we add a hidden entrypoint to `bound` which proxies to `bound`'s Durable Object? The error message says they can do it themselves but, if binding directly to DOs from another worker works in production, shouldn't we hook this up for them?

mrbbot (approved):
Approving on behalf of Greg as he can't approved his own PR. Chatted to him, and he's happy. 🙂

packages/wrangler/templates/facade.d.ts
RamIdeas: Are we going to remove this file before merging?

packages/wrangler/src/api/startDevWorker/events.ts
```diff
@@ -149,4 +150,5 @@ export type ProxyData = {
 	liveReload?: boolean;
 	proxyLogsToController?: boolean;
 	internalDurableObjects?: CfDurableObject[];
+	entrypointAddresses: WorkerEntrypointsDefinition | undefined;
```
RamIdeas, line 153: Noticed you've added this property and set it everywhere it's needed but I don't see where it's read/used

RamIdeas (approved):

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
mrbbot: > Can we add a hidden entrypoint to `bound` which proxies to `bound`'s Durable Object? The error message says they can do it themselves...

What I'm recommending here is implementing an entrypoint in front of your Durable Object that implements a public API for your Durable Objects. It's not possible to directly expose a Durable Object and its RPC members directly. Whilst it is in theory possible to define a `Proxy`-prototyped `WorkerEntrypoint` that intercepts accesses to any possible RPC method and forwards them onto a Durable Object, the tricky thing is working out which Durable Object ID to send it to. You could define a "client" `DurableObject` that added the Durable Object ID as a parameter to every call, and unwrapped this on the "server" `WorkerEntrypoint` that then forwarded the request to the actual Durable Object, but this feels like it might be tricky to get right, and could introduce behaviour differences between the local and remote RPC implementations.

.changeset/eight-seahorses-serve.md
mrbbot: If named entrypoints are supported in that with Pages then probably, but given this is just a changeset and not the formal docs, and most  `pages dev` users will be familiar with the `--service` flag over a config file, I think this is fine.

packages/wrangler/src/api/startDevWorker/events.ts
mrbbot: Yeah, I think we should be able to tidy this up then. 👍

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
GregBrimble: Okay, I'm just going to commit my suggestion here then as-is, and we can see what user feedback looks like :)

packages/wrangler/templates/facade.d.ts
GregBrimble: Thanks for the reminder. Yes, I'll do that now.

GregBrimble:
Okay, once these pass, I'm merging and will follow up with another PR to bump to [the latest workerd version](url) 🚀

PR #5218, opened 2024-03-11
[C3] Reorganize helper scripts

jculvey:

This is a re-organization of c3's helper scripts. Previously, the `common.ts` and to some degree `command.ts` became junk drawers that we filled with a variety of helper functions. These files became large and unweildy, had low cohesion, and weren't conducive to unit testing.

I added unit test coverage in several places where it was lacking and improved the docstrings for the most key functions.

- Tests
  - [ ] TODO (before merge)
  - [ x ] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ x ] Included
  - [  ] Not necessary because: 
- Public documentation
  - [ ] TODO (before merge)
  - [ x ] Not necessary because: No user facing changes


packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -1,9 +1,9 @@
 import { existsSync } from "fs";
 import { cp } from "fs/promises";
 import { join } from "path";
-import { retry } from "helpers/command";
-import { sleep } from "helpers/common";
-import { detectPackageManager } from "helpers/packages";
+import { detectPackageManager } from "helpers/packageManagers";
+import { retry } from "helpers/retry";
+import { sleep } from "helpers/sleep";
```
dario-piotrowicz, line 6: This is a generic comment for this and all other files, but wouldn't it be nicer to have a "barrel" (`index.ts`) file in `helpers` that exports from all the other files?

So that here and in all the other consumers you could just have:
```diff
- import { detectPackageManager } from "helpers/packageManagers";
- import { retry } from "helpers/retry";
- import { sleep } from "helpers/sleep";
+ import { detectPackageManager, retry, sleep } from "helpers";
```

dario-piotrowicz (approved):
LGTM 🙂 

except the small stuff I commented on

Also, you've commented in the PR's description:
> almost no behavioral changes to c3 (a few corner cases discovered when adding unit tests were fixed)
ideally if there is even a slight behavioral change it could still be worth/proper to add a changeset for it?

packages/create-cloudflare/src/wrangler/__tests__/accounts.test.ts
```diff
@@ -0,0 +1,169 @@
...
+			const mock = vi
+				.mocked(runCommand)
+				.mockReturnValueOnce(Promise.resolve(loggedInWhoamiOutput));
+
+			const loggedIn = await wranglerLogin();
+
+			expect(loggedIn).toBe(true);
+			expect(mock).toHaveBeenCalledWith(
+				["npx", "wrangler", "whoami"],
+				expect.anything()
```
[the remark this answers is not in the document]
jculvey: This is because each call to `runCommand` will also include an options object with things like `env` and `silent` which we don't care about so much. In many cases we just want to validate whether a command with the correct arguments were called.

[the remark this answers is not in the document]
dario-piotrowicz: ah! duh yeah :sweat_smile:

sorry my bad, I misread the code and thought this was for the return value :sweat: :bow:

packages/create-cloudflare/src/git.ts
```diff
@@ -0,0 +1,249 @@
...
+	} finally {
+		s.stop(`${brandColor("initialized")} ${dim(`git`)}`);
+	}
+}
+
+export async function getProductionBranch(cwd: string) {
+	try {
+		const productionBranch = await runCommand(
+			// "git branch --show-current", // git@^2.22
+			["git", "rev-parse", "--abbrev-ref", "HEAD"], // git@^1.6.3
```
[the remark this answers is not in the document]
jculvey: I think they're intentional, I didn't introduce them here they just came over in the move.

After looking at it, 2.22 was released in early 2019, and I think that's long enough to warrant just switching to that and delete the comment.

packages/create-cloudflare/e2e-tests/frameworks.test.ts
jculvey: I like this idea because it's great in dev, but this causes a huge problem with mocking. 

The mocks currently `vi.mock('helpers/command'` and similar, and if I combine them all into a single import then we lose the ability to mock some of those helpers but not others.

dario-piotrowicz: yeah.... I know that splitting the utils into multiple files was one of the main points here

I didn't consider that using a barrel would defeat the purpose here 😓

it is annoying.... but if that's what vitest forces you to do I guess there's no much we can do about it 😓

dario-piotrowicz: yeah.... I know that splitting the utils into multiple files was one of the main points here

I didn't consider that using a barrel would defeat the purpose here 😓

it is annoying.... but if that's what vitest forces you to do I guess there's no much we can do about it 😓

PR #5225, opened 2024-03-11
Restrict renaming dispatch namespaces

matthewdavidrodgers:
For WC-2014

Workers for Platforms metrics are currently tied to the name of the dispatch namespace, and so changing that unexpectedly changes the metrics of your invocations. So for now, we want to restrict updates to namespaces, making the name stable.

While this could be considered breaking, it doesn't seem that this command has ever been used - only a select few dispatch namespaces have ever been updated through the API.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [X] Not necessary because: functionality removed
  - [ ] TODO (before merge)
  - [X] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [X] Cloudflare docs PR(s): [removing command docs pr](https://github.com/cloudflare/cloudflare-docs/pull/13369)
  - [ ] Not necessary because:


packages/wrangler/src/dispatch-namespace.ts
```diff
@@ -181,31 +181,5 @@ export function workerNamespaceCommands(
...
-						describe: "New name of the dispatch namespace",
-						type: "string",
-						demandOption: true,
-					});
-			},
-			async (args) => {
-				await printWranglerBanner();
-				const config = readConfig(args.config, args);
-				const accountId = await requireAuth(config);
-				await renameWorkerNamespace(accountId, args.oldName, args.newName);
```
rozenmd, line 205: Did you need to remove this function too?

matthewdavidrodgers: correct, thank you. removed

matthewdavidrodgers:
if we have a problem with removing this command, we can keep the command in place but return a helpful error when it's used

.changeset/lemon-vans-film.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": minor
```
penalosa, line 2: This is a breaking change, and would have to be a `major` bump.

We can probably bundle this into v4, but in the meantime could you update this PR to deprecate this command instead of removing it (i.e. preserve the functionality but print a warning)

penalosa (changes requested):

matthewdavidrodgers:
I'm going to decline/close this - instead of removing this command we're going to continue allowing it, but add a warning about its effect on metrics (until we improve that situation)

PR #5226, opened 2024-03-11
Remove second wrangler banner from namespace rename

DaniFoldi:

After noticing #5225 and definitely not renaming a dispatch namespace a few times to see if the team notices it, I found that the banner was being printed twice in that command.

While the linked PR removes the yargs handler, it leaves the function in place, so I thought I'd fix it to aid a potential reintroduction in the future by ensuring the consistency between these functions.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Banners are excluded from the snapshot the tests are checked against
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: small banner fix

### DIFF
--- a/.changeset/young-chefs-change.md
+++ b/.changeset/young-chefs-change.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix(wrangler): Remove second banner from `wrangler dispatch-namespace rename`
--- a/.changeset/young-chefs-change.md
+++ b/.changeset/young-chefs-change.md
@@ -2,4 +2,4 @@
 "wrangler": patch
 ---
 
-fix(wrangler): Remove second banner from `wrangler dispatch-namespace rename`
+fix: remove second Wrangler banner from `wrangler dispatch-namespace rename`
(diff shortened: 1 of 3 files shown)


rozenmd (approved):

petebacondarwin (changes requested):
The fact that there needs to be no change to a test for this to pass CI makes me worried about the test coverage here.
Could we add tests for this command? It should be quite straightforward, mocking out the fetch request expectation.

nora-soderlund:
is a community PR the best place to add _missing_ test coverage though? feels like something that should go to the employees backlog.

petebacondarwin:
> is a community PR the best place to add _missing_ test coverage though? feels like something that should go to the employees backlog.

Good call @nora-soderlund - I didn't notice that @DaniFoldi is not an employee. I totally don't expect him to backfill all our tests. But equally, this PR is fixing something that is not right in our codebase, so a test to prove it would be appreciated.

DaniFoldi:
Hey @petebacondarwin 👋 

I had a look at this, however it looks like it would require significant time to test that without having to change most tests.
If my investigation is correct, the header is disabled in tests to avoid tests failing every time when it changes (new version available, etc).
I think it should be possible to mock the header function and count its number of invocations. Failing on the second one could be possible by throwing, however detecting if a command is missing the header (r2 commands seem to have no header, for example) I'm not even sure how I'd approach.

Feel free to close this PR, or merge as-is, and the header stuff could be tracked in a separate issue.

petebacondarwin (approved):
I updated a test to prove this fix.

PR #5229, opened 2024-03-12
add `bindingNames` to `getPlatformProxy` result

### LINKED ISSUE #4792: 🚀 Feature Request [getBindingsProxy]: add a convenient way to get hold of bindings of a specific types

dario-piotrowicz:

This PR adds a new `bindingNames` field in the `getPlatformProxy` result that users can use to get the names of their bindings grouped by type.

For example they can use it to get all the bindings of a specific type in the following way:
```ts
const { env, bindingNames } = getPlatformProxy();

const allKvNames = bindingNames.kv; // ['MY_KV', 'MY_KV_1']
const allR2Names = bindingNames.r2; // ['MY_R2']
// ...

const extract = (names; string[]) => Object.entries(env).map(
   [key, binding]) => names.includes[key] ? binding ; null
).filter(Boolean);
const allKvs = extract(allKvNames);
const allR2s = extract(allR2Names);
// ...
```

Notes:
 - this uses duck typing checks to discern what a binding type is, this checking is used only for the `getPlatformProxy` utility, but in the future if beneficial it could be exposed to users as a generic (also usable in workerd) utility?
 - another check that could be performed is using the Miniflare custom inspector like so:
   ```ts
   import { inspect, } from "node:util";

   const inspectStr = inspect(env.MY_KV);
   const match = inspectStr.match(/ProxyStub \{ name: '(.*?)', poisoned: (?:false|true) \}/);
   const name = match[1]; // 'KvNamespace'
   ```
   this would be much simpler but much more brittle and it would only work with the Miniflare magic proxies
 - I implemented `bindingNames` which only returns the names because I found this to be the simplest/cleanest
   way not to have issues with types, alternatively I could have had something like `bindingsByType` or `groupedEnv`,
   etc... but with such solutions they binding types would come from a specific `@cloudflare/workers-types` entrypoint
   which might not be what the user wants/expects...

fixes #4792 

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Not necessary because:


alexanderniebuhr:
@dario-piotrowicz thanks for letting me know of this PR.

Just want to give some insights from Astro's side. We would love to see this introduced to the API. We are currently building a DevToolbarApp, which will allow users to interact with the bindings directly from the browser without going back to the terminal, this is a significant improvement in DX.

Currently there is no way for us to know which binding is of which type. We would love to know which bindings are of a specific type.
We tried to workaround this and found some manually way for the user to type, manually reading the `wrangler.toml` or using the inspector check, mentioned above. - All of those are either not very bullet proof or require a lot of custom logic, which was the main reason for us to adopt `getPlatformProxy` to get rid of custom implementations.

We also believe this feature could be useful for other frameworks and users, even if they don't have it requested yet.

I'm happy to discuss this further in any of our dedicated communication channels, just let me know!

PR #5234, opened 2024-03-12
[WIP] feat(wrangler): Implement environment inheritance for Pages

CarmenPopoviciu:

coming soon...

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:


packages/wrangler/src/config/validation.ts
```diff
@@ -155,6 +159,18 @@ export function normalizeAndValidateConfig(
 				rawConfig
 			);
 			diagnostics.addChild(envDiagnostics);
+		} else if (isPagesConfig(rawConfig)) {
```
petebacondarwin, line 162: Do we need to run the `normalizeAndValidateEnvironment()` in this case at all?
We already have all the configuration from the top level validation above.

How about instead, on line [174](https://github.com/cloudflare/workers-sdk/pull/5234/files#diff-0ec779a40744e36b45a9b28dee06739afa8aa2f00bb9388b68556e7fd87ab99eR174) we just change it to 

```
} else if (!isPagesConfig(rawConfig)) {
```

packages/wrangler/src/config/validation.ts
```diff
@@ -161,10 +165,21 @@ export function normalizeAndValidateConfig(
...
 			diagnostics.addChild(envDiagnostics);
-		} else {
-			// An environment was specified, but no configuration for it was found.
-			// To cover any legacy environment cases, where the `envName` is used,
-			// Let's create a fake active environment with the specified `envName`.
+		} else if (!isPagesConfig(rawConfig)) {
+			/**
+			 * An environment was specified, but no configuration for it was found.
+			 *
+			 * ========== PAGES =========================================================
```
CarmenPopoviciu, line 172: // TODO @CarmenPopoviciu move comment to somewhere more appropriate

packages/wrangler/src/__tests__/configuration.pages.test.ts
```diff
@@ -0,0 +1,846 @@
...
+					pagesRawConfig,
+					undefined,
+					{
+						env: undefined,
+					}
+				);
+
+				expect(diagnostics.renderWarnings()).toMatchInlineSnapshot(`
+			"Processing wrangler configuration:
+			  - Hyperdrive Bindings are currently in beta to allow the API to evolve before general availability.
```
CarmenPopoviciu, line 36: is there a way to silence these Hyperdrive-relate warnings in tests?

packages/wrangler/src/__tests__/configuration.pages.test.ts
```diff
@@ -0,0 +1,846 @@
...
+								binding: "TEST_AED_BINDING",
+							},
+						],
+						ai: {
+							binding: "TEST_AI_BINDING",
+						},
+					})
+				);
+
+				/*======================================*/
```
CarmenPopoviciu, line 734: just calling this test out to highlight that this happens because we do the Pages-specific validation **after* `normalizeAndValidateConfig()`

CarmenPopoviciu: // TODO @CarmenPopoviciu add a comment to explain the above, if we decide to keep this test as part of the suite

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: ✅

packages/wrangler/src/__tests__/configuration.pages.test.ts
```diff
@@ -0,0 +1,846 @@
+import { normalizeAndValidateConfig } from "../config/validation";
+import {
+	generateRawConfigForPages,
+	generateRawEnvConfigForPages,
+} from "./helpers/generate-wrangler-config";
+import type { RawConfig, RawEnvironment } from "../config";
+
+describe("normalizeAndValidateConfig()", () => {
```
CarmenPopoviciu, line 8: are we OK with these tests living in a separate file? it's rather unconventional to test the same fn over 2 separate test suites, but:
-`configuration.test.ts` is already 5k lines long so had mixed feelings abt adding more to it.
- this made it easy to write the tests
- this way we keep Pages specific tests in one place. Granted we could achieve that via a dedicated `describe` block inside `configuration.test.ts`

I have no strong opinions, but if anyone else does, happy to readjust.

packages/wrangler/src/__tests__/configuration.pages.test.ts
CarmenPopoviciu: comments added ✅

packages/wrangler/src/__tests__/helpers/generate-wrangler-config.ts
```diff
@@ -0,0 +1,200 @@
...
+		name: projectName,
+		compatibility_date: "2024-01-01",
+		compatibility_flags: ["COMPATIBILITY_FLAG"],
+		limits: { cpu_ms: 100 },
+		placement: { mode: "smart" },
+
+		/** NON-INHERITABLE ENVIRONMENT FIELDS **/
+		vars: { VAR: "TEST_VAR" },
+		durable_objects: {
+			bindings: [{ name: "TEST_DO_BINDING", class_name: "TEST_DO_CLASS" }],
```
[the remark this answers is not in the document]
CarmenPopoviciu: the `script_name` seems to be optional? Do we want to add it to make this test binding more feature complete?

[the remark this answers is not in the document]
penalosa: It's optional in general, but it should be required for Pages, since Pages doesn't support DOs defined within the same worker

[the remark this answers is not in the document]
CarmenPopoviciu: oh, nvm. I understand why: https://github.com/cloudflare/workers-sdk/pull/5166#discussion_r1528513794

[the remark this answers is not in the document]
CarmenPopoviciu: ✅

packages/wrangler/src/__tests__/configuration.pages.test.ts
CarmenPopoviciu: ✅

packages/wrangler/src/__tests__/configuration.pages.test.ts
```diff
@@ -0,0 +1,785 @@
...
+					})
+				);
+
+				/*======================================*/
+				/*      Pages unsupported env name      */
+				/*======================================*/
+				/**
+				 * While Pages config only supports "preview" & "production" named
+				 * environments, the following is a valid test case, because all
+				 * ages-specific validation (incl. env naming) is performed after
```
penalosa, line 342:
```suggestion
				 * Pages-specific validation (incl. env naming) is performed after
```

packages/wrangler/src/__tests__/configuration.pages.test.ts
```diff
@@ -0,0 +1,785 @@
...
+							},
+						],
+						ai: {
+							binding: "TEST_AI_BINDING",
+						},
+					})
+				);
+
+				/*========================*/
+				/*      "production"      */
```
penalosa, line 236: Is there a reason these are in one test case, rather than a test case per environment?

packages/wrangler/src/__tests__/configuration.pages.test.ts
```diff
@@ -0,0 +1,785 @@
+import path from "node:path";
+import { normalizeAndValidateConfig } from "../config/validation";
+import {
+	generateRawConfigForPages,
+	generateRawEnvConfigForPages,
+} from "./helpers/generate-wrangler-config";
+import type { RawConfig, RawEnvironment } from "../config";
+
+describe("normalizeAndValidateConfig()", () => {
+	describe("Pages configuration", () => {
```
penalosa, line 10: Is this `describe()` necessary? It seems like `
describe("normalizeAndValidateConfig()")` has no other children

penalosa (approved):
A few comments, but overall looks good!

packages/wrangler/src/__tests__/configuration.pages.test.ts
CarmenPopoviciu: ✅

packages/wrangler/src/__tests__/configuration.pages.test.ts
CarmenPopoviciu: it's not necessary per se, it's merely to have some logical grouping in place, and separate the Pages test suite from the suite in `configuration.test.ts`. We already have a `describe("normalizeAndValidateConfig()",...)` [block](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/__tests__/configuration.test.ts#L11) in place in `configuration.test.ts`, and since we're duplicating that suite name here, having an extra grouping layer felt like a good idea. If this feels like overkill, we could rename to top level grouping to smth like `normalizeAndValidateConfig() for Pages` and get rid of the secondary grouping layer

packages/wrangler/src/__tests__/configuration.pages.test.ts
CarmenPopoviciu: I had these separated at first, but as I was writing them, felt like a bit of overkill. Since these tests have basically the same setup, and test for the same thing, with the only diff being the env name, I figured this grouping would make more sense. But perhaps we lost some readability and tidiness in the process. I'll regroup

CarmenPopoviciu: ✅

PR #5238, opened 2024-03-13
fix: move annotations under metadata in create-worker-upload-form

RamIdeas:

This fixes annotations in versions upload to match where the api expects to find them

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: no existing tests – will add in follow-up
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: experimental

### DIFF
--- a/packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
+++ b/packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
@@ -95,7 +95,8 @@ export type WorkerMetadataBinding =
 			destination: string;
 	  };
 
-export interface WorkerMetadata {
+// for PUT /accounts/:accountId/workers/scripts/:scriptName
+export type WorkerMetadataPut = {
 	/** The name of the entry point module. Only exists when the worker is in the ES module format */
 	main_module?: string;
 	/** The name of the entry point module. Only exists when the worker is in the service-worker format */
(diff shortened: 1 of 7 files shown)


packages/wrangler/src/versions/upload.ts
```diff
@@ -384,6 +384,11 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			placement,
 			tail_consumers: config.tail_consumers,
 			limits: config.limits,
+			annotations: {
+				"workers/message": props.message,
+				"workers/tag": props.tag,
+				"workers/triggered_by": "wrangler",
```
taylorlee, line 390: Can we remove triggered_by while we're here? As this will no longer be overridable soon.

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -111,6 +111,7 @@ export interface WorkerMetadata {
 	placement?: CfPlacement;
 	tail_consumers?: CfTailConsumer[];
 	limits?: CfUserLimits;
+	annotations?: Record<string, string>;
```
taylorlee, line 114: Since this is the single field that's not supported in script upload, but supported in version upload, is it worth calling that out in a comment? Is it fairly awkward to split the types into the set of fields supported by either API?

taylorlee (approved):

packages/wrangler/src/versions/upload.ts
```diff
@@ -422,15 +427,6 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			try {
 				const body = createWorkerUploadForm(worker);
 
-				body.set(
```
penalosa, line 425: Is this a breaking change for current users?

taylorlee: It's a bug fix, so it changes behavior that message was ignored and now it's actually set. But version-upload command is gated both by the `--experimental-gradual-rollouts` wrangler flag as well at gated at the server level, so we're ok to make breaking changes here if needed currently.

RamIdeas: It's not a breaking change. The API wasn't expecting annotations there but under metadata instead – so it's a fix.

petebacondarwin (approved):

petebacondarwin:
Snapshots need updating to accommodate the extra empty line that was added.

packages/wrangler/src/versions/upload.ts
penalosa: Ah, I didn't realise this wasn't in the wrangler-wide form data creation

petebacondarwin:
I added an issue to track adding the missing tests https://github.com/cloudflare/workers-sdk/issues/5254

PR #5243, opened 2024-03-14
ci: add dependabot configuration to handle miniflare workerd dependency updates

petebacondarwin:

Sets up Dependabot to autogenerate PRs on every update to workerd.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: CI configuration change only
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: No public facing change
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: No public facing change


petebacondarwin:
> Do we need to modify https://github.com/cloudflare/workers-sdk/blob/main/.github/workflows/c3-dependabot-versioning-prs.yml to generate changesets for these PRs too?

Good call. I forgot we need changesets to trigger a release of Miniflare.

.github/workflows/c3-dependabot-versioning-prs.yml
mrbbot, line 1: Should this file be renamed to something more generic? Does the `workflows` `README` need updating too?

tools/dependabot/generate-dependabot-pr-changesets.ts
```diff
@@ -0,0 +1,164 @@
...
+		"${packageName}": patch
+		---
+		`;
+}
+
+export function generateCommitMessage(
+	packageName: string,
+	changes: Map<string, Change>
+): string {
+	const t = new CLITable({
```
mrbbot, line 111: Do we not want to use Markdown tables like the previous script here? A CLI table would rely on monospace formatting which is not guaranteed when rendering markdown.

mrbbot (commented):
Nice! A few questions, but looks good! ✅

.github/workflows/c3-dependabot-versioning-prs.yml
mrbbot: Should the workflow be renamed too, and the `pull_request_target` `paths` be updated too.

petebacondarwin: Yes and yes. Sorry - brain melt there.

tools/dependabot/generate-dependabot-pr-changesets.ts
petebacondarwin: Ah good call. I didn't realised that it was a markdown table before. I just assumed it was poorly formatted 😄 .

.github/workflows/dependabot-versioning-prs.yml
```diff
@@ -1,11 +1,12 @@
 # workflow to update the framework cli versioning PRs we get
 # from dependabot for C3
 
-name: "C3: Generate changeset for dependabot PRs"
+name: "Generate changesets for dependabot PRs"
 on:
   pull_request_target:
     paths:
       - "packages/create-cloudflare/src/frameworks/package.json"
+      - "packages/miniflare/package.json"
```
jculvey, line 9: I wonder if this is what we want, or if we actually want separate workflows for miniflare and c3 each with their respective triggers. I'm thinking about the case where one package has changes but the other doesn't.

It seems like `generate-dependabot-pr-changesets.ts` will tolerate being called on a target even if there are no changes (it will just throw a warning saying so)

jculvey (commented):
Nice tests!

Left a comment about workflow organization but other than that lgtm. I don't expect this would break anything on the c3 side (but it's kinda hard to tell until it's run once)

jculvey (approved):

.github/workflows/dependabot-versioning-prs.yml
petebacondarwin: Hmm. Good point. Let's split them into two but keep the shared helper.

mrbbot (approved):

PR #5250, opened 2024-03-14
c3: update DO example

elithrar:

Makes it clearer what a "Durable Object" is. A user using c3 or Cloudflare for the first time will have no idea. I have also moved it down the list so that "web application" (a far more common use-case) is higher. 

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Human language changes only
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Human language changes only

### DIFF
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -143,14 +143,14 @@ export const getTemplateMap = async () => {
 		"hello-world": (await import("../templates/hello-world/c3")).default,
 		"hello-world-python": (await import("../templates/hello-world-python/c3"))
 			.default,
-		"hello-world-durable-object": (
-			await import("../templates/hello-world-durable-object/c3")
-		).default,
 		// Dummy record -- actual template config resolved in `selectFramework`
 		"web-framework": { displayName: "Website or web app" } as TemplateConfig,
 		common: (await import("../templates/common/c3")).default,
 		scheduled: (await import("../templates/scheduled/c3")).default,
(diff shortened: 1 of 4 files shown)


petebacondarwin (approved):
Assuming that the CI tests are not broken by this change, which they might as some have to use "down arrow" key presses to get to the correct item in the list when testing the interactive flow.

petebacondarwin:
Oh and this file needs reformatting:  packages/create-cloudflare/templates/hello-world-durable-object/c3.ts

packages/create-cloudflare/src/templates.ts
```diff
@@ -143,14 +143,14 @@ export const getTemplateMap = async () => {
...
-			await import("../templates/hello-world-durable-object/c3")
-		).default,
 		// Dummy record -- actual template config resolved in `selectFramework`
 		"web-framework": { displayName: "Website or web app" } as TemplateConfig,
 		common: (await import("../templates/common/c3")).default,
 		scheduled: (await import("../templates/scheduled/c3")).default,
 		queues: (await import("../templates/queues/c3")).default,
+		"hello-world-durable-object": (
+			await import("../templates/hello-world-durable-object/c3")
+		).default,
```
petebacondarwin, line 153: Moving this is likely to have broken the C3 e2e tests that rely upon the ordering...

petebacondarwin (changes requested):

petebacondarwin (approved):

PR #5258, opened 2024-03-14
wrangler: Try to URL decode hyperdrive connection string components before sending them to API

OilyLime:

Fixes SQC-153
LDW CR-851053

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Does not change the Hyperdrive command or its options.

### DIFF
--- a/packages/wrangler/src/hyperdrive/update.ts
+++ b/packages/wrangler/src/hyperdrive/update.ts
@@ -1,13 +1,13 @@
 import { readConfig } from "../config";
 import { UserError } from "../errors";
 import { logger } from "../logger";
+import { patchConfig } from "./client";
+import { hyperdriveBetaWarning } from "./utils";
 import type {
 	CommonYargsArgv,
 	StrictYargsOptionsToInterface,
 } from "../yargs-types";
 import type { PatchHyperdriveBody } from "./client";
-import { patchConfig } from "./client";
-import { hyperdriveBetaWarning } from "./utils";
 
 export function options(yargs: CommonYargsArgv) {
 	return yargs
--- a/packages/wrangler/src/hyperdrive/update.ts
+++ b/packages/wrangler/src/hyperdrive/update.ts
@@ -1,12 +1,13 @@
 import { readConfig } from "../config";
+import { UserError } from "../errors";
 import { logger } from "../logger";
-import { patchConfig } from "./client";
-import { hyperdriveBetaWarning } from "./utils";
 import type {
 	CommonYargsArgv,
 	StrictYargsOptionsToInterface,
 } from "../yargs-types";
 import type { PatchHyperdriveBody } from "./client";
+import { patchConfig } from "./client";
+import { hyperdriveBetaWarning } from "./utils";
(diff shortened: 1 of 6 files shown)


elithrar (approved):

packages/wrangler/src/hyperdrive/update.ts
```diff
@@ -78,17 +83,17 @@ export async function handler(
...
 	);
 
 	if (!allOriginFieldsSet && !noOriginFieldSet) {
 		throw new Error(
-			`When updating the origin, all of the following must be set: ${requiredOriginOptions.join(
-				", "
-			)}`
+			`When updating the origin, all of the following must be set: ${requiredOriginOptions
+				.map((option) => camelToKebab(option))
+				.join(", ")}`
```
mrbbot, line 96: Should there be tests for this error? I noticed you didn't need to update any snapshots for this change.

packages/wrangler/src/hyperdrive/update.ts
```diff
@@ -78,17 +83,17 @@ export async function handler(
...
-		isOptionSet(args, field as keyof typeof options)
+		isOptionSet(args, field)
 	);
 	const noOriginFieldSet = requiredOriginOptions.every(
-		(field) => !isOptionSet(args, field as keyof typeof options)
+		(field) => !isOptionSet(args, field)
 	);
 
 	if (!allOriginFieldsSet && !noOriginFieldSet) {
 		throw new Error(
```
mrbbot, line 93: So this doesn't prompt the user to report this to Sentry...
```suggestion
		throw new UserError(
```

packages/wrangler/src/hyperdrive/update.ts
OilyLime: There is a test: in `src/__tests__/hyperdrive.test.ts` -- `"should throw an exception when updating a hyperdrive config's origin but not all fields are set"`

packages/wrangler/src/hyperdrive/update.ts
OilyLime: Done.

mrbbot (approved):
Looks good! If this can wait until after the LDW, we can merge this then. Otherwise, you'll need a CR for this change.

OilyLime:
> It looks like you'll need to run `pnpm run prettify` too to format your code.

Ran it but no files were changed 🤷

penalosa:
> Ran it but no files were changed 🤷 

It looks like the specific file with issues is `packages/wrangler/src/hyperdrive/update.ts`—could you try formatting that directly?

OilyLime:
> > Ran it but no files were changed 🤷
> 
> It looks like the specific file with issues is `packages/wrangler/src/hyperdrive/update.ts`—could you try formatting that directly?

Still no dice.

Issue #5168, opened 2024-03-05
Remove ':' from secret:bulk command


RamIdeas:
### Describe the solution

Deprecate – but keep – the existing secret:bulk command

Rename to something consistent with other commands similar to https://github.com/cloudflare/workers-sdk/issues/5062

Issue #5165, opened 2024-03-05
Remove deprecated `wrangler version` command


RamIdeas:
https://github.com/cloudflare/workers-sdk/pull/5069 marked the `wrangler version` command as deprecated since `wrangler --version` is the recommended approach and having this command may cause confusions with the upcoming `wrangler versions ...` commands

cometkim:
Can it add an option to print only the version? It has too much unnecessary format around, making it difficult to use it with some scripts.

penalosa:
Closed by #7352

PR #5265, opened 2024-03-15
General v4 command changes

RamIdeas:

#5168 #5165

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/index.ts
+++ b/packages/wrangler/src/index.ts
@@ -387,6 +387,7 @@ export function createCLIParser(argv: string[]) {
 		}
 	);
 
+	// [DEPRECATED] secret:bulk
 	wrangler.command(
 		"secret:bulk [json]",
 		"🗄️  Bulk upload secrets for a Worker",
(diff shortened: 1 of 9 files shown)


packages/wrangler/src/index.ts
```diff
@@ -387,36 +388,64 @@ export function createCLIParser(argv: string[]) {
 		}
 	);
 
+	// [DEPRECATED] secret:bulk
```
mrbbot, line 391: Below your `[DEPRECATED]` comments show the new command that should be used, rather than the deprecated one? Should this be ```[DEPRECATED] Use `secret bulk` instead```?

packages/wrangler/src/index.ts
```diff
@@ -387,36 +388,64 @@ export function createCLIParser(argv: string[]) {
...
 	wrangler.command(
```
mrbbot, line 392: Can we hide the deprecated commands from the help output?

PR #5277, opened 2024-03-18
add `build-cf-types` script to Next.js projects and add testing for it

dario-piotrowicz:

This PR:
 - adds a `build-cf-types` script to Next.js projects
 - slightly simplified other projects' d.ts files and adds tests for them too

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: the script itself doesn't need to be documented and the d.ts files are already documented


packages/create-cloudflare/templates/next/env.d.ts
```diff
@@ -1,7 +1,5 @@
+// Generated by Wrangler
```
jculvey, line 1: Was the lack of this causing an initial build failure 😄 ?

jculvey (approved):

packages/create-cloudflare/templates/next/env.d.ts
dario-piotrowicz: it was causing the `build-cf-types` script to fail since that would refuse to override the file 🙂 

https://github.com/cloudflare/workers-sdk/blob/1720f0a12a6376093b3c5799d74f47c522ae8571/packages/wrangler/src/type-generation.ts#L331-L340

jculvey (approved):

PR #5282, opened 2024-03-18
wrangler: Add source map support for Workers

maxwellpeterson:
- Add `upload_source_maps` boolean config option
- Include source maps in Worker upload when enabled

Closes WP-821 and WP-822. Allows users to enable source map support for Workers when using Wrangler. For more context see "SPEC: Workers Source maps" on the internal wiki.

### How to Test

1. Ask @maxwellpeterson to ungate source maps support for your account.
2. Deploy a Worker with `upload_source_maps = true` using the pre-release build for this PR.
3. View uploaded contents using the [Download Worker](https://developers.cloudflare.com/api/operations/worker-script-download-worker) API endpoint.
4. Uploaded contents should include all source maps output by the build step, and each source map should include `sources` and `sourceRoot` values that match the project directory structure.

- Tests
  - [x] Included
  - [x] Included
- Public documentation
  - [x] Not necessary because: This feature is not publicly available yet.


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -965,6 +1001,88 @@ function deployWfpUserWorker(
...
+
+/**
+ * Find source maps by scanning module contents for special `//#
+ * sourceMappingURL=` comments pointing to source map files.
+ */
+function scanSourceMaps(modules: CfModule[]): CfWorkerSourceMap[] {
+	const maps: CfWorkerSourceMap[] = [];
+	for (const module of modules) {
+		const commentPrefix = "//# sourceMappingURL=";
+		const commentIndex = module.content.indexOf(commentPrefix);
```
zebp, line 1055: nit: we should probably just check last line for correctness, probably wont actually impact anyone but ya-know.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -965,6 +1001,88 @@ function deployWfpUserWorker(
...
+}
+
+/**
+ * Find source maps by scanning module contents for special `//#
+ * sourceMappingURL=` comments pointing to source map files.
+ */
+function scanSourceMaps(modules: CfModule[]): CfWorkerSourceMap[] {
+	const maps: CfWorkerSourceMap[] = [];
+	for (const module of modules) {
+		const commentPrefix = "//# sourceMappingURL=";
```
zebp, line 1054: There's a note in the sourcemap spec that we *might* want to accommodate, up to you though. I can't remember a time where I saw `//@`

> Note: The prefix for this annotation was initially “//@” however this conflicts with Internet Explorer’s Conditional Compilation and was changed to “//#”.  It is reasonable for tools to also accept “//@” but “//#” is preferred.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -965,6 +1001,88 @@ function deployWfpUserWorker(
...
+			throw new Error(
+				`Invalid source map path in ${module.filePath}: ${wranglerPath} does not exist.`
+			);
+		}
+		const map = JSON.parse(
+			readFileSync(wranglerPath).toString()
+		) as SourceMapV3;
+		// Overwrite the file property of the sourcemap to match the name of the
+		// corresponding module in the multipart upload.
+		map.file = module.name;
```
zebp, line 1077: Do we want to update the `sources` here too? When testing with a custom build command with esbuild (`pnpm esbuild --outdir=dist --sourcemap=external src/**/*.ts`) I ended up with entries in my sources that contained one or more `../`'s with paths like `../../src/foo/bar.ts`. We want to display these as `./src/foo/bar.ts` *one day* in the quick editor so we probably want to do that transformation here in wrangler. I guess a naive call to `substring` if the string starts with `../` would suffice?

zebp (approved):
Mostly LGTM, some concerns about scanning sourcemaps for custom builds.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -487,46 +497,52 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
-								props.compatibilityFlags ?? config.compatibility_flags
-							),
-						}
-				  );
+		const {
+			modules,
+			dependencies,
+			resolvedEntryPointPath,
+			bundleType,
+			...bundle
```
petebacondarwin, line 505:
```suggestion
			sourceMapPath,
			sourceMapMetadata,
```

This will save on destructuring bundle again further down and also makes it clear here what is actually being used from the result of bundling.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -635,17 +651,37 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
+		const main: CfModule = {
+			name: entryPointName,
+			filePath: resolvedEntryPointPath,
+			content: content,
+			type: bundleType,
+		};
+		let sourceMaps: CfWorkerSourceMap[] | undefined = undefined;
+		if (props.sourceMaps ?? config.source_maps) {
+			const { sourceMapPath, sourceMapMetadata } =
+				bundle as Partial<BundleResult>;
```
petebacondarwin, line 663: Can we just do the destructuring above?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -965,6 +1001,88 @@ function deployWfpUserWorker(
...
+		const commentIndex = module.content.indexOf(commentPrefix);
+		if (commentIndex === -1 || module.filePath === undefined) {
+			continue;
+		}
+		// Assume the source map path in the comment is relative to the
+		// generated file it appears in.
+		const commentPath = module.content
+			.slice(commentIndex + commentPrefix.length)
+			.toString()
+			.trim();
```
petebacondarwin, line 1064: Be aware that this `commentPath` can actually be any URL, including at data URL!
We might want a bit of URL parsing here and validation to ensure that we get what we expect.

packages/wrangler/src/deploy/deploy.ts
petebacondarwin: According to the spec I think that the plan is only to support inline content for the sources?

packages/wrangler/src/deploy/deploy.ts
maxwellpeterson: Since this change was made over 10 years ago, I feel fine ignoring `//@`

packages/wrangler/src/config/environment.ts
```diff
@@ -302,6 +302,12 @@ interface EnvironmentInheritable {
 	 */
 	logpush: boolean | undefined;
 
+	/**
+	 * Include source maps when uploading this worker.
+	 * @inheritable
+	 */
+	source_maps: boolean | undefined;
```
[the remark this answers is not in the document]
maxwellpeterson: Changed to `upload_source_maps` in fixup commit

maxwellpeterson:
Pushed a fixup commit that addresses feedback and adds the config option to the `versions upload` command

packages/wrangler/src/deploy/deploy.ts
maxwellpeterson: @zebp Pushed a change to clean up the `sources` array as you suggested. The custom build part of this PR still feels pretty brittle, and I'm not sure how much we can do about it.

@petebacondarwin The plan was to only to support ["linked"](https://esbuild.github.io/api/#sourcemap) source maps. We could support inline as well but I think we want to deter people from unnecessarily increasing their JS bundle size.

packages/wrangler/src/deploy/index.ts
```diff
@@ -191,6 +191,10 @@ export function deployOptions(yargs: CommonYargsArgv) {
 				describe:
 					"Send Trace Events from this worker to Workers Logpush.\nThis will not configure a corresponding Logpush job automatically.",
 			})
+			.option("source-maps", {
+				type: "boolean",
+				describe: "Include source maps when uploading this worker.",
```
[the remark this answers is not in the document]
maxwellpeterson: Yes, added in fixup commit

packages/wrangler/src/__tests__/helpers/mock-upload-worker.ts
```diff
@@ -15,7 +15,7 @@ export function mockUploadWorkerRequest(
 		expectedMainModule?: string;
 		expectedType?: "esm" | "sw";
 		expectedBindings?: unknown;
-		expectedModules?: Record<string, string>;
+		expectedModules?: Record<string, string | null>;
```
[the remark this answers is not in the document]
maxwellpeterson: I wanted to test that source maps were not included in upload when the config flag was disabled. Previously there was no way to assert the absence of a file from the upload form.

packages/wrangler/src/deploy/deploy.ts
maxwellpeterson: Added a basic check that the URL doesn't start with `data:` since we don't want inline content

packages/wrangler/src/deploy/deploy.ts
maxwellpeterson: This ended up getting refactored into a separate function

packages/wrangler/src/deploy/deploy.ts
maxwellpeterson: This was a bit of a type hack since the return type of `noBundleWorker` doesn't include the two source map fields (so they can't be destructured here). The type of `bundle` is `{}` which is then assignable to `Partial<BundleResult>` later on.

packages/wrangler/src/deployment-bundle/source-maps.ts
```diff
@@ -0,0 +1,136 @@
...
+function loadSourceMap(
+	{ name, filePath }: CfModule,
+	sourceMapPath: string,
+	{ entryDirectory }: SourceMapMetadata
+): CfWorkerSourceMap[] {
+	if (filePath === undefined) {
+		return [];
+	}
+	const map = JSON.parse(
+		fs.readFileSync(sourceMapPath).toString()
```
petebacondarwin, line 49:
```suggestion
		fs.readFileSync(sourceMapPath, "utf8")
```

packages/wrangler/src/deployment-bundle/source-maps.ts
```diff
@@ -0,0 +1,136 @@
...
+		// the source map is located.
+		return loadSourceMap(main, sourceMapPath, sourceMapMetadata);
+	} else {
+		// Don't know where source maps are located, so need to find
+		// them by scanning the contents of the user-specified modules.
+		return scanSourceMaps([main, ...modules]);
+	}
+}
+
+interface SourceMapV3 {
```
petebacondarwin, line 26: Elsewhere in Wrangler we actually use a library for this:

https://github.com/cloudflare/workers-sdk/blob/bd00dc827769145df8d42a677393173bd751e643/packages/wrangler/src/dev/inspect.ts#L14

packages/wrangler/src/deployment-bundle/source-maps.ts
```diff
@@ -0,0 +1,136 @@
...
+			module.filePath === undefined
+		) {
+			continue;
+		}
+		// Assume the source map path in the comment is relative to the
+		// generated file it appears in.
+		const commentPath = stripPrefix(commentPrefix, lastLine).trim();
+		if (commentPath.startsWith("data:")) {
+			throw new Error(
+				`Invalid source map path in ${module.filePath}: expected file path but found data URL.`
```
petebacondarwin, line 93: Strictly this is not invalid, it is just unsupported by this code.

packages/wrangler/src/deployment-bundle/source-maps.ts
```diff
@@ -0,0 +1,136 @@
...
+		const map = JSON.parse(
+			fs.readFileSync(wranglerPath).toString()
+		) as SourceMapV3;
+		// Overwrite the file property of the sourcemap to match the name of the
+		// corresponding module in the multipart upload.
+		map.file = module.name;
+		map.sourceRoot = cleanPathPrefix(map.sourceRoot);
+		map.sources = map.sources.map(cleanPathPrefix);
+		maps.push({
+			name: module.name + ".map",
```
petebacondarwin, line 112: If `module.name + ".map"` does not match `commentPath` then the downstream tooling might be unable to link the source to this source map.

packages/wrangler/src/deployment-bundle/source-maps.ts
```diff
@@ -0,0 +1,136 @@
...
+		});
+	}
+	return maps;
+}
+
+/**
+ * Removes leading "." and ".." segments from the given file path.
+ */
+function cleanPathPrefix(filePath: string): string {
+	return stripPrefix("..", stripPrefix("." + path.sep, filePath));
```
petebacondarwin, line 123: Should this be:

```suggestion
	return stripPrefix(".." + path.sep, stripPrefix("." + path.sep, filePath));
```

Moreover it is not guaranteed that these paths will have the same path separator as the current OS. I.e. on Windows `path.sep` will be `\` but these paths here might well have `/`.

I would be lenient and actually handle both `/` and `\` in thes path.

petebacondarwin: Similarly in the `withTrailingSep()` function below.

petebacondarwin (approved):
I have a few concerns that there are some corner cases that will end up with invalid source-maps that won't work well at runtime. But otherwise this looks fine.

Have you been able to test this works in actual life?

packages/wrangler/src/deployment-bundle/source-maps.ts
maxwellpeterson: Updated to use the type from this library

packages/wrangler/src/deployment-bundle/source-maps.ts
maxwellpeterson: Updated the error message to say "unsupported" instead of "invalid"

packages/wrangler/src/deployment-bundle/source-maps.ts
maxwellpeterson: By "downstream tooling" do you mean the runtime? We ignore both the `//# sourceMappingURL=` comment and the source map module name, and instead match the JS module name with the `file` property of the source map (which this PR overwrites to match the corresponding JS module name).

petebacondarwin: OK, that is slightly strange behaviour from the runtime.
But I assume that in the future there will be other consumers that will read the source file and then request the source map? Like quick editor dashboard?
Or is there going to be some special API for these consumers?

maxwellpeterson:
> Have you been able to test this works in actual life?

Yes, I tested it with several different Workers with benchmarking EWC upload performance. This was mostly using Wrangler's own bundling, and haven't done as much with custom build setups.

packages/wrangler/src/deployment-bundle/source-maps.ts
maxwellpeterson: The Worker source and source map are downloaded using the same endpoint and can be downloaded together in a single request. If clients want to correlate source files and source maps they can match them using the same logic that the runtime does (JS module name and source map `file` field). Or at least that was the plan.

petebacondarwin: Yeah but many SourceMap consumers will parse the source file and use the path in the comment if it is there.
They only fall back to appending the `.map` extension if there is no comment.
It might be better to strip out the comment if you don't expect it to be used.

Not a big issue but might be worth keeping an eye on as this starts to be used.

PR #5284, opened 2024-03-18
Config file support in `pages dev`

CarmenPopoviciu:

This PR adds `wrangler.toml` support in `pages dev`

♫♫♫
```
Everybody spread the word
We're gonna have a celebration
All across the world
In every nation
It's time for the good times
Forget about the bad times, oh yeah
One day to come together to release the pressure
We need a holiday
```
♫♫♫

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not required at this stage of the work

## Test Plan 🚀
Apart from unit/integration/fixtures and e2e tests, this PR was manually tested based on the following test plan:

- [x] **`pages dev` should pick up configuration from `wrangler.toml`, if file exists and `pages_build_output_dir` is specified in the configuration**
```
npx wrangler pages dev
```
Screenshot 2024-03-24 at 18 02 13

- [x] **`pages dev` should ignore `wrangler.toml`, if it exists but `pages_build_output_dir` is not specified in the configuration**
```
npx wrangler pages dev
```
Screenshot 2024-03-24 at 18 09 23

- [x] **`pages dev` should fail if `wrangler.toml` file is not valid**
```
# wrangler.toml

[env.staging]
compatibility_date = "2024-01-01"
```

```
npx wrangler pages dev
```
Screenshot 2024-03-24 at 18 12 52

- [x] **`pages dev` should apply top level non environment config specified in `wrangler.toml`**
- [x] **`pages dev` should apply top level local development config specified in `wrangler.toml`**
```
# wrangler.toml

[dev]
ip = "127.0.0.1"
port = 8181
inspector_port = 8989
local_protocol = "http"
upstream_protocol = "http"
```
```
npx wrangler pages dev
```
=== terminal ===
Screenshot 2024-03-24 at 18 19 00

=== browser ===
Screenshot 2024-03-24 at 18 19 26

=== dev-tools ===
Screenshot 2024-03-24 at 18 23 31

- [x] **`pages dev` should apply top level inheritable environment config specified in `wrangler.toml`**
```
# wrangler.toml

# test by commenting out `compatibility_date`
#compatibility_date = "2024-01-01"
```
```
npx wrangler pages dev
```

Screenshot 2024-03-24 at 18 28 41

- [x] **`pages dev` should apply top level non-inheritable environment config specified in `wrangler.toml`**
```
npx wrangler pages dev
```
Screenshot 2024-03-24 at 17 57 49

- [x] **`pages dev` should merge cmd line args with config from `wrangler.toml`**
```
npx wrangler pages dev 
  --kv kv_binding_args 
  --binding VAR3="var3" 
  --do do_binding_args=DO@a 
  --d1 d1_binding_args 
  --r2 r2_binding_args 
  --ai ai_binding_args 
  --service service_binding_args=worker-args
```

Screenshot 2024-03-24 at 18 50 38

- [x] **when a config field is specified in both `wrangler.toml` and as a command arg to `pages dev`, the arg value should take precedence and override the value specified in `wrangler.toml`**
```
npx wrangler pages dev
  --kv KV_BINDING_TOML=new_kv_id 
  --binding VAR1="new_var_1" 
  --do DO_BINDING_TOML=new_do@new_do_script 
  --d1 D1_BINDING_TOML=new_d1_binding 
  --r2 R2_BINDING_TOML=new_r2_binding 
  --ai ai_binding_args 
  --service SERVICE_BINDING_TOML=new_service_name
```
Screenshot 2024-03-24 at 18 48 23

- [x] **`pages dev` should not apply any environment (`preview` | `production`) configuration, if specified. It should always defer to the top level config.**
```
# wrangler.toml

## Preview Environment
[env.preview]
[[env.preview.kv_namespaces]]
binding = "KV_BINDING_PREVIEW_TOML"
id = "KV_ID_PREVIEW_TOML"

## Production Environment
[env.production]
[[env.production.kv_namespaces]]
binding = "KV_BINDING_PRODUCTION_TOML"
id = "KV_ID_PRODUCTION_TOML"
```
```
npx wrangler pages dev
```
Screenshot 2024-03-24 at 18 53 31

- [x] **`pages dev <directory>` should pick up `wrangler.toml` configuration, even if `pages_build_output_dir` is not specified**
```
# wrangler.toml
name = "pages-project"
compatibility_date = "2024-01-01"

[vars]
VAR1 = "holiday"
VAR2 = "celebrate"
```
```
npx wrangler pages dev public
```
Screenshot 2024-03-25 at 18 11 54

### DIFF
--- a/packages/wrangler/e2e/pages-dev.test.ts
+++ b/packages/wrangler/e2e/pages-dev.test.ts
@@ -575,4 +575,274 @@ describe("pages dev", () => {
 			);
 		});
 	});
+
+	it("should support wrangler.toml", async () => {
+		await runDevSession(workerPath, "", async (session) => {
+			await seed(workerPath, {
+				"public/_worker.js": dedent`
+						export default {
+							async fetch(request, env) {
+								return new Response("Pages supports wrangler.toml ⚡️⚡️")
+							}
+						}`,
+				"wrangler.toml": dedent`
+					name = "pages-project"
+					pages_build_output_dir = "public"
+					compatibility_date = "2023-01-01"
+				`,
+			});
+
+			const { text, stderr } = await retry(
+				(s) => s.status !== 200,
+				async () => {
+					const r = await fetch(`http://127.0.0.1:${session.port}`);
+					return {
+						text: await r.text(),
+						status: r.status,
+						stderr: session.stderr,
+					};
+				}
+			);
+			expect(normalizeOutput(stderr)).toMatchInlineSnapshot(`""`);
+			expect(text).toMatchInlineSnapshot(
+				'"Pages supports wrangler.toml ⚡️⚡️"'
+			);
+		});
+	});
+
+	it("should use top-level configuration specified in `wrangler.toml`", async () => {
+		await runDevSession(workerPath, "", async (session) => {
+			await seed(workerPath, {
+				"public/_worker.js": dedent`
+						export default {
+							async fetch(request, env) {
+								return new Response(env.PAGES + " " + "supports wrangler.toml")
+							}
+						}`,
+				"wrangler.toml": dedent`
+					name = "pages-project"
+					pages_build_output_dir = "public"
+					# commenting this out would result in a warning. If there is no "compatibility_date"
+					# related warning in stdout, then this value was picked up
+					compatibility_date = "2023-01-01"
+
+					[vars]
+					PAGES = "⚡️ Pages ⚡️"
+
+					[[kv_namespaces]]
+					binding = "KV_BINDING_TOML"
+					id = "KV_ID_TOML"
+				`,
+			});
+
+			const { text, stdout, stderr } = await retry(
+				(s) => s.status !== 200,
+				async () => {
+					const r = await fetch(`http://127.0.0.1:${session.port}`);
+					return {
+						text: await r.text(),
+						status: r.status,
+						stdout: session.stdout,
+						stderr: session.stderr,
+					};
+				}
+			);
+
+			expect(text).toMatchInlineSnapshot(
+				'"⚡️ Pages ⚡️ supports wrangler.toml"'
+			);
+			expect(normalizeOutput(stderr)).toMatchInlineSnapshot(`""`);
+			expect(normalizeOutput(stdout).replace(/\s/g, "")).toContain(
+				`
+					Your worker has access to the following bindings:
+					- KV Namespaces:
+						- KV_BINDING_TOML: KV_ID_TOML
+					- Vars:
+						- PAGES: "⚡️ Pages ⚡️"
+				`.replace(/\s/g, "")
+			);
+		});
+	});
+
+	it("should merge (with override) `wrangler.toml` configuration with configuration provided via the command line, with command line args taking precedence", async () => {
+		const flags = [
+			` --binding VAR1=NEW_VAR_1 VAR3=VAR_3_ARGS`,
+			` --kv KV_BINDING_1_TOML=NEW_KV_ID_1 KV_BINDING_3_ARGS=KV_ID_3_ARGS`,
+			` --do DO_BINDING_1_TOML=NEW_DO_1@NEW_DO_SCRIPT_1 DO_BINDING_3_ARGS=DO_3_ARGS@DO_SCRIPT_3_ARGS`,
+			` --d1 D1_BINDING_1_TOML=NEW_D1_NAME_1 D1_BINDING_3_ARGS=D1_NAME_3_ARGS`,
+			` --r2 R2_BINDING_1_TOML=NEW_R2_BUCKET_1 R2_BINDING_3_TOML=R2_BUCKET_3_ARGS`,
+			` --service SERVICE_BINDING_1_TOML=NEW_SERVICE_NAME_1 SERVICE_BINDING_3_TOML=SERVICE_NAME_3_ARGS`,
+			` --ai AI_BINDING_2_TOML`,
+		];
+		await runDevSession(workerPath, `${flags}`, async (session) => {
+			await seed(workerPath, {
+				"public/_worker.js": dedent`
+						export default {
+							async fetch(request, env) {
+								return new Response("Pages supports wrangler.toml ⚡️")
+							}
+						}`,
+				"wrangler.toml": dedent`
+					name = "pages-project"
+					pages_build_output_dir = "public"
+					compatibility_date = "2023-01-01"
+
+					[vars]
+					VAR1 = "VAR_1_TOML" # to override
+					VAR2 = "VAR_2_TOML" # to merge
+
+					# to override
+					[[kv_namespaces]]
+					binding = "KV_BINDING_1_TOML"
+					id = "KV_ID_1_TOML"
+
+					# to merge as is
+					[[kv_namespaces]]
+					binding = "KV_BINDING_2_TOML"
+					id = "KV_ID_2_TOML"
+
+					# to override
+					[[durable_objects.bindings]]
+					name = "DO_BINDING_1_TOML"
+					class_name = "DO_1_TOML"
+					script_name = "DO_SCRIPT_1_TOML"
+
+					# to merge as is
+					[[durable_objects.bindings]]
+					name = "DO_BINDING_2_TOML"
+					class_name = "DO_2_TOML"
+					script_name = "DO_SCRIPT_2_TOML"
+
+					# to override
+					[[d1_databases]]
+					binding = "D1_BINDING_1_TOML"
+					database_id = "D1_ID_1_TOML"
+					database_name = "D1_NAME_1_TOML"
+
+					# to merge as is
+					[[d1_databases]]
+					binding = "D1_BINDING_2_TOML"
+					database_id = "D1_ID_2_TOML"
+					database_name = "D1_NAME_2_TOML"
+
+					# to override
+					[[r2_buckets]]
+					binding = 'R2_BINDING_1_TOML'
+					bucket_name = 'R2_BUCKET_1_TOML'
+
+					# to merge as is
+					[[r2_buckets]]
+					binding = 'R2_BINDING_2_TOML'
+					bucket_name = 'R2_BUCKET_2_TOML'
+
+					# to override
+					[[services]]
+					binding = "SERVICE_BINDING_1_TOML"
+					service = "SERVICE_NAME_1_TOML"
+
+					# to merge as is
+					[[services]]
+					binding = "SERVICE_BINDING_2_TOML"
+					service = "SERVICE_NAME_2_TOML"
+
+					# to override
+					[ai]
+					binding = "AI_BINDING_1_TOML"
+				`,
+			});
+
+			const { text, stdout, stderr } = await retry(
+				(s) => s.status !== 200,
+				async () => {
+					const r = await fetch(`http://127.0.0.1:${session.port}`);
+					return {
+						text: await r.text(),
+						status: r.status,
+						stdout: session.stdout,
+						stderr: session.stderr,
+					};
+				}
+			);
+
+			expect(text).toMatchInlineSnapshot('"Pages supports wrangler.toml ⚡️"');
+			expect(normalizeOutput(stderr)).toMatchInlineSnapshot(`
+				"▲ [WARNING] WARNING: You have Durable Object bindings that are not defined locally in the worker being developed.
+				  Be aware that changes to the data stored in these Durable Objects will be permanent and affect the live instances.
+				  Remote Durable Objects that are affected:
+				  - {"name":"DO_BINDING_1_TOML","class_name":"DO_1_TOML","script_name":"DO_SCRIPT_1_TOML"}
+				  - {"name":"DO_BINDING_2_TOML","class_name":"DO_2_TOML","script_name":"DO_SCRIPT_2_TOML"}
+				▲ [WARNING] This worker is bound to live services: SERVICE_BINDING_1_TOML (SERVICE_NAME_1_TOML), SERVICE_BINDING_2_TOML (SERVICE_NAME_2_TOML)
+				▲ [WARNING] ⎔ Support for service bindings in local mode is experimental and may change.
+				▲ [WARNING] ⎔ Support for external Durable Objects in local mode is experimental and may change.
+				▲ [WARNING] Using Workers AI always accesses your Cloudflare account in order to run AI models, and so will incur usage charges even in local development."
+			`);
+			expect(normalizeOutput(stdout).replace(/\s/g, "")).toContain(
+				`
+				Your worker has access to the following bindings:
+				- Durable Objects:
+				  - DO_BINDING_1_TOML: NEW_DO_1 (defined in NEW_DO_SCRIPT_1)
+				  - DO_BINDING_2_TOML: DO_2_TOML (defined in DO_SCRIPT_2_TOML)
+				  - DO_BINDING_3_ARGS: DO_3_ARGS (defined in DO_SCRIPT_3_ARGS,)
+				- KV Namespaces:
+				  - KV_BINDING_1_TOML: NEW_KV_ID_1
+				  - KV_BINDING_2_TOML: KV_ID_2_TOML
+				  - KV_BINDING_3_ARGS: KV_ID_3_ARGS,
+				- D1 Databases:
+				  - D1_BINDING_1_TOML: local-D1_BINDING_1_TOML=NEW_D1_NAME_1 (NEW_D1_NAME_1)
+				  - D1_BINDING_2_TOML: D1_NAME_2_TOML (D1_ID_2_TOML)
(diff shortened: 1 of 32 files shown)


packages/wrangler/src/pages/dev.ts
```diff
@@ -913,3 +811,185 @@ async function spawnProxyProcess({
 
 	return port;
 }
+
+/**
+ *
+ */
+function resolvePagesDevServerSettings(
```
CarmenPopoviciu, line 818: in need of a better naming here 🤔 I *think*. Maybe `reconcilePagesDevServerSettings` or `mergePagesDevServerSettings`? Or is `resolvePagesDevServerSettings` descriptive enough?

packages/wrangler/src/dev.tsx
```diff
@@ -874,125 +878,160 @@ export function getBindings(
...
+			logger.log(
+				`Found a non-empty WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING variable for binding. Hyperdrive will connect to this database during local development.`
+			);
+			hyperdrive.localConnectionString = connectionStringFromEnv;
+		}
+
+		return hyperdrive;
+	});
+
+	// Queues bindings ??
```
CarmenPopoviciu, line 991: //todo @CarmenPopoviciu dbl check this is correct

packages/wrangler/src/dev.tsx
```diff
@@ -1003,3 +1042,26 @@ export function getBindings(
 
 	return bindings;
 }
+
+/**
+ * Merge two arrays of objects, overriding elements in `source` with
+ * elements from `override`, if they have the same `key` value
+ */
+function mergeWithOverride<T>(
```
CarmenPopoviciu, line 1050: we should place this fn in a `utils`-like folder/file, but can't tell what would be appropriate, or what the convention is for these things. Suggestions welcome. Will also add tests once we settle on where it should live ^.^

packages/wrangler/src/dev.tsx
```diff
@@ -1003,3 +1042,26 @@ export function getBindings(
...
+	source: Array<T>,
+	override: Array<T>,
+	key: keyof T
```
petebacondarwin, line 1053:
```suggestion
	keyProperty: keyof T
```

packages/wrangler/src/dev.tsx
```diff
@@ -1003,3 +1042,26 @@ export function getBindings(
...
+
+	// preserve everything in `override` as it takes precedence over `source`,
+	// and add remaining elements from `source` that were not overriden
+	override.forEach((el) => {
+		if (sourceMap.has(el[key])) {
+			sourceMap.delete(el[key]);
+		}
+	});
+
+	return [...override, ...sourceMap.values()];
```
petebacondarwin, line 1066: This seems a bit of a complicated approach. Would the following also work?

```suggestion
	override.forEach((el) => sourceMap.set(el[key], el));
	return Array.from(sourceMap.values());
```

packages/wrangler/src/pages/dev.ts
```diff
@@ -930,3 +830,203 @@ async function spawnProxyProcess({
...
+		inspectorPort: args.inspectorPort ?? config.dev?.inspector_port,
+		localProtocol: args.localProtocol ?? config.dev?.local_protocol,
+	};
+}
+
+/**
+ * Parses the arguments specified by `pages dev` for environment bindings, and converts them
+ * into arrays of appropriate environment binding structures
+ */
+function getBindings(args: PagesDevArguments): Partial<
```
petebacondarwin, line 873: I was thrown by the fact that this has the same function name as the one used by the normal dev command. This is really converting the command line args into override bindings that will be passed through to unstable_dev to be merged with those parsed from the wrangler.toml config, right?

Could we give this a bit more of an explicit name to call that out? E.g. `getBindingsFromArgs()` or `getOverrideBindings()` or something like that.

packages/wrangler/src/pages/dev.ts
```diff
@@ -229,99 +236,86 @@ export function Options(yargs: CommonYargsArgv) {
...
+		services,
+		ai,
+	} = getBindings(args);
+
+	if (services && services.find(({ environment }) => !!environment)) {
+		// We haven't yet properly defined how environments of service bindings should
+		// work, so if the user is using an environment for any of their service
+		// bindings we warn them that they are experimental
+		logger.warn("Support for service binding environments is experimental.");
 	}
```
petebacondarwin, line 311: I think it would be nice to move this to the `getBindings()` function.

packages/wrangler/src/pages/dev.ts
```diff
@@ -930,3 +830,203 @@ async function spawnProxyProcess({
 
 	return port;
 }
+
+/**
+ * Reconciles top-level & local development settings, with `pages dev`command line args taking
+ * precedence over `wrangler.toml` configuration
```
petebacondarwin, line 836: Might be worth noting that this does not deal with bindings.

petebacondarwin:
Regarding this test case:

- pages dev should ignore wrangler.toml, if it exists but pages_build_output_dir is not specified in the configuration

Should we consider outputting a warning to the user, in case they were actually trying to create a toml for Pages and just missed that field?

packages/wrangler/src/dev.tsx
```diff
@@ -874,125 +878,160 @@ export function getBindings(
...
-						// to create a preview namespace for the user automatically
-						throw new UserError(
-							`In development, you should use a separate kv namespace than the one you'd use in production. Please create a new kv namespace with "wrangler kv:namespace create <name> --preview" and add its id as preview_id to the kv_namespace "${binding}" in your wrangler.toml`
-						); // Ugh, I really don't like this message very much
-					}
-					return {
-						binding,
-						id: preview_id ?? id,
-					};
+	/**
```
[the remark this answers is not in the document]
CarmenPopoviciu: this was totally by accident. will fix

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -574,4 +574,269 @@ describe("pages dev", () => {
...
+	// 			(s) => s.status !== 200,
+	// 			async () => {
+	// 				const r = await fetch(`https://127.0.0.1:${session.port}`);
+	// 				return { text: await r.text(), status: r.status, stderr: session.stderr };
+	// 			}
+	// 		);
+	// 		expect(normalizeOutput(stderr)).toMatchInlineSnapshot(`""`);
+	// 		expect(text).toMatchInlineSnapshot('"Pages supports wrangler.toml ⚡️"');
+	// 	});
+	// });
```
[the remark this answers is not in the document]
CarmenPopoviciu: oh, completely didn't notice it. it's leftover code and was meant to be removed. I removed it

packages/wrangler/src/dev.tsx
CarmenPopoviciu: oh ofc, because Map keys are unique. Completely overlooked that 😓 . Fixed!

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: do you mean the one in this file?

petebacondarwin: Yep

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: fixed

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: renamed to `getBindingsFromArgs()`

packages/wrangler/src/dev.tsx
CarmenPopoviciu: fixed

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: very good callout! fixed!

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -574,4 +574,269 @@ describe("pages dev", () => {
...
+		});
+	});
+
+	it("should use top-level configuration specified in `wrangler.toml`", async () => {
+		await runDevSession(workerPath, "", async (session) => {
+			await seed(workerPath, {
+				"public/_worker.js": dedent`
+						export default {
+							async fetch(request, env) {
+								return new Response("Pages supports wrangler.toml ⚡️")
```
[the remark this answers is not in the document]
CarmenPopoviciu: done! also updated the readme and cleaned up the `wrangler.toml` file for both fixtures

penalosa:
> Regarding this test case:
> 
>     * pages dev should ignore wrangler.toml, if it exists but `pages_build_output_dir` is not specified in the configuration

One slight issue here is that this represents a break when compared to existing behaviour (where currently `pages dev` unconditionally reads `wrangler.toml`). I wonder if we want to have a brief period where `pages dev` supports a config file without `pages_build_output_dir`, but warns loudly that the behaviour is about to change?

CarmenPopoviciu:
> > Regarding this test case:
> > ```
> > * pages dev should ignore wrangler.toml, if it exists but `pages_build_output_dir` is not specified in the configuration
> > ```
> 
> One slight issue here is that this represents a break when compared to existing behaviour (where currently `pages dev` unconditionally reads `wrangler.toml`). I wonder if we want to have a brief period where `pages dev` supports a config file without `pages_build_output_dir`, but warns loudly that the behaviour is about to change?

Is this really a breaking change tho? Given that current behaviour is that Pages does not support `wrangler.toml`, and the reason why `pages dev` works with `wrangler.toml` is purely coincidental. There is nowhere, that I am aware of at least, where we made this contract public

penalosa:
> > > Regarding this test case:
> > > ```
> > > * pages dev should ignore wrangler.toml, if it exists but `pages_build_output_dir` is not specified in the configuration
> > > ```
> > 
> > 
> > One slight issue here is that this represents a break when compared to existing behaviour (where currently `pages dev` unconditionally reads `wrangler.toml`). I wonder if we want to have a brief period where `pages dev` supports a config file without `pages_build_output_dir`, but warns loudly that the behaviour is about to change?
> 
> Is this really a breaking change tho? ...given that current behaviour is that Pages does not support `wrangler.toml`, and the reason why `pages dev` works with `wrangler.toml` is purely coincidental. There is nowhere, that I am aware of at least, where we made this contract public

I think `getBindingsProxy()` and a lot of the recent frameworks work depends on it

CarmenPopoviciu:
I see...I was not aware of that. Any thoughts on this Dario?

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -574,4 +574,240 @@ describe("pages dev", () => {
...
+						- KV_BINDING_TOML: KV_ID_TOML
+					- Vars:
+						- VAR1: "var1_toml"
+						- VAR2: "var2_toml"
+				`.replace(/\s/g, "")
+			);
+		});
+	});
+
+	it("should merge (with override) `wrangler.toml` configuration with configuration provided via the command line, with command line args taking precedence", async () => {
```
penalosa, line 668: Love this test!

dario-piotrowicz:
@CarmenPopoviciu yes, @penalosa is right this would be a bit breaking for `getPlatformProxy`...

I mean not the utility itself obviously but the workflow with it and `wrangler pages dev` 😕 

I think it would indeed be beneficial if we could not include this mandatory field for now and just show a warning regarding it until make the next major version of wrangler, giving users and frameworks some time to adopt/migrate over 🤔 

___

> Is this really a breaking change tho? ...given that current behaviour is that Pages does not support wrangler.toml, and the reason why pages dev works with wrangler.toml is purely coincidental. There is nowhere, that I am aware of at least, where we made this contract public

We recently did make this contract public: https://developers.cloudflare.com/pages/functions/bindings/ (https://github.com/cloudflare/cloudflare-docs/pull/12731)

So I would indeed think that this would be a breaking change regardless of `getPlatformProxy`, no? 🤔 (cc. @admah)

packages/wrangler/src/dev.tsx
penalosa: We have a `utils` folder in Wrangler where this could live?

packages/wrangler/src/dev.tsx
```diff
@@ -1003,3 +1042,21 @@ export function getBindings(
...
+function mergeWithOverride<T>(
+	source: Array<T>,
+	override: Array<T>,
+	keyProperty: keyof T
+): Array<T> {
+	const sourceMap: Map<T[typeof keyProperty], T> = new Map();
+	source.forEach((el) => sourceMap.set(el[keyProperty], el));
+
+	// preserve everything in `override` as it takes precedence over `source`,
+	// and add remaining elements from `source` that were not overriden
```
penalosa, line 1059: Is this comment correct? Specifically "add remaining elements from `source` that were not overriden"?

packages/wrangler/src/pages/dev.ts
```diff
@@ -229,99 +236,90 @@ export function Options(yargs: CommonYargsArgv) {
...
+		);
+	}
 
 	if (directory !== undefined && command.length > 0) {
 		throw new FatalError(
 			"Specify either a directory OR a proxy command, not both.",
 			1
 		);
 	} else if (directory === undefined) {
-		logger.warn(
```
penalosa, line 294: Is there a reason we're removing this deprecation warning?

penalosa (approved):
Couple nits (and the change from ignoring to warning on invalid config files) but otherwise LGTM!

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -574,4 +574,269 @@ describe("pages dev", () => {
...
+				"wrangler.toml": dedent`
+					name = "pages-project"
+					pages_build_output_dir = "public"
+					# commenting this out would result in a warning. If there is no "compatibility_date"
+					# related warning in stdout, then this value was picked up
+					compatibility_date = "2023-01-01"
+
+					[dev]
+					#ip = "127.0.0.1"
+					#port = 8181
```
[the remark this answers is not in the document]
CarmenPopoviciu: whoo, nice catch! I meant to remove them, but forgot. removed!

packages/wrangler/e2e/pages-dev.test.ts
CarmenPopoviciu: yay 🎉

packages/wrangler/src/dev.tsx
CarmenPopoviciu: yes, that is what the function does. So for example given:
```
const source = [{id: 1, name: "1"}, {id: 2, name: "2"}, {id: 3, name: "3"}]
const override = [{id: 1, name: "new1"}, {id: 2, name: "new2"}, {id: 4, name: "4"}]
```
the result of `mergeWithOverride(source, override)` is:
```
[{id: 1, name: "new1"},{id: 2, name: "new2"},{id: 3, name: "3"},{id: 4, name: "4"}]
```

Otherwise, args would override the entire config, which is not what we want I guess

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: we're not removing it, just moved it inside `spawnProxyProcess`, and ensured we surface it only when `(directory === undefined && (command.length > 0 || port !== undefined))`. Otherwise, it got surfaced in some `pages dev` corner cases, where we'd actually not want it to

packages/wrangler/src/dev.tsx
penalosa: Sorry, yeah—I meant that it felt like that comment applied to the _whole_ function, not just the line below

CarmenPopoviciu: removed comment in favor of fn jsdoc

CarmenPopoviciu:
> > > Regarding this test case:
> > > ```
> > > * pages dev should ignore wrangler.toml, if it exists but `pages_build_output_dir` is not specified in the configuration
> > > ```
> > 
> > 
> > One slight issue here is that this represents a break when compared to existing behaviour (where currently `pages dev` unconditionally reads `wrangler.toml`). I wonder if we want to have a brief period where `pages dev` supports a config file without `pages_build_output_dir`, but warns loudly that the behaviour is about to change?
> 
> Is this really a breaking change tho? ...given that current behaviour is that Pages does not support `wrangler.toml`, and the reason why `pages dev` works with `wrangler.toml` is purely coincidental. There is nowhere, that I am aware of at least, where we made this contract public

After talking abt this internally, we're going to skip any warning for this use case, and keep the current behaviour as it is. I have added a test case in my test plan to reflect this behaviour. Pls see:
```
pages dev <directory> should pick up wrangler.toml configuration, even if pages_build_output_dir is not specified
```

petebacondarwin (approved):

PR #5296, opened 2024-03-19
explicitly 'keep' secret_text and secret_key bindings

RamIdeas:

This PR changes deployments to explicitly 'keep' (inherit from previous Version) secret_text and secret_key bindings. These were implicitly inherited bindings before but we want to explicitly inherit them now because we no longer implicitly inherit plain_text bindings in version uploads so we need to explicitly only inherit the secret bindings. This also opens the possibility of maybe eventually specifying secret bindings (without values, of course) in wrangler.toml in the future.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: no user-facing change
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: no user-facing change

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -8947,7 +8947,7 @@ export default{
 			writeWranglerToml();
 			writeWorkerSource();
 			mockSubDomainRequest();
-			mockUploadWorkerRequest({ keepVars: true });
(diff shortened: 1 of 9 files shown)


petebacondarwin (approved):
No version upload test 😭

PR #5301, opened 2024-03-19
[C3]: align wrangler.toml template file for Next.js with those of the other frameworks

dario-piotrowicz:

the wrangler.toml file for Next.js has an extra `MY_KV` binding that is used for the scaffold example we provide to the user

this can be a bit inconvenient as this is the only toml among all the framework that has such extra binding, this requires us to take extra care of the toml when making modifications while all the other toml template files instead are all the same

fix this by alinging the toml file with that all the other frameworks and adding to it whatever is necessary for the scaffold KV example programmatically

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: this flow is already all tested
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: the binding name is included in the template README and not something we document elsewhere

### DIFF
--- a/packages/create-cloudflare/templates/next/c3.ts
+++ b/packages/create-cloudflare/templates/next/c3.ts
@@ -16,6 +16,7 @@ import {
 } from "helpers/files";
 import { detectPackageManager } from "helpers/packageManagers";
 import { installPackages } from "helpers/packages";
+import MagicString from "magic-string";
 import { getTemplatePath } from "../../src/templates";
 import type { TemplateConfig } from "../../src/templates";
 import type { C3Args, C3Context } from "types";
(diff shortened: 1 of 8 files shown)


packages/create-cloudflare/templates/next/c3.ts
```diff
@@ -27,10 +28,30 @@ const generate = async (ctx: C3Context) => {
...
-		join(ctx.project.path, "wrangler.toml")
+	const wranglerToml = readFile(join(getTemplatePath(ctx), "wrangler.toml"));
+
+	const newToml = new MagicString(wranglerToml);
+
+	// Note: here we add `# KV Example:` to the toml file for the KV example, we don't actually
+	//       include the comment in the template wrangler.toml file just so to keep it identical
+	//       and consistent with that of all the other frameworks
+	//       (instead of making it a special case which needs extra care)
+	newToml.replace(
```
dario-piotrowicz, line 39: PS: honestly I don't see any benefit in using `MagicString` here to do the replacing as Regex replacement would have worked just as well... but I figured I'd use `MagicString` to be consistent with the codebase 🤔

jculvey (approved):

packages/create-cloudflare/templates/next/c3.ts
dario-piotrowicz: ah ok cool, I guess I'll go old school and use a regex then 🙂👍

PR #5302, opened 2024-03-19
fix: ensure `tsc-all.mjs` can run on Windows

mrbbot:

Should hopefully fix tests on `main`. `fixtures/vitest-pool-workers-examples/tsc-all.mjs` wasn't compatible with Windows, but checks run on Windows on `main`.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: fixing internal type checking script
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: fixing internal type checking script
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: fixing internal type checking script

### DIFF
--- a/fixtures/vitest-pool-workers-examples/tsc-all.mjs
+++ b/fixtures/vitest-pool-workers-examples/tsc-all.mjs
@@ -1,3 +1,4 @@
+import assert from "node:assert";
 import childProcess from "node:child_process";
 import fs from "node:fs/promises";
 import path from "node:path";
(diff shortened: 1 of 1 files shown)


fixtures/vitest-pool-workers-examples/tsc-all.mjs
```diff
@@ -17,10 +17,12 @@ async function* walkTsConfigs(rootPath) {
...
 for await (const tsconfigPath of walkTsConfigs(__dirname)) {
 	console.log(`Checking ${path.relative(__dirname, tsconfigPath)}...`);
-	const result = childProcess.spawnSync(
-		"node_modules/.bin/tsc",
-		["-p", tsconfigPath],
-		{ stdio: "inherit", cwd: __dirname, shell: true }
-	);
+	let tsc = path.resolve(__dirname, "node_modules/.bin/tsc");
+	if (process.platform === "win32") tsc += ".cmd";
+	const result = childProcess.spawnSync(tsc, ["-p", tsconfigPath], {
```
petebacondarwin, line 22: Could we not just run `pnpm tsc` or similar here?

mrbbot: Hmmm, yeah that would work. `pnpm` gives a slightly annoying warning about `.npmrc` files. Given we're already running this with `pnpm check:type`, `tsc` should be on the path. Updated to use that directly instead. 👍

petebacondarwin (approved):

PR #5310, opened 2024-03-20
[wrangler] Watch the entire module root + watch requirements.txt

penalosa:

In `--no-bundle` mode watch the entire module root for changes, rather than just the entrypoint file. Additionally, watch the `requirements.txt` file for Python workers.

Fixes [EW-8223](https://jira.cfdata.org/browse/EW-8223)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: bugfix to how this behaviour should work

### DIFF
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -91,7 +91,7 @@ async function runDevSession(
 
 		in ${workerPath} {
 			exits {
-        $$ ${WRANGLER} dev ${flags}
+        $ ${WRANGLER} dev ${flags}
 			}
 		}
 			`;
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -5,11 +5,12 @@ import * as nodeNet from "node:net";
 import path from "node:path";
 import { setTimeout } from "node:timers/promises";
 import shellac from "shellac";
+import dedent from "ts-dedent";
 import { Agent, fetch, setGlobalDispatcher } from "undici";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { normalizeOutput } from "./helpers/normalize";
 import { retry } from "./helpers/retry";
-import { dedent, makeRoot, seed } from "./helpers/setup";
+import { makeRoot, seed } from "./helpers/setup";
 import { WRANGLER } from "./helpers/wrangler-command";
 
 // Use `Agent` with lower timeouts so `fetch()`s inside `retry()`s don't block for a long time
@@ -90,7 +91,7 @@ async function runDevSession(
 
 		in ${workerPath} {
 			exits {
-        $ ${WRANGLER} dev ${flags}
+        $$ ${WRANGLER} dev ${flags}
 			}
 		}
 			`;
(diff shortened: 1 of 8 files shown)


dom96 (approved):
LGTM, let's merge :)

packages/wrangler/e2e/dev.test.ts
```diff
@@ -285,10 +286,15 @@ describe("basic dev python tests", () => {
 					compatibility_date = "2023-01-01"
 					compatibility_flags = ["python_workers"]
 			`,
+			"arith.py": dedent`
```
CarmenPopoviciu, line 289: can we pls use the full noun here? `arith` might not be immediately meaningful for everyone

packages/wrangler/src/dev/use-esbuild.ts
```diff
@@ -199,11 +199,23 @@ export function useEsbuild({
...
 			stopWatching = bundleResult?.stop;
 
-			// if "noBundle" is true, then we need to manually watch the entry point and
-			// trigger "builds" when it changes
+			// if "noBundle" is true, then we need to manually watch all modules and
+			// trigger "builds" when any change
 			if (noBundle) {
-				const watcher = watch(entry.file, {
+				const watching = [path.resolve(entry.moduleRoot)];
+				// Should we try and watch a Python reqyurements.txt file?
```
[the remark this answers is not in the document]
petebacondarwin:
```suggestion
				// Check whether we need to watch a Python requirements.txt file.
```

packages/wrangler/src/dev/use-esbuild.ts
```diff
@@ -199,11 +199,23 @@ export function useEsbuild({
...
+				const pythonRequirements =
```
petebacondarwin, line 207:
```suggestion
				const watchPythonRequirements =
```

packages/wrangler/src/dev/use-esbuild.ts
```diff
@@ -199,11 +199,23 @@ export function useEsbuild({
...
+						? path.resolve(entry.directory, "requirements.txt")
+						: undefined;
+
+				if (pythonRequirements) {
+					watching.push(pythonRequirements);
+				}
+
+				const watcher = watch(watching, {
 					persistent: true,
+					ignored: [".git", "node_modules"],
```
petebacondarwin, line 218: Should we be pulling in the `.gitignore` file here, perhaps?

I always get a bit wary when this stuff is hard-coded, as has been noted recently in Pages where it refuses to upload files within a node_modules folder.

petebacondarwin (changes requested):

packages/wrangler/src/dev/use-esbuild.ts
penalosa: This was intended to mimic the behaviour in https://github.com/cloudflare/workers-sdk/blob/d67836a599d55e5f0932cc5f1cc901e640015f47/packages/wrangler/src/deployment-bundle/find-additional-modules.ts#L177-L193

petebacondarwin: LOL which is code I wrote 😆 - OK, given that this is already the behaviour, I guess we are safe.

packages/wrangler/e2e/dev.test.ts
```diff
@@ -90,7 +91,7 @@ async function runDevSession(
 
 		in ${workerPath} {
 			exits {
-        $ ${WRANGLER} dev ${flags}
+        $$ ${WRANGLER} dev ${flags}
```
petebacondarwin, line 94: I think this results in unwanted output in the CI logs. Can we revert this?

PR #5322, opened 2024-03-20
Don't run checks or tests on main

penalosa:

All checks + tests are run on PRs as well as on Version Packages before a release, and so we don't need to re-run them on `main`. The relevant checks are `test:ci` and `check`, both of which are run on pull requests.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: CI change
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: CI change
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: CI change


GregBrimble:
We don't enforce up-to-date branches, so would mean that a conflict might not be caught until you go to make a release. And worst-case, the Version Packages PR itself could potentially be out-of-date?

penalosa:
> We don't enforce up-to-date branches, so would mean that a conflict might not be caught until you go to make a release. And worst-case, the Version Packages PR itself could potentially be out-of-date?

We'll be keeping track of Version Packages for this (and we lock main as we do releases, so there's no chance of Version Packages getting out of sync with main)

RamIdeas (approved):

PR #5327, opened 2024-03-20
[pages] Add `wrangler pages config download`

penalosa:

Adds a Wrangler command to support downloading a Pages project's config into a `wrangler.toml` file

Fixes [DEVX-1178](https://jira.cfdata.org/browse/DEVX-1178)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Docs will be coming soon


.changeset/fresh-comics-carry.md
```diff
@@ -0,0 +1,5 @@
+---
```
CarmenPopoviciu, line 1: are you planning to land these as two separate commits? That would be the tidier way IMHO

penalosa: That's fair, I can definitely split it out into another PR (I did it like this initially because I needed a pre-release with both changes)

packages/wrangler/src/pages/index.ts
```diff
@@ -44,7 +45,7 @@ export function pages(yargs: CommonYargsArgv) {
 						Build.Handler
 					)
 					.command(
-						"build-env",
+						"build-env [projectDir]",
```
CarmenPopoviciu, line 48: I imagine we'd always run this from the Pages project directory. Is this needed for CI?

packages/wrangler/src/pages/build-env.ts
```diff
@@ -9,17 +12,37 @@ import type {
...
-	return yargs;
+	return yargs.positional("projectDir", {
+		type: "string",
+		description: "The location of the Pages project",
+	});
 }
 
 export const Handler = async (args: PagesBuildEnvArgs) => {
-	const config = readConfig(undefined, {
+	if (!args.projectDir) {
```
CarmenPopoviciu, line 22: what's the reason for making this mandatory. Would it make sense to default to the current working directory?

.changeset/fresh-comics-carry.md
CarmenPopoviciu: if you can split this PR in two dedicated commit, that's already helpfu

packages/wrangler/src/pages/index.ts
penalosa: This is needed for CI, yeah

packages/wrangler/src/pages/build-env.ts
```diff
@@ -9,17 +12,37 @@ import type {
...
 		...args,
 		// eslint-disable-next-line turbo/no-undeclared-env-vars
 		env: process.env.PAGES_ENVIRONMENT,
 	});
+	// Fail if the config file exists but isn't valid for Pages
 	if (!config.pages_build_output_dir) {
-		throw new FatalError("No Pages config file found");
+		throw new FatalError(
+			"No Pages config file found",
+			EXIT_CODE_NO_CONFIG_FOUND
```
CarmenPopoviciu, line 44: I'm assuming we need to exit with this exit code here? Cause technically we did find a `wrangler.toml` file, just it doesn't have the mandatory `pages_build_output_dir` Pages field. IMO the err message should reflect that, but not sure if we can exit with the same exit code but a diff err message than on line https://github.com/cloudflare/workers-sdk/pull/5327/files?diff=unified&w=1#diff-f0b20ef6f77561affbcad6b9afae83c2b49211c0e1d7af12a81abf191b72bbafR31 ?

packages/wrangler/src/pages/build-env.ts
penalosa: Potentially, but given this is a hidden command only meant to be used in CI (where the argument will always be set), I'm not sure if we should?

packages/wrangler/src/pages/build.ts
```diff
@@ -45,6 +45,10 @@ export function Options(yargs: CommonYargsArgv) {
 				type: "string",
 				description: "The location for the output config file",
 			},
+			"build-metadata-path": {
```
CarmenPopoviciu, line 48: what is this for?

packages/wrangler/src/metrics/send-event.ts
```diff
@@ -71,7 +71,8 @@ export type EventNames =
 	| "list worker versions"
 	| "view versioned deployment"
 	| "view latest versioned deployment"
-	| "list versioned deployments";
+	| "list versioned deployments"
+	| "pages download config";
```
CarmenPopoviciu, line 75: looking at these events, the naming convention for pages seems to be `<action> pages ...`. So perhaps we can keep that convention here to and use `download pages config` instead?

.changeset/fresh-comics-carry.md
petebacondarwin: It won't make much difference outside of PR review, since we squash all the commits from a PR into a single commit when merging. So unless this will aid review, I wouldn't bother.

packages/wrangler/src/__tests__/pages/pages.test.ts
```diff
@@ -24,6 +24,7 @@ describe("pages", () => {
 
 		Commands:
 		  wrangler pages dev [directory] [-- command..]  🧑‍💻 Develop your full-stack Pages application locally
+		  wrangler pages download                        ⚡️ Download settings from your project
```
petebacondarwin, line 27: Should this be a hidden command to avoid having to document and support it publicly?

CarmenPopoviciu, line 27: anything speaking against placing this command at the bottom of the Pages cmds list, after `wrangler pages deploy [directory]`? I guess my thinking is that as a user I'd be first interested to know how to create/deploy a project, before I'd want to know how to download the config. But maybe that's just me?

packages/wrangler/src/pages/build-env.ts
```diff
@@ -9,17 +12,37 @@ import type {
...
 
 export const Handler = async (args: PagesBuildEnvArgs) => {
-	const config = readConfig(undefined, {
+	if (!args.projectDir) {
+		throw new FatalError("No Pages project location specified");
+	}
+	const configPath = path.resolve(args.projectDir, "wrangler.toml");
+	// Fail early if the config file doesn't exist
+	try {
+		await stat(configPath);
```
petebacondarwin, line 28: Why not `existsSync()` to avoid the try-catch?

packages/wrangler/src/pages/build-env.ts
```diff
@@ -9,17 +12,37 @@ import type {
...
+			EXIT_CODE_NO_CONFIG_FOUND
+		);
+	}
+	const config = readConfig(path.resolve(args.projectDir, "wrangler.toml"), {
 		...args,
 		// eslint-disable-next-line turbo/no-undeclared-env-vars
 		env: process.env.PAGES_ENVIRONMENT,
 	});
+	// Fail if the config file exists but isn't valid for Pages
 	if (!config.pages_build_output_dir) {
```
petebacondarwin, line 41: It feels like this would be cleaner in a helper that made it clear we are just checking whether it is a valid pages config, which would avoid the need for a comment explaining this.

```suggestion
	if (!isPagesConfig(config)) {
```

packages/wrangler/src/pages/build-env.ts
petebacondarwin: Yeah it feels a bit weird to exit with "no config found".
From the user's point of view they will see that error and get confused because they will see that there is a wrangler.toml and maybe not realise that it is just malformed.

packages/wrangler/src/pages/build-env.ts
CarmenPopoviciu: that would have been my next suggestion, to make it hidden if it's only needed in CI 😅 . I didn't notice it already was. Then totes fine the way it is now

packages/wrangler/src/pages/download-config.ts
```diff
@@ -0,0 +1,226 @@
...
+import { PAGES_CONFIG_CACHE_FILENAME } from "./constants";
+import type { RawEnvironment } from "../config";
+import type {
+	CommonYargsArgv,
+	StrictYargsOptionsToInterface,
+} from "../yargs-types";
+import type { PagesConfigCache } from "./types";
+import type { Project } from "@cloudflare/types";
+
+// TODO: fix the Project definition
```
petebacondarwin, line 20: Ticket?

.changeset/fresh-comics-carry.md
CarmenPopoviciu: oh right, those will automatically get squashed :/. nvm then

packages/wrangler/src/__tests__/pages/pages.test.ts
penalosa: I think this list is sorted alphabetically, but I'm not certain

packages/wrangler/src/pages/download-config.ts
CarmenPopoviciu, line 20: todo as part of this PR?

packages/wrangler/src/pages/download-config.ts
```diff
@@ -0,0 +1,226 @@
...
+	project: PagesDeploymentConfig,
+	accountId: string
+): Promise<RawEnvironment> {
+	const configObj = {} as RawEnvironment;
+	configObj.compatibility_date =
+		project.compatibility_date ?? new Date().toISOString().substring(0, 10);
+	if (project.compatibility_flags?.length)
+		configObj.compatibility_flags = project.compatibility_flags;
+
+	for (const [name, envVar] of Object.entries(project.env_vars ?? {}).filter(
```
petebacondarwin, line 71: NIT: Using a filter here is a bit overkill. You could just check the type in the body of the for-loop, avoiding the creation of an extra array for nothing.

packages/wrangler/src/pages/download-config.ts
```diff
@@ -0,0 +1,226 @@
...
+			dataset,
+		});
+	}
+	for (const [name] of Object.entries(project.ai_bindings ?? {})) {
+		configObj.ai = { binding: name };
+	}
+	return configObj;
+}
+async function writeWranglerToml(toml: RawEnvironment) {
+	const wranglerCommandUsed = ["wrangler", ...process.argv.slice(2)].join(" ");
```
petebacondarwin, line 164: Using `process.argv` here seems a bit brittle. Could we get this via yargs and pass it through? E.g. `yargs.parsed.argv`?

petebacondarwin: Or at least use yargs `hideBin()` function.

packages/wrangler/src/pages/download-config.ts
```diff
@@ -0,0 +1,226 @@
...
+
+	// Pages does not support custom wrangler.toml locations, so always write to ./wrangler.toml
+	await writeFile(
+		"wrangler.toml",
```
petebacondarwin, line 168: Will this command always be run from the place where wrangler.toml needs to be written?

Would we want to consider looking for the nearest package.json?

packages/wrangler/src/pages/download-config.ts
```diff
@@ -0,0 +1,226 @@
...
+			TOML.stringify(toml as TOML.JsonMap),
+		].join("\n")
+	);
+}
+
+async function downloadProject(accountId: string, projectName: string) {
+	const project = await fetchResult<PagesProject>(
+		`/accounts/${accountId}/pages/projects/${projectName}`
+	);
+	console.log(JSON.stringify(project, null, 2));
```
petebacondarwin, line 181: What about using `logger` here? So that users can control the logging output?

petebacondarwin: And perhaps make it `logger.debug()`?

CarmenPopoviciu, line 181: do we need this log here?

packages/wrangler/src/metrics/send-event.ts
petebacondarwin: Yes that is correct @CarmenPopoviciu  - sparrow metrics events have a specific convention.

packages/wrangler/src/__tests__/pages/pages.test.ts
CarmenPopoviciu: judging by
```
wrangler pages project
wrangler pages deployment
wrangler pages deploy [directory]
```
doesn't seem to be. Not sure if there is any convention in how we list these commands tbh

packages/wrangler/src/pages/build-env.ts
CarmenPopoviciu: indeed, that was my thinking as well

packages/wrangler/src/pages/build-env.ts
CarmenPopoviciu: I agree. We are using this check in multiple places, so having a helper would make thing tidier and more maintainable on the long term. I remember adding this helper in one of my PRs, but not sure if it isn't in the `pages dev` PR which didn't land yet...

CarmenPopoviciu: oh, found it: https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/config/validation.ts#L45-L47

CarmenPopoviciu: it prolly needs some adjusting, since rn it take a raw config object as param.

packages/wrangler/src/__tests__/pages/pages.test.ts
CarmenPopoviciu: that's a good callout.

penalosa: We're planning to document this publicly as part of the launch AFAIK

packages/wrangler/src/pages/build-env.ts
penalosa: Added a new exit code

packages/wrangler/src/metrics/send-event.ts
penalosa: Changed

packages/wrangler/src/pages/build.ts
penalosa: This is for integration with CI—perhaps it shouldn't be in this PR

packages/wrangler/src/pages/build-env.ts
penalosa: Changed to using `isPagesConfig()`

packages/wrangler/src/pages/download-config.ts
penalosa: I think this should always be run from the root of the Pages project. If a user ends up running it in the wrong place they can move it to the root, but I think that's more predictable? Open to changing this though

petebacondarwin: Fair enough. It might be weird for it to appear somewhere else.

packages/wrangler/src/pages/download-config.ts
penalosa: I've removed this entirely—it doesn't really need to be in the generated header

packages/wrangler/src/__tests__/pages/pages.test.ts
penalosa: Moved

packages/wrangler/src/pages/download-config.ts
penalosa: It depends on [SHIP-8984](https://jira.cfdata.org/browse/SHIP-8984)

packages/wrangler/src/pages/build-env.ts
CarmenPopoviciu: I guess this will prolly need to be persisted in the BE work as well, so just calling that out

penalosa: Yep, I'll add a task

CarmenPopoviciu (approved):

PR #5330, opened 2024-03-21
chore(wrangler): Add `pages dev` tests

CarmenPopoviciu:

This PR adds some sanity-check unit and e2e tests for `wrangler pages dev`.

Screenshot 2024-03-22 at 13 33 27

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: not user facing. Only adding tests
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not user facing. Only adding tests

### DIFF
--- a/packages/wrangler/e2e/pages-dev.test.ts
+++ b/packages/wrangler/e2e/pages-dev.test.ts
@@ -5,36 +5,87 @@ import getPort from "get-port";
 import shellac from "shellac";
 import { fetch } from "undici";
 import { beforeEach, describe, expect, it } from "vitest";
+import { normalizeOutput } from "./helpers/normalize";
 import { retry } from "./helpers/retry";
 import { dedent, makeRoot, seed } from "./helpers/setup";
 import { WRANGLER } from "./helpers/wrangler-command";
 
+type MaybePromise<T = void> = T | Promise<T>;
+
+interface SessionData {
+	port: number;
+	stdout: string;
+	stderr: string;
+}
+
+const waitUntilOutputContains = async (
+	session: SessionData,
+	substring: string,
+	intervalMs = 100
+) => {
+	await retry(
+		(stdout) => !stdout.includes(substring),
+		async () => {
+			await setTimeout(intervalMs);
+			return session.stdout + "\n\n\n" + session.stderr;
+		}
+	);
+};
+
 async function runDevSession(
 	workerPath: string,
 	flags: string,
-	session: (port: number) => Promise<void>
+	session: (sessionData: SessionData) => MaybePromise
 ) {
 	let pid;
 	try {
-		const port = await getPort();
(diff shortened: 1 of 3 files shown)


packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -5,36 +5,87 @@ import getPort from "get-port";
 import shellac from "shellac";
 import { fetch } from "undici";
 import { beforeEach, describe, expect, it } from "vitest";
+import { normalizeOutput } from "./helpers/normalize";
 import { retry } from "./helpers/retry";
 import { dedent, makeRoot, seed } from "./helpers/setup";
 import { WRANGLER } from "./helpers/wrangler-command";
 
+type MaybePromise<T = void> = T | Promise<T>;
```
CarmenPopoviciu, line 13: technically, all of `MaybePromise`, `SessionData`, `waitUntilOutputContains` and `runDevSession` are duplicated from wrangler proper. The only diff is the `dev` command `runDevSession` runs under the hood (`wrangler dev` vs `wrangler pages dev`).

We could keep things as they are, in the idea that convergence will make one of these obsolete, or we could extract the common functionality in a helper to clean up the code duplication. No major blocker, just calling it out

CarmenPopoviciu:
Ironically, the `E2E Tests` CI job is skipped, so y'all just have to take my word that all these tests are 🍏  ;)

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -227,4 +574,137 @@ describe("pages dev tests", () => {
 			"Unexpected error building Functions directory"
 		);
 	});
+
+	it("should modify _routes.json during dev session (_worker)", async () => {
```
petebacondarwin, line 578: Is this message correct? 
Shouldn't it be "should support modifying ..."?

packages/wrangler/src/__tests__/pages/dev.test.ts
```diff
@@ -0,0 +1,34 @@
+import { runInTempDir } from "../helpers/run-in-tmp";
+import { runWrangler } from "../helpers/run-wrangler";
+
+/**
+ * Given that `runWrangler()` mocks out the underlying implementation
+ * (see "jest.setup.ts") there's only so much worth testing here.
```
petebacondarwin, line 6: I believe this can be turned off if needed. But perhaps for the future...

petebacondarwin (approved):

packages/wrangler/e2e/pages-dev.test.ts
CarmenPopoviciu: yes indeed. fixed! ✅

PR #5359, opened 2024-03-22
chore: ensure c3 frameworks dependabot changesets get correct filename

petebacondarwin:


tools/dependabot/generate-dependabot-pr-changesets.ts
```diff
@@ -36,21 +37,29 @@ function processArgs(): Args {
...
-			- PackageJSON: The path to the package JSON being updated by Dependabot`);
+			- PackageJSON: The path to the package JSON being updated by Dependabot
+			- Changeset prefix: The prefix to go on the front of the filename of the generated changeset.
+		`);
 	}
 	return {
 		prNumber: args[0],
 		packageName: args[1],
 		packageJSONPath: args[2],
+		changesetPrefix: args[3],
```
dario-piotrowicz, line 53: what about making the fourth argument optional? 🤔 
as in
```ts
changesetPrefix: args[3] ?? 'dependabot-update'
```

.github/workflows/c3-dependabot-versioning-prs.yml
```diff
@@ -32,4 +32,4 @@ jobs:
           git config --global user.name 'Wrangler automated PR updater'
 
       - name: Generate C3 changesets
-        run: node -r esbuild-register tools/dependabot/generate-dependabot-pr-changesets.ts ${{ github.event.number }} create-cloudflare packages/create-cloudflare/src/frameworks/package.json
+        run: node -r esbuild-register tools/dependabot/generate-dependabot-pr-changesets.ts ${{ github.event.number }} create-cloudflare packages/create-cloudflare/src/frameworks/package.json c3-frameworks-update
```
dario-piotrowicz, line 35: shall we also add a comment to try to avoid this getting updated in the future without also updating the other workflow?
```suggestion
        # Note: the c3-frameworks-update prefix needs to be such as it is used by the .github/workflows/c3-e2e-dependabot.yml workflow
        run: node -r esbuild-register tools/dependabot/generate-dependabot-pr-changesets.ts ${{ github.event.number }} create-cloudflare packages/create-cloudflare/src/frameworks/package.json c3-frameworks-update
```

dario-piotrowicz (approved):
LGTM 😃

.github/workflows/c3-dependabot-versioning-prs.yml
petebacondarwin: Good idea.

tools/dependabot/generate-dependabot-pr-changesets.ts
petebacondarwin: Since it is internal there is no real benefit to making it optional. Better to force people to be explicit.

PR #5363, opened 2024-03-24
add new Env Wrangler integration type

dario-piotrowicz:
Usage example: https://github.com/dario-piotrowicz/nitro-generic-wrangler-env-type-example

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Not necessary because:


dario-piotrowicz:
I've been looking into adding it to `workers-types` instead ([example](https://github.com/dario-piotrowicz/workers-types-env-poc)) as that could be the best place for such a type.

The work on `workers-types` is however non-trivial and we are not sure if the perceived benefits would be worth it, so we're pausing this for now and considering circling back to it in April.

PR #5373, opened 2024-03-25
Versions rollback

RamIdeas:

This PR implements versioned rollbacks via `wrangler rollback [version-id] --experimental-versions`.

Please note, the `experimental-versions` flag is required to use the new behaviour and the original `wrangler rollback` command is unchanged if run without this flag. 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/big-hounds-breathe.md
+++ b/.changeset/big-hounds-breathe.md
@@ -1,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+feature: Implement versioned rollbacks via `wrangler rollback [version-id] --experimental-versions`.
+
+Please note, the `experimental-versions` flag is required to use the new behaviour. The original `wrangler rollback` command is unchanged if run without this flag.
--- a/packages/wrangler/e2e/versions.test.ts
+++ b/packages/wrangler/e2e/versions.test.ts
@@ -16,6 +16,9 @@ function matchWhoamiEmail(stdout: string): string {
 function matchVersionId(stdout: string): string {
 	return stdout.match(/Version ID:\s+([a-f\d-]+)/)?.[1] as string;
 }
+function countOccurences(stdout: string, substring: string) {
+	return stdout.split(substring).length - 1;
+}
 
 describe("versions deploy", () => {
 	let root: string;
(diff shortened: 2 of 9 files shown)


.changeset/big-hounds-breathe.md
mrbbot, line 1: It might be helpful to explain what `wrangler rollback [version-id]` does here. I guess deploying with 100% of `[version-id]`, or the last 100% version if `[version-id]` is omitted?

packages/wrangler/e2e/versions.test.ts
```diff
@@ -282,7 +285,121 @@ describe("versions deploy", () => {
...
+			Version(s):  (100%) 00000000-0000-0000-0000-000000000000
+			                 Created:  TIMESTAMP
+			                     Tag:  -
+			                 Message:  -"
+		`);
+
+		console.log({
+			versionId1: countOccurences(list.stdout, versionId1),
+			versionId2: countOccurences(list.stdout, versionId2),
+		});
```
mrbbot, line 398:
```suggestion
```

packages/wrangler/src/versions/rollback/index.ts
```diff
@@ -0,0 +1,147 @@
...
+			({ percentage }) => percentage === 100
+		);
+
+		if (stableVersion) return stableVersion.version_id;
+	}
+
+	// if we get here, we did not find a stable version
+	throw new Error(
+		"Could not find stable Worker Version to rollback to. Please try again with an explicit Version ID."
+	);
```
mrbbot, line 146: How many deployments does `fetchLatestDeployments()` return? Does it make sense to paginate the deployments response here?

packages/wrangler/e2e/versions.test.ts
```diff
@@ -282,7 +285,121 @@ describe("versions deploy", () => {
 		expect(list.stdout).toContain(versionId2);
 	});
 
-	// TODO: rollback, when supported for --x-versions
+	it("rollback to first worker version", async () => {
```
mrbbot, line 288: Should there be a test for rolling back to a specific version ID specified on the command line too?

RamIdeas: https://github.com/cloudflare/workers-sdk/pull/5373/commits/b0db6adf2d5fda0c737c222aba3b06d31692f4a7

packages/wrangler/src/versions/rollback/index.ts
mrbbot: 10/20 is fine. Pagination definitely isn't required, just something that would be nice if easy to add. No worries. 👍

mrbbot (approved):

PR #5376, opened 2024-03-26
C3: Improve help output

jculvey:
Overhauls the output of c3's --help to improve readability and provide more information.

Uses a custom help implementation rather than the one built into `yargs` which has a number of pitfalls when it comes to formatting multiline text and lists of options.

Before:
image

After:
image

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Existing coverage sufficient
  - [ ] TODO (before merge)
  - [xg] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Already documented

### DIFF
--- a/packages/create-cloudflare/package.json
+++ b/packages/create-cloudflare/package.json
@@ -74,6 +74,7 @@
 		"execa": "^7.1.1",
 		"glob": "^10.3.3",
 		"haikunator": "^2.1.2",
+		"indent-string": "^5.0.0",
 		"magic-string": "^0.30.5",
 		"open": "^8.4.0",
 		"pnpm": "^8.10.0",
@@ -84,6 +85,7 @@
 		"vite-tsconfig-paths": "^4.0.8",
 		"which-pm-runs": "^1.1.0",
 		"wrangler": "workspace:*",
+		"wrap-ansi": "^9.0.0",
 		"yargs": "^17.7.1",
 		"yarn": "^1.22.19"
 	},
--- a/packages/create-cloudflare/src/helpers/args.ts
+++ b/packages/create-cloudflare/src/helpers/args.ts
@@ -98,11 +98,10 @@ const cliDefinition: ArgumentsDefinition = {
 
       When using the --framework option, C3 will dispatch to the official creation tool used by the framework (ex. "create-remix" is used for Remix).
 
-      You may specify additional arguments to be passed directly to these underlying tools by adding them after "--" in the argument string, like so:
+      You may specify additional arguments to be passed directly to these underlying tools by adding them after a "--" argument, like so:
 
       npm create cloudflare -- --framework next -- --ts
-
-      Note that there are 2 occurrences of "--" in the above arg string. The first "--" is required by "npm create" when passing --flags. The second "--" signals to C3 that the remaining arguments should be passed to "create-next-app".
+      pnpm create clouldfare --framework next -- --ts
       `,
 			values: [
 				{ name: "angular" },
@@ -193,6 +192,7 @@ const cliDefinition: ArgumentsDefinition = {
 		},
 		{
 			name: "help",
+			alias: "h",
 			description: "Show help and exit",
 			type: "boolean",
 			hidden: true,
@@ -215,7 +215,6 @@ export const parseArgs = async (argv: string[]): Promise<Partial<C3Args>> => {
 		.version(version)
 		.alias("v", "version")
 		.help(false)
-		.positional("potato", { type: "string" })
 		// note: we use strictOptions since `strict()` seems not to handle `positional`s correctly
 		.strictOptions() as unknown as Argv<C3Args>;
 
--- a/packages/create-cloudflare/src/helpers/args.ts
+++ b/packages/create-cloudflare/src/helpers/args.ts
@@ -29,7 +29,7 @@ export type AllowedValueDefinition = {
 export type ArgumentsDefinition = {
 	intro: string;
 	positionals?: ArgDefinition[];
-	options?: OptionDefinition[];
+	options: OptionDefinition[];
 };
 
 const cliDefinition: ArgumentsDefinition = {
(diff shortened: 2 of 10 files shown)


packages/create-cloudflare/src/helpers/args.ts
```diff
@@ -1,88 +1,239 @@
...
+			alias: "f",
 			type: "string",
-			requiresArg: true,
-			description:
-				"The type of framework to use to create a web application (when using this option `--type` is ignored)",
-		})
-		.option("deploy", {
+			description: `The type of framework to use to create a web application (when using this option "--type" is coerced to "web-framework")
+
+      When using the --framework option, C3 will dispatch to the official creation tool used by the framework (ex. "create-remix" is used for Remix).
```
dario-piotrowicz, line 99: it feels almost like we give Remix a preferential treatment here 😅

packages/create-cloudflare/src/helpers/args.ts
```diff
@@ -1,88 +1,239 @@
...
+
+      You may specify additional arguments to be passed directly to these underlying tools by adding them after "--" in the argument string, like so:
+
+      npm create cloudflare -- --framework next -- --ts
+
+      Note that there are 2 occurrences of "--" in the above arg string. The first "--" is required by "npm create" when passing --flags. The second "--" signals to C3 that the remaining arguments should be passed to "create-next-app".
```
dario-piotrowicz, line 105: instead of this and having to clarify the double `--`s I think you could say something like:

```
You may specify additional arguments to be passed directly to these underlying tools by adding them after a "--" argument, like so:
       npm create cloudflare -- --framework next -- --ts
       pnpm create clouldfare --framework next -- --ts
```

so providing multiple PMs here could clarify the `--` usage?

I'm not really sure, it's just a thought 🙂

packages/create-cloudflare/src/helpers/args.ts
```diff
@@ -1,88 +1,239 @@
...
+	const additionalArgs =
+		doubleDashesIdx < 0 ? [] : argv.slice(doubleDashesIdx + 1);
+
+	const yargsObj = yargs(hideBin(c3Args))
+		.scriptName("create-cloudflare")
+		.usage("$0 [args]")
 		.version(version)
 		.alias("v", "version")
+		.help(false)
+		.positional("potato", { type: "string" })
```
dario-piotrowicz, line 218: if add a `potato` arg I think we should also add a `carrot` one, either that or we should remove this argument 🥔

dario-piotrowicz:
could you add an entry in the help message regarding `--help` and `-h`? 🙂 

regarding `-h`, it seems like that's not recognized, is that intentional?
Screenshot 2024-03-26 at 14 50 35

dario-piotrowicz:
the help message I get if I provide a wrong flag seems to be doubly wrapped 😕 
(while the one I get with `--help` is not)

I don't imagine this to be intentional, is it?

| with `--help` | with `--NONEXISTENT` |
|--------|--------|
| Screenshot 2024-03-26 at 14 54 35 | Screenshot 2024-03-26 at 14 54 56 |

packages/create-cloudflare/src/helpers/args.ts
jculvey: Yeah, but having an in-context usage example seems useful. Any thoughts on an alternative?

jculvey:
> the help message I get if I provide a wrong flag seems to be doubly wrapped 😕 (while the one I get with `--help` is not)
> 
> I don't imagine this to be intentional, is it?
> 

Ugh 🤦 . Good catch, no not intentional, I just missed it. It appears that `strictOptions` doesn't throw an error that we can catch, it just has a default behavior of showing its internal help and exiting.

packages/create-cloudflare/package.json
```diff
@@ -74,6 +74,7 @@
 		"execa": "^7.1.1",
 		"glob": "^10.3.3",
 		"haikunator": "^2.1.2",
+		"indent-string": "^5.0.0",
```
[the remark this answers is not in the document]
jculvey: I'm tempted too but I figured since they're small, using these gives us one less thing to worry about for now.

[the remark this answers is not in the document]
dario-piotrowicz: that's fine by me, I honestly don't have a huge preference either way 🙂 :+1:

jculvey:
ldw exception CR: lwd-exempt CR: https://jira.cfdata.org/browse/CR-850376

jculvey:
> the help message I get if I provide a wrong flag seems to be doubly wrapped 😕 (while the one I get with `--help` is not)
> 
> I don't imagine this to be intentional, is it?

Ok fixed this by making a custom `strictOptions`

packages/create-cloudflare/package.json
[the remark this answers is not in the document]
IgorMinar: Since this is a devDep rather than dep, it's less bad but looking at the package contents I have to roll my eyes just a little :-) — it's a single simple function.

On the other hand we need to be very careful about deps and consider them a liability. They don't add much and only poses a risk of supply chain attacks or other issues (remember leftpad?). 

In general the fewer deps we have, the lower the risk that something goes awry.

IgorMinar (changes requested):
I still see the initial help prose having trouble with text wrapping, all the remaining text seems to wrap well, so it's just the first paragraph. Can you please take another look?

Screenshot 2024-03-26 at 5 27 20 PM
Screenshot 2024-03-26 at 5 28 26 PM

jculvey:
> I still see the initial help prose having trouble with text wrapping, all the remaining text seems to wrap well, so it's just the first paragraph. Can you please take another look?
> 

Fixed

IgorMinar (approved):
I tested the latest build and it looks great now. Thanks James!

PR #5375, opened 2024-03-25
Skip comments in requirements and test dependencies

kflansburg:
## What this PR solves / how to test

Fixes EW-8270.

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): <https://github.com/cloudflare/cloudflare-docs/pull/>...
  - [ ] Not necessary because:

PR #5398, opened 2024-03-26
Skip comments in requirements and test dependencies

garrettgu10:

Fixes EW-8270. Revised version of #5375 ; this PR is separate because they're both done with forks. 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: we will be documenting this in other places

### DIFF
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -393,7 +393,7 @@ describe("python dependency tests", () => {
 		}));
 	});
 
-	it("can parse requirments file and use dependencies", async () => {
+	it("can parse requirements file and use dependencies", async () => {
 		await worker.runDevSession("", async (session) => {
 			const { text } = await retry(
 				(s) => s.status !== 200,
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -362,6 +362,51 @@ describe("basic dev python tests", () => {
 	});
 });
 
+describe("python dependency tests", () => {
+	let worker: DevWorker;
+
+	beforeEach(async () => {
+		worker = await makeWorker();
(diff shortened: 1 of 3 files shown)


packages/wrangler/e2e/dev.test.ts
```diff
@@ -362,6 +362,51 @@ describe("basic dev python tests", () => {
...
+					{
+						"name": "${workerName}",
+						"version": "0.0.0",
+						"private": true
+					}
+					`,
+		}));
+	});
+
+	it("can parse requirments file and use dependencies", async () => {
```
penalosa, line 396:
```suggestion
	it("can parse requirements file and use dependencies", async () => {
```

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -39,6 +39,16 @@ function isValidPythonPackageName(name: string): boolean {
 	return regex.test(name);
 }
 
+/**
+ * Strips comments and surrounding whitespace from a python requirements.txt line
+ * See https://pip.pypa.io/en/latest/reference/requirements-file-format/ (Comments)
+ * @param requirement The requirement to sanitize
+ */
+function stripPythonRequirement(requirement: string): string {
```
penalosa, line 47:
```suggestion
function stripComments(requirement: string): string {
```

penalosa (approved):
I'm a tad concerned about how fragile this might be, but let's land and see how it goes

petebacondarwin:
I think this may have caused the e2e test failing: https://github.com/cloudflare/workers-sdk/actions/runs/8661443570/job/23752569600#step:11:129

PR #5400, opened 2024-03-26
Log note of upcoming change of "Deployment ID" to "Version ID"

RamIdeas:

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: logging note of future change

### DIFF
--- a/packages/wrangler/src/__tests__/deployments.test.ts
+++ b/packages/wrangler/src/__tests__/deployments.test.ts
@@ -110,7 +110,7 @@ describe("deployments", () => {
 			🟩 Active
 
 
-			NOTE: Deployment ID is now referred to as Version ID. The output of this command will be changing in a future version of Wrangler to reflect this. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
+			NOTE: \\"Deployment ID\\" in this output will be changed to \\"Version ID\\" in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
 		`);
 			});
 
@@ -145,7 +145,7 @@ describe("deployments", () => {
 			🟩 Active
 
 
-			NOTE: Deployment ID is now referred to as Version ID. The output of this command will be changing in a future version of Wrangler to reflect this. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
+			NOTE: \\"Deployment ID\\" in this output will be changed to \\"Version ID\\" in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
 		`);
 			});
 
@@ -180,7 +180,7 @@ describe("deployments", () => {
 
 
 
-			NOTE: Deployment ID is now referred to as Version ID. The output of this command will be changing in a future version of Wrangler to reflect this. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
+			NOTE: \\"Deployment ID\\" in this output will be changed to \\"Version ID\\" in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
 		`);
 			});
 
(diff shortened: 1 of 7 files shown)


packages/wrangler/src/__tests__/deployments.test.ts
```diff
@@ -107,7 +107,10 @@ describe("deployments", () => {
 			Author:        Jean-Luc-Picard@federation.org
 			Source:        Rollback from Wrangler 🤠
 			Rollback from: MOCK-DEPLOYMENT-ID-2222
-			🟩 Active"
+			🟩 Active
+
+
+			NOTE: Deployment ID is now referred to as Version ID. The output of this command will be changing in a future version of Wrangler to reflect this. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
```
petebacondarwin, line 113: "... is now referred to ..." feels a bit ambiguous. I ask myself, "Where is it referred to as Version ID?", "Where is it not?", "How do I know which and when?"

Should this messages explicitly answer these questions?
Or perhaps it would be enough to phrase this as:

```suggestion
			NOTE: The term "Deployment ID" in this output will be changed to "Version ID" in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
```

packages/wrangler/src/utils/deployment-id-version-id-change.ts
```diff
@@ -0,0 +1,12 @@
+import { logger } from "../logger";
+
+export function logVersionIdChange() {
+	const docsLink =
+		"https://developers.cloudflare.com/workers/configuration/versions-and-deployments";
+
+	logger.log("");
+	logger.log("");
```
petebacondarwin, line 8: Why not just add some `\n` to the start of the log below?

petebacondarwin (commented):
If the experimental versions flag is enabled, does this output already display `Version ID` rather than `Deployment ID`? That would seem like a good step for people opting into this new flow. If that is the case, then perhaps this message should only be displayed to people not opted-in?

packages/wrangler/src/__tests__/deployments.test.ts
RamIdeas: cc @tanushreesharma-cf

RamIdeas:
> If the experimental versions flag is enabled, does this output already display `Version ID` rather than `Deployment ID`? That would seem like a good step for people opting into this new flow. If that is the case, then perhaps this message should only be displayed to people not opted-in?

That is indeed the case. This message is logged in the old `deployments list` handler. The new handler – used if `--x-versions` is specified – omits Deployment IDs entirely and only shows Version IDs. You can try it now in the latest wrangler in the worker-app fixture: `npx wrangler@latest deployments list --x-versions`

packages/wrangler/src/utils/deployment-id-version-id-change.ts
RamIdeas: I first logged the message with logger.warn and the prefix WARNING was added before the "\n\n"

packages/wrangler/src/__tests__/deployments.test.ts
tanushreesharma-cf: Minor tweak to Pete's suggestion. Removed "The term" since it's not just the term that's changing but the definition. 

```suggestion
			NOTE: "Deployment ID" in this output will be changed to "Version ID" in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
```

RamIdeas: https://github.com/cloudflare/workers-sdk/pull/5400/commits/64ed9086802df55c13e6862510f2881fbc25ed2a

packages/wrangler/src/utils/deployment-id-version-id-change.ts
RamIdeas: https://github.com/cloudflare/workers-sdk/pull/5400/commits/64ed9086802df55c13e6862510f2881fbc25ed2a

petebacondarwin (approved):

petebacondarwin:
This needs to link to a LDW CR before we merge.

RamIdeas:
> This needs to link to a LDW CR before we merge.

Thanks for the reminder! Added it to the description.

PR #5405, opened 2024-03-27
Catch erroneous use of `deployments view [deployment-id] --experimental-versions` command

RamIdeas:

Adds a `view` command to the new `deployments` subcommands opted-in via the `--experimental-versions` flag which shows a deprecation message and points the user to either run `wrangler deployments status --experimental-versions` or `wrangler versions view <version-id> --experimental-versions` instead.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/versions/deployments/deployments.view.test.ts
+++ b/packages/wrangler/src/__tests__/versions/deployments/deployments.view.test.ts
@@ -1,18 +1,22 @@
+import { UserError } from "../../../errors";
 import { runWrangler } from "../../helpers/run-wrangler";
 
 describe("deployments view", () => {
-	test("depracted error when run with no args", async () => {
+	test("error when run with no args", async () => {
 		const result = runWrangler("deployments view  --x-versions");
 
 		await expect(result).rejects.toMatchInlineSnapshot(
-			`[Error: \`wrangler deployments view\` is deprecated and will be removed in a future major version. Please use \`wrangler deployments status --x-versions\` instead.]`
+			`[Error: \`wrangler deployments view\` has been renamed \`wrangler deployments status --x-versions\`. Please use that command instead.]`
 		);
+		await expect(result).rejects.toBeInstanceOf(UserError);
 	});
-	test("depracted error when run with positional arg", async () => {
+
+	test("error when run with positional arg", async () => {
 		const result = runWrangler("deployments view dummy-id  --x-versions");
 
 		await expect(result).rejects.toMatchInlineSnapshot(
-			`[Error: \`wrangler deployments view <deployment-id>\` is deprecated and will be removed in a future major version. Deployment ID is now referred to as Version ID. Please use \`wrangler versions view [version-id] --x-versions\` instead.]`
+			`[Error: \`wrangler deployments view <deployment-id>\` has been renamed \`wrangler versions view [version-id] --x-versions\`. Please use that command instead.]`
 		);
+		await expect(result).rejects.toBeInstanceOf(UserError);
 	});
 });
--- a/packages/wrangler/src/__tests__/versions/deployments/deployments.view.test.ts
+++ b/packages/wrangler/src/__tests__/versions/deployments/deployments.view.test.ts
@@ -1,0 +1,18 @@
+import { runWrangler } from "../../helpers/run-wrangler";
+
+describe("deployments view", () => {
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/versions/deployments/view.ts
```diff
@@ -0,0 +1,41 @@
...
+	await printWranglerBanner();
+
+	if (args.deploymentId === undefined) {
+		throw new UserError(
+			"`wrangler deployments view` is deprecated and will be removed in a future major version. Please use `wrangler deployments status --x-versions` instead."
+		);
+	} else {
+		throw new UserError(
+			"`wrangler deployments view <deployment-id>` is deprecated and will be removed in a future major version. Deployment ID is now referred to as Version ID. Please use `wrangler versions view [version-id] --x-versions` instead."
+		);
```
RamIdeas, line 39: Instead of throwing here, it could log this message as a warning and run `wrangler versions view <version-id>` on behalf of the user

packages/wrangler/src/__tests__/versions/deployments/deployments.view.test.ts
```diff
@@ -0,0 +1,18 @@
+import { runWrangler } from "../../helpers/run-wrangler";
+
+describe("deployments view", () => {
+	test("depracted error when run with no args", async () => {
```
petebacondarwin, line 4:
```suggestion
	test("deprecated error when run with no args", async () => {
```

packages/wrangler/src/__tests__/versions/deployments/deployments.view.test.ts
```diff
@@ -0,0 +1,18 @@
...
+		const result = runWrangler("deployments view  --x-versions");
+
+		await expect(result).rejects.toMatchInlineSnapshot(
+			`[Error: \`wrangler deployments view\` is deprecated and will be removed in a future major version. Please use \`wrangler deployments status --x-versions\` instead.]`
+		);
+	});
+	test("depracted error when run with positional arg", async () => {
```
petebacondarwin, line 11:
```suggestion
	test("deprecated error when run with positional arg", async () => {
```

petebacondarwin (approved):
What is the aim of adding a "new" command that is already deprecated?

Is it to catch people who are used to running a `view` command under a different sub-command?

Although running the "correct" command feels like it is helping users, I would say that in this case an error is better since it prevents users from starting down the wrong path at all, and in fact would allow this fake command to be removed in the future without a breaking change, since the only user facing change would be the replacement of the "helpful" error message with a more generic one.

I think one could actually go further and rather than call this "deprecated" you just say that we are providing a more helpful error message for an invalid command.

packages/wrangler/src/versions/deployments/view.ts
RamIdeas, line 39: @petebacondarwin how about this instead?

```suggestion
	if (args.deploymentId === undefined) {
		throw new UserError(
			"`wrangler deployments view` has been renamed `wrangler deployments status --x-versions`. Please use that command instead."
		);
	} else {
		throw new UserError(
			"`wrangler deployments view <deployment-id>` has been renamed `wrangler versions view [version-id] --x-versions`. Please use that command instead."
		);
	}
```

RamIdeas:
> What is the aim of adding a "new" command that is already deprecated?
> 
> Is it to catch people who are used to running a `view` command under a different sub-command?

Yeah pretty much. You'll note this command is marked as 'hidden' too.

> Although running the "correct" command feels like it is helping users, I would say that in this case an error is better since it prevents users from starting down the wrong path at all, and in fact would allow this fake command to be removed in the future without a breaking change, since the only user facing change would be the replacement of the "helpful" error message with a more generic one.

Thanks for confirming – I agree.

> I think one could actually go further and rather than call this "deprecated" you just say that we are providing a more helpful error message for an invalid command.

Yeah I like this. See my other comment for suggested change.

packages/wrangler/src/versions/deployments/view.ts
petebacondarwin: LGTM!

PR #5411, opened 2024-03-27
[C3] fix: improve error message when no accounts found during deployment

mrbbot:

Ref: CUSTESC-39344

Previously, C3 would display `TypeError: Cannot read properties of undefined (reading 'value')` if you were logged in as a user without access to any accounts. This change ensures a more appropriate error message is displayed in this case.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: it doesn't look like these code paths are tested at the moment. Happy to add some though if they are?
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: improving error messaging

### DIFF
--- a/.changeset/large-pumpkins-try.md
+++ b/.changeset/large-pumpkins-try.md
@@ -1,0 +1,7 @@
+---
+"create-cloudflare": patch
+---
+
+fix: display helpful error message when no accounts found during deployment
+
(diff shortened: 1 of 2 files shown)


dario-piotrowicz (approved):
Looks good to me 🙂 

I could not reproduce the issue locally but the change anyways looks valid to me (as it makes the C3 code more robust anyways)

petebacondarwin (approved):

petebacondarwin:
Thanks @mrbbot - let's land this after the LDW.

PR #5412, opened 2024-03-27
Implement `--json` in `{versions|deployments} list`

### LINKED ISSUE #5408: 🚀 Feature Request: `--output` option for `wrangler versions list/view`
### Describe the solution

Currently, there's no way to get just the `version_id` as output from the `wrangler version list/view` commands. Which means if I were to extract the `version_id` for my CI pipelines, I would need to parse the whole output which comes as, 
```
 ⛅️ wrangler 3.38.0
-------------------
Version ID:  0154fc52-5bd8-4538-af68-a4ac54443b05
Created:     2024-03-27T07:47:52.692Z
Author:      sayantan@ironforge.cloud
Source:      Unknown (version_upload)
Tag:         deploy/1
Message:     -
```

Maybe we could have a `--output json` sorta flag which I can then easily parse for the `version_id`, `tag`, etc.

RamIdeas:

Fixes #5408.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: discoverable in help text


.changeset/thirty-mayflies-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feature: adds `--json` flag to `wrangler deployments list --experimental-versions` and `wrangler deployments status --experimental-versions`, `wrangler versions list --experimental-versions` and `wrangler versions view --experimental-versions` which will format the output as clean JSON. The ` --experimental-versions` flag is still required for these commands.
```
penalosa, line 5:
```suggestion
feature: adds `--json` flag to `wrangler deployments list --experimental-versions`, `wrangler deployments status --experimental-versions`, `wrangler versions list --experimental-versions`, and `wrangler versions view --experimental-versions` which will format the output as clean JSON. The ` --experimental-versions` flag is still required for these commands.
```

packages/wrangler/src/versions/view.ts
```diff
@@ -55,6 +62,11 @@ export async function versionsViewHandler(args: VersionsViewArgs) {
 
 	const version = await fetchVersion(accountId, workerName, args.versionId);
 
+	if (args.json) {
+		logRaw(JSON.stringify(version, null, 2));
```
penalosa, line 66: Why the `null, 2`? In general, I don't think we should be outputting prettified JSON

packages/wrangler/src/versions/deployments/list.ts
```diff
@@ -52,6 +60,12 @@ export async function versionsDeploymentsListHandler(
 	}
 
 	const deployments = await fetchLatestDeployments(accountId, workerName);
+
+	if (args.json) {
+		logRaw(JSON.stringify(deployments, null, 2));
```
penalosa, line 65: Same comment as above

packages/wrangler/src/utils/render-labelled-values.ts
```diff
@@ -67,5 +67,17 @@ export default function formatLabelledValues(
 		);
 	});
 
-	return formattedLines.join(lineSeparator);
+	const output = formattedLines.join(lineSeparator);
+
+	return collapseWhiteSpaceLines(output);
+}
+
+// UTILS
```
penalosa, line 75:
```suggestion

```

penalosa (approved):
Some nits, but mostly LGTM

packages/wrangler/src/versions/view.ts
RamIdeas: Curious why you think we shouldn't output prettified JSON?

Seems to me like there's no benefit in _not_ prettifying it but if a user needed to debug their usage it would be helpful for it to be slightly more readable

PR #5417, opened 2024-03-27
Add Hello World Python template to c3

garrettgu10:

Adds a "Hello World" template to create-cloudflare

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: I don't think this uses any new functionality
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: this is effectively public documentation

### DIFF
--- a/packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
+++ b/packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
@@ -1,0 +1,4 @@
+name = "hello-python-worker"
+main = "src/entry.py"
+compatibility_flags = ["python_workers"]
+compatibility_date = "2024-03-20"
--- a/packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
+++ b/packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
@@ -1,4 +1,4 @@
-name = "hello-python-worker"
+name = "<TBD>"
 main = "src/entry.py"
 compatibility_flags = ["python_workers"]
-compatibility_date = "2024-03-20"
+compatibility_date = "<TBD>"
--- a/.changeset/long-pears-sell.md
+++ b/.changeset/long-pears-sell.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": minor
+---
+
+Add Hello World Python Template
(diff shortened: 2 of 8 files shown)


packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
```diff
@@ -0,0 +1,4 @@
+name = "hello-python-worker"
+main = "src/entry.py"
+compatibility_flags = ["python_workers"]
+compatibility_date = "2024-01-29"
```
kflansburg, line 4: Should we bump this to `2024-03-20`?

packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
```diff
@@ -0,0 +1,4 @@
+name = "hello-python-worker"
```
jculvey, line 1: You should use a placeholder here:

`name = "<TBD>"`

C3 will automatically update this field with the name the user selects in step 1. Otherwise, the user might pick a name like `crimson-sky-1234` but end up with `hello-python-worker` in their `wrangler.toml`

packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
```diff
@@ -0,0 +1,4 @@
...
+main = "src/entry.py"
+compatibility_flags = ["python_workers"]
+compatibility_date = "2024-03-20"
```
jculvey, line 4: I think you might want: 

```suggestion
compatibility_date = "<TBD>"
```

By default, C3 will automatically look up the latest compat date and update this field unless a specific compat date has already been specified. Setting `2024-03-20` explicitly will have the effect of holding at this date even when newer compat dates become available.

jculvey (changes requested):
Do you have the permissions to push branches directly to the workers-sdk repo? If so, it would be better to PR from that rather than a fork. Otherwise, the builds won't be run with the secrets required to have the e2e tests run cleanly.

Also, I think it would be a good idea to add an e2e test for this template [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/create-cloudflare/e2e-tests/workers.test.ts).

packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
irvinebroque:
```suggestion
name = "<TBD>"
```

garrettgu10:
Redone in #5475

PR #5419, opened 2024-03-27
add helpful logging to --experimental-versions commands

RamIdeas:

This PR adds helpful logging to `--experimental-versions` commands.

The changes in packages/cli are a refactor and do not effect the output by default but now there are options to override the default formatting.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: chore

### DIFF
--- a/packages/cli/index.ts
+++ b/packages/cli/index.ts
@@ -67,16 +67,24 @@ export const newline = () => {
 	log("");
 };
 
+type FormatOptions = {
+	linePrefix?: string;
+	firstLinePrefix?: string;
+	newlineBefore?: boolean;
+	newlineAfter?: boolean;
+	formatLine?: (line: string) => string;
+	multiline?: boolean;
+};
 export const format = (
 	msg: string,
 	{
 		linePrefix = gray(shapes.bar),
-		firstLinePrefix = gray(shapes.bar),
+		firstLinePrefix = linePrefix,
 		newlineBefore = false,
 		newlineAfter = false,
 		formatLine = (line: string) => white(line),
 		multiline = true,
-	}
+	}: FormatOptions = {}
 ) => {
 	const lines = multiline ? msg.split("\n") : [msg];
 	const formattedLines = lines.map(
--- a/packages/cli/index.ts
+++ b/packages/cli/index.ts
@@ -40,6 +40,7 @@ export const status = {
 	warning: bgYellow(` WARNING `),
 	info: bgBlue(` INFO `),
 	success: bgGreen(` SUCCESS `),
+	cancel: white.bgRed(` X `),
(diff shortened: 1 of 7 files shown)


packages/cli/index.ts
```diff
@@ -66,18 +67,38 @@ export const newline = () => {
...
-		.slice(1)
-		.map((ln) => `${gray(shapes.bar)} ${white(ln)}`);
-	logRaw(`${gray(shapes.leftT)} ${lines[0]}`);
-	if (restLines.length) {
-		logRaw(restLines.join("\n"));
+export const format = (
+	msg: string,
+	{
+		linePrefix = gray(shapes.bar),
+		firstLinePrefix = gray(shapes.bar),
```
mrbbot, line 74: Should this default to `linePrefix`? Maybe make it `undefined` by default?

packages/wrangler/src/versions/upload.ts
```diff
@@ -499,6 +500,22 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
+	const cmdTriggersDeploy = blue(
+		"wrangler triggers deploy --experimental-versions"
+	);
+	logger.info(
+		gray(`
+To deploy this version to production traffic use the command ${cmdVersionsDeploy}
+
+NOTE: Changes to non-versioned settings (config properties 'logpush' or 'tail_consumers') take effect after your next deployment. Use the command ${cmdVersionsDeploy} to deploy these changes
+
+NOTE: Changes to triggers (routes, custom domains, cron schedules, etc) must be applied with the command ${cmdTriggersDeploy}
```
mrbbot, line 516: I feel like you don't need the `NOTE`s here, and you could rephrase this a little to only include the deploy command once?
```suggestion
To deploy this version to production traffic use the command \`${cmdVersionsDeploy}\`. 

Changes to non-versioned settings (config properties 'logpush' or 'tail_consumers') take effect after your next deployment.

Changes to triggers (routes, custom domains, cron schedules, etc) must be applied with the command \`${cmdTriggersDeploy}\`
```

mrbbot (approved):
Looks good! Very minor non-blocking comments, but feel free to merge.

packages/wrangler/src/versions/upload.ts
RamIdeas: Removed note. Kept the repeated command as I think there's functional value to that part of the copy.

We can remove later if Meaghan confirms it's unnecessary

PR #5421, opened 2024-03-27
Implement `version_metadata` binding

RamIdeas:

This PR implements the `version_metadata` binding. This binding already exists in the platform and has been used so far as an "unsafe.binding".

The interface of the binding is a plain object with two string properties: `id` and `tag`. This shape has been emulated locally via a json var binding.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:


WalshyDev:
Out of interest, why are we preferring vresion_metadata over metadata? What's the path for those currently using metadata?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -360,7 +360,13 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 	}
 
 	const bindingOptions = {
-		bindings: bindings.vars,
+		bindings: {
+			...bindings.vars,
+			// emulate version_metadata binding via a JSON var
+			...(bindings.version_metadata
+				? { [bindings.version_metadata.binding]: { id: randomUUID(), tag: "" } }
```
penalosa, line 367: This feels like it should be in miniflare proper—why did you choose to emulate it with a JSON binding?

packages/wrangler/src/versions/upload.ts
```diff
@@ -390,6 +391,8 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			},
 		};
 
+		console.log(worker.bindings);
```
penalosa, line 394:
```suggestion

```

packages/wrangler/src/versions/version-metadata-binding.ts
```diff
@@ -0,0 +1 @@
+export default function version_metadata() {}
```
penalosa, line 1: What's the purpose of this file?

fixtures/worker-app/src/index.js
penalosa, line 1: Could you add a pages fixture test, as well?

penalosa (commented):
This could probably do with some more testing around deploy and Pages

packages/wrangler/src/dev/miniflare.ts
mrbbot: I think this is probably fine. There's no real functionality to emulate here, and it would be very easy for users to include a binding like this themselves if they were using the `Miniflare` API. We already have bindings like AI that don't have any specific Miniflare implementation, although I guess that's more of a special case because it requires authentication.

packages/wrangler/src/versions/version-metadata-binding.ts
RamIdeas: Accidental commitage

fixtures/worker-app/src/index.js
RamIdeas: https://github.com/cloudflare/workers-sdk/pull/5421/commits/15dcacb8d579d80388f2aa7577e9c86fdcdfdeaf

penalosa (approved):

PR #5426, opened 2024-03-28
Implement `triggers deploy` command

RamIdeas:

This PR implements the `wrangler triggers deploy` command behind the `--experimental-versions` flag.

I extracted the logic from `wrangler deploy` which I've now made call into the extracted function. The test coverage for `wrangler deploy` should cover `wrangler triggers deploy`.

The only difference this makes to `wrangler deploy` is:
* the log line "Published <worker-name> (<timing>)" is now "Deployed <worker-name> triggers (<timing>)"
* and, instead of getting the value `available_on_subdomain` from the response of the PUT script call, it now makes an additional GET call to ".../subdomain" to get the value

It is possible to mitigate these two differences but I don't think the log line change or minimal added latency necessitates it. But I'm open to others' opinions :)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:


.changeset/big-pumas-count.md
```diff
@@ -0,0 +1,11 @@
...
+"wrangler": minor
+---
+
+feature: added a new `wrangler triggers deploy` command
+
+This command currently requires the `--experimental-versions` flag.
+
+This command extracts the trigger deployment logic from `wrangler deploy` and allows users to update their currently deployed Worker's triggers without doing another deployment. This is primarily useful for users of `wrangler versions upload` and `wrangler versions deploy` who can then run `wrangler triggers deploy` to apply trigger changes to their currently deployed Worker Versions.
+
+The command can also be used even if not using the `wrangler versions ...` commands. And, in fact, are already using it implicitly when running `wrangler deploy`.
```
penalosa, line 11: This is missing a word. Maybe "*developers* are already using it implicitly when running `wrangler deploy`?

packages/wrangler/src/__tests__/deploy.test.ts
penalosa, line 1: These snapshots should not be changed, but I assume this has been fixed in a followup?

packages/wrangler/src/__tests__/deprecated-usage-model.test.ts
penalosa, line 1: Same comment as for `deploy.test.ts`

packages/wrangler/src/triggers/index.ts
```diff
@@ -0,0 +1,88 @@
...
+		.option("legacy-env", {
+			type: "boolean",
+			describe: "Use legacy environments",
+			hidden: true,
+		})
+		.option("dispatch-namespace", {
+			describe:
+				"Name of a dispatch namespace to deploy the Worker to (Workers for Platforms)",
+			type: "string",
+		});
```
penalosa, line 56: I realise these options are copied from `deploy`, but this argument doesn't make sense for triggers

packages/wrangler/src/triggers/index.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export async function triggersDeployHandler(
+	args: StrictYargsOptionsToInterface<typeof triggersDeployOptions>
+) {
+	await printWranglerBanner();
+
+	const configPath = args.config || findWranglerToml();
+	const config = readConfig(configPath, args);
```
penalosa, line 65: You don't need to resolve the config path outside of `readConfig()`. The first argument should be `undefined` here

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -0,0 +1,256 @@
...
+			if (route.pattern.includes("/")) {
+				throw new UserError(
+					`Cannot use "${route.pattern}" as a Custom Domain; paths are not allowed`
+				);
+			}
+			customDomainsOnly.push(route);
+		} else {
+			routesOnly.push(route);
+		}
+	}
```
penalosa, line 57: You've added logic for publishing routes here, but haven't removed it from `wrangler deploy`. Are you sure this won't cause duplicate deploys?

packages/wrangler/src/triggers/index.ts
RamIdeas: Fixed in https://github.com/cloudflare/workers-sdk/pull/5436

packages/wrangler/src/triggers/index.ts
RamIdeas: Fixed in https://github.com/cloudflare/workers-sdk/pull/5436

packages/wrangler/src/triggers/deploy.ts
RamIdeas: Definitely not causing duplicate deploys. This was left for the validation. The arrays are dead code which have been removed in https://github.com/cloudflare/workers-sdk/pull/5436

packages/wrangler/src/__tests__/deploy.test.ts
RamIdeas: Yes fixed in 006f1aefc48250c636160bf6b907bcf2de4cab2e c234dadea496ce1abb36353c63dd19cdbb602827

packages/wrangler/src/__tests__/deprecated-usage-model.test.ts
RamIdeas: Yes fixed in 006f1aefc48250c636160bf6b907bcf2de4cab2e c234dadea496ce1abb36353c63dd19cdbb602827

.changeset/big-pumas-count.md
RamIdeas: Fixed in https://github.com/cloudflare/workers-sdk/pull/5436

lrapoport-cf:
for posterity, i am approving this PR -- the concerns raised are addressed in the linked PRs/commits, which themselves have been approved

PR #5429, opened 2024-03-28
Add Wrangler support for R2 storage classes

ocsfrank:

Not exempt from LDW, but I would like to get this reviewed before next week.

Fixes R2-1963

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: pre-release feature. Docs are planned for next week and will be retro-actively linked here @jonesphillip

### DIFF
--- a/.changeset/unlucky-laws-fetch.md
+++ b/.changeset/unlucky-laws-fetch.md
@@ -1,0 +1,15 @@
+---
+"wrangler": minor
+---
+
+R2 will introduce storage classes soon. Wrangler allows you to interact with storage classes once it is
+enabled on your account.
+
+Wrangler supports an `-s` flag that allows the user to specify a storage class when creating a bucket,
+changing the default storage class of a bucket, and uploading an object.
+
+```bash
+wrangler r2 bucket create ia-bucket -s InfrequentAccess
+wrangler r2 bucket update storage-class my-bucket -s InfrequentAccess
+wrangler r2 object put bucket/ia-object -s InfrequentAccess --file foo
+```
--- a/packages/wrangler/src/__tests__/r2.test.ts
+++ b/packages/wrangler/src/__tests__/r2.test.ts
@@ -6,7 +6,7 @@ import { actionsForEventCategories } from "../r2/helpers";
 import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
 import { mockConsoleMethods } from "./helpers/mock-console";
 import { useMockIsTTY } from "./helpers/mock-istty";
-import { createFetchResult, msw, mswSuccessR2handlers } from "./helpers/msw";
+import { createFetchResult, msw, mswR2handlers } from "./helpers/msw";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { runWrangler } from "./helpers/run-wrangler";
 import type {
@@ -18,7 +18,7 @@ import type {
 
 describe("r2", () => {
 	const std = mockConsoleMethods();
-	beforeEach(() => msw.use(...mswSuccessR2handlers));
+	beforeEach(() => msw.use(...mswR2handlers));
 
 	runInTempDir();
 
@@ -43,6 +43,7 @@ describe("r2", () => {
 
 			Commands:
 			  wrangler r2 bucket create <name>  Create a new R2 bucket
+			  wrangler r2 bucket update         Update bucket state
 			  wrangler r2 bucket list           List R2 buckets
 			  wrangler r2 bucket delete <name>  Delete an R2 bucket
 			  wrangler r2 bucket sippy          Manage Sippy incremental migration on an R2 bucket
@@ -121,7 +122,8 @@ describe("r2", () => {
 			  -v, --version                   Show version number  [boolean]
 
 			Options:
-			  -J, --jurisdiction  The jurisdiction where the new bucket will be created  [string]"
+			  -J, --jurisdiction   The jurisdiction where the new bucket will be created  [string]
+			  -s, --storage-class  The default storage class for objects uploaded to this bucket  [string]"
 		`);
 				expect(std.err).toMatchInlineSnapshot(`
 				            "[31mX [41;31m[[41;97mERROR[41;31m][0m [1mNot enough non-option arguments: got 0, need at least 1[0m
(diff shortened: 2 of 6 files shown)


packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -42,11 +42,12 @@ describe("r2", () => {
 			Manage R2 buckets
 
 			Commands:
-			  wrangler r2 bucket create <name>  Create a new R2 bucket
-			  wrangler r2 bucket list           List R2 buckets
-			  wrangler r2 bucket delete <name>  Delete an R2 bucket
-			  wrangler r2 bucket sippy          Manage Sippy incremental migration on an R2 bucket
-			  wrangler r2 bucket notification   Manage event notifications for an R2 bucket
+			  wrangler r2 bucket create <name>             Create a new R2 bucket
+			  wrangler r2 bucket put-storage-class <name>  Update the default storage class of an existing R2 bucket
```
sdnts, line 46: Have you considered making a `put` subcommand so we can perhaps nest other bucket updates under it?
`wrangler r2 bucket put storage-class <name>`

ocsfrank: Good idea -- we indeed should.

ocsfrank: Added as a subcommand.

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -196,12 +199,123 @@ describe("r2", () => {
...
+			});
+
+			it("should create a bucket with the expected default storage class", async () => {
+				msw.use(
+					rest.post(
+						"*/accounts/:accountId/r2/buckets",
+						async (request, response, context) => {
+							const { accountId } = request.params;
+							expect(accountId).toEqual("some-account-id");
+							expect(request.headers.get("cf-r2-storage-class")).toEqual(
```
petebacondarwin, line 214: I know we do this all over the place, but really we should not be using `expect` checks inside the mock service worker handlers.
Instead, we should simulate what the backend would do, which I assume would be some kind of 4xx response in the case that the storage class is invalid?

Then in these tests we can simply check that Wrangler does not error itself, which proves that it must have passed the correct header.
Equally we can check that Wrangler handles the error response correctly if the header is invalid.

.changeset/unlucky-laws-fetch.md
```diff
@@ -0,0 +1,15 @@
...
+
+R2 will introduce storage classes soon. Wrangler allows you to interact with storage classes once it is
+enabled on your account.
+
+Wrangler supports an `-s` flag that allows the user to specify a storage class when creating a bucket,
+changing the default storage class of a bucket, and uploading an object.
+
+```bash
+wrangler r2 bucket create ia-bucket -s InfrequentAccess
+wrangler r2 bucket put-storage-class my-bucket -s InfrequentAccess
```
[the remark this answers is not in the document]
petebacondarwin: Oh from looking at the code it should be:

```suggestion
wrangler r2 bucket put storage-class my-bucket -s InfrequentAccess
```

The extra noun "put" seems a bit off here. The object of the command is the `bucket` still and the operation is `update-storage-class`, no?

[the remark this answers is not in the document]
petebacondarwin: So I would imagine this command should be:

```suggestion
wrangler r2 bucket update-storage-class my-bucket -s InfrequentAccess
```

packages/wrangler/src/r2/helpers.ts
```diff
@@ -48,19 +48,47 @@ export async function listR2Buckets(
...
 		body: JSON.stringify({ name: bucketName }),
 		headers,
 	});
 }
 
+/**
+ * Update the default storage class to `storageClass` of a bucket with the given `bucketName`
+ * within the account given by `accountId`.
+ */
+export async function putR2BucketStorageClass(
```
petebacondarwin, line 72: `put...` seems like the wrong name here?
This is updating the storage class using a `PATCH` verb not `PUT`.
It feels like "put" is being used because it was copied over from when we actually "PUT" to the REST API?

petebacondarwin (changes requested):

.changeset/unlucky-laws-fetch.md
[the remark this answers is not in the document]
ocsfrank: Ah, sorry -- that's my bad. I updated the command itself, but not the changeset file. The reason for using `put` even though it's a `PATCH` request is because the action itself is called `PutBucketStorageClass` in the API to be consistent with other operations that update a bucket settings (`PutBucketLifecycleConfiguration`, `PutBucketCors`, etc.). So, this is why there's a `put` subcommand as we'd also like to expand Wrangler to include the other bucket actions.

packages/wrangler/src/r2/helpers.ts
ocsfrank: It's because of the naming convention in the R2 API itself (https://github.com/cloudflare/workers-sdk/pull/5429#discussion_r1546741564).

packages/wrangler/src/__tests__/r2.test.ts
ocsfrank: Good point, going to update the test.

.changeset/unlucky-laws-fetch.md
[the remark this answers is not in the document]
ocsfrank: Although, looking at this again, I do agree that `update` makes it much more intuitive from a user perspective as the operations **is** a `PATCH` request. Updating this to replace the `put` subcommand with `update` subcommand.

ocsfrank:
Updated to using `update storage-class` instead of `put storage-class` as it's more intuitive. Also updated the tests to mimic the response from R2.

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -850,6 +975,18 @@ describe("r2", () => {
 		`);
 			});
 
+			it("should upload R2 object with storage class to bucket", async () => {
```
ocsfrank, line 978: Does this test run against real R2? If that's the case, we'll have to remove it as it's currently private preview.

petebacondarwin (approved):

packages/wrangler/src/__tests__/r2.test.ts
ocsfrank: Ah, it doesn't look like it -- it's using mock api and token. I think we're good.

PR #5449, opened 2024-03-29
Handle invalid wrangler.toml files

penalosa:

Handle non-Pages wrangler.toml files better. This PR ensures that `wrangler pages functions build-env` and `wrangler pages functions build` fail fast with the appropriate error message if a non-pages wrangler.toml file is found. This ensure that config validation isn't run, which was inadvertently throwing errors for non-pages wrangler.toml files.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Not yet documented

### DIFF
--- a/packages/wrangler/src/config/index.ts
+++ b/packages/wrangler/src/config/index.ts
@@ -5,7 +5,7 @@ import { FatalError, UserError } from "../errors";
 import { logger } from "../logger";
 import { EXIT_CODE_INVALID_PAGES_CONFIG } from "../pages/errors";
 import { parseJSONC, parseTOML, readFileSync } from "../parse";
-import { normalizeAndValidateConfig } from "./validation";
+import { isPagesConfig, normalizeAndValidateConfig } from "./validation";
 import { validatePagesConfig } from "./validation-pages";
 import type { CfWorkerInit } from "../deployment-bundle/worker";
 import type { CommonYargsOptions } from "../yargs-types";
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/config/index.ts
```diff
@@ -30,7 +31,8 @@ export function readConfig<CommandArgs>(
 	configPath: string | undefined,
 	// Include command specific args as well as the wrangler global flags
 	args: CommandArgs &
-		Pick<OnlyCamelCase<CommonYargsOptions>, "experimentalJsonConfig">
+		Pick<OnlyCamelCase<CommonYargsOptions>, "experimentalJsonConfig">,
+	requirePagesConfig?: boolean
```
petebacondarwin, line 35: And with this new parameter our journey to the darkside is complete as we have ended up leaking the pages specific stuff into the general wrangler config stuff. 😢

packages/wrangler/src/config/index.ts
```diff
@@ -55,6 +57,12 @@ export function readConfig<CommandArgs>(
 	 * so we should error if `wrangler.json` is detected in a Pages project
 	 */
 	const isPagesConfig = rawConfig.pages_build_output_dir !== undefined;
```
petebacondarwin, line 59: By the way, there is actually a helper function for this in `validation.ts` - `isPagesConfig(rawConfig)` - which we should use.

packages/wrangler/src/config/index.ts
petebacondarwin: Let's land this as-is since it does fix the problem for DevWeek. But I would love it if we could circle back and have another look at how to make this stuff a bit less convoluted.

petebacondarwin (approved):

packages/wrangler/src/config/index.ts
penalosa: Yeah, I don't love this, but it fixes the problem for now. We should definitely revisit, but I think there might be some pieces here that we fundamentally can't pull apart, if we want to maintain a single `readConfig()` interface.

petebacondarwin:
And don't forget the changeset...

PR #5458, opened 2024-03-31
chore: bump vitest to 1.5.0 and loosen version requirement for vitest-pool-workers

### LINKED ISSUE #5615: 🚀 Feature Request: Update vitest dependency to 1.5 to support running and debugging tests directly from VSCode
### Describe the solution

When I try to run my vitest tests directly from Visual Studio Code with `@cloudflare/vitest-pool-workers` version `0.1.19` I get following error:
Screenshot 2024-04-14 at 13 37 06

If you would update the vitest dependency to at least 1.4 it would allow running tests directly with the vitest VSCode integration.

Cherry:

Bumps vitest to 1.5.0, to keep this inline with latest changes. All tests pass locally for me, and I tested this on a couple of my own projects without issue.

Closes #5615 

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/13706
  - [ ] Not necessary because:

### DIFF
--- a/packages/vitest-pool-workers/package.json
+++ b/packages/vitest-pool-workers/package.json
@@ -70,9 +70,9 @@
 		"zod": "^3.20.6"
 	},
 	"peerDependencies": {
-		"@vitest/runner": "1.4.0",
-		"@vitest/snapshot": "1.4.0",
-		"vitest": "1.4.0"
+		"@vitest/runner": "1.5.0",
+		"@vitest/snapshot": "1.5.0",
+		"vitest": "1.5.0"
 	},
 	"workers-sdk": {
 		"prerelease": true
--- a/packages/vitest-pool-workers/package.json
+++ b/packages/vitest-pool-workers/package.json
@@ -56,6 +56,7 @@
 		"@cloudflare/workers-tsconfig": "workspace:*",
 		"@cloudflare/workers-types": "^4.20240419.0",
 		"@types/node": "20.8.3",
+		"@types/semver": "^7.5.1",
 		"capnp-ts": "^0.7.0",
 		"capnpc-ts": "^0.7.0",
 		"undici": "5.28.3"
@@ -66,13 +67,14 @@
 		"devalue": "^4.3.0",
 		"esbuild": "0.17.19",
 		"miniflare": "workspace:*",
+		"semver": "^7.5.1",
 		"wrangler": "workspace:*",
 		"zod": "^3.20.6"
 	},
 	"peerDependencies": {
-		"@vitest/runner": "1.5.0",
-		"@vitest/snapshot": "1.5.0",
-		"vitest": "1.5.0"
+		"@vitest/runner": "1.3.x - 1.5.x",
+		"@vitest/snapshot": "1.3.x - 1.5.x",
+		"vitest": "1.3.x - 1.5.x"
 	},
 	"workers-sdk": {
 		"prerelease": true
--- a/packages/vitest-pool-workers/package.json
+++ b/packages/vitest-pool-workers/package.json
@@ -70,9 +70,9 @@
 		"zod": "^3.20.6"
 	},
 	"peerDependencies": {
-		"@vitest/runner": "1.3.0",
-		"@vitest/snapshot": "1.3.0",
-		"vitest": "1.3.0"
+		"@vitest/runner": "1.4.0",
+		"@vitest/snapshot": "1.4.0",
+		"vitest": "1.4.0"
 	},
 	"workers-sdk": {
 		"prerelease": true
--- a/packages/vitest-pool-workers/src/pool/index.ts
+++ b/packages/vitest-pool-workers/src/pool/index.ts
@@ -56,7 +56,7 @@ import type {
 } from "vitest";
 import type { ProcessPool, Vitest, WorkspaceProject } from "vitest/node";
 
-// https://github.com/vitest-dev/vitest/blob/v1.4.0/packages/vite-node/src/client.ts#L386
+// https://github.com/vitest-dev/vitest/blob/v1.5.0/packages/vite-node/src/client.ts#L386
 declare const __vite_ssr_import__: unknown;
 assert(
 	typeof __vite_ssr_import__ === "undefined",
--- a/packages/vitest-pool-workers/src/pool/index.ts
+++ b/packages/vitest-pool-workers/src/pool/index.ts
@@ -20,6 +20,7 @@ import {
 	testRegExps,
 	WebSocket,
 } from "miniflare";
+import semverSatisfies from "semver/functions/satisfies.js";
 import { createMethodsRPC } from "vitest/node";
 import { createChunkingSocket } from "../shared/chunking-socket";
 import { OPTIONS_PATH, parseProjectOptions } from "./config";
@@ -775,13 +776,13 @@ function assertCompatibleVitestVersion(ctx: Vitest) {
 		"Expected to find `vitest`'s version"
 	);
 
-	if (expectedVitestVersion !== actualVitestVersion) {
+	if (!semverSatisfies(actualVitestVersion, expectedVitestVersion)) {
 		const message = [
-			`You're running \`vitest@${actualVitestVersion}\`, but this version of \`@cloudflare/vitest-pool-workers\` only supports \`vitest@${expectedVitestVersion}\`.`,
+			`You're running \`vitest@${actualVitestVersion}\`, but this version of \`@cloudflare/vitest-pool-workers\` only officially supports \`vitest ${expectedVitestVersion}\`.`,
 			"`@cloudflare/vitest-pool-workers` currently depends on internal Vitest APIs that are not protected by semantic-versioning guarantees.",
-			`Please install \`vitest@${expectedVitestVersion}\` to continue using \`@cloudflare/vitest-pool-workers\`.`,
+			`Your tests may work without issue, but we can not guarantee compatibility outside of the above version range.`,
 		].join("\n");
-		throw new Error(message);
+		log.warn(message);
 	}
 }
 
--- a/packages/vitest-pool-workers/src/pool/index.ts
+++ b/packages/vitest-pool-workers/src/pool/index.ts
@@ -56,7 +56,7 @@ import type {
 } from "vitest";
 import type { ProcessPool, Vitest, WorkspaceProject } from "vitest/node";
 
-// https://github.com/vitest-dev/vitest/blob/v1.3.0/packages/vite-node/src/client.ts#L386
+// https://github.com/vitest-dev/vitest/blob/v1.4.0/packages/vite-node/src/client.ts#L386
 declare const __vite_ssr_import__: unknown;
 assert(
 	typeof __vite_ssr_import__ === "undefined",
--- a/.changeset/brave-numbers-wash.md
+++ b/.changeset/brave-numbers-wash.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vitest-pool-workers": patch
+---
+
+chore: bump vitest dependency to version 1.4.0
--- a/.changeset/brave-numbers-wash.md
+++ b/.changeset/brave-numbers-wash.md
@@ -2,4 +2,4 @@
 "@cloudflare/vitest-pool-workers": patch
 ---
 
-chore: bump vitest dependency to version 1.5.0
+fix: loosen the peer dependency version on vitest to support versions ranging from 1.3.0 to 1.5.0
--- a/.changeset/brave-numbers-wash.md
+++ b/.changeset/brave-numbers-wash.md
@@ -2,4 +2,4 @@
 "@cloudflare/vitest-pool-workers": patch
 ---
 
-chore: bump vitest dependency to version 1.4.0
+chore: bump vitest dependency to version 1.5.0
--- a/fixtures/vitest-pool-workers-examples/package.json
+++ b/fixtures/vitest-pool-workers-examples/package.json
@@ -15,7 +15,7 @@
 		"miniflare": "workspace:*",
 		"toucan-js": "^3.3.1",
 		"typescript": "^5.3.3",
-		"vitest": "1.3.0",
+		"vitest": "1.4.0",
 		"wrangler": "workspace:*"
 	}
 }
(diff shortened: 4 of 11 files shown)


Cherry:
While everything appears to be working fine, there's an odd type conflict it seems in: `fixtures\vitest-pool-workers-examples\hyperdrive`:
```
hyperdrive/global-setup.ts:19:10 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.

19  provide("echoServerPort", port);
            ~~~~~~~~~~~~~~~~

hyperdrive/vitest.config.ts:9:35 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.

9     const echoServerPort = inject("echoServerPort");
       
```

Cherry:
I'm not sure what the issues are with the C3 E2E tests, but most everything else appears to be fine. If anyone has any insight here, that would be awesome.

johtso:
@Cherry excited that this is happening! Do you know of any way I can take it for a spin in a local project? Not having much luck installing from github because of it being a monorepo. Tried using https://gitpkg.vercel.app but get ` ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In ../..: "miniflare@workspace:*" is in the dependencies but no package named "miniflare" is present in the workspace`

Cherry:
@johtso If you add:

```json
"@cloudflare/vitest-pool-workers": "https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/8561120366/npm-package-cloudflare-vitest-pool-workers-5458"
```

Or just run:
`npm install https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/8561120366/npm-package-cloudflare-vitest-pool-workers-5458 -D`

to your `package.json` dev dependencies, you should be able to test with this.

jculvey (approved):
C3 changes lgtm

johtso:
> @johtso If you add:
> 
> ```json
> "@cloudflare/vitest-pool-workers": "https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/8561120366/npm-package-cloudflare-vitest-pool-workers-5458"
> ```
> 
> Or just run: `npm install https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/8561120366/npm-package-cloudflare-vitest-pool-workers-5458 -D`
> 
> to your `package.json` dev dependencies, you should be able to test with this.

This is exactly what I was looking for, thank you so much!

jculvey:
> I'm not sure what the issues are with the C3 E2E tests, but most everything else appears to be fine. If anyone has any insight here, that would be awesome.

The issue is that the C3 E2E tests run actual deployments and need credentials to run properly. When PRs are made from forks, the `CLOUDFLARE_ACCOUNT_ID` and `CLOUDFLARE_API_TOKEN` secrets are unset which makes the tests fail.

Cherry:
Updated this for 1.5.0. Tested and all seems to be working well.

packages/vitest-pool-workers/package.json
```diff
@@ -70,9 +70,9 @@
 		"zod": "^3.20.6"
 	},
 	"peerDependencies": {
-		"@vitest/runner": "1.3.0",
-		"@vitest/snapshot": "1.3.0",
-		"vitest": "1.3.0"
+		"@vitest/runner": "1.5.0",
+		"@vitest/snapshot": "1.5.0",
+		"vitest": "1.5.0"
```
petebacondarwin, line 75: Should we make these peer dependencies a bit more lenient?
I feel like bumping these strict dependencies is a breaking change to the library, forcing developers to update to 1.5.0 even if they would prefer to stay on 1.3.0 or 1.4.0.

The reason why this was not originally set to `^1.3.0` was that we are using some internal APIs that could have a breaking change within a minor release.
But that being said, the Vitest team are aware of what APIs we are using and are onboard with these becoming public and stabilized.

We should fast-track that stabilization but in the meantime, should we be more lenient here?

Perhaps we could go with:

```suggestion
		"@vitest/runner": "1.3.x||1.4.x||1.5.x",
		"@vitest/snapshot": "1.3.x||1.4.x||1.5.x",
		"vitest": "1.3.x||1.4.x||1.5.x"
```

petebacondarwin (changes requested):

johtso:
Also maybe consider warning the developer rather than throwing an error if they choose to install a newer version of vitest than this packages has been tested with?

petebacondarwin:
> Also maybe consider warning the developer rather than throwing an error if they choose to install a newer version of vitest than this packages has been tested with?

I believe most package managers will only warn (not error) if you have installed a peer dependency that is outside the range given.

johtso:
> I believe most package managers will only warn (not error) if you have installed a peer dependency that is outside the range given.

The package currently explicitly asserts that the installed vitest version matches the dependency version: [assertCompatibleVitestVersion](https://github.com/cloudflare/workers-sdk/blob/21878f5338e4736fb263f50e4b5ac3856cade28f/packages/vitest-pool-workers/src/pool/index.ts#L753)

petebacondarwin:
Cripes! OK. So I think we probably should relax that a bit. I think a warning should be enough.

packages/vitest-pool-workers/package.json
Cherry: Thanks Pete! I'm very happy to do this, and tweak the specific version enforcer to just yell loudly instead of exiting entirely. Something like `>=1.3.0 || <= 1.5.x` might make this version a little easier to maintain in future as 1.6.0 gets released, etc.

Cherry:
Done! Here's what the warning now looks like:



Very happy to update this if someone has any better wording.

packages/vitest-pool-workers/package.json
Cherry: Done!

packages/vitest-pool-workers/package.json
```diff
@@ -66,13 +67,14 @@
...
 		"wrangler": "workspace:*",
 		"zod": "^3.20.6"
 	},
 	"peerDependencies": {
-		"@vitest/runner": "1.3.0",
-		"@vitest/snapshot": "1.3.0",
-		"vitest": "1.3.0"
+		"@vitest/runner": "1.3.x - 1.5.x",
+		"@vitest/snapshot": "1.3.x - 1.5.x",
+		"vitest": "1.3.x - 1.5.x"
```
petebacondarwin, line 77: TIL - hyphenated ranges!

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -20,6 +20,7 @@ import {
 	testRegExps,
 	WebSocket,
 } from "miniflare";
+import semverSatisfies from "semver/functions/satisfies.js";
```
petebacondarwin, line 23: Why is there a `.js` on the end of this module specifier?

petebacondarwin (approved):

packages/vitest-pool-workers/package.json
Cherry: Same! `>=1.3.0 || <= 1.5.x` didn't work as I expected, because `1.2.0` satisfied this, as it's < 1.5.x, but hyphenated worked great!

packages/vitest-pool-workers/src/pool/index.ts
Cherry: I'm not entirely sure, but I assume it's just due to how `semver` is built and published in a very CJS fashion?

Without it, I got an error like this:
```
Error: Cannot find module 'E:\xxx\node_modules\semver\functions\satisfies' imported from E:\xxx\node_modules\@cloudflare\vitest-pool-workers\dist\pool\index.mjs
Did you mean to import semver/functions/satisfies.js?
 ❯ finalizeResolution node:internal/modules/esm/resolve:264:11
 ❯ moduleResolve node:internal/modules/esm/resolve:917:10
 ❯ defaultResolve node:internal/modules/esm/resolve:1130:11
 ❯ ModuleLoader.defaultResolve node:internal/modules/esm/loader:396:12
 ❯ ModuleLoader.resolve node:internal/modules/esm/loader:365:25
 ❯ ModuleLoader.getModuleJob node:internal/modules/esm/loader:240:38
 ❯ ModuleWrap.<anonymous> node:internal/modules/esm/module_job:85:39
 ❯ link node:internal/modules/esm/module_job:84:36
```

And when swapping over to `semver/functions/satisfies.js` everything was happy, so I didn't dig any further.

Cherry: Another option would be to just `import semver from 'semver'` and then using `semver.satisfies`, if that's preferred. Let me know.

Cherry:
macOS tests keep flaking with `RangeError: WebAssembly.instantiate(): Out of memory: wasm memory`, but otherwise this is good to merge. Not sure what else can be done that other than keep trying to run until it passes 😅

I think this might be not helped by a change to GitHub Actions earlier this month: https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/

Looking at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories, this now seems to mean the `macos-latest` is running macOS 14, on M1, but with half of the RAM than previously.

Cherry:
Rebased for lockfile changes and all tests are passing now. 😀

punkeel (approved):

DaniFoldi:
Hey @petebacondarwin,
I see that you have been rebasing this branch and looking at e2e tests.

Vitest 1.6.0 was released yesterday, and there are some fixes in recent versions that we depend on. Do you think you'll have a chance to land this soon? Many people on Discord are also asking about updates, and a link to this PR is the only answer we can give, unfortunately.

Ps. happy to take a look at 1.6.0 to see if it breaks something, but even 1.5.x would be great.

Cherry:
I tried to bump to v1.6.0 but hit a roadblock that I believe to be caused by this change:

https://github.com/vitest-dev/vitest/commit/d8304bb4fbe16285d014f63aa71ef9969865c691#diff-2a750ef38c91574e0278bc98b00b991a28ad7e03949fd6574bf8bb049615781cR33

After updating, all tests were failing with `Error: readFileSync() is not yet implemented in Workers`.

I believe this is due to `readFileSync` now actually being used in the vitest runner, which has since been stubbed here to error as per: https://github.com/cloudflare/workers-sdk/blob/027f971975a48a564603275f3583d21e9d053229/packages/vitest-pool-workers/src/worker/lib/node/fs.ts

I'm not familiar enough with vitest internals or this code to dive much deeper in stubbing/patching this, but using 1.5.x works great and would be good to at least get that merged in soon until someone else can take a look at 1.6.x.

petebacondarwin:
I'm trying to land it but the tests are not playing along. I'll try to dig in and sort it this coming week

Cherry:
Anything I can do to help here @petebacondarwin? It seems the standard tests keep passing fine, and it's really just the C3 ones which are failing?

petebacondarwin:
Let's give it one more go and then I'll dig in a bit more deeply.
It is possible to run the C3 e2e tests locally, which I will try if this doesn't go well

petebacondarwin:
Looks like the issues are npm ones:

```
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: tmp-e2e-c3e2da7571-hello-world-js@0.0.0
npm ERR! Found: vitest@1.5.0
npm ERR! node_modules/vitest
npm ERR!   dev vitest@"1.5.0" from the root project
npm ERR!   peer vitest@"1.5.0" from @vitest/browser@1.5.0
npm ERR!   node_modules/@vitest/browser
npm ERR!     peerOptional @vitest/browser@"1.5.0" from vitest@1.5.0
npm ERR!   1 more (@vitest/ui)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer vitest@"1.3.0" from @cloudflare/vitest-pool-workers@0.1.19
npm ERR! node_modules/@cloudflare/vitest-pool-workers
npm ERR!   dev @cloudflare/vitest-pool-workers@"^0.1.0" from the root project
```

petebacondarwin:
I think we need to bump this to "^0.2.7": https://github.com/cloudflare/workers-sdk/blob/f673c66373e2acd8d9cc94d5afa87b07dd3d750c/packages/create-cloudflare/templates/hello-world/js/package.json#L14

Cherry:
Hmm, interesting. Wouldn't it need to pin to the workspace version, since 0.2.7 will also be pinned to the old 1.3.0?

petebacondarwin:
> Hmm, interesting. Wouldn't it need to pin to the workspace version with the updated peer deps, since 0.2.7 will also be pinned to the old vitest 1.3.0?

I don't think we can use the workspace because these are templates that get used outside of the workers-sdk repo. I.e. when running C3 externally. So it has to point to a publicly deployed version.

I think it is a bit of a chicken and egg thing? We need the C3 e2e tests to pass before we release the new version and we need the new version released to run the C3 e2e tests.

Perhaps we just don't update these C3 templates to vitest 1.5.0 for the time being?

Cherry:
> Perhaps we just don't update these C3 templates to vitest 1.5.0 for the time being?

That might be the best solution here, yeah. And then I can make a followup PR once this lands in a release, updating the C3 templates afterwards.

Thanks for digging in here.

RamIdeas:
> I think it is a bit of a chicken and egg thing? We need the C3 e2e tests to pass before we release the new version and we need the new version released to run the C3 e2e tests.

Can an override in the top-level package.json help here? The template can point to the released public version (for anyone cloning the template) and the e2e tests will respect the override and use the workspace version

petebacondarwin:
> > I think it is a bit of a chicken and egg thing? We need the C3 e2e tests to pass before we release the new version and we need the new version released to run the C3 e2e tests.
> 
> Can an [override](https://pnpm.io/package_json#pnpmoverrides) in the top-level package.json help here? The template can point to the released public version (for anyone cloning the template) and the e2e tests will respect the override and use the workspace version

Yeah I thought about that. But I think that is overkill for this situation. These templates are not really part of the key infrastructure that we want to test against local versions of stuff. They are just for external users who will always be using the publicly deployed version of c3. So I don't think it is worth the extra hassle of switching out the versions just for our C3 e2e tests.

PR #5466, opened 2024-04-02
Ai Finetune commands

celso:

Adds Workers AI finetune commands

- Tests
  - [ ] TODO (before merge)
  - [X] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [X] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/tree/workersai-devweek
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/grumpy-bats-change.md
+++ b/.changeset/grumpy-bats-change.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Adds Workers AI finetune commands
--- a/.changeset/grumpy-bats-change.md
+++ b/.changeset/grumpy-bats-change.md
@@ -1,5 +1,5 @@
 ---
-"wrangler": patch
+"wrangler": minor
 ---
 
-Adds Workers AI finetune commands
+feature: add Workers AI finetune commands
--- a/packages/wrangler/src/ai/createFinetune.tsx
+++ b/packages/wrangler/src/ai/createFinetune.tsx
@@ -1,0 +1,103 @@
+import fs, { existsSync, statSync } from "fs";
+import path from "path";
+import { fetchResult } from "../cfetch";
+import { withConfig } from "../config";
+import { logger } from "../logger";
+import { requireAuth } from "../user";
+import { asJson } from "../yargs-types";
+import { getErrorMessage } from "./utils";
+import type {
+	CommonYargsArgv,
+	StrictYargsOptionsToInterface,
+} from "../yargs-types";
+import type { Finetune } from "./types";
+
+export function options(yargs: CommonYargsArgv) {
+	return asJson(yargs);
+}
+
+const requiredAssets = ["adapter_config.json", "adapter_model.safetensors"];
+
+type HandlerOptions = StrictYargsOptionsToInterface<typeof options>;
+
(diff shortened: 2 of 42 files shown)


packages/wrangler/src/user/user.ts
```diff
@@ -349,7 +349,7 @@ const DefaultScopes = {
 	"zone:read": "Grants read level access to account zone.",
 	"ssl_certs:write": "See and manage mTLS certificates for your account",
 	"constellation:write": "Manage Constellation projects/models",
-	"ai:read": "List AI models",
+	"ai:write": "See and change Workers AI catalog and assets",
```
mchenco, line 352: on dash this is ai:edit, does it matter

.changeset/grumpy-bats-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Adds Workers AI finetune commands
```
mrbbot, line 5:
```suggestion
---
"wrangler": minor
---

feature: add Workers AI finetune commands
```

packages/wrangler/src/ai/createFinetune.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { asJson } from "../yargs-types";
+import { getErrorMessage } from "./utils";
+import type {
+	CommonYargsArgv,
+	StrictYargsOptionsToInterface,
+} from "../yargs-types";
+import type { Finetune } from "./types";
+
+export function options(yargs: CommonYargsArgv) {
+	return asJson(yargs);
```
mrbbot, line 16: You'll want to define your positional arguments here with `.positional()`, similar to:

https://github.com/cloudflare/workers-sdk/blob/38f7d5491bf47690401c861ea6b86e520b0cb7cc/packages/wrangler/src/type-generation.ts#L19-L24

That will give you type checking for options, and allow you to drop the `as any`s below.

packages/wrangler/src/ai/createFinetune.tsx
```diff
@@ -0,0 +1,103 @@
...
+}
+
+const requiredAssets = ["adapter_config.json", "adapter_model.safetensors"];
+
+type HandlerOptions = StrictYargsOptionsToInterface<typeof options>;
+
+export const handler = withConfig<HandlerOptions>(
```
mrbbot, line 23: It doesn't look like this handler supports JSON output, even though you've used `asJson` in `options` to add a `--json` flag. Did you mean to add support for this here, or should the `asJson` above be removed?

packages/wrangler/tsconfig.json
```diff
@@ -3,7 +3,8 @@
 	"compilerOptions": {
 		"module": "CommonJS",
 		"types": ["node"],
-		"jsx": "react"
+		"jsx": "react",
+		"lib": ["dom"]
```
mrbbot, line 7: We really don't want to type `wrangler` with `dom` types. What type error were you seeing without this?
```suggestion
```

PR #5477, opened 2024-04-02
chore(wrangler): Updated Queues client to use new 'by-id' endpoints.

pmiguel:

Updates the Queues client to use new 'by-id' endpoints rather than the old 'by name' endpoints, which are in deprecation.

Fixes # MQ-553 internally

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: no change in the way wrangler is used, this is transparent.


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -1068,25 +1053,23 @@ export async function updateQueueConsumers(
...
-			const body: PutConsumerBody = {
+			if (config.name === undefined) {
+				// TODO: how can we reliably get the current script name?
+				throw new UserError(
+					"Script name is required to update queue consumers"
+				);
+			}
+
+			const body: PostTypedConsumerBody = {
+				type: "worker",
```
pmiguel, line 1070: Q: should we still map this to `consumer.type`, and set `type=worker` as the default in the toml parser?

packages/wrangler/src/queues/client.ts
```diff
@@ -46,103 +48,148 @@ export type Consumer = ScriptReference & {
...
-		new URLSearchParams({ page: page.toString() })
-	);
+	const params = new URLSearchParams({ page: page.toString() });
+
+	if (queueName) {
+		// TODO final parameter name
+		params.append("queueName", queueName);
+	}
+
+	return fetchResult(`/accounts/${accountId}/queues`, {}, params);
```
mattdeboard, line 117: Should we factor out the hard-coded URLs into something like 

```ts
function queuesURL(accountId: string, queueId?: string): string {
  let url = "/accounts/${accountId}/queues"
  if (queueId) {
    url = `${url}/${queueId}`
  }
  return url
} 
```

?

pmiguel: Good call.

pmiguel: Added a small URL builder.

CarmenPopoviciu:
LGTM. Have these changes been validated by someone on the queues team as well?

packages/wrangler/src/versions/upload.ts
```diff
@@ -567,18 +566,7 @@ async function ensureQueuesExist(config: Config) {
...
-			const queueErr = err as FetchError;
-			if (queueErr.code === 11000) {
-				// queue_not_found
-				throw new UserError(
-					`Queue "${queue}" does not exist. To create it, run: wrangler queues create ${queue}`
-				);
-			}
-			throw err;
-		}
+		await getQueue(config, queue);
```
w-kuhn, line 569: could this loop be replaced with a single call to get multiple queues by name?

packages/wrangler/src/queues/client.ts
```diff
@@ -46,208 +48,248 @@ export type Consumer = ScriptReference & {
...
+			);
+		}
+		this.url += "/consumers";
+		if (consumerId) {
+			this.url += `/${consumerId}`;
+		}
+		return this;
+	}
+
+	build(): string {
```
w-kuhn, line 100: if the invocations of the `withConsumer` and `withQueueId` methods are only ever chained with `build()` - can we eliminate the build method and just return `this.url` from the methods that update it?

packages/wrangler/src/queues/client.ts
```diff
@@ -46,208 +48,248 @@ export type Consumer = ScriptReference & {
...
+	withQueueId(queueId: string): QueuesUrlBuilder {
+		this.url += `/${queueId}`;
+		this.hasQueueId = true;
+		return this;
+	}
+
+	withConsumer(consumerId?: string): QueuesUrlBuilder {
+		if (!this.hasQueueId) {
+			throw new UserError(
+				"Can't access consumers without a queueId specified."
```
w-kuhn, line 90: it is surprising to see a UserError thrown from the depths of a utility function - can this be removed? I think the UserError thrown higher up covers this case, right? `throw new UserError(`No http_pull consumer exists for queue ${queueName}`);`

packages/wrangler/src/queues/client.ts
pmiguel: `withQueueId` can be optionally chained with `withConsumer`, that's why I added a "return" method that actually returns the fully formed string.

packages/wrangler/src/queues/client.ts
pmiguel: It might not hit that `UserError` on some code paths. But tbh this is an internal thing. I'm ok with it just forming a malformed URL, would get a 404/400 from the API.

packages/wrangler/src/versions/upload.ts
pmiguel: Yap, just requires a few utility functions. Will do that.

packages/wrangler/src/queues/cli/commands/consumer/worker/remove.ts
```diff
@@ -26,6 +26,6 @@ export async function handler(
 	const config = readConfig(args.config, args);
 
 	logger.log(`Removing consumer from queue ${args.queueName}.`);
-	await deleteConsumer(config, args.queueName, args.scriptName, args.env);
+	await deleteWorkerConsumer(config, args.queueName, args.scriptName);
```
w-kuhn, line 29: doesn't this still need the `env` to find the correct consumer id?

pmiguel: Did some parkour on this, should correctly find the ID of service workers now.

w-kuhn (approved):

CarmenPopoviciu (approved):

PR #5482, opened 2024-04-02
Use new Discord server url

DaniFoldi:
## What this PR solves / how to test

Hi :wave:,

I was originally reproducing and fixing a completely unrelated bug in c3, but noticed the old url in the summary after creating a project. Since the new URL is preferred, I thought I'd PR it, and a quick search found another 20 or so matches.

The only remaining instances of the old .gg/cloudflaredev one are in changelog entries, which I didn't want to modify.

The changeset for this is a patch to every touched package, not sure if this is worth releasing kv-asset-handler for. Also since Hyperdrive is now GA (🎉), the banner might go away so if this would cause merge conflicts, happy to undo any conflicting ones (also Constellation).

Fixes #[insert GH or internal issue number(s)].

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [x] Included, but something broke and didn't run locally so 🤞 
  - [ ] Not necessary because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): <https://github.com/cloudflare/cloudflare-docs/pull/>...
  - [x] Not necessary because: most places in the docs repo already use the new link

PR #5483, opened 2024-04-02
[C3] Skip empty summary rows

DaniFoldi:

Hi :wave:,

Before getting side-tracked with #5482, I found that when using `.` as the project location, there was nowhere to `cd` to, so the first line in the next steps output summary was skipped. Instead of being properly skipped, `undefined undefined` was printed (see screenshot). This change should skip over these lines where a step is excluded.

image

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Didn't see any tests for the summary
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)

### DIFF
--- a/.changeset/quick-cheetahs-attack.md
+++ b/.changeset/quick-cheetahs-attack.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+Skip empty next steps in the output summary. Previously, they would log undefined undefined.
--- a/.changeset/quick-cheetahs-attack.md
+++ b/.changeset/quick-cheetahs-attack.md
@@ -2,4 +2,4 @@
 "create-cloudflare": patch
 ---
 
-Skip empty next steps in the output summary. Previously, they would log undefined undefined.
+fix: make sure that using the current directory doesn't generate an `undefined undefined` log in the output summary.
--- a/packages/create-cloudflare/src/summary.ts
+++ b/packages/create-cloudflare/src/summary.ts
@@ -59,7 +59,7 @@ export const printSummary = async (ctx: C3Context) => {
 	}
 
 	newline();
-	nextSteps.forEach((entry) => {
+	nextSteps.filter(entry => entry.length > 0).forEach((entry) => {
 		log(`${dim(entry[0])} ${blue(entry[1])}`);
 	});
 	newline();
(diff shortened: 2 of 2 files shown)


dario-piotrowicz:
Hi @DaniFoldi 😄👋 

Thanks a lot for the fix 🙂 

but I feel like the change is not the right solution, I think that instead of filtering the thing out we should just not add it in the summary in the first place

I think that this code is slightly off:
https://github.com/cloudflare/workers-sdk/blob/940ad89713fa086f23d394570c328716bfb1bd59/packages/create-cloudflare/src/summary.ts#L15-L17

And changing it in the following way should fix the issue:
```diff
 
        const dirRelativePath = relative(ctx.originalCWD, ctx.project.path);
        const nextSteps = [
-               dirRelativePath
-                       ? ["Navigate to the new directory", `cd ${dirRelativePath}`]
-                       : [],
+               ...(dirRelativePath
+                       ? [["Navigate to the new directory", `cd ${dirRelativePath}`]]
+                       : []),
                [
                        "Run the development server",
                        quoteShellArgs([npm, "run", ctx.template.devScript ?? "start"]),
```

what do you think? 🙂

DaniFoldi:
Hey @dario-piotrowicz,

Thanks, good idea! I've pushed a new commit changing it to your suggestion, I agree it's cleaner. Looked at the option of adding an explicit type annotation to prevent future issues, but ts couldn't validate all of them as `[string, string][]`, so it was going to require more type assertions than I'd be comfortable with for such a small change.

.changeset/quick-cheetahs-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+Skip empty next steps in the output summary. Previously, they would log undefined undefined.
```
dario-piotrowicz, line 5: given the PR's change I would update the changeset 🙏 

something along these lines would be great 🙂
```suggestion
fix: make sure that using the current directory doesn't generate an `undefined undefined` log in the output summary.
```

the `fix:` prefix is as per our [changesets convention](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changeset-message-format)

dario-piotrowicz (approved):
Looks good to me, thanks a bunch @DaniFoldi ❤️ 🙂

dario-piotrowicz:
> Thanks, good idea! I've pushed a new commit changing it to your suggestion, I agree it's cleaner. Looked at the option of adding an explicit type annotation to prevent future issues, but ts couldn't validate all of them as [string, string][], so it was going to require more type assertions than I'd be comfortable with for such a small change.

yeah I agree, if it's going to add unnecessary complexity it doesn't really feel worth it 🙂

jculvey (approved):

PR #5491, opened 2024-04-03
[miniflare] fix: make sure the magic proxy can handle multiple parallel R2 stream reads

### LINKED ISSUE #5360: `wrangler.getBindingsProxy` will freeze the server when reading two bucket streams in parallel
(Context: This issue was originally reported by @atinux for [Nuxt Hub](https://hub.nuxt.com/), we have later narrowed down this issue to [nitro-cf-dev](https://github.com/pi0/nitro-cloudflare-dev/tree/main) and then finally to this minimal reproduction.)

## Description

Using [`wrangler.getBindingsProxy`](https://github.com/cloudflare/workers-sdk/pull/5002) to read R2 bucket data, the readable stream will be stuck on big sizes (1MB+) if used with concurrency of 2+ promises. The second stream will never end and the Node.js process is stuck. This issue does not happen in production.

## Minimal reproduction

I have made a repo with two simple (plain script and plain node server) examples that reproduce this. https://github.com/pi0/cf-proxy-repro

(tested on `wrangler@3.37.0`)

```toml
r2_buckets = [
  { binding = "BLOB", bucket_name = "default" },
]
```

```js
import { getPlatformProxy } from "wrangler" // tested on 3.37.0

const proxy = await getPlatformProxy({})
const { BLOB: bucket } = proxy.env

console.log('Fetching cat image...')
const imgReadableStream = await fetch('https://images.unsplash.com/photo-1579168765467-3b235f938439?ixlib=rb-4.0.3&q=85&fm=jpg&crop=entropy&cs=srgb&dl=yoo-hoo-E3LcqpQxtTU-unsplash.jpg').then(r => r.body)

console.log('Converting image to buffer...')
const buff = await readableStreamToBuff(imgReadableStream)
console.log('Buffer length (response):', buff.length)

console.log('Putting cats to the bucket...')
await bucket.put('cat1.jpg', buff.buffer)
await bucket.put('cat2.jpg', buff.buffer)

console.log('Reading from bucket...')

// This will make process hang
const [cat1, cat2] = await Promise.all([
  bucket.get('cat1.jpg').then(obj => readableStreamToBuff(obj.body)),
  bucket.get('cat2.jpg').then(obj => readableStreamToBuff(obj.body)),
])

// This works
// const cat1 = await bucket.get('cat1.jpg').then(obj => readableStreamToBuff(obj.body))
// const cat2 = await bucket.get('cat2.jpg').then(obj => readableStreamToBuff(obj.body))

console.log('Buffer length (bucket):', [cat1.length, cat2.length])

await proxy.dispose()

async function readableStreamToBuff(stream) {
  console.log('reading stream to buffer... ')
  const chunks = [];
  await stream.pipeTo(
    new WritableStream({
      write(chunk) {
        process.stdout.write(` +${chunk.length}`)
        chunks.push(chunk);
      },
      close() {
        console.log('stream close')
      }
    }),
  )
  console.log('stream read!')
  return Buffer.concat(chunks);
}

```

Used with parallel `Promise.all`: (process hangs)

image

Used without parallel promises: (works)

image

## Possible workaround

Using a mutex to run only one at a time. Used for [nitro-cloudflare-dev](https://github.com/pi0/nitro-cloudflare-dev/blob/main/src/runtime/r2-patch.ts)

dario-piotrowicz:

Currently trying to read multiple R2 streams in parallel (via `Promise.all` for example)
generates a deadlock which prevents any of the target streams to be read, this is caused
by the magic proxy underlying implementation only allowing a single HTTP connection to the
workerd process at a time. 

This PR fixes such issue by instead allowing any number of parallel HTTP connections at the same time.

Fixes #5360

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: this is a bug fix

### DIFF
--- a/.changeset/twelve-geckos-melt.md
+++ b/.changeset/twelve-geckos-melt.md
@@ -4,8 +4,4 @@
 
 fix: make sure the magic proxy can handle multiple parallel r2 stream reads
 
-currently trying to read multiple R2 streams in parallel (via `Promise.all` for example)
-generates a deadlock which prevents any of the target streams to be read, this is caused
-by the magic proxy underlying implementation only allowing a single HTTP connection to the
-workerd process at a time. Fix such issue by instead allowing any number of parallel HTTP
-connections at the same time.
+Currently trying to read multiple R2 streams in parallel (via `Promise.all` for example) leads to deadlock which prevents any of the target streams from being read. This is caused by the underlying implementation only allowing a single HTTP connection to the Workers runtime at a time. This change fixes the issue by allowing multiple parallel HTTP connections.
--- a/.changeset/twelve-geckos-melt.md
+++ b/.changeset/twelve-geckos-melt.md
@@ -1,0 +1,11 @@
+---
+"miniflare": patch
+---
+
+fix: make sure the magic proxy can handle multiple parallel r2 stream reads
+
+currently trying to read multiple R2 streams in parallel (via `Promise.all` for example)
+generates a deadlock which prevents any of the target streams to be read, this is caused
+by the magic proxy underlying implementation only allowing a single HTTP connection to the
+workerd process at a time. Fix such issue by instead allowing any number of parallel HTTP
+connections at the same time.
--- a/packages/miniflare/src/plugins/core/proxy/fetch-sync.ts
+++ b/packages/miniflare/src/plugins/core/proxy/fetch-sync.ts
@@ -39,23 +39,23 @@ const { notifyHandle, port, filename } = workerData;
 // with "Error: Cannot find module 'undici'". Instead we need to create a
 // 'require' using the '__filename' of the host... :(
 const actualRequire = createRequire(filename);
-const { Client, fetch } = actualRequire("undici");
+const { Pool, fetch } = actualRequire("undici");
 
-let clientUrl;
-let client;
+let dispatcherUrl;
+let dispatcher;
 
 port.addEventListener("message", async (event) => {
   const { id, method, url, headers, body } = event.data;
-  if (clientUrl !== url) {
-    clientUrl = url;
-    client = new Client(url, {
+  if (dispatcherUrl !== url) {
+    dispatcherUrl = url;
+    dispatcher = new Pool(url, {
       connect: { rejectUnauthorized: false },
     });
   }
   headers["${CoreHeaders.OP_SYNC}"] = "true";
   try {
     // body cannot be a ReadableStream, so no need to specify duplex
-    const response = await fetch(url, { method, headers, body, dispatcher: client });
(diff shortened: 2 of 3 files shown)


RamIdeas (approved):
LGTM

I would wait for @mrbbot to review before merging – there may have been a reason to limit to 1 HTTP connection or not

dario-piotrowicz:
> LGTM
> 
> I would wait for @mrbbot to review before merging – there may have been a reason to limit to 1 HTTP connection or not

@mrbbot came up with the solution, so hopefully he should be cool with it 😁

but yeah I'm happy to wait for his approval 🙂

.changeset/twelve-geckos-melt.md
```diff
@@ -0,0 +1,11 @@
...
+"miniflare": patch
+---
+
+fix: make sure the magic proxy can handle multiple parallel r2 stream reads
+
+currently trying to read multiple R2 streams in parallel (via `Promise.all` for example)
+generates a deadlock which prevents any of the target streams to be read, this is caused
+by the magic proxy underlying implementation only allowing a single HTTP connection to the
+workerd process at a time. Fix such issue by instead allowing any number of parallel HTTP
+connections at the same time.
```
mrbbot, line 11:
```suggestion
Currently trying to read multiple R2 streams in parallel (via `Promise.all` for example) leads to deadlock which prevents any of the target streams from being read. This is caused by the underlying implementation only allowing a single HTTP connection to the Workers runtime at a time. This change fixes the issue by allowing multiple parallel HTTP connections.
```

mrbbot (approved):
Looks good! ✅ I've made a couple suggestions to the changeset.

Dario and I paired on fixing this. To elaborate on this issue here, attempting to access `R2ObjectBody#body` causes a synchronous network request via `fetch-sync.ts` to get the `body` `ReadableStream`. Whilst the stream itself is returned synchronously, the chunks are not. When another `get()` request is sent, and `R2ObjectBody#body` is accessed again, this `Atomics.wait()` call is hit:

https://github.com/cloudflare/workers-sdk/blob/bd00dc827769145df8d42a677393173bd751e643/packages/miniflare/src/plugins/core/proxy/fetch-sync.ts#L130

This blocks the main thread that the stream from the 1st request has been transferred to, preventing it from being read. Because we were using a single-connection `Client`, `undici` had to wait for the response to be read before sending the next request. This meant the `Atomics.wait()` call never unblocked, as we were blocking the response being read, so had cyclic dependencies and deadlock.

PR #5506, opened 2024-04-04
Add prompt to `wrangler pages download config`

penalosa:

Adds an interactive prompt to `wrangler pages download config` if downloading the config would overwrite an existing `wrangler.toml` file.

Fixes [DEVX-1243](https://jira.cfdata.org/browse/DEVX-1243)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Not yet documented

### DIFF
--- a/.changeset/wise-grapes-prove.md
+++ b/.changeset/wise-grapes-prove.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feat: Add interactive prompt to `wrangler pages download config` if an exsting `wrangler.toml` file exists
--- a/.changeset/wise-grapes-prove.md
+++ b/.changeset/wise-grapes-prove.md
@@ -2,4 +2,4 @@
 "wrangler": minor
 ---
 
-feat: Add interactive prompt to `wrangler pages download config` if an exsting `wrangler.toml` file exists
+feat: Add interactive prompt to `wrangler pages download config` if an existing `wrangler.toml` file exists
--- a/packages/wrangler/src/__tests__/pages/pages-download-config.test.ts
+++ b/packages/wrangler/src/__tests__/pages/pages-download-config.test.ts
@@ -602,7 +602,11 @@ describe("pages download config", () => {
 		it("should not overwrite existing file w/o --force (non-interactive)", async () => {
 			setIsTTY(false);
 			await writeWranglerToml({ name: "some-project" });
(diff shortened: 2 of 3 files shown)


.changeset/wise-grapes-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feat: Add interactive prompt to `wrangler pages download config` if an exsting `wrangler.toml` file exists
```
petebacondarwin, line 5:
```suggestion
feat: Add interactive prompt to `wrangler pages download config` if an existing `wrangler.toml` file exists
```

packages/wrangler/src/__tests__/pages/pages-download-config.test.ts
```diff
@@ -447,4 +456,329 @@ describe("pages download config", () => {
...
+					  [[env.production.analytics_engine_datasets]]
+					  binding = \\"AE\\"
+					  dataset = \\"data\\"
+
+					  [env.production.ai]
+					  binding = \\"AI\\"
+					"
+			`);
+		});
+		it("should not overwrite existing file w/o --force (non-interactive)", async () => {
```
petebacondarwin, line 602: Should we check what messages are logged in this case?

petebacondarwin (changes requested):
Is `false` the correct default option here for non-interactive?
I feel like it should actually fail... because otherwise if someone is expecting their CI to download and keep the wrangler.toml updated before deploying, and haven't specified `--force` it will just deploy the old wrangler.toml instead.

petebacondarwin (approved):

packages/wrangler/src/pages/download-config.ts
```diff
@@ -225,7 +225,7 @@ export const Handler = async ({ projectName, force }: DownloadConfigArgs) => {
 			{ fallbackValue: false }
 		);
 		if (!overwrite) {
-			return;
+			throw new FatalError("Not overwriting existing `wrangler.toml` file");
```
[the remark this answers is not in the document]
admah: I think it's ok for now. We can revisit if needed.

PR #5530, opened 2024-04-05
docs: add contributing guides for Miniflare/Vitest integration

mrbbot:

This PR adds some high-level contributing guides for Miniflare and the Workers Vitest integration. This includes a new high-level architecture diagram for Miniflare, that describes the services included in Miniflare's `workerd` config.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: internal docs
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: internal docs
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: internal docs


CarmenPopoviciu (approved):

packages/miniflare/src/workers/core/entry.worker.ts
petebacondarwin, line 1: The changes in this file look a bit out of place...?

CarmenPopoviciu: I had a look through the codebase and it seems like this is the only place these are used. I'm also seeing that one of the diagrams introduced by this PR uses the new type as well, so I'm thinking this change was probably intentional. I'll keep as is and go ahead and merge

PR #5531, opened 2024-04-05
Write build-env to file

### LINKED ISSUE #5525: 🐛 BUG: Unable to deploy Pages when using a `worker.toml` with Hyperdrive binding

Pages, Other

Latest everything

18.18.2

macOS 14.4.1

### Observed behavior
I am using the Cloudflare Pages Github application to automatically deploy my app. Today, I added the new `wrangler.toml` file since support was just added to Pages. I have a Hyperdrive binding in it. When I push my code to automatically deploy, I get this opaque error message:
```
20:47:19.442	wrangler.toml config file detected
20:47:20.465	Failed: an internal error occurred. If this continues, contact support: https://cfl.re/3WgEyrH
```

### Expected behavior
Because I have seen this issue in development, I assume that the problem is due to this error:
```
Error: When developing locally, you should use a local Postgres connection string to emulate Hyperdrive functionality. Please setup Postgres locally and set the value of the 'WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_HYPERDRIVE' variable or "HYPERDRIVE"'s "localConnectionString" to the Postgres connection string.
```

When the Cloudflare Pages application builds my Pages site (built with SvelteKit) that has a `wrangler.toml` with a Hyperdrive binding, it needs a Hyperdrive local connection string. How can I provide a local connection string to the Cloudflare Pages application while building? Setting one in the Pages console environment variables and secrets doesn't work.

### Steps to reproduce
1. Create a new SvelteKit project with Cloudflare.
2. Add automatic deployments with the Cloudflare Pages Github Application.
3. Add a `wrangler.toml` file with a Hyperdrive binding.
4. Push your code.
5. Observe that the deployment will fail with a vague error message.

Note that deploying directly using `npx wrangler pages deploy` works because it builds my site locally, which uses my local environment variables that have local connection strings for Hyperdrive. However, I would prefer not to use this method and have my deployments managed by Github.

_No response_

_No response_

penalosa:

This ensures that `wrangler pages functions build-env` outputs to a file, rather than stdout.

Fixes #5525 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: internal only

### DIFF
--- a/packages/wrangler/src/pages/build-env.ts
+++ b/packages/wrangler/src/pages/build-env.ts
@@ -1,8 +1,7 @@
-import { existsSync } from "node:fs";
+import { existsSync, writeFileSync } from "node:fs";
 import path from "node:path";
 import { readConfig } from "../config";
 import { FatalError } from "../errors";
-import { logger } from "../logger";
 import { EXIT_CODE_NO_CONFIG_FOUND } from "./errors";
 import type {
 	CommonYargsArgv,
@@ -12,16 +11,27 @@ import type {
 export type PagesBuildEnvArgs = StrictYargsOptionsToInterface<typeof Options>;
 
 export function Options(yargs: CommonYargsArgv) {
-	return yargs.positional("projectDir", {
-		type: "string",
-		description: "The location of the Pages project",
-	});
+	return yargs
+		.positional("projectDir", {
+			type: "string",
+			description: "The location of the Pages project",
+		})
+		.options({
+			outfile: {
+				type: "string",
+				description: "The location to write the build environment file",
+			},
+		});
 }
 
 export const Handler = async (args: PagesBuildEnvArgs) => {
 	if (!args.projectDir) {
 		throw new FatalError("No Pages project location specified");
 	}
+	if (!args.outfile) {
+		throw new FatalError("No outfile specified");
+	}
+
 	const configPath = path.resolve(args.projectDir, "wrangler.toml");
 	// Fail early if the config file doesn't exist
 	if (!existsSync(configPath)) {
@@ -46,7 +56,8 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
 		Object.entries(config.vars).filter(([_, v]) => typeof v === "string")
 	);
 
-	logger.log(
+	writeFileSync(
+		args.outfile,
 		JSON.stringify({
 			vars: textVars,
 			pages_build_output_dir: path.relative(
--- a/packages/wrangler/src/pages/build-env.ts
+++ b/packages/wrangler/src/pages/build-env.ts
@@ -2,6 +2,7 @@ import { existsSync, writeFileSync } from "node:fs";
 import path from "node:path";
 import { readConfig } from "../config";
 import { FatalError } from "../errors";
+import { logger } from "../logger";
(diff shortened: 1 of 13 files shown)


petebacondarwin (approved):

petebacondarwin (approved):

packages/wrangler/src/pages/build-env.ts
```diff
@@ -40,9 +41,10 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
 			EXIT_CODE_NO_CONFIG_FOUND
 		);
 	}
+	logger.log("Reading build configuration from your wrangler.toml file...");
```
CarmenPopoviciu, line 44: will this show up in the CI logs?

packages/wrangler/src/pages/build-env.ts
```diff
@@ -56,14 +58,26 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
...
+	const buildConfiguration = {
+		vars: textVars,
+		pages_build_output_dir: path.relative(
+			args.projectDir,
+			config.pages_build_output_dir
+		),
+	};
+
+	writeFileSync(args.outfile, JSON.stringify(buildConfiguration));
+	logger.debug(`Build configuration written to ${args.outfile}`);
```
CarmenPopoviciu, line 70: what abt these debug logs?

penalosa: These won't show up in CI

packages/wrangler/src/pages/build-env.ts
penalosa: These will, if we print the stdout of this commend. They won't show up by default

CarmenPopoviciu: yes we do

packages/wrangler/src/pages/build-env.ts
CarmenPopoviciu: kk

packages/wrangler/src/pages/build-env.ts
CarmenPopoviciu: I think `build configuration` might be confusing here. perhaps `environment configuration` might be better

packages/wrangler/src/pages/build-env.ts
```diff
@@ -46,13 +58,26 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
...
 
+	const buildConfiguration = {
+		vars: textVars,
+		pages_build_output_dir: path.relative(
+			args.projectDir,
+			config.pages_build_output_dir
+		),
+	};
+
+	writeFileSync(args.outfile, JSON.stringify(buildConfiguration));
```
CarmenPopoviciu, line 69: same comment here wrt `build configuration` wording

packages/wrangler/src/pages/build-env.ts
penalosa: It includes the build output directory as well, which is why I called it build configuration

PR #5570, opened 2024-04-09
MQ-605: Extend miniflare's queues simulator to support delayed delivery

sesteves:
This patch updates the miniflare's queue broker to support delayed delivery of messages, both when sending the message from a producer and when retrying the message from a consumer.
The implementation is based on miniflare's timers, and currently there is no support for specifying the delivery delay in queue settings for a producer.

Fixes #MQ-605

- Tests
  - [ ] TODO (before merge)
  - [X] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [X] TODO (before merge)
  - [ ] Not necessary because:


packages/miniflare/src/workers/queues/schemas.ts
```diff
@@ -1,12 +1,15 @@
...
+export const QueueMessageDelaySchema = z.number().int().min(0).max(43200).optional();
+
 export const QueueConsumerOptionsSchema = /* @__PURE__ */ z.object({
 	// https://developers.cloudflare.com/queues/platform/configuration/#consumer
 	// https://developers.cloudflare.com/queues/platform/limits/
 	maxBatchSize: z.number().min(0).max(100).optional(),
 	maxBatchTimeout: z.number().min(0).max(30).optional(), // seconds
 	maxRetires: z.number().min(0).max(100).optional(),
 	deadLetterQueue: z.ostring(),
+	retryDelay: QueueMessageDelaySchema,
```
penalosa, line 12: What does this represent? It looks like users can configure the delay when calling `.retry()`—why is it also specified in the consumer settings?

packages/miniflare/src/workers/queues/broker.worker.ts
```diff
@@ -320,28 +340,36 @@ export class QueueBrokerObject extends MiniflareDurableObject<QueueBrokerObjectE
...
 		const contentType = validateContentType(req.headers);
+		const delay = validateMessageDelay(req.headers);
 		const body = Buffer.from(await req.arrayBuffer());
 
 		// If we don't have a consumer, drop the message
 		const consumer = this.#maybeConsumer;
 		if (consumer === undefined) return new Response();
 
-		this.#enqueue([{ contentType, body }]);
+		this.#enqueue([{ contentType, delaySecs: undefined, body }], delay);
```
penalosa, line 372: Why does this set the `globalDelay` instead of the message delay?

packages/miniflare/src/workers/queues/broker.worker.ts
```diff
@@ -320,28 +340,36 @@ export class QueueBrokerObject extends MiniflareDurableObject<QueueBrokerObjectE
...
+
+			const fn = () => {
+				this.#messages.push(msg);
+				this.#ensurePendingFlush();
+			}
+
+			const delay = message.delaySecs ?? globalDelay;
+			this.timers.setTimeout(fn, delay * 1000);
 		}
-		this.#ensurePendingFlush();
```
penalosa, line 331: This is a change in behaviour from ensuring a flush at the end of all messages to ensureing a flush at the end of each message. Will this cause problems?

packages/miniflare/src/workers/queues/schemas.ts
sesteves: Please refer to https://developers.cloudflare.com/queues/reference/batching-retries/#delay-on-retry
"Delays can also be configured in wrangler.toml...."

packages/miniflare/src/workers/queues/broker.worker.ts
sesteves: This is the same in practice; i.e. only one delay can be specified when sending a single message.

penalosa: Coudl you add a comment here to that effect?

penalosa (approved):

penalosa:
Approved pending e2e tests

packages/miniflare/test/plugins/queues/index.spec.ts
```diff
@@ -834,3 +833,142 @@ test("validates message size", async (t) => {
...
+	async function sendBatch(delay: number, ...messages: unknown[]) {
+		await mf.dispatchFetch("http://localhost/batch", {
+			method: "POST",
+			headers: { "X-Msg-Delay-Secs": delay.toString() },
+			body: JSON.stringify(messages),
+		});
+	}
+
+	// Send batch of messages.
+	sendBatch(1, { body: 10, delaySeconds: 0 }, { body: 11 }, { body: 12, delaySeconds: 2 })
```
petebacondarwin, line 942: For future reference... the first parameter here is saying that messages in the batch that do not provide a delaySeconds property will default to 1 second delay.

petebacondarwin (approved):
This looks great. Thanks for updating Miniflare!
I left a couple of nits but not blocking landing if you are keen merge today.

packages/miniflare/src/workers/queues/schemas.ts
```diff
@@ -1,12 +1,29 @@
...
+	/* @__PURE__ */ z.record(QueueProducerSchema);
+
 export const QueueConsumerOptionsSchema = /* @__PURE__ */ z.object({
 	// https://developers.cloudflare.com/queues/platform/configuration/#consumer
 	// https://developers.cloudflare.com/queues/platform/limits/
 	maxBatchSize: z.number().min(0).max(100).optional(),
 	maxBatchTimeout: z.number().min(0).max(30).optional(), // seconds
 	maxRetires: z.number().min(0).max(100).optional(),
 	deadLetterQueue: z.ostring(),
+	retryDelay: QueueMessageDelaySchema,
```
petebacondarwin, line 26: Can you update the Miniflare README with this new option?

https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/README.md#interface-queueconsumeroptions

petebacondarwin (changes requested):
Actually the change to the configuration API concerns me. We even document it publicly here: https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/README.md#queues

packages/miniflare/test/plugins/queues/index.spec.ts
```diff
@@ -311,8 +311,7 @@ test("retries messages", async (t) => {
 	const log = new TestLog(t);
 	const mf = new Miniflare({
 		log,
-
-		queueProducers: { QUEUE: "queue" },
+		queueProducers: { QUEUE: { queueName: "queue" } },
```
petebacondarwin, line 314: This change is no longer needed, right?

petebacondarwin (approved):

PR #5572, opened 2024-04-09
fix(wrangler): fix `build-env` to exit with code rather than throw fatal error

CarmenPopoviciu:

Currently `pages functions build-env` throws a fatal error if a config file does not exit, or if it is
invalid. This causes issues for the CI system. We
should instead exit with a specific code, if any of those situations arises.

You can test by running `npx wrangler pages function build-env [directory] --outfile=<file-name.json>`

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not public facing api

## How this PR was tested

- with valid Pages config file
Screenshot 2024-04-11 at 12 06 05

- with non-Pages config file + `WRANGLER_LOG="debug"`
Screenshot 2024-04-11 at 12 07 08

- with invalid Pages config file + `WRANGLER_LOG="debug"`
Screenshot 2024-04-11 at 12 07 08

- with invalid non-Pages config file + `WRANGLER_LOG="debug"`
Screenshot 2024-04-11 at 12 07 08

- with no config file + `WRANGLER_LOG="debug"`
Screenshot 2024-04-11 at 12 06 42

PREVIEW OF CI ERR MESSAGING

Screenshot 2024-04-11 at 13 48 14

Screenshot 2024-04-11 at 13 53 30

### DIFF
--- a/packages/wrangler/src/pages/build-env.ts
+++ b/packages/wrangler/src/pages/build-env.ts
@@ -3,7 +3,11 @@ import path from "node:path";
 import { readConfig } from "../config";
 import { FatalError } from "../errors";
 import { logger } from "../logger";
-import { EXIT_CODE_NO_CONFIG_FOUND } from "./errors";
+import {
+	EXIT_CODE_INVALID_PAGES_CONFIG,
+	EXIT_CODE_NO_CONFIG_FOUND,
+} from "./errors";
+import type { Config } from "../config";
 import type {
 	CommonYargsArgv,
 	StrictYargsOptionsToInterface,
@@ -34,24 +38,30 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
 	}
 
 	const configPath = path.resolve(args.projectDir, "wrangler.toml");
-	// Fail early if the config file doesn't exist
 	if (!existsSync(configPath)) {
-		throw new FatalError(
-			"No Pages config file found",
-			EXIT_CODE_NO_CONFIG_FOUND
-		);
+		logger.log("No Pages config file found. Skipping...");
+		process.exit(EXIT_CODE_NO_CONFIG_FOUND);
 	}
+
 	logger.log("Reading build configuration from your wrangler.toml file...");
 
-	const config = readConfig(
-		configPath,
-		{
-			...args,
-			// eslint-disable-next-line turbo/no-undeclared-env-vars
-			env: process.env.PAGES_ENVIRONMENT,
-		},
-		true
-	);
+	let config: Omit<Config, "pages_build_output_dir"> & {
+		pages_build_output_dir: string;
+	};
+	try {
+		config = readConfig(
+			configPath,
+			{
+				...args,
+				// eslint-disable-next-line turbo/no-undeclared-env-vars
+				env: process.env.PAGES_ENVIRONMENT,
+			},
+			true
+		);
+	} catch (err) {
+		logger.log("Invalid Pages config file found. Skipping...");
+		process.exit(EXIT_CODE_INVALID_PAGES_CONFIG);
+	}
 
 	// Ensure JSON variables are not included
 	const textVars = Object.fromEntries(
--- a/packages/wrangler/src/pages/build-env.ts
+++ b/packages/wrangler/src/pages/build-env.ts
@@ -39,8 +39,9 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
 
 	const configPath = path.resolve(args.projectDir, "wrangler.toml");
 	if (!existsSync(configPath)) {
-		logger.log("No Pages config file found. Skipping...");
-		process.exit(EXIT_CODE_NO_CONFIG_FOUND);
+		logger.log("No Pages configuration file found. Exiting.");
+		process.exitCode = EXIT_CODE_NO_CONFIG_FOUND;
+		return;
 	}
 
 	logger.log("Reading build configuration from your wrangler.toml file...");
@@ -59,8 +60,9 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
 			true
 		);
 	} catch (err) {
-		logger.log("Invalid Pages config file found. Skipping...");
-		process.exit(EXIT_CODE_INVALID_PAGES_CONFIG);
+		logger.log("Invalid Pages configuration file found. Exiting.");
+		process.exitCode = EXIT_CODE_INVALID_PAGES_CONFIG;
+		return;
 	}
 
 	// Ensure JSON variables are not included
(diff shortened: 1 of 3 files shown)


penalosa (approved):

packages/wrangler/src/pages/build-env.ts
```diff
@@ -34,24 +38,30 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
...
 
 	const configPath = path.resolve(args.projectDir, "wrangler.toml");
-	// Fail early if the config file doesn't exist
 	if (!existsSync(configPath)) {
-		throw new FatalError(
-			"No Pages config file found",
-			EXIT_CODE_NO_CONFIG_FOUND
-		);
+		logger.log("No Pages config file found. Skipping...");
+		process.exit(EXIT_CODE_NO_CONFIG_FOUND);
```
petebacondarwin, line 43: Let's not actually call `process.exit()` deep in the codebase. We should only do this at the top level. 

There are a couple of other alternatives: 

1. Throw a well defined typed error that can be caught at the very top level. Perhaps even a `FatalError` that has a "no log" flag. But preferably a error something like `ExitWithCodeError`.
2. Set the exit code using `process. exitCode = ...` and return here, which will then allow the main to exit with that code.

petebacondarwin (changes requested):
Please can we remove the calls to `process.exit()`

petebacondarwin (approved):
I wonder if, separately to this we should mention in the logs (especially for when there is a warning) that wrangler.toml usage is beta?

packages/wrangler/src/pages/build-env.ts
```diff
@@ -34,24 +38,32 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
...
+	let config: Omit<Config, "pages_build_output_dir"> & {
+		pages_build_output_dir: string;
+	};
+	try {
+		config = readConfig(
+			configPath,
+			{
+				...args,
+				// eslint-disable-next-line turbo/no-undeclared-env-vars
+				env: process.env.PAGES_ENVIRONMENT,
```
RamIdeas, line 58: Is this a pre-existing thing inside Pages CI or can we formalise into a WRANGLER_ env var with `getEnvironmentVariableFactory`?

packages/wrangler/src/pages/build-env.ts
```diff
@@ -34,24 +38,32 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
...
+			},
+			true
+		);
+	} catch (err) {
+		logger.log("Invalid Pages configuration file found. Exiting.");
```
RamIdeas, line 63: This should be a warning using logger.warn which will write to stderr instead of stdout

penalosa: This shouldn't be a warning, which will show up as a scary messaging, when this is an expected case. This should be a debug log

packages/wrangler/src/pages/build-env.ts
penalosa: This is a pre-existing thing in Pages CI

packages/wrangler/src/pages/build-env.ts
```diff
@@ -34,24 +38,32 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
...
-		{
-			...args,
-			// eslint-disable-next-line turbo/no-undeclared-env-vars
-			env: process.env.PAGES_ENVIRONMENT,
-		},
-		true
-	);
+	let config: Omit<Config, "pages_build_output_dir"> & {
+		pages_build_output_dir: string;
+	};
```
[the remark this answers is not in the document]
petebacondarwin: It is to make the property required I believe. In `Config` the property is optional.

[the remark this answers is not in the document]
CarmenPopoviciu: that is correct yes. `pages_build_output_dir` is an optional `Config` property, but in this particular case it has to be mandatory. It is also the return type of `readConfig`

packages/wrangler/src/pages/build-env.ts
petebacondarwin: I think this underlines the need for the text of these messages to be less scary themselves.

CarmenPopoviciu: I went ahead and done the following wrt messaging:
1. introduced a one liner `BETA` log (`Checking for configuration in a wrangler.toml (BETA)`) as per our conversation. Keeping in wrangler so we can remove more easily when time comes
2. used `logger.debug` instead of `logger.log` in the wrangler command. This gets rid of the double messaging we were seeing in the CI logs. Makes running the `build-env` command in the terminal a bit unfriendly if not run with `WRANGLER_LOG="debug"`. However, since this is not a public API, I'm OK with the tradeoff
3. Updated the CI logs as per our conversation

I'll update description with how the final messaging looks like in CI, once all things are built

petebacondarwin: I assume the (BETA) message is in pages-infra?

petebacondarwin (approved):

packages/wrangler/src/pages/build-env.ts
CarmenPopoviciu: I kept the (BETA) message in wrangler, so it stays within our scope, and can remove more easily when time comes

petebacondarwin (approved):

packages/wrangler/src/pages/build-env.ts
```diff
@@ -33,25 +37,37 @@ export const Handler = async (args: PagesBuildEnvArgs) => {
...
+			{
+				...args,
+				// eslint-disable-next-line turbo/no-undeclared-env-vars
+				env: process.env.PAGES_ENVIRONMENT,
+			},
+			true
+		);
+	} catch (err) {
+		logger.debug("wrangler.toml file is invalid. Exiting.");
+		process.exitCode = EXIT_CODE_INVALID_PAGES_CONFIG;
```
jahands, line 68: Why did we do this format instead of throwing a fatal error with a custom error code like we do for other errors communicated to Pages CI? e.g. https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/config/index.ts#L93

simulified:
There is a problem
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 1000; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

packages/wrangler/src/pages/build-env.ts
penalosa: We don't exit with `FatalError`s any more (that one you linked is caught by this catch handler)

PR #5576, opened 2024-04-10
C3: Update Python workers template wrangler toml to include bindings …

jculvey:

Adds commented out bindings examples to the `wrangler.toml` of the python template to align it with the others.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: trivial change
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: trivial change

### DIFF
--- a/packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
+++ b/packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
@@ -3,3 +3,97 @@ name = "<TBD>"
 main = "src/entry.py"
 compatibility_flags = ["python_workers"]
 compatibility_date = "<TBD>"
+
+# Variable bindings. These are arbitrary, plaintext strings (similar to environment variables)
+# Note: Use secrets to store sensitive data.
+# Docs: https://developers.cloudflare.com/workers/wrangler/configuration/#environment-variables
+# [vars]
+# MY_VARIABLE = "production_value"
+
+# Bind the Workers AI model catalog. Run machine learning models, powered by serverless GPUs, on Cloudflare’s global network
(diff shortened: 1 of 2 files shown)


packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
```diff
@@ -3,3 +3,97 @@ name = "<TBD>"
 main = "src/entry.py"
 compatibility_flags = ["python_workers"]
 compatibility_date = "<TBD>"
+
+# Variable bindings. These are arbitrary, plaintext strings (similar to environment variables)
+# Note: Use secrets to store sensitive data.
```
dario-piotrowicz, line 8: should we also add a link to the secrets doc?

(https://developers.cloudflare.com/workers/configuration/secrets/)

dario-piotrowicz (approved):
Looks good to me 🙂

packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
IgorMinar: I like this suggestion.

IgorMinar (approved):

packages/create-cloudflare/templates/hello-world-python/py/wrangler.toml
jculvey: That is a good suggestion indeed.

I'll do this in a followup PR since I want to make the change to all the `wrangler.toml` files. In that same PR I'd like to update the worker templates to use `build-cf-types` as well and get rid of the redundant definition/docs in their source files.

PR #5580, opened 2024-04-10
Allow custom `_worker.bundle` and `_build-metadata.json` files for advanced framework use cases

### LINKED ISSUE #5579: 🚀 Feature Request: support deployments with dynamically generated wrangler.toml (or runtime config in another format)
### Describe the solution

NuxtHub, Nuxt, and Nitro would like to be able to hide / abstract away wrangler.toml configuration for full-stack app deployments via `wrangler pages deploy` and Pages CI.

In the ideal scenario, Nitro (powering Nuxt and NuxtHub) would as part of the build generate a config file (wrangler.toml or more likely a subset of it stored in a simpler format), and this configuration would be picked up by wrangler when the application is deployed to pages.

The configuration supported in this way is limited to JS runtime config (compat flags and date and bindings config). This would enable to frameworks to completely manage runtime configuration on behalf of the user without exposing them to often unnecessary deployment configuration.

Without any other changes, the only way to approximate this behavior now is by the framework generating wrangler.toml and writing it into the source directory of the project, which increases noise for developers and creates a potential for clobbering developer's modifications.

penalosa:

This is pretty much untested—here be dragons 🐉 

Fixes #5579

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/pages/build.ts
+++ b/packages/wrangler/src/pages/build.ts
@@ -1,6 +1,6 @@
 import { createHash } from "node:crypto";
 import { existsSync, lstatSync, mkdirSync, writeFileSync } from "node:fs";
-import { readFile } from "node:fs/promises";
+import { readFile, writeFile } from "node:fs/promises";
 import path, {
 	basename,
 	dirname,
@@ -14,6 +14,7 @@ import { FatalError, UserError } from "../errors";
 import { logger } from "../logger";
 import * as metrics from "../metrics";
 import { isNavigatorDefined } from "../navigator-user-agent";
+import { parseJSON } from "../parse";
 import { buildFunctions } from "./buildFunctions";
 import {
 	EXIT_CODE_FUNCTIONS_NO_ROUTES_ERROR,
@@ -32,6 +33,11 @@ import type {
 	CommonYargsArgv,
 	StrictYargsOptionsToInterface,
 } from "../yargs-types";
+
+type BuildMetadata = {
+	wrangler_config_hash: string;
+	build_output_directory: string;
+};
 
 export type PagesBuildArgs = StrictYargsOptionsToInterface<typeof Options>;
 
@@ -130,6 +136,7 @@ export const Handler = async (args: PagesBuildArgs) => {
 	const validatedArgs = await validateArgs(args);
 
 	let bundle: BundleResult | undefined = undefined;
+	let workerBundleContents: Buffer | undefined = undefined;
 
 	if (validatedArgs.plugin) {
 		const {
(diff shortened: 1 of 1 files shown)


IgorMinar (commented):
I just skimmed through this and it looks roughly right. Thanks for putting it together so quickly @penalosa! We'll need tests, etc for this but this is a great start.

@dario-piotrowicz I believe you are familiar with this part of the code base as well, can you please take a look too?

GregBrimble:
Now we have `wrangler.toml` (and the potential for `main` and `rules` — I don't know if they're actually hooked up yet?), what does `_worker.bundle` give users that they couldn't reasonably express with config options and source themselves?

packages/wrangler/src/pages/build.ts
```diff
@@ -448,18 +470,31 @@ const validateArgs = async (args: PagesBuildArgs): Promise<ValidatedArgs> => {
...
-		if (foundWorkerScript) {
+		const foundWorkerScript = existsSync(prospectiveWorkerScriptPath);
+		if (foundWorkerBundle) {
+			workerBundlePath = prospectiveWorkerBundlePath;
+		} else if (foundWorkerScript) {
 			workerScriptPath = prospectiveWorkerScriptPath;
-		} else if (!foundWorkerScript && !existsSync(args.directory)) {
+		} else if (
+			!foundWorkerScript &&
+			!foundWorkerBundle &&
```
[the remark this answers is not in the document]
petebacondarwin: I appreciate that the previous code also misses this point.

packages/wrangler/src/pages/build.ts
```diff
@@ -483,22 +518,54 @@ We looked for the Functions directory (${basename(
...
+
+			// Ensure we only add known properties to the build metadata
+			buildMetadata =
+				maybeBuildMetadata?.wrangler_config_hash &&
+				maybeBuildMetadata?.build_output_directory
+					? {
+							wrangler_config_hash: maybeBuildMetadata.wrangler_config_hash,
+							build_output_directory: maybeBuildMetadata.build_output_directory,
+					  }
+					: undefined;
```
petebacondarwin, line 545: In the case that this file is not valid, should we be logging a warning and/or perhaps falling through to generating the buildMetadata using the logic below?

packages/wrangler/src/pages/build.ts
```diff
@@ -483,22 +518,54 @@ We looked for the Functions directory (${basename(
 		);
 	}
 
+	// The build metadata for this build (a hash of the config file it was based on, and the output directory that was used) is usually generated from wrangler.toml
+	// However, frameworks may want to allow users to use a completely different config file, and so may want to fully generate a `_build-metadata.json` file themselves
+	// This is an advanced use case, and should go along with a custom `_worker.bundle` file. The vast majority of users won't need to cutomise this!
```
dario-piotrowicz, line 523: would it make sense here to error if a custom `_build-metadata.json` file is used but no custom `_worker.bundle` file is? (e.g. `if (existsSync(prospectiveBuildMetadataPath) && !foundWorkerBundle) throw`)

dario-piotrowicz (commented):
Looks good to me, though I agree with @GregBrimble that ideally the same should be applied to workers so that we don't diverge workers and pages

IgorMinar:
from meeting with @pi0: would enabling a runtime flag via this feature disable the binding config ui in the dash?

pi0:
I hate to be the one to ping on here but it is a huge blocker for us to be able to skip double wrangler transforms for Nitro build output.

penalosa:
Closing in favour of a new approach from @petebacondarwin

PR #5589, opened 2024-04-11
Fix `"module": "....js"` package resolution in vitest-pool-workers

### LINKED ISSUE #5588: 🐛 BUG: Vitest-pool-workers doesn't handle ESM modules well

Workers Vitest Integration

vitest-pool-workers: 0.1.18

20.12.1

Mac Sonoma 14.4.1

### Observed behavior

When importing an NPM library that has `"module": "foo.js"` in it's package JSON, but doesn't have `"type": "module"`, vitest-pool-workers fails to resolve the dependency, and you get an error linked below.

### Expected behavior
Vitest-pool-workers should handle `"module"` in the package.json.

### Steps to reproduce
Write a test with a worker that imports a dependency using `"module"`, e.g. `toucan-js`. I can provide an internal repo if needed, run vitest, observe the error.

_No response_

```js
workerd/server/server.c++:2802: error: Fallback service failed to fetch module; payload = ; spec = /?specifier=%2F<snip>%2Fnode_modules%2F.pnpm%2Ftoucan-js%403.3.1%2Fnode_modules%2Ftoucan-js%2Fdist%2Findex.esm.js&referrer=%2F<snip>node_modules%2F.pnpm%2F<snip>%2Fnode_modules%2F<snip>%2Flib%2Fsentry.js

 FAIL  tests/index.test.ts [ tests/index.test.ts ]
Error: No such module "<snip>/node_modules/.pnpm/<snip>/node_modules/<snip>/lib/toucan-js".
  imported from "<snip>/node_modules/.pnpm/<snip>/node_modules/<snip>/lib/sentry.js"
```

Skye-31:

Fixes #5588.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Bugfix


fixtures/vitest-pool-workers-examples/external-package-resolution/src/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
 	"extends": "../../tsconfig.workerd.json",
-	"include": ["./**/*.ts"]
+	"include": ["./**/*.ts"],
+	"compilerOptions": {
+		"skipLibCheck": true
```
petebacondarwin, line 5: Let's put a comment here explaining why we are not checking libs - i.e. that the toucan-js references a different workers-types version which breaks overall type checking.

petebacondarwin (approved):
I think this is a reasonable fix, although I note from https://nodejs.org/api/esm.html#resolution-algorithm-specification that the use of `.esm.js` in toucan-js, while not providing a `"type": "module"` property in package.json appears to be wrong.

It might be that esbuild is more lenient than that algorithm, which is why it has not been a problem so far.

I think we should try to upstream a fix to toucan if possible.

.changeset/empty-llamas-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vitest-pool-workers": patch
+---
+
+Fix `"module": "..."` resolution in vitest-pool-workers
```
petebacondarwin, line 5: This is a very vague changeset and not really very actionable by the person reading the released change log. (Also the format is slightly off).

Can I propose we fill it out a bit:

```suggestion
fix: Support importing ES modules from libraries that do not correctly provide `"type"="module"` not use `.mjs` extensions

The toucan-js library has an entry point of `"module": "dist/index.esm.js"`. This file does not use the standard `.mjs` extension, nor does it specify `"type"="module"`, so the resolution and loading algorithm fails to identify this file as an ES Module, defaulting to CommonJS, breaking Vitest.

Fixes #5588
```

Skye-31:
> while not providing a "type": "module" property in package.json appears to be wrong
As far as I understand, this `"module"` field was added before `exports` became a thing, and there was no way to indicate you supported both ESM & CJS exports. 

While we could upstream the fix to toucan, there are a number of other libraries (from a [github search](https://github.com/search?q=path%3Apackage.json+AND+language%3Ajson+%22%5C%22module%5C%22%3A+%5C%22%22&type=code), apparently more than 600 thousand!) doing this same thing

PR #5599, opened 2024-04-12
Add basic support for `wrappedBindings` and RPC entrypoints to Miniflare's Magic Proxy

### LINKED ISSUE #5575: 🐛 BUG: `getPlatformProxy` doesn't provide an AI binding when it should

Wrangler core

`wrangler@3.48.0`

(the issue has been introduced in `3.42.0`, earlier versions don't present this issue)

20

Mac

### Observed behavior

`getPlatformProxy` with a `wrangler.toml` file containing an AI binding declaration:
```toml
[ai]
binding = "AI"
```
does not produce a proxy to the AI binding

### Expected behavior

It should produce an AI binding, meaning that with:
```js
import { getPlatformProxy } from 'wrangler';

const { env } = await getPlatformProxy();
```
`env.AI` should contain a valid proxy object

https://github.com/dario-piotrowicz/getPlatformProxy-ai-undefined-repro

### LINKED ISSUE #5577: 🐛 BUG: Miniflare's `getBindings` does not properly expose RPC entrypoints

Miniflare

miniflare@3.20240404.0

20

Mac

### Observed behavior
[RPC entrypoints](https://blog.cloudflare.com/javascript-native-rpc) can be defined and used within Miniflare (meaning that they do work for scripts running in miniflare itself).

The issue is that they don't seem available from `Miniflare#getBindings`, the method does return valid proxies for such bindings but trying to access the RPC methods does not work and generates the following [workerd warning](https://github.com/cloudflare/workerd/blob/f77cdc6f08be2a29ef1e35118f86fbd4fdc008b7/src/workerd/api/http.c%2B%2B#L2117-L2122) (regardless on the compat date and flags used):
Screenshot 2024-04-10 at 18 09 53

### Expected behavior

`Miniflare#getBindings` should return bindings with RPC that can be used in the same manner as they can inside Miniflare itself.

https://github.com/dario-piotrowicz/miniflare-no-rpc-binding-repro/tree/main

penalosa:

This fixes a number of issues with the Miniflare magic proxy, including:
- Fixes #5575 
- Fixes #5577 (partial support, not including stubs or calling returned functions)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: bugfixes

### DIFF
--- a/.changeset/quick-parents-float.md
+++ b/.changeset/quick-parents-float.md
@@ -1,0 +1,48 @@
+---
+"miniflare": patch
+---
+
+fix: add support for wrapped bindings in magic proxy
+
+currently `Miniflare#getBindings()` does not return proxies to provided `wrappedBindings`, make sure that appropriate proxies are instead returned
+
+Example:
+
+```ts
+import { Miniflare } from "miniflare";
+
+const mf = new Miniflare({
+	workers: [
+		{
+			wrappedBindings: {
+				Greeter: {
+					scriptName: "impl",
+				},
+			},
+			modules: true,
+			script: `export default { fetch(){ return new Response(''); } }`,
+		},
+		{
+			modules: true,
+			name: "impl",
+			script: `
+        class Greeter {
+          sayHello(name) {
+            return "Hello " + name;
+          }
+        }
+
+        export default function (env) {
+          return new Greeter();
+        }
+      `,
+		},
+	],
+});
+
+const { Greeter } = await mf.getBindings();
+
+console.log(Greeter.sayHello("world")); // <--- prints 'Hello world'
+
+await mf.dispose();
+```
--- a/.changeset/quick-parents-float.md
+++ b/.changeset/quick-parents-float.md
@@ -26,16 +26,16 @@ const mf = new Miniflare({
 			modules: true,
 			name: "impl",
 			script: `
-        class Greeter {
-          sayHello(name) {
-            return "Hello " + name;
-          }
-        }
+				class Greeter {
+					sayHello(name) {
+						return "Hello " + name;
+					}
+				}
 
-        export default function (env) {
-          return new Greeter();
-        }
-      `,
+				export default function (env) {
+					return new Greeter();
+				}
+			`,
 		},
 	],
 });
--- a/fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
+++ b/fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
@@ -156,8 +156,15 @@ describe("getPlatformProxy - bindings", () => {
 	});
 
 	type EntrypointService = Service<
-		NamedEntrypoint & Rpc.WorkerEntrypointBranded
-	>;
+		Omit<NamedEntrypoint, "getCounter"> & Rpc.WorkerEntrypointBranded
+	> & {
+		getCounter: () => Promise<
+			Promise<{
+				value: Promise<number>;
+				increment: (amount: number) => Promise<number>;
+			}>
+		>;
+	};
 
 	describe("provides rpc service bindings to external local workers", () => {
 		let rpc: EntrypointService;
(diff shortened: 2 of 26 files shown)


packages/miniflare/src/workers/core/proxy.worker.ts
```diff
@@ -255,7 +259,27 @@ export class ProxyServer implements DurableObject {
 			}
 			assert(Array.isArray(args));
 			try {
-				result = func.apply(target, args);
+				if (func.constructor.name === "RpcProperty") {
```
Skye-31, line 262: Do you need to support `DurableObject` as well?

dario-piotrowicz: durable objects should already be supported 🤔

Skye-31: I meant `class ... extends DurableObject` - these support RPC methods

dario-piotrowicz: I see... I assume that those will just be wrapped into an `RpcPromise` here and their properties will be 
be constructed as `RpcProperty`, I'll have a look to make sure! 🙂👍 

(PS: I'm taking over the PR as Samuel's OOO and we'd like to land this soon-ish)

dario-piotrowicz: Yeah I can confirm that in DOs we also get the functions as `RpcProperty`

And that the current implementation works as you'd expect it to 🙂 

Screenshot 2024-04-13 at 15 34 03

Screenshot 2024-04-13 at 15 34 13

However I don't think that's going to work currently with `getPlatformProxy` since DO rpc stubs are not supported with multiple `wrangler dev` sessions (which is what `getPlatformProxy` relies on):
Screenshot 2024-04-12 at 19 29 43

We'll have to implement that too, but that feels tangentially related and I'd leave it as a followup not to scope creep this PR 🙂

dario-piotrowicz:
With the correct env variables the wrangler AI e2e passes locally:
Screenshot 2024-04-13 at 16 57 54

on Github I think we need to set:
`WRANGLER_IMPORT` (to: `packages/wrangler/wrangler-dist/cli.js`)
and make sure that the `CLOUDFLARE_API_TOKEN` grants access to Workers AI (I've tested with this with my own Cloudflare account id and api token)

I don't think I have permissions to do the above, maybe someone could help me? 🙂 @petebacondarwin? 🙏

dario-piotrowicz:
> [!NOTE]
> The `wrappedBindings` support is not perfect, it seems to break in some cases: https://github.com/dario-piotrowicz/miniflare-wrappedBindings-getBindings-function-stringification-error-repro
> but it is good enough for the AI binding, so if it'd be acceptable I'd like to proceed with the current solution
> (to enable AI binding proxies as soon as we can) and address the above issue as a follow up

fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
```diff
@@ -164,8 +193,6 @@ describe("getPlatformProxy - bindings", () => {
 		await dispose();
 	});
 
-	// Note: the following test is skipped due to flakiness caused by the local registry not working reliably
-	//       when we run all our fixtures together (possibly because of race condition issues)
```
dario-piotrowicz, line 168: The test below needs to still be skipped at it doesn't seem to be passing 😢

It is also not passing on the latest stable wrangler release so it's likely an unrelated regression 😓

RamIdeas (approved):

.changeset/quick-parents-float.md
```diff
@@ -0,0 +1,48 @@
...
+			script: `
+        class Greeter {
+          sayHello(name) {
+            return "Hello " + name;
+          }
+        }
+
+        export default function (env) {
+          return new Greeter();
+        }
```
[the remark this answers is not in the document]
dario-piotrowicz: Fixed, thanks 😄

fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
dario-piotrowicz: This is being addressed in https://github.com/cloudflare/workers-sdk/pull/5669 🙂

PR #5612, opened 2024-04-13
Feat(vitest-pool-workers): Support specifying an environment in tests

### LINKED ISSUE #5611: 🚀 Feature Request: Support environments in vitest-pool-workers
### Describe the solution

Currently, there's no way to choose an environment when using the `wrangler: { configPath }` option in the vitest integration, meaning if your bindings aren't available at the root level, you can't use them in your test

Skye-31:

Fixes #5611 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): <https://github.com/cloudflare/cloudflare-docs/pull/14014>
  - [ ] Not necessary because:

### DIFF
--- a/fixtures/vitest-pool-workers-examples/d1/wrangler.toml
+++ b/fixtures/vitest-pool-workers-examples/d1/wrangler.toml
@@ -3,7 +3,7 @@ main = "src/index.ts"
 compatibility_date = "2024-01-01"
 compatibility_flags = ["nodejs_compat"]
 
-[[ d1_databases ]]
+[[env.production.d1_databases]]
 binding = "DATABASE"
 database_name = "database"
 database_id = "00000000-0000-0000-0000-000000000000"
--- a/.changeset/rich-jokes-chew.md
+++ b/.changeset/rich-jokes-chew.md
@@ -1,0 +1,24 @@
+---
+"@cloudflare/vitest-pool-workers": minor
+---
+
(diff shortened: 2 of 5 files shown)


fixtures/vitest-pool-workers-examples/d1/wrangler.toml
```diff
@@ -3,7 +3,7 @@ main = "src/index.ts"
 compatibility_date = "2024-01-01"
 compatibility_flags = ["nodejs_compat"]
 
-[[ d1_databases ]]
+[[env.production.d1_databases]]
```
CarmenPopoviciu, line 6: can we please keep both top-level and production-specific configuration for D1, and adjust the tests such that we test for:
1. behaviour is as expected if we don't pass the `environment` argument
2. behaviour is as expected if we do pass in the `environment` arg
🙏

CarmenPopoviciu: turns out we can't test for both scenarios with just one fixture. We'd need to create a completely new fixture for this, since config is specified at the fixture level. 

We can consider that current fixture tests use-case `2`, and all the others carry on working without specifying an environment, so use-case `1`.

CarmenPopoviciu (approved):

PR #5619, opened 2024-04-15
fix(wrangler) wrangler kv:key put does not work as expected

### LINKED ISSUE #2911: 🐛 BUG: wrangler kv:key put does not work as expected

KV, Wrangler

2.12.3

Linux

### Describe the Bug

wrangler **kv:key put** does not work as expected when using --path with --metadata commandline options.
Meaning of "as expected": the same bytes can not be read with **kv:key get**, that were previously put.

### Simple way to test:

**This works: the same file can be get, which were put:**
md5 original.png
81a3478c5f2b12823f3ff0253c3388e7
wrangler kv:key put --binding=FILES kvimage.png --path=original.png
wrangler kv:key get --binding=FILES kvimage.png > copy.png
md5 copy.png
81a3478c5f2b12823f3ff0253c3388e7

**This does NOT work, md5 does not match, the bytes are different:**
md5 original.png
81a3478c5f2b12823f3ff0253c3388e7
wrangler kv:key put --binding=FILES kvimage.png --path=original.png --metadata='{"contentType":"image/png"}'
wrangler kv:key get --binding=FILES kvimage.png > copy.png
md5 copy.png
40621887293ab018c535120fe0b6bf92

### Looking code:

I think the problem is that wrangler is putting the Buffer data into FormData, to be sent to KV. And this is a wrong type of data to be added into FormData.

--path file is read with readFileSyncToBuffer, which returns node-Buffer
https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/kv/index.ts#L236

This is given to function putKVKeyValue
https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/kv/helpers.ts#L203

which tries to append the Buffer into undici FormData object.
https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/kv/helpers.ts#L193

undici FormData append expects value to be eg. Blob or File, but receives Buffer. So Buffer is converted to string.
https://github.com/nodejs/undici/blob/main/types/formdata.d.ts#L27

gmemstr:

Currently, when passing a path to `kv:key put --path`, the contents of the file are read in as a Buffer, which would normally be fine - however, when we *also* have `--metadata` defined, the payload is transformed into `FormData`. `FormData` can take a `string`, `Blob` or `File`, and will fall back to `string` for everything else. Because we loaded the file as a Buffer, the file contents are then transformed into a string by the library being used (`undici`). *Normally* this is okay *unless* you are not uploading a text-based document, e.g an image file.

Fixes #2911.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: 

Minor change correcting behaviour to what is expected

### DIFF
--- a/packages/wrangler/src/kv/helpers.ts
+++ b/packages/wrangler/src/kv/helpers.ts
@@ -1,3 +1,4 @@
+import { Blob } from "node:buffer";
 import { URLSearchParams } from "node:url";
 import { Miniflare } from "miniflare";
 import { FormData } from "undici";
@@ -200,7 +201,7 @@ function asFormData(fields: Record<string, unknown>): FormData {
 	const formData = new FormData();
 
 	for (const [name, value] of Object.entries(fields)) {
-		formData.append(name, value);
+		formData.append(name, Buffer.isBuffer(value) ? new Blob([value]) : value);
 	}
 
 	return formData;
--- a/.changeset/purple-cars-run.md
+++ b/.changeset/purple-cars-run.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: correctly handle non-text based files for kv put
+
+The current version of the kv:key put command with the --path argument will treat file contents as a string because it is not one of Blob or File when passed to the form helper library. We should turn it into a Blob so it's not mangling inputs.
--- a/packages/wrangler/src/__tests__/kv.test.ts
+++ b/packages/wrangler/src/__tests__/kv.test.ts
@@ -505,6 +505,29 @@ describe("wrangler", () => {
 				expect(std.err).toMatchInlineSnapshot(`""`);
 			});
 
+			it("should put a key with a binary value and metadata", async () => {
+				const buf = Buffer.from(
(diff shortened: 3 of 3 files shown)


packages/wrangler/src/kv/helpers.ts
```diff
@@ -199,7 +199,10 @@ export function unexpectedKVKeyValueProps(keyValue: KeyValue): string[] {
 function asFormData(fields: Record<string, unknown>): FormData {
 	const formData = new FormData();
 
-	for (const [name, value] of Object.entries(fields)) {
+	for (let [name, value] of Object.entries(fields)) {
```
gmemstr, line 202: Trips up the prefer-const eslint line, not sure if we want to make the function a bit messier with an if/else or disable the rule for this line.

petebacondarwin: Let's go with:

```ts
for (const [name, value] of Object.entries(fields)) {
  formData.append(name, Buffer.isBuffer(value) ? new Blob(value) : value);
}
```

Issue #1586, opened 2022-07-30
🐛 BUG: Windows+node:fs=> ALL Pages uploads/updates fail with 2000+ files [ERROR EMFILE]


torcd:
### System tested

`OS:` Windows 64-bit
`pm:` nvm-windows
`node.js:` 16.7.0 (lowest supported), 16.16.0 (lts), 17.9.0, 18.5.0, 18.7.0 (latest), likely all versions affected
`wrangler:` 2.0.17, 2.0.23, likely all versions >2.0.6 affected
`dependencies:` all updated to "wanted" versions
`ERROR src:` -`node:fs`, -`chokidar`, -wrangler pages `publish/upload` code (eg lack of _graceful_ fs queuing/retries)

### Issue:

On win64, when publishing a few thousand files via `wrangler pages publish`, the uploads fail due to chronic `ERROR EMFILE: too many files open '<C:\path\to\last\read\file>'`  errors.

The uploads either fail:
1. during the dir walk/read/check stage if a previous deployment exceeds [6000-8000] files -> thus, one can't reliably update even a single file if larger previous deployments went through

OR

2. during the upload stage when the total batch of files is in the [2K-4K]+ range -> ie, even the first deployment will fail if you start with a larger batch

### Similar bugs/fixes elsewhere:
**[#141 chokidar: Use graceful-fs to avoid EMFILE?](https://github.com/paulmillr/chokidar/issues/141)
[ #624 IPFS: upload directory with subdirectories [WIP]](https://github.com/ipfs/ipfs-desktop/pull/624#issuecomment-375607420)**

### Potential Fix/Solution:
The bug likely stems from use of the `core node fs` module and how files are watched and read in the pipeline. While `chokidar+fsevents` might be preventing EMFILE errors on Mac, no such mitigations appear to be used on Windows systems. Instead of queuing/waiting/retrying, the uploads fail entirely once a single EMFILE error is thrown, rendering Pages projects with a few thousand files impossible to maintain even though wrangler technically allows 20,000 files per upload. :((

Can a test build with **_graceful-fs +wrangler publish/upload code `fs` replacements_** be temporarily made available at least until a permanent Windows-specific fix is released? Thanks.

Cherry:
Ran into this myself today with a large project. `graceful-fs` seems to be a good drop-in replacement to queue up the calls and more gracefully handle `EMFILE` errors.

PR #5632, opened 2024-04-17
Batch file uploads to avoid EMFILE

horo-fox:

Fixes https://github.com/cloudflare/workers-sdk/issues/1586 (I think, at least for me).

I checked this helps by modifying my `node_modules` installed version. I'm pretty sure I copied over the edits, but not completely sure. Is there a way to test this? I'm not too familiar with JS and package managers.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: hard to test
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because:

### DIFF
--- a/packages/wrangler/src/pages/upload.tsx
+++ b/packages/wrangler/src/pages/upload.tsx
@@ -209,16 +209,18 @@ export const upload = async (
 
 			for (let i = 0; i < bucket.files.length; i += 1000) {
 				// only read up to 1000 files, from disk, at a time to avoid `EMFILE` error (on Windows)
-				payload.push(...(await Promise.all(
-					bucket.files.slice(i * 1000, (i + 1) * 1000).map(async (file) => ({
-						key: file.hash,
-						value: (await readFile(file.path)).toString("base64"),
-						metadata: {
-							contentType: file.contentType,
-						},
-						base64: true,
-					}))
-				)));
+				payload.push(
+					...(await Promise.all(
+						bucket.files.slice(i * 1000, (i + 1) * 1000).map(async (file) => ({
+							key: file.hash,
+							value: (await readFile(file.path)).toString("base64"),
+							metadata: {
+								contentType: file.contentType,
+							},
+							base64: true,
+						}))
+					))
+				);
 			}
 
 			try {
--- a/packages/wrangler/src/pages/upload.tsx
+++ b/packages/wrangler/src/pages/upload.tsx
@@ -208,7 +208,7 @@ export const upload = async (
 			const payload: UploadPayloadFile[] = [];
 
 			for (let i = 0; i < bucket.files.length; i += 1000) {
-				// upload files 1k at once to avoid EMFILE
+				// only read up to 1000 files, from disk, at a time to avoid `EMFILE` error (on Windows)
 				payload.push(...(await Promise.all(
 					bucket.files.slice(i * 1000, (i + 1) * 1000).map(async (file) => ({
 						key: file.hash,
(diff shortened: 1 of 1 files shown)


packages/wrangler/src/pages/upload.tsx
```diff
@@ -200,16 +200,21 @@ export const upload = async (
...
-					metadata: {
-						contentType: file.contentType,
-					},
-					base64: true,
-				}))
-			);
+			const payload: UploadPayloadFile[] = [];
+
+			for (let i = 0; i < bucket.files.length; i += 1000) {
+				// upload files 1k at once to avoid EMFILE
```
petebacondarwin, line 206: I think what we mean here is:

```suggestion
				// only read up to 1000 files, from disk, at a time to avoid `EMFILE` error (on Windows)
```

petebacondarwin (commented):
Let me see if I can create a test to prove this works...

packages/wrangler/src/pages/upload.tsx
petebacondarwin: Should we just consider using graceful-fs instead?

horo-fox: Maybe, but:

1) this avoids a dep
2) this is faster (presumably) and more obviously correct

The only case that handles better is if there's a weird OS where the limit on the # of file descriptors is <1k which is not something to be worried about IMO. But we can use `graceful-fs` here ultimately, I'm not sure how to weigh the pros vs cons.

horo-fox:
The workflows seem like they need an approval to start...

horo-fox:
I don't think the error is caused by this PR, though it happening multiple times is weird... But I don't know what here could cause an "ELIFECYCLE" (whatever that means).

This seems to work though:

```
PS > npx https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/8722250031/npm-package-wrangler-5632 pages deploy dist    
Need to install the following packages:
https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/8722250031/npm-package-wrangler-5632
Ok to proceed? (y)
npm WARN deprecated rollup-plugin-inject@3.0.2: This package has been deprecated and is no longer maintained. Please use @rollup/plugin-inject.
npm WARN deprecated sourcemap-codec@1.4.8: Please use @jridgewell/sourcemap-codec instead
🌏  Uploading... (16520/16520)

✨ Success! Uploaded 3965 files (12555 already uploaded) (113.58 sec)

✨ Deployment complete! Take a peek over at https://8efe3fc4.classes-1ao.pages.dev
```

(there's the extra concern that my uploaded pages 500 despite not having a function or anything, but that's a different issue presumably)

petebacondarwin:
Just as an update, I am not able to create a test that fails on Windows with this error yet. See https://github.com/cloudflare/workers-sdk/pull/5635

petebacondarwin:
Good question. I think we could land this without a test that proves it fixes the Windows problem. We have tests that the feature basically works in normal scenarios.

petebacondarwin:
@horo-fox - we had to revert this as it had a bug. Notably I see this comment from you earlier:

> (there's the extra concern that my uploaded pages 500 despite not having a function or anything, but that's a different issue presumably)

That is actually the bug in action. Above 1000 files this was not uploading correctly, and so the 500s you were seeing were missing files that had not been uploaded.

PR #5654, opened 2024-04-18
cache + speedup windows e2e

penalosa:

This adds additional turbo caching to e2e builds, as well as speeding up Wrangler installs when running e2e for Windows

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: testing change
- Public documentation
  - [ ] TODO (before merge)


petebacondarwin (approved):

.github/workflows/changesets.yml
```diff
@@ -39,7 +39,15 @@ jobs:
       - name: Check the changesets
         run: node -r esbuild-register tools/deployments/validate-changesets.ts
 
-      - name: Check the build
+      - name: "Create .env file"
```
petebacondarwin, line 42: NIT: the description of this step is not very helpful. Can we describe the intent instead?

```suggestion
      - name: "Configure turbo repo cache environment variables"
```

.github/workflows/e2e.yml
```diff
@@ -61,34 +69,23 @@ jobs:
           NODE_ENV: "production"
         working-directory: packages/wrangler
 
-      - name: Find Wrangler
+      - name: Find pnpm global dir
```
petebacondarwin, line 72: No need for abbreviation here

```suggestion
      - name: Find pnpm global directory
```

petebacondarwin (changes requested):
There are quite a few changes to what trigger different jobs to run. Please can you update `.github/workflows/README.md` with this information.
Can you also explain the reasoning behind moving some tests from PRs to main? We recently did the opposite. One of the reasons was that it was not as easy to be notified if there is a failure on main compared to on a PR (such as Version Packages) and such failed jobs don't block PRs from landing. Perhaps I am misunderstanding the changes here.

vitest.shared.ts
```diff
@@ -17,5 +17,8 @@ export default defineConfig({
 			useAtomics: true,
 		},
 		restoreMocks: true,
+		// A lot of the fixture tests are extremely flaky because of the dev registry
+		// Retry tests by default so that only real errors are reported
+		retry: 2,
```
CarmenPopoviciu, line 22: I'm not against this change, but was wondering if adding these retries won't increase our build times even more

penalosa:
> Can you also explain the reasoning behind moving some tests from PRs to main?

The reason for moving these to main is that on the version packages PR the versions of packages to be released are bumped. This invalidates the turbo cache (and isn't configurable), and so we'd lose the benefit of caching for running tests before release. With our new process of locking main pre release, we'll just have to make sure that main is green before merging the changesets PR.

vitest.shared.ts
penalosa: This only affects packages which run tests with vitest, which are all so fast it doesn't make much of a difference (it excludes wrangler and miniflare, for instance). I think the extra robustness should make up for potentially longer build times

.github/workflows/test-and-check.yml
```diff
@@ -1,11 +1,16 @@
-name: Pull Request
+name: Tests + Checks
 
-on: pull_request
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
CarmenPopoviciu, line 7: is this left "empty" on purpose?

CarmenPopoviciu:
with the relatively limited context I have, this LGTM. Left all the comments I had for things that stood out. Let's pls get one more approval before we land this, and thank you for working on it Samuel <3

petebacondarwin:
Discussed offline: we agreed not to run the tests on main but leave the on the Version Packages PR.
Needs a bit of a rebase.

petebacondarwin (approved):
Magic!

PR #5669, opened 2024-04-20
[wrangler] fix: fix local Durable Object proxies not working

### LINKED ISSUE #5620: 🐛 BUG: `getPlatformProxy` DO proxy "undefined" is not a valid JSON

Wrangler core

wrangler@3.50.0

20

Mac

### Observed behavior
When using `getPlatformProxy` with recent version of wrangler (v>=3.46.0) it seems like DurableObject proxies don't properly work and when trying to fetch from them an response saying that `undefined` is not a valid JSON value is presented:
ko

### Expected behavior
Fetching from a DurableObject proxy should work as expected (as it worked with wrangler versions <= 3.45.0):
ok

### Minimal reproduction

https://github.com/dario-piotrowicz/getPlatformProxy-DO-JSON-issue-repro

### LINKED ISSUE #5658: 🐛 BUG: requests from workers to DOs (in miniflare) broken in wrangler 3.46+

Wrangler core, Miniflare

3.51.2 [wrangler]

v18.19.0

Mac OS Sonoma 14.4.1

### Expected behavior

The behavior I expect is:

- I can bind a durable object that runs under one worker (worker A) to another worker (worker B)
- From worker B, I can create a durable object stub and perform a request on worker A's DO.

### Observed behavior

Instead, when I make a request from worker B, the request does not reach the DO class, and instead I see an exception:

```
SyntaxError: "undefined" is not valid JSON
    at async Object.fetch (file:///Users/jesseditson/Desktop/cloudflare-test/node_modules/miniflare/dist/src/workers/core/entry.worker.js:954:22)
```

### Steps to reproduce

I've posted a minimal repro repo here: https://github.com/jesseditson/wrangler-do-error-repro

To set it up, clone the repo then run `npm install`.

To reproduce the issue, run (in two seperate terminals)

`npm run do`
`npm run api`

Then visit `http://localhost:8777/test` in a browser or run `curl http://localhost:8777/test`

### Investigation

Looking at the actual entry.worker.js, it looks like some headers are auto-injected - when I log the request path and headers, the offending call has:

```
path: /__WRANGLER_EXTERNAL_DURABLE_OBJECTS_WORKER

Headers(6) {
  'accept-encoding' => 'br, gzip',
  'host' => 'localhost:8787',
  'x-miniflare-durable-object-cf-blob' => 'undefined',
  'x-miniflare-durable-object-id' => '899153785ebdad0ee7fe9208aae7ed48307ff327588dddf7e767a75daee36277',
  'x-miniflare-durable-object-name' => 'TestDO',
  'x-miniflare-durable-object-url' => 'http://do/hello',
  [immutable]: false
}
```

The `x-miniflare-durable-object-cf-blob` appears to be the issue, and sure enough when I add a line before the request:

`request.headers.delete("x-miniflare-durable-object-cf-blob")`

things work as expected.

### Resolution

I've verified that this is a regression by running `npm i --save-dev wrangler@3.45` in the posted repro, which then no longer reproduces this error.

I have not yet looked into the patches merged between .45 and .51, but it also seems possible that the error exists outside this repo, as I can't grep any reads of the `x-miniflare-durable-object-cf-blob` header.

https://github.com/jesseditson/wrangler-do-error-repro

_No response_

dario-piotrowicz:

This PR solves the regression of local Durable Object proxies not working (a regression introduced in wrangler 3.46.0, more specifically in https://github.com/cloudflare/workers-sdk/pull/5215)

To test the fix please try out the minimal reproductions of the linked issues 👇 with both the latest wrangler version and this [PR's prerelease](https://github.com/cloudflare/workers-sdk/pull/5669#issuecomment-2067679382) (the former will not work while the latter would work as intended)

Fixes #5620
Fixes #5658

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: bug/regression fix

### DIFF
--- a/.changeset/odd-guests-beam.md
+++ b/.changeset/odd-guests-beam.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: fix broken local Durable Object proxies
--- a/.changeset/odd-guests-beam.md
+++ b/.changeset/odd-guests-beam.md
@@ -2,4 +2,10 @@
 "wrangler": patch
 ---
 
-fix: fix broken local Durable Object proxies
+fix: fix broken Durable Object local proxying (when no `cf` property is present)
+
+A regression was introduced in wrangler 3.46.0 (https://github.com/cloudflare/workers-sdk/pull/5215)
+which made it so that missing `Request#cf` properties are serialized as `"undefined"`, this in turn
+throws a syntax parse error when such values are parsed via `JSON.parse` breaking the communication
+with Durable Object local proxies. Fix such issue by serializing missing `Request#cf` properties as
+`"{}"` instead.
--- a/fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
+++ b/fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
@@ -207,7 +207,7 @@ describe("getPlatformProxy - bindings", () => {
 		await dispose();
 	});
 
-	it.skip("correctly obtains functioning DO bindings (provided by external local workers)", async () => {
+	it("correctly obtains functioning DO bindings (provided by external local workers)", async () => {
 		const { env, dispose } = await getPlatformProxy<Env>({
 			configPath: wranglerTomlFilePath,
 		});
--- a/packages/wrangler/src/dev/miniflare.ts
+++ b/packages/wrangler/src/dev/miniflare.ts
@@ -97,10 +97,10 @@ function createProxyPrototypeClass(handlerSuperKlass, getUnknownPrototypeKey) {
 				return getUnknownPrototypeKey(key);
 			}
 		});
-	
+
 		return Reflect.construct(handlerSuperKlass, [ctx, env], klass);
 	}
-		
+
 	Reflect.setPrototypeOf(klass.prototype, handlerSuperKlass.prototype);
 	Reflect.setPrototypeOf(klass, handlerSuperKlass);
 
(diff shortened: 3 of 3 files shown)


.changeset/odd-guests-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: fix broken local Durable Object proxies
```
petebacondarwin, line 5: NIT: this description is a bit enigmatic... could you provide a bit more description? I.e. the fact that previously if there was not a `cf` property on the request it was being rendered in the header as `undefined` that could not then be JSON parsed causing the "DO proxy "undefined" is not a valid JSON" error.

petebacondarwin (approved):

.changeset/odd-guests-beam.md
dario-piotrowicz: sure I'll expand on this 🙂 

I didn't go too much in details since the breakage was all internal and not something related to the user code so I figured that keeping it generic in the changeset made sense 🤔

dario-piotrowicz: I've updated the changeset, please have another look 🙏

petebacondarwin (approved):
Love it!

PR #5670, opened 2024-04-20
[miniflare] fix: make the magic proxy able to proxy objects containing functions

dario-piotrowicz:

This PR makes it so that "magic" proxy objects can contain functions, something that is not currently possible.

To test the changes in this PR you can try this [minimal reproduction/example](https://github.com/dario-piotrowicz/miniflare-wrappedBindings-getBindings-function-stringification-error-repro) with both the latest version of miniflare and this PR's miniflare prerelease. You can see how the example errors with the former but succeeds with the latter.

> [!NOTE]
> This improves the situation but the magic proxy's handling of functions is not yet perfect, for example functions inside arrays (e.g. `return [() => 'Hello']`) still don't work. Also functions returning functions do not work either.
>
> In any case I believe that the changes here improve the situation and are unlikely to conflict with any fix for the above points, so I think that it should be safe to proceed with these changes (and address the other points separately).

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)

### DIFF
--- a/.changeset/wet-taxis-share.md
+++ b/.changeset/wet-taxis-share.md
@@ -2,4 +2,4 @@
 "miniflare": patch
 ---
 
-fix: make the magic proxy able to serialize objects containing functions
+fix: make the magic proxy able to proxy objects containing functions
--- a/.changeset/wet-taxis-share.md
+++ b/.changeset/wet-taxis-share.md
@@ -2,4 +2,6 @@
 "miniflare": patch
 ---
 
-fix: make the magic proxy able to proxy objects containing functions
+fix: Allow the magic proxy to proxy objects containing functions
+
+This was previously prevented but this change removes that restriction.
--- a/.changeset/wet-taxis-share.md
+++ b/.changeset/wet-taxis-share.md
@@ -1,0 +1,5 @@
+---
+"miniflare": patch
+---
+
+fix: make the magic proxy able to serialize objects containing functions
--- a/packages/miniflare/src/workers/core/proxy.worker.ts
+++ b/packages/miniflare/src/workers/core/proxy.worker.ts
@@ -55,12 +55,37 @@ function isPlainObject(value: unknown) {
 	if(value?.constructor?.name === 'RpcStub') {
 		return false;
 	}
+	if(isObject(value)) {
+		const valueAsRecord = value as Record<string, unknown>;
+		if (objectContainsFunctions(valueAsRecord)){
+			return false;
+		}
+	}
 	return (
 		proto === Object.prototype ||
 		proto === null ||
 		Object.getOwnPropertyNames(proto).sort().join("\0") === objectProtoNames
 	);
 }
+function objectContainsFunctions(obj: Record<string, unknown>): boolean {
+	for(const [, entry] of Object.entries(obj)) {
+		if(typeof entry === 'function') {
+			return true;
+		}
+		if(isObject(entry)) {
+			if(objectContainsFunctions(entry as Record<string, unknown>)) {
+				return false;
+			}
+		}
+	}
+
+	return false;
+}
+
+function isObject(value: unknown) {
+	return value && typeof value === 'object';
+}
+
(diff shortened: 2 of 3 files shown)


petebacondarwin (approved):
Is it worth adding tests that reads non-function properties off an object that also has functions?

dario-piotrowicz:
> Is it worth adding tests that reads non-function properties off an object that also has functions?

yeah that's a good shout, I'll look into adding that 🙂👍

.changeset/wet-taxis-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": patch
+---
+
+fix: make the magic proxy able to serialize objects containing functions
```
RamIdeas, line 5: Can you make clear that this fix doesn't allow those functions to be callable

petebacondarwin: Is that true? JS RPC does allow methods and callback to be called I believe.

RamIdeas: This isn't JS RPC – it's the miniflare magic proxy

petebacondarwin: But isn't it using JS RPC under the hood? Perhaps I am misunderstanding.

dario-piotrowicz: @petebacondarwin no this is unrelated to JS RPC 🙂, it is using the magic proxy and saving such object on the heap (in workerd) so that they can be referenced later.

JS RPC is implemented in the same sort of way: https://github.com/cloudflare/workers-sdk/blob/66bdad08834b403100d1e4d6cd507978cc50eaba/packages/miniflare/src/workers/core/proxy.worker.ts#L55-L57
but one is not necessary related to the other

dario-piotrowicz: PS: I've updated the changeset as I can see that `serialize` here is not really the right word and can lead to confusion 🙂 
(my changes are not actually serializing such objects but allowing us to handle them)

petebacondarwin (approved):

.changeset/wet-taxis-share.md
RamIdeas: Ah yeah sorry I didn't see any implementation changes so assumed you were simply allowing the object to be _serialised_ whereas before it was prevented from _proxying_.

Am I right to now think that proxying objects with functions _always_ worked? If so, do we know why we explicitly prevented it?

.changeset/wet-taxis-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": patch
+---
+
+fix: make the magic proxy able to proxy objects containing functions
```
RamIdeas, line 5:
```suggestion
fix: allow the magic proxy to proxy objects containing functions

This was previously prevented but this change removes that restriction.
```

RamIdeas:
This PR needs rebasing ("Required Github CI checks have changed for workers-sdk. Please rebase on main to get the required jobs to run")

.changeset/wet-taxis-share.md
dario-piotrowicz: > Am I right to now think that proxying objects with functions always worked?

No, they never worked, only the rpc stubs ones but that's because we're specifically handling them ad-hoc

>  If so, do we know why we explicitly prevented it?

The thing is that with my changes we check if an object contains functions and if so we mark as a non-plain object.
Basically my changes make the `!isPlainObject(object)` trigger here:
https://github.com/cloudflare/workers-sdk/blob/c0b057b9d196dc46f54530f91e8c87dee88d6cfa/packages/miniflare/src/workers/core/proxy.worker.ts#L88-L93
making it so that the object doesn't get sent to devalue which would otherwise complain when trying to serialize it (saying that it doesn't know how to serialize functions).

PR #5674, opened 2024-04-21
Implement Early Hints for `modulepreload`

itsmatteomanf:

This PR adds support for `<link rel="modulepreload" href="...">` entries in the HTML, to allow them to be added to the Link header.

- Tests
  - [X] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [X] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [X] Not necessary because it's an expansion of the current feature set, which is already expected. Some more docs might be nice in general, but not related to this specific addition.


petebacondarwin (approved):

petebacondarwin (changes requested):
Actually there are tests for this... 

https://github.com/cloudflare/workers-sdk/blob/f514934d6be8ea3b6777435f35e236c5ad0528e2/packages/pages-shared/__tests__/asset-server/handler.test.ts#L520

itsmatteomanf:
@petebacondarwin I have merged the changes on main as there was a conflict, tell me if I red it wrong, but there is now a cache control header on the early hints cache, which makes sense, so merged it in here.

itsmatteomanf:
> Actually there are tests for this...
> 
> https://github.com/cloudflare/workers-sdk/blob/f514934d6be8ea3b6777435f35e236c5ad0528e2/packages/pages-shared/__tests__/asset-server/handler.test.ts#L520

Does it make sense to test for the specific module preload part? The preconnect one isn't checked either...

petebacondarwin:
I realised that we were not checking the linting on pages-shared, so I have just pushed an extra commit that cleans that up.
Meanwhile I think it is valuable to add a test for this new feature. It looks like it won't be too difficult. The fact that other features are not tested is not really a good argument for not testing new features.
I'm happy to add that test now, if you don't mind.

itsmatteomanf:
> I'm happy to add that test now, if you don't mind.

Might make sense to just add the `preconnect` test, too? It's not hard to do and would cover all cases, should I or should you?

petebacondarwin:
> > I'm happy to add that test now, if you don't mind.
> 
> Might make sense to just add the `preconnect` test, too? It's not hard to do and would cover all cases, should I or should you?

If you have time, please do. Thanks.

petebacondarwin (approved):

petebacondarwin:
Added a small fixup commit because some snapshot tests needed updating. But otherwise looks good. We just need a review from a pages team member and then we are good to merge.

WalshyDev (approved):

WalshyDev:
I'll get this released along with https://github.com/cloudflare/workers-sdk/pull/5875 hopefully tomorrow or Thursday @itsmatteomanf 
Appreciate it!

ngbrown:
This added Early Hints for `rel="modulepreload"` in the HTML head. Can you do the same for `rel=modulepreload` in the `Link` header? Currently, it only works when the `Link` header uses `rel=preload`.

itsmatteomanf:
@ngbrown can't now, you are free to open a PR yourself :)

PR #5675, opened 2024-04-21
C3: Improve bindings DX in hono template

jculvey:

The `hono` template has been updated as follows:
- Bumps `create-hono` to `0.7.0`
- Automatically installs dependencies and specifies the detected package manager to avoid interactive prompts
- Adds a `wrangler.toml` file with commented out examples of all available bindings to match other templates.
- Adds a `cf-typegen` package script to automatically regenerate types for `Bindings` from `wrangler.toml`

Example of resulting project with an added kv example: https://github.com/jculvey/cf-hono-example. 

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: existing coverage sufficient
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Improvements are backwards compatible


packages/create-cloudflare/templates/hono/snippets/appDeclaration.ts
```diff
@@ -0,0 +1 @@
+new Hono<{ Bindings: Bindings }>()
```
dario-piotrowicz, line 1:
```suggestion
new Hono<{ Bindings: Bindings }>()

```

packages/create-cloudflare/templates/hono/snippets/bindingsType.ts
```diff
@@ -0,0 +1 @@
+type Bindings = Record<string, unknown> & CloudflareBindings;
```
dario-piotrowicz, line 1:
```suggestion
type Bindings = Record<string, unknown> & CloudflareBindings;

```

packages/create-cloudflare/templates/hono/c3.ts
```diff
@@ -1,24 +1,77 @@
...
+			if (n.node.id.type === "Identifier" && n.node.id.name === "app") {
+				n.node.init = snippets
+					.appDeclarationTs[0] as recast.types.namedTypes.NewExpression;
+
+				return false;
+			}
+		},
+	});
+
+	s.stop(`${brandColor("updated")} \`${indexFile}\``);
```
dario-piotrowicz, line 55:
```suggestion
	s.stop(`${brandColor("updated")} \`${dim(indexFile)}\``);
```

Usually the updated file name is dimmed right?
Screenshot 2024-04-22 at 10 07 28

(only in Qwik I see an exception to this and I'd imagine we'd want to amend that too? 🤔)

packages/create-cloudflare/templates/hono/snippets/bindingsType.ts
dario-piotrowicz: Also I think hono in its template index.ts does not use semicolons:
```suggestion
type Bindings = Record<string, unknown> & CloudflareBindings
```

Yours being the only one present:
Screenshot 2024-04-22 at 10 12 00

dario-piotrowicz (approved):
Looks good to me (with a few minor suggestions) 🙂 

Also if you're adding the toml file support, I'm thinking that it might be appropriate to add an e2e for the bindings access as you did for the other frameworks?
https://github.com/cloudflare/workers-sdk/blob/1f2c1e382ea309e8a06e48c2d6a77b4b24515cf0/packages/create-cloudflare/e2e-tests/frameworks.test.ts#L76-L79

yusukebe (approved):
Looks good to me! Thanks!

Cherry:
@jculvey Looking at your generated repo at ..., everything on `env` is now `unknown`, so for example:

```ts
c.env.ddd
```
in your repo doesn't throw any errors, it just reports as `unknown`. Vs with a more traditional bindings definition, you end up with a useful error that tells you it doesn't exist on `env`:

```
Property 'ddd' does not exist on type 'CloudflareBindings'.ts(2339)
```

Is this the intent here? I feel like the DX is much worsened when everything on `env` is `unknown`.

jculvey:
@Cherry  thanks for catching this!

In short, no, the intent here wasn't just to regress the type safety of env 😅 .

In short we're trying to simplify the process of consuming the type definition generated by `wrangler types` for `hono` to bring it in line with the rest of the C3 templates. 

Unfortunately If you try to assign the result of that command directly to `Bindings` you end up with this error:

image

As you point out, if the generated `CloudflareBindings` was a `type` and not an `interface`, there wouldn't be an issue.

What I was trying to accomplish was to augment the existing type with the generated one:
image

Agreed this is worse DX. I'll make a PR to patch this up. Thanks again for the report!

Cherry:
Thanks @jculvey! 🧡

PR #5687, opened 2024-04-23
C3: Fix issue with hono Bindings type

jculvey:

This fixes an issue introduced with https://github.com/cloudflare/workers-sdk/pull/5675 where the `Bindings` type ends up being too permissive (essentially `Record<string, unknown>`).

Open to suggestions on a better implementation of the new type.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Backwards compatible fix

### DIFF
--- a/packages/create-cloudflare/templates/hono/snippets/bindingsType.ts
+++ b/packages/create-cloudflare/templates/hono/snippets/bindingsType.ts
@@ -1,1 +1,3 @@
-type Bindings = Record<string, unknown> & CloudflareBindings
+type Bindings = {
+  [key in keyof CloudflareBindings]: CloudflareBindings[key]
+}
--- a/.changeset/shaggy-pugs-switch.md
+++ b/.changeset/shaggy-pugs-switch.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+fix: Fixes an issue with the `Bindings` type genereated by C3 being too permissive.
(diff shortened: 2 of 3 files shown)


yusukebe:
Hi @jculvey 

As mentioned by also @Cherry, it should not allow `Record<string, unknown>`. I think your way of defining `Bindings` is best, though anyone can write more shortly. Thanks!

packages/create-cloudflare/templates/hono/snippets/bindingsType.ts
```diff
@@ -1 +1,3 @@
-type Bindings = Record<string, unknown> & CloudflareBindings
+type Bindings = {
+  [key in keyof CloudflareBindings]: CloudflareBindings[key]
+}
```
dario-piotrowicz, line 3:
```suggestion
}

```

dario-piotrowicz (approved):

packages/create-cloudflare/templates/hono/snippets/bindingsType.ts
jculvey: I've really gotta setup some tooling to fix this 🤦

PR #5689, opened 2024-04-23
Warn if unexpected key is in the toml

### LINKED ISSUE #5681: 🚀 Feature Request: Warn if unexpected key is in the toml
### Describe the solution

On more than one occasion I have been on a call with a user confused at why something from their wrangler toml is not applying correctly. 

They have been appending some config, often recommended by some docs, or a guide. To their `wrangler.toml` and are confused that the resource is not created.

```toml
[[kv_namespaces]]
binding = "KV"
id = "<NAMESPACE_ID>"

tail_consumer = [{ ... }]
```

On `wrangler deploy` it ignores the additional key in the kv_namespaces array and deploys the worker

It would be better if when running `wrangler deploy` the extension raised warnings whilst deploying.

```bash
Warning: Unexpected Key: tail_consumer is not a valid key for kv_namespaces.
```

The deployment should still succeed to avoid this being a breaking change.

Ankcorn:

Fixes #5681

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because:


fixtures/messy-config/wrangler.toml
```diff
@@ -0,0 +1,13 @@
...
+compatibility_date = "2024-04-19"
+compatibility_flags = ["nodejs_compat"]
+
+
+[[d1_databases]]
+binding = "MY_DB"
+database_name = "my-database"
+database_id = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
+
+tail_consumers = [{service = "<TAIL_WORKER_NAME>"}]
```
Ankcorn, line 13: This will now add a warning as the toml is invalid

packages/cli/package.json
```diff
@@ -24,7 +24,7 @@
 	"devDependencies": {
 		"@clack/core": "^0.3.2",
 		"@clack/prompts": "^0.6.3",
-		"@cloudflare/eslint-config-worker": "*",
+		"@cloudflare/eslint-config-worker": "workspace:*",
```
Ankcorn, line 27: I ran into some issues with pnpm trying to resolve these from npm.

Does anyone have any idea what I can do here? I can remove these if its not desired

packages/wrangler/src/config/validation-helpers.ts
```diff
@@ -589,6 +589,16 @@ const isRecord = (
...
+export const warnForUnnecessaryProperies = (diagnostic: Diagnostics, field: string, value: Object, keys: string[]) => {
+	for (const prop of Object.keys(value)) {
+		if(!keys.includes(prop)) {
+			diagnostic.warnings.push(
+				`"${field}" has unexpected key '${prop}'. Allowed keys are${keys.map(key => ` '${key}'`)}.`
+			)
+		}
+	}
+}
+
```
petebacondarwin, line 601: Did you see this helper?

https://github.com/cloudflare/workers-sdk/blob/9aa155d43a3ad36ab620efa236a9ba2357bbde5a/packages/wrangler/src/config/validation-helpers.ts#L516-L534

Is it possible to use that instead?

fixtures/messy-config/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "messy-config",
+  "version": "0.0.0",
+  "private": true,
+  "scripts": {
+    "deploy": "wrangler deploy",
+    "dev": "wrangler dev",
+    "start": "wrangler dev",
+    "test": "vitest"
```
petebacondarwin, line 9: Thanks for creating this fixture. As it stands I don't think this is running any tests in the CI. We could do with a vitest.config.js and then an actual vitest test that runs Wrangler to trigger the config warnings.

packages/cli/package.json
petebacondarwin: I'm not sure why you needed to do these changes. They have not been needed previously. Is there something funny in your local setup perhaps?
Arguably putting workspace makes these dependencies tighter so they will only work if the package matches within the workspace. But given that in both cases we have `*` then both should match anything in the workspace.

petebacondarwin (changes requested):
Thanks for improving the experience for Wrangler users. This looks like a great improvement. A couple of suggestions if you have time...

packages/cli/package.json
Ankcorn: There is definitely something funny in my local setup, totally happy to put these back.

I couldn't get it to resolve them without it. There are some places that already had `workspace:*`

There shouldn't be anything too funny with my setup. I'm using volta to install the node.js version required in the package.json and the latest pnpm.

packages/wrangler/src/config/validation-helpers.ts
Ankcorn: Didn't spot it! Thanks for pointing it out :)

fixtures/messy-config/package.json
Ankcorn: I'm not sure a full e2e test is warranted for this change. I realise I may have misunderstood the intent behind the fixtures when creating this. I treated it as more of an examples folder where I could put some code.

If you think it's best I'm happy to add one but looking through the existing tests in the repo a unit test [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/__tests__/deploy.test.ts) checking that the output contains the warnings would give me reasonable confidence in this change.

Thoughts?

packages/wrangler/src/config/validation-helpers.ts
```diff
@@ -589,6 +589,7 @@ const isRecord = (
 ): value is Record<string | number | symbol, unknown> =>
 	typeof value === "object" && value !== null && !Array.isArray(value);
 
+
```
petebacondarwin, line 592: I suspect this extra line is breaking the formatting checks.

petebacondarwin (approved):
Cool. Thanks for updating. Try running:

```
pnpm run fix
```

to sort out the formatting issues


The other failures are most likely flakes

Ankcorn:
Thanks @petebacondarwin 

I have ran those formatting checks.

Hopefully the CI flakes in my favour this time 😃

Ankcorn:
The other thing that is worth flagging is I used the json schema and a script to generate the lists of optional fields so I didn't have to write them by hand. 

There is a chance that a missing field in my configuration could lead to unexpected warnings in customers wrangler tomls.

I could think of a good way to verify this but here is the script I used. 

```javascript
const schema = require('./config-schema.json')

const environment = schema.definitions.Environment.properties

function generateCode(properties) {
	return `validateAdditionalProperties(diagnostics, field, Object.keys(value), [${
		properties.reduce((sum, el) => `${sum}\n\t"${el}",`, '')
	}
]);

	`
}
for(const service of schema.definitions.Environment.required) {
	console.log(service, '\n')
	const serviceConfig = environment[service].properties || environment[service].items

	if(serviceConfig && !serviceConfig.additionalProperties && serviceConfig.properties) {

		// Generate code
		console.log(generateCode(Object.keys(serviceConfig.properties)))
	} else if(serviceConfig) {
		for(const configuration of Object.values(serviceConfig)) {
			if(configuration.items?.additionalProperties === false) {
				console.log(service, serviceConfig)
				console.log(generateCode(Object.keys(configuration.items.properties)))
			}
		}
	}
}
```

PR #5696, opened 2024-04-24
D1: execute --file --remote now uses dedicated import API

geelen:

Fixes CFSQL-794:

This hits a new API endpoint 'import', that initially returns a signed R2 PUT url for uploading SQL, then returns polling info as the import completes. This completely bypasses the existing split-chunk-and-execute implementation that often left your DB in an invalid state

D1 server-side code is up and running so this should be fully functional for testing & feedback.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: no user-facing changes


packages/wrangler/src/d1/constants.ts
```diff
@@ -2,4 +2,3 @@ export const DEFAULT_MIGRATION_PATH = "./migrations";
 export const DEFAULT_MIGRATION_TABLE = "d1_migrations";
 export const LOCATION_CHOICES = ["weur", "eeur", "apac", "oc", "wnam", "enam"];
 // Max number of statements to send in a single /execute call
```
rozenmd, line 4: stray comment

packages/wrangler/src/d1/execute.tsx
```diff
@@ -399,23 +524,19 @@ function logResult(r: QueryResult | QueryResult[]) {
...
-	return batches;
-}
-
 function shorten(query: string | undefined, length: number) {
 	return query && query.length > length
 		? query.slice(0, length) + "..."
 		: query;
 }
+
+async function checkForSQLiteBinary(filename: string) {
```
rozenmd, line 533: Reckon we can test this function?

packages/wrangler/src/d1/execute.tsx
```diff
@@ -348,41 +346,168 @@ async function executeRemotely({
...
+	if (etag !== etagResponse.replace(/^"|"$/g, "")) {
+		throw new UserError(
+			`File contents did not upload successfully. Please retry.`
+		);
 	}
-	return results;
+	logger.log(`🌀 Uploading complete.`);
+
+	return await d1ApiPost<
+		ImportPollingResponse | { success: false; error: string }
```
rozenmd, line 460: we repeat `{ success: false; error: string }` a few times, could be worth having its own type

packages/wrangler/src/d1/execute.tsx
```diff
@@ -315,30 +320,23 @@ async function executeRemotely({
 	config,
 	name,
 	shouldPrompt,
-	batches,
-	json,
```
rozenmd, line 319: we still want `json` mode though

geelen: Fixed!

image

packages/wrangler/package.json
```diff
@@ -172,6 +172,7 @@
 		"jest": "^29.7.0",
 		"jest-fetch-mock": "^3.0.3",
 		"jest-websocket-mock": "^2.5.0",
+		"md5-file": "^5.0.0",
```
penalosa, line 175: Do we need a dependency for this?

packages/wrangler/src/d1/execute.tsx
```diff
@@ -81,7 +88,7 @@ export function Options(yargs: CommonYargsArgv) {
 		.option("batch-size", {
 			describe: "Number of queries to send in a single batch",
 			type: "number",
-			default: DEFAULT_BATCH_SIZE,
+			deprecated: true,
```
penalosa, line 91:
```suggestion
			deprecated: true,
			hidden: true
```

packages/wrangler/src/d1/execute.tsx
```diff
@@ -315,30 +320,23 @@ async function executeRemotely({
...
-	json: boolean | undefined;
+	input: ExecuteInput;
 	preview: boolean | undefined;
 }) {
-	const multiple_batches = batches.length > 1;
-	// in JSON mode, we don't want a prompt here
-	if (multiple_batches && !json) {
-		const warning = `⚠️  Too much SQL to send at once, this execution will be sent as ${batches.length} batches.`;
+	if (input.file) {
+		const warning = `⚠️ This process may take some time, during which your D1 will be unavailable to serve queries.`;
```
penalosa, line 333:
```suggestion
		const warning = `⚠️ This process may take some time, during which your D1 database will be unavailable to serve queries.`;
```

packages/wrangler/src/d1/execute.tsx
```diff
@@ -315,30 +320,23 @@ async function executeRemotely({
...
 		if (shouldPrompt) {
-			const ok = await confirm(
-				`${warning}\nℹ️  Each batch is sent individually and may leave your DB in an unexpected state if a later batch fails.\n⚠️  Make sure you have a recent backup. Ok to proceed?`
-			);
+			const ok = await confirm(`${warning}\n  Ok to proceed?`);
 			if (!ok) return null;
-			logger.log(`🌀 Let's go`);
 		} else {
-			logger.error(warning);
+			logger.log(warning);
```
penalosa, line 339:
```suggestion
			logger.warn(warning);
```

packages/wrangler/src/d1/execute.tsx
```diff
@@ -348,41 +346,194 @@ async function executeRemotely({
...
-		);
-		logger.error(error.message);
-		throw error;
+	if (preview) {
+		if (!db.previewDatabaseUuid) {
+			const error = new UserError(
+				"Please define a `preview_database_id` in your wrangler.toml to execute your queries against a preview database"
+			);
+			logger.error(error.message);
+			throw error;
```
penalosa, line 355:
```suggestion
			throw new UserError(
				"Please define a `preview_database_id` in your wrangler.toml to execute your queries against a preview database"
			);
```

packages/wrangler/src/d1/execute.tsx
```diff
@@ -348,41 +346,194 @@ async function executeRemotely({
...
+		`/accounts/${accountId}/d1/database/${db.uuid}/${action}`,
+		{
+			method: "POST",
+			headers: {
+				"Content-Type": "application/json",
+				...(db.internal_env ? { "x-d1-internal-env": db.internal_env } : {}),
+			},
+			body: JSON.stringify(body),
+		}
+	);
```
penalosa, line 536: Is this needed? We already have utility methods for calling the Cloudflare API

packages/wrangler/src/d1/execute.tsx
```diff
@@ -348,41 +346,194 @@ async function executeRemotely({
...
+
+	return await d1ApiPost<ImportPollingResponse | PollingFailure>(
+		accountId,
+		db,
+		"import",
+		{ action: "ingest", filename, etag }
+	);
+}
+
+async function pollUntilComplete(
```
penalosa, line 488: It would be great if this could display a spinner UI like Cloudchamber

packages/wrangler/src/d1/execute.tsx
```diff
@@ -348,41 +346,194 @@ async function executeRemotely({
...
+					etag,
+					initResponse
+			  )
+			: initResponse;
+
+		const finalResponse = await pollUntilComplete(
+			firstPollResponse,
+			accountId,
+			db
+		);
```
penalosa, line 401: Can we split this two functions out a bit more? Right now the upload function also requests the first polling response—it would be nice to have a function that _just_ did upload, and then handoff to a function that _just_ did polling?

packages/wrangler/src/d1/execute.tsx
geelen: Isn't that what `fetchResult` is? I just extracted this so that all staging DB calls sent the right header, but if there's another internal api I should use for doing the fetch then lmk

packages/wrangler/src/d1/execute.tsx
geelen: Sadly the API conflates the two a bit in the interests of performance. The "upload" endpoint does actually return a "polling response", which can also be "complete" (pretty common for things like migrations). The server side actually adds a small delay (1s currently) to increase the chance of things happening in a single round-trip.

You only need to "pollUntilComplete" if the first response isn't "complete", so I'll add a quick comment to clarify. But this was the best way I could figure out to capture the _actual_ behaviour of the API, rather than the logical order of operations...

packages/wrangler/package.json
geelen: We're expecting 5-10GB files so I had a look around for something that had a good implementation for larger files. Would be pretty easy to vendor the particular code path we use, if you'd prefer. For now I'll just pin to the exact version at least.

packages/wrangler/src/d1/migrations/apply.tsx
```diff
@@ -37,7 +33,7 @@ export function ApplyOptions(yargs: CommonYargsArgv) {
 	return MigrationOptions(yargs).option("batch-size", {
 		describe: "Number of queries to send in a single batch",
 		type: "number",
-		default: DEFAULT_BATCH_SIZE,
+		deprecated: true,
```
RamIdeas, line 36: Even if this is being deprecated, I think we should keep the default value or else it'd be a breaking change in behaviour

RamIdeas: Unless this option isn't being used at all anymore, in which case it's a breaking change no matter what?

packages/wrangler/src/d1/execute.tsx
RamIdeas: There's a util called `spinnerWhile` which you can wrap you async stuff in pretty easily to get the spinner

packages/wrangler/src/d1/migrations/apply.tsx
geelen: The old version couldn't handle too much SQL in one call, so chunked it up using this `batch-size`. The new one does it all in one go, so there's no batching.

So yes, it's no longer used at all, but I don't think that constitutes it being a breaking change. The commands still do exactly what they did previously, just the `batch-size` option has no effect any more, so I deprecated it.

packages/wrangler/src/d1/execute.tsx
geelen: Thanks for that. I've added it in a couple of places, but for the main import it conflicts with the log lines coming down from the polling endpoint itself, so I'm leaving it off there:

image

rozenmd (approved):

.changeset/fluffy-files-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Improved `d1 execute --file --remote` performance & added support for much larger SQL files within a single transaction.
```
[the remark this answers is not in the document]
RamIdeas:
```suggestion
feature: Improved `d1 execute --file --remote` performance & added support for much larger SQL files within a single transaction.
```

RamIdeas (approved):

PR #5703, opened 2024-04-24
Check if middleware is an array

### LINKED ISSUE #5420: 🐛 BUG: wrangler dev throws if finding a non-iterable middleware export

Wrangler core

3.39.0 [Wrangler]

20.10.0

Mac Ventura 13

## Observed Behavior
`wrangler dev` throws if it finds an export named "middleware", but cannot iterate it.

## Expected Behavior
If requesting a hidden exported property (`middleware`) and iterating over it, there should at least be checks to **verify if the item is iterable** before attempting to iterate.

_TLDR; It should not throw._

## Error Reproduction
Create a simple file with the following content, then launch it using `wrangler dev`:

```ts
export default {
  middleware: () => {}, // wrangler dev attempts to iterate this, throwing
}
```

The same script runs fine in production/deployed but cannot be run locally.

## Why this matters
`itty-router` routers have a Proxy trap on the prototype chain.  If we export our router directly, a request from the runtime to "middleware" returns a function (via that trap).  This throws when wrangler tries to iterate it without checking to see if it's iterable/an array.

```ts
import { Router } from 'itty-router'

export default Router() // this doesn't work locally

export default { ...Router() } // but this does, as the prototype chain is stripped
```

_No response_

image

## Fun Fact
We used a `Proxy` to identify that the property "middleware" was being requested from the export (the logs showed nothing).

penalosa:

Use the named export `__INTERNAL_WRANGLER_MIDDLEWARE__` for listing the middleware to be injected in module-format workers. To inject middleware for testing, use `__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__`.

This removes support for adding a `middleware` array to your exported handler.

Fixes #5420.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not a documented feature

### DIFF
--- a/packages/wrangler/src/deployment-bundle/apply-middleware.ts
+++ b/packages/wrangler/src/deployment-bundle/apply-middleware.ts
@@ -62,12 +62,12 @@ export async function applyMiddlewareLoaderFacade(
 			dedent/*javascript*/ `
 				import worker, * as OTHER_EXPORTS from "${prepareFilePath(entry.file)}";
 				${imports}
-				
+
 				worker.middleware = [
 					${middlewareFns.join(",")},
-					...(worker.middleware ?? []),
+					...(Array.isArray(worker.middleware) ? worker.middleware : []),
 				].filter(Boolean);
-				
+
 				export * from "${prepareFilePath(entry.file)}";
 				export default worker;
 			`
--- a/packages/wrangler/src/deployment-bundle/apply-middleware.ts
+++ b/packages/wrangler/src/deployment-bundle/apply-middleware.ts
@@ -54,7 +54,9 @@ export async function applyMiddlewareLoaderFacade(
 		)
 		.join("\n");
 
-	const middlewareFns = middlewareIdentifiers.map(([m]) => `${m}.default`);
+	const middlewareFns = middlewareIdentifiers
+		.map(([m]) => `${m}.default`)
+		.join(",");
 
 	if (entry.format === "modules") {
 		await fs.promises.writeFile(
@@ -63,12 +65,9 @@ export async function applyMiddlewareLoaderFacade(
 				import worker, * as OTHER_EXPORTS from "${prepareFilePath(entry.file)}";
 				${imports}
 
-				worker.middleware = [
-					${middlewareFns.join(",")},
-					...(Array.isArray(worker.middleware) ? worker.middleware : []),
-				].filter(Boolean);
-
 				export * from "${prepareFilePath(entry.file)}";
+
+				export const __INTERNAL_WRANGLER_MIDDLEWARE__ = [${middlewareFns}]
 				export default worker;
 			`
 		);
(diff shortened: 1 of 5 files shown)


petebacondarwin (commented):
So this means that it won't crash for itty-router anymore.
But do we actually want to ignore middleware that has been proxied such as is happening there?

penalosa:
I believe we do, yes. As far as I know we've never documented this middleware behaviour, and I think we implemented it for testing purposes.

petebacondarwin:
I don't think it is just for testing... we add these middlewares to handle specific `wrangler dev` scenarios such as the `__scheduled` path handling. See https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/templates/middleware

If the itty-router is just wrapping the array containing these middlewares in a proxy that is then ignored by our bundler then these middlewares are no longer applied and so this functionality would be unavailable to them.

petebacondarwin:
Shouldn't the fix be for the itty-router to pass through this middleware property untouched by the Proxy that they create?

penalosa:
Sorry, that wasn't super clear—we do definitely use middleware, but I think this way of defining middleware (a user specified array in the exported handler object) was only for testing?

petebacondarwin (commented):
As discussed offline. I wonder if we should avoid reading the `middleware` property from the default export at all? Perhaps move it to a non-default export?
I believe we could do this if this is only an internal API...

kwhitley:
> Shouldn't the fix be for the itty-router to pass through this middleware property untouched by the Proxy that they create?

We don't actually include/expose a middleware property at all.  Our router is an object with a Proxy prototype.  Any attributes on the object are accessed normally, and unknown attributes are intercepted by the proxy, interpreting them as route handlers.

So a accessing `router.middleware` returns a function like you might use here:
```ts
router.middleware('/path', () => 'this will match the MIDDLEWARE HTTP verb...')
```

As a result, something this will throw a similar error to what we're seeing here, as it DOES find router.middleware, but it's a function... not an iterable array:
```ts
for (const handler of router.middleware) {
  // do something
}
```

kwhitley:
Any update on this?

We'd love to remove our disclaimer for using itty with the very environment it was designed for! :)

penalosa:
> As discussed offline. I wonder if we should avoid reading the `middleware` property from the default export at all? Perhaps move it to a non-default export? I believe we could do this if this is only an internal API...

This should be ready for a re-review now

packages/wrangler/src/deployment-bundle/apply-middleware.ts
```diff
@@ -54,21 +54,23 @@ export async function applyMiddlewareLoaderFacade(
...
-					${middlewareFns.join(",")},
-					...(worker.middleware ?? []),
-				].filter(Boolean);
-				
+
 				export * from "${prepareFilePath(entry.file)}";
+
+				export const __INTERNAL_WRANGLER_MIDDLEWARE__ = [
+					...(OTHER_EXPORTS.__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__ ?? []),
+					${middlewareFns}
```
petebacondarwin, line 72: NIT: I would prefer to put the `join(",")` here rather than in the variable declaration above.

petebacondarwin (approved):
Love it!

PR #5705, opened 2024-04-24
Add workers ai staging flag

G4brym:

This adds support for `staging` flag for Ai binding, this binding only works in specific workers.
For normal users trying to use it, the flag will be ignored.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: internal tool
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: internal tool


packages/wrangler/src/deployment-bundle/worker.ts
```diff
@@ -111,6 +111,7 @@ export interface CfBrowserBinding {
 
 export interface CfAIBinding {
 	binding: string;
+	staging?: boolean;
```
RamIdeas, line 114: Do we want to mark this with a JSDoc `@internal` attribute or some other way to clarify to external users that they shouldn't use this?

RamIdeas: We already have `first_party_worker`, `logfwdr` and `unsafe` bindings so no need to block this PR now. We can follow up later if this is applicable.

RamIdeas (approved):

Issue #5614, opened 2024-04-14
🐛 BUG: `vercel/ai` streaming broken with `3.50.0` in dev | unexpected `gzip` compression


flexchar:
### Which Cloudflare product(s) does this pertain to?

Workers Runtime

### What version(s) of the tool(s) are you using?

3.50.0

### What version of Node are you using?

v21.7.1

### What operating system and version are you using?

macOS 14.4.1

### Describe the Bug

`vercel/ai` library provides easy to implement streaming for OpenAI. After upgrade to latest version the streaming is not output until it's completely done buffering. This is in development. I cannot deploy as that would break the clients. Once I can, I will test.

I am also using HonoJS however it did not change. I am not using any compression middlewares etc.

### Observed behavior
The streaming is buffered until AI provider is finished.

### Expected behavior
Should not be buffered.

### Steps to reproduce

Reproducible https://github.com/flexchar/workers-sdk-issue-5614

### Please provide any relevant error logs

I noticed that the broken response includes `Content-Encoding: gzip` header. Compressing would explain why buffering is not working. 

As such, appending header to disable compression alleviates the issue:
```ts
return new StreamingTextResponse(stream, {
    headers: {
        // Disable any compression to prevent buffering
        'Content-Encoding': 'identity',
    },
});
```

Reference https://sdk.vercel.ai/docs/api-reference/streaming-text-response#example

Might be related:
- https://github.com/cloudflare/cloudflare-docs/issues/12060
- https://github.com/cloudflare/workers-sdk/pull/5713

RamIdeas:
Hi @flexchar! Can you please provide a minimal reproduction of this bug where running with `npx wrangler@3.50.0 dev` would not stream the response and where `npx wrangler@3.49.0 dev` will stream the response?

zwily:
I'm am seeing the same thing with my AI-based app. on 3.49.0, I get responses streamed back (with no `Content-Encoding: gzip`) header, and with 3.50, `Content-Encoding: gzip` is added and the entire response is buffered and sent at once.

Unfortunately my code is not public and getting all the pieces in places for a minimal reproduction would take quite awhile so I'm going to pin to 3.49.0 for now.

zwily:
A couple notes... If I add a `Content-Encoding` on my response from the worker, like `Content-Encoding: identity`, then it works. (That's not a valid value for Content-Encoding, but it skips the logic that adds the gzipping.)

[...]

PR #5409, opened 2024-03-27
[miniflare] feature: respect incoming `Accept-Encoding` header

mrbbot:
## What this PR solves / how to test

Fixes #5246 (CUSTESC-39344).

This PR updates Miniflare to respect the incoming `Accept-Encoding` header, and set the correct value of `Accept-Encoding`/`request.cf.clientAcceptEncoding` inside the Worker. This is to match the production/documented behaviour here: https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#accept-encoding.

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): <https://github.com/cloudflare/cloudflare-docs/pull/>...
  - [x] Not necessary because: matching already documented behaviour

PR #5713, opened 2024-04-26
Set `Accept-Encoding: identity` for AI API requests

mhart:
Streaming doesn't currently work with `wrangler dev` if the responses are gzipped. This may be due to workerd not being able to stream gzipped bodies – still under investigation. But in the meantime we can explicitly opt out of gzip responses by setting `Accept-Encoding: identity` when making requests to the API.

Due to miniflare/workerd's behaviour with gzipped bodies, you'll need to disable gzip encoding from your Worker/Pages output too – unless you can guarantee that your client (eg, curl) won't request with `accept-encoding: gzip`. Unfortunately most browsers will. Here's an example of how to do that:

```
    return new Response(stream, {
      headers: {
        'content-encoding': 'none',
        'content-type': 'text/event-stream',
      },
    });
```

Fixes #APIOPS-8486.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/ai/fetcher.ts
+++ b/packages/wrangler/src/ai/fetcher.ts
@@ -19,6 +19,11 @@ export async function AIFetcher(request: Request): Promise<Response> {
 	request.headers.delete("Host");
 	request.headers.delete("Content-Length");
 
+	// Streaming doesn't currently work with `wrangler dev` if the responses are gzipped.
+	// We can explicitly opt out of gzip responses by setting
+	// `Accept-Encoding: identity` when making requests to the API.
+	request.headers.set("Accept-Encoding", "identity");
+
 	const res = await performApiFetch(`/accounts/${accountId}/ai/run/proxy`, {
 		method: "POST",
 		headers: Object.fromEntries(request.headers.entries()),


irvinebroque:
We must be able to stream gzipped bodies from workerd, right? Test it out just with workerd to validate?

penalosa:
I'm not sure this is the right solution here: this feels like a bug we should fix, rather than workaround?

penalosa:
I _think_ this is a bug in `workerd`. The following worker fails to stream when run directly through workerd locally:

```js
export default {
	async fetch(request, env, ctx) {
		return new Response(
			ReadableStream.from(
				(async function* () {
					for (let i = 0; i < 512; i++) {
						yield "01234567890abcdef".repeat(16) + "\n";
						await scheduler.wait(100);
					}
				})()
			).pipeThrough(new TextEncoderStream()),
      {headers: {
        "Content-Encoding":"gzip"
      }}
		);
	},
};
```

mhart:
> I _think_ this is a bug in `workerd`. The following worker fails to stream when run directly through workerd locally:

Yeah it is. There's just an unknown ETA on a fix for that. This is a workaround in the meantime – and only for AI bindings in local dev.

penalosa:
@mhart This feels potentially more serious to me—if `workerd` doesn't support streaming for gzipped bodies then I think we're in a situation where Wrangler/Miniflare will effectively _never_ stream responses in local dev. I wonder if we need to disable gzip/brotli entirely when running locally?

mhart:
> @mhart This feels potentially more serious to me—if `workerd` doesn't support streaming for gzipped bodies then I think we're in a situation where Wrangler/Miniflare will effectively _never_ stream responses in local dev. I wonder if we need to disable gzip/brotli entirely when running locally?

That's exactly right – that's the behaviour I've observed – eg via service bindings or similar

penalosa:
We made a change here recently (https://github.com/cloudflare/workers-sdk/pull/5409) which started adding `Content-Encoding` headers to user worker responses, based on their `Accept-Encoding` header. I think we should probably partially revert that, which will allow use worker responses to be streamed.

mhart:
Btw, should I change the example to use `globalThis.MINIFLARE`? Is that still the recommended way to test if you're in the local dev server? ie:

```js
    return new Response(stream, {
      headers: {
        ...(globalThis.MINIFLARE ? { 'content-encoding': 'none' } : {}),
        'content-type': 'text/event-stream',
      },
    });
```

frandiox:
The added `content-encoding` is also problematic when proxying Miniflare from Node because you can't use `mf.dispatchFetch` anymore: https://github.com/cloudflare/workers-sdk/pull/5409#issuecomment-2057431634

flexchar:
This seems to be a potential fix for https://github.com/cloudflare/workers-sdk/issues/5614.

mhart:
Ok, @penalosa update to this: I actually think part of the problem is that miniflare is applying encoding in cases where cloudflare's production server doesn't. Eg, with `text/event-stream`. This isn't compressed in production (even if `accept-encoding` is sent) – whereas it is in `miniflare`. I suspect we're not taking into account the same mime types to determine when to apply encoding or not.

Fixing that would actually fix Workers AI streaming (I think!), even without workerd changes.

Still doesn't help other use cases (eg, HTML streaming – as `text/html` *is* encoded by default). But still, seems like the right thing to do.

RamIdeas:
This PR has been superseded by https://github.com/cloudflare/workers-sdk/pull/5798

PR #5732, opened 2024-05-01
feat: handle `?raw` suffix

### LINKED ISSUE #5731: 🚀 Feature Request: handle `?raw` suffix
### Describe the solution

Vite allows for [importing assets as strings](https://vitejs.dev/guide/assets#importing-asset-as-string). I have a monorepo and my main packages are using vite. Therefore, I always import static assets by using this technique. If I want to re-use those helper functions and import them in a worker, I get an error that no module loader is defined for that file extension (e.g. html).
It would be great if `?raw` suffix was supported.

hrueger:

Fixes #5731

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/cuddly-items-repair.md
+++ b/.changeset/cuddly-items-repair.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feat: handle `?raw` suffix for imports
--- a/packages/wrangler/src/deployment-bundle/find-additional-modules.ts
+++ b/packages/wrangler/src/deployment-bundle/find-additional-modules.ts
@@ -113,6 +113,18 @@ export async function findAdditionalModules(
 	return modules;
 }
 
+/**
+ * Remove a trailing `?raw` suffix from the file name
+ * @param filePath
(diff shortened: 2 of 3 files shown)


GregBrimble (changes requested):
Thanks for creating a PR!

If we were to add support for `?raw`, I think that would take the form of a adhoc [rule](https://developers.cloudflare.com/workers/wrangler/configuration/#:~:text=to%20triggers.-,rules,-Rule%20optional) so that we can ensure we handle binary files, for example. Additionally, it looks like a lot of the JS ecosystem is starting to converge on [import attributes](https://github.com/tc39/proposal-import-attributes) which, potentially, might be preferred over query params on the import identifier. If it's alright, I'd hold off and wait for some official product direction here before continuing with this PR. Thanks again for the PR and highlighting the need for something in this space though!

PR #5740, opened 2024-05-01
Add Version ID to all outputs in addition to the deprecated Deployment ID

WalshyDev:

Fixes WC-2110

We need the version ID in the output so users can migrate before we fully remove it.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Nothing about the specific output and terminology is already version ID

### DIFF
--- a/packages/wrangler/e2e/c3-integration.test.ts
+++ b/packages/wrangler/e2e/c3-integration.test.ts
@@ -68,7 +68,8 @@ describe("c3 integration", () => {
 			Published tmp-e2e-wrangler (TIMINGS)
 			  https://tmp-e2e-wrangler.SUBDOMAIN.workers.dev
 			Current Deployment ID: 00000000-0000-0000-0000-000000000000
-			NOTE: "Deployment ID" in this output will be changed to "Version ID" in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
+			Current Version ID: 00000000-0000-0000-0000-000000000000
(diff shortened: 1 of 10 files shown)


packages/wrangler/e2e/c3-integration.test.ts
```diff
@@ -68,6 +68,7 @@ describe("c3 integration", () => {
 			Published tmp-e2e-wrangler (TIMINGS)
 			  https://tmp-e2e-wrangler.SUBDOMAIN.workers.dev
 			Current Deployment ID: 00000000-0000-0000-0000-000000000000
+			Current Version ID: 00000000-0000-0000-0000-000000000000
 			NOTE: "Deployment ID" in this output will be changed to "Version ID" in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments"
```
tanushree-sharma, line 72: @WalshyDev can we also update the copy on the "NOTE". Sorry if the change wasn't super clear on the ticket. 

`Note: Deployment ID has been renamed to Version ID. Deployment ID is present to maintain compatibility with the previous behavior of this command. This output will change in a future version of Wrangler. To learn more visit: https://developers.cloudflare.com/workers/configuration/versions-and-deployments.`

RamIdeas (approved):

PR #5744, opened 2024-05-02
Load sourcemaps relative to entry directory

penalosa:

Load sourcemaps relative to the entry directory, not cwd.

Fixes CUSTESC-40775.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)


petebacondarwin (approved):
LGTM!

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -9716,6 +9716,29 @@ export default{
...
+				"some/dir/wrangler.toml"
+			);
+			writeWorkerSource({ format: "ts" });
+
+			mockSubDomainRequest();
+			mockUploadWorkerRequest({
+				expectedMainModule: "index.js",
+				expectedModules: {
+					"index.js.map": expect.stringContaining(
+						`"sources":["../../another.ts","../../index.ts"],"sourceRoot":""`
```
zebp, line 9735: Can this test explicitly specify the `file` field of this? It's critical that it matches the name of a file in the multipart but has `.map` otherwise the stacktrace remapper in the runtime can't correlate module names to sourcemaps.

zebp: (not that I think this PR would chance that field, but we should have been explicitly testing this in the first place)

zebp (approved):

PR #5750, opened 2024-05-03
skip header rules on asset handler error

aaronlisman:

Fixes CUSTESC-40911

- Tests
  - [X] Included
  - [X] Included
- Public documentation
  - [?] Not necessary because: bugfix, though I could see how it may be worth noting?

### DIFF
--- a/packages/pages-shared/__tests__/asset-server/handler.test.ts
+++ b/packages/pages-shared/__tests__/asset-server/handler.test.ts
@@ -783,20 +783,22 @@ describe("asset-server handler", () => {
 				rules: [
 					{
 						path: "/*",
-						headers: {"x-unwanted-header": "foo"},
+						headers: { "x-unwanted-header": "foo" },
 						unsetHeaders: [],
 					},
 				],
 			},
 			redirects: {
 				invalid: [],
-				rules: [{
-					from: "/here",
-					to: "/there",
-					status: 301,
-					lineNumber: 1,
-				}],
-			}
+				rules: [
+					{
+						from: "/here",
+						to: "/there",
+						status: 301,
+						lineNumber: 1,
+					},
+				],
+			},
 		}) as Metadata;
 
 		const findAssetEntryForPath = async (path: string) => {
@@ -815,10 +817,12 @@ describe("asset-server handler", () => {
 			});
 
 			expect(response.status).toBe(500);
-			expect(Object.fromEntries(response.headers)).not.toHaveProperty("x-unwanted-header");
+			expect(Object.fromEntries(response.headers)).not.toHaveProperty(
+				"x-unwanted-header"
+			);
 		});
 
-		test("404 skips headers", async () => {
+		test("404 doesn't skip headers", async () => {
 			const { response } = await getTestResponse({
 				request: "https://foo.com/404",
 				metadata,
@@ -826,10 +830,12 @@ describe("asset-server handler", () => {
 			});
 
 			expect(response.status).toBe(404);
-			expect(Object.fromEntries(response.headers)).not.toHaveProperty("x-unwanted-header");
+			expect(Object.fromEntries(response.headers)).toHaveProperty(
+				"x-unwanted-header"
+			);
 		});
 
-		test("301 still has headers", async () => {
+		test("301 doesn't skip headers", async () => {
 			const { response } = await getTestResponse({
 				request: "https://foo.com/here",
 				metadata,
@@ -837,7 +843,9 @@ describe("asset-server handler", () => {
 			});
 
 			expect(response.status).toBe(301);
-			expect(Object.fromEntries(response.headers)).toHaveProperty("x-unwanted-header");
+			expect(Object.fromEntries(response.headers)).toHaveProperty(
+				"x-unwanted-header"
+			);
 		});
 	});
 });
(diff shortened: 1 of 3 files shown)


packages/pages-shared/asset-server/handler.ts
```diff
@@ -446,7 +446,10 @@ export async function generateHandler<
 				};
 			}
 		);
-		const matches = headersMatcher({ request });
+
+		// If the response was an internal error from the asset handler, skip applying header rules.
+		// This avoids a case where we can cache 500s for a long time, depending on headers set by a user.
+		const matches = response.ok ? headersMatcher({ request }) : [];
```
aaronlisman, line 452: @WalshyDev is `response.ok` the proper check here? It's apparently false for 3XX responses.

WalshyDev: [response.ok](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) will only check for 2xx codes. We probably want a `response.status < 400 && response.status > 599` or similar.

aaronlisman: should be better: https://github.com/cloudflare/workers-sdk/compare/c2ebd1fa5cea326202e32df4728019a7ce705156..3cc829d7f12f65b7686e2f56e4b651969049691f

packages/pages-shared/asset-server/handler.ts
```diff
@@ -446,7 +446,11 @@ export async function generateHandler<
 				};
 			}
 		);
-		const matches = headersMatcher({ request });
+
+		// If the response was an internal error from the asset handler, skip applying header rules.
```
WalshyDev, line 450:
```suggestion
		// If the response was an internal error or 404 from the asset handler, skip applying header rules.
```

packages/pages-shared/asset-server/handler.ts
```diff
@@ -446,7 +446,11 @@ export async function generateHandler<
...
+		// This avoids a case where we can cache 500s for a long time, depending on headers set by a user.
```
WalshyDev, line 451:
```suggestion
		// This avoids a cases where we can cache undesirable files for a long time, depending on headers set by a user.
```

WalshyDev (approved):

packages/pages-shared/__tests__/asset-server/handler.test.ts
```diff
@@ -774,6 +774,36 @@ describe("asset-server handler", () => {
 			expect(isPreservationCacheResponseExpiring(res)).toBe(true);
 		});
 	});
+
+	test("internal asset error doesn't set headers", async () => {
```
WalshyDev, line 778: probably worth it to add a test for 404 too

GregBrimble (approved):
I'm a little concerned about omitting some headers even on 404s here. For example, some users might be hardening their site with security headers that will now be dropped on 404s. I had previously considered expanding the `_headers` and `_redirects` specs to support matching on request/response properties (incl. status) but never got that over the line. We maybe consider revisiting that idea instead here?

Edit: oops, didn't mean to approve.

GregBrimble (changes requested):
<blocking while we consider implications>

aaronlisman:
> I'm a little concerned about omitting some headers even on 404s here. For example, some users might be hardening their site with security headers that will now be dropped on 404s. I had previously considered expanding the `_headers` and `_redirects` specs to support matching on request/response properties (incl. status) but never got that over the line. We maybe consider revisiting that idea instead here?
> 
> Edit: oops, didn't mean to approve.

instead of matching on the status code, should we catch exceptions from `fetchAsset` and plumb through a flag that says this was definitely an error? that seems closer to the intent of the bug fix, which is 'something definitely went wrong while fetching the asset, and we don't want to apply user headers'.

or at least just match on 5XX statuses?

WalshyDev:
> I'm a little concerned about omitting some headers even on 404s here. For example, some users might be hardening their site with security headers that will now be dropped on 404s.

True but we also don't really want to potentially cache a 404 for hours/days. If I hit 404 a js file while the site is propagating, I could easily break the site for all users hitting this colo. 

We could also just remove cache-control and not the rest?

GregBrimble:
Yeah, totally understand why we want this — avoiding caching 404s. But we can't add this if it breaks something else (potentially more important to some people).

For example, let's say someone is attaching an `X-Frame-Options: DENY` with `_headers`, and the page the attacker is embedding 404s, that doesn't necessarily say that there is nothing of interest on the page. It might still contain buttons with actions which shouldn't be clickjackable.

5XX are definitely _less_ likely to be in this scenario since we'll control the error page, but I'm still a bit concerned about a full-stack site that potentially is then manipulating the asset page in some way, and where users are still reliant on controls they've placed in `_headers`.

WalshyDev:
Indeed, my comment didn't disagree and offered an alternative. Fully agree with you, just also think we should solve 404s here too (safely).

GregBrimble:
@aaronlisman and I chatted and decided that we're okay with not attaching headers on a Cloudflare-owned error page. Our thinking was that, if Workers went down, then we'd be serving a hard error from FL or whatever. @aaronlisman is going to change this PR to check for an actual error from asset serving, and we'll just leave it at that for the moment.

If we need to provide more control over conditionally including/excluding when these headers are applied, we can raise a feature request for that extension as new work and plan it as normal.

He's going to make that change here, and update docs to reflect that headers won't be served in that instance. I'm going to add docs to recommend that any security headers related to Functions code should be attached in that Functions code, rather than relying on pass through from the "upstream" assets with `_headers`.

GregBrimble (approved):
I tweaked it, so I kinda invalidate my own review here, but I'm no longer blocking :) @WalshyDev , want to take a quick second look, and then we'll get this merged?

WalshyDev (approved):
lgtm

Issue #5729, opened 2024-04-30
🐛 BUG: Regression with hanging scripts after reload with custom builds


Cherry:
### Which Cloudflare product(s) does this pertain to?

Wrangler core

### What version(s) of the tool(s) are you using?

3.53.0

### What version of Node are you using?

20.12.2

### What operating system and version are you using?

Windows 11

### Observed behavior

When using a simple custom build, if you save the file multiple times, wrangler's server appears to hang and no longer accept requests.

### Expected behavior

wrangler should continue to accept requests after the custom build runs.

### Steps to reproduce

- Clone https://github.com/Cherry/wrangler-custom-builds-hang-repro
- npm ci
- npm run dev
- Observe requests to `http://127.0.0.1:8787` working fine
- Make multiple changes to `src/index.ts`. Add a console.log, etc. but ensure the file is written to at least twice
- Observe requests are now just hanging

---

-  `npx wrangler@3.53.0 dev` - hangs with above steps
-  `npx wrangler@3.19.0 dev` - hangs with above steps
-  `npx wrangler@3.18.0 dev` - works without issue
-  `npx wrangler@2.21.1 dev` - works without issue

It appears this regression was introduced in 3.19.0, cc @RamIdeas.

### Please provide a link to a minimal reproduction

https://github.com/Cherry/wrangler-custom-builds-hang-repro

Cherry:
This was fixed in https://github.com/cloudflare/workers-sdk/pull/5754, so going to close. Thanks!

PR #5741, opened 2024-05-02
Don't trigger bundle starts with custom builds

penalosa:
## What this PR solves / how to test

Previously, running a custom build which didn't change the output file would cause Wrangler to hang, since we were relying on esbuild to run in order to update the proxy. However, esbuild will not re-run in watch mode if all that's changed about one of its input files is the modified time—it seems to require an actual content change. 

Fixes #5729 

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: manually tested
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): <https://github.com/cloudflare/cloudflare-docs/pull/>...
  - [x] Not necessary because: bugfix

PR #5754, opened 2024-05-03
fix: custom build noop causes esbuild to skip bundling

RamIdeas:
now, if esbuild doesn't start bundling within a timeout of the custom build finishing, we assume it has decided to skip bundling and continue with the previous bundle output

This is alternative fix from #5741 but the guarantee of requests using your latest code changes is maintained

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/sixty-pans-nail.md
+++ b/.changeset/sixty-pans-nail.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: when using custom builds, the `wrangler dev` proxy server was sometimes left in a paused state
+
+This could be observed as the browser loading indefinitely, after saving a source file (unchanged) when using custom builds. This is now fixed by ensuring the proxy server is unpaused after a short timeout period.
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -91,7 +91,7 @@ async function runDevSession(
 
 		in ${workerPath} {
 			exits {
-        $ ${WRANGLER} dev ${flags}
+		$ ${WRANGLER} dev ${flags}
 			}
 		}
 			`;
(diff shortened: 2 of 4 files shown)


RamIdeas:
@Cherry can you confirm this resolves #5729 please?

Cherry:
This doesn't appear to resolve the issue unfortunately, no. If I open a file and just re-save it without any changes a few times, requests hang again.

RamIdeas:
Ah ok I tried it myself and the problem isn't just that esbuild isn't rebundling, it's that miniflare therefore isn't reloading either.

I have updated this PR and verified that it fixes your reproduction. If you have time to verify it, please let us know, thanks.

Cherry:
This does appear to resolve it now, yes, thank you!

I did get the following error once or twice when I spammed saving:
```
✘ [ERROR] No matching export in "dist/index.mjs" for import "default"
[wrangler]
[wrangler]     .wrangler/tmp/bundle-JUf0Z6/middleware-insertion-facade.js:1:11:
[wrangler]       1 │         import worker, * as OTHER_EXPORTS from "E:\\GitHub\\xxx...
[wrangler]         ╵                ~~~~~~
```

It didn't appear to impact anything though.

RamIdeas:
Thanks for confirming!

My guess is that error is because, while you spammed save, the imported file was still being written to.

RamIdeas:
I've added an e2e regression test for this bug but the test passes with AND without the fix too – not entirely sure why since I can reliably reproduce the bug manually and this PR reliably fixes the bug.

This codepath will be rewritten in the next couple weeks (to avoid useEffect dependency nightmares like this) amongst other planned work. I am happy to merge this as-is.

petebacondarwin (approved):

PR #5767, opened 2024-05-06
Specify pnpm version 8 in package.json engines field

geelen:

Now that PNPM is up to `9.0.6`, the chances of a user having upgraded to 9 is non-zero. This will give them an actual error message they can follow up, rather than a bunch of cryptic ones:

image

See https://github.com/cloudflare/workers-sdk/pull/5766 for what's potentially required to update to pnpm 9.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] TODO (before merge)
  - [ ] Included
- Public documentation
  - [ ] TODO (before merge)


rozenmd (approved):

petebacondarwin (approved):
We can land this but the long term fix is to add support for v9 by fixing up the dependencies to use `workspace:*` rather than `*`.

PR #5780, opened 2024-05-08
chore: add inputs to miniflare turbo config to improve cache hits

petebacondarwin:

By specifying inputs to miniflare we can ensure that the build and test steps only run when they actually need to.
Previously a change to a test would also cause the main build step to run again, for example.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: build tooling configuration change
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: build tooling configuration change
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: build tooling configuration change


CarmenPopoviciu (approved):

packages/miniflare/turbo.json
```diff
@@ -3,10 +3,24 @@
...
+				"*.json"
+			],
+			"outputs": ["dist/**", "bootstrap.js"],
 			"env": [
 				"CI_OS"
 			]
+		},
+		"test": {
+			"inputs": [
+				"test/**"
```
Skye-31, line 22: Doesn't this need to include `src` too?

petebacondarwin: In the root turbo.json there is a relationship between test and build, which covers the src.
But I just realised that this should be "test:ci" and also it is probably better to make that dependency explicit here. So this should be more like:

```
		"test:ci": {
			"inputs": ["tests/**"],
			"dependsOn": ["build"]
		},
```

packages/miniflare/turbo.json
```diff
@@ -3,10 +3,23 @@
 	"extends": ["//"],
 	"pipeline": {
 		"build": {
-			"outputs": ["dist/**"],
+			"inputs": [
+				"src/**",
+				"scripts/**",
+				"types/**",
+				"worker-metafiles/**",
```
mrbbot, line 10: Hey! 👋 `worker-metafiles` should probably be classified as `outputs`, they get written in `scripts/build.mjs`:

https://github.com/cloudflare/workers-sdk/blob/f3c0fff3a9d4c217146b358be2e74c2004dd5a43/packages/miniflare/scripts/build.mjs#L122-L129

petebacondarwin: Oh! I see. Thanks @mrbbot - the directory itself is not gitignored but the files in it are!

PR #5784, opened 2024-05-08
Fix ai binding tests

G4brym:

The next workerd version, includes some changes in the ai binding, this pr fixes the tests that will start failing

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because:

### DIFF
--- a/fixtures/ai-app/tests/index.test.ts
+++ b/fixtures/ai-app/tests/index.test.ts
@@ -27,6 +27,7 @@ describe("'wrangler dev' correctly renders pages", () => {
 		expect((content as Record<string, object>).run).toEqual("function");
 	});
 
+	// TODO: unskip when https://github.com/cloudflare/workerd/pull/2095 is merged and released
 	it.skip("ai binding properties", async ({ expect }) => {
 		const response = await fetch(`http://${ip}:${port}/`);
 		const content = await response.json();
--- a/fixtures/ai-app/tests/index.test.ts
+++ b/fixtures/ai-app/tests/index.test.ts
@@ -20,18 +20,21 @@ describe("'wrangler dev' correctly renders pages", () => {
 		await stop?.();
 	});
 
-	it("ai binding is defined ", async ({ expect }) => {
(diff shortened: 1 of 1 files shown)


G4brym:
@Cherry a new workerd release would take a few days, and block other features
I think its better to remove the test that is triggering the build to fail currently

Cherry:
Whoops, it seems I missed the removal of `logs` when mentioning the breaking change that removing `getLogs` would have been in https://github.com/cloudflare/workerd/pull/2044.

Perhaps skipping the test here temporarily (`it.skip`) and then restoring it after `workerd` is patched to revert the breaking change would be a better approach, instead of removing the `binding` test entirely? It's a good thing this was caught here - and it would be good to catch other breaking changes like this in the future too instead of removing it entirely.

G4brym:
Agree, i've just openned a pr in workerd to re add it
https://github.com/cloudflare/workerd/pull/2095/files

fixtures/ai-app/tests/index.test.ts
```diff
@@ -20,18 +20,21 @@ describe("'wrangler dev' correctly renders pages", () => {
...
-				logs: [],
-				options: {},
-			},
-			fetch: "function",
-			run: "function",
+		expect((content as Record<string, object>).fetch).toEqual("function");
+		expect((content as Record<string, object>).run).toEqual("function");
+	});
+
+	it.skip("ai binding properties", async ({ expect }) => {
```
RamIdeas, line 30: Would be good to comment when this test can be unskipped

```suggestion
	// TODO: unskip when https://github.com/cloudflare/workerd/pull/2095 is merged and released
	it.skip("ai binding properties", async ({ expect }) => {
```

RamIdeas (approved):

PR #5792, opened 2024-05-09
Auto add e2e label + PR description reminder for e2e

RamIdeas:

Sometimes we review PRs that change snapshot values in unit-tests and e2e tests but we don't notice the e2e snapshots were updated.

This PR makes sure that the 'e2e' label is added to a PR if any files within wrangler's e2e/ directory are touched. If the e2e tests change, we undoubtedly want to run the job to make sure they still pass.

This PR also adds to the PR description a reminder for the author and the reviewers to check whether an e2e test run is required.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: github action workflow change
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: internal repo change
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: internal repo change

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,6 +8,10 @@ Fixes #[insert GH or internal issue number(s)].
   - [ ] TODO (before merge)
   - [ ] Included
   - [ ] Not necessary because:
+- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
+  - [ ] I don't know
+  - [ ] Required / Maybe required
+  - [ ] Not required because:
 - Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
   - [ ] TODO (before merge)
   - [ ] Included
--- a/.github/pr-add-e2e-label.config.yml
+++ b/.github/pr-add-e2e-label.config.yml
@@ -1,0 +1,5 @@
+# add 'e2e' label to any changes within the 'packages/wrangler/e2e' directory
+e2e:
+  - all:
+      - changed-files:
+          - any-glob-to-any-file: "packages/wrangler/e2e/**/*"
--- a/.github/workflows/pr-add-e2e-label.yml
+++ b/.github/workflows/pr-add-e2e-label.yml
@@ -1,0 +1,17 @@
+name: "Pull Request Labeler"
+on:
+  - pull_request_target
+
+jobs:
+  labeler:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/labeler@v5
+        with:
+          configuration-path: .github/pr-add-e2e-label.config.yml


.github/pull_request_template.md
```diff
@@ -8,6 +8,10 @@ Fixes #[insert GH or internal issue number(s)].
   - [ ] TODO (before merge)
   - [ ] Included
   - [ ] Not necessary because:
+- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
+  - [ ] I don't know
```
RamIdeas, line 12: This is useful for external contributors who don't know whether their change requires an e2e run and reminds us, as reviewers, to decide for ourselves.

.github/pull_request_template.md
```diff
@@ -8,6 +8,10 @@ Fixes #[insert GH or internal issue number(s)].
...
+  - [ ] Required / Maybe required
```
RamIdeas, line 13: This is useful for internal and external contributors.

For internal contributors, this will serve as a reminder to add the "e2e" label before the PR is merged. Especially when you don't want to run them while the PR is in-flight to save runners.

For external contributors, they won't be able to run the e2e tests themselves (due to security implications). This will serve as a reminder to add the "e2e" label and also for us, as maintainers, to run the e2e tests in a separate PR (only once we have reviewed the changes and confirmed no malicious intent)

.github/pull_request_template.md
```diff
@@ -8,6 +8,10 @@ Fixes #[insert GH or internal issue number(s)].
...
+  - [ ] Not required because:
```
RamIdeas, line 14: This is useful to confirm there is no need for the e2e test run. And to make sure they weren't run only because we forgot to add the label.

.github/workflows/e2e.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - changeset-release/main
+    paths:
+      - packages/wrangler/e2e/**
```
petebacondarwin, line 8: Is this enough? It feels like this actually restricts the job to run only on pushes that match both the branch and the path?

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore:~:text=the%20path%20again.-,If%20you%20define%20both%20branches/branches%2Dignore%20and%20paths/paths%2Dignore%2C%20the%20workflow%20will%20only%20run%20when%20both%20filters%20are%20satisfied.,-The%20paths%20and

RamIdeas: Good catch! Any idea how to make it run if either match?

Trying to make "push:" an array so I can specify the same event twice here https://github.com/cloudflare/workers-sdk/pull/5793/commits/0daaaf008ca205bbf36b37cac03a3e60c35b8f71

RamIdeas: I've switched strategy. Rather than mess with the existing job, I've added a job to automatically add the 'e2e' label if files within the dir change (this feature comes out-of-the-box with actions/labeler)

petebacondarwin (approved):
Nice!

PR #5795, opened 2024-05-09
DRAFT chore: match dependency versions across the repo

threepointone:
This isn't ready for review yet, but putting it up for CI to have a look.

---

tl;dr: makes it so we never use different versions of a package anywhere in the repo. No functional changes.

- Installs @manypkg/cli (https://github.com/Thinkmill/manypkg)
- Run `manypkg check` which returns a list of all the mismatched dependencies
- Manually fix each dependency failure
- Run `manypkg fix` to fix any extra failures (like unordered dependencies in `package.json`s)
- Run `pnpm check`, `pnpm test`, `pnpm prettify` and fix any failures from them
- Added `check:repo` npm task that runs `manypkg check`, and added it to `npm run check` so it runs in CI

--- 

Notes: 
- The most confusing part: when running tests, wrangler/../deploy.tsts.ts and wrangler/../pages/deploy.test.ts fail (on macos and ubuntu respectively). Of note, it appears like `runInTmpDir()` fails to apply itself, so the tests run in the source directory, and fail expectations. Been fighting this for a week and I can't figure out why or how, yet. 
- These packages hanf when running test:ci on my machine (with node@16.17.1): 
  - additional-modules 
  - interactive-dev-tests
  - pages-d1-shim
  - pages-workerjs-directory
  - type-generation-fixture
  - worker-app
- Running tests appears to leave orphaned workerd process 
- I added `/templates/*` to workspaces and gave it the same treatment, but removed it from workspaces. We'll revisit that folder in the future, I don't want to slow down the builds atm.


packages/miniflare/types/streams.d.ts
```diff
@@ -11,7 +11,7 @@ declare module "stream/web" {
 	}
 
 	interface ReadableStreamBYOBReader {
-		readonly closed: Promise<void>;
+		readonly closed: Promise<undefined>;
```
threepointone, line 14: updating typescript made this a little more strict (I think), so I had to fix this here

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -508,6 +508,7 @@ function buildProjectMiniflareOptions(
 		return {
 			...SHARED_MINIFLARE_OPTIONS,
 			unsafeModuleFallbackService: moduleFallbackService,
+			// @ts-expect-error SUNILFIX
```
threepointone, line 511: this is failing typecheck, will fix before landing

packages/miniflare/scripts/build.mjs
```diff
@@ -186,8 +187,10 @@ async function buildPackage() {
 			"esbuild",
 		],
 		plugins: [embedWorkersPlugin],
-		logLevel: watch ? "info" : "warning",
-		watch,
+		logLevel: "warning",
```
threepointone, line 190: disabled watch mode here because we updated esbuild and it didn't have the same api. I'll being it back later.

packages/wrangler/src/__tests__/pages/project-upload.test.ts
```diff
@@ -479,22 +480,3 @@ describe("pages project upload", () => {
...
- *
- * Note that we don't always get a progress message for every item uploaded.
- * These upload counts are not deterministic and can change from test run to test run.
- *
- * Also if you run the tests in --runInBand mode then we never see any of
- * these progress messages in the tests!!
- *
- * So this helper removes these message from the snapshots to keep them consistent.
- */
-export function normalizeProgressSteps(str: string): string {
```
threepointone, line 498: exporting this function and reusing it elsewhere was causing this suite to be run twice, so I extracted it into its own helper.

packages/wrangler/src/__tests__/pages/secret.test.ts
```diff
@@ -22,16 +25,6 @@ function createFetchResult(result: unknown, success = true) {
 	};
 }
 
-export function mockGetMemberships(
```
threepointone, line 25: just like nomralize-progress-steps, importing this function was causing this suite to be run twice. we already had this helper in /helpers/mock-oauth-flow, so I replaced it with that

packages/wrangler/src/__tests__/secret.test.ts
```diff
@@ -27,16 +30,6 @@ function createFetchResult(result: unknown, success = true) {
 	};
 }
 
-export function mockGetMemberships(
```
threepointone, line 30: we should write a lint rule that prevents exports from *.test.ts/*.spec.ts files

PR #5798, opened 2024-05-09
miniflare: compress responses more like Cloudflare FL

RamIdeas:

TODO (collecting some links first)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: fix


RamIdeas:
@mhart @frandiox can you confirm whether or not this fixes your issues please?

packages/miniflare/test/index.spec.ts
```diff
@@ -303,9 +307,13 @@ test("Miniflare: custom service using Content-Encoding header", async (t) => {
...
 	const test = async (encoding: string) => {
 		const res = await mf.dispatchFetch("http://localhost", {
-			headers: { "X-Test-Encoding": encoding },
+			headers: {
+				"Accept-Encoding": encoding,
+				"X-Test-Encoding": encoding,
+			},
 		});
 		t.is(await res.text(), testBody, encoding);
+		t.is(res.headers.get('Content-Encoding'), encoding, encoding);
```
RamIdeas, line 316: I noticed this was test wasn't actually testing anything – the encoding was always "gzip". I added this assertion and it was failing – I had to add the "Accept-Encoding" header above so we are now actually testing each encoding

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -223,6 +224,12 @@ function ensureAcceptableEncoding(
 	if (encodings.length === 0) return response;
 
 	const contentEncoding = response.headers.get("Content-Encoding");
+	const contentType = response.headers.get("Content-Type");
+
+    // if cloudflare's FL does not compress this mime-type, then don't compress locally either
+    if (!isCompressedByCloudflareFL(contentType)) {
+        return response;
+    }
```
RamIdeas, line 232: This is the main fix, in miniflare's entry worker

packages/miniflare/test/index.spec.ts
```diff
@@ -445,6 +491,18 @@ test("Miniflare: negotiates acceptable encoding", async (t) => {
...
+	res = await fetch(defaultCompressedUrl);
+	t.is(res.headers.get("Content-Type"), "text/html");
+	t.is(res.headers.get("Content-Encoding"), 'gzip');
+	t.is(await res.text(), testBody);
+
+	// Check `Content-Type: text/event-stream` is not compressed (FL does not compress this mime type)
+	res = await fetch(defaultUncompressedUrl);
+	t.is(res.headers.get("Content-Type"), "text/event-stream");
+	t.is(res.headers.get("Content-Encoding"), null);
+	t.is(await res.text(), testBody);
```
RamIdeas, line 505: These were the tests confirmed to fail without the fix, passing with the fix

mhart:
@RamIdeas It works! At least, in this fairly common case of Workers AI use:

```js
const body = await env.AI.run('@cf/meta/llama-2-7b-chat-int8', {
  prompt: 'Say hi',
  stream: true,
});
return new Response(body, { headers: { 'content-type': 'text/event-stream' } });
```

Tried with some other content types like `text/plain` and `text/html` and they do get encoded – which means they also don't stream (and will continue not to until some workerd changes happen).

So, users will still see issues in use cases like streaming HTML rendering – but I suspect there's nothing more for miniflare to do here. 🚀

frandiox:
> can you confirm whether or not this fixes your issues please?

No luck. A simple `text/html` response would still go through [this code path](https://github.com/cloudflare/workers-sdk/blob/e425172e0c0f124da226ee3bdb7ba332bdc8d30d/packages/miniflare/src/workers/core/entry.worker.ts#L276) and set the content-encoding. 

image


`mf.dispatchFetch(...)` returns a decompressed response with content-encoding `gzip`, so we can't forward it to the browser as is. Since undici's fetch always decompresses the response, perhaps `mf.dispatchFetch` should remove the content-encoding header?

From the [original issue](https://github.com/cloudflare/workers-sdk/pull/5409#issuecomment-2057431634):

>Running mf.dispatchFetch(...) (or plain fetch(workerUrl)) is decompressing the response but keeping Content-Encoding: gzip or similar in the response headers. Then, once we forward this response from Node (e.g. Vite server) to the browser, the browser expects encoded content but it is not encoded anymore.

>Unfortunately, I think there's no way to avoid decompressing the response in undici. The workaround is to remove Content-Encoding from the Miniflare response or recompressing it in Node -- or perhaps use Node's http.get instead of fetch to get the response... which is not intuitive with mf.dispatchFetch().

Not sure if `mf.dispatchFetch` is supposed to work in Node proxies. Maybe not and this is not an issue. Perhaps I just need to use undici's `request` directly to communicate with the worker instead of `mf.dispatchFetch`, although it feels counter intuitive.

packages/miniflare/src/shared/mime-types.ts
```diff
@@ -0,0 +1,59 @@
...
+    "application/manifest+json",
+    "application/ld+json",
+    "application/graphql+json",
+    "application/geo+json",
+]);
+
+export function isCompressedByCloudflareFL(contentTypeHeader: string | undefined | null) {
+    if(!contentTypeHeader) return true; // Content-Type inferred as text/plain
+
+    const [contentType] = contentTypeHeader?.split(';');
```
penalosa, line 56: Why the `?.` given the check above?

packages/miniflare/src/shared/mime-types.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    return compressedByCloudflareFL.has(contentType);
+}
```
penalosa, line 59:
```suggestion
}

```

penalosa (approved):

packages/miniflare/src/shared/mime-types.ts
RamIdeas: Yeah saw this last night. Will remove.

I originally had: `const [contentType] = contentTypeHeader?.split(';') ?? ['text/plain'];`

RamIdeas:
> Not sure if `mf.dispatchFetch` is supposed to work in Node proxies. Maybe not and this is not an issue. Perhaps I just need to use undici's `request` directly to communicate with the worker instead of `mf.dispatchFetch`, although it feels counter intuitive.

Ah I thought my change to `_transformsForContentEncodingAndContentType` handled this case but seems the test uses miniflare in front of a node server, not the other way around.

Since you posted your original comment, I've thought the response to be malformed. If the body is decompressed, the Content-Encoding header shouldn't imply it is compressed. So I agree we should remove the header. I have done in a following commit. Would you mind trying again?

This actually means you will be proxying an uncompressed response. Maybe your proxy will apply its own compression. But that seems wasteful so, in fact, I think miniflare should use undici.request not undici.fetch but that seemed like a larger change (the options don't match up) so will leave it uncompressed for now.

penalosa (approved):
Approved pending confirmation

RamIdeas:
Thanks for confirming. Will merge now and it'll go out in tomorrow's release

PR #5809, opened 2024-05-13
Update pnpm etc

petebacondarwin:

Updates the monorepo to use npnm 9.1.1 both locally and in GH actions.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: GH Action configuration change
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: GH Action configuration change
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: GH Action configuration change


.github/actions/install-dependencies/action.yml
```diff
@@ -1,19 +1,33 @@
...
 
-    - name: Install Node.js ${{ env.node-version }}
+    - name: Install Node.js ${{ inputs.node-version }}
       uses: actions/setup-node@v3
       with:
-        node-version: ${{ env.node-version }}
+        node-version: ${{ inputs.node-version }}
         cache: "pnpm"
 
+    - name: "Create .env file"
```
penalosa, line 21: Could we rename this as we have in other actions (I think it was to something like "Configure turbo repo cache environment variables")

packages/wrangler/src/deployment-bundle/module-collection.ts
```diff
@@ -334,7 +334,7 @@ export function createModuleCollector(props: {
 
 								// Finally, load the file and hash it
 								// If we didn't do any smart resolution above, this will attempt to load as an absolute path
-								const fileContent = await readFile(filePath);
+								const fileContent = await readFile(stripQuery(filePath));
```
penalosa, line 337: Is this intended to be part of this PR?

.github/actions/install-dependencies/action.yml
petebacondarwin: Absolutely!!

packages/wrangler/src/deployment-bundle/module-collection.ts
petebacondarwin: Nope!

petebacondarwin:
```
ERROR: This version of pnpm requires at least Node.js v18.12
The current version of Node.js is v16.18.1
```

😢 

So fixing our node.js dependency must come first...

petebacondarwin:
Node.js 18 is in, so this PR is now ready to go!

penalosa (approved):

PR #5813, opened 2024-05-13
Add gateway failure retries for Pages project uploads

GregBrimble:

Attempts to mitigate a number of customer escalations we've seen recently because of a seemingly increased failure rate of KV.

- Reduces the maximum bucket size from 50 MiB to 40 MiB.
- Reduces the maximum asset count from 5000 to 2000.
- Allows for more retries (with increased sleep between attempts) when encountering an API gateway failure.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: just internal stuff


packages/wrangler/src/parse.ts
```diff
@@ -71,6 +71,20 @@ export class ParseError extends UserError implements Message {
...
+	#status?: number;
+	constructor({ status, ...rest }: Message & { status?: number }) {
+		super(rest);
+		this.name = this.constructor.name;
+		this.#status = status;
+	}
+
+	isGatewayError() {
+		if (this.#status !== undefined) {
+			return [524].includes(this.#status);
```
CarmenPopoviciu, line 83: out of curiosity .. why not `this.#status === 524`?

pnpm-lock.yaml
```diff
@@ -945,7 +945,7 @@ importers:
         version: 8.49.0
       eslint-config-turbo:
         specifier: latest
-        version: 1.13.2(eslint@8.49.0)
```
CarmenPopoviciu, line 948: are these changes commited on purpose? dbl checking...

packages/wrangler/src/parse.ts
GregBrimble: Figure it's likely that we'd want to capture additional status codes in the future, so just dumped that pattern here while I was there. Premature? Yes, absolutely.

pnpm-lock.yaml
GregBrimble: nope! I'll remove, thanks for catching :)

CarmenPopoviciu (approved):

.changeset/forty-insects-lick.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": patch
```
CarmenPopoviciu, line 2: we sure this is a `patch` not `minor`. I'm always a bit confused, but these changes feel more like a minor to me?

GregBrimble: I'd call this a bug fix personally: absolutely no functional change and the intent is to fix flakey failures that users see.

CarmenPopoviciu: isn't the fact that we're adjusting the retry logic in case of API gateway failures a functional change? Or the fact that we're reducing the asset count?

CarmenPopoviciu: not pushing for either, mainly thinking out loud...and reconsidering all my changeset `minor` life choices 😅

GregBrimble: I'd personally say it's not a functional change as far as the user is concerned, no. But yeah, don't personally care at all and happy to be told I'm wrong!

CarmenPopoviciu: fair enough. 

> don't personally care at all and happy to be told I'm wrong
lol, same here

I'd just go ahead and merge as is

PR #5820, opened 2024-05-14
Node 18

petebacondarwin:

Updates the monorepo to support node 18.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: internal changes
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: internal changes

### DIFF
--- a/.github/workflows/c3-dependabot-versioning-prs.yml
+++ b/.github/workflows/c3-dependabot-versioning-prs.yml
@@ -19,10 +19,10 @@ jobs:
       - uses: pnpm/action-setup@v2
         with:
           version: 8.8.0
-      - name: Use Node.js 16.18
+      - name: Use Node.js 18
         uses: actions/setup-node@v3
         with:
-          node-version: 16.18
+          node-version: 18.20.2
           cache: "pnpm"
       - name: Install NPM Dependencies
         run: pnpm install --frozen-lockfile
--- a/packages/wrangler/src/__tests__/cfetch-utils.test.ts
+++ b/packages/wrangler/src/__tests__/cfetch-utils.test.ts
@@ -1,6 +1,7 @@
 import { rest } from "msw";
 import { hasMorePages } from "../cfetch";
 import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
+import { mockConsoleMethods } from "./helpers/mock-console";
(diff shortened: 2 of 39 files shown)


.github/workflows/c3-dependabot-versioning-prs.yml
```diff
@@ -19,10 +19,10 @@ jobs:
       - uses: pnpm/action-setup@v2
         with:
           version: 8.8.0
-      - name: Use Node.js 16.18
+      - name: Use Node.js 18
         uses: actions/setup-node@v3
         with:
-          node-version: 16.18
+          node-version: 18.20.2
```
RamIdeas, line 25: can we pin this to latest v18.x.x rather than the specific v18.20.2?

petebacondarwin: I'm not keen on doing that because that means that running a build at two different times might have different results. I am also not worried about keeping on the leading edge given that we have spent the last year stuck on v16 😄

petebacondarwin: In the pnpm PR that will follow this, I move all the dependency installations into a single common action so we only have to change the version in one place.

RamIdeas (approved):

packages/wrangler/src/__tests__/cfetch-utils.test.ts
```diff
@@ -1,6 +1,7 @@
 import { rest } from "msw";
 import { hasMorePages } from "../cfetch";
 import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
+import { mockConsoleMethods } from "./helpers/mock-console";
 import { createFetchResult, msw } from "./helpers/msw";
 import { runInTempDir } from "./helpers/run-in-tmp";
```
RamIdeas, line 6: The cleanup for this util is failing on windows again (resource lock) – I thought we solved that? (maybe by just wrapping in try..catch but that seemed fine)

RamIdeas: I've rerun the job anyway in case it was just a flake

PR #5840, opened 2024-05-15
Fix the worker-speedtest template (#5830)

### LINKED ISSUE #5830: 🐛 BUG: worker-speedtest template can't be deployed

Workers for Platforms

 ⛅️ wrangler 3.56.0

v20.10.0

Fedora 39

### Observed behavior
Running `npm create cloudflare@latest my-app -- --template cloudflare/workers-sdk/templates/worker-speedtest` as described in the [Quickstarts](https://developers.cloudflare.com/workers/get-started/quickstarts) results in a project that can't be deployed.

### Expected behavior
I expect the code generated by the template to be in a state where it can be deployed.

### Steps to reproduce
Please provide the following:
Run `npm create cloudflare@latest my-app -- --template cloudflare/workers-sdk/templates/worker-speedtest` and as it to deploy the worker.

The reason is that `packages.json` wants to deploy `src/index.js`, while the template contains `src/index.ts`. It seems this was refactored at some point from .js to .ts without updating `packages.json`?

After fixing this and other references to `.js` files in the code, `npm run deploy` worked and deployed the worker.

However searching through the source for other occurrences of `.js`, showed that the test code also still referred to javascript files, so `npm run test` failed, and simply changing that to `.ts` still left some errors.

So I think that makes this a bit more than a trivial change, which is why I'm opening this issue instead of opening a PR directly.
I'll send a PR to fix both the deploy issue and the unit tests later today or tomorrow.

_No response_

_No response_

nielsreijers:
The code had been converted to typescript in 026d944d, but packages.json and several other places still refered to .js files.

In addition a refactoring of the test code meant tests for both /up and /down urls were now being sent to the handler for /up, causing the tests to fail. I removed some of the test in up.test.ts since they were testing different values of the ?bytes= parameter, but this parameter is only used in down.ts, not in up.ts.

Fixes #5830.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: small change to an example template, not touching main production code
- Public documentation
  - [ ] TODO (before merge)


petebacondarwin (approved):

nielsreijers:
Thanks for the approve @petebacondarwin !
Any idea what's causing the failed test? It doesn't seem related to my change, does it?

PR #5869, opened 2024-05-20
Implement vitest pool e2e meta tests

petebacondarwin:
Commit 2 from #5508 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: test refactor
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: test refactor

### DIFF
--- a/packages/vitest-pool-workers/package.json
+++ b/packages/vitest-pool-workers/package.json
@@ -47,6 +47,8 @@
 		"dev": "node scripts/bundle.mjs watch",
 		"check:lint": "eslint .",
 		"check:type": "tsc && tsc -p src/worker/tsconfig.json && tsc -p types/tsconfig.json",
+		"test": "vitest run",
+		"test:ci": "vitest run",
 		"capnp:rtti": "capnpc -o ts scripts/rtti/rtti.capnp"
 	},
 	"devDependencies": {
@@ -57,6 +59,7 @@
 		"@types/semver": "^7.5.1",
 		"capnp-ts": "^0.7.0",
 		"capnpc-ts": "^0.7.0",
+		"ts-dedent": "^2.2.0",
 		"undici": "5.28.4"
 	},
 	"dependencies": {
--- a/packages/vitest-pool-workers/test/global-setup.ts
+++ b/packages/vitest-pool-workers/test/global-setup.ts
@@ -1,0 +1,88 @@
+import assert from "node:assert";
+import childProcess from "node:child_process";
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import type { GlobalSetupContext } from "vitest/node";
+
+function packPackage(packDestinationPath: string, packagePath: string) {
+	const output = childProcess.execSync(
+		`pnpm pack --pack-destination ${packDestinationPath}`,
+		{ cwd: packagePath, encoding: "utf8" }
+	);
+	const tarballPath = output.split("\n").find((line) => line.endsWith(".tgz"));
+	assert(
+		tarballPath !== undefined && path.isAbsolute(tarballPath),
+		`Expected absolute tarball path in ${JSON.stringify(output)}`
+	);
+	return tarballPath;
+}
+
+// Using a global setup means we can modify tests without having to re-install
+// packages into our temporary directory
+export default async function ({ provide }: GlobalSetupContext) {
+	console.log("Installing packages to temporary directory...");
+
+	// Create temporary directory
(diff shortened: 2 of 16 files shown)


packages/vitest-pool-workers/test/global-setup.ts
penalosa, line 1: Not blocking, but we should probably move this to use the `pnpm deploy` approach that you implemented for the other e2e tests. Maybe we want an `e2e-tests` package that runs all these tests?

packages/vitest-pool-workers/test/helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+			},
+		},
+	}
+});
+`;
+
+export function waitFor<T>(callback: Parameters<typeof vi.waitFor<T>>[0]) {
+	// The default timeout of `vi.waitFor()` is only 1s, which is a little
+	// short for some of these tests, especially on Windows.
+	return vi.waitFor(callback, { timeout: 10_000, interval: 500 });
```
penalosa, line 33: TIL! This would be pretty useful in lots of places

packages/vitest-pool-workers/test/helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+}
+
+async function seed(root: string, files: Record<string, string>) {
+	for (const [name, contents] of Object.entries(files)) {
+		const filePath = path.resolve(root, name);
+		await fs.mkdir(path.dirname(filePath), { recursive: true });
+		await fs.writeFile(filePath, contents);
+	}
+}
```
penalosa, line 42: Not blocking, but there are already three `seed()`/`seedFs()` functions in various places in workers-sdk

packages/vitest-pool-workers/test/helpers.ts
```diff
@@ -0,0 +1,158 @@
...
+	}) => Promise<Process>;
+	vitestDev: (options?: { flags?: string[]; maxBuffer?: number }) => Process;
+}>({
+	// Fixture for creating a temporary directory
+	async tmpPath({}, use) {
+		const tmpPoolInstallationPath = inject("tmpPoolInstallationPath");
+		const tmpPathBase = path.join(tmpPoolInstallationPath, "test-");
+		const tmpPath = await fs.mkdtemp(tmpPathBase);
+		await use(tmpPath);
+		await fs.rm(tmpPath, { recursive: true, maxRetries: 10 });
```
penalosa, line 96: This type of cleanup occasionally fails on Windows in CI with `EBUSY`, so we may want to catch and ignore this erroring

penalosa (approved):

packages/vitest-pool-workers/test/helpers.ts
petebacondarwin: New to me too...

packages/vitest-pool-workers/test/helpers.ts
petebacondarwin: Right but it is not so easy to share across packages. We might look into a shared package that is not deployed (similar to pages-shared).

packages/vitest-pool-workers/test/helpers.ts
petebacondarwin: Interestingly I haven't seen any flakes of this in Windows testing. (And I was running them a lot today!)
So I suspect that using `maxRetries` is a good solution here.

packages/vitest-pool-workers/test/global-setup.ts
petebacondarwin: I tried doing that a few days ago and I struggled to get it to work.
I think the problem is that we are not installing the package as a global (or accessing the binary directly like we do in the e2e tests). Instead we are creating a separate dummy package that has these as dependencies.

packages/vitest-pool-workers/package.json
```diff
@@ -47,6 +47,8 @@
 		"dev": "node scripts/bundle.mjs watch",
 		"check:lint": "eslint .",
 		"check:type": "tsc && tsc -p src/worker/tsconfig.json && tsc -p types/tsconfig.json",
+		"test": "vitest run",
```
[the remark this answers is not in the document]
petebacondarwin: Doh! Sorry I missed this. Will do in a follow up.

PR #5875, opened 2024-05-20
Don't cache 404s from the asset-server

WalshyDev:

Fixes WC-2233

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: no e2e tests touched
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: we already tell users not to cache, this just prevents a common footgun when caching

### DIFF
--- a/packages/pages-shared/__tests__/asset-server/handler.test.ts
+++ b/packages/pages-shared/__tests__/asset-server/handler.test.ts
@@ -640,6 +640,7 @@ describe("asset-server handler", () => {
 			expect(Object.fromEntries(response3.headers)).toMatchInlineSnapshot(`
 				{
 				  "access-control-allow-origin": "*",
+				  "cache-control": "no-store",
 				  "referrer-policy": "strict-origin-when-cross-origin",
 				}
 			`);
@@ -721,6 +722,7 @@ describe("asset-server handler", () => {
 			expect(Object.fromEntries(response2.headers)).toMatchInlineSnapshot(`
 				{
 				  "access-control-allow-origin": "*",
+				  "cache-control": "no-store",
 				  "referrer-policy": "strict-origin-when-cross-origin",
 				}
 			`);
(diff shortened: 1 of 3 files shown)


packages/pages-shared/asset-server/handler.ts
```diff
@@ -481,7 +481,19 @@ export async function generateHandler<
 	if (responseWithoutHeaders.status >= 500) {
 		return responseWithoutHeaders;
 	}
-	return await attachHeaders(responseWithoutHeaders);
+
+	const finalResponse = await attachHeaders(responseWithoutHeaders);
```
dario-piotrowicz, line 485: nit: I'd probably call this `responseWithHeaders` as it sounds a bit clearer then `finalResponse`, but that might be just me 😅 :+1:

packages/pages-shared/__tests__/asset-server/handler.test.ts
```diff
@@ -802,15 +804,17 @@ describe("asset-server handler", () => {
 		}) as Metadata;
 
 		const findAssetEntryForPath = async (path: string) => {
-			if (path.startsWith("/asset")) return "some-asset";
+			if (path.startsWith("/asset")) {
+				return "some-asset";
```
dario-piotrowicz, line 808: ❤️

dario-piotrowicz (approved):
Looks good to me 😃👍

petebacondarwin (approved):
Would it be better to add a very short cache such as a few minutes?

WalshyDev:
Interestingly enough not really, it's very very cheap for us to find a 404 so even for a definitely should be 404 asset, it's super quick and cheap.
For the cases though where it _shouldn't_ 404, a few mins is still quite a bit more impact than we'd want (definitely not as bad as a cache-control of a year but still).

PR #5220, opened 2024-03-11
feat: first-cut of using unenv to create a hybrid node.js compatibility setting

petebacondarwin:
## What this PR solves / how to test

Fixes #[insert GH or internal issue number(s)].

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): <https://github.com/cloudflare/cloudflare-docs/pull/>...
  - [ ] Not necessary because:

PR #5878, opened 2024-05-21
feat: second-cut of using unenv to create a hybrid node.js compatibility setting (via inject)

IgorMinar:
rebased version of https://github.com/cloudflare/workers-sdk/pull/5220/files which uses unenv's `inject` to introduce globals

- Tests
  - [ ] TODO (before merge)
  - [x] Included - the update to the fixture tests this
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: no e2e tests are affected by node-compat
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: still experimental but we should follow up with a docs ticket.


IgorMinar:
@petebacondarwin I rebased your PR but didn't want to force-push, so I'm creating a new PR. Let's chat tomorrow to discuss how we want to handle these PRs.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -0,0 +1,95 @@
...
+
+			// esbuild expects alias paths to be absolute
+			const aliasAbsolute = Object.fromEntries(
+				Object.entries(alias).map(([key, value]) => [
+					key,
+					require.resolve(value).replace(/\.cjs$/, ".mjs"),
+				])
+			);
+
+			build.onResolve(
```
IgorMinar, line 24: @petebacondarwin I don't understand what this resolve block is doing and why, could you please explain or suggest an inline comment?

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -0,0 +1,95 @@
...
+				};
+			});
+
+			build.onLoad({ filter: UNENV_GLOBALS_RE }, ({path}) => {
+				const globalName = path.match(UNENV_GLOBALS_RE)![1];
+				const [moduleName, exportName] = inject[globalName];
+
+				return {
+					contents: `
+						import { ${exportName} } from "${moduleName}";
```
IgorMinar, line 84: todo: add global and globalThis assignment...

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -0,0 +1,95 @@
...
+						// If we match an alias then return that and potentially mark it as external
+						return { path: result, external: external.includes(result) };
+					} else {
+						// Otherwise continue
+						return undefined;
+					}
+				}
+			);
+
+			// Requires of node-built-in-modules are converted to embed a file that does an ESM import
```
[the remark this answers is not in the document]
petebacondarwin: Workers do not support:

```
const { resolve } = require("node:path");
```

We only support 

```
import { resolve } from "node:path";
```

In the first case you get:

```

✘ [ERROR] service core:user:worker-app: Uncaught Error: Dynamic require of "node:path" is not supported
```

So what this hack does is replace the dynamic require of the node with a load from a synthetic virtual file that contains:

```
export * from 'node:path';
```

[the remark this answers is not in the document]
petebacondarwin: I stole this from @dario-piotrowicz which he added to next-on-pages: https://sourcegraph.com/npm/cloudflare/next-on-pages@58cdf844a05934ade288ee4966a23bff1e983179/-/blob/dist/index.js

It is also used in:

- SST Remix: https://github.com/sst/ion/blob/4e1f667da8a555104d28efddd3a3e203566289f7/pkg/platform/src/components/cloudflare/remix.ts
- Open Next: https://github.com/sst/open-next/blob/65f7a6032d0dba552084becae50da9cf22145e09/packages/open-next/src/plugins/edge.ts
- Astro: https://github.com/withastro/astro/blob/8e465d621e7acc4fd36d78bef63450c6550e0ec1/packages/integrations/vercel/src/serverless/middleware.ts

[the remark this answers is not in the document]
IgorMinar: Thanks for the explanation, Pete. I'll have to ponder this a bit, but I think it all makes sense.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -0,0 +1,175 @@
...
+						`_virtual_unenv_global_polyfill-${globalName}.js`
+					)
+				),
+			];
+
+			build.onResolve({ filter: UNENV_GLOBALS_RE }, ({ path }) => {
+				return {
+					path: path,
+				};
+			});
```
petebacondarwin, line 85:
```suggestion
			build.onResolve({ filter: UNENV_GLOBALS_RE }, ({ path }) => ({ path }));
```

petebacondarwin: Is the point of this resolve block to avoid esbuild trying to resolve the path itself, which it can't because these paths are virtual?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -386,12 +386,38 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 
 	const compatibilityFlags =
 		props.compatibilityFlags ?? config.compatibility_flags;
-	const nodejsCompat = compatibilityFlags.includes("nodejs_compat");
+
+	const nodejsCompatV2 = compatibilityFlags.includes(
```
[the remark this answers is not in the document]
IgorMinar: hahahahahaha... i think it's duplicated 5 or 6 times. I was really itching to refactor this code but then I realized that each location has a slightly different way to emit errors and some other needs. I fear that refactoring this is a much bigger job than it seems, and I didn't want to expand the scope of this PR any more. I suggest we merge this as is, and then have someone on the wrangler team with the right knowledge refactor this to dedupe.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
IgorMinar: yeah. I think I asked the question first a while back, and then figure it out and left a comment. or something along those lines.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -405,6 +431,12 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 		);
 	}
 
+	if (props.noBundle && nodejsCompatV2) {
+		logger.warn(
+			"`nodejs_compat_v2` compatibility flag and `--no-bundle` can't be used together. If you want to polyfill Node.js built-ins and disable Wrangler's bundling, please polyfill as part of your own bundling process."
+		);
+	}
```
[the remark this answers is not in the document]
IgorMinar: you are right on both counts. 

perhaps we should drop this completely?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -386,12 +386,38 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
+		compatibilityFlags.includes("nodejs_compat_v2");
+	if (nodejsCompatV2) {
+		// strip the "experimental:" prefix because workerd doesn't understand it yet.
+		compatibilityFlags[
+			compatibilityFlags.indexOf("experimental:nodejs_compat_v2")
+		] = "nodejs_compat_v2";
+	}
+	// nodejsCompatV2 supersedes nodejsCompat, so disable nodejsCompat if nodejsCompatV2 is enabled
+	const nodejsCompat =
+		!nodejsCompatV2 ?? compatibilityFlags.includes("nodejs_compat");
```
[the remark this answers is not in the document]
IgorMinar: what I wanted to do was to support the case when both flags are present in the compat flags list.

I just wanted to be nice and I also didn't want to add more code to error when both flags are present, since `nodejs_compat_v2` will be rolled into `nodejs_compat` eventually.

I don't feel very strongly about this though.

packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -325,7 +328,10 @@ export async function bundleWorker(
 						NodeModulesPolyfills(),
 					]
 				: []),
-			nodejsCompatPlugin(!!nodejsCompat),
+			// Runtime Node.js compatibility
+			...(nodejsCompat ? [nodejsCompatPlugin(!!nodejsCompat)] : []),
```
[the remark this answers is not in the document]
IgorMinar: yeah. I think I misunderstood the original intent. I'll defer to you on this one.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -0,0 +1,175 @@
+import nodePath from "node:path";
+import { cloudflare, env, nodeless } from "unenv";
+import { getBasePath } from "../../paths";
+import type { Plugin } from "esbuild";
+
+const REQUIRED_NODE_BUILT_IN_NAMESPACE = "node-built-in-modules";
+
+export const nodejsHybridPlugin: () => Plugin = () => {
+	const { alias, inject, /* polyfill, */ external } = env(nodeless, cloudflare);
```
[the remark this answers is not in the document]
IgorMinar: sgtm

packages/wrangler/src/deploy/deploy.ts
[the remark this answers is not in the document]
petebacondarwin: OK

packages/wrangler/src/deploy/deploy.ts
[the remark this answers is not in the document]
petebacondarwin: I understand that but this approach assumes the user implicitly understands the behaviour when both flags are enabled. It is kind of intuitive but...?

petebacondarwin (commented):
Last small thing:

- a changeset (with a minor bump) - or are we claiming that since this is internal/experimental it should not appear in the changelog? I don't feel we should try to hide it by doing that: users can see the code being added so it is not a secret, and also no changeset means that the package will not be published.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
IgorMinar: yes

packages/wrangler/package.json
```diff
@@ -82,6 +82,7 @@
 		"resolve.exports": "^2.0.2",
 		"selfsigned": "^2.0.1",
 		"source-map": "0.6.1",
+		"unenv": "npm:unenv-nightly@1.10.0-1717522572.87b9352",
```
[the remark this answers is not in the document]
IgorMinar: yeah. but this won't happen until shortly before we drop the `experimental:` prefix.

RamIdeas (approved):
LGTM just needs the changeset before merge

packages/wrangler/src/deploy/deploy.ts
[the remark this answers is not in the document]
petebacondarwin: I tweaked the messaging.

packages/wrangler/src/deploy/deploy.ts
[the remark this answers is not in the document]
petebacondarwin: I'm going to make this a hard error for now. We can relax it if we find this is too restrictive and annoying.

fixtures/nodejs-hybrid-app/tests/index.test.ts
```diff
@@ -0,0 +1,21 @@
...
+	it("should work when running code requiring polyfills", async ({
+		expect,
+	}) => {
+		const { ip, port, stop } = await runWranglerDev(
+			resolve(__dirname, "../src"),
+			["--port=0", "--inspector-port=0"]
+		);
+		try {
+			const response = await fetch(`http://${ip}:${port}`);
+			await expect(response.text()).resolves.toMatchInlineSnapshot(`"{"id":"1","timestamp":"2017-04-30T23:00:00.000Z","userstamp":"RNACEN","descr":"ENA","current_release":884,"full_descr":"ENA","alive":"Y","for_release":" ","display_name":"ENA","project_id":"","avg_length":"412","min_length":"10","max_length":"900074","num_sequences":"12086180","num_organisms":"814855","description":"provides a comprehensive record of the world's nucleotide sequencing information","url":"https://www.ebi.ac.uk/ena/browser/","example":[{"upi":"URS00002D0E0C","taxid":10090},{"upi":"URS000035EE7E","taxid":9606},{"upi":"URS0000000001","taxid":77133}],"reference":[{"title":"The European Nucleotide Archive in 2017","authors":"Silvester et al.","journal":"Nucleic Acids Res. 2017","pubmed_id":"29140475"}]}"`);
```
IgorMinar: is this output stable enough not to break the CI unexpectidly in the future? maybe a smaller snapshot would be better (e.g. just check for presence of `{"id":"`

IgorMinar (commented):
LGTM except for one regression that was introduced during the refactor. I didn't have enough time to debug it, but I discovered it in the matrix app and moved a reduction into the fixture as 0c2fc6d4fd19ab897c8313640cd63648fd1965e6

somehow don't tell esbuild that `node:assert` is a built-in, so it freaks out when it's not resolved to an absolute path. did we somehow forget to configure `externals`?

fixtures/nodejs-hybrid-app/tests/index.test.ts
petebacondarwin: Indeed! The timestamp was non-deterministic.
I have changed it to only check for the id.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -0,0 +1,191 @@
...
+											/*
+										// TODO: should we try to preserve globalThis.${globalName} if it exists?
+										*/ ""
+										}
+										return globalThis.${globalName} = globalVar;
+									})();
+
+								export {
+									exportable as '${globalName}',
+									exportable as 'globalThis.${globalName}',
```
RamIdeas, line 145: Does this work?

petebacondarwin: Our fixture doesn't test it but it is being used in the node.js compat matrix app that Igor and James C built. I believe that gives us enough confidence for today. But I'll see if I can add a quick check in the fixture.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -0,0 +1,191 @@
...
+									// TODO: should we try to preserve globalThis.${globalName} if it exists?
+									*/ ""
+									}
+									return globalThis.${globalName} = ${exportName};
+							})();
+
+							export {
+								exportable as '${globalName}',
+								exportable as 'global.${globalName}',
+								exportable as 'globalThis.${globalName}'
```
RamIdeas, line 186: @petebacondarwin can you check these too in the fixture test you proposed?

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * Currently we require that the v2 mode is configured via `experimental:nodejs_compat_v2` compat flag,
+ * where the `experimental:` prefix is stripped before being passed to the runtime since that does not
+ * understand this prefix.
+ *
+ * We assert that only one of these modes can be specified at a time.
+ * We assert that you must prefix v2 mode with `experimental`.
+ * We warn if using legacy or v2 mode.
+ */
+export function validateNodeCompat(
```
RamIdeas, line 23: 👏

packages/wrangler/src/dev.tsx
```diff
@@ -567,6 +569,12 @@ export async function startApiDev(args: StartDevOptions) {
 		additionalModules,
 	} = await validateDevServerSettings(args, config);
 
+	const nodejsCompatMode = validateNodeCompat(
+		args.nodeCompat ?? config.node_compat ?? false,
+		args.compatibilityFlags ?? config.compatibility_flags,
+		args.noBundle ?? config.no_bundle ?? false
+	);
```
RamIdeas, line 576: nit but these booleans could end up passed in the wrong order – better to make the params an object of named options

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
petebacondarwin: Added 0ec2e8632ced68a2db402223a73f84b893643be7

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
petebacondarwin: Done in 0ec2e8632ced68a2db402223a73f84b893643be7

packages/wrangler/src/dev.tsx
petebacondarwin: Good call. I'll update.

packages/wrangler/src/pages/functions/buildPlugin.ts
```diff
@@ -45,11 +45,10 @@ export function buildPluginFromFunctions({
 		minify,
 		sourcemap,
 		watch,
-		legacyNodeCompat,
 		// We don't currently have a mechanism for Plugins 'requiring' a specific compat date/flag,
 		// but if someone wants to publish a Plugin which does require this new `nodejs_compat` flag
 		// and they document that on their README.md, we should let them.
-		nodejsCompat: true,
+		nodejsCompatMode: nodejsCompatMode ?? "v1",
```
RamIdeas, line 51: does this comment still apply?

petebacondarwin: Yes, basically what this does is set the nodejsCompat (aka v1) to true unless there is a mode provided. In this case there will never be a mode provided since there is still no way for the user to specify one.

petebacondarwin: I could have just hard-coded it to v1 but this way, when we do finally provide such a mechanism, this will pick it up straight away.

RamIdeas (approved):
You managed to add this extra compat flag and make the codebase cleaner 👏

packages/wrangler/src/pages/functions/buildPlugin.ts
RamIdeas: yeah that's what threw me off – I thought the mechanism had been provided since it tries using the var before defaulting to v1

RamIdeas (approved):

petebacondarwin:
I'm going to override the GitGuardian check failure because it is picking up a DB secret that is actually a publicly accessible DB - so a false positive.

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -0,0 +1,103 @@
+import { UserError } from "../errors";
+import { logger } from "../logger";
+
+export type NodeJSCompatMode = "legacy" | "v1" | "v2" | null;
```
penalosa, line 4: Feel free to ignore this, but what about naming this with a bit more verbosity to describe what they do? Something like `legacy-build-time`/`runtime-only`/`hybrid-build-and-runtime`

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	const nodejsCompatV2 = compatibilityFlags.includes(
+		"experimental:nodejs_compat_v2"
+	);
+	const nodejsCompatV2NotExperimental =
+		compatibilityFlags.includes("nodejs_compat_v2");
+
+	if (nodejsCompatV2) {
+		// strip the "experimental:" prefix because workerd doesn't understand it yet.
```
penalosa, line 45: Given the flag is no longer experimental in workerd, do we want to keep this prefix stripping?

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -0,0 +1,103 @@
...
+	if (legacyNodeCompat && nodejsCompat) {
+		throw new UserError(
+			`The ${nodejsCompat ? "`nodejs_compat`" : "`nodejs_compat_v2`"} compatibility flag cannot be used in conjunction with the legacy \`--node-compat\` flag. If you want to use the Workers ${nodejsCompat ? "`nodejs_compat`" : "`nodejs_compat_v2`"} compatibility flag, please remove the \`--node-compat\` argument from your CLI command or \`node_compat = true\` from your config file.`
+		);
+	}
+
+	if (legacyNodeCompat && (nodejsCompat || nodejsCompatV2)) {
+		throw new UserError(
+			`The ${nodejsCompat ? "`nodejs_compat`" : "`nodejs_compat_v2`"} compatibility flag cannot be used in conjunction with the legacy \`--node-compat\` flag. If you want to use the Workers ${nodejsCompat ? "`nodejs_compat`" : "`nodejs_compat_v2`"} compatibility flag, please remove the \`--node-compat\` argument from your CLI command or \`node_compat = true\` from your config file.`
+		);
```
penalosa, line 66: This seems duplicative?

penalosa (approved):

packages/wrangler/src/deployment-bundle/node-compat.ts
petebacondarwin: I see where you are coming from but I think the more verbose strings will actually make the code less readable.
I'll add a JSDOC comment to explain the values here.

packages/wrangler/src/deployment-bundle/node-compat.ts
petebacondarwin: No we still want to require users of Wrangler to add this - it was orthogonal to workerd's experimental constraints - and is left here to avoid people trying to use this for production while we stabilize it.

packages/wrangler/src/deployment-bundle/node-compat.ts
petebacondarwin: Agh! Cut and paste mistake. Good spot.

petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/5991/commits/852db02ad3c8ff499c5910180c037b5a137f2f1d

packages/wrangler/src/deployment-bundle/node-compat.ts
petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/5991/commits/852db02ad3c8ff499c5910180c037b5a137f2f1d

fixtures/nodejs-hybrid-app/wrangler.toml
```diff
@@ -0,0 +1,11 @@
...
+main = "src/index.ts"
+compatibility_date = "2024-06-03"
+compatibility_flags = ["experimental:nodejs_compat_v2"]
+
+[vars]
+DB_HOSTNAME = "hh-pgsql-public.ebi.ac.uk"
+DB_PORT = "5432"
+DB_NAME = "pfmegrnargs"
+DB_USERNAME = "reader"
+DB_PASSWORD = "NWDMCE5xdipIjRrp"
```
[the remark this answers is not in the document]
petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/5991/commits/6298cd7199ecf1c5d760c3ba3fd34231e3c158db

fixtures/nodejs-hybrid-app/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Welcome to Cloudflare Workers! This is your first worker.
+ *
+ * - Run `npm run dev` in your terminal to start a development server
+ * - Open a browser tab at http://localhost:8787/ to see your worker in action
+ * - Run `npm run deploy` to publish your worker
+ *
+ * Learn more at https://developers.cloudflare.com/workers/
+ */
```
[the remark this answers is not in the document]
petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/5991/commits/91d257651513d23b1b12f82d3b4fd412772b0c03

PR #5896, opened 2024-05-22
Jest -> Vitest

penalosa:

Migrate from Jest to Vitest!

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: not public facing
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: not public facing


packages/wrangler/src/__tests__/ai.test.ts
```diff
@@ -117,6 +117,10 @@ describe("ai commands", () => {
 	});
 
 	it("should truncate model description", async () => {
+		const original = process.stdout.columns;
+		// Arbitrary fixed value for testing
```
penalosa, line 121: I picked this because it meant the snapshot could stay the same. It's an entirely arbitrary choice though

packages/wrangler/src/__tests__/cloudchamber/create.test.ts
```diff
@@ -171,7 +171,7 @@ describe("cloudchamber create", () => {
 		await expect(
 			runWrangler("cloudchamber create")
 		).rejects.toThrowErrorMatchingInlineSnapshot(
-			`"image is required but it's not passed as an argument"`
+			`[Error: image is required but it's not passed as an argument]`
```
penalosa, line 174: This is the first of many snapshot format changes. Unfortunately a change was required, since Jest formats errors in `toThrowErrorMatchingInlineSnapshot()` and `toMatchInlineSnapshot()` differently to each other (while Vitest formats them the same, and doesn't allow this behaviour to be customised)

packages/wrangler/src/__tests__/constellation.test.ts
```diff
@@ -75,7 +76,7 @@ describe("constellation commands", () => {
 
 	beforeEach(() => {
 		// @ts-expect-error we're using a very simple setTimeout mock here
-		jest.spyOn(global, "setTimeout").mockImplementation((fn, _period) => {
+		vi.spyOn(global, "setTimeout").mockImplementation((fn, _period) => {
```
penalosa, line 79: We use this pattern in a bunch of places—we should refactor to use Vitest fake timers at some point. I didn't do that here because I wanted to keep the diff clean

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -255,6 +256,7 @@ describe("deploy", () => {
 
 			it("should not throw an error in non-TTY if 'CLOUDFLARE_API_TOKEN' & 'account_id' are in scope", async () => {
 				process.env = {
+					...process.env,
```
penalosa, line 259: Spreading `process.env` here is required, as otherwise `os.tmpdir()` fails on Windows. I'm not really sure how this worked previously with Jest

packages/wrangler/src/__tests__/metrics.test.ts
```diff
@@ -17,18 +18,17 @@ import { clearDialogs, mockConfirm } from "./helpers/mock-dialogs";
 import { useMockIsTTY } from "./helpers/mock-istty";
 import { msw, mswSuccessOauthHandlers } from "./helpers/msw";
 import { runInTempDir } from "./helpers/run-in-tmp";
+import type { MockInstance } from "vitest";
 
 declare const global: { SPARROW_SOURCE_KEY: string | undefined };
+vi.unmock("../metrics/metrics-config");
```
penalosa, line 24: This replaces the use of a module level `useOriginal` flag, which wasn't working properly in Vitest

packages/wrangler/src/__tests__/pages/deploy.test.ts
```diff
@@ -34,6 +37,7 @@ describe("pages deploy", () => {
 	beforeEach(() => {
 		actualProcessEnvCI = process.env.CI;
 		process.env.CI = "true";
+		setIsTTY(false);
```
penalosa, line 40: This is more reliable than the `process.env.CI` setting (which wasn't working properly locally). I think we can remove the `process.env.CI` setting in a followup (I didn't remove it in this PR in order to minimise the diff)

packages/wrangler/src/__tests__/pages/pages-deployment-tail.test.ts
```diff
@@ -20,15 +22,41 @@ import type {
 import type { RequestInit } from "undici";
 import type WebSocket from "ws";
 
+vi.mock("ws", async (importOriginal) => {
```
penalosa, line 25: This has been moved out of `vitest.setup.ts` since we only need to mock `ws` in two files

packages/wrangler/src/__tests__/pages/pages-deployment-tail.test.ts
```diff
@@ -80,10 +102,11 @@ describe("pages deployment tail", () => {
 
 			await api.closeHelper();
 			expect(api.requests.deletion.count).toStrictEqual(1);
+			await api.closeHelper();
```
penalosa, line 105: For some reason this didn't work in an `afterEach()`... I've copied this line to the end of every test that needs it, but longer-term we should find a better way to test websockets, since this approach has layers upon layers of mocks

packages/wrangler/src/__tests__/sentry.test.ts
```diff
@@ -40,6 +40,7 @@ describe("sentry", () => {
 		msw.resetHandlers();
 	});
 	describe("non interactive", () => {
+		beforeEach(() => setIsTTY(false));
```
penalosa, line 43: Otherwise this didn't work when just running this test file

packages/wrangler/src/__tests__/sentry.test.ts
```diff
@@ -192,6 +193,10 @@ describe("sentry", () => {
 			const fakeInstallPath = "/wrangler/";
 			for (const exception of event.data.exception?.values ?? []) {
 				for (const frame of exception.stacktrace?.frames ?? []) {
+					if (frame.module.startsWith("@mswjs")) {
+						frame.module =
+							"@mswjs.interceptors.src.interceptors.fetch:index.ts";
+					}
```
penalosa, line 199: For some reason this module name was different on windows 🤦

packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
```diff
@@ -29,8 +30,10 @@ describe("versions deploy", () => {
 	runInTempDir();
 	mockConsoleMethods();
 	const std = collectCLIOutput();
+	const { setIsTTY } = useMockIsTTY();
 
 	beforeEach(() => {
+		setIsTTY(false);
```
penalosa, line 36: Without this the snapshots failed in CI

packages/wrangler/src/__tests__/vitest.setup.ts
```diff
@@ -247,42 +217,27 @@ jest.mock("execa", () => {
...
-			),
-		};
-	},
-	{ virtual: true }
-);
-
-jest.mock("@cloudflare/cli/streams", () => {
-	const { PassThrough } = jest.requireActual("node:stream");
+// make vitest understand virtual `worker:` imports
+// TODO: Actual build these with Vite
```
penalosa, line 224: This will come in a fast follow with the local runtime work

packages/wrangler/src/cfetch/index.ts
```diff
@@ -162,6 +162,10 @@ function throwFetchError(
 	resource: string,
 	response: FetchResult<unknown>
 ): never {
+	// This is an error from within an MSW handler
+	if (typeof vitest !== "undefined" && !("errors" in response)) {
+		throw response;
+	}
```
penalosa, line 168: This is admittedly a horrible hack. We should move away from the pattern of using `expect()` within MSW handler, which will remove the need for this.

packages/wrangler/src/d1/formatTimeAgo.ts
```diff
@@ -2,7 +2,9 @@ import TimeAgo from "javascript-time-ago";
 import en from "javascript-time-ago/locale/en";
 import prettyBytes from "pretty-bytes";
 
-TimeAgo.addDefaultLocale(en);
+TimeAgo.addLocale(en);
+TimeAgo.setDefaultLocale("en");
```
penalosa, line 6: Previously, `javascript-time-ago` complained about multiple calls of `addDefaultLocale()`

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -676,17 +676,10 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			limits: config.limits,
 		};
 
-		// As this is not deterministic for testing, we detect if in a jest environment and run asynchronously
-		// We do not care about the timing outside of testing
-		const bundleSizePromise = printBundleSize(
+		await printBundleSize(
```
penalosa, line 679: Contrary to the comment that was previously here—while we don't care about the value, we do care about the order of log statements for snapshot testing (and `printBundleSize()` writes to stdout)

packages/wrangler/src/pages/upload.tsx
```diff
@@ -355,6 +355,13 @@ export const upload = async (
 
 // Decode and check that the current JWT has not expired
 function isJwtExpired(token: string): boolean | undefined {
+	// During testing we don't use valid JWTs, so don't try and parse them
+	if (
+		typeof vitest !== "undefined" &&
+		(token === "<<funfetti-auth-jwt>>" || token === "<<funfetti-auth-jwt2>>")
+	) {
+		return false;
+	}
```
penalosa, line 364: Previously this was causing an unhandled error that Jest appeared to ignore

packages/wrangler/src/__tests__/ai.test.ts
RamIdeas: Clicked through to the diff expecting a round number – wasn't expecting... 186! 😆

packages/wrangler/src/__tests__/metrics.test.ts
RamIdeas: I thought the main goal of moving from jest to vitest was to unravel the global mocking and unmocking and avoid the pattern – do we have a plan to do this in a follow-up?

packages/wrangler/src/d1/formatTimeAgo.ts
RamIdeas: Why is one `en` and the other `"en"`?

packages/wrangler/src/__tests__/metrics.test.ts
penalosa: One of the issues with Jest was mocking modules globally and then unmocking them with a module level `useOriginal` flag (rather than calling `jest.unmock()`). Using `vi.unmock()` here makes it clearer what's going on, I think. 

We should definitely improve this in a followup though (and make sure to only mock metrics-config when needed). I kinda started that in this PR for some things (`ws` is no longer mocked globally, for instance, it's only mocked in the two files that need it), but I didn't want to make this a refactor as well as a migration.

packages/wrangler/src/d1/formatTimeAgo.ts
penalosa: One is a string, and the other is an object import from `TimeAgo`

packages/wrangler/package.json
```diff
@@ -171,11 +141,10 @@
 		"ink-table": "^3.0.0",
 		"is-ci": "^3.0.1",
 		"javascript-time-ago": "^2.5.4",
-		"jest": "^29.7.0",
-		"jest-websocket-mock": "^2.5.0",
 		"md5-file": "5.0.0",
 		"mime": "^3.0.0",
 		"minimatch": "^5.1.0",
+		"mock-socket": "^9.3.1",
```
RamIdeas, line 147: Noticing this added dep and `vitest-websocket-mock` – what's the reasoning for both?

penalosa: We'd been using it previously (just transitively through the dependencies of `jest-websocket-mock`). `mock-socket` is the client, and `vitest-websocket-mock` is the server

RamIdeas (approved):

PR #5910, opened 2024-05-24
chore: tweak the test:ci configuration

petebacondarwin:
Super simple update...


penalosa (commented):
This isn't quite the same-this will let turbo run fixtures concurrently, I think?

petebacondarwin:
> This isn't quite the same-this will let turbo run fixtures concurrently, I think?

That's what I want. Locally I am happy to allow a few flakes because if I run it multiple times the result will be full turbo.
So I only want the limited concurrency in the CI jobs.

penalosa (approved):

PR #5915, opened 2024-05-24
test: use structuredClone polyfill in tests as well

petebacondarwin:
Small refactor to fix JSRPC tests later.


penalosa (approved):
Do we need this polyfill anymore? Vitest doesn't support Node versions below 18

petebacondarwin:
> Do we need this polyfill anymore? Vitest doesn't support Node versions below 18

I think you are right. Now that we are testing on Node 18.
I'll leave this open but not land it until I have confirmed.

petebacondarwin:
Closing as our vitest integration requires Node 18 and so this polyfill is not needed.

PR #5916, opened 2024-05-24
feature: add support for JSRPC

### LINKED ISSUE #5386: Ensure JSRPC/named entrypoints are supported for `SELF` binding and Durable Object bindings to same worker in Vitest integration
To support integration testing with hot-module-reloading, the Vitest integration wraps `export default` and any user-defined Durable Object classes. This wrapping will need to be updated for JSRPC. In particular, we need to make sure RPC methods can still be called on `WorkerEntrypoint` `default` exports, and `DurableObject` subclasses. Only prototype methods can be called over RPC, so we'll need to be think about how we expose methods from a dynamically `import()`ed class on the prototype of a class. For `DurableObject`s, we can likely do this by modifying the prototype in a `blockConcurrencyWhile()` callback. For `WorkerEntrypoint`s, we might have to do something super hacky with a `Proxy` prototype (I assume this is allowed 🙈)?

### LINKED ISSUE #5389: `Expected Vitest to start running before importing modules` message should call out checking for multiple `vitest` versions
This message is almost always caused by multiple `vitest` versions being installed. We should call this out explicitly, and suggest using `npm/pnpm/yarn why vitest` to debug the issue.

https://github.com/cloudflare/workers-sdk/blob/426906a96b378823295067fc9ca8a495712ae94e/packages/vitest-pool-workers/src/worker/durable-objects.ts#L190

Ref: https://discord.com/channels/595317990191398933/1218150105777963101/1218608536741875894

petebacondarwin:

Fixes #5386 
Fixes #5389

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: covered by fixtures and doesn't affect e2es
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Not necessary because: covered by.

### DIFF
--- a/fixtures/vitest-pool-workers-examples/rpc/src/index.ts
+++ b/fixtures/vitest-pool-workers-examples/rpc/src/index.ts
@@ -1,0 +1,138 @@
+import { DurableObject, RpcTarget, WorkerEntrypoint } from "cloudflare:workers";
+
+export class Counter extends RpcTarget {
+	#value: number;
+
+	constructor(value: number) {
+		super();
+		this.#value = value;
+	}
+
+	get value() {
+		return this.#value;
+	}
+
+	increment(by = 1) {
+		return (this.#value += by);
+	}
+
+	clone() {
+		return new Counter(this.#value);
+	}
+}
+
+export class TestObject extends DurableObject<Env> {
+	#value: number = 0;
+
+	constructor(ctx: DurableObjectState, env: Env) {
+		super(ctx, env);
+		void ctx.blockConcurrencyWhile(async () => {
+			this.#value = (await ctx.storage.get("count")) ?? 0;
+		});
+	}
+
+	get value() {
+		return this.#value;
+	}
+
+	increment(by = 1) {
+		this.#value += by;
+		void this.ctx.storage.put("count", this.#value);
+		return this.#value;
+	}
+
+	scheduleReset(afterMillis: number) {
+		void this.ctx.storage.setAlarm(Date.now() + afterMillis);
+	}
+
+	async fetch(request: Request) {
+		return Response.json({
+			source: "TestObject",
+			method: request.method,
+			url: request.url,
+			ctxWaitUntil: typeof this.ctx.waitUntil,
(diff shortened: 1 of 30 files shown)


fixtures/vitest-pool-workers-examples/rpc/src/index.ts
```diff
@@ -0,0 +1,138 @@
...
+	increment(by = 1) {
+		return (this.#value += by);
+	}
+
+	clone() {
+		return new Counter(this.#value);
+	}
+}
+
+export class TestObject extends DurableObject<Env> {
```
Skye-31, line 24: What's the state of returning non-primitives over RPC? If it works, it would be great to have a test showing this - if it doesn't, the isolated storage assertion should be updated to mention that objects over DO RPC aren't currently supported

petebacondarwin: I think that objects and functions are generally OK but classes need to derive from `RpcTarget`?

See https://developers.cloudflare.com/workers/runtime-apis/rpc/#structured-clonable-types-and-more

Skye-31: When we were looking at the larger PR, there was a bug with returning plain objects over RPC, specifically from durable objects - would be great to know whether this is now supported or needs work

Skye-31: Is it possible to get a test showing this?

penalosa (approved):
This is good to go, I think

fixtures/vitest-pool-workers-examples/rpc/src/index.ts
penalosa: https://github.com/cloudflare/workers-sdk/pull/5916/commits/6e6292de5e0b3e002185d9a41b85745cdd6b7084

Skye-31 (approved):
Thanks very much!

PR #5939, opened 2024-05-29
Remote runtime controller

penalosa:

Supersedes #5867, fixes [DEVX-855](https://jira.cfdata.org/browse/DEVX-855), [DEVX-857](https://jira.cfdata.org/browse/DEVX-857), [DEVX-1280](https://jira.cfdata.org/browse/DEVX-1280), [DEVX-1281](https://jira.cfdata.org/browse/DEVX-1281)

- Tests
  - [ ] TODO (before merge)
  - [x] Included: The `RemoteRuntimeController` can't be usefully unit tested, and so it's primarily tested by the e2e test runs with the `--x-devenv-runtime --remote` flags
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: internal refactor


packages/wrangler/src/api/dev.ts
```diff
@@ -176,7 +177,7 @@ export async function unstable_dev(
 		bundle: options?.bundle,
 		compatibilityDate: options?.compatibilityDate,
 		compatibilityFlags: options?.compatibilityFlags,
-		ip: options?.ip,
+		ip: options?.ip ?? "127.0.0.1",
```
CarmenPopoviciu, line 180: we're using `"win32" ? "127.0.0.1" : "localhost"` as the default port in all our code. Is this something we want to consider here as well?

packages/wrangler/src/api/dev.ts
```diff
@@ -79,6 +79,7 @@ export interface UnstableDevOptions {
 		testScheduled?: boolean; // Test scheduled events by visiting /__scheduled in browser
 		watch?: boolean; // unstable_dev doesn't support watch-mode yet in testMode
 	};
+	experimentalDevenvRuntime?: boolean;
```
CarmenPopoviciu, line 82:
```suggestion
	experimentalDevEnvRuntime?: boolean;
```

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -65,6 +66,22 @@ export interface StartDevWorkerOptions {
 		exclude?: string[];
 	};
 
+	// -- PASSTHROUGH -- FROM OLD CONFIG TO NEW CONFIG (TEMP)
```
CarmenPopoviciu, line 69: can you pls expand on this 🙏 ? What's the `old` vs the `new` config?

penalosa: One of the projects that's wrapped up in Start Dev Worker is refactoring Wrangler's internal data structures from a mishmash of props and CfWorkerInit to just use StartDevWorkerOptions. However, since we're doing this in stages it'd be difficult (and inefficient) to shift all of the code over to use the new internal structure immediately, since that would involve converting back and forth in the course of a normal `wrangler dev` session. Instead, for some of these properties we're passing through the old data structure and can clean it up once all the controllers are in place and the data flow is finalised (Config Controller + Bundle Controller are the main ones that need done at this stage)

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -48,118 +48,20 @@ function getName(config: StartDevWorkerOptions) {
...
 
-	const persistence = event.config.dev?.persist
-		? getLocalPersistencePath(
-				typeof event.config.dev?.persist === "object"
-					? event.config.dev?.persist.path
-					: undefined,
-				event.config.config?.path
-			)
-		: null;
+	const persistence = getLocalPersistencePath(
```
CarmenPopoviciu, line 59: what changed that makes the `event.config.dev?.persist` check irrelevant here? Since `persist` is optional, won't `event.config.dev?.persist.path` potentially break?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -220,18 +122,16 @@ async function convertToConfigBundle(
 		initialPort: undefined,
 		initialIp: "127.0.0.1",
 		rules: [],
-		inspectorPort:
-			event.config.dev?.inspector?.port ??
-			(await getPort({ port: DEFAULT_INSPECTOR_PORT })),
+		inspectorPort: 0,
```
CarmenPopoviciu, line 125: does this mean we're forever ignoring whatever's been passed in `dev.inspector.port`?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
penalosa: This matches existing behaviour (the diff is because I'd previously misunderstood how this should work). See https://github.com/cloudflare/workers-sdk/blob/7d93a60deeb8b64ca8ff8be66860ffbda4b860ad/packages/wrangler/src/dev.tsx#L848-L851.

Essentially, `getLocalPersistencePath()` treats the first argument as `string | undefined`

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
+				workerAccount,
+				workerContext,
+				this.#abortController.signal
+			);
+		} catch (err: unknown) {
+			assert(err && typeof err === "object");
+			// instead of logging the raw API error to the user,
+			// give them friendly instructions
+			// for error 10063 (workers.dev subdomain required)
+			if ("code" in err && err.code === 10063) {
```
CarmenPopoviciu, line 63: is it worth mapping the err code here to a more human readable name? So smth like:
`const SUBDOMAIN_REQUIRED_ERROR = 10063` ?

I know Pages has https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/pages/errors.ts. Not sure if an equivalent exists for wrangler proper?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
penalosa: No, it's used as the port for the inspector proxy (this value is for the user worker inspector port). See https://github.com/cloudflare/workers-sdk/blob/7d93a60deeb8b64ca8ff8be66860ffbda4b860ad/packages/wrangler/src/dev.tsx#L773 in the existing flow, although I'm realising that this should probably include a `getPort` call as well

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
penalosa: Potentially, but this is copied from the existing flow and I didn't want to change too much code. We tend to use numeric error codes in too many places in Wrangler proper in general

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
+				logger.error(`${errorMessage}\n${solutionMessage}\n${onboardingLink}`);
+			} else if (
+				"cause" in err &&
+				(err.cause as { code: string; hostname: string })?.code === "ENOTFOUND"
+			) {
+				logger.error(
+					`Could not access \`${(err.cause as { code: string; hostname: string }).hostname}\`. Make sure the domain is set up to be proxied by Cloudflare.\nFor more details, refer to https://developers.cloudflare.com/workers/configuration/routing/routes/#set-up-a-route`
+				);
+			} else if (err instanceof UserError) {
+				logger.error(err.message);
```
CarmenPopoviciu, line 78: should we prefix this error as well, similarly to [here](https://github.com/cloudflare/workers-sdk/pull/5939/files#diff-f3311016ad4bd6233c4740b176f304287ceeb2d88e865bdfed4ba09d0eaa3087R83)?
```
logger.error("Error while creating remote dev session:", err);
```

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
+			return workerPreviewToken;
+		} catch (err: unknown) {
+			assert(err && typeof err === "object");
+			// we want to log the error, but not end the process
+			// since it could recover after the developer fixes whatever's wrong
+			// instead of logging the raw API error to the user,
+			// give them friendly instructions
+			if ((err as unknown as { code: string }).code !== "ABORT_ERR") {
+				// code 10049 happens when the preview token expires
+				if ("code" in err && err.code === 10049) {
```
CarmenPopoviciu, line 142: [similar](https://github.com/cloudflare/workers-sdk/pull/5939/files#r1624162679) comment here

packages/wrangler/src/api/startDevWorker/utils.ts
```diff
@@ -41,3 +44,158 @@ export function urlFromParts(
...
+	hook: Hook<T>
+): T | Promise<T>;
+export function unwrapHook<T extends string | number | object>(
+	hook: Hook<T> | undefined
+): T | Promise<T> | undefined;
+export function unwrapHook<T extends string | number | object>(hook: Hook<T>) {
+	return typeof hook === "function" ? hook() : hook;
+}
+
+export class MissingConfigError extends Error {
```
CarmenPopoviciu, line 58: would this be better placed in [errors.ts](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/errors.ts) ?

packages/wrangler/src/api/startDevWorker/utils.ts
```diff
@@ -41,3 +44,158 @@ export function urlFromParts(
...
+}
+
+export async function getBinaryFileContents(file: File<string | Uint8Array>) {
+	if ("contents" in file) {
+		if (file.contents instanceof Buffer) {
+			return file.contents;
+		}
+		return Buffer.from(file.contents);
+	}
+	return readFile(file.path);
```
CarmenPopoviciu, line 71: do we need to specify the encoding here?

packages/wrangler/src/dev.tsx
```diff
@@ -287,6 +287,13 @@ export function devOptions(yargs: CommonYargsArgv) {
 					"Show interactive dev session  (defaults to true if the terminal supports interactivity)",
 				type: "boolean",
 			})
+			.option("experimental-devenv-runtime", {
```
CarmenPopoviciu, line 290:
```suggestion
			.option("experimental-dev-env-runtime", {
```

packages/wrangler/src/dev.tsx
```diff
@@ -287,6 +287,13 @@ export function devOptions(yargs: CommonYargsArgv) {
...
+				alias: ["x-devenv-runtime"],
```
CarmenPopoviciu, line 291:
```suggestion
				alias: ["x-dev-env-runtime"],
```

packages/wrangler/src/dev.tsx
```diff
@@ -512,6 +519,7 @@ export async function startDev(args: StartDevOptions) {
 					sendMetrics={configParam.send_metrics}
 					testScheduled={args.testScheduled}
 					projectRoot={projectRoot}
+					experimentalDevenvRuntime={args.experimentalDevenvRuntime}
```
CarmenPopoviciu, line 522:
```suggestion
					experimentalDevEnvRuntime={args.experimentalDevEnvRuntime}
```

packages/wrangler/src/dev.tsx
```diff
@@ -633,12 +642,12 @@ export async function startApiDev(args: StartDevOptions) {
 			showInteractiveDevSession: args.showInteractiveDevSession,
 			forceLocal: args.forceLocal,
 			enablePagesAssetsServiceBinding: args.enablePagesAssetsServiceBinding,
-			local: !args.remote,
 			firstPartyWorker: configParam.first_party_worker,
 			sendMetrics: configParam.send_metrics,
 			testScheduled: args.testScheduled,
 			disableDevRegistry: args.disableDevRegistry ?? false,
 			projectRoot,
+			experimentalDevenvRuntime: args.experimentalDevenvRuntime,
```
CarmenPopoviciu, line 650:
```suggestion
			experimentalDevEnvRuntime: args.experimentalDevEnvRuntime,
```

packages/wrangler/src/dev/dev.tsx
```diff
@@ -295,35 +370,67 @@ function DevSession(props: DevSessionProps) {
...
 
 	const onBundleStart = useCallback(() => {
 		devEnv.proxy.onBundleStart({
 			type: "bundleStart",
 			config: startDevWorkerOptions,
 		});
 	}, [devEnv, startDevWorkerOptions]);
+	const onBundleComplete = useCallback(
+		(bundle: EsbuildBundle) => {
+			if (props.experimentalDevenvRuntime) {
```
CarmenPopoviciu, line 414:
```suggestion
			if (props.experimentalDevEnvRuntime) {
```

packages/wrangler/src/dev/dev.tsx
```diff
@@ -466,11 +557,11 @@ function DevSession(props: DevSessionProps) {
 			);
 		}
 
-		if (bundle.current) {
+		if (bundle) {
 			latestReloadCompleteEvent.current = {
 				type: "reloadComplete",
 				config: startDevWorkerOptions,
-				bundle: bundle.current,
+				bundle: bundle,
```
CarmenPopoviciu, line 564:
```suggestion
				bundle,
```

packages/wrangler/src/dev/dev.tsx
```diff
@@ -539,8 +630,12 @@ function DevSession(props: DevSessionProps) {
 			host={props.host}
 			routes={props.routes}
 			onReady={announceAndOnReady}
-			sourceMapPath={bundle.current?.sourceMapPath}
+			sourceMapPath={bundle?.sourceMapPath}
 			sendMetrics={props.sendMetrics}
+			// startDevWorker
+			accountId={accountId}
+			setAccountId={setAccountIdAndResolveDeferred}
+			experimentalDevenvRuntime={props.experimentalDevenvRuntime}
```
CarmenPopoviciu, line 638:
```suggestion
			experimentalDevEnvRuntime={props.experimentalDevEnvRuntime}
```

packages/wrangler/src/dev/local.tsx
```diff
@@ -53,6 +53,7 @@ export interface LocalProps {
 	testScheduled?: boolean;
 	sourceMapPath: string | undefined;
 	services: Config["services"] | undefined;
+	experimentalDevenvRuntime: boolean;
```
CarmenPopoviciu, line 56:
```suggestion
	experimentalDevEnvRuntime: boolean;
```

packages/wrangler/src/dev/local.tsx
```diff
@@ -170,6 +162,19 @@ function useLocalWorker(props: LocalProps) {
 		}
 	}, [props.bindings.durable_objects?.bindings]);
 
+	if (!props.experimentalDevenvRuntime) {
```
CarmenPopoviciu, line 165:
```suggestion
	if (!props.experimentalDevEnvRuntime) {
```

packages/wrangler/src/dev/remote.tsx
```diff
@@ -68,33 +68,39 @@ interface RemoteProps {
...
-		usageModel: props.usageModel,
-		env: props.env,
-		legacyEnv: props.legacyEnv,
-		host: props.host,
-		routes: props.routes,
-		onReady: props.onReady,
-		sendMetrics: props.sendMetrics,
-		port: props.port,
-	});
+	if (!props.experimentalDevenvRuntime) {
```
CarmenPopoviciu, line 80:
```suggestion
	if (!props.experimentalDevEnvRuntime) {
```

CarmenPopoviciu: isn't this breaking existing behaviour, with us now requiring the `experimentalDevEnvRuntime` flag to be set in order to run `useWorker()`?

packages/wrangler/src/dev/remote.tsx
```diff
@@ -68,33 +68,39 @@ interface RemoteProps {
 		| undefined;
 	sourceMapPath: string | undefined;
 	sendMetrics: boolean | undefined;
+
+	setAccountId: (accountId: string) => void;
+	experimentalDevenvRuntime: boolean;
```
CarmenPopoviciu, line 73:
```suggestion
	experimentalDevEnvRuntime: boolean;
```

packages/wrangler/src/dev/start-server.ts
```diff
@@ -99,12 +104,31 @@ export async function startDevServer(
...
+					const accountChoices = await getAccountChoices();
+					if (accountChoices.length === 1) {
+						saveAccountToCache({
+							id: accountChoices[0].id,
+							name: accountChoices[0].name,
+						});
+						accountId = accountChoices[0].id;
+					} else {
+						throw logger.error(
+							"In a non-interactive environment, it is mandatory to specify an account ID, either by assigning its value to CLOUDFLARE_ACCOUNT_ID, or as `account_id` in your `wrangler.toml` file."
```
CarmenPopoviciu, line 125:
```suggestion
							"Missing account ID.
In a non-interactive environment, it is mandatory to specify an account ID. Please: 
❯❯ specify it in your wrangler.toml file: `account_id = <your_account_id>, or
❯❯ assign its value to the CLOUDFLARE_ACCOUNT_ID environment variable: CLOUDFLARE_ACCOUNT_ID = <your_account_id>
```

packages/wrangler/src/dev/start-server.ts
```diff
@@ -148,7 +172,34 @@ export async function startDevServer(
 		),
 	});
 
+	console.log({ experimentalDevenvRuntime: props.experimentalDevenvRuntime });
```
CarmenPopoviciu, line 175: do we mean to keep this log?

packages/wrangler/src/dev/start-server.ts
```diff
@@ -148,7 +172,34 @@ export async function startDevServer(
...
+
+	if (props.experimentalDevenvRuntime) {
```
CarmenPopoviciu, line 177: what about the logs in this `if` block? are these meant to be kept as well?

packages/wrangler/src/dev/start-server.ts
```diff
@@ -209,15 +260,17 @@ export async function startDevServer(
 			workerDefinitions,
 			sourceMapPath: bundle?.sourceMapPath,
 			services: props.bindings.services,
+			experimentalDevenvRuntime: props.experimentalDevenvRuntime,
```
CarmenPopoviciu, line 263:
```suggestion
			experimentalDevEnvRuntime: props.experimentalDevenvRuntime,
```

packages/wrangler/src/dev/start-server.ts
```diff
@@ -148,7 +172,34 @@ export async function startDevServer(
...
+				const { proxyWorker } = await devEnv.proxy.ready.promise;
+				const url = await proxyWorker.ready;
+
+				props.onReady?.(url.hostname, parseInt(url.port), ev.proxyData);
+			});
+		});
+	}
+
 	if (props.local) {
+		console.log("UNSTABLE_DEV LOCAL MODE");
```
CarmenPopoviciu, line 202: are we keeping this?

packages/wrangler/src/dev/start-server.ts
```diff
@@ -209,15 +260,17 @@ export async function startDevServer(
...
 		});
 
 		return {
 			stop: async () => {
 				await Promise.all([stop(), stopWorkerRegistry(), devEnv.teardown()]);
 			},
-			// TODO: inspectorUrl,
 		};
 	} else {
+		console.log("UNSTABLE_DEV REMOTE MODE");
```
CarmenPopoviciu, line 272: are we keeping this log?

packages/wrangler/src/dev/start-server.ts
```diff
@@ -260,12 +313,14 @@ export async function startDevServer(
 			},
 			sourceMapPath: bundle?.sourceMapPath,
 			sendMetrics: props.sendMetrics,
+			experimentalDevenvRuntime: props.experimentalDevenvRuntime,
```
CarmenPopoviciu, line 316:
```suggestion
			experimentalDevEnvRuntime: props.experimentalDevEnvRuntime,
```

CarmenPopoviciu, line 316:
```suggestion
			experimentalDevEnvRuntime: props.experimentalDevEnvRuntime,
```

packages/wrangler/src/dev/dev.tsx
```diff
@@ -274,17 +278,88 @@ type DevSessionProps = DevProps & {
...
-		() => ({
+	const startDevWorkerOptions: StartDevWorkerOptions = useMemo(() => {
+		const routes =
+			props.routes?.map((r) =>
+				typeof r === "string"
+					? ({
+							type: "route",
+							pattern: r,
+						} as const)
+					: ({ type: "route", ...r } as const)
```
andyjessop, line 314: Another option would be to define `Trigger` as the generic argument to `map` in these three definitions. That could provide more type safety if we come to modify or expand these object constructions.

packages/wrangler/src/dev/use-esbuild.ts
```diff
@@ -288,5 +290,12 @@ export function useEsbuild({
 		onStart,
 		defineNavigatorUserAgent,
 	]);
+
+	useEffect(() => {
+		if (bundle) {
+			onComplete(bundle);
```
andyjessop, line 296: Nice decoupling 👍

packages/wrangler/src/dev/dev.tsx
```diff
@@ -183,6 +185,8 @@ export type DevProps = {
 	sendMetrics: boolean | undefined;
 	testScheduled: boolean | undefined;
 	projectRoot: string | undefined;
+
```
andyjessop, line 188:
```suggestion
```

packages/wrangler/src/dev/dev.tsx
```diff
@@ -295,35 +370,67 @@ function DevSession(props: DevSessionProps) {
...
 	const onBundleStart = useCallback(() => {
 		devEnv.proxy.onBundleStart({
 			type: "bundleStart",
 			config: startDevWorkerOptions,
 		});
 	}, [devEnv, startDevWorkerOptions]);
+	const onBundleComplete = useCallback(
+		(bundle: EsbuildBundle) => {
+			if (props.experimentalDevenvRuntime) {
+				devEnv.runtimes.forEach((runtime) =>
```
andyjessop, line 415: Is there a chance of a race condition here? I'm not completely clear yet, but I'm wondering if something like this could happen:

1. User sets remote flag.
2. Bundler starts with remote runtime `onBundleComplete`.
3. User sets local flag.
4. Bundler starts with local runtime `onBundleComplete`.
5. Remote runtime `onBundleComplete` fires.

Perhaps we can add an E2E test to confirm the intended behaviour here to avoid regressions.

andyjessop (approved):

packages/wrangler/src/dev/start-server.ts
CarmenPopoviciu: or smth along those lines...

packages/wrangler/src/api/dev.ts
RamIdeas: No this is how yargs camel cases `--experimental-devenv-runtime` – best to just be consistent and not force it

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
RamIdeas: Let's let typescript checks handle these kind of nits – fair comment if the typescript checks fail but in this case they are passing

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
+
+		void this.#mutex.runWith(() => this.#onBundleComplete(ev, id));
 	}
 	onPreviewTokenExpired(_: PreviewTokenExpiredEvent): void {
 		notImplemented(this.onPreviewTokenExpired.name, this.constructor.name);
 	}
 
 	async teardown() {
-		notImplemented(this.teardown.name, this.constructor.name);
+		// There's no way to teardown remote preview sessions, and so this is ignored in remote mode
```
RamIdeas, line 300: We should nullify the `#sessions` field so it can be garbage collected and not reused (although nothing should happen after a teardown but we might have some async callbacks which we don't bother cleaning up that might use it)

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
 	BundleCompleteEvent,
 	BundleStartEvent,
 	PreviewTokenExpiredEvent,
 	ReloadCompleteEvent,
 	ReloadStartEvent,
 } from "./events";
+import type { Trigger } from "./types";
 
 export class RemoteRuntimeController extends RuntimeController {
+	#abortController = new AbortController();
```
RamIdeas, line 39: I copied the pattern of use AbortController for cancellations from miniflare (/wrangler?) but, to be useful, we need to check `abortController.aborted` after each `await`

RamIdeas: We can wrap that in a `this.checkAborted()` call and have that as a one-liner after each await

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
+					},
+					userWorkerInspectorUrl: {
+						protocol: token.inspectorUrl.protocol,
+						hostname: token.inspectorUrl.hostname,
+						port: token.inspectorUrl.port.toString(),
+						pathname: token.inspectorUrl.pathname,
+					},
+					userWorkerInnerUrlOverrides: {
+						hostname: token.host,
+					},
```
RamIdeas, line 247: This can't actually be used in remote-mode – the hostname is going to be `token.host` anyway

packages/wrangler/src/api/dev.ts
CarmenPopoviciu: oh, I didn't realise that. kk, carry on then :)

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
 	onBundleStart(_: BundleStartEvent) {
-		notImplemented(this.onBundleStart.name, this.constructor.name);
+		// Abort any previous operations when a new bundle is started
+		this.#abortController.abort();
+		this.#abortController = new AbortController();
 	}
-	onBundleComplete(_: BundleCompleteEvent) {
-		notImplemented(this.onBundleComplete.name, this.constructor.name);
+	onBundleComplete(ev: BundleCompleteEvent) {
+		const id = ++this.#currentBundleId;
```
RamIdeas, line 281: I'd prefer we didn't have this incrementing id stuff – we can accomplish the same thing with identity checks of `bundle` and `config` (see `latestConfig` in ProxyController)

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
RamIdeas: Actually, we have an AbortController to abort

packages/wrangler/src/api/startDevWorker/types.ts
RamIdeas: Samuel is correct but I have a commit I didn't push that removes the need for passthrough data structures

packages/wrangler/src/dev.tsx
RamIdeas: This is fair. I was trying to avoid too many dashes but since we're going to remove `-runtime` lgtm 👍

packages/wrangler/src/dev/dev.tsx
RamIdeas: That's fair. It would need to be `Extract<Trigger, { type: '...' }>` for each .map

packages/wrangler/src/dev/dev.tsx
RamIdeas: I agree we should have a e2e test for this common/likely scenario.

I've made an effort to keep as many methods synchronous (vs async) to avoid race conditons. Which is the case here. The async work inside those event handlers will be the only parts that need to consider races and we do that with equality checks on bundle/config with the latest event (currently accomplished with incrementing ids)

packages/wrangler/src/dev/dev.tsx
RamIdeas: Besides consistency, there's no downside to this syntax. The lint rule will be more annoying than beneficial imo

packages/wrangler/src/dev/remote.tsx
RamIdeas: The flag has to be NOT set to run useWorker

packages/wrangler/src/dev/start-server.ts
RamIdeas: This error message is likely from existing code. More interesting, why are we throwing logger.error (which returns void)?

packages/wrangler/src/dev/start-server.ts
RamIdeas: Nope

packages/wrangler/src/dev/start-server.ts
```diff
@@ -148,7 +172,34 @@ export async function startDevServer(
...
+				type: "bundleComplete",
+				config: startDevWorkerOptions,
+				bundle,
+			});
+		});
+
+		// to comply with the current contract of this function, call props.onReady on reloadComplete
+		devEnv.runtimes.forEach((runtime) => {
+			runtime.on("reloadComplete", async (ev) => {
+				console.log("devEnv.runtimes.onReloadComplete");
```
RamIdeas, line 192: 3 `console.log`s to remove

packages/wrangler/src/dev/start-server.ts
RamIdeas: no

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
```diff
@@ -1,30 +1,303 @@
...
+					(trigger): trigger is Extract<Trigger, { type: "route" }> =>
+						trigger.type === "route"
+				)
+				.map((trigger) => {
+					const { type: _, ...route } = trigger;
+					if (!("custom_domain" in route)) {
+						return route.pattern;
+					}
+					return route;
+				});
```
RamIdeas, line 171: I had this code (unpushed 🤦) on my branch:

```suggestion
				.map((trigger): Route => {
					const { type: _, ...route } = trigger;

					if (
						"custom_domain" in route ||
						"zone_id" in route ||
						"zone_name" in route
					) {
						return route;
					} else {
						return route.pattern; // SimpleRoute
					}
				});
```

Just want to ensure it's functionally identical..?

packages/wrangler/src/dev/dev.tsx
penalosa: Arguably one of the main points of linting _is_ consistency though

packages/wrangler/src/dev/start-server.ts
penalosa: No, I don't think any are—I'll go through and remove

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
penalosa: I modified the flow to use abort controllers in the same way as the existing flow—happy to change to a different approach, but I'm not quite sure what you mean here? Why do we need to check `abortController.aborted`?

RamIdeas: After each await (each async step) the abort controller could've become aborted. Unless we check `abortController.aborted` and return/throw if it is true, the code will continue to run even though it doesn't need to (it's an optimisation but the entire usage of AbortController is essentially an optimisation)

penalosa:
The history and diff here got a tad messed up by the squash-merging of the local runtime controller. I'll fix it up once Rahul has added an extra commit on top to handle binding coercion

RamIdeas:
> The history and diff here got a tad messed up by the squash-merging of the local runtime controller. I'll fix it up once Rahul has added an extra commit on top to handle binding coercion

As discussed offline, no need to block this on my commit – I can add it in a follow-up PR

RamIdeas (approved):

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
penalosa: I've refactored the error code to pull out shared handling logic between the new and old

packages/wrangler/src/api/startDevWorker/utils.ts
penalosa: No, because we're returning a `Buffer` rather than a string

packages/wrangler/src/dev/dev.tsx
penalosa: 567f10bf92069c7022f8b6f0866dac3373004797

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
penalosa: 83d914f61aed0379bb2a6ab26754f1314001df38

packages/wrangler/src/api/startDevWorker/types.ts
penalosa: We're going to address this in a followup

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
penalosa: `persist` is optional, but `path` is required within `persist`, and so we don't need an extra `?`

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
penalosa: Potentially, but let's address this in a followup since the messaging is the same as the existing flow

packages/wrangler/src/dev/dev.tsx
penalosa: I've fixed this for now—let's revisit our linting/prettier rules here in a followup

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
penalosa: I don't think that's needed here. The primary async operations are `fetch` calls, and we pass in the abort signal to those (and handle the thrown abort errors)

packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts
penalosa: Let's address in a followup—for now this mimics the patterns in the local runtime controller

packages/wrangler/src/dev/dev.tsx
penalosa: @andyjessop I'll add an e2e test to verify this, but the behaviour should be:
1. User sets remote flag.
2. Bundler triggers remote runtime `onBundleComplete` (and local runtime, which is ignored)
3. User sets local flag.
4. Bundler triggers local runtime `onBundleStart`  (and remote runtime, cancelling any in-progress operations)
5. Bundler triggers local runtime `onBundleComplete`  (and remote runtime, which is ignored, cancelling any in-)
6. Remote runtime finishes (if cancellation failed), but doesn't fire `reloadComplete` because the current bundle has changed.

penalosa: d676c6d6847007f4798089b66d33d2e6c754a471

packages/wrangler/e2e/dev-env.test.ts
```diff
@@ -0,0 +1,130 @@
...
+					devEnv.runtimes.forEach((runtime) =>
+						runtime.onBundleComplete({
+							type: "bundleComplete",
+							config,
+							bundle,
+						})
+					);
+
+					// Immediately switch runtime
+					config.dev.remote = !firstRemote
```
RamIdeas, line 83: Don't mutate. The object identity will not be reliable. Please update immutably. The ConfigController will expose a method to deeply update immutably but for now please do it manually:

```
config = Object.assign({}, config, { dev: { remote: !firstRemote } });
```

PR #5943, opened 2024-05-30
Update turbo.json

penalosa:

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because:

### DIFF
--- a/.github/actions/run-c3-e2e/action.yml
+++ b/.github/actions/run-c3-e2e/action.yml
@@ -58,7 +58,7 @@ runs:
         TEST_PM: ${{ inputs.packageManager }}
         TEST_PM_VERSION: ${{ inputs.packageManagerVersion }}
         SAVE_DIFFS: ${{ inputs.saveDiffs }}
-        CI_OS: ${{ runner.os }}
+        # CI_OS: ${{ runner.os }}
 
     - name: Upload Logs
       uses: actions/upload-artifact@v3
--- a/.github/actions/run-c3-e2e/action.yml
+++ b/.github/actions/run-c3-e2e/action.yml
@@ -58,6 +58,7 @@ runs:
         TEST_PM: ${{ inputs.packageManager }}
         TEST_PM_VERSION: ${{ inputs.packageManagerVersion }}
         SAVE_DIFFS: ${{ inputs.saveDiffs }}
+        CI_OS: ${{ runner.os }}
 
     - name: Upload Logs
       uses: actions/upload-artifact@v3
--- a/.github/actions/run-c3-e2e/action.yml
+++ b/.github/actions/run-c3-e2e/action.yml
@@ -58,7 +58,7 @@ runs:
         TEST_PM: ${{ inputs.packageManager }}
         TEST_PM_VERSION: ${{ inputs.packageManagerVersion }}
         SAVE_DIFFS: ${{ inputs.saveDiffs }}
-        # CI_OS: ${{ runner.os }}
+        CI_OS: ${{ runner.os }}
 
     - name: Upload Logs
       uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


dario-piotrowicz (approved):
Looks good to me, hopefully it fixes things 😄🤞 

I'm also wondering, `turbo.json` doesn't allow for comments...
it'd be nice to somehow document why you're adding these, so that in the future if someone thinks of deleting them (as they seem unnecessary) they can understand why they've been added in the first place.

If there is no right/convenient place for document these please at least make the commit message on merge if very clear and explanatory so that at least git blame/the git history can help a bit here 🙂🙏

packages/create-cloudflare/turbo.json
```diff
@@ -1,11 +1,9 @@
 {
 	"$schema": "http://turbo.build/schema.json",
 	"extends": ["//"],
-
 	"pipeline": {
 		"build": {
 			"env": [
-				"VITEST",
```
CarmenPopoviciu, line 8: removing this after having run `pnpm run check:lint` and confirming that we don't use that anywhere in `create-cloudflare`

CarmenPopoviciu, line 8: removing `VITEST` since this is declared as a `globalEnv` in https://github.com/cloudflare/workers-sdk/blob/main/turbo.json#L6

.github/actions/run-c3-e2e/action.yml
```diff
@@ -58,6 +58,7 @@ runs:
         TEST_PM: ${{ inputs.packageManager }}
         TEST_PM_VERSION: ${{ inputs.packageManagerVersion }}
         SAVE_DIFFS: ${{ inputs.saveDiffs }}
+        CI_OS: ${{ runner.os }}
```
CarmenPopoviciu, line 61: not 100% sure if we need this, but seeing this declared in all our other relevant actions, so ¯\_(ツ)_/¯

penalosa: We do, otherwise turbo won't consider the OS part of the cache key

packages/create-cloudflare/turbo.json
```diff
@@ -23,7 +21,8 @@
 				"E2E_PROJECT_PATH",
 				"E2E_RETRIES",
 				"E2E_NO_DEPLOY"
-			]
+			],
+			"dependsOn": ["build"]
```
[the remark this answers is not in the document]
CarmenPopoviciu: this also seems to mean, that all env vars declared at `build` level get inherited by `test:e2e`, which is why I have not added them again here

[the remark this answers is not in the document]
penalosa: I think some env vars have been removed from this PR? Initially CI_OS was added here

.github/actions/run-c3-e2e/action.yml
dario-piotrowicz: But, it is already here: https://github.com/cloudflare/workers-sdk/blob/9a804b9061006f2c2d508151d86246400055e4fd/turbo.json#L6

👆 isn't this sufficient?

packages/create-cloudflare/turbo.json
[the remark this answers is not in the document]
CarmenPopoviciu: that is a `glabalEnv` var, as Dario pointed out. That is why I removed it

.github/actions/run-c3-e2e/action.yml
penalosa: No—that forces turbo to consider it, but we also need to make sure it's defined

dario-piotrowicz: damn.... sorry I read the comments and not the file properly... yeah this is the yml file actually defining it! sorry my bad! 🙈

dario-piotrowicz (approved):

PR #3735, opened 2023-08-10
[Wrangler] Update the commands list

lrapoport-cf:
## What this PR solves / how to test:
This PR cleans up and standardises the look and feel of all `wrangler` commands as displayed by `wrangler --help` and `wrangler <cmd> --help`. 

<details>
<summary>Please refer to original designs here</summary>
Screenshot 2024-06-19 at 20 38 13
</details>

**Before**
Screenshot 2024-02-26 at 18 50 41

**After**
Screenshot 2024-06-19 at 20 32 12

Here is the list of all public facing wrangler commands, in their updated version:

<details>
<summary>**wrangler deployments --help**</summary>
Screenshot 2024-06-19 at 19 58 58
</details>

<details>
<summary>**wrangler deployments --x-versions --help**</summary>
Screenshot 2024-06-19 at 19 59 30
</details>

<details>
<summary>**npx wrangler rollback --help**</summary>
Screenshot 2024-06-19 at 20 12 21
</details>

<details>
<summary>**npx wrangler rollback --x-versions --help**</summary>
Screenshot 2024-06-19 at 20 13 05
</details>

<details>
<summary>**npx wrangler versions --x-versions --help**</summary>
Screenshot 2024-06-19 at 20 14 33
</details>

<details>
<summary>**wrangler triggers --x-versions --help**</summary>
Screenshot 2024-06-19 at 20 15 36
</details>

<details>
<summary>**wrangler kv --help**</summary>
Screenshot 2024-06-19 at 20 18 59
</details>

<details>
<summary>**wrangler queues --help**</summary>
Screenshot 2024-06-19 at 20 18 25
</details>

<details>
<summary>**wrangler r2 --help**</summary>
Screenshot 2024-06-19 at 20 19 31
</details>

<details>
<summary>**wrangler d1 --help**</summary>
Screenshot 2024-06-19 at 20 21 44
</details>

<details>
<summary>**wrangler vectorize --help**</summary>
Screenshot 2024-06-19 at 20 22 49
</details>

<details>
<summary>**wrangler hyperdrive --help**</summary>
Screenshot 2024-06-19 at 20 23 34
</details>

<details>
<summary>**wrangler pages --help**</summary>
Screenshot 2024-06-19 at 20 24 30
</details>

<details>
<summary>**wrangler mtls-certificate --help**</summary>
Screenshot 2024-06-19 at 20 25 22
</details>

<details>
<summary>**wrangler cloudchamber --help**</summary>
[...]

PR #5950, opened 2024-05-31
Version secrets + secret behaviour changes

WalshyDev:

* Adds new versions secret commands (put, bulk, list and view).
* Adds a --pretty option to `wrangler secret list`
* Handles the new API error in rollback when attempting to rollback to a version with changed secrets from current

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (after merge - experimental features)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

## Screenshots

**Version in dash**
image

### DIFF
--- a/.changeset/little-rice-promise.md
+++ b/.changeset/little-rice-promise.md
@@ -1,0 +1,13 @@
+---
+"wrangler": minor
+---
+
+feat: add `wrangler versions secret put`, `wrangler versions secret bulk` and `wrangler versions secret list`
+
+`wrangler versions secret put` allows for you to add/update a secret even if the latest version is not fully deployed. A new version with this secret will be created, the existing secrets and config are copied from the latest version.
+
+`wrangler versions secret bulk` allows you to bulk add/update multiple secrets at once, this behaves the same as `secret put` and will only make one new version.
+
+`wrangler versions secret list` lists the secrets available to the currently deployed versions. `wrangler versions secret list --latest-version` or `wrangler secret list` will list for the latest version.
+
+Additionally, we will now prompt for extra confirmation if attempting to rollback to a version with different secrets than the currently deployed.
--- a/packages/wrangler/src/__tests__/rollback.test.ts
+++ b/packages/wrangler/src/__tests__/rollback.test.ts
@@ -1,0 +1,235 @@
+import { http, HttpResponse } from "msw";
+import { describe, expect, test } from "vitest";
+import { CANNOT_ROLLBACK_WITH_MODIFIED_SECERT_CODE } from "../versions/api";
+import { collectCLIOutput } from "./helpers/collect-cli-output";
+import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
+import { mockConfirm, mockPrompt } from "./helpers/mock-dialogs";
+import { useMockIsTTY } from "./helpers/mock-istty";
+import { createFetchResult, msw } from "./helpers/msw";
+import { runWrangler } from "./helpers/run-wrangler";
+import type { ApiDeployment } from "../versions/types";
+
+describe("rollback", () => {
+	const std = collectCLIOutput();
+	const { setIsTTY } = useMockIsTTY();
+	mockAccountId();
+	mockApiToken();
+
+	function mockGetDeployments(multiVersion = false) {
+		const versions = multiVersion
+			? [
+					{ version_id: "version-id-1", percentage: 50 },
+					{ version_id: "version-id-2", percentage: 50 },
+				]
+			: [{ version_id: "version-id-1", percentage: 100 }];
+
+		msw.use(
+			http.get(
+				`*/accounts/:accountId/workers/scripts/:scriptName/deployments`,
+				async ({ params }) => {
+					expect(params.accountId).toEqual("some-account-id");
+					expect(params.scriptName).toEqual("script-name");
+
+					return HttpResponse.json(
+						createFetchResult({
+							deployments: [
+								{
+									id: "deployment-id",
+									source: "api",
+									strategy: "percentage",
+									versions,
+								},
+							] as ApiDeployment[],
+						})
+					);
+				},
+				{ once: true }
+			)
+		);
+	}
+
+	function mockGetVersion(versionId: string) {
+		msw.use(
+			http.get(
+				`*/accounts/:accountId/workers/scripts/:scriptName/versions/${versionId}`,
+				async ({ params }) => {
+					expect(params.accountId).toEqual("some-account-id");
+					expect(params.scriptName).toEqual("script-name");
+
+					return HttpResponse.json(
+						createFetchResult({
+							id: versionId,
+							metadata: {},
+							number: 2,
+							resources: {
+								bindings: [
+									{
+										type: "secret_text",
+										name: "SECRET_1",
+										text: "First secret",
+									},
+									{
+										type: "secret_text",
+										name: "SECRET_2",
+										text: "Second secret",
+									},
+									{
+										type: "secret_text",
+										name: "SECRET_3",
+										text: "Third secret",
+									},
+								],
+								script: {
+									etag: "etag",
+									handlers: ["fetch"],
+									last_deployed_from: "api",
+								},
+								script_runtime: {
+									usage_model: "standard",
+									limits: {},
+								},
+							},
+						})
+					);
+				},
+				{ once: true }
+			)
+		);
+	}
+
+	function mockPostDeployment(forced = false) {
+		msw.use(
+			http.post(
+				`*/accounts/:accountId/workers/scripts/:scriptName/deployments${forced ? "?force=true" : ""}`,
+				async ({ params }) => {
+					expect(params.accountId).toEqual("some-account-id");
+					expect(params.scriptName).toEqual("script-name");
+
+					return HttpResponse.json(createFetchResult({}));
+				},
+				{ once: true }
+			)
+		);
+	}
+
+	test("can rollback to an earlier version", async () => {
+		mockGetDeployments();
+		mockGetVersion("version-id-1");
+		mockGetVersion("rollback-version");
+		mockPostDeployment();
+
+		mockPrompt({
+			text: "Please provide a message for this rollback (120 characters max, optional)?",
+			result: "Test rollback",
+		});
+
+		mockConfirm({
+			text: "Are you sure you want to deploy this Worker Version to 100% of traffic?",
+			result: true,
+		});
+
+		await runWrangler(
+			"rollback --name script-name --version-id rollback-version --x-versions"
+		);
+
+		// Unable to test stdout as the output has weird whitespace. Causing lint to fail with "no-irregular-whitespace"
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	test("rolling back with changed secrets prompts confirmation", async () => {
+		mockGetDeployments();
+		mockGetVersion("version-id-1");
+		mockGetVersion("rollback-version");
+
+		// Deployment will fail due to changed secret
+		msw.use(
+			http.post(
+				`*/accounts/:accountId/workers/scripts/:scriptName/deployments`,
+				async ({ params }) => {
+					expect(params.accountId).toEqual("some-account-id");
+					expect(params.scriptName).toEqual("script-name");
+
+					return HttpResponse.json(
+						createFetchResult(null, false, [
+							{
+								code: CANNOT_ROLLBACK_WITH_MODIFIED_SECERT_CODE,
+								message:
+									"A secret has changed since this version was active. If you are sure this is ok, add a ?force=true query parameter to allow the rollback. The following secrets have changed: SECRET, SECRET_TWO",
+							},
+						]),
+						{ status: 400 }
+					);
+				},
+				{ once: true }
+			)
+		);
+
+		// Now deploy again with force and succeed
+		mockPostDeployment(true);
+
+		mockPrompt({
+			text: "Please provide a message for this rollback (120 characters max, optional)?",
+			result: "Test rollback",
+		});
+
+		mockConfirm({
+			text: "Are you sure you want to deploy this Worker Version to 100% of traffic?",
+			result: true,
+		});
+
+		// We will have an additional confirmation
+		mockConfirm({
+			text:
+				"The following secrets have changed since the target version was deployed. Please confirm you wish to continue with the rollback" +
+				"\n  * SECRET\n  * SECRET_TWO",
+			result: true,
+		});
+
+		await runWrangler(
+			"rollback --name script-name --version-id rollback-version --x-versions"
+		);
+
+		// Unable to test stdout as the output has weird whitespace. Causing lint to fail with "no-irregular-whitespace"
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	test("rolling back with changed secrets (non-interactive)", async () => {
+		setIsTTY(false);
+		mockGetDeployments();
+		mockGetVersion("version-id-1");
+		mockGetVersion("rollback-version");
+
+		// Deployment will fail due to changed secret
+		msw.use(
+			http.post(
+				`*/accounts/:accountId/workers/scripts/:scriptName/deployments`,
+				async ({ params }) => {
+					expect(params.accountId).toEqual("some-account-id");
+					expect(params.scriptName).toEqual("script-name");
+
+					return HttpResponse.json(
+						createFetchResult(null, false, [
+							{
+								code: CANNOT_ROLLBACK_WITH_MODIFIED_SECERT_CODE,
+								message:
+									"A secret has changed since this version was active. If you are sure this is ok, add a ?force=true query parameter to allow the rollback. The following secrets have changed: SECRET, SECRET_TWO",
+							},
+						]),
+						{ status: 400 }
+					);
+				},
+				{ once: true }
+			)
+		);
+
+		// Now deploy again with force and succeed
+		mockPostDeployment(true);
+
+		await runWrangler(
+			"rollback --name script-name --version-id rollback-version --x-versions"
+		);
+
+		// Unable to test stdout as the output has weird whitespace. Causing lint to fail with "no-irregular-whitespace"
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+});
--- a/packages/wrangler/src/__tests__/versions/secrets/put.test.ts
+++ b/packages/wrangler/src/__tests__/versions/secrets/put.test.ts
@@ -1,0 +1,223 @@
+import { describe, expect, test } from "vitest";
+import { mockAccountId, mockApiToken } from "../../helpers/mock-account-id";
+import { mockConsoleMethods } from "../../helpers/mock-console";
+import { clearDialogs, mockPrompt } from "../../helpers/mock-dialogs";
+import { useMockIsTTY } from "../../helpers/mock-istty";
+import { useMockStdin } from "../../helpers/mock-stdin";
+import { runInTempDir } from "../../helpers/run-in-tmp";
+import { runWrangler } from "../../helpers/run-wrangler";
+import writeWranglerToml from "../../helpers/write-wrangler-toml";
+import { mockPostVersion, mockSetupApiCalls } from "./utils";
+
+describe("versions secret put", () => {
+	const std = mockConsoleMethods();
+	const { setIsTTY } = useMockIsTTY();
+	runInTempDir();
+	mockAccountId();
+	mockApiToken();
+	afterEach(() => {
+		clearDialogs();
+	});
+
+	test("can add a new secret (interactive)", async () => {
+		setIsTTY(true);
+
+		mockPrompt({
+			text: "Enter a secret value:",
+			options: { isSecret: true },
+			result: "the-secret",
+		});
+
+		mockSetupApiCalls();
+		mockPostVersion((metadata) => {
+			expect(metadata.bindings).toStrictEqual([
+				{ type: "secret_text", name: "NEW_SECRET", text: "the-secret" },
+			]);
+			expect(metadata.keep_bindings).toStrictEqual([
+				"secret_key",
+				"secret_text",
+			]);
+		});
+		await runWrangler(
+			"versions secret put NEW_SECRET --name script-name --x-versions"
+		);
+
+		expect(std.out).toMatchInlineSnapshot(`
+			"🌀 Creating the secret for the Worker \\"script-name\\"
+			✨ Success! Created version id with secret NEW_SECRET. To deploy this version with secret NEW_SECRET to production traffic use the command \\"wrangler versions deploy --x-versions\\"."
+		`);
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	// For some reason, this always hangs. Not sure why
+	test.skip("can add a new secret (non-interactive)", async () => {
+		setIsTTY(false);
+		const mockStdIn = useMockStdin({ isTTY: false });
+
+		mockSetupApiCalls();
+		mockPostVersion((metadata) => {
+			expect(metadata.bindings).toStrictEqual([
+				{ type: "secret_text", name: "NEW_SECRET", text: "the-secret" },
+			]);
+			expect(metadata.keep_bindings).toStrictEqual([
+				"secret_key",
+				"secret_text",
+			]);
+		});
+
+		mockStdIn.send(
+			`the`,
+			`-`,
+			`secret
+			` // whitespace & newline being removed
+		);
+		await runWrangler(
+			"versions secret put NEW_SECRET --name script-name --x-versions"
+		);
+
+		expect(std.out).toMatchInlineSnapshot(`
+			"🌀 Creating the secret for the Worker \\"script-name\\"
+			✨ Success! Created version id with secret NEW_SECRET.
+			➡️  To deploy this version with secret NEW_SECRET to production traffic use the command wrangler versions deploy."
+		`);
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	test("can add a new secret, read Worker name from wrangler.toml", async () => {
+		writeWranglerToml({ name: "script-name" });
+
+		setIsTTY(true);
+
+		mockPrompt({
+			text: "Enter a secret value:",
+			options: { isSecret: true },
+			result: "the-secret",
+		});
+
+		mockSetupApiCalls();
+		mockPostVersion((metadata) => {
+			expect(metadata.bindings).toStrictEqual([
+				{ type: "secret_text", name: "NEW_SECRET", text: "the-secret" },
+			]);
+			expect(metadata.keep_bindings).toStrictEqual([
+				"secret_key",
+				"secret_text",
+			]);
+		});
+		await runWrangler("versions secret put NEW_SECRET --x-versions");
+
+		expect(std.out).toMatchInlineSnapshot(`
+			"🌀 Creating the secret for the Worker \\"script-name\\"
+			✨ Success! Created version id with secret NEW_SECRET. To deploy this version with secret NEW_SECRET to production traffic use the command \\"wrangler versions deploy --x-versions\\"."
+		`);
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	test("can add a new secret with message", async () => {
+		setIsTTY(true);
+
+		mockPrompt({
+			text: "Enter a secret value:",
+			options: { isSecret: true },
+			result: "the-secret",
+		});
+
+		mockSetupApiCalls();
+		mockPostVersion((metadata) => {
+			expect(metadata.bindings).toStrictEqual([
+				{ type: "secret_text", name: "NEW_SECRET", text: "the-secret" },
+			]);
+			expect(metadata.keep_bindings).toStrictEqual([
+				"secret_key",
+				"secret_text",
+			]);
+
+			expect(metadata.annotations).not.toBeUndefined();
+			expect(
+				(metadata.annotations as Record<string, string>)["workers/message"]
+			).toBe("Deploy a new secret");
+		});
+		await runWrangler(
+			"versions secret put NEW_SECRET --name script-name --message 'Deploy a new secret' --x-versions"
+		);
+
+		expect(std.out).toMatchInlineSnapshot(`
+			"🌀 Creating the secret for the Worker \\"script-name\\"
+			✨ Success! Created version id with secret NEW_SECRET. To deploy this version with secret NEW_SECRET to production traffic use the command \\"wrangler versions deploy --x-versions\\"."
+		`);
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	test("can add a new secret with message + tag", async () => {
+		setIsTTY(true);
+
+		mockPrompt({
+			text: "Enter a secret value:",
+			options: { isSecret: true },
+			result: "the-secret",
+		});
+
+		mockSetupApiCalls();
+		mockPostVersion((metadata) => {
+			expect(metadata.bindings).toStrictEqual([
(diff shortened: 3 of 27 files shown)


tanushree-sharma:
Know you're still polishing up so ignore if these are on your to-do:

1. Only thing that should be updated / evaluated with the wrangler versions secret put command is secrets. Wrangler.toml warnings should not be shown
image
image
image

2. Copy when creating a new secret:

`npx wrangler versions secret put <KEY>
? Enter a secret value: › ***
🌀 Creating the secret for Worker worker-name
✨ Success! Created version XXX with secret FOO.
➡️ To deploy this version with secret FOO to production traffic use the command wrangler versions deploy.`

Deleting a secret should reflect similar copy.

Nice one with the automated versions message! Can we add one on delete of a secret too?
image
image

tanushree-sharma:
Looking great. Some more feedback:

1. Nit: Can we add a space between ➡️ and "To" 
image

2. Forgot the flag in the copy
`➡️ To deploy this version with secret new-secret to production traffic use the command wrangler versions deploy --x-versions.`

3. Not seeing the version message populated anymore (I did wrangler secret put)
image

4. `wrangler secret list` is showing warnings (similar to #1 in my message above)
image

5. Not seeing updated output for wrangler versions view. Also shows the warning. 
image

.changeset/little-rice-promise.md
```diff
@@ -0,0 +1,13 @@
...
+"wrangler": minor
+---
+
+feat: add `wrangler versions secret put`, `wrangler versions secret bulk` and `wrangler versions secret list`
+
+`wrangler versions secret put` allows for you to add/update a secret even if the latest version is not fully deployed. A new version with this secret will be created, the existing secrets and config are copied from the latest version.
+
+`wrangler versions secret bulk` allows you to bulk add/update multiple secrets at once, this behaves the same as `secret put` and will only make one new version.
+
+`wrangler versions secret list` lists the secrets available to the currently deployed versions. `wrangler secret list` will list for the latest version.
```
RamIdeas, line 11: I think this behaviour should be inverted. If `wrangler versions secret put/bulk` amends secrets from the _latest_ version, then `wrangler versions secret list` should list the secrets from the latest version.

For users not using gradual rollouts, `wrangler secret list` and `wrangler versions secret list` will output the secrets from the same (deployed) version. Therefore it's not a breaking change for `wrangler secret list` to list the secrets from the deployed version.

WalshyDev: Yeah agreed but we would break `wrangler secret list` behaviour today if we did make it output deployed versions (since right now it just dumps a JSON and we'd need to at least dump two JSONs + a header of what version it is)

Could do a `--latest` on `versions secret list for now`, these commands are still experimental and may change defaults later

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -13,31 +13,43 @@ import type {
...
+		throw new TypeError("Unsupported module: " + mimeType);
+	}
+
+	return moduleType;
 }
 
 export type WorkerMetadataBinding =
 	// If you add any new binding types here, also add it to safeBindings
 	// under validateUnsafeBinding in config/validation.ts
+	| { type: "inherit"; name: string }
```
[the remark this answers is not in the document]
WalshyDev: I decided against it just because it is a weird one to use properly in wrangler, it's much more for actual API usage. 
We have keep_bindings which handles this by type already in wrangler

It just needed to be added here so I didn't have type errors in the API calls

packages/wrangler/src/config/index.ts
```diff
@@ -112,7 +114,7 @@ export function readConfig<CommandArgs>(
 		args
 	);
 
-	if (diagnostics.hasWarnings()) {
+	if (diagnostics.hasWarnings() && hideWarnings !== true) {
```
CarmenPopoviciu, line 117: why are we doing this? and why only for warnings?

WalshyDev: For the secret commands, it doesn't make sense to warn on config validation. The config isn't actually being used except for the name (which would error if wrong).

Something being unsafe or in the wrong env doesn't impact the command itself and just creates clutter

packages/wrangler/src/secret/index.ts
```diff
@@ -282,11 +282,17 @@ export const secret = (secretYargs: CommonYargsArgv) => {
...
-					type: "string",
-					requiresArg: true,
-				});
+				return yargs
+					.option("name", {
+						describe: "Name of the Worker",
+						type: "string",
+						requiresArg: true,
+					})
+					.option("pretty", {
```
CarmenPopoviciu, line 291: instead of adding this as a cmd option, why not pretty print by default? Correct me if I'm wrong, but I'm assuming we're adding this so that users can make sense of whatever `JSON.stringify(await fetchResult(url), null, "  ")` will print? If we log things that are not easy to decipher, IMHO we either should not log those at all, or log them in a format that is helpful for our users. But maybe I'm missing context and am way off 😅 ?

WalshyDev: My thinking is that this would be a breaking change doing it by default, we could have users expecting it to be a JSON today and parsing it. Imo, not an invalid use case, means you could avoid pushing secrets that already exist

packages/wrangler/src/secret/index.ts
```diff
@@ -323,7 +339,7 @@ export const secret = (secretYargs: CommonYargsArgv) => {
  * Remove trailing white space from inputs.
  * Matching Wrangler legacy behavior with handling inputs
  */
-function trimTrailingWhitespace(str: string) {
+export function trimTrailingWhitespace(str: string) {
```
CarmenPopoviciu, line 342: a duplicate of this fn is declared [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/pages/secret/index.ts#L30-L32) as well. Should we consider moving the fn to  a utilities file, and import it in both wrangler proper and pages?

packages/wrangler/src/secret/index.ts
```diff
@@ -333,7 +349,7 @@ function trimTrailingWhitespace(str: string) {
  * This function can be used to grab the incoming stream of data from, say,
  * piping the output of another process into the wrangler process.
  */
-function readFromStdin(): Promise<string> {
+export function readFromStdin(): Promise<string> {
```
CarmenPopoviciu, line 352: same goes with this one 👉  https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/pages/secret/index.ts#L40-L65

packages/wrangler/src/versions/api.ts
```diff
@@ -104,28 +108,62 @@ export async function createDeployment(
...
+			const errorMsg = e.notes[0].text.replace(
+				` [code: ${CANNOT_ROLLBACK_WITH_MODIFIED_SECERT_CODE}]`,
+				""
+			);
+			const targetString = "The following secrets have changed:";
+			const changedSecrets = errorMsg
+				.substring(errorMsg.indexOf(targetString) + targetString.length + 1)
+				.split(", ");
+
+			const confirmed = await confirm(
```
CarmenPopoviciu, line 146: what happens if we are not in interactive mode? should we check for `isInteractive()`?

WalshyDev: It defaults and moves on - added a test for this :)

packages/wrangler/src/secret/index.ts
WalshyDev: We should, moved!

packages/wrangler/src/versions/api.ts
```diff
@@ -104,28 +108,62 @@ export async function createDeployment(
...
+			e instanceof APIError &&
+			e.code === CANNOT_ROLLBACK_WITH_MODIFIED_SECERT_CODE
+		) {
+			// This is not great but is the best way I could think to handle for now
+			const errorMsg = e.notes[0].text.replace(
+				` [code: ${CANNOT_ROLLBACK_WITH_MODIFIED_SECERT_CODE}]`,
+				""
+			);
+			const targetString = "The following secrets have changed:";
+			const changedSecrets = errorMsg
```
CarmenPopoviciu, line 142: are we guaranteed the message will have this format or that `e.notes[0].text` will always be returned?

WalshyDev: Yes, if this error code is returned. This error will always be this way.

This was the best way I could think to pass the secret list down to Wrangler while also giving a helpful error to users so it's a bit jank but does work

packages/wrangler/src/config/index.ts
CarmenPopoviciu: I see. Let me think abt this a bit more 🤔

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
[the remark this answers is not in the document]
CarmenPopoviciu: can we flag that with a comment then pls 🙏 so as to avoid any future confusions

packages/wrangler/src/secret/index.ts
CarmenPopoviciu: what prompted us to add `--pretty`? what's the use case?

packages/wrangler/src/versions/index.ts
```diff
@@ -254,5 +257,12 @@ export default function registerVersionsSubcommands(
 			"Safely roll out new Versions of your Worker by splitting traffic between multiple Versions [beta]",
 			versionsDeployOptions,
 			versionsDeployHandler
+		)
+		.command(
+			"secret",
+			"🤫 Generate a secret that can be referenced in a Worker",
```
CarmenPopoviciu, line 263:
```suggestion
			"Generate a secret that can be referenced in a Worker",
```

CarmenPopoviciu: because we're not doing emojis for sub-commands anymore, as per https://github.com/cloudflare/workers-sdk/pull/3735

packages/wrangler/src/secret/index.ts
WalshyDev: Creates a better view for people, dumping JSON out isn't great. Moving the default view in the next major version to the pretty one is the goal. Just wasn't done to avoid breaking changes here.

packages/wrangler/src/versions/rollback/index.ts
```diff
@@ -78,14 +79,12 @@ export async function versionsRollbackHandler(args: VersionsRollbackArgs) {
...
-	const message = await inputPrompt({
-		type: "text",
-		label: "Message",
-		question:
-			"Please provide a message for this rollback (120 characters max, optional)?",
-		defaultValue: args.message ?? "Rollback",
-		acceptDefault: args.yes,
-	});
+	const message = await prompt(
+		"Please provide a message for this rollback (120 characters max, optional)?",
```
CarmenPopoviciu, line 83:
```suggestion
		"Please provide an optional message for this rollback (120 characters max)?",
```

CarmenPopoviciu: maybe? up to you

packages/wrangler/src/versions/rollback/index.ts
```diff
@@ -95,16 +94,11 @@ export async function versionsRollbackHandler(args: VersionsRollbackArgs) {
...
-		defaultValue: args.yes, // defaultValue: false.    if --yes, defaultValue: true
-		acceptDefault: args.yes,
-	});
-
-	if (!confirm) {
+	const confirmed = await confirm(
+		"Are you sure you want to deploy this Worker Version to 100% of traffic?",
+		{ defaultValue: true }
+	);
+	if (!confirmed) {
```
CarmenPopoviciu, line 101: I noticed that in https://github.com/cloudflare/workers-sdk/pull/5950/files#diff-7b1535cea8d8067fdfcbc158145aa6ea986e82b46d3db6eb89feb254844f1175R161 we're throwing and I wondered if we wouldn't want to do smth similar to here, meaning `cli.cancel("....")`

WalshyDev: The cli cancel won't stop the rest of the execution which means it'll print "Aborting rollback...\nSuccess! Worker Version  ... has been deployed ..."

Because we'd be returning in createDeployment but not exiting the whole rollback execution flow

I tried that first :(

packages/wrangler/src/secret/index.ts
CarmenPopoviciu: right, thx for the context. Totally up to you, but maybe for consistency purposes, would it make sense to use a similar option as we do for [tailing](url)? That's really the only place, AFAICT, where we're giving a choice between `json` and `pretty` output. So smth like 👇  as per https://developers.cloudflare.com/workers/wrangler/commands/#tail
Screenshot 2024-06-20 at 13 33 56

packages/wrangler/src/versions/rollback/index.ts
CarmenPopoviciu: kk 👍

packages/wrangler/src/versions/secrets/bulk.ts
```diff
@@ -0,0 +1,128 @@
...
+			describe: `The JSON file of key-value pairs to upload, in form {"key": value, ...}`,
+			type: "string",
+		})
+		.option("name", {
+			describe: "Name of the Worker",
+			type: "string",
+			requiresArg: true,
+		})
+		.option("message", {
+			describe: "Description of this deployment (optional)",
```
CarmenPopoviciu, line 31: Generally we don't seem to be marking `optional` options (:meta: I know) this way. Is there a particular reason we want to postfix these descriptions with `(optional)`? Lack of docs perhaps?

packages/wrangler/src/versions/secrets/bulk.ts
```diff
@@ -0,0 +1,128 @@
...
+	let content: Record<string, string>;
+	if (args.json) {
+		const jsonFilePath = path.resolve(args.json);
+		try {
+			content = parseJSON<Record<string, string>>(
+				readFileSync(jsonFilePath),
+				jsonFilePath
+			);
+		} catch (e) {
+			return logger.error(`🚨 Please provide a JSON file or valid JSON pipe`);
```
CarmenPopoviciu, line 72:
```suggestion
			return logger.error(`Please provide a JSON file or valid JSON pipe`);
```

CarmenPopoviciu: because I think the big, red error label will be enough 😅 . Also because https://github.com/cloudflare/workers-sdk/blob/main/STYLEGUIDE.md#general-error

CarmenPopoviciu: + could we pls give a short & concise description of what the error is? As per same styleguide

packages/wrangler/src/versions/secrets/bulk.ts
```diff
@@ -0,0 +1,128 @@
...
+	} else {
+		try {
+			const rl = readline.createInterface({ input: process.stdin });
+			let pipedInput = "";
+			for await (const line of rl) {
+				pipedInput += line;
+			}
+			content = parseJSON<Record<string, string>>(pipedInput);
+		} catch {
+			return logger.error(`🚨 Please provide a JSON file or valid JSON pipe`);
```
CarmenPopoviciu, line 84:
```suggestion
			return logger.error(`Please provide a JSON file or valid JSON pipe`);
```

CarmenPopoviciu: and same comment as above

packages/wrangler/src/versions/secrets/bulk.ts
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	if (!content) {
+		return logger.error(`🚨 No content found in JSON file or piped input.`);
```
CarmenPopoviciu, line 89:
```suggestion
		return logger.error(`No content found in JSON file or piped input.`);
```

packages/wrangler/src/versions/secrets/bulk.ts
```diff
@@ -0,0 +1,128 @@
...
+		versionMessage: args.message,
+		versionTag: args.tag,
+		sendMetrics: config.send_metrics,
+	});
+
+	for (const secret of secrets) {
+		logger.log(`✨ Successfully created secret for key: ${secret.name}`);
+	}
+	logger.log(
+		"\nFinished processing secrets JSON file:" +
```
CarmenPopoviciu, line 124: is there any point, in all these log/warn/error messages in this file, at which it would be worth mentioning the JSON file path? Just a nit, not a must, if it's even valuable to begin with

packages/wrangler/src/versions/secrets/bulk.ts
WalshyDev: No idea, I was just following how we've been doing it already for secret stuff haha

packages/wrangler/src/versions/secrets/bulk.ts
```diff
@@ -0,0 +1,128 @@
...
+) {
+	await printWranglerBanner();
+	const config = readConfig(args.config, args, false, true);
+
+	const scriptName = getLegacyScriptName(args, config);
+	if (!scriptName) {
+		const error = new UserError(
+			"Required Worker name missing. Please specify the Worker name in wrangler.toml, or pass it as an argument with `--name <worker-name>`"
+		);
+		logger.error(error.message);
```
CarmenPopoviciu, line 53: I'm not sure we need to both log **and** throw the error. I think throwing should be enough?

packages/wrangler/src/versions/secrets/bulk.ts
WalshyDev: Well the user is providing the path so probably not too much value in showing it again. Not sure where we'd add it either tbh

packages/wrangler/src/versions/secrets/bulk.ts
WalshyDev: ha yep indeed, this is weird. This is a copy-pasta from the regular secret bulk

CarmenPopoviciu: I **think**
```
throw new FatalError("Required Worker name missing. Please specify the Worker name in wrangler.toml, or pass it as an argument with `--name <worker-name>`")
```
should do it

packages/wrangler/src/versions/secrets/delete.ts
```diff
@@ -0,0 +1,119 @@
...
+			describe: "The variable name to be accessible in the Worker",
+			type: "string",
+		})
+		.option("name", {
+			describe: "Name of the Worker",
+			type: "string",
+			requiresArg: true,
+		})
+		.option("message", {
+			describe: "Description of this deployment (optional)",
```
CarmenPopoviciu, line 28: [same](https://github.com/cloudflare/workers-sdk/pull/5950/files#r1647441543) comment

packages/wrangler/src/versions/secrets/bulk.ts
WalshyDev: FatalError would try and log to Sentry, UserError should be enough no?

packages/wrangler/src/versions/secrets/delete.ts
```diff
@@ -0,0 +1,119 @@
...
+	const scriptName = getLegacyScriptName(args, config);
+	if (!scriptName) {
+		throw new UserError(
+			"Required Worker name missing. Please specify the Worker name in wrangler.toml, or pass it as an argument with `--name <worker-name>`"
+		);
+	}
+
+	if (args.key === undefined) {
+		throw new UserError(
+			"Secret name is required, please specify the name of your secret."
```
CarmenPopoviciu, line 54:
```suggestion
			"Secret name is required. Please specify the name of your secret."
```

packages/wrangler/src/secret/index.ts
WalshyDev: yep, agreed, done

packages/wrangler/src/versions/api.ts
WalshyDev: yeah not sure it is :/
to me that's less clear

packages/wrangler/src/versions/secrets/index.ts
```diff
@@ -0,0 +1,332 @@
...
+	scriptName,
+	versionId,
+	secrets,
+	versionMessage,
+	versionTag,
+	sendMetrics,
+	overrideAllSecrets,
+}: CopyLatestWorkerVersionArgs) {
+	// Grab the specific version info
+	const versionInfo = await fetchResult<VersionDetails>(
```
CarmenPopoviciu, line 132: do we need to handle the case when this API returns an error?

packages/wrangler/src/versions/secrets/list.ts
```diff
@@ -0,0 +1,94 @@
...
+		);
+
+		const secrets = (version as VersionDetails).resources.bindings.filter(
+			(binding) => binding.type === "secret_text"
+		);
+		for (const secret of secrets) {
+			logger.log(`Secret Name: ${secret.name}`);
+		}
+
+		logger.log();
```
CarmenPopoviciu, line 92: no params to `log()`?

packages/wrangler/src/versions/secrets/put.ts
```diff
@@ -0,0 +1,99 @@
...
+	const scriptName = getLegacyScriptName(args, config);
+	if (!scriptName) {
+		throw new UserError(
+			"Required Worker name missing. Please specify the Worker name in wrangler.toml, or pass it as an argument with `--name <worker-name>`"
+		);
+	}
+
+	if (args.key === undefined) {
+		throw new UserError(
+			"Secret name is required, please specify the name of your secret."
```
CarmenPopoviciu, line 55:
```suggestion
			"Secret name is required. Please specify the name of your secret."
```

packages/wrangler/src/versions/secrets/put.ts
```diff
@@ -0,0 +1,99 @@
...
+
+	// Grab the latest version
+	const versions = (
+		await fetchResult<{ items: WorkerVersion[] }>(
+			`/accounts/${accountId}/workers/scripts/${scriptName}/versions`
+		)
+	).items;
+	if (versions.length === 0) {
+		throw new UserError(
+			"There are currently no uploaded versions of this Worker - please upload a version before uploading a secret."
```
CarmenPopoviciu, line 80:
```suggestion
			"There are currently no uploaded versions of this Worker. Please upload a version before uploading a secret."
```

packages/wrangler/src/versions/secrets/list.ts
WalshyDev: Nah just ensures there's an empty line between versions

packages/wrangler/src/versions/secrets/put.ts
```diff
@@ -0,0 +1,99 @@
...
+		versionId: latestVersion.id,
+		secrets: [{ name: args.key, value: secretValue }],
+		versionMessage: args.message,
+		versionTag: args.tag,
+		sendMetrics: config.send_metrics,
+	});
+
+	logger.log(
+		`✨ Success! Created version ${newVersion.id} with secret ${args.key}.\n` +
+			`➡️  To deploy this version with secret ${args.key} to production traffic use the command wrangler versions deploy --x-versions.`
```
CarmenPopoviciu, line 97:
```suggestion
			`➡️  To deploy this version with secret ${args.key} to production traffic use the command "wrangler versions deploy --x-versions".`
```

packages/wrangler/src/versions/secrets/index.ts
WalshyDev: It shouldn't ever generate an error, we already know the version exists. If it does, I'm happy treating that like all usual fatalerrors and having it reported to us

packages/wrangler/src/__tests__/rollback.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+		await runWrangler(
+			"rollback --name script-name --version-id rollback-version --x-versions"
+		);
+
+		// Unable to test stdout as the output has weird whitespace. Causing lint to fail with "no-irregular-whitespace"
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	test("rolling back with changed secrets (non-interactive)", async () => {
```
CarmenPopoviciu, line 196: <3

packages/wrangler/src/versions/secrets/bulk.ts
```diff
@@ -0,0 +1,130 @@
...
+		secrets,
+		versionMessage: args.message,
+		versionTag: args.tag,
+		sendMetrics: config.send_metrics,
+	});
+
+	for (const secret of secrets) {
+		logger.log(`✨ Successfully created secret for key: ${secret.name}`);
+	}
+	logger.log(
```
CarmenPopoviciu, line 125:
```suggestion
	logger.log(\n✨ Success! Created version ${newVersion.id} with ${secrets.length} secrets` +
	"\nTo deploy this version to production traffic use the command `wrangler versions deploy --x-versions`."
```

packages/wrangler/src/versions/secrets/delete.ts
```diff
@@ -0,0 +1,119 @@
...
+			secrets: newSecrets,
+			versionMessage: args.message,
+			versionTag: args.tag,
+			sendMetrics: config.send_metrics,
+			overrideAllSecrets: true,
+		});
+
+		logger.log(
+			`✨ Success! Created version ${newVersion.id} with deleted secret ${args.key}.\n` +
+				`➡️  To deploy this version without the secret ${args.key} to production traffic use the command wrangler versions deploy --x-versions.`
```
CarmenPopoviciu, line 116:
```suggestion
				"To deploy this version without the secret ${args.key} to production traffic use the command `wrangler versions deploy --x-versions`."
```

packages/wrangler/src/versions/secrets/put.ts
```diff
@@ -0,0 +1,99 @@
...
+		versionId: latestVersion.id,
+		secrets: [{ name: args.key, value: secretValue }],
+		versionMessage: args.message,
+		versionTag: args.tag,
+		sendMetrics: config.send_metrics,
+	});
+
+	logger.log(
+		`✨ Success! Created version ${newVersion.id} with secret ${args.key}.\n` +
+			`➡️  To deploy this version with secret ${args.key} to production traffic use the command "wrangler versions deploy --x-versions".`
```
CarmenPopoviciu, line 97:
```suggestion
			`To deploy this version with secret ${args.key} to production traffic use the command "wrangler versions deploy --x-versions".`
```

packages/wrangler/src/__tests__/versions/secrets/put.test.ts
```diff
@@ -0,0 +1,228 @@
...
+		expect(std.out).toMatchInlineSnapshot(`
+			"🌀 Creating the secret for the Worker \\"script-name\\"
+			✨ Success! Created version id with secret NEW_SECRET.
+			➡️  To deploy this version with secret NEW_SECRET to production traffic use the command \\"wrangler versions deploy --x-versions\\"."
+		`);
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	// For some reason, this always hangs. Not sure why
+	test.skip("can add a new secret (non-interactive)", async () => {
```
CarmenPopoviciu, line 54: not sure if you have time to fix this test today, but if not, let's pls test manually that this works...if you haven't done that already...and maybe take fixing the test as a afast-follow?

WalshyDev: yeah this has been tested manually, fixing will be a fast follow
I think some prompt or something isn't being mocked fully - i just wish i knew which haha

packages/wrangler/src/versions/secrets/delete.ts
```diff
@@ -0,0 +1,119 @@
...
+			scriptName,
+			versionId: latestVersion.id,
+			secrets: newSecrets,
+			versionMessage: args.message,
+			versionTag: args.tag,
+			sendMetrics: config.send_metrics,
+			overrideAllSecrets: true,
+		});
+
+		logger.log(
```
CarmenPopoviciu, line 114:
```suggestion
		logger.log(`✨ Success! Created version ${newVersion.id} with deleted secret ${args.key}. To deploy this version without the secret ${args.key} to production traffic use the command "wrangler versions deploy --x-versions".`);
```

packages/wrangler/src/versions/secrets/put.ts
```diff
@@ -0,0 +1,99 @@
...
+		accountId,
+		scriptName,
+		versionId: latestVersion.id,
+		secrets: [{ name: args.key, value: secretValue }],
+		versionMessage: args.message,
+		versionTag: args.tag,
+		sendMetrics: config.send_metrics,
+	});
+
+	logger.log(
```
CarmenPopoviciu, line 95:
```suggestion
	logger.log(`✨ Success! Created version ${newVersion.id} with secret ${args.key}. To deploy this version with secret ${args.key} to production traffic use the command "wrangler versions deploy --x-versions".`);
```

packages/wrangler/src/versions/secrets/bulk.ts
CarmenPopoviciu: oh yeah, you're right 👍

packages/wrangler/src/versions/secrets/bulk.ts
CarmenPopoviciu: fair!

packages/wrangler/src/config/index.ts
CarmenPopoviciu: I spoke with Walshy directly, but for posterity

1. Not showing these warnings is a **prod** requirement (see [here](https://github.com/cloudflare/workers-sdk/pull/5950#issuecomment-2151047650)), so we can't go around that
2. an alternative to passing this `hideWarnings` flag, would be for the command to read the `rawConfig` from the config file, validate the Worker name, and decide what it needs to do further based on that validation. Since that's quite cumbersome, and basically would duplicate code we already have in `readConfig`, perhaps having this flag is the way to go

packages/wrangler/src/config/index.ts
```diff
@@ -41,13 +41,15 @@ export function readConfig<CommandArgs>(
...
+	requirePagesConfig?: boolean,
+	hideWarnings?: boolean
 ): Config;
 export function readConfig<CommandArgs>(
 	configPath: string | undefined,
 	// Include command specific args as well as the wrangler global flags
 	args: ReadConfigCommandArgs<CommandArgs>,
-	requirePagesConfig?: boolean
+	requirePagesConfig?: boolean,
+	hideWarnings?: boolean
```
CarmenPopoviciu, line 52:
```suggestion
	hideWarnings = false
```

packages/wrangler/src/config/index.ts
CarmenPopoviciu, line 117:
```suggestion
	if (diagnostics.hasWarnings() && !hideWarnings) {
```

CarmenPopoviciu (approved):
thx Walshy :meouw-heart:

tanushree-sharma:
Know this has already been merged but I have feedback:

**1. Rollback guardrail: If we detect that a user is rolling back to a version that has a different value for a secret, we should warn the user but give them the option to proceed anyway:**

I set up a case to hit the guardrail and wrangler is getting stuck here:
image

It isn't until I cancel the operation that I see the warning
image

Also, this is what I'm seeing in dash: 
image

We need a popup window rather than a banner. 

**2. I want to make the next step after wrangler versions secret put/delete clear to users and the best way to do that is to have “To deploy..” on the next line**

Instead of:
image

This:
image

3. Can we add back the automatic version message when a secret has been updated? Currently the version message is blank 

4. Reminder to implement `wrangler versions view` updates

PR #5955, opened 2024-06-02
Fix: Correctly escape newlines in `constructType` function

### LINKED ISSUE #5699: 🐛 BUG: wrangler types generates invalid JS for multi-line strings

Wrangler core

3.51.2

v21.7.1

Mac Sonoma 14.4.1

### Observed behavior

When using a multi-line string as environment variable, `wrangler types` generates invalid JS.

### Expected behavior

`wrangler types` should generate valid JS using either \` or by inserting \n.

### Steps to reproduce
- Add a multi-line string: https://github.com/Le0Developer/wrangler-types-multiline/blob/master/wrangler.toml#L13
- Run `wrangler types`
- Observe invalid JS: https://github.com/Le0Developer/wrangler-types-multiline/blob/master/worker-configuration.d.ts#L5-L6

https://github.com/Le0Developer/wrangler-types-multiline

_No response_

harugon:

Fixes #5699

This PR fixes the issue with `constructType` function handling multiline strings improperly. It ensures that newlines in strings are correctly escaped, preventing invalid JavaScript code generation. 

- Tests
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: This change affects type generation only
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Internal functionality change only


packages/wrangler/src/type-generation.ts
```diff
@@ -144,7 +144,7 @@ export function constructType(
 		if (useRawVal) {
 			return `${typeKey}: ${value};`;
 		}
-		return `${typeKey}: "${value.replace(/"/g, '\\"')}";`;
+		return `${typeKey}: "${value.replace(/"/g, '\\"').replace(/\n/g, '\\n')}";`;
```
penalosa, line 147: I wonder if we'd be able to get away with something simpler here—maybe `JSON.stringify()` of the vlaue?

harugon: Thank you for the suggestion! I've updated the constructType function to use JSON.stringify() for escaping string values. Please review the changes and let me know if there are any further improvements needed.

packages/wrangler/src/type-generation.ts
```diff
@@ -144,7 +144,7 @@ export function constructType(
 		if (useRawVal) {
 			return `${typeKey}: ${value};`;
 		}
-		return `${typeKey}: "${value.replace(/"/g, '\\"')}";`;
+		return `${typeKey}: "${JSON.stringify(value)}";`;
```
penalosa, line 147: I don't think this needs wrapped in quotation marks?
```suggestion
		return `${typeKey}: ${JSON.stringify(value)};`;
```

petebacondarwin (approved):

petebacondarwin:
After running the tests, I noticed that the original backslash escaping tests were not correct either. So this fix actually fixes that too.

PR #5957, opened 2024-06-03
Bundle controller

penalosa:

Fixes [DEVX-860](https://jira.cfdata.org/browse/DEVX-860)

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: not user-facing


packages/wrangler/src/routes.ts
```diff
@@ -24,7 +24,8 @@ export async function getWorkersDevSubdomain(
 			);
 
 			const wantsToRegister = await confirm(
-				"Would you like to register a workers.dev subdomain now?"
+				"Would you like to register a workers.dev subdomain now?",
+				{ fallbackValue: false }
```
penalosa, line 28: This is unrelated to the bundle controller—just a fix to make this default to false in non-interactive mode

packages/wrangler/src/__tests__/api/startDevWorker/BundleController.test.ts
```diff
@@ -0,0 +1,522 @@
...
+                export default {
+                    fetch(request, env, ctx) {
+                        //comment
+                        return new Response("hello world" + name)
+                    }
+                } satisfies ExportedHandler
+            `,
+			"src/other.ts": dedent/* javascript */ `
+                export default "someone"
+            `,
```
RamIdeas, line 155: nit the indentation here is using spaces

packages/wrangler/src/user/user.ts
```diff
@@ -1120,12 +1120,7 @@ export function listScopes(message = "💁 Available scopes:"): void {
 	// TODO: maybe a good idea to show usage here
 }
 
-export async function getAccountId(): Promise<string | undefined> {
-	const apiToken = getAPIToken();
-	if (!apiToken) {
-		return;
-	}
-
+export async function getAccountId(): Promise<string> {
```
RamIdeas, line 1123: This seems to now be used in the React flow. Does `await select("Select an account",` work with react/ink?

penalosa: This is only used in the React flow if `--x-dev-env` is enabled, so it shouldn't break any existing users. It slightly changes the output for people with the experimental flag turned on, but the functionality is preserved.

This is the current prompt output for remote mode:
Screenshot 2024-06-07 at 16 12 50

This is the current final output for remote mode:
Screenshot 2024-06-07 at 16 11 10

And this is the new prompt output:
Screenshot 2024-06-07 at 16 14 35

And the new final output:
Screenshot 2024-06-07 at 16 15 23

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -1,37 +1,280 @@
...
+				},
+			});
+		} catch (err) {
+			logger.error("Custom build failed:", err);
+			this.emitErrorEvent({
+				type: "error",
+				reason: "Custom build failed",
+				cause: castErrorCause(err),
+				source: "BundlerController",
+				data: undefined,
```
RamIdeas, line 155: I think I expected this to be used for the inputs that caused the step to run – on error it will help dry-run the function if the inputs are known

```suggestion
				data: { config, filePath },
```

packages/wrangler/src/user/user.ts
penalosa: The reason I made this change was that they rendering of the account list happened in the `Remote` component, which is no longer rendered

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -1,37 +1,280 @@
...
+					source: "BundlerController",
+					data: undefined,
+				})
+		);
+	}
+
+	#tmpDir?: EphemeralDirectory;
+
+	async onConfigUpdate(event: ConfigUpdateEvent) {
+		await this.teardown();
```
RamIdeas, line 244: This isn't how teardown should be used – ideally it should only be called when the dev env is not expected to be used anymore. And this also causes onConfigUpdate to become async.

Ideally, each step would manage itself more granularly. This allows only the parts that need to be async to be async and reduce unnecessary async boundaries

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -1,37 +1,280 @@
...
+			if (buildAborter.signal.aborted) {
+				return;
+			}
+			assert(this.#tmpDir);
+			assert(config._entry, "config._entry");
+			assert(config._additionalModules, "config._additionalModules");
+			assert(config.build?.moduleRules, "config.build?.moduleRules");
+			assert(config.build?.define, "config.build?.define");
+			if (!config.build?.bundle) {
+				// if we're not building, let's just copy the entry to the destination directory
```
petebacondarwin, line 72:
```suggestion
				// if we're not bundling, let's just copy the entry to the destination directory
```

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -1,37 +1,280 @@
...
+			}
+
+			const entryDirectory = path.dirname(config._entry.file);
+			const moduleCollector = createModuleCollector({
+				wrangler1xLegacyModuleReferences: getWrangler1xLegacyModuleReferences(
+					entryDirectory,
+					config._entry.file
+				),
+				entry: config._entry,
+				// `moduleCollector` doesn't get used when `props.noBundle` is set, so
```
petebacondarwin, line 87: There is no `props`:

```suggestion
				// `moduleCollector` doesn't get used when `noBundle` is set, so
```

packages/wrangler/src/api/startDevWorker/BundlerController.ts
penalosa: That's fair—do you have a suggestion for how this could/should be done differently? The things that need done on every config update are the same as those that need done when doing a teardown, and unfortunately some of those are async (closing the esbuild and filesystem watchers)

petebacondarwin (commented):
Interim comments.

petebacondarwin:
In terms of review, I can't see anything glaring that @RamIdeas will not pick up or has not already picked up.

packages/wrangler/src/dev/dev.tsx
```diff
@@ -458,24 +553,30 @@ function DevSession(props: DevSessionProps) {
...
+			});
+		}
+	}, [devEnv, startDevWorkerOptions, props.experimentalDevEnv]);
 
+	// This value is static, so doesn't cause the below hooks to be called conditionally
+	if (props.experimentalDevEnv) {
+		return null;
+	}
+	// eslint-disable-next-line react-hooks/rules-of-hooks
+	useCustomBuild(props.entry, props.build, onBundleStart, onCustomBuildEnd);
```
RamIdeas, line 577:
```suggestion
if (!props.experimentalDevEnv) {
	// eslint-disable-next-line react-hooks/rules-of-hooks
	useCustomBuild(props.entry, props.build, onBundleStart, onCustomBuildEnd);
}
```

packages/wrangler/src/dev/dev.tsx
```diff
@@ -458,24 +553,30 @@ function DevSession(props: DevSessionProps) {
...
+				type: "configUpdate",
+				config: startDevWorkerOptions,
+			});
+		}
+	}, [devEnv, startDevWorkerOptions, props.experimentalDevEnv]);
 
+	// This value is static, so doesn't cause the below hooks to be called conditionally
+	if (props.experimentalDevEnv) {
+		return null;
+	}
```
RamIdeas, line 575: In the interest of derisking, we may want to be more granular with the if-block. See my comments below

packages/wrangler/src/dev/dev.tsx
```diff
@@ -458,24 +553,30 @@ function DevSession(props: DevSessionProps) {
...
+	// eslint-disable-next-line react-hooks/rules-of-hooks
+	useCustomBuild(props.entry, props.build, onBundleStart, onCustomBuildEnd);
+
+	// eslint-disable-next-line react-hooks/rules-of-hooks
 	const bundle = useEsbuild({
```
RamIdeas, line 580:
```suggestion
	const bundle = props.experimentalDevEnv ? undefined : useEsbuild({
```

packages/wrangler/src/dev/dev.tsx
```diff
@@ -350,8 +424,9 @@ function DevSession(props: DevSessionProps) {
 					: undefined,
 			dev: {
 				auth: async () => {
+					const selectedAccountId = await getAccountId();
 					return {
-						accountId: await accountIdDeferred.promise,
+						accountId: selectedAccountId,
```
RamIdeas, line 429: With my other suggestions, we would still be able to keep the rendering of the components Local and Remote, and therefore still keep the same auth flow.

packages/wrangler/src/user/user.ts
RamIdeas: See my other suggestions for how we could keep that component rendered

packages/wrangler/src/dev/dev.tsx
RamIdeas: The already of the experimentalDevEnv props passed through so can individually return null if needed

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -1,37 +1,280 @@
...
+		await this.#bundlerCleanup?.();
+		this.#customBuildAborter?.abort();
+		this.#tmpDir?.remove();
+	}
 
 	emitBundleStartEvent(data: BundleStartEvent) {
 		this.emit("bundleStart", data);
 	}
-	emitBundleCompletetEvent(data: BundleCompleteEvent) {
+	emitBundleCompletedEvent(data: BundleCompleteEvent) {
```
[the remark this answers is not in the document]
RamIdeas: nit
```suggestion
	emitBundleCompleteEvent(config: Config, bundle: Bundle) {
		this emit("bundleStart", { config, bundle });
```

RamIdeas (approved):
Most comments left are subjective, or at least there is justification for the current approach. As discussed offline this is looking mostly solid. A few changes to make but Samuel will decide the extent to which are required now and which will come later. LGTM.

packages/wrangler/src/dev/dev.tsx
```diff
@@ -297,12 +301,57 @@ function DevSession(props: DevSessionProps) {
...
 	}, []);
 
-	const [devEnv] = useState(() => new DevEnv());
+	const [devEnv] = useState(() => {
+		if (!props.experimentalDevEnv) {
+			return new DevEnv();
+		}
+		const proxy = new ProxyController();
+		const local = new LocalRuntimeController();
+		const remote = new RemoteRuntimeController();
```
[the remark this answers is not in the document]
penalosa: What's the benefit of doing it this way vs explicitly constructing the controllers?

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -1,37 +1,280 @@
...
+						minify: config.build.minify,
+						nodejsCompatMode: config.build.nodejsCompatMode,
+						define: config.build.define,
+						checkFetch: true,
+						assets: config._assets,
+						// enable the cache when publishing
+						bypassAssetCache: false,
+						// We want to know if the build is for development or publishing
+						// This could potentially cause issues as we no longer have identical behaviour between dev and deploy?
+						targetConsumer: "dev",
```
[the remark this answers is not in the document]
penalosa: It's also used to decide when to apply middleware, which is I suppose slightly more complicated

packages/wrangler/src/dev/dev.tsx
penalosa: Given the long term goal is to move away from React, is this definitely a direction we want to go in? It feels like starting to move away from React with things like rendering the account choices list is the right direction, but I'm happy to defer that to a later stage if you think that's too risky

JAYCRIBZ04:
> ## What this PR solves / how to test
> Fixes [DEVX-860](https://jira.cfdata.org/browse/DEVX-860)
> 
> ## Author has addressed the following
> * Tests
>   
>   * [x]  TODO (before merge)
>   * [ ]  Included
>   * [ ]  Not necessary because:
> * E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
>   
>   * [ ]  I don't know
>   * [x]  Required / Maybe required
>   * [ ]  Not required because:
> * Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
>   
>   * [x]  TODO (before merge)
>   * [ ]  Included
>   * [ ]  Not necessary because:
> * Public documentation
>   
>   * [ ]  TODO (before merge)
>   * [ ]  Cloudflare docs PR(s):
>   * [x]  Not necessary because: not user-facing

gh pr checkout 5957

RamIdeas (approved):

Issue #4269, opened 2023-10-24
🐛 BUG: Wrangler is injecting development code in production builds


Skye-31:
### Which Cloudflare product(s) does this pertain to?

Wrangler core

### What version(s) of the tool(s) are you using?

3.14.0

### What version of Node are you using?

v18.17.0

### What operating system are you using?

Mac

### Describe the Bug

Wrangler has started injecting some sort of "modules watch" code into production builds, when it shouldn't be. These functions are getting called for every import that is made in your worker script, which has an increase in both script size of ~5kb, and script startup time.

```
// wrangler-modules-watch:wrangler:modules-watch
var init_wrangler_modules_watch = __esm({
  "wrangler-modules-watch:wrangler:modules-watch"() {
    init_modules_watch_stub();
  }
});

// node_modules/.pnpm/wrangler@3.14.0/node_modules/wrangler/templates/modules-watch-stub.js
var init_modules_watch_stub = __esm({
  "node_modules/.pnpm/wrangler@3.14.0/node_modules/wrangler/templates/modules-watch-stub.js"() {
    init_wrangler_modules_watch();
  }
});
```

This behaviour started occurring in 3.11.0, and didn't occur in 3.10.0, so I'm almost certain it's the result of https://github.com/cloudflare/workers-sdk/pull/3726, cc @petebacondarwin 

### Please provide a link to a minimal reproduction

Any worker that imports anything -> `npx wrangler deploy --outdir dist`, look at the outputted code to see these files which shouldn't be present

petebacondarwin:
It appears this code is only added if esbuild is having to deal with a Common JS import. If all the code in a Worker is ESM then this virtual module gets stripped from the output.

PR #5960, opened 2024-06-03
fix: avoid injecting esbuild watch stubs into production Worker code

petebacondarwin:

When we added the ability to include additional modules in the deployed bundle of a Worker,
we inadvertently also included some boiler plate code that is only needed at development time.

This fix ensures that this code is only injected if we are running esbuild in watch mode
(e.g. `wrangler dev`) and not when building for deployment.

It is interesting to note that this boilerplate only gets included in the production code
if there is an import of CommonJS code in the Worker, which esbuild needs to convert to an
ESM import.

Fixes [#4269](https://github.com/cloudflare/workers-sdk/issues/4269)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -2601,7 +2601,18 @@ addEventListener('fetch', event => {});`
 				find_additional_modules: true,
 				rules: [{ type: "ESModule", globs: ["**/*.mjs"] }],
 			});
-			writeWorkerSource({ type: "esm" });
+			// Create a Worker that imports a CommonJS module to trigger esbuild to add
+			// extra boilerplate to convert to ESM imports.
+			fs.writeFileSync(`another.cjs`, `module.exports.foo = 100;`);
+			fs.writeFileSync(
+				`index.js`,
+				`import { foo } from "./another.cjs";
+					export default {
+						async fetch(request) {
+							return new Response('Hello' + foo);
+						},
+					};`
+			);
 			fs.mkdirSync("a/b/c", { recursive: true });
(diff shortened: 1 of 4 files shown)


CarmenPopoviciu (approved):

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2621,6 +2632,11 @@ addEventListener('fetch', event => {});`
 			);
 			writeAssets(assets);
 			mockUploadWorkerRequest({
+				expectedEntry(entry) {
+					// Ensure that we have not included the watch stub in production code.
+					// This is only needed in `wrangler dev`.
+					expect(entry).not.toMatch(/modules-watch-stub\.js/);
```
CarmenPopoviciu, line 2638: is it worth checking the watch stub *did* get added for `wrangler dev`?

RamIdeas (approved):

packages/wrangler/src/__tests__/deploy.test.ts
petebacondarwin: There is a test in the fixture that would fail if the stub was not there because the watchers would not be wired up: https://github.com/cloudflare/workers-sdk/blob/7ad8ddd732c31441c9d245d7edd36199b768fc54/fixtures/additional-modules/test/index.test.ts#L97

PR #5976, opened 2024-06-05
fix: define defineWorkersConfig using overload signatures

m-shaka:
The type definition of `defineWorkersConfig` doesn't work with `mergeConfig` of `vitest/config` because of type mismatch
This function should be an overload function like [defineConfig](https://github.com/vitest-dev/vitest/blob/f15b4e99da1d8e5531ecc900792e3bb490fe1c7a/packages/vitest/src/config.ts#L23-L29)

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] ~Not necessary because:~ Sorry but I'm not sure if I need to write additional tests
  - [x] I don't know
  - [ ] Not required because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: no breaking change

### DIFF
--- a/packages/vitest-pool-workers/src/config/index.ts
+++ b/packages/vitest-pool-workers/src/config/index.ts
@@ -31,25 +31,25 @@ globalThis.structuredClone ??= function (value, options) {
 	return message.message;
 };
 
-type ConfigFn<T extends UserConfig> = ((env: ConfigEnv) => T | Promise<T>);
+type ConfigFn<T extends UserConfig> = (env: ConfigEnv) => T | Promise<T>;
 
 export type AnyConfigExport<T extends UserConfig> =
 	| T
 	| Promise<T>
-	| ConfigFn<T>
+	| ConfigFn<T>;
 
 function mapAnyConfigExport<T extends UserConfig, U extends UserConfig>(
 	f: (t: T) => U,
 	config: T
-): U
+): U;
 function mapAnyConfigExport<T extends UserConfig, U extends UserConfig>(
 	f: (t: T) => U,
 	config: Promise<T>
-): Promise<U>
+): Promise<U>;
 function mapAnyConfigExport<T extends UserConfig, U extends UserConfig>(
 	f: (t: T) => U,
 	config: ConfigFn<T>
-): ConfigFn<U>
+): ConfigFn<U>;
 function mapAnyConfigExport<T extends UserConfig, U extends UserConfig>(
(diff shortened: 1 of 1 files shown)


packages/vitest-pool-workers/src/config/index.ts
```diff
@@ -161,17 +144,35 @@ function ensureWorkersConfig<T extends UserConfig>(config: T): T {
...
+	config: Promise<WorkersUserConfigExport>
+): Promise<WorkersUserConfigExport>;
+export function defineWorkersConfig(
+	config: ConfigFn<WorkersUserConfigExport>
+): ConfigFn<WorkersUserConfigExport>;
 export function defineWorkersConfig(
 	config: AnyConfigExport<WorkersUserConfigExport>
 ): AnyConfigExport<WorkersUserConfigExport> {
-	return mapAnyConfigExport(ensureWorkersConfig, config);
+	if (typeof config === "function") {
```
m-shaka, line 159: I merged `defineWorkersConfig` and `mapAnyConfigExport`. Otherwise,  I would have to write overload signatures  to both methods

packages/vitest-pool-workers/src/config/index.ts
```diff
@@ -161,15 +176,43 @@ function ensureWorkersConfig<T extends UserConfig>(config: T): T {
...
+export function defineWorkersConfig(
+	config: Promise<WorkersUserConfigExport>
+): Promise<WorkersUserConfigExport>
+export function defineWorkersConfig(
+	config: ConfigFn<WorkersUserConfigExport>
+): ConfigFn<WorkersUserConfigExport>
 export function defineWorkersConfig(
 	config: AnyConfigExport<WorkersUserConfigExport>
 ): AnyConfigExport<WorkersUserConfigExport> {
+	if (typeof config === 'function') {
```
m-shaka, line 191: Type guard is needed because `mayAnyConfigExport` is also overloaded.
I could remove `mayAnyConfigExport` and write its body directly into `defineWorkersConfig` and `defineWorkersProject` to avoid many overload signatures, but  I didn't know the original intent of `mapAnyConfigExport` so I decided to keep it

threepointone (approved):

threepointone:
Looks good, thank you for the PR!

PR #5978, opened 2024-06-05
chore: add launch.json to debug single (current) test file

andyjessop:

This is a quality-of-life improvement for debugging and testing packages locally. If you're using VSCode, and your current buffer is a test file, you can select the debug panel and you will see a "Debug Current Test File" option:

Screenshot 2024-06-05 at 15 34 19

Hit `f5` or click the green triangle and the test will start to run in the VSCode console:

Screenshot 2024-06-05 at 15 37 48

This specific implementation runs as a command attached to the `tools` package, running it with `pnpm` as with any other package.json script in the repo.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: QoL update - no user-facing code touched.
  - [ ] I don't know
  - [x] Not required because: as above ^
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: as above ^ 
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: as above ^

### DIFF
--- a/tools/package.json
+++ b/tools/package.json
@@ -14,6 +14,7 @@
 		"@typescript-eslint/eslint-plugin": "^5.55.0",
 		"@typescript-eslint/parser": "^5.55.0",
 		"ts-dedent": "^2.2.0",
-		"undici": "5.28.4"
+		"undici": "5.28.4",
+		"find-up": "^6.3.0"
 	}
 }
--- a/tools/package.json
+++ b/tools/package.json
@@ -5,7 +5,8 @@
 	"scripts": {
 		"check:type": "tsc",
 		"test:ci": "vitest run",
-		"check:lint": "eslint ."
+		"check:lint": "eslint .",
+		"test:file": "node -r esbuild-register test/run-test-file.ts"
 	},
 	"devDependencies": {
 		"@cloudflare/eslint-config-worker": "workspace:*",
--- a/tools/test/run-test-file.ts
+++ b/tools/test/run-test-file.ts
@@ -23,7 +23,9 @@ testProcess.on("close", (code) => {
 	process.exit(code ?? 1);
 });
 
-// Function to find the nearest package.json
+/**
+ * Finds the nearest parent package.json to the provided file.
+ */
 function findNearestPackageJson(file: string) {
 	let dir = path.dirname(file);
 
--- a/tools/test/run-test-file.ts
+++ b/tools/test/run-test-file.ts
@@ -1,0 +1,38 @@
+import { spawn } from "child_process";
+import fs from "fs";
+import path from "path";
+
+const currentFile = process.argv[2];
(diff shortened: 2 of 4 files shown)


tools/package.json
```diff
@@ -5,7 +5,8 @@
 	"scripts": {
 		"check:type": "tsc",
 		"test:ci": "vitest run",
-		"check:lint": "eslint ."
+		"check:lint": "eslint .",
+		"test:file": "node -r esbuild-register test/run-test-file.ts"
```
petebacondarwin, line 9: I was literally thinking of asking if you could make it TS and add it to `tools` package!!

tools/test/run-test-file.ts
```diff
@@ -0,0 +1,40 @@
+import { spawn } from "child_process";
+import fs from "fs";
+import path from "path";
+
+const currentFile = process.argv[2];
+const packageJsonPath = findNearestPackageJson(currentFile);
```
petebacondarwin, line 6: There is a library that we use elsewhere called `find-up`, which can be used for this. E.g. https://github.com/cloudflare/workers-sdk/blob/cfde0587c40cbcdfcb5e7981e8105e1b99f031fa/packages/wrangler/src/pages/utils.ts#L50

petebacondarwin (approved):

tools/test/run-test-file.ts
andyjessop: Great, I'll use that instead then.

tools/package.json
```diff
@@ -5,14 +5,16 @@
...
 	},
 	"devDependencies": {
 		"@cloudflare/eslint-config-worker": "workspace:*",
 		"@cloudflare/workers-tsconfig": "workspace:*",
 		"@typescript-eslint/eslint-plugin": "^5.55.0",
 		"@typescript-eslint/parser": "^5.55.0",
 		"ts-dedent": "^2.2.0",
-		"undici": "5.28.4"
+		"undici": "5.28.4",
+		"find-up": "^6.3.0"
```
andyjessop, line 18: @petebacondarwin I put the dependency in this package because that's what we seem to do elsewhere (install dependencies in the package where they are consumed). Didn't know if this sort of thing we would hoist to the root. I used the same version as we use with `wrangler` so that they would be easier to upgrade together if necessary.

tools/test/run-test-file.ts
andyjessop: https://github.com/cloudflare/workers-sdk/pull/5978/commits/8489118fae0c4d694d9c641ea2d297ae6fabdf95

tools/package.json
petebacondarwin: That's the right thing to do. The TS file will be executed from inside this package.

PR #5988, opened 2024-06-06
rename `secret:bulk` top-level command to `secret bulk` subcommand

### LINKED ISSUE #5168: Remove ':' from secret:bulk command

RamIdeas:
and deprecate `secret:bulk` – command still exists with same functionality but also with a deprecation notice

Fixes #5168

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: unit tested
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/secret.test.ts
+++ b/packages/wrangler/src/__tests__/secret.test.ts
@@ -564,12 +564,12 @@ describe("wrangler secret", () => {
 		});
 	});
 
-	describe("secret:bulk", () => {
-		it("should fail secret:bulk w/ no pipe or JSON input", async () => {
+	describe("bulk", () => {
+		it("should fail secret bulk w/ no pipe or JSON input", async () => {
 			vi.spyOn(readline, "createInterface").mockImplementation(
 				() => null as unknown as Interface
 			);
-			await runWrangler(`secret:bulk --name script-name`);
+			await runWrangler(`secret bulk --name script-name`);
 			expect(std.out).toMatchInlineSnapshot(
 				`"🌀 Creating the secrets for the Worker \\"script-name\\" "`
 			);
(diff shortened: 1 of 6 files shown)


CarmenPopoviciu:
should we mark https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/index.ts#L399-L404 as `DEPRECATED` with a comment? like we do for other deprecated cmds?

packages/wrangler/src/__tests__/secret.test.ts
```diff
@@ -745,6 +766,13 @@ describe("wrangler secret", () => {
 
 			"
 		`);
+			expect(std.warn).toMatchInlineSnapshot(`
+			"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1m\`wrangler secret:bulk\` is deprecated and will be removed in a future major version.[0m
+
+			  Please use \`wrangler secret bulk\` instead, which accepts exactly the same arguments.
```
andyjessop, line 772: Not for this PR of course, but I'm wondering if it's worth using an abstraction to remove the boilerplate from these snapshots. I suppose the pro is that it's better to be explicit about the output, and the cons are that it's less readable and potentially less maintainable.

andyjessop (approved):
Looks good! KV next..?

CarmenPopoviciu (approved):

packages/wrangler/src/__tests__/secret.test.ts
RamIdeas: In this case, this was intentional to test the deprecation message shows. But as mentioned on chat, I'm going to make that a bit more explicit like in https://github.com/cloudflare/workers-sdk/pull/5989

PR #5992, opened 2024-06-06
feat: wrangler deploy prompts warning with deployment

danielrs:

Shows a warning to avoid accidental overwrite of a percentage-based deployment.

- Tests
  - [ ] TODO (before merge)
  - [X] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -43,6 +43,10 @@ import {
 	mswSuccessOauthHandlers,
 	mswSuccessUserHandlers,
 } from "./helpers/msw";
+import {
+	mswListNewDeploymentsLatestFiftyFifty,
+	mswListNewDeploymentsLatestFull,
+} from "./helpers/msw/handlers/versions";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { runWrangler } from "./helpers/run-wrangler";
 import { writeWorkerSource } from "./helpers/write-worker-source";
@@ -78,6 +82,7 @@ describe("deploy", () => {
 		setIsTTY(true);
 		mockLastDeploymentRequest();
 		mockDeploymentsListRequest();
+		msw.use(...mswListNewDeploymentsLatestFull);
 		logger.loggerLevel = "log";
 	});
 
@@ -421,6 +426,34 @@ describe("deploy", () => {
 
 			"
 		`);
+		});
+
+		it("should warn user when worker has deployment with multiple versions", async () => {
+			msw.use(...mswListNewDeploymentsLatestFiftyFifty);
+			writeWranglerToml();
+			writeWorkerSource();
+			mockSubDomainRequest();
+			mockUploadWorkerRequest();
+			mockConfirm({
+				text: `"wrangler deploy" will upload a new version and deploy it globally immediately.\nAre you sure you want to continue?`,
+				result: false,
+			});
+
+			await runWrangler("deploy ./index");
+
+			expect(std.warn).toMatchInlineSnapshot(`
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mYour last deployment has multiple versions. To progress this deployment use \\"wrangler versions deploy\\" instead.[0m
+
+				  Currently deployed versions:
+
+				  Version(s):  (50%) test-name:version:0
+				                   Created:  2021-01-01T00:00:00.000Z
+
+				               (50%) test-name:version:1
+				                   Created:  2021-01-01T00:00:00.000Z
+
+				"
+			`);
 		});
 
 		it("should warn user when additional properties are passed to a services config", async () => {
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -43,10 +43,7 @@ import {
 	mswSuccessOauthHandlers,
 	mswSuccessUserHandlers,
 } from "./helpers/msw";
-import {
-	mswListNewDeploymentsLatestFiftyFifty,
-	mswListNewDeploymentsLatestFull,
-} from "./helpers/msw/handlers/versions";
+import { mswListNewDeploymentsLatestFull } from "./helpers/msw/handlers/versions";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { runWrangler } from "./helpers/run-wrangler";
 import { writeWorkerSource } from "./helpers/write-worker-source";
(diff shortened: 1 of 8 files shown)


packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,6 +209,30 @@ function getConfig(
...
+ */
+export async function confirmLatestDeploymentOverwrite(
+	accountId: string,
+	scriptName: string
+) {
+	try {
+		const latest = await fetchLatestDeployment(accountId, scriptName);
+		if (latest && latest.versions.length >= 2) {
+			logger.warn(
+				`Your latest deployment has multiple versions.\n"wrangler deploy" will upload a new version and deploy it globally immediately.`
```
RamIdeas, line 223: Should this also include a mention of "wrangler versions deploy" as the alternative?

tanushree-sharma:
We should return the current status of the deployment + allow users to override and wrangler deploy. 

Copy:

```> wrangler deploy

▲ [WARNING] Your last deployment has multiple versions. To progress this deployment use "wrangler versions deploy". "wrangler deploy" will upload a new version and deploy it globally immediately.

Currently deployed versions:

Version(s):  (50%) d58c1fb8-6015-410f-8fb7-e2818788b337
                    Created:  2024-06-07T13:21:14.974Z
             
                   (50%) 2ea0b4e6-f593-41cf-8a70-f0ef64bb0b02
                   Created:  2024-06-05T22:17:53.355Z
 
 Are you sure you want to continue? (y/n). ```

File: packages/wrangler/src/__tests__/deploy.test.ts [hunk 2]
```diff
@@ -423,6 +428,34 @@ describe("deploy", () => {
 		`);
 		});
 
+		it("should warn user when worker has deployment with multiple versions", async () => {
```
petebacondarwin, line 431: Can we have a test that checks what happens if the user says yes? 
Also what happens if the console is non-interactive? Can that be tested too?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,6 +209,68 @@ function getConfig(
...
+				return `${percentage} ${version.id}\n${details}`;
+			});
+
+			// Format deployment.
+
+			const formattedDeployment = formatLabelledValues({
+				"Version(s)": formattedVersions.join("\n\n"),
+			});
+
+			// Print message and confirmation.
```
RamIdeas, line 252: Can this all be replaced with a call to printLatestDeployment (the function right below this one)?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,6 +209,68 @@ function getConfig(
...
+			logger.warn(
+				`Your last deployment has multiple versions. To progress this deployment use "wrangler versions deploy" instead.
+Currently deployed versions:
+
+${formattedDeployment}`
+			);
+
+			return await confirm(
+				`"wrangler deploy" will upload a new version and deploy it globally immediately.\nAre you sure you want to continue?`
+			);
```
RamIdeas, line 263: Can you use inputPrompt instead of confirm please? There are other usages in this file you can use for reference

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -433,6 +436,14 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 
 	const { format } = props.entry;
 
+	if (!props.dispatchNamespace && prod && accountId && scriptName) {
+		const yes = await confirmLatestDeploymentOverwrite(accountId, scriptName);
+		if (!yes) {
+			cancel("Aborting deploy...");
```
RamIdeas, line 442: This is mixing new style (c3/cli) with a command using the old style (wrangler). You can just `logger.error` in this command but I don't think it's a big deal to leave as-is (nit)

packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,6 +209,61 @@ function getConfig(
...
+			});
+
+			const traffic = new Map(
+				latest.versions.map((v) => [v.version_id, v.percentage])
+			);
+
+			// Print message and confirmation.
+
+			cli.warn(
+				`Your last deployment has multiple versions. To progress this deployment use "wrangler versions deploy" instead.`,
```
RamIdeas, line 241: copy nit: "that" instead of "this" because "this" deployment should refer to one being triggered by this command whereas "that" deployment would refer to the "last" deployment mentioned in this line?

cc @tanushree-sharma

```suggestion
				`Your last deployment has multiple versions. To progress that deployment use "wrangler versions deploy" instead.`,
```

packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,6 +209,61 @@ function getConfig(
...
+				`${leftT} Your last deployment has ${versions.length} version(s):`
+			);
+			printVersions(versions, traffic);
+
+			return inputPrompt<boolean>({
+				type: "confirm",
+				question: `"wrangler deploy" will upload a new version and deploy it globally immediately.\nAre you sure you want to continue?`,
+				label: "",
+				defaultValue: true, // defaults to true to keep current behavior.
+				acceptDefault: !isInteractive() || CI.isCI(),
```
RamIdeas, line 255: Thanks for calling out this default is for back-compat

Really this default should be `false` as we're trying to prevent a footgun. But `false` default means we're half-preventing the footgun in interactive mode but not in CI (for back-compat)

I would suggest the following, which would make the default false in interactive mode to prevent a user in autopilot hitting enter and instead needing to explicitly respond yes. And still default to true in CI for back-compat.
```suggestion
				defaultValue: !isInteractive() || CI.isCI(), // defaults to true in CI for back-compat
				acceptDefault: !isInteractive() || CI.isCI(),
```

packages/wrangler/src/deploy/deploy.ts
danielrs: Leaving this as-is with `cancel` because `logger.error` seems to write to a log file somewhere? Not what we want in this case.

packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,6 +209,61 @@ function getConfig(
...
+
+			const versions = latest.versions.map((v) => {
+				const version = versionCache.get(v.version_id);
+				assert(version);
+				return version;
+			});
+
+			const traffic = new Map(
+				latest.versions.map((v) => [v.version_id, v.percentage])
+			);
```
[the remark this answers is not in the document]
danielrs: What I didn't like about `printLatestDeployment` is that it would fetch the deployment again when I already had it. Ended up refactoring to a function named `printDeployment` that lets you pass and print any deployment as a parameter.

packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,43 +215,112 @@ function getConfig(
...
+		"current",
+		versionCache
+	);
+}
 
+export async function printDeployment(
+	accountId: string,
+	workerName: string,
+	deployment: ApiDeployment | undefined,
+	adjetive: "current" | "last",
```
RamIdeas, line 286: typo 
```suggestion
	adjective: "current" | "last",
```

packages/wrangler/src/versions/deploy.ts
```diff
@@ -205,43 +215,112 @@ function getConfig(
...
+) {
+	const [versions, traffic] = await fetchDeploymentVersions(
+		accountId,
+		workerName,
+		deployment,
+		versionCache
+	);
 	cli.logRaw(
-		`${leftT} Your current deployment has ${versions.length} version(s):`
+		`${leftT} Your ${adjetive} deployment has ${versions.length} version(s):`
```
RamIdeas, line 296: typo
```suggestion
		`${leftT} Your ${adjective} deployment has ${versions.length} version(s):`
```

packages/wrangler/src/deploy/deploy.ts
RamIdeas: all logs are meant to go to the log file for debug-ability but no big deal to leave this

packages/wrangler/src/versions/deploy.ts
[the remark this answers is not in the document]
RamIdeas: I had an idea to pass a callback param like `skipIf(deployment: ApiDeployment) => boolean` which would enable flow control using the one api request

but this is fine for now, and I can refactor in a follow up if we decide it's worth it

RamIdeas (approved):

Issue #729, opened 2022-03-30
feat: allow durable objects migrations within environments 


threepointone:
We let folks define `[durable_objects]` under environments (in fact, we demand it), so one of the two should be true - 
- Either we should verify that the shape of `[durable_objects]` should be the same across environments (except for DOs with `script_name`?)
- Or `[migrations]` should be defined under environments as well. 

I think it should be the former, but open to suggestions.

petebacondarwin:
What happens if you are developing a new DurableObject class, not yet on `production`, and you want to deploy to `staging` for testing? It seems that the only way is to decouple the migrations for each environment...

petebacondarwin:
Given that we are currently sticking with "non-service environments", each environment declared in wrangler.toml results in a new Worker which can have its own Durable Object (DO) declarations. It is very reasonable to have an environment that declares a DO that is not declared in other environments - e.g. adding a new DO in staging before sending it to production, removing a redundant DO in staging before removing it from production.

Therefore I think it is right that migration should also be defined in environments rather than being only at the top level config.

To avoid backward compatibility concerns it is best to make this an "inheritable" configuration property, which means that it can be simply defined at the top level and then apply to all environments without having to redefined it for each environment.

WalshyDev:
Hit this today too, we should definitely be supporting this per-env.
cc: @petebacondarwin @RamIdeas 

I imagine others will hit this too (is this something you have ran into @jahands with your work?)

PR #5995, opened 2024-06-07
Allow do migrations in environments

petebacondarwin:

Allow Durable Object migrations to be overridable in environments. By making the `migrations` key inheritable, users can provide different migrations for each wrangler.toml environment.

Resolves https://github.com/cloudflare/workers-sdk/issues/729

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/smooth-pumpkins-breathe.md
+++ b/.changeset/smooth-pumpkins-breathe.md
@@ -1,0 +1,10 @@
+---
+"wrangler": minor
+---
+
+feat: allow Durable Object migrations to be overridable in environments
+
+By making the `migrations` key inheritable, users can provide different migrations
+for each wrangler.toml environment.
+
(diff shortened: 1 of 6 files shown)


threepointone (approved):
Does this make migrations under environments mandatory, or will it pick up the top one if we don't define it?

petebacondarwin:
> Does this make migrations under environments mandatory, or will it pick up the top one if we don't define it?

To avoid a breaking change it is inheritable so it will be picked up from the top level.

Scamreno (commented):
Thanks

PR #5999, opened 2024-06-08
fix: avoid esbuild warning when running dev/bundle

threepointone:
I've been experimenting with esbuild 0.21.4 with wrangler. It's mostly been fine. But I get this warning every time

```
▲ [WARNING] Import "__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__" will always be undefined because there is no matching export in "src/index.ts" [import-is-undefined]

    .wrangler/tmp/bundle-Z3YXTd/middleware-insertion-facade.js:8:23:
      8 │ .....(OTHER_EXPORTS.__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__ ?? []),
        ╵
```

This is because esbuild@0.18.5 enabled a warning by default whenever an undefined import is accessed on an imports object. However we abuse imports to inject stuff in `middleware.test.ts`. We should probably fix that with a better solution, but an immediate fix to this warning is to make that access not be statically analyzable. This patch does so by defining the export name as a regular variable and using that when reading from it.

### DIFF
--- a/packages/wrangler/src/__tests__/middleware.test.ts
+++ b/packages/wrangler/src/__tests__/middleware.test.ts
@@ -861,7 +861,19 @@ describe("middleware", () => {
 					.replace(/\/\/ .*/g, "")
 					.trim()
 			).toMatchInlineSnapshot(`
-				"var src_default = {
+				"var __defProp = Object.defineProperty;
+				var __export = (target, all) => {
+				  for (var name in all)
+				    __defProp(target, name, { get: all[name], enumerable: true });
+				};
+
+
+				var src_exports = {};
+				__export(src_exports, {
+				  DurableObjectExample: () => DurableObjectExample,
+				  default: () => src_default
+				});
+				var src_default = {
 				  async fetch(request, env) {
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/__tests__/middleware.test.ts
```diff
@@ -861,7 +861,19 @@ describe("middleware", () => {
 					.replace(/\/\/ .*/g, "")
 					.trim()
 			).toMatchInlineSnapshot(`
-				"var src_default = {
+				"var __defProp = Object.defineProperty;
```
threepointone, line 864: of note, it adds some extra code for reading that dynamically. which is why we need a proper fix later so we don't ship any extra code. but this should be fine for now.

threepointone:
Alternately, we should delete middleware.test.ts, it's testing an implementation detail that's explicitly added for the test, which is weird. The test should be for actual middleware.

petebacondarwin:
Let get #6006 green and land that since it also avoids adding code to production bundles.

PR #6001, opened 2024-06-10
Config controller

penalosa:

This change removes some 'passthrough' config options from the BundleController PR and now uses expected StartDevWorkerOptions.

A thin implementation of the config controller has been implemented and integrated into the react-flow behind the existing -x-dev-env flag.

This change is covered by local+remote e2e tests both with+without the –x-dev-env flag. The dev-env fixture tests will be updated in a follow-up PR.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because: refactor behind experimental flag
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: refactor behind experimental flag

### DIFF
--- a/packages/wrangler/src/api/startDevWorker/BundlerController.ts
+++ b/packages/wrangler/src/api/startDevWorker/BundlerController.ts
@@ -183,7 +183,6 @@ export class BundlerController extends Controller<BundlerControllerEventMap> {
 			ignoreInitial: true,
 		});
 		this.#customBuildWatcher.on("ready", () => {
-			assert(config.build?.custom?.watch);
 			void this.#runCustomBuild(config, String(pathsToWatch));
 		});
 
--- a/packages/wrangler/src/dev.tsx
+++ b/packages/wrangler/src/dev.tsx
@@ -628,7 +628,7 @@ export async function startApiDev(args: StartDevOptions) {
 	);
 
 	const devEnv = new DevEnv();
-	if (!args.disableDevRegistry) {
+	if (!args.disableDevRegistry && args.experimentalDevEnv) {
 		devEnv.runtimes.forEach((runtime) => {
 			runtime.on("reloadComplete", async (reloadEvent: ReloadCompleteEvent) => {
 				if (!reloadEvent.config.dev?.remote) {
--- a/packages/wrangler/src/dev/dev.tsx
+++ b/packages/wrangler/src/dev/dev.tsx
@@ -603,12 +603,11 @@ function DevSession(props: DevSessionProps) {
 	]);
 
 	const onBundleStart = useCallback(() => {
-		if (!props.experimentalDevEnv) {
-			devEnv.proxy.onBundleStart({
-				type: "bundleStart",
-				config: startDevWorkerOptions,
-			});
-		}
+		// NOTE: this callback won't be called if props.experimentalDevEnv
+		devEnv.proxy.onBundleStart({
+			type: "bundleStart",
+			config: startDevWorkerOptions,
+		});
 	}, [devEnv, startDevWorkerOptions, props.experimentalDevEnv]);
 	const onBundleComplete = useCallback(
 		(bundle: EsbuildBundle) => {


packages/wrangler/src/__tests__/api/startDevWorker/BundleController.test.ts
```diff
@@ -68,20 +67,15 @@ describe("happy path bundle + watch", () => {
...
-			script: unusable(),
-			_entry: {
-				file: path.resolve("src/index.ts"),
-				directory: path.resolve("src"),
-				format: "modules",
-				moduleRoot: path.resolve("src"),
-				name: "worker-name",
-			},
-			_additionalModules: [],
+			script: { path: path.resolve("src/index.ts") },
```
RamIdeas, line 70: As a user I would expect this filepath to be relative to the cwd. So the user shouldn't have to call path.resolve – the ConfigController should do that on their behalf

packages/wrangler/src/__tests__/api/startDevWorker/LocalRuntimeController.test.ts
```diff
@@ -695,8 +695,8 @@ describe("Bindings", () => {
 		expect(res.status).toBe(404);
 		res = await fetch(new URL("/secrets.txt", url));
 		expect(res.status).toBe(200);
-
-		config.site = { path: tmp, exclude: ["secrets.txt"] };
+		config.legacy = {};
+		config.legacy.site = { bucket: tmp, exclude: ["secrets.txt"] };
```
RamIdeas, line 699: Need to avoid mutating

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -41,79 +43,92 @@ export class BundlerController extends Controller<BundlerControllerEventMap> {
...
 				);
 			}
 
-			const entryDirectory = path.dirname(config._entry.file);
+			const entry: Entry = {
+				file: config.script.path,
+				directory: config.directory,
+				format: config.build.format,
+				moduleRoot: config.build.moduleRoot,
+			};
```
RamIdeas, line 85: 👏

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -154,22 +170,21 @@ export class BundlerController extends Controller<BundlerControllerEventMap> {
...
+		if (!config.build?.custom?.command) {
 			return;
 		}
 
-		assert(config._entry);
-		assert(config.build?.custom.watch);
+		assert(config.build?.custom?.watch);
 
-		this.#customBuildWatcher = watch(config.build?.custom.watch, {
+		this.#customBuildWatcher = watch(config.build?.custom?.watch, {
```
RamIdeas, line 179: the assert above should make these nullish operators unnecessary, right? in fact, we'd never want to pass `undefined` to `watch()`

packages/wrangler/src/dev/dev.tsx
```diff
@@ -448,6 +583,25 @@ function DevSession(props: DevSessionProps) {
 				testScheduled: props.testScheduled,
 				getRegisteredWorker: (name) => workerDefinitions[name],
 			},
+			legacy: {
+				site:
```
RamIdeas, line 587: I got this wrong originally, it should be plural: `sites`.

Wonder if it's worth being super-explicit with the product name like `legacy.WorkersSites`

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -50,31 +51,23 @@ export interface StartDevWorkerOptions {
 	 * This is the `main` property of a wrangler.toml.
 	 * You can specify a file path or provide the contents directly.
 	 */
-	script: File<string>;
```
penalosa, line 53: This should be `main` or `entrypoint`

packages/wrangler/src/dev.tsx
```diff
@@ -584,6 +629,25 @@ export async function startApiDev(args: StartDevOptions) {
 		{ sendMetrics: config.send_metrics, offline: !args.remote }
 	);
 
+	const devEnv = new DevEnv();
+	if (!args.disableDevRegistry) {
```
penalosa, line 631: This should be wrapped in `args.experimentalDevEnv`

packages/wrangler/src/dev/dev.tsx
```diff
@@ -647,22 +608,16 @@ function DevSession(props: DevSessionProps) {
 	]);
 
 	const onBundleStart = useCallback(() => {
-		devEnv.proxy.onBundleStart({
-			type: "bundleStart",
-			config: startDevWorkerOptions,
-		});
-	}, [devEnv, startDevWorkerOptions]);
+		if (!props.experimentalDevEnv) {
```
penalosa, line 606: This isn't needed

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -174,18 +174,18 @@ export class BundlerController extends Controller<BundlerControllerEventMap> {
...
 			// TODO: add comments re this ans ready
 			ignoreInitial: true,
 		});
-		this.#customBuildWatcher.on(
-			"ready",
-			() =>
-				void this.#runCustomBuild(config, String(config.build?.custom?.watch))
-		);
+		this.#customBuildWatcher.on("ready", () => {
+			assert(config.build?.custom?.watch);
```
penalosa, line 186: This should be removed

petebacondarwin (approved):
LGTM - let's see how it flies!

packages/wrangler/src/dev/dev.tsx
```diff
@@ -489,22 +603,15 @@ function DevSession(props: DevSessionProps) {
 	]);
 
 	const onBundleStart = useCallback(() => {
+		// NOTE: this callback won't be called if props.experimentalDevEnv
 		devEnv.proxy.onBundleStart({
 			type: "bundleStart",
 			config: startDevWorkerOptions,
 		});
-	}, [devEnv, startDevWorkerOptions]);
+	}, [devEnv, startDevWorkerOptions, props.experimentalDevEnv]);
```
threepointone, line 611: @RamIdeas What is the expected behaviour here? I think your comment was incomplete, so I don't understand this. Did you mean to write `if(props.experimentalDevEnv) { devEnv.proxy.onBundleStart(/*...*/) }`

RamIdeas: > Did you mean to write if(props.experimentalDevEnv) { devEnv.proxy.onBundleStart(/*...*/) }

No, explicitly the other way around. The faking of events is what wrangler main was doing. The experimental flag is newer and is the code path of real events within DevEnv.

Slightly confusing, so I put an if statement around it `if(!props.experimentalDevEnv) { devEnv.proxy.onBundleStart(/*...*/) }` (note the `!`) to be explicit but Samuel and I then then decided it was unnecessary as `onBundleStart` is not called unless `props.experimentalDevEnv === false`. So we settled on a comment instead. The prop left in the dependency array was a mistake but is not an issue since it is static and won't change. I'll make a follow-up PR to remove the extraneous dependency array item.

RamIdeas: Actually I'm going to put the `if` statement back just to be super explicit and avoid confusion. Even if it's redundant, it being code and not a comment will avoid confusion as to whether it is correct.

RamIdeas: Addressed here https://github.com/cloudflare/workers-sdk/pull/6116

PR #6003, opened 2024-06-11
fix: add thread pool to runTests to avoid process starvation

### LINKED ISSUE #5381: Vitest integration doesn't respect `os.availableParallelism()`/`os.cpus().length`
Currently, the Workers Vitest integration will spawn all required `workerd` processes at once, rather than batching these up and only using `os.availableParallelism()`/`os.cpus().length` at once. Excess parallelism like this can result in starvation and slow tests down. We should make sure to respect these values.

andyjessop:

Fixes #5381.

In order to ensure that test runs don't fight for resources, which could lead to slower tests or even deadlocks in some circumstances (where two threads are waiting for each other to release resources), this PR limits the number of active threads used by the test runner.

It creates a generic `ResourcePool`, which counts the number of resources that are "active". If an attempt is made to attain a resource when the maximum number has been reached, instead a `Promise` is returned, which resolves the next time a thread becomes available.

## Testing

Although unit tests have been added, it would have been better to also include integration tests. Several attempts were made here, but none was viable.

The integration testing helper used in the `vitest-pool-workers` package takes any number of fixture files, and then spawns a child process in which to run those with Vitest. One issue with this is that we are unable to mock `os.availableParallelism` because any mocks defined in the main test file are not passed through to the child process.

One potential way to solve this would be to add console logs in our process pool code, which could then be extracted from `stdout` in the main test file. I'm open to arguments as to whether or not this is a desirable course of action.

- Tests
  - [ ] TODO (before merge)
  - [x] Included (**see caveat above**)
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: This doesn't touch `wrangler` commands E2E
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: This doesn't touch any user-facing APIs.


packages/vitest-pool-workers/src/pool/helpers.ts
```diff
@@ -24,3 +25,56 @@ export function getRelativeProjectPath(
...
+
+type Resolve = () => void;
+
+/**
+ * A generic pool that can be used to limit the number of active resources. The resources
+ * are anonymous, so we don't care about their implementation. The goal is just to count
+ * the number of resources in-use.
+
+ */
+export class ResourcePool {
```
petebacondarwin, line 41: I wonder if this should actually be called a `Semaphore` (see https://learn.microsoft.com/en-us/dotnet/api/system.threading.semaphore?view=net-8.0).

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -649,15 +651,37 @@ function maybeGetResolvedMainPath(project: Project): string | undefined {
...
+
+	return async function run(...args: RunTestsArgs): Promise<void> {
+		await threadPool.nextAvailableResource();
+
+		// Promise.finally will be called in the event of either resolving or rejecting,
+		// so the thread will always be released unless the test run hangs.
+		return runTests(...args).finally(() => {
+			threadPool.releaseResource();
+		});
+	};
```
petebacondarwin, line 678: Could we lock this section of code using a form such as:

```
const limiter = new Semaphore(calculateAvailableThreads);
limiter.runWith(() => runTests(...args));
});
```

where the `runWith()` function will handle locking, waiting and unlocking the semaphore?

andyjessop: I like that - it's a simpler API.

packages/vitest-pool-workers/src/pool/helpers.ts
andyjessop: That looks pretty accurate.

https://en.wikipedia.org/wiki/Semaphore_(programming)

> A useful way to think of a semaphore as used in a real-world system is as a record of how many units of a particular resource are available, coupled with operations to adjust that record safely (i.e., to avoid [race conditions](https://en.wikipedia.org/wiki/Race_condition)) as units are acquired or become free, and, if necessary, wait until a unit of the resource becomes available.

packages/vitest-pool-workers/src/pool/helpers.ts
```diff
@@ -24,3 +25,61 @@ export function getRelativeProjectPath(
...
+
+	get availableResourcesCount(): number {
+		return this.#maxResources - this.#activeResources;
+	}
+
+	async runWith<T>(fn: () => Promise<T>): Promise<T> {
+		if (this.#activeResources < this.#maxResources) {
+			this.#activeResources++;
+			return fn().finally(() => {
+				void this.#releaseResource();
```
petebacondarwin, line 61: I wonder if this should actually return the promise here?

PR #6002, opened 2024-06-10
Revert "Batch file uploads to avoid EMFILE"

GregBrimble:
Reverts cloudflare/workers-sdk#5632

PR #6009, opened 2024-06-11
fix: reduce the number of parallel file reads on Windows to avoid EMFILE type errors

RamIdeas:

Reapplies the batching reverted in https://github.com/cloudflare/workers-sdk/pull/6002 but with a tested helper method to ensure batches are correctly compiled

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because:

### DIFF
--- a/packages/wrangler/src/pages/constants.ts
+++ b/packages/wrangler/src/pages/constants.ts
@@ -1,10 +1,12 @@
 import { version as wranglerVersion } from "../../package.json";
 
+const isWindows = process.platform === "win32";
+
 export const MAX_ASSET_COUNT = 20_000;
 export const MAX_ASSET_SIZE = 25 * 1024 * 1024;
 export const PAGES_CONFIG_CACHE_FILENAME = "pages.json";
 export const MAX_BUCKET_SIZE = 40 * 1024 * 1024;
-export const MAX_BUCKET_FILE_COUNT = 2000;
+export const MAX_BUCKET_FILE_COUNT = isWindows ? 1000 : 2000;
 export const BULK_UPLOAD_CONCURRENCY = 3;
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/utils/create-batches.ts
```diff
@@ -0,0 +1,8 @@
+export function* createBatches<T>(
```
andyjessop, line 1: Is there a reason this needs to be a generator rather than a standard function returning an array?

andyjessop (approved):
LGTM 👍

packages/wrangler/src/pages/constants.ts
```diff
@@ -1,10 +1,12 @@
 import { version as wranglerVersion } from "../../package.json";
 
+const isWindows = process.platform === "win32";
+
 export const MAX_ASSET_COUNT = 20_000;
 export const MAX_ASSET_SIZE = 25 * 1024 * 1024;
 export const PAGES_CONFIG_CACHE_FILENAME = "pages.json";
 export const MAX_BUCKET_SIZE = 40 * 1024 * 1024;
-export const MAX_BUCKET_FILE_COUNT = 2000;
+export const MAX_BUCKET_FILE_COUNT = isWindows ? 1000 : 2000;
```
petebacondarwin, line 9: Let's add a comment here to explain why Windows needs a lower limit, so that people don't try to change it in the future.

petebacondarwin (approved):

PR #6010, opened 2024-06-11
test: add checks of uploads of very large number of Pages assets

petebacondarwin:

Adds a test to avoid accidentally breaking Pages uploads for large numbers of assets.

**This test would have failed after the #5632 change.**

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: just a new non-e2e test added
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: just adding a test
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: just adding a test

### DIFF
--- a/packages/wrangler/src/__tests__/pages/project-upload.test.ts
+++ b/packages/wrangler/src/__tests__/pages/project-upload.test.ts
@@ -516,6 +516,67 @@ describe("pages project upload", () => {
 	`);
 	});
 
+	it("should handle a very large number of assets", async () => {
+		const assets = new Set<string>();
+		// Create a large number of asset files to upload
+		for (let i = 0; i < 10_000; i++) {
+			const path = `file-${i}.txt`;
+			const content = `contents of file-${i}.txt`;
+			assets.add(content);
+			writeFileSync(path, content);
+		}
+
+		mockGetUploadTokenRequest(
+			"<<funfetti-auth-jwt>>",
+			"some-account-id",
+			"foo"
+		);
+
+		// Accumulate multiple requests then assert afterwards
+		const uploadedAssets = new Set<string>();
+		msw.use(
+			http.post("*/pages/assets/check-missing", async ({ request }) => {
+				const body = (await request.json()) as { hashes: string[] };
+
+				return HttpResponse.json(
+					{
+						success: true,
+						errors: [],
+						messages: [],
+						result: body.hashes,
+					},
+					{ status: 200 }
+				);
+			}),
+			http.post<never, UploadPayloadFile[]>(
+				"*/pages/assets/upload",
+				async ({ request }) => {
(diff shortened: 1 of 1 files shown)


GregBrimble:
Now we have the bot installed, I'm just testing Holopin really quickly here, sorry! Don't mind me!

packages/wrangler/src/__tests__/pages/project-upload.test.ts
```diff
@@ -516,6 +516,67 @@ describe("pages project upload", () => {
 	`);
 	});
 
+	it("should handle a very large number of assets", async () => {
+		const assets = new Set<string>();
+		// Create a large number of asset files to upload
+		for (let i = 0; i < 10_000; i++) {
```
CarmenPopoviciu, line 522: maybe worth making this number a tad more random, like `10_003` for example? Batches tend to be round numbers like `10`|`100`|`10_000`, so perhaps it'd be good to test against a large number of files that is not divisible by `batch`, and the division remainder is != 0

packages/wrangler/src/__tests__/pages/project-upload.test.ts
```diff
@@ -516,6 +516,67 @@ describe("pages project upload", () => {
...
+			const path = `file-${i}.txt`;
+			const content = `contents of file-${i}.txt`;
+			assets.add(content);
+			writeFileSync(path, content);
```
CarmenPopoviciu, line 526: any memory concerns here regarding writing 10k files? I understand the file size is small, but worth asking nonetheless

petebacondarwin: I'm not worried. The total memory used in this is going to be less than a megabyte.

packages/wrangler/src/__tests__/pages/project-upload.test.ts
```diff
@@ -516,6 +516,67 @@ describe("pages project upload", () => {
...
+					{
+						success: true,
+						errors: [],
+						messages: [],
+						result: body.hashes,
+					},
+					{ status: 200 }
+				);
+			}),
+			http.post<never, UploadPayloadFile[]>(
```
CarmenPopoviciu, line 551: how many times do we expect this endpoint to be called? should we assert that as well?

petebacondarwin: There are other tests that check how the bucketing works and exactly what requests are made.
This test is purely looking at whether the correct number of assets arrive at the server or not.

CarmenPopoviciu (approved):

PR #6014, opened 2024-06-12
ci: fix typos in C3 E2E actions

petebacondarwin:
It is possible that because of this we were not testing C3 against yarn!

### DIFF
--- a/.github/workflows/c3-e2e-dependabot.yml
+++ b/.github/workflows/c3-e2e-dependabot.yml
@@ -39,7 +39,7 @@ jobs:
     # Note: please keep this job in sync with the e2e one
     #       in .github/workflows/c3-e2e.yml
     needs: [get-dependabot-bumped-framework]
-    name: ${{ format('Dependabot E2E ({0}@{1})', matrix.pm.name, matrix.pm.version) }}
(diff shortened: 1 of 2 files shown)


CarmenPopoviciu (approved):

.github/workflows/c3-e2e-dependabot.yml
```diff
@@ -49,7 +49,7 @@ jobs:
             { name: npm, version: "0.0.0" },
             { name: pnpm, version: "9.1.3" },
             { name: bun, version: "1.0.3" },
-            { name: yarn, verison: "1.0.0" },
```
CarmenPopoviciu, line 52: 😅

packages/create-cloudflare/src/helpers/packageManagers.ts
```diff
@@ -27,7 +27,7 @@ export const detectPackageManager = () => {
 		version = process.env.TEST_PM_VERSION;
 		process.env.npm_config_user_agent = name;
 	}
-
+	//
```
petebacondarwin, line 30: Only here to trigger tests - will be removed before landing

Issue #3840, opened 2023-08-25
🐛 BUG: wrangler pages dev doesn't recompile worker when imported dependency changes


mariusGundersen:
### Which Cloudflare product(s) does this pertain to?

Pages

### What version(s) of the tool(s) are you using?

3.6.0

### What version of Node are you using?

18.13.0

### What operating system are you using?

Windows

### Describe the Bug

When running `wrangler pages dev` the worker is compiled automatically whenever a file in /functions is changed. However, if a dependency of one of those files changes it does not recompile.

For example, given
```js
// functions/test.js
import lib from '../lib/lib.js';

export async function onRequest() {
  return new Response(`<h1>${lib()}</h1>`, {
    headers: {
      'Content-Type': 'text/html'
    }
  })
}

// lib/lib.js
export default function () {
  return 'Testing';
}
```

Then when running `wrangler pages dev` any change to `functions/test.js` will cause the worker to recompile and refreshing the browser page will reflect that. If however the string in `lib/lib.js` is changed the worker is not recompiled and refreshing the page in the browser will not reflect those changes.

CarmenPopoviciu:
hi @mariusGundersen 

Thank you so much for raising this issue and providing all the helpful details that you did <3. 

I was able to reproduce the behaviour you described, so this is indeed an issue. In fact, we have a similar report in https://github.com/cloudflare/workers-sdk/issues/4824. As I mentione there

> reloading on dependency changes actually used to work, but we seem to have introduced a regression with
> https://github.com/cloudflare/workers-sdk/pull/3311. I need to double check this, but preliminary testing would seem to indicate that. I 
> don't have enough context on those changes to know whether the regression was intentional or not, and I do not know if re-enabling 
> esbuild's watch mode for Pages would not introduce the same performance issues as those reported in 
> https://github.com/cloudflare/workers-sdk/issues/3258. Something I need to look into. But I agree with the general sentiment that this 
> is something we should support

I will be picking up these two issues as part of https://github.com/cloudflare/workers-sdk/milestone/10 and will update here once the fix for this particular report is ready.

PR #6022, opened 2024-06-12
fix(wrangler): Watch for external dependencies changes in `pages dev`

CarmenPopoviciu:

The watch mode in `pages dev` for Pages Functions projects is currently partially broken, as it only watches for file system changes in the `/functions` directory, but not for changes in any of the Functions' dependencies. This means that given a Pages Function `math-is-fun.ts`, defined as follows:

```
import { ADD } from "../math/add";

export async function onRequest() {
  return new Response(`${ADD} is fun!`);
}
```

`pages dev` will reload for any changes in `math-is-fun.ts` itself, but not for any changes in `math/add.ts`, which is its dependency.

Similarly, `pages dev` will not reload for any changes in non-JS module imports, such as wasm/html/binary module
imports.

This PR fixes all these things, plus adds some extra polish to the `pages dev` watch mode experience.

Fixes [#3840](https://github.com/cloudflare/workers-sdk/issues/3840)

## Solutions we considered

Pages Functions projects cannot rely on esbuild's watch mode alone. That's because the watch mode that's built into esbuild is designed specifically for only triggering a rebuild when esbuild's build inputs are changed (see https://github.com/evanw/esbuild/issues/3705). With Functions, we would actually want to trigger a rebuild every time a new file is added to the "/functions" directory.

One solution would be to use an esbuild plugin, and "teach" esbuild to watch the "/functions" directory. But it's more complicated than that. When we build Functions, one of the steps is to generate the routes based on the file tree (see [generateConfigFileFromTree](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/pages/buildFunctions.ts#L74-L77)). These routes are then injected into the esbuild entrypoint (see [templates/pages-template-worker.ts](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/templates/pages-template-worker.ts)). Delegating the "/functions" dir watching to an esbuild plugin, would mean delegating the routes generation to that plugin as well. This gets very hairy very quickly.

Another solution, is to use a combination of dependencies watching, via esbuild, and file system watching, via chokidar. The downside of this approach is that a lot of syncing between the two watch modes must be put in place, in order to avoid triggering building Functions multiple times over one single change (like for example renaming file that's part of the
dependency graph).

Another solution, which is the one we opted for here, is to delegate file watching entirely to a file system watcher, chokidar in this case. While not entirely downside-free
  - we still rely on esbuild to provide us with the dependency graph
  - we need some logic in place to pre-process and filter the dependencies we pass to chokidar
  - we need to keep track of which dependencies are being watched 
  
this solution keeps all things watch mode in one place, makes things easier to read, reason about and maintain, separates Pages <-> esbuild concerns better, and gives all the flexibility we need.

## How we tested the changes
This PR comes with e2e tests, but in addition to them, we've manually tested these changes in the following scenarios:

- add/change/delete Function file in `functions`
- add/change/delete Function file in `functions/<subdir>/**/<subdir>`
- add/change/delete `_middleware.ts` in `functions`
- add/change/delete `_middleware.ts` in `functions/<subdir>/**/<subdir>`
- change in external dependency (eg `import { ADD } from "../<dir>/**/<dir>/add"`)
- change in external module (eg `import html from "../<dir>/**/<dir>/my-html.html"`)
- bulk file add/delete
- bulk directory add/delete

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: watch mode is not documented afaik

### DIFF
--- a/packages/wrangler/e2e/pages-dev.test.ts
+++ b/packages/wrangler/e2e/pages-dev.test.ts
@@ -134,71 +134,6 @@ describe("pages dev", () => {
 		);
 	});
 
-	it("should modify worker during dev session (_worker)", async () => {
-		const helper = new WranglerE2ETestHelper();
-		await helper.seed({
-			"_worker.js": dedent`
-					export default {
-						fetch(request, env) {
-							return new Response("Hello World!")
-						}
-					}`,
-		});
-		const port = await getPort();
-		const worker = helper.runLongLived(`wrangler pages dev --port ${port} .`);
-		const { url } = await worker.waitForReady();
-
-		await expect(
-			fetch(url).then((r) => r.text())
-		).resolves.toMatchInlineSnapshot('"Hello World!"');
-
-		await helper.seed({
-			"_worker.js": dedent`
-						export default {
-							fetch(request, env) {
-								return new Response("Updated Worker!")
-							}
-						}`,
-		});
-
-		await worker.waitForReload();
-
-		await expect(
-			fetch(url).then((r) => r.text())
-		).resolves.toMatchInlineSnapshot('"Updated Worker!"');
-	});
-
-	it("should modify worker during dev session (Functions)", async () => {
-		const helper = new WranglerE2ETestHelper();
-		const port = await getPort();
-		const worker = helper.runLongLived(`wrangler pages dev --port ${port} .`);
-
-		await helper.seed({
-			"functions/_middleware.js": dedent`
-					export async function onRequest() {
-						return new Response("Hello World!")
-					}`,
-		});
-
-		const { url } = await worker.waitForReady();
-
-		await expect(
-			fetch(url).then((r) => r.text())
-		).resolves.toMatchInlineSnapshot('"Hello World!"');
-
-		await helper.seed({
-			"functions/_middleware.js": dedent`
-					export async function onRequest() {
-						return new Response("Updated Worker!")
-					}`,
-		});
-
-		await worker.waitForReload();
-
-		await expect(
-			fetch(url).then((r) => r.text())
-		).resolves.toMatchInlineSnapshot('"Updated Worker!"');
-	});
 	it("should support wrangler.toml", async () => {
 		const helper = new WranglerE2ETestHelper();
 		await helper.seed({
@@ -313,61 +248,6 @@ describe("pages dev", () => {
 		await expect(
 			fetch(url).then((r) => r.text())
 		).resolves.toMatchInlineSnapshot('"Updated Worker!"');
-	});
-
-	it("should support modifying _routes.json during dev session", async () => {
-		const helper = new WranglerE2ETestHelper();
-		await helper.seed({
-			"_worker.js": dedent`
-					export default {
-						async fetch(request, env) {
-							const url = new URL(request.url);
-							if (url.pathname === "/foo") {
-								return new Response("foo");
-							}
-							if (url.pathname === "/bar") {
-								return new Response("bar");
-							}
-							return new Response("Hello _routes.json")
-						}
-					}`,
-			"_routes.json": dedent`
-				{
-					"version": 1,
-					"include": ["/foo", "/bar"],
-					"exclude": []
-				}
-			`,
-			"index.html": dedent`
-				hello world
-			`,
-		});
-		const port = await getPort();
-		const worker = helper.runLongLived(`wrangler pages dev --port ${port} .`);
-		const { url } = await worker.waitForReady();
-
-		const foo = await fetchText(`${url}/foo`);
-		expect(foo).toMatchInlineSnapshot('"foo"');
-
-		const bar = await fetchText(`${url}/bar`);
-		expect(bar).toMatchInlineSnapshot('"bar"');
-
-		await helper.seed({
-			"_routes.json": dedent`
-				{
-					"version": 1,
-					"include": ["/foo"],
-					"exclude": ["/bar"]
-				}
-			`,
-		});
-		await worker.waitForReload();
-
-		const foo2 = await fetchText(`${url}/foo`);
-		expect(foo2).toMatchInlineSnapshot('"foo"');
-
-		const bar2 = await fetchText(`${url}/bar`);
-		expect(bar2).toMatchInlineSnapshot('"hello world"');
 	});
 
 	it("should validate _routes.json during dev session, and fallback to default value", async () => {
@@ -593,6 +473,7 @@ describe("pages dev", () => {
 
 	it("should pick up wrangler.toml configuration even in cases when `pages_build_output_dir` was not specified, but the <directory> command argument was", async () => {
 		const helper = new WranglerE2ETestHelper();
+
 		await helper.seed({
 			"public/_worker.js": dedent`
 					export default {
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/pages/utils.ts
```diff
@@ -75,3 +75,22 @@ export function getPagesTmpDir(): string {
 	tmpDirCacheProjectRoot = projectRoot;
 	return tmpDirCache;
 }
+
+/**
+ * Creates a basic debounced function that delays invoking `fn` until after
+ * `delayMs` milliseconds have elapsed since the last time the debounced
+ * function was invoked.
+ */
+export function debounce(fn: () => void, delayMs = 100) {
```
CarmenPopoviciu, line 84: happy to place this somewhere else if needed

packages/wrangler/src/pages/dev.ts
```diff
@@ -497,9 +607,28 @@ export const Handler = async (args: PagesDevArguments) => {
...
 					} else {
-						throw e;
+						/*
+						 * don't break developer flow in watch mode by throwing an error
+						 * here. Many times errors will be just the result of unfinished
+						 * typing. Instead, log the error, point out we are still serving
+						 * the last successfully built Functions, and allow developers to
+						 * write their code to completion
+						 */
+						logger.warn(
```
CarmenPopoviciu, line 617: should this be an error?

packages/wrangler/src/pages/dev.ts
```diff
@@ -509,9 +638,13 @@ export const Handler = async (args: PagesDevArguments) => {
...
 				// Resolve anyway and run without Functions
 				onEnd();
-				// Turn off Functions
 				usingFunctions = false;
 			} else {
+				/*
+				 * do not start the `pages dev` session if we encounter errors
+				 * while attempting to build Functions. These flag underlying
+				 * issues in the Functions code, and should be addressed before
+				 */
```
CarmenPopoviciu, line 647: should we throw a `FatalError`?

packages/wrangler/src/pages/dev.ts
```diff
@@ -462,9 +462,54 @@ export const Handler = async (args: PagesDevArguments) => {
...
+		// always watch the "/functions" directory
+		const watcher = watch([functionsDirectory], {
+			persistent: true,
+			ignoreInitial: true,
+		});
+
 		try {
 			const buildFn = async () => {
-				await buildFunctions({
+				const crrBundleDependencies = new Set();
```
penalosa, line 510: Can we name this something more verbose? Is `crr` `current`?

packages/wrangler/src/pages/dev.ts
```diff
@@ -476,15 +521,80 @@ export const Handler = async (args: PagesDevArguments) => {
...
+					 *   - everything in "./.wrangler", as it's mostly cache and
+					 *     temporary files
+					 *   - any bundle dependencies we are already watching
+					 *   - anything outside of the current working directory, since we
+					 *     are expecting `wrangler pages dev` to be run from the Pages
+					 *     project root folder
+					 */
+					if (
+						!resolvedDep.match(/\/functions\/./) &&
+						!resolvedDep.match(/\.wrangler\/./) &&
```
penalosa, line 539: Does this affect middleware/templates? I think we sometimes put those in `.wrangler`?

packages/wrangler/src/pages/dev.ts
```diff
@@ -476,15 +521,80 @@ export const Handler = async (args: PagesDevArguments) => {
...
+					 * EXCLUDE:
+					 *   - the "/functions" directory because we're already watching it
+					 *   - everything in "./.wrangler", as it's mostly cache and
+					 *     temporary files
+					 *   - any bundle dependencies we are already watching
+					 *   - anything outside of the current working directory, since we
+					 *     are expecting `wrangler pages dev` to be run from the Pages
+					 *     project root folder
+					 */
+					if (
```
penalosa, line 537: How does this deal with things like packages in node_modules? Should we exclude those too?

packages/wrangler/src/pages/dev.ts
```diff
@@ -476,15 +521,80 @@ export const Handler = async (args: PagesDevArguments) => {
...
+					}
+				}
+
+				/*
+				 *`bundle.dependencies` and `bundle.modules` will always contain the
+				 * latest dependency list of the current bundle. If we are currently
+				 * watching any dependency files not in that list, we should remove
+				 * them, as they are no longer relevant to the compiled Functions.
+				 */
+				watchedBundleDependencies.forEach(async (path) => {
```
penalosa, line 570: Does `watcher.getWatched()` work for this use case? I wonder if it would allow you to remove the `watchedBundleDependencies` `Set`

packages/wrangler/src/pages/dev.ts
penalosa: I feel like this should be an error? AFAIK it's a recoverable error (i.e. saving the file again should re-trigger the build), but it feels like we shouldn't hide build errors?

packages/wrangler/src/pages/functions/buildWorker.ts
```diff
@@ -349,17 +349,34 @@ export function buildNotifierPlugin(onEnd: () => void): Plugin {
...
+						const warningMessages = result.warnings.reduce(
+							(acc, warning) => `${acc}\n` + `▶︎ ${warning.text}`,
+							""
+						);
+
+						logger.warn(
+							`Got ${result.warnings.length} warning${result.warnings.length > 1 ? "s" : ""} when compiling Worker:\n` +
+								`${warningMessages}`
+						);
+					}
```
penalosa, line 378: Could we lean on esbuild's formatting for build errors here, in the same way as we do for Workers proper? https://github.com/cloudflare/workers-sdk/blob/ab763c37454f45a6d7cd2be7963a863c49fe2d49/packages/wrangler/src/logger.ts#L149

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: yeah, what we'll see is smth like
```
[WARNING] Error while building Functions.....

<error details>
```

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: it is! will do!

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -639,4 +513,231 @@ describe("pages dev", () => {
 			);
 		}
 	);
+
+	describe("watch mode", () => {
+		e2eTest(
```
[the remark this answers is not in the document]
CarmenPopoviciu: ✅

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: converted to `logger.error` ✅

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -639,4 +513,231 @@ describe("pages dev", () => {
...
+						}`,
+				});
+
+				await seed({
+					"functions/hi.js": dedent`
+						import { hi } from "../utils/greetings"
+						export async function onRequest() {
+							return new Response(hi)
+						}`,
+				});
```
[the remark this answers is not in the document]
CarmenPopoviciu: ✅

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -639,4 +513,231 @@ describe("pages dev", () => {
...
+
+				const { url } = await waitForReady(worker);
+
+				await expect(
+					fetch(`${url}/greetings/hello`).then((r) => r.text())
```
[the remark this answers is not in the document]
CarmenPopoviciu: ✅

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: ✅

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: FWICT`watcher.getWatched()` seem to be rather reliable (https://github.com/paulmillr/chokidar/issues?q=is%3Aissue+getwatched+is%3Aclosed), which is why I avoided it. But we can do something like `watcher.unwatch([file1, file2, ..., file_n])`, if that's what you meant.

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: yes it indeed does. those will be filtered out. It *seemed* like that would be OK, since technically that's not something users should tamper with? In my head the files that'd need to be watched here, are user modified files (with "modified" being add/change/delete). Furthermore, but I'd have to dbl check this again, I seem to remember that not filtering that `.wrangler/tmp` folder would result in some circular dependency situation? where chokidar would enter a perpetual state of triggering change events

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: I thought abt that, but decided not to exclude them. My main concern was hitting perf bottleneck due to watching too many files. I have no feel whatsoever of how many deps a project has on average. If we are worried abt perf hits or other things, we can def exclude. FWICT, watching `node_modules` is not an expectation users have?

packages/wrangler/src/pages/dev.ts
```diff
@@ -476,42 +521,130 @@ export const Handler = async (args: PagesDevArguments) => {
...
+						if (!watchedBundleDependencies.has(resolvedDep)) {
+							watchedBundleDependencies.add(resolvedDep);
+							watcher.add(resolvedDep);
+						}
+					}
+				}
+
+				/*
+				 *`bundle.dependencies` and `bundle.modules` will always contain the
+				 * latest dependency list of the current bundle. If we are currently
```
CarmenPopoviciu, line 566: soooo...turns out, while testing the `_worker.js` work, that there is at least one corner case where this seems not to be true.

Say our Function `math-is-fun.ts` imports `../math/add.ts` and calls `add()`. So far so good. Now imagine we delete the contents of `add.ts` but leave everything else as it is. We'll get a compile warning saying:

`Import "add" will always be undefined because the file "../math/add.ts" has no exports`

which is correct. At this point, IMHO, `add.ts` is technically still a dependency of `math-is-fun.ts` (though maybe we can argue abt that later), so my expectation would be that if we were to undo the changes we made in `add.ts`, that would trigger a rebuild. This works as expected if esbuild was in `watch` mode, but it does not trigger anything with the current implementation. The reason for that is that the `metafile` returned by esbuild does not list `add.ts` as an `input` to the entrypoint, which means we won't return it as part of `bundle.dependecies` 😕. The returned metafile BTW, looks exactly the same in both esbuild/chokidar watch modes. So, it's either that esbuild watches more files than what it returns in the metafile, or we're making some wrong assumptions wehn we process that metafile to return `bundle.dependencies`, or there is a plugin somewhere that strips out empty imports? (meaning of 0kb in size).

Whatever it is, the finding is a tad worrying...cause it makes me wonder...what other corner case I didn't think of yet, would result in a similar behaviour

it's never easy is it 😅 🍦 (`i-scream` 😉 )

packages/wrangler/src/pages/functions/buildWorker.ts
CarmenPopoviciu: ✅

packages/wrangler/src/pages/dev.ts
penalosa: That makes sense! Users should ideally not be modifying these files directly

packages/wrangler/src/pages/dev.ts
penalosa: Ah, I didn't realise it was so unreliable—in that case your `Set` solution is great!

penalosa (approved):

packages/wrangler/src/pages/dev.ts
```diff
@@ -462,9 +462,54 @@ export const Handler = async (args: PagesDevArguments) => {
...
+		 * watching entirely to a file system watcher, chokidar in this case. While
+		 * not entirely downside-free
+		 *   - we still rely on esbuild to provide us with the dependency graph
+		 *   - we need some logic in place to pre-process and filter the dependencies
+		 *     we pass to chokidar
+		 *   - we need to keep track of which dependencies are being watched
+		 * this solution keeps all things watch mode in one place, makes things easier
+		 * to read, reason about and maintain, separates Pages <-> esbuild concerns
+		 * better, and gives all the flexibility we needed.
+		 */
```
andyjessop, line 501: Fantastic comment 🥇

packages/wrangler/src/pages/dev.ts
andyjessop: Similar to here, I think it makes sense. If a user has an error, it would be good to let them know as soon as possible: https://github.com/cloudflare/workers-sdk/pull/6150/files/b4006823e7ddeb4b5d8e656ba7f2f63756fe3602#diff-c90ff1ca6d929baaafb7d67d6d6ccd573593762adf4e7a9b6917343ceeceb68bR443

andyjessop (approved):
LGTM 🔥

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: ✅

CarmenPopoviciu:
thank you everyone for reviewing! It is much much appreciated <3. Merged! 🎉

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: raised here: https://github.com/cloudflare/workers-sdk/issues/6182

PR #6025, opened 2024-06-13
fix(wrangler): avoid path collisions between performance and Performance polyfills

IgorMinar:
It turns out that ESBuild paths are case insensitive, which can result in path collisions between polyfills for globalThis.performance and globalThis.Performance.

This change ensures that we encode all global names to lowercase and decode them appropriately.

The version of unenv used at HEAD currently doesn't suffer from this path collision, but if not fixed the change introduced in https://github.com/unjs/unenv/pull/257/files would result in Node.js API coverage regressions. This change prevents those regressions from happening.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because:

### DIFF
--- a/.changeset/lazy-cycles-pull.md
+++ b/.changeset/lazy-cycles-pull.md
@@ -1,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+fix: avoid path collisions between performance and Performance Node.js polyfills
+
+It turns out that ESBuild paths are case insensitive, which can result in path collisions between polyfills for `globalThis.performance` and `globalThis.Performance`, etc.
+
+This change ensures that we encode all global names to lowercase and decode them appropriately.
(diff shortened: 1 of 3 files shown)


petebacondarwin (approved):
This fix looks fine to land.
But I think that actually this is not the best approach for these virtual files.
Rather than using virtual fake paths (which in extreme cases could collide with real paths), I think the correct approach is to use esbuild namespaces here to filter these resolves to loaders.
We might still need to encode/decode upper case letters from globals to distinguish between globals that only differ in case, but we could avoid having to use `getBasePath()` and the concern of path collision.

petebacondarwin:
It needs a changeset before I merge...

PR #6029, opened 2024-06-13
Normalize some dependencies in workers-sdk

threepointone:
This is the first of a few expected patches that normalize dependency versions, This normalizes `undici`, `concurrently`, `@types/node`, `react`, `react-dom`, `@types/react`, `@types/react-dom`, `eslint`, `typescript`. There are no functional code changes (but there are a couple of typecheck fixes).

--- 

Let's see what CI says.


packages/miniflare/scripts/build.mjs
```diff
@@ -4,7 +4,6 @@ import esbuild from "esbuild";
 import { getPackage, pkgRoot } from "./common.mjs";
 
 const argv = process.argv.slice(2);
-const watch = argv[0] === "watch";
```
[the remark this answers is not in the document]
petebacondarwin: later .... 😄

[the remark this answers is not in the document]
petebacondarwin: Tracking at https://github.com/cloudflare/workers-sdk/issues/6042

templates/worker-sites-react/package.json
```diff
@@ -27,8 +27,8 @@
 	},
 	"dependencies": {
 		"@cloudflare/kv-asset-handler": "workspace:*",
-		"react": "^16.9.0",
-		"react-dom": "^16.9.0"
+		"react": "^18.2.0",
+		"react-dom": "^18.2.0"
```
petebacondarwin, line 31: I don't think we have any tests of this template.
Is it definitely going to work with a big major bump like this?

petebacondarwin (approved):

templates/worker-sites-react/package.json
threepointone: tested, works fine

packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -455,6 +455,7 @@ describe.concurrent(`E2E: Web frameworks`, () => {
 					await verifyBuildScript(framework, projectPath, logStream);
 					await storeDiff(framework, projectPath);
 				} catch (e) {
+					console.error("ERROR", e);
```
CarmenPopoviciu, line 458: @threepointone was this meant to be `console.error` or `logger.error`?

CarmenPopoviciu: oh, totally missed this was in `create-cloudflare`. Is the log relevant here then? I see we're asserting on the fail message...not sure if that needs to be logged for that assertion to pass

threepointone: I added it while debugging, but it's not a bad idea to keep it here, because otherwise you have to see logs, which are hard to do in CI (unless I'm mistaken?)

PR #6006, opened 2024-06-11
Remove middleware warning

penalosa:
## What this PR solves / how to test

Alternative fix to https://github.com/cloudflare/workers-sdk/pull/5999

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [ ] Required / Maybe required
  - [x] Not required because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Change to testing coe
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Not user-facing

PR #6043, opened 2024-06-14
fix: avoid esbuild warning when running dev/bundle

threepointone:
(Another simpler take on https://github.com/cloudflare/workers-sdk/pull/5999, inspired by https://github.com/cloudflare/workers-sdk/pull/6006)

I've been experimenting with esbuild 0.21.4 with wrangler. It's mostly been fine. But I get this warning every time

```
▲ [WARNING] Import "__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__" will always be undefined because there is no matching export in "src/index.ts" [import-is-undefined]

    .wrangler/tmp/bundle-Z3YXTd/middleware-insertion-facade.js:8:23:
      8 │ .....(OTHER_EXPORTS.__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__ ?? []),
        ╵
```

This is because esbuild@0.18.5 enabled a warning by default whenever an undefined import is accessed on an imports object. However we abuse imports to inject stuff in `middleware.test.ts`. A simple fix is to only inject that code in tests.

### DIFF
--- a/packages/wrangler/src/deployment-bundle/apply-middleware.ts
+++ b/packages/wrangler/src/deployment-bundle/apply-middleware.ts
@@ -68,7 +68,7 @@ export async function applyMiddlewareLoaderFacade(
 				export * from "${prepareFilePath(entry.file)}";
 
 				export const __INTERNAL_WRANGLER_MIDDLEWARE__ = [
-					...(OTHER_EXPORTS.__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__ ?? []),
+					${process.env.NODE_ENV === "test" ? `...(OTHER_EXPORTS.__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__ ?? []),` : ""}
 					${middlewareFns}
 				]
 				export default worker;
--- a/.changeset/rotten-pears-hope.md
+++ b/.changeset/rotten-pears-hope.md
@@ -1,0 +1,17 @@
+---
+"wrangler": patch
+---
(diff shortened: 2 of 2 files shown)


petebacondarwin (approved):
Love it!

packages/wrangler/src/deployment-bundle/apply-middleware.ts
```diff
@@ -68,7 +68,7 @@ export async function applyMiddlewareLoaderFacade(
 				export * from "${prepareFilePath(entry.file)}";
 
 				export const __INTERNAL_WRANGLER_MIDDLEWARE__ = [
-					...(OTHER_EXPORTS.__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__ ?? []),
+					${process.env.NODE_ENV === "test" ? `...(OTHER_EXPORTS.__INJECT_FOR_TESTING_WRANGLER_MIDDLEWARE__ ?? []),` : ""}
```
petebacondarwin, line 71: Does Vitest set this?

petebacondarwin: Yes it does!

https://github.com/vitest-dev/vitest/blob/e2d69c63b290841fff741a30b0025405e0d4c68f/docs/guide/migration.md?plain=1#L282

threepointone: yup

PR #6050, opened 2024-06-15
Normalize more deps

threepointone:
This is the last of the patches that normalize dependencies across the codebase. In this batch: `ws`, `zod` , `rimraf`, `@types/rimraf`, `ava`, `source-map`, `glob`, `cookie`, `@types/cookie`, `@microsoft/api-extractor`, `@types/mime`, `@types/yargs`, `devtools-protocol`, `@vitest/ui`, `strip-ansi`

This patch also sorts dependencies in every `package.json`

--- 

There are 2 remaining after this; `execa` and `vitest`. Changing their versions seems to break the build, so those will be for the next PR.


packages/miniflare/test/setup.mjs
```diff
@@ -42,3 +43,11 @@ process.on("exit", () => {
 	);
 	throw new Error(message);
 });
+
+// https://github.com/avajs/ava/discussions/3259
```
threepointone, line 47: I had to make this change to update miniflare's ava to v6

packages/wrangler/package.json
```diff
@@ -159,16 +159,16 @@
...
 		"signal-exit": "^3.0.7",
-		"strip-ansi": "^7.0.1",
+		"strip-ansi": "^7.1.0",
 		"supports-color": "^9.2.2",
 		"timeago.js": "^4.0.2",
 		"ts-dedent": "^2.2.0",
 		"ts-json-schema-generator": "^1.5.0",
 		"undici": "^5.28.4",
 		"update-check": "^1.5.4",
-		"vitest": "^1.6.0",
```
threepointone, line 169: I had to bring this version down to normalize, we'll bring it back up when it works for everything

threepointone:
Wow, actual E2E errors on every platform? I'll have a look on monday.

packages/wrangler/package.json
threepointone: this isn't true anymore, reverted this one.

petebacondarwin (approved):

PR #6052, opened 2024-06-15
Add `.wrangler` and `.DS_Store` to `.gitignore` generated by `wrangler init`

threepointone:
^ Like it says. I know we're deprecating `init` sometime, but for those who still use it (like me), a small QOL improvement.


petebacondarwin (approved):
To be clear for future reviewers. This file is copied into newly generated Workers projects created via `wrangler init`. See https://github.com/cloudflare/workers-sdk/blob/c643a8193a3c0739b33d3c0072ae716bc8f1565b/packages/wrangler/src/init.ts#L337

PR #6055, opened 2024-06-15
chore: change vitest config to be ESM (for sure) to avoid warnings

petebacondarwin:
Without this we often see:

```
 The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
```


threepointone:
omg yes

threepointone (approved):

threepointone (approved):

petebacondarwin:
An alternative solution to changing the file name is to set `"type": "module"` in various `package.json` files but that seemed like a more dangerous change.

CarmenPopoviciu (approved):

andyjessop:
> An alternative solution to changing the file name is to set `"type": "module"` in various `package.json` files but that seemed like a more dangerous change.

I wonder if it's worth doing just that, as a fail-fast mechanism. If users have `"type": "module"` and it breaks things, we likely want to know about it as soon as possible.

petebacondarwin:
> > An alternative solution to changing the file name is to set `"type": "module"` in various `package.json` files but that seemed like a more dangerous change.
> 
> I wonder if it's worth doing just that, as a fail-fast mechanism. If users have `"type": "module"` and it breaks things, we likely want to know about it as soon as possible.

By dangerous change, I meant that switching the default format for the packages in our repo could have consequences outside of Vitest that I didn't want to start messing with. Not that we potentially don't support such packages.

andyjessop (approved):

RamIdeas (approved):

PR #6057, opened 2024-06-15
test: fix vitest-pool-workers tests that rely on local workspace packages

petebacondarwin:

These tests create a fake package that relies upon the local versions of vitest-pool-workers and its dependencies. To make this work we package up all the local dependencies and add them to pnpm overrides.

Unfortunately we missed the transitive `@cloudflare/kv-asset-handler` dependency, so when that got a version bump the tests broke.

This is the simple, naive fix - adding in this package to those being packed. We ought to find a better solution that doesn't require manual dependency tracking.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: it is fixing tests
  - [ ] I don't know
  - [x] Not required because: it is only touching non-e2e tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: just a test fix
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: just a test fix

### DIFF
--- a/packages/vitest-pool-workers/test/global-setup.ts
+++ b/packages/vitest-pool-workers/test/global-setup.ts
@@ -28,9 +28,13 @@ export default async function ({ provide }: GlobalSetupContext) {
 		await fs.mkdtemp(path.join(os.tmpdir(), "vitest-pool-workers"))
 	);
 
-	// Pack `miniflare`, `wrangler` and `vitest-pool-workers` into tarballs
+	// Pack `miniflare`, `wrangler`, `kv-asset-handler` and `vitest-pool-workers` into tarballs
 	const packDestinationPath = path.join(tmpPath, "packed");
 	const packagesRoot = path.resolve(__dirname, "../..");
+	const kvAssetHandlerTarballPath = packPackage(
+		packDestinationPath,
+		path.join(packagesRoot, "kv-asset-handler")
+	);
 	const miniflareTarballPath = packPackage(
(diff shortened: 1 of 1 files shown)


packages/vitest-pool-workers/test/global-setup.ts
```diff
@@ -28,9 +28,14 @@ export default async function ({ provide }: GlobalSetupContext) {
...
 
-	// Pack `miniflare`, `wrangler` and `vitest-pool-workers` into tarballs
+	// Pack `miniflare`, `wrangler`, `kv-asset-handler` and `vitest-pool-workers` into tarballs
 	const packDestinationPath = path.join(tmpPath, "packed");
 	const packagesRoot = path.resolve(__dirname, "../..");
+	const kvAssetHandlerTarballPath = packPackage(
+		packDestinationPath,
+		path.join(packagesRoot, "kv-asset-handler")
+	);
+	console.log(path.join(packagesRoot, "kv-asset-handler"), kvAssetHandlerTarballPath)
```
threepointone, line 38: stray log?

threepointone (approved):
looks good, ship it!

packages/vitest-pool-workers/test/global-setup.ts
petebacondarwin: oops

PR #6060, opened 2024-06-15
run react eslint rules only where needed

threepointone:
Our default eslint config includes react linting, and was applied to every package. However that meant we kept seeing warnings about react not being detected etc. This patch splits out the react part, and applies it only where required.

### DIFF
--- a/packages/wrangler/.eslintrc.js
+++ b/packages/wrangler/.eslintrc.js
@@ -1,6 +1,6 @@
 module.exports = {
 	root: true,
-	extends: ["@cloudflare/eslint-config-worker"],
+	extends: ["@cloudflare/eslint-config-worker/react"],
 	ignorePatterns: [
 		"vendor",
 		"*-dist",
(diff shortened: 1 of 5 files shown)


packages/wrangler/.eslintrc.js
```diff
@@ -1,13 +1,18 @@
...
 		"*-dist",
 		"pages/functions/template-worker.ts",
 		"templates",
 		"emitted-types",
 	],
+	settings: {
+		react: {
+			version: "detect",
+		},
+	},
```
petebacondarwin, line 15: Why note have this setting in `packages/eslint-config-worker/react.js`?

petebacondarwin (approved):

packages/wrangler/.eslintrc.js
petebacondarwin: @threepointone will check this and follow up if necessary

PR #6065, opened 2024-06-16
Update actions/upload-artifact from v3 to v4

threepointone:
This removes the warning about it using an outdated version of node. 

I don't think we need to make any changes here, but here's a reference https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md


threepointone:
ok, maybe not, the paths have to be changed. I'll leave it for next week.

threepointone:
Closing this one because it's a little more fiddly.

PR #6073, opened 2024-06-18
Added `d1 export --local` support

geelen:
This uses a new miniflare-only debug pragma `PRAGMA miniflare_d1_export` to trigger the dump as D1 provides no in-Worker API (which Miniflare uses) to generate exports.

I'm open to other ways to implement this.

Fixes limitation with the previous iteration of the export functionality. Previously, all `d1 export --local` commands would fail.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: covered by local tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: documentation doesn't say `--local` doesn't work, it just emphasises `--remote` (which still makes sense to be the default).

### DIFF
--- a/packages/miniflare/src/workers/d1/dumpSql.ts
+++ b/packages/miniflare/src/workers/d1/dumpSql.ts
@@ -1,0 +1,160 @@
+// NOTE: this function duplicates the logic inside SQLite's shell.c.in as close
+// as possible, with any deviations noted.
+import { SqlStorage } from "@cloudflare/workers-types/experimental";
+
+export function* dumpSql(
+	db: SqlStorage,
+	options?: {
+		noSchema?: boolean;
+		noData?: boolean;
+		tables?: string[];
+	}
+) {
+	yield `PRAGMA defer_foreign_keys=TRUE;`;
+
+	// Empty set means include all tables
+	const filterTables = new Set(options?.tables || []);
+	const { noData, noSchema } = options || {};
+
+	// Taken from SQLite shell.c.in https://github.com/sqlite/sqlite/blob/105c20648e1b05839fd0638686b95f2e3998abcb/src/shell.c.in#L8463-L8469
+	// @ts-ignore (SqlStorageStatement needs to be callable)
+	const tables_cursor = db.prepare(`
+    SELECT name, type, sql 
+      FROM sqlite_schema AS o 
+    WHERE (true) AND type=='table' 
+      AND sql NOT NULL 
+    ORDER BY tbl_name='sqlite_sequence', rowid;
+  `)();
+	const tables: any[] = Array.from(tables_cursor);
+
+	for (const { name: table, sql } of tables) {
+		if (filterTables.size > 0 && !filterTables.has(table)) continue;
+
+		if (table === "sqlite_sequence") {
+			if (!noSchema) yield `DELETE FROM sqlite_sequence;`;
+		} else if (table.match(/^sqlite_stat./)) {
+			// This feels like it should really appear _after_ the contents of sqlite_stat[1,4] but I'm choosing
+			// to match SQLite's dump output exactly so writing it immediately like they do.
+			if (!noSchema) yield `ANALYZE sqlite_schema;`;
+		} else if (sql.startsWith(`CREATE VIRTUAL TABLE`)) {
+			throw new Error(
+				`D1 Export error: cannot export databases with Virtual Tables (fts5)`
+			);
+		} else if (table.startsWith("_cf_") || table.startsWith("sqlite_")) {
+			continue;
+		} else {
+			// SQLite dump has an extremely weird behaviour where, if the table was explicitly
+			// quoted i.e. "Table", then in the dump it has `IF NOT EXISTS` injected. I don't understand
+			// why, but on the off chance there's a good reason to I am following suit.
+			if (sql.match(/CREATE TABLE ['"].*/)) {
+				if (!noSchema) yield `CREATE TABLE IF NOT EXISTS ${sql.substring(13)};`;
+			} else {
+				if (!noSchema) yield `${sql};`;
+			}
+		}
+
+		if (noData) continue;
+		const columns_cursor = db.exec(`PRAGMA table_info="${table}"`);
+		const columns = Array.from(columns_cursor);
+		const select = `SELECT ${columns.map((c) => c.name).join(", ")}
+                            FROM "${table}";`;
+		const rows_cursor = db.exec(select);
+		for (const dataRow of rows_cursor.raw()) {
+			const formattedCells = dataRow.map((cell: unknown, i: number) => {
+				const colType = columns[i].type;
+				const cellType = typeof cell;
+				if (cell === null) {
+					return "NULL";
+				} else if (colType === "INTEGER" || cellType === "number") {
+					return cell;
+				} else if (colType === "TEXT" || cellType === "string") {
+					return outputQuotedEscapedString(cell);
+				} else if (cell instanceof ArrayBuffer) {
+					return `X'${Array.prototype.map
+						.call(new Uint8Array(cell), (b) => b.toString(16))
+						.join("")}'`;
+				} else {
+					console.log({ colType, cellType, cell, column: columns[i] });
+					return "ERROR";
+				}
+			});
+
+			yield `INSERT INTO ${sqliteQuote(table)} VALUES(${formattedCells.join(",")});`;
+		}
+	}
+
+	if (!noSchema) {
+		// Taken from SQLite shell.c.in https://github.com/sqlite/sqlite/blob/105c20648e1b05839fd0638686b95f2e3998abcb/src/shell.c.in#L8473-L8478
+		const rest_of_schema = db.exec(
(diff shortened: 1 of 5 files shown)


packages/miniflare/src/workers/d1/dumpSql.ts
```diff
@@ -0,0 +1,152 @@
+// NOTE: this function duplicates the logic inside SQLite's shell.c.in as close
+// as possible, with any deviations noted.
+import { SqlStorage } from "@cloudflare/workers-types/experimental";
+
+export function* dumpSql(
```
andyjessop, line 5: Is it worth building the array here instead of using a generator? It could simplify the code and would mean it wouldn't need to convert to an array in the caller. Or is there another reason why the sql return value needs to be in iterable?

packages/miniflare/src/workers/d1/dumpSql.ts
```diff
@@ -0,0 +1,152 @@
...
+				if (cell === null) {
+					return "NULL";
+				} else if (colType === "INTEGER" || cellType === "number") {
+					return cell;
+				} else if (colType === "TEXT" || cellType === "string") {
+					// Reproduce `output_quoted_escaped_string` from https://github.com/sqlite/sqlite/blob/master/src/shell.c.in#L1799-L1862
+					let lfs = false;
+					let crs = false;
+					// First escape any ' as ''
+					let output_string = `'${(cell as string).replace(
```
andyjessop, line 75: Nit: this could potentially be extracted to a separate function to make the `sqlDump` function more readable.

packages/miniflare/src/workers/d1/dumpSql.ts
```diff
@@ -0,0 +1,152 @@
...
+    ORDER BY tbl_name='sqlite_sequence', rowid;
+  `)();
+	const tables: any[] = Array.from(tables_cursor);
+
+	for (const { name: table, sql } of tables) {
+		if (filterTables.size > 0 && !filterTables.has(table)) continue;
+
+		if (table === "sqlite_sequence") {
+			if (!noSchema) yield `DELETE FROM sqlite_sequence;`;
+		} else if (table.startsWith(`sqlite_stat`)) {
```
justin-mp, line 35: This matches too many table names!  It's supposed to only match sqlite_stat plus exactly one more character.   The code in shell.in.c's `dump_callback` is `sqlite3_strglob("sqlite_stat?", zTable)==0`, and the `?` operator in a glob matches exactly one character.  I don't know if JS has globbing, but the equivalent regexp would be `/sqlite_stat./`.

packages/miniflare/src/workers/d1/dumpSql.ts
```diff
@@ -0,0 +1,152 @@
...
+		tables?: string[];
+	}
+) {
+	yield `PRAGMA defer_foreign_keys=TRUE;`;
+
+	// Empty set means include all tables
+	const filterTables = new Set(options?.tables || []);
+	const { noData, noSchema } = options || {};
+
+	// Taken from SQLite shell.c.in https://github.com/sqlite/sqlite/blob/master/src/shell.c.in#L8444-L8450
```
justin-mp, line 19: Can you put a permalink here?  I just pulled master and this SQL is on line 8464 now.

packages/miniflare/src/workers/d1/dumpSql.ts
```diff
@@ -0,0 +1,152 @@
...
+					console.log({ colType, cellType, cell, column: columns[i] });
+					return "ERROR";
+				}
+			});
+
+			yield `INSERT INTO ${sqliteQuote(table)} VALUES(${formattedCells.join(",")});`;
+		}
+	}
+
+	if (!noSchema) {
```
justin-mp, line 108: Does it matter that this query appears first in shell.c.in?

packages/wrangler/src/__tests__/d1/export.test.ts
```diff
@@ -14,19 +14,90 @@ describe("execute", () => {
...
 			`Missing required argument: output`
 		);
 	});
 
-	it("should reject --local mode (for now)", async () => {
-		await expect(
-			runWrangler("d1 export db --local --output /tmp/test.sql")
-		).rejects.toThrowError(
-			`Local imports/exports will be coming in a future version of Wrangler.`
+	it("should handle local", async () => {
```
justin-mp, line 25: You should probably add tests for the edge cases like quoted names adding the "IF NOT EXISTS" as well as properly quoting things like keywords.

justin-mp (commented):
For my own education, does this PR mean that `PRAGMA miniflare_d1_export(?,?,?);` is part of our public API now?  I don't think we want it to be so.  How do we prevent people from writing that in their own queries?

packages/miniflare/src/workers/d1/dumpSql.ts
geelen: This is for parity with internal D1 code which uses a generator as it needs to stream the results out with backpressure. For Miniflare we keep the same code (duplicated as it didn't feel worth extracting a shared package for a single file that should very rarely change) but accumulate into an array in the caller.

packages/miniflare/src/workers/d1/dumpSql.ts
geelen: Yeah I wrote this file while trying my best to copy line-by-line the functionality in shell.c.in, but I think it makes sense to then slightly restructure it for readability after the fact.

packages/miniflare/src/workers/d1/dumpSql.ts
geelen: Though, in saying that, it would be nice to have a separate repo with a bunch of unit tests that cover regressions. At the moment that's inside D1's internal codebase, so I think I'll add a cleanup ticket to extract this out and unify the references. It would be nice to have this released in the mean time though.

packages/miniflare/src/workers/d1/dumpSql.ts
geelen: Good catch! Fixed.

packages/miniflare/src/workers/d1/dumpSql.ts
geelen: 👍

packages/miniflare/src/workers/d1/dumpSql.ts
geelen: Does it? This is supposed to match up with https://github.com/sqlite/sqlite/blob/105c20648e1b05839fd0638686b95f2e3998abcb/src/shell.c.in#L8472, i.e. directly after `run_schema_dump_query` (which dumps schema & data), it checks `SHFLG_DumpDataOnly)` is not set before exporting all `('index','trigger','view')` rows of `sqlite_schema`

Have I got that wrong?

packages/wrangler/src/__tests__/d1/export.test.ts
geelen: I have all these inside the D1 codebase, but see https://github.com/cloudflare/workers-sdk/pull/6073#discussion_r1655712242 I think it makes sense for this file to live in a separate package with a whole suite of tests (as well as a test runner that runs native SQLite for comparison like the D1 suite does).

Imo that's a follow-up task. For the moment I'm just keeping the two implementations in sync manually.

geelen:
> For my own education, does this PR mean that `PRAGMA miniflare_d1_export(?,?,?);` is part of our public API now? I don't think we want it to be so. How do we prevent people from writing that in their own queries?

It's part of miniflare's implementation of D1, but not present anywhere else. It's not accessible from `wrangler d1 execute --local` since you can't pass parameters but technically if someone wrote it in a worker and passed in at least two params they'd get a sql dump in local dev (and an error in production).

We could refactor `dumpSql` in this PR to use the public D1 API rather than the internal DO one (and then `wrangler export` would just send lots of queries to the D1 rather than a single debug pragma), but that would mean it's much harder to reuse the specific `dumpSql` code that's present in D1's internal codebase. And so I've erred on the side of making it easier to keep them consistent and accept leaking an implementation detail.

Ideally, Miniflare would make it possible to do an end-run around the D1 shim and hit an object directly with an RPC call, but as far as I can find out that's not possible.

packages/miniflare/src/workers/d1/dumpSql.ts
andyjessop: Makes sense, thank you!

andyjessop (approved):

packages/miniflare/src/workers/d1/dumpSql.ts
justin-mp: No, you're correct.

packages/wrangler/src/__tests__/d1/export.test.ts
justin-mp: I agree it could be a follow-up task.

justin-mp (approved):

Issue #6074, opened 2024-06-18
chore: Run eslint on workers-playground


threepointone:
### Describe the solution

Just discovered that we don't lint the workers-playground package, and there are a bunch of failing rules. Making this issue to track the work on it, I'll get to it myself.

PR #6075, opened 2024-06-18
chore: remove eslint config from workers-playground

threepointone:
We don't actually lint this package, so removing it for now. Followup issue: https://github.com/cloudflare/workers-sdk/issues/6074

### DIFF
--- a/packages/eslint-config-worker/react.js
+++ b/packages/eslint-config-worker/react.js
@@ -8,4 +8,9 @@ module.exports = {
 			extends: ["plugin:react/recommended", "plugin:react-hooks/recommended"],
 		},
 	],
+	settings: {
+		react: {
+			version: "detect",
+		},
+	},
 };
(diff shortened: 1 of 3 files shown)


packages/wrangler/.eslintrc.js
```diff
@@ -8,11 +8,6 @@ module.exports = {
 		"templates",
 		"emitted-types",
 	],
-	settings: {
```
threepointone, line 11: also move this to the base react eslint config so we don't have to write this everywhere

packages/eslint-config-worker/react.js
```diff
@@ -5,7 +5,16 @@ module.exports = {
 	overrides: [
 		{
 			files: ["*.ts", "*.tsx"],
-			extends: ["plugin:react/recommended", "plugin:react-hooks/recommended"],
+			extends: [
+				"plugin:react/jsx-runtime",
```
threepointone, line 9: added an additional plugin to extend from here

threepointone: nevermind this, it had other issues, removed.

petebacondarwin (approved):

PR #6077, opened 2024-06-18
test: improve Windows reliability of e2e tests

petebacondarwin:

- `child_process.kill()` doesn't really work on Windows. Using `kill-tree` library instead
- `killAllWranglerDev()` helper could mess up parallel tests since it kills all wrangler/workerd processes regardless of whether they are part of the current test
- `shellac` library has issues on Windows when running background tasks. Using simple `execSync` and `spawn` functions instead
- tightened up the timeouts a bit to avoid long running false negatives
- reworked the helpers to use a class `WranglerE2ETestHelper` rather than the `e2eTest` extension.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: test refactor
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: test refactor

### DIFF
--- a/packages/wrangler/e2e/README.md
+++ b/packages/wrangler/e2e/README.md
@@ -118,7 +118,7 @@ it(`can modify worker during ${cmd}`, async () => {
 			{
 				"name": "worker",
 				"version": "0.0.0",
-			  	"private": true
+				"private": true
 			}
 		`,
 	});
--- a/packages/wrangler/e2e/README.md
+++ b/packages/wrangler/e2e/README.md
@@ -47,24 +47,24 @@ describe("uploading Worker versions", () => {
 	beforeAll(async () => {
 		await helper.seed({
 			"wrangler.toml": dedent`
-			name = "${workerName}"
-			main = "src/index.ts"
-			compatibility_date = "2023-01-01"
-		`,
+				name = "${workerName}"
+				main = "src/index.ts"
+				compatibility_date = "2023-01-01"
+			`,
 			"src/index.ts": dedent`
-			export default {
-			fetch(request) {
-				return new Response("Hello World!")
-			}
-			}
-		`,
+				export default {
+					fetch(request) {
+						return new Response("Hello World!")
+					}
+				}
+			`,
 			"package.json": dedent`
-			{
-			"name": "${workerName}",
-			"version": "0.0.0",
-			"private": true
-			}
-		`,
+				{
+					"name": "${workerName}",
+					"version": "0.0.0",
+					"private": true
+				}
+			`,
 		});
 	});
 	it("deploy worker", async () => {
--- a/packages/wrangler/e2e/README.md
+++ b/packages/wrangler/e2e/README.md
@@ -9,7 +9,7 @@ You can then run the e2e test suite with the following commands (run them in roo
 ```zsh
 pnpm i
 pnpm build
-CLOUDFLARE_ACCOUNT_ID=8d783f274e1f82dc46744c297b015a2f CLOUDFLARE_API_TOKEN=<cloudflare-testing-api-token> WRANGLER="node --no-warnings $PWD/.tmp/wrangler/bin/wrangler.js" WRANGLER_IMPORT="$PWD/.tmp/wrangler/wrangler-dist/cli.js" pnpm --filter wrangler run test:e2e --retry 0
+CLOUDFLARE_ACCOUNT_ID=8d783f274e1f82dc46744c297b015a2f CLOUDFLARE_API_TOKEN=<cloudflare-testing-api-token> WRANGLER="node --no-warnings $PWD/packages/wrangler/bin/wrangler.js" WRANGLER_IMPORT="$PWD/packages/wrangler/wrangler-dist/cli.js" pnpm --filter wrangler run test:e2e
 ```
 
 > Make sure to replace `<cloudflare-testing-api-token>` with the actual API token you generated.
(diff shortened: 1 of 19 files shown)


packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -26,20 +22,23 @@ const WASM_ADD_MODULE = Buffer.from(
 	"base64"
 );
 
-describe.each(RUNTIMES)("Core: $flags", ({ runtime, flags }) => {
+describe.skip.each(RUNTIMES)("Core: $flags", ({ runtime, flags }) => {
```
RamIdeas, line 25: Is this a mistake?

packages/wrangler/e2e/README.md
```diff
@@ -6,69 +6,140 @@ You can also run these tests locally, but you'll need access to the `8d783f274e1
...
 
-```sh
-rm -rf .tmp
-pnpm --filter wrangler deploy .tmp/wrangler
-rm .tmp/wrangler/templates/tsconfig.json
-CLOUDFLARE_ACCOUNT_ID=8d783f274e1f82dc46744c297b015a2f CLOUDFLARE_API_TOKEN=$CLOUDFLARE_TESTING_API_TOKEN WRANGLER="node --no-warnings $PWD/.tmp/wrangler/bin/wrangler.js" WRANGLER_IMPORT="$PWD/.tmp/wrangler/wrangler-dist/cli.js" pnpm --filter wrangler run test:e2e --retry 0
+```zsh
+pnpm i
+pnpm build
+CLOUDFLARE_ACCOUNT_ID=8d783f274e1f82dc46744c297b015a2f CLOUDFLARE_API_TOKEN=<cloudflare-testing-api-token> WRANGLER="node --no-warnings $PWD/.tmp/wrangler/bin/wrangler.js" WRANGLER_IMPORT="$PWD/.tmp/wrangler/wrangler-dist/cli.js" pnpm --filter wrangler run test:e2e --retry 0
```
penalosa, line 12: This no longer works, since you've remove the `pnpm deploy` step and `.tmp/wrangler` won't be generated

penalosa (commented):
I'll review this in detail next week, but it looks like a big change here is migrating from vitest fixtures to a new helper class that needs to be instantiated in each test. To me, that seems harder to use, and adds boilerplate to each test run. What was the reasoning for moving away from vitest fixtures here?

It also looks like there's a change from a single `run()` function to a `runWranglerLongLived()` and a `run()` function.

packages/wrangler/e2e/README.md
petebacondarwin: Yes you're right! I forgot to update that bit. I have been happily running the e2e tests without deploying a standalone package. I'll update the command.

petebacondarwin:
> I'll review this in detail next week, but it looks like a big change here is migrating from vitest fixtures to a new helper class that needs to be instantiated in each test. To me, that seems harder to use, and adds boilerplate to each test run. What was the reasoning for moving away from vitest fixtures here?
> 
> It also looks like there's a change from a single `run()` function to a `runWranglerLongLived()` and a `run()` function.

The boilerplate is actually quite small. The main difference is that you instantiate a class to get access to the methods rather than changing the `it` calls to `e2eTest` calls.

The primary motivation was that it was not possible to use `e2eTest` to create a context that persisted across `it` calls, which is an approach used in a bunch of the tests.

But also, for me, the `e2eTest` approach used a bunch of Vitest magic that was not apparent on first looking at the code and required me to spend time digging around and reading up on the Vitest docs to understand what was going on and how it works. Simple class instantiation can be understood by anyone with a basic understanding of JS.

The move to split the `run` command in two is somewhat subjective, but I feel that it is more discoverable since they have quite different flows. It was only when I was updating the docs that I realised that you could just await the original `run` function to get a promise that effectively had very different usage than not awaiting it. While quite clever this is hard to discover and not intuitive. Finally the implementation of the new `run` command is much less complex and avoids having to worry about killing the async child process since it just uses standard Node.js methods to run the process to completion.

packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
```diff
@@ -1,123 +1,83 @@
 import assert from "node:assert";
 import crypto from "node:crypto";
-import { test } from "vitest";
+import { onTestFinished } from "vitest";
```
penalosa, line 3: TIL! This is cool

penalosa (approved):

packages/wrangler/e2e/vitest.config.mts
```diff
@@ -3,17 +3,17 @@ import { defineConfig } from "vitest/config";
...
 				singleThread: true,
 			},
 		},
-		retry: 2,
 		include: ["e2e/**/*.test.ts"],
 		// eslint-disable-next-line turbo/no-undeclared-env-vars
 		outputFile: process.env.TEST_REPORT_PATH ?? ".e2e-test-report/index.html",
 		globalSetup: path.resolve(__dirname, "./validate-environment.ts"),
 		reporters: ["verbose", "html"],
+		bail: 1,
```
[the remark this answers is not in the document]
petebacondarwin: From the docs:

> Stop test execution when given number of tests have failed 

So this will stop when the first test fails. The default of zero means to never bail.

PR #6083, opened 2024-06-19
chore: update to turbo v2

threepointone:
Updates to turbo v2.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] I don't know
  - [x] Not required because: already covered
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: no functional change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: chore


packages/create-cloudflare/turbo.json
```diff
@@ -1,13 +1,28 @@
 {
 	"$schema": "http://turbo.build/schema.json",
 	"extends": ["//"],
-	"pipeline": {
+	"tasks": {
 		"build": {
-			"env": ["TEST_PM", "TEST_PM_VERSION", "npm_config_user_agent", "CI"],
+			"env": [
+				"TMPDIR",
```
threepointone, line 7: turbo needs you to whitelist any nv vars that you'd use. We have to add these new ones so that os.tmpdir() works correctly (which depend on them)

packages/miniflare/turbo.json
```diff
@@ -1,15 +1,16 @@
 {
 	"$schema": "http://turbo.build/schema.json",
 	"extends": ["//"],
-	"pipeline": {
+	"tasks": {
 		"build": {
 			"inputs": ["src/**", "scripts/**", "types/**", "*.mjs", "*.js", "*.json"],
 			"outputs": ["dist/**", "bootstrap.js", "worker-metafiles/**"],
-			"env": ["CI_OS"]
+			"env": ["CI_OS", "NODE_EXTRA_CA_CERTS"]
```
threepointone, line 8: without this, the miniflare test `Miniflare: HTTPS fetches using browser CA certificates` will fail

packages/wrangler/turbo.json
```diff
@@ -58,12 +61,97 @@
...
+				"PWD",
+				"LC_ALL",
+				"WRANGLER_SEND_METRICS",
+				"https_proxy",
+				"HTTPS_PROXY",
+				"http_proxy",
+				"HTTP_PROXY",
+				"CI_OS",
+				"SENTRY_DSN",
+				"SYSTEMROOT",
```
petebacondarwin, line 149: All these feel unnecessary. If they are only in the source files then they are already covered by the `build` task. Here we should only be declaring the env vars that are referenced in the input files (i.e. `"e2e/**"`).

packages/cli/turbo.json
```diff
@@ -1,6 +1,9 @@
 {
 	"$schema": "http://turbo.build/schema.json",
 	"extends": ["//"],
-
-	"pipeline": {}
+	"tasks": {
+		"build": {
+			"outputs": ["dist/**"]
+		}
```
petebacondarwin, line 7: Does overriding this here also override the `dependsOn` field at the top level?
Do we need to include that?
I see that this package depends upon the eslint and tsconfig packages, so we need those deps marked in turbo so that we don't fail to rebuild this package when those change.

petebacondarwin (changes requested):
A couple of concerns that need resolving (either by fixing or telling me I'm wrong).

packages/miniflare/turbo.json
```diff
@@ -1,15 +1,16 @@
...
 		},
 		"test:ci": {
 			"inputs": ["test/**"],
-			"dependsOn": ["build"]
+			"dependsOn": ["build"],
+			"env": ["CI_OS", "NODE_EXTRA_CA_CERTS"]
```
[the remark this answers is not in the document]
threepointone: the test runner reads it dynamically, and it has to be whitelisted here for it to be added to the environment

packages/wrangler/turbo.json
```diff
@@ -1,11 +1,11 @@
...
 	"$schema": "http://turbo.build/schema.json",
 	"extends": ["//"],
-	"pipeline": {
+	"tasks": {
 		"build": {
 			"inputs": [
-				"!**/__tests__/**",
-				"!**/.wrangler/**",
+				"!*/**/__tests__/**",
+				"!*/**/.wrangler/**",
```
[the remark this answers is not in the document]
threepointone: turbo v2 breaking change for this pattern. this change was made by the migration codemod `npx @turbo/codemod migrate`

packages/wrangler/turbo.json
threepointone: If they're read dynamically, then they have to be whitelisted. I considered being more precise, but that would make this more confusing for the future. Instead, I duplicate them across tasks because it's easier to make changes in the future too (where the instruction will be "hey just add it to both"

packages/cli/turbo.json
petebacondarwin: I guess it improves the upstream tasks that rely upon this task?

penalosa:
Apologies, completely missed this PR, and did this separetly in #7076

PR #6090, opened 2024-06-19
chore: run eslint and tsc in missing packages

threepointone:
Some of our packages didn't have a lint/typecheck rule setup, so I added it. Also fixed a couple of failures along the way. I _didn't_ enable linting for kv-asset-handler because that looks like a bit more work, which we can do later.

No changeset / tests for this, no functional changes.

### DIFF
--- a/packages/create-cloudflare/package.json
+++ b/packages/create-cloudflare/package.json
@@ -30,7 +30,6 @@
 		"dev:codemod": "node -r esbuild-register scripts/codemodDev.ts",
 		"check:lint": "eslint .",
 		"check:type": "tsc",
-		"lint": "eslint",
 		"prepublishOnly": "pnpm -w run build",
 		"test:e2e": "vitest run --config ./vitest-e2e.config.ts",
 		"test:e2e:npm": "pnpm run build && cross-env TEST_PM=npm vitest run --config ./vitest-e2e.config.ts",
--- a/packages/kv-asset-handler/.eslintrc.js
+++ b/packages/kv-asset-handler/.eslintrc.js
@@ -1,0 +1,4 @@
+module.exports = {
+	root: true,
+	extends: ["@cloudflare/eslint-config-worker"],
+};
--- a/packages/workers-playground/src/QuickEditor/PreviewTab/useRefreshableIframe.tsx
+++ b/packages/workers-playground/src/QuickEditor/PreviewTab/useRefreshableIframe.tsx
@@ -42,6 +42,7 @@ export function useRefreshableIframe(
 				second.removeEventListener("load", onLoadEvent);
 			};
 		}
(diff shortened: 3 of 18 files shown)


packages/create-cloudflare/package.json
```diff
@@ -30,7 +30,6 @@
 		"dev:codemod": "node -r esbuild-register scripts/codemodDev.ts",
 		"check:lint": "eslint .",
 		"check:type": "tsc",
-		"lint": "eslint",
```
threepointone, line 33: we already have check:lint above this which actually lints the package, this command doesn't do anything

packages/workers-playground/src/QuickEditor/PreviewTab/useRefreshableIframe.tsx
```diff
@@ -42,6 +42,7 @@ export function useRefreshableIframe(
 				second.removeEventListener("load", onLoadEvent);
 			};
 		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
threepointone, line 45: this one is a tricky one to fix, excluding it for now

packages/workers-playground/src/QuickEditor/PreviewTab/useRefreshableIframe.tsx
```diff
@@ -69,6 +70,7 @@ export function useRefreshableIframe(
 		if (src) {
 			setUrl(src);
 		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
threepointone, line 73: this one is a tricky one to fix, excluding it for now

packages/kv-asset-handler/.eslintrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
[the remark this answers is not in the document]
CarmenPopoviciu: I'll leave it up to you, but in case we are not planning on fixing those failures any time soon, we might want to consider flagging what you said ☝️ with a comment, just in case someone wakes up one day with the urge of using this config 😅 . Again, abs up to you!

packages/workers-playground/src/QuickEditor/PreviewTab/useRefreshableIframe.tsx
CarmenPopoviciu: same question here

CarmenPopoviciu (approved):
Thank you for putting this up Sunil <3

packages/kv-asset-handler/.eslintrc.js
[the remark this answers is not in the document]
threepointone: I plan on doing it myself soon, filed https://github.com/cloudflare/workers-sdk/issues/6101

PR #6097, opened 2024-06-19
Support --x-dev-env for Pages

RamIdeas:

Fixes [DEVX-1296](https://jira.cfdata.org/browse/DEVX-1296)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: experimental flag


packages/wrangler/src/pages/dev.ts
```diff
@@ -680,6 +694,8 @@ export const Handler = async (args: PagesDevArguments) => {
 			showInteractiveDevSession: args.showInteractiveDevSession,
 			testMode: false,
 			watch: true,
+			fileBasedRegistry: args.experimentalRegistry,
```
penalosa, line 697: This shouldn't be added here—bad rebase maybe?

packages/wrangler/src/pages/dev.ts
```diff
@@ -239,9 +239,23 @@ export function Options(yargs: CommonYargsArgv) {
...
+					"Use the experimental DevEnv instantiation (unified across wrangler dev and unstable_dev)",
+				default: false,
+			},
+			"experimental-registry": {
+				alias: ["x-registry"],
+				type: "boolean",
+				describe:
+					"Use the experimental file based dev registry for multi-worker development",
+				default: false,
+			},
```
penalosa, line 258: As below

packages/wrangler/src/pages/dev.ts
```diff
@@ -239,9 +239,23 @@ export function Options(yargs: CommonYargsArgv) {
 			},
 			"show-interactive-dev-session": {
 				describe:
-					"Show interactive dev session (defaults to true if the terminal supports interactivity)",
+					"Use the experimental DevEnv instantiation (unified across wrangler dev and unstable_dev)",
```
penalosa, line 242: Typo

penalosa (dismissed):

packages/wrangler/e2e/__snapshots__/pages-dev.test.ts.snap
```diff
@@ -0,0 +1,84 @@
...
+  - SERVICE_BINDING_3_TOML: SERVICE_NAME_3_ARGS
+- AI:
+  - Name: AI_BINDING_2_TOML
+- Vars:
+  - VAR1: "(hidden)"
+  - VAR2: "VAR_2_TOML"
+  - VAR3: "(hidden)"
+▲ [WARNING] ⎔ Support for service bindings in local mode is experimental and may change.
+▲ [WARNING] ⎔ Support for external Durable Objects in local mode is experimental and may change.
+(node:75279) ExperimentalWarning: The Ed25519 Web Crypto API algorithm is an experimental feature and might change at any time
```
RamIdeas, line 40: The number in `(node:75279) ...` changes on every run which will cause the snapshots to fail

This line isn't in the other snapshot. Need to figure out where it's coming from

RamIdeas: I just needed to add `--no-warnings` to the wrangler command used in my e2e command

packages/wrangler/src/dev/start-server.ts
```diff
@@ -135,6 +135,9 @@ export async function startDevServer(
 			format: props.entry.format,
 			moduleRoot: props.entry.moduleRoot,
 		},
+		legacy: {
+			enablePagesAssetsServiceBinding: props.enablePagesAssetsServiceBinding,
+		},
```
petebacondarwin, line 140: I think this is OK to put in legacy for now. But we might need to revisit this when we get Assets for Workers up and running.

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -92,6 +92,23 @@ async function convertToConfigBundle(
 		event.bundle = { ...event.bundle, modules: [] };
 	}
 
+	if (event.config.legacy?.enablePagesAssetsServiceBinding !== undefined) {
```
petebacondarwin, line 95: should this be `==true`?

```suggestion
	if (event.config.legacy?.enablePagesAssetsServiceBinding === true) {
```

petebacondarwin (approved):
LGTM with a few nits.

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
petebacondarwin: since setting that flag to false would pass this if test

RamIdeas: This wasn't a boolean flag. It was an object which if set opted into using the binding with the object as config.

I realised this binding, when resolved, ended up being just a fetcher which startWorker support as-is. So I've updated this PR to pass it in directly so startWorker has no knowledge of the PagesAssetsServiceBinding and no extra options to support

packages/wrangler/src/dev/start-server.ts
RamIdeas: I've updated this PR to pass in a fetcher binding and therefore do not need to add this extra option to the API and support it forever

petebacondarwin (approved):
Still looks good. There are some flakes, perhaps?

PR #6105, opened 2024-06-20
Add `demandCommand()` to R2 Commands

### LINKED ISSUE #5928: 🐛 BUG: wrangler r2 bucket command logs nothing

Wrangler core

3.57.1

20.13.1

Windows 11

### Observed behavior

- Run `npx wrangler@3.57.1 r2 bucket`
- Observe no output. It just exits immediately

### Expected behavior

I would expect to see some output detailing the commands/help available under `bucket`, like `create`, `delete`, `list`, etc. as per https://developers.cloudflare.com/workers/wrangler/commands/#r2-bucket

Like what you see with `npx wrangler pages deployment`, and other commands.

helloimalastair:

Fixes #5928.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Change affects noop command output only.
  - [ ] I don't know
  - [x] Not required because: Same as above
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Same as above
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Same as above

### DIFF
--- a/packages/wrangler/src/index.ts
+++ b/packages/wrangler/src/index.ts
@@ -532,7 +532,7 @@ export function createCLIParser(argv: string[]) {
 
 	// r2
 	wrangler.command("r2", "📦 Manage R2 buckets & objects", (r2Yargs) => {
-		return r2(r2Yargs.command(subHelp));
+		return r2(r2Yargs, subHelp);
 	});
 
 	// d1
--- a/packages/wrangler/src/r2/index.ts
+++ b/packages/wrangler/src/r2/index.ts
@@ -24,7 +24,7 @@ import {
 } from "./helpers";
 import * as Notification from "./notification";
 import * as Sippy from "./sippy";
-import type { CommonYargsArgv } from "../yargs-types";
+import type { CommonYargsArgv, SubHelp } from "../yargs-types";
 import type { R2PutOptions } from "@cloudflare/workers-types/experimental";
 
 const CHUNK_SIZE = 1024;
@@ -50,10 +50,12 @@ async function createFileReadableStream(filePath: string) {
 	});
 }
 
-export function r2(r2Yargs: CommonYargsArgv) {
+export function r2(r2Yargs: CommonYargsArgv, subHelp: SubHelp) {
 	return r2Yargs
+		.command(subHelp)
 		.command("object", "Manage R2 objects", (r2ObjectYargs) => {
 			return r2ObjectYargs
+				.demandCommand()
 				.command(
 					"get <objectPath>",
 					"Fetch an object from an R2 bucket",
@@ -427,6 +429,7 @@ export function r2(r2Yargs: CommonYargsArgv) {
 		})
 
 		.command("bucket", "Manage R2 buckets", (r2BucketYargs) => {
+			r2BucketYargs.demandCommand();
 			r2BucketYargs.command(
 				"create <name>",
 				"Create a new R2 bucket",
--- a/.changeset/chilled-ties-occur.md
+++ b/.changeset/chilled-ties-occur.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+feat: Add help messages to all invalid `r2` commands
--- a/packages/wrangler/src/__tests__/r2.test.ts
+++ b/packages/wrangler/src/__tests__/r2.test.ts
@@ -2,6 +2,7 @@ import * as fs from "node:fs";
 import { http, HttpResponse } from "msw";
 import { MAX_UPLOAD_SIZE } from "../r2/constants";
 import { actionsForEventCategories } from "../r2/helpers";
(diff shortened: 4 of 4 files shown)


threepointone (approved):
Premptively approving, but will wait for @CarmenPopoviciu's call on including this in the imminent release.

CarmenPopoviciu (approved):

CarmenPopoviciu:
thank you @helloimalastair for fixing this <3. Can we please add a test to verify the changes 🙏 . There's a reason why we didn't caught this ourselves

CarmenPopoviciu (changes requested):
thank you @helloimalastair for fixing this <3. Can we please add a test to verify the changes 🙏 ? There's a reason why we didn't caught this ourselves 😋

helloimalastair:
@CarmenPopoviciu just to clarify, should the test run across just the commands that have changed here, or every command with subcommands within `wrangler`? And should it just check if *something* is logged, or that an exact message is logged?

CarmenPopoviciu:
basically we want to test that running `wrangler r2 bucket` prints
Screenshot 2024-06-20 at 15 47 06

and `wrangler r2 object` prints
Screenshot 2024-06-20 at 15 47 55

which is what this PR fixes

helloimalastair:
@CarmenPopoviciu should be ready with tests now? Had to replace `SubHelp` with `demandCommand()` to get them all to pass though.

packages/wrangler/src/index.ts
```diff
@@ -531,8 +531,8 @@ export function createCLIParser(argv: string[]) {
 	});
 
 	// r2
-	wrangler.command("r2", "📦 Manage R2 buckets & objects", (r2Yargs) => {
-		return r2(r2Yargs.command(subHelp));
+	wrangler.command("r2", "📦 Interact with an R2 store", (r2Yargs) => {
```
CarmenPopoviciu, line 534:
```suggestion
	wrangler.command("r2", "📦 Manage R2 buckets & objects", (r2Yargs) => {
```

CarmenPopoviciu: wording was changed recently based on a request from the design team, so let's please keep it in place. Similarly, wording across multiple other commands was changed in the same way to bring consistency across the cli

packages/wrangler/src/r2/index.ts
```diff
@@ -50,10 +50,12 @@ async function createFileReadableStream(filePath: string) {
 	});
 }
 
-export function r2(r2Yargs: CommonYargsArgv) {
+export function r2(r2Yargs: CommonYargsArgv, subHelp: SubHelp) {
 	return r2Yargs
+		.command(subHelp)
 		.command("object", "Manage R2 objects", (r2ObjectYargs) => {
 			return r2ObjectYargs
+				.demandCommand()
```
CarmenPopoviciu, line 58: just to make sure this is what you want here...

calling `demandCommand()` will throw an error if `wrangler r2 object` is called without any options. So basically this 👇 
Screenshot 2024-06-21 at 17 59 19

An alternative would be to instead call `.command(subHelp)` which would ensure we still show the help menu, without actually throwing. So basically this 👇 
Screenshot 2024-06-21 at 18 00 46

As far as I can see this type of behaviour is not standardized across the wrangler codebase, with commands such as `ai` and `d1` using `.demandCommand`, and all the other not. With the lack of a clear standard UX, I'm not sure what to advise here. I guess either of the two options are OK, but perhaps worth discussing with the r2 team/product what you'd like this behaviour to be?

packages/wrangler/src/__tests__/r2/r2.test.ts
```diff
@@ -0,0 +1,119 @@
...
+	runInTempDir();
+	it("should show help when no argument is passed", async () => {
+		await runWrangler("r2");
+		await endEventLoop();
+		expect(std.out).toMatchInlineSnapshot(`
+		"wrangler r2
+
+		📦 Interact with an R2 store
+
+		Commands:
```
CarmenPopoviciu, line 17: please make sure you rebase your branch on top of latest `main`. We've had some changes in there that would cause these tests to fail

packages/wrangler/src/index.ts
helloimalastair: Ooh, good catch. Not entirely sure why I changed it back, but will change it back again...

packages/wrangler/src/r2/index.ts
helloimalastair: Just confirmed with the team, and it's fine as is.

helloimalastair:
Think I screwed something up here, let me see if I can fix it...

helloimalastair:
Ok, should be good to go now, and all tests should be in a single file.

PR #6111, opened 2024-06-20
chore: fail build on eslint warnings

threepointone:
Modify all the eslint commands to fail if any warnings are present.

--- 
This should fail, and pass once https://github.com/cloudflare/workers-sdk/pull/6001/files#r1648173873 is resolved

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -17,7 +17,7 @@
 	"license": "MIT OR Apache-2.0",
 	"author": "wrangler@cloudflare.com",
 	"scripts": {
-		"check:lint": "eslint .",
+		"check:lint": "eslint . --max-warnings=0",
 		"check:type": "tsc",
 		"test:ci": "vitest run"
 	},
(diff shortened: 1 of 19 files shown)


DaniFoldi:
Hey Sunil :wave:,

I'm not sure if changing the scripts here is the best approach for contributors.

In IDEs warnings and errors have separate colours for highlighting, and most people are probably used to red being errors, and yellow being warnings. Obviously, I'd avoid introducing new warnings whenever possible, however if the goal is for a rule (or all rules) to prevent a change from being merged, wouldn't the best way to achieve that be setting the severity of that rule (or all rules) to `error`? That way local (IDE) and CI experience is consistent. Flat eslint configs (while a definite increase in complexity over the old system) are just js arrays/objects, so mapping warnings to errors should be relatively simple even with presets.

threepointone:
Warnings signal that, with good judgement, they can be disabled. While an error signals that it's usually a mistake and may cause problems down the line. This doesn't change anything about the dx, but we don't want to land code that hasn't had a call taken. For example the linked code shouldn't have landed and should have broken the build, even though it wasn't super serious. I wouldn't want that standing out as an error while developing, feels too aggressive.

CarmenPopoviciu (approved):

PR #6116, opened 2024-06-21
fix: onBundleStart dependency array

RamIdeas:

Fixes extraneous item in onBundleStart dependency array. Addresses comment [here](https://github.com/cloudflare/workers-sdk/pull/6001/files#r1648173873).

This prop was left in the dependency array after removing a redundant surrounding if-condition using the prop. It was decided the comment was enough but I am putting the redundant if-condition back to be ultra specific for readers of the code – the interim period where we're faking some events, but not when the experimental flag is on, is admittedly a bit confusing.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: tested by existing tests
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: no change in behaviour
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because:

### DIFF
--- a/packages/wrangler/src/dev/dev.tsx
+++ b/packages/wrangler/src/dev/dev.tsx
@@ -603,11 +603,12 @@ function DevSession(props: DevSessionProps) {
 	]);
 
 	const onBundleStart = useCallback(() => {
-		// NOTE: this callback won't be called if props.experimentalDevEnv
-		devEnv.proxy.onBundleStart({
-			type: "bundleStart",
-			config: startDevWorkerOptions,
-		});
+		if (!props.experimentalDevEnv) {
+			devEnv.proxy.onBundleStart({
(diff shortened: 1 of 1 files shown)


CarmenPopoviciu (approved):

penalosa:
This seems likely to cause more confusion to me—can we at least retain the comment?

petebacondarwin:
> This seems likely to cause more confusion to me—can we at least retain the comment?

There is a comment down where `onBundleStart()` is called.
I guess there is no chance that the `experimentalDevEnv` can change at runtime? If it can then it seems that this is a reasonable check to have in place?

PR #6124, opened 2024-06-21
startWorker tests

RamIdeas:

This PR rewrites the fixtures/dev-env tests as e2e tests. Whereas the former were written when only the ProxyController had been implemented and therefore faked all other events, these new tests assume a full implementation.

The steps in each test are identical but have been rewritten for e2e usage. Notable difference: the script contents is written to disk instead of passing it via config because `config.entrypoint.contents` has been descoped for now.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:


packages/wrangler/src/__tests__/api/startDevWorker/LocalRuntimeController.test.ts
```diff
@@ -125,7 +126,7 @@ describe("Core", () => {
 
 		const config: StartDevWorkerOptions = {
 			name: "worker",
-			entrypoint: unusable(),
+			entrypoint: unusable() as FilePath,
```
petebacondarwin, line 129:
```suggestion
			entrypoint: unusable<FilePath>(),
```

And so on.

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -14,30 +13,43 @@ type Options = StartDevWorkerOptions;
...
+				define: {},
+				format: "modules",
+				moduleRoot: directory,
+				...input.build,
+			},
+			...input,
+		};
 		this.latest = input;
-		this.emitConfigUpdateEvent(this.latest);
+		this.emitConfigUpdateEvent(this.config);
```
petebacondarwin, line 52: This looks like a bug fix? If so what is it fixing? Is there a relevant test that would fail if this got reverted?
What is the difference between `this.latest` and `this.config`?

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -14,30 +13,43 @@ type Options = StartDevWorkerOptions;
 export class ConfigController extends Controller<ConfigControllerEventMap> {
 	config?: Options;
 
-	public set(input: Hook<Options, [Readonly<Options> | undefined]>) {
+	public set(input: Options) {
 		const config = unwrapHook(input, this.latest);
```
petebacondarwin, line 17: I don't think that `unwrapHook()` needed here. The type suggests that `input` cannot be a function or a Promise.

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -51,7 +60,7 @@ export interface StartDevWorkerOptions {
 	 * This is the `main` property of a wrangler.toml.
 	 * You can specify a file path or provide the contents directly.
 	 */
-	entrypoint: FilePath;
+	entrypoint?: FilePath;
```
petebacondarwin, line 63: What does it mean for a worker to have no entry-point?

packages/wrangler/src/api/startDevWorker/ConfigController.ts
RamIdeas: You're right. I forgot to remove that. Not a problem since that function will return input as-is. But addressed by removing in https://github.com/cloudflare/workers-sdk/pull/6145

packages/wrangler/src/api/startDevWorker/ConfigController.ts
RamIdeas: It is a bug fix of the startWorker programmatic interface only (not wrangler_dev or unstable_dev). There are tests in this PR that use the `worker.config` (which is a getter for this ConfigController.config) value and will fail if it is not set

packages/wrangler/src/api/startDevWorker/types.ts
RamIdeas: As discussed offline, it means you will have to have a `main` field in your config file. There is existing runtime code to check one of the possible ways to provide an entrypoint is actually provided and throws if not. The existing code was a bit too hairy to create a static type union but will be possible in the future as more deadcode paths are removed.

penalosa (approved):

PR #6126, opened 2024-06-22
Add cloudchamber curl command

IRCody:
Adds a cloudchamber curl command. Example usage:

`wrangler cloudchamber curl /ssh-public-keys`

[CC-3877](https://jira.cfdata.org/browse/CC-3877)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: I don't believe any of the cloudchamber commands are documented there atm.

### DIFF
--- a/.changeset/new-dragons-bow.md
+++ b/.changeset/new-dragons-bow.md
@@ -4,5 +4,4 @@
 
 feature: Add 'cloudchamber curl' command
 
-Adds a cloudchamber curl command which allows easy access to arbitrary
-cloudchamber API endpoints.
+Adds a cloudchamber curl command which allows easy access to arbitrary cloudchamber API endpoints.
--- a/.changeset/new-dragons-bow.md
+++ b/.changeset/new-dragons-bow.md
@@ -1,0 +1,8 @@
+---
+"wrangler": minor
+---
+
+feature: Add 'cloudchamber curl' command
+
+Adds a cloudchamber curl command which allows easy access to arbitrary
+cloudchamber API endpoints.
--- a/packages/wrangler/src/__tests__/cloudchamber/curl.test.ts
+++ b/packages/wrangler/src/__tests__/cloudchamber/curl.test.ts
@@ -156,104 +156,104 @@ describe("cloudchamber curl", () => {
 			"├ Loading account 
 			│ 
 			>> Body
-				 [
-				     {
-				         \\"id\\": \\"1\\",
-				         \\"type\\": \\"default\\",
-				         \\"created_at\\": \\"123\\",
-				         \\"account_id\\": \\"123\\",
-				         \\"vcpu\\": 4,
-				         \\"memory\\": \\"400MB\\",
-				         \\"version\\": 1,
-				         \\"image\\": \\"hello\\",
-				         \\"location\\": {
-				             \\"name\\": \\"sfo06\\",
-				             \\"enabled\\": true
-				         },
-				         \\"network\\": {
-				             \\"ipv4\\": \\"1.1.1.1\\"
-				         },
-				         \\"placements_ref\\": \\"http://ref\\",
-				         \\"node_group\\": \\"metal\\"
-				     },
-				     {
-				         \\"id\\": \\"2\\",
-				         \\"type\\": \\"default\\",
-				         \\"created_at\\": \\"1234\\",
-				         \\"account_id\\": \\"123\\",
-				         \\"vcpu\\": 4,
-				         \\"memory\\": \\"400MB\\",
-				         \\"version\\": 2,
-				         \\"image\\": \\"hello\\",
-				         \\"location\\": {
-				             \\"name\\": \\"sfo06\\",
-				             \\"enabled\\": true
-				         },
-				         \\"network\\": {
-				             \\"ipv4\\": \\"1.1.1.2\\"
-				         },
-				         \\"current_placement\\": {
-				             \\"deployment_version\\": 2,
-				             \\"status\\": {
-				                 \\"health\\": \\"running\\"
-				             },
-				             \\"deployment_id\\": \\"2\\",
-				             \\"terminate\\": false,
-				             \\"created_at\\": \\"123\\",
-				             \\"id\\": \\"1\\"
-				         },
-				         \\"placements_ref\\": \\"http://ref\\",
-				         \\"node_group\\": \\"metal\\"
-				     },
-				     {
-				         \\"id\\": \\"3\\",
-				         \\"type\\": \\"default\\",
-				         \\"created_at\\": \\"123\\",
-				         \\"account_id\\": \\"123\\",
-				         \\"vcpu\\": 4,
-				         \\"memory\\": \\"400MB\\",
-				         \\"version\\": 1,
-				         \\"image\\": \\"hello\\",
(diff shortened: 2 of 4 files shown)


ichernetsky-cf (approved):

gabivlj (approved):

threepointone:
as you can tell, I've been poking at this PR. Looks like there's an actual failure on windows. I don't have a windows machine, so I'm debugging with CI. I'll keep poking at it.

threepointone:
I tried debugging, but it's a bit confusing, so I'm going to move away from this, cc @petebacondarwin

.changeset/new-dragons-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feature: Add 'cloudchamber curl command
```
petebacondarwin, line 5: Please can we add a bit more content here. If a user is browsing the changelog and see this they would not know what this command does and/or is for.

packages/wrangler/src/__tests__/cloudchamber/curl.test.ts
```diff
@@ -0,0 +1,126 @@
...
+			vcpu: 3,
+			memory: "400GB",
+			network: { assign_ipv4: "predefined" },
+		};
+
+		await runWrangler(
+			`cloudchamber curl /deployments/v2 --json -X POST -D '` +
+				JSON.stringify(deployment) +
+				`'`
+		);
```
petebacondarwin, line 85: NIT given you are using backticks then you may as well interpolate.

```suggestion
		await runWrangler(
			`cloudchamber curl /deployments/v2 --json -X POST -D '${JSON.stringify(deployment)}'`
		);
```

packages/wrangler/src/__tests__/cloudchamber/curl.test.ts
```diff
@@ -0,0 +1,126 @@
...
+			  -v, --verbose             Show version number  [boolean]
+			      --use-stdin, --stdin  Equivalent of using --data-binary @- in curl  [boolean]"
+		`);
+	});
+
+	it("should be able to use data flag", async () => {
+		setIsTTY(false);
+		setWranglerConfig({});
+		msw.use(
+			http.post("*/deployments/v2", async ({ request }) => {
```
petebacondarwin, line 59: None of these tests check what the format of this `*` section of the URL is.
The command could be posting to a completely random URL.

packages/wrangler/src/cloudchamber/curl.ts
petebacondarwin, line 1: Pretty much all the functions in this file already exist in packages/wrangler/src/cloudchamber/client/core/request.ts.

Can you just reuse those (e.g. just the `request` and `sendRequest`) functions from there here.

packages/wrangler/src/cloudchamber/index.ts
```diff
@@ -54,5 +55,11 @@ export const cloudchamber = (
 		)
 		.command("registries", "Configure registries via Cloudchamber", (args) =>
 			registriesCommand(args).command(subHelp)
+		)
+		.command(
+			"curl <path>",
+			"perform curl in wrangler",
```
petebacondarwin, line 61: Can you add some more information about what this command does.
Even from looking at the code, it is not 100% clear to me.
I think it is just a way to send an arbitrary request to the CC API?

petebacondarwin (changes requested):

packages/wrangler/src/cloudchamber/curl.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+async function requestFromCmd(
+	args: {
+		path: string;
+		method: string;
+		header: (string | number)[] | undefined;
+		data?: string;
+		silent?: boolean;
+		verbose?: boolean;
```
petebacondarwin, line 70: Do these have meaning over and above the `WRANGLER_LOG` env var that can control what `logger` methods actually write to the console?
Is there a reason you are not using the `logger`?

petebacondarwin (approved):

petebacondarwin (approved):
LGTM - might be worth getting second set of eyes from @RamIdeas, since he is interested in how we do CLI input and output and JSON etc.

packages/wrangler/src/cloudchamber/curl.ts
```diff
@@ -0,0 +1,154 @@
...
+				for (const header in headers) {
+					logRaw("\t" + yellow(`${header}: ${headers[header]}`));
+				}
+			}
+			logRaw(cyanBright(">> Body"));
+			const text = JSON.stringify(res, null, 4);
+			logRaw(
+				text
+					.split("\n")
+					.map((line) => `${yellow(`\t`)} ${brandColor(line)}`)
```
RamIdeas, line 136: Is there a need for `\t` to be yellow?

packages/wrangler/src/cloudchamber/curl.ts
```diff
@@ -0,0 +1,154 @@
...
+					null,
+					4
+				)
+			);
+		} else if (!args.json && !args.silent) {
+			if (args.verbose) {
+				logRaw(cyanBright(">> Headers"));
+				for (const header in headers) {
+					logRaw("\t" + yellow(`${header}: ${headers[header]}`));
+				}
```
RamIdeas, line 129: There is a `formatLabelledValues` util in `utils/render-labelled-values` which will do this for you

packages/wrangler/src/cloudchamber/curl.ts
RamIdeas: I'm also not sure what this .split/.map/.join is doing, since JSON.stringify(..., 4) already indents the output

packages/wrangler/src/cloudchamber/curl.ts
```diff
@@ -0,0 +1,154 @@
...
+						: {
+								res,
+								headers: headers,
+								request_id: requestId,
+							},
+					null,
+					4
+				)
+			);
+		} else if (!args.json && !args.silent) {
```
RamIdeas, line 124: This condition seems unnecessary
```suggestion
		} else {
```

packages/wrangler/src/cloudchamber/curl.ts
RamIdeas: I think we'd need to be careful with the output here. Is it expected that this command's output will be piped to other processes? If so, would changes to the output be considered a breaking change?

Since we've provided a --json flag, we should guide scripting users to that terse output which will reflect the backcompat of the api being hit. Leaving us more free to iterate on the human-readable output without fear of breaking anyone's workflow.

I don't see a problem with the `--verbose` flag, since `WRANGLER_LOG` or `--log-level` don't have a concept of verbosity, only debug logging. My only issue is if it encourages users to script against stdout, preventing us from iterating on this command.

RamIdeas (changes requested):

RamIdeas (approved):

PR #6129, opened 2024-06-22
update typescript to 5.5.2

threepointone:
Updating this so we can play with isolateDeclarations later

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: no functional code changes
  - [ ] I don't know
  - [x] Not required because: no functional code changes
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: no functional code changes
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: no functional code changes

### DIFF
--- a/packages/vitest-pool-workers/src/worker/lib/mlly.ts
+++ b/packages/vitest-pool-workers/src/worker/lib/mlly.ts
@@ -1,4 +1,4 @@
-function isObject(value: unknown) {
+function isObject(value: unknown): value is Record<string, unknown> {
 	return value !== null && typeof value === "object";
 }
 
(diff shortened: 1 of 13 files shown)


packages/vitest-pool-workers/src/worker/lib/mlly.ts
```diff
@@ -16,7 +16,7 @@ export function interopDefault(sourceModule: any): any {
 	for (const key in sourceModule) {
 		if (key === "default") {
 			try {
-				if (!(key in newModule)) {
+				if (!(key in (newModule as Record<string, unknown>))) {
```
andyjessop, line 19: I think you can avoid the type assertions in this file by moving them to the `isObject` function above:

```
function isObject(value: unknown): value is Record<string, unknown> {
	return value !== null && typeof value === "object";
}
```

Then add the same check for the `default` object:


```
if (
		!isObject(sourceModule) ||
		!("default" in sourceModule) ||
		!isObject(sourceModule.default)
	) {
		return sourceModule;
	}
```

threepointone: awesome, thank you!

petebacondarwin (approved):

PR #6137, opened 2024-06-24
fix c3 astro e2es

dario-piotrowicz:

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/packages/create-cloudflare/e2e-tests/frameworks.test.ts
+++ b/packages/create-cloudflare/e2e-tests/frameworks.test.ts
@@ -696,7 +696,15 @@ const verifyBuildScript = async (
 	const TEST_PORT = Math.ceil(Math.random() * 1000) + 20000;
 
 	const devProc = spawnWithLogging(
-		[npx, "wrangler", "pages", "dev", outputDir, "--port", `${TEST_PORT}`],
+		[
+			npx,
+			"wrangler@3.60.3",
+			"pages",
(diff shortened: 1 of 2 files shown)


packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -696,15 +697,23 @@ const verifyBuildScript = async (
...
+		],
 		{
 			cwd: projectPath,
 		},
 		logStream,
 	);
 
 	// Wait a few seconds for dev server to spin up
-	await sleep(7000);
+	await sleep(15000);
```
andyjessop, line 716: Is it possible to ping the server instead of waiting for an arbitrary? It might take only a few seconds on some machines.

dario-piotrowicz: yes definitely 🙂

I did had it that way earlier in this PR too

anyways I am just messing around and trying to understand why the CI e2es are failing so this is not really code I necessarily want to commit 😅

Issue #5542, opened 2024-04-07
🐛 BUG: OAUTH bug, redirects the user to localhost:8976


ranjancse26:
### Which Cloudflare product(s) does this pertain to?

Wrangler core

### What version(s) of the tool(s) are you using?

Wrangler 3.48.0

### What version of Node are you using?

v20.11.1

### What operating system and version are you using?

Windows 11

### Observed behavior
Wrangler login or publish is not working. It's taking the user to https://localhost:8976/oauth/callback

### Expected behavior
Wrangler should login
Wrangler should be able to publish from local machine

### Steps to reproduce
Please provide the following:
- Select any worker repo
- Perform the wrangler login using npx wrangler login
- The consent form is show
- Once the user approves the content, it's taking the user to https://localhost:8976/oauth/callback

### Please provide a link to a minimal reproduction

https://github.com/cloudflare/python-worker-hello-world

### Please provide any relevant error logs

No error logs displayed on screen.

penalosa:
@ranjancse26 This is expected behaviour. What happens when you are redirected to `localhost:8976`? What do the Wrangler output logs say?

ranjancse26:
@penalosa I don't think redirecting to localost 8976 is the expected behavior. Now I not seeing a different error. The localhost redirect is not happening. 

User consent screen is not showing up. Instead, I am seeing the console error.

![image](https://github.com/cloudflare/workers-sdk/assets/2565797/74ef9ea5-b33f-4d61-8ccf-49f6ad4f22e4)

![image](https://github.com/cloudflare/workers-sdk/assets/2565797/8ee0464e-6103-4767-a649-6d04b56a458b)

Could you please spend some time in recreating this issue by yourself. Follow the steps which I have mentioned above. See for yourself if you are able to just log in using wrangler and then do the deployment using the "deploy" command. Check if you are able to successfully deploy the worker using wrangler.

petebacondarwin:
I have tried to reproduce this using the steps given above. I cloned the python repository you linked.

Within this repository I ran:

```sh
 npx wrangler@3.48.0 login
 ⛅️ wrangler 3.48.0 (update available 3.51.0)
-------------------------------------------------------
Attempting to login via OAuth...
[...]

Issue #3672, opened 2023-07-24
🐛 BUG: `wrangler login` behind a VPN


rigor789:
### Which Cloudflare product(s) does this pertain to?

Wrangler core

### What version(s) of the tool(s) are you using?

3.3.0 [wrangler]

### What version of Node are you using?

20.4.0

### What operating system are you using?

Mac

### Describe the Bug

```
$ wrangler login
 ⛅️ wrangler 3.3.0
------------------
Attempting to login via OAuth...
Opening a link in your default browser: https://dash.cloudflare.com/oauth2/auth.....
```

-> Browser opens, I press **Allow**
-> Browser redirects to `localhost:8976/...` but shows "Unable to connect"

There's an exception in the terminal:
```
/Users/rigor789/.nvm/versions/node/v20.4.0/lib/node_modules/wrangler/wrangler-dist/cli.js:30633
            throw a;
            ^

SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (/Users/rigor789/.nvm/versions/node/v20.4.0/lib/node_modules/wrangler/wrangler-dist/cli.js:6565:19)
    at successSteps (/Users/rigor789/.nvm/versions/node/v20.4.0/lib/node_modules/wrangler/wrangler-dist/cli.js:6536:27)
    at /Users/rigor789/.nvm/versions/node/v20.4.0/lib/node_modules/wrangler/wrangler-dist/cli.js:5099:83
    at node:internal/process/task_queues:140:7
    at AsyncResource.runInAsyncScope (node:async_hooks:206:9)
    at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v20.4.0
```

Turns out this is caused by certain network conditions such as using a VPN that seemingly triggers a bot verification, that's not accounted for in the `wrangler login` flow.

Inspecting the response that's being parsed shows a html page titled "Just a moment..." with some javascript presumably for verifying the user. Seems like this contains sensitive info, hence I'm not sharing it here.

**Disabling the VPN fixes the issue.**

### Additional notes

Sounds like others have ran into this issue recently based on a Discord conversation:
https://discord.com/channels/595317990191398933/799437470004412476/1133135405928173638

Cherry:
[...]

PR #6140, opened 2024-06-25
fix: add extra error logging to auth response errors

petebacondarwin:

May help diagnosing issues like https://github.com/cloudflare/workers-sdk/issues/3672, https://github.com/cloudflare/workers-sdk/issues/5542 and https://jira.cfdata.org/browse/CUSTESC-42645

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: just adding logging for errors
  - [ ] I don't know
  - [x] Not required because: just adding logging for errors that are not hit in e2e tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: just adding logging to errors

### DIFF
--- a/packages/wrangler/src/user/user.ts
+++ b/packages/wrangler/src/user/user.ts
@@ -246,6 +246,7 @@ import { generateAuthUrl } from "./generate-auth-url";
 import { generateRandomState } from "./generate-random-state";
 import type { ChooseAccountItem } from "./choose-account";
 import type { ParsedUrlQuery } from "node:querystring";
+import type { Response } from "undici";
 
 export type ApiCredentials =
 	| {
@@ -728,7 +729,7 @@ async function exchangeRefreshTokenForAccessToken(): Promise<AccessContext> {
 		}
 	} else {
 		try {
-			const json = (await response.json()) as TokenResponse;
+			const json = (await getJSONFromResponse(response)) as TokenResponse;
 			if ("error" in json) {
 				throw json.error;
 			}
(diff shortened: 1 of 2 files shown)


packages/wrangler/src/user/user.ts
```diff
@@ -1249,3 +1252,27 @@ async function fetchAuthToken(body: URLSearchParams) {
 		headers,
 	});
 }
+
+async function getJSONFromResponse(response: Response) {
+	const text = await response.text();
```
andyjessop, line 1257: Very minor, but if you passed in the text string to this function, it would reduce its responsibilities:

```
async function getJSONFromText<T>(text: string): T {
  try ...
}

...

const json = getJSONFromText<TokenResponse>(await response.text());
```

andyjessop (approved):
One very minor comment. Otherwise, looks good 👍

packages/wrangler/src/user/user.ts
petebacondarwin: RIght, but then we would have to extract the text on each of the three calls, which seems a bit repetitive; but more importantly, the logs grab the status from the response so that would need to be passed in too.

petebacondarwin: Also, we might also want to use the status code as part of the analysis of what went wrong with the response (e.g. I think that the Bot Management failure would have a 403 status).

petebacondarwin:
We're going to hold off landing this given our limited deployment window this week. But if you are experiencing issues as linked in the description above, you could try this version of Wrangler via:

```
npx https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/9657631663/npm-package-wrangler-6140 login
```

Or perhaps installing it into your project (and using it as normal) via:

```
npm install --save-dev https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/9657631663/npm-package-wrangler-6140
```

PR #6144, opened 2024-06-25
fix: some peer dependency warnings

threepointone:
This fixes some peer dependency warning we were seeing when doing installs/builds

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] I don't know
  - [x] Not required because: no functional changes
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: no api/behaviour changes

### DIFF
--- a/.changeset/tiny-queens-brake.md
+++ b/.changeset/tiny-queens-brake.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vitest-pool-workers": patch
+---
+
+fix: some peer dependency warnings
--- a/fixtures/vitest-pool-workers-examples/package.json
+++ b/fixtures/vitest-pool-workers-examples/package.json
@@ -11,6 +11,7 @@
 		"@cloudflare/vitest-pool-workers": "workspace:*",
 		"@cloudflare/workers-types": "^4.20240620.0",
 		"@types/node": "20.8.3",
+		"@vitest/ui": "1.5.0",
 		"jose": "^5.2.2",
 		"miniflare": "workspace:*",
 		"toucan-js": "^3.3.1",
--- a/package.json
+++ b/package.json
@@ -72,8 +72,7 @@
 			"@types/react-tabs>@types/react": "^18",
 			"@types/react-transition-group>@types/react": "^18",
 			"@cloudflare/elements>@types/react": "^18",
-			"capnpc-ts>typescript": "5.5.5",
-			"vitest-pool-workers>vite": "5.1.0"
+			"capnpc-ts>typescript": "5.5.2"
 		},
 		"patchedDependencies": {
 			"ink@3.2.0": "patches/ink@3.2.0.patch",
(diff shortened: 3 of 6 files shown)


threepointone:
@dario-piotrowicz are you familiar with the vitest-pool-workers codebase? I see the type errors here which seem legit, but I'm not sure what the fix should be. Is this something you can help with? No rush at all on this, or lemme know if someone else can help on it. If not, I'll be happy to dive in myself.

dario-piotrowicz:
> @dario-piotrowicz are you familiar with the vitest-pool-workers codebase? I see the type errors here which seem legit, but I'm not sure what the fix should be. Is this something you can help with? No rush at all on this, or lemme know if someone else can help on it. If not, I'll be happy to dive in myself.

@threepointone no sorry I am not really familiar with the vitest-pool-workers codebase, I am still totally happy to give it a look when I find the time if you want 🙂 (sometime next week I think)

also @petebacondarwin or @penalosa could be familiar with it

andyjessop:
Vitest Pool Workers relies on a specific version of Vitest, which I think could cause issues such as this. When you `pnpm install` on this branch, you get:

```
 WARN  Issues with peer dependencies found
.
└─┬ @vitest/ui 1.6.0
  └── ✕ unmet peer vitest@1.6.0: found 1.2.2
```

I believe the actual issue is to do with a conflict in the way Vitest extends the `vite` types. It adds this ambient declaration:

```ts
declare module 'vite' {
    interface UserConfig {
        /**
         * Options for Vitest
         */
        test?: VitestInlineConfig;
    }
}
```

And for some reason, it's not picking this up, and is instead using the `UserConfig` from `vite`, which doesn't contain this `test` property.

andyjessop:
@threepointone @petebacondarwin I think there is a solution to this, but I don't believe it's a long term solution.

[This line](https://github.com/cloudflare/workers-sdk/pull/6144/files#diff-614d2d01eb1ba352d6f48212291ef416dc8b226f827b1614d514ba5efe9fdcdeR36) is causing the Vite version to resolve to 5.3.1 in the root. And this version of vite is causing this type mismatch in my comment above.

The solution I have is to lock the vite version in `vitest-pool-workers` to 5.1.0:

```
"vite": "5.1.0",
```

(Note that in order to update this correctly, you will have to first checkout the `pnpm-lock.yaml` file from `main` and then `pnpm install` again, because there is some apparent hysteresis in the `pnpm-lock.yaml`, which means that the `vitest-pool-workers` package will still see the 5.3.1 version if you only `pnpm install`).

The other solution I would think is to upgrade Vitest in `vitest-pool-workers`, but I read that the version there is locked.

andyjessop:
Referencing this here, because it's a different fix to the same issue:

https://github.com/cloudflare/next-on-pages/pull/811#discussion_r1652657818

threepointone:
@andyjessop feel free to take a call and own this PR?

PR #4877, opened 2024-01-30
[wrangler] fix: stop rebuild attempts when sources are missing (#3886)

magnusdahlstrand:
Fixes #3886.

**What this PR solves / how to test:**

As I've been working on a site using sveltekit and its adapter-cloudflare alongside workers-sdk I've been seeing lots of errors in the console (as those in the issue linked above) referring to missing files. This is due to rebuilds of the worker file being triggered on all file changes, including "unlink" events, which with lots of build pipelines happen at the beginning of a build.

In effect this means that we watch files for change, notice that a file has been deleted, and then try to rebuild, although we can guarantee that at this point the build will file, since the watched file no longer exists, and instead we should just hold on a few seconds until the build finishes and the "add" or "change" event is emitted.

**Author has addressed the following:**

This PR updates the behaviour of the dev.ts' worker script building to ignore "unlink" events on the worker file as well as the route files, so rebuilds are only triggered on events which add or modify said files. This means that no more `ENOENT` errors for missing `_worker.js` or `_routes.json` when watching for changes.

- Tests
  - [ ] Included
  - [x] Not necessary because: Removes intermittent errors during development watch experience which doesn't seem to be tested for.
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [x] Issue(s)/PR(s): #3886
  - [ ] Not necessary because:

**Note for PR author:**

We want to celebrate and highlight awesome PR review! If you think this PR received a particularly high-caliber review, please assign it the label `highlight pr review` so future reviewers can take inspiration and learn from it.

Issue #3886, opened 2023-09-01
🐛 BUG: wrangler pages dev 


tengweiherr:
### Which Cloudflare product(s) does this pertain to?

Pages

### What version(s) of the tool(s) are you using?

3.5.0 [Wrangler]

### What version of Node are you using?

16.18.1

### What operating system are you using?

Mac

### Describe the Bug

I was following the [local development guide of Next.js on @cloudflare/next-on-pages](https://github.com/cloudflare/next-on-pages/tree/main/packages/next-on-pages#local-development). The following is the summary: 

1. Enable Next.js pages local development in watch mode
`npx @cloudflare/next-on-pages --watch`
2. Then in a separate terminal run:
`npx wrangler pages dev .vercel/output/static --compatibility-flag=nodejs_compat`

On every change in the directory, it does trigger the build. During every build, the files in `.vercel/output` will be removed and replaced by updated asset. The building process takes quite some time, resulting in the following error from the wrangler:
```
✘ [ERROR] Could not validate _routes.json at /Users/wteng/Desktop/web-app/.vercel/output/static: Error: ENOENT: no such file or directory, open '/Users/wteng/Desktop/web-app/.vercel/output/static/_routes.json'

▲ [WARNING] Falling back to the following _routes.json default: {

    "version": 1,
    "include": [
      "/*"
    ],
    "exclude": []
  }

/usr/local/lib/node_modules/wrangler/wrangler-dist/cli.js:30632
            throw a;
            ^

[Error: ENOENT: no such file or directory, scandir '/Users/wteng/Desktop/web-app/.vercel/output/static/_worker.js'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'scandir',
  path: '/Users/wteng/Desktop/web-app/.vercel/output/static/_worker.js'
}
```

I also tried `npx wrangler pages dev --live-reload .vercel/output/static --compatibility-flag=nodejs_compat` but still the result.

## Expected result
It should work like a proper hot reloading application without error.

### Please provide any relevant error logs

✘ [ERROR] Could not validate _routes.json at /Users/wteng/Desktop/web-app/.vercel/output/static: Error: ENOENT: no such file or directory, open '/Users/wteng/Desktop/web-app/.vercel/output/static/_routes.json'

▲ [WARNING] Falling back to the following _routes.json default: {

    "version": 1,
    "include": [
      "/*"
    ],
    "exclude": []
  }

[...]

PR #6150, opened 2024-06-26
fix(wrangler): Fix `pages dev` watch mode [_worker.js]

### LINKED ISSUE #4824: 🚀 Feature Request: Pages functions with _worker.js does not hot reload for any changes other than the _worker.js file
### Describe the solution

It would be nice for wrangler to reload on any changes in the directory and subdirectories. Currently it will only reload if _worker.js is changed. It won't reload if any of its dependencies are modified including files imported from the same directory.

Using:

```
npx wrangler pages dev ./public
```

CarmenPopoviciu:

The watch mode in `pages dev` for Advanced Mode projects is currently partially broken, as it only watches for changes in the `_worker.js` file, but not for changes in any of its imported dependencies. This means that given the following `_worker.js` file

```
import { graham } from "./graham-the-dog";
export default {
    fetch(request, env) {
	return new Response(graham)
    }
}
```

`pages dev` will reload for any changes in the `_worker.js` file itself, but not for any changes in `graham-the-dog.js`, which is its dependency.

Similarly, `pages dev` will not reload for any changes in non-JS module imports, such as wasm/html/binary module imports.

This commit addresses all the aforementioned issues.

Fixes #4824

## Solutions we considered
Our original solution was implemented using esbuild's `watch` mode (see https://github.com/cloudflare/workers-sdk/pull/6150/commits/dd6cd19707f2070d7dba63e8d79872fce9887d82), which basically meant just enabling the `watch` flag in `buildRawWorker()`. Turned out though (via failed test in CI), that this solution introduced a regression for a frameworks corner case...the deletion of `_worker.js` itself (please see https://github.com/cloudflare/workers-sdk/pull/4877 and https://github.com/cloudflare/workers-sdk/issues/3886). 

When used in the context of frameworks, the assets directory served by `pages dev` gets rebuilt (removed + added back with updated assets) by a separate process, external to wrangler (I'm no expert here on the exact details, but I assume this more or less to be the case). This affects key Pages files, such as `_worker.js` and `_routes.json`, which get deleted/added back as part of that assets directory, which in turn, triggers a `_worker.js` rebuild. If the `_worker.js` re-build is triggered before the external process gets a chance to add the file back, `pages dev` will show a bunch of errors in the terminal, which, while accurate, seem to be confusing/annoying for our users. This was fixed by https://github.com/cloudflare/workers-sdk/pull/4877 by completely turning off Worker rebuilds for chokidar "unlink` events.

Unfortunately, the behaviour for this particular use case is very hard to port to esbuild watch mode, since esbuild gives us very limited information wrt to what changed ....so here we are 😢 

This lead us to the second implementation (see https://github.com/cloudflare/workers-sdk/pull/6150/commits/5c8fcf02018607b0308e595dd63f69ec33274756), which, [similar to Functions](https://github.com/cloudflare/workers-sdk/pull/6022), delegates the entirety of the watch mechanism to `chokidar`. 

## How we tested the changes
This PR comes with e2e tests, but in addition to them, we've manually tested these changes in the following scenarios:

- change/remove/add `_worker.js`
- change in external dependency (eg import { ADD } from "../<dir>/**/<dir>/add")
- change in external module (eg import html from "../<dir>/**/<dir>/my-html.html")
- adding/removing external dependencies
- errors in `_worker.js`
- errors in dependencies

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: watch mode is not documented

### DIFF
--- a/packages/wrangler/src/pages/dev.ts
+++ b/packages/wrangler/src/pages/dev.ts
@@ -17,12 +17,7 @@ import { printWranglerBanner } from "../update-check";
 import * as shellquote from "../utils/shell-quote";
 import { buildFunctions } from "./buildFunctions";
 import { ROUTES_SPEC_VERSION, SECONDS_TO_WAIT_FOR_PROXY } from "./constants";
-import {
-	FunctionsBuildError,
-	FunctionsNoRoutesError,
-	getFunctionsBuildWarning,
-	getFunctionsNoRoutesWarning,
-} from "./errors";
+import { FunctionsNoRoutesError, getFunctionsNoRoutesWarning } from "./errors";
 import {
 	buildRawWorker,
 	checkRawWorker,
@@ -504,27 +499,22 @@ export const Handler = async (args: PagesDevArguments) => {
 		 * every detection period, the signal will not be sent!"
 		 * (http://unscriptable.com/2009/03/20/debouncing-javascript-methods/)
 		 */
-		const debouncedRunBuild = (delayMS = 50) => {
-			const debounceFn = debounce(async () => {
-				try {
-					await runBuild();
-				} catch (e) {
-					/*
-					 * don't break developer flow in watch mode by throwing an error
-					 * here. Many times errors will be just the result of unfinished
-					 * typing. Instead, log the error, point out we are still serving
-					 * the last successfully built Worker, and allow developers to
-					 * write their code to completion
-					 */
-					logger.error(
-						`Error while attempting to build ${singleWorkerScriptPath}. Skipping to last successfully built version of the Worker.\n` +
-							`${e}`
-					);
-				}
-			}, delayMS);
-
-			debounceFn();
-		};
+		const debouncedRunBuild = debounce(async () => {
+			try {
+				await runBuild();
+			} catch (e) {
+				/*
+				 * don't break developer flow in watch mode by throwing an error
+				 * here. Many times errors will be just the result of unfinished
+				 * typing. Instead, log the error, point out we are still serving
+				 * the last successfully built Worker, and allow developers to
+				 * write their code to completion
+				 */
+				logger.warn(
+					`Failed to build ${singleWorkerScriptPath}. Continuing to serve the last successfully built version of the Worker.`
+				);
+			}
+		}, 50);
 
 		try {
 			await runBuild();
@@ -532,13 +522,16 @@ export const Handler = async (args: PagesDevArguments) => {
 			watcher.on("all", async (eventName, path) => {
 				logger.debug(`🌀 "${eventName}" event detected at ${path}.`);
 
-				// if "_worker.js" was deleted wait for 5s before re-building. It
-				// might be we are dealing with a temporary delete (as part of some
-				// frameworks-related process - see ...), and the file will be
-				// re-added immediately
-				const delayMS =
-					eventName === "unlink" && path.includes(workerScriptPath) ? 5000 : 50;
-				debouncedRunBuild(delayMS);
+				// Skip re-building the Worker if "_worker.js" was deleted.
+				// This is necessary for Pages projects + Frameworks, where
+				// the file was potentially deleted as part of a build process,
+				// which will add the updated file back.
+				// see https://github.com/cloudflare/workers-sdk/issues/3886
+				if (eventName === "unlink") {
+					return;
+				}
+
+				debouncedRunBuild();
 			});
 		} catch (e: unknown) {
 			/*
@@ -546,9 +539,7 @@ export const Handler = async (args: PagesDevArguments) => {
 			 * Worker. These flag underlying issues in the _worker.js code, and
 			 * should be addressed before starting the dev process
 			 */
-			throw new FatalError(
-				`Error while attempting to build ${singleWorkerScriptPath}\n` + `${e}`
-			);
+			throw new FatalError(`Failed to build ${singleWorkerScriptPath}.`);
 		}
 	} else if (usingFunctions) {
 		// Try to use Functions
@@ -688,11 +679,9 @@ export const Handler = async (args: PagesDevArguments) => {
 				await buildFn();
 			} catch (e) {
 				if (e instanceof FunctionsNoRoutesError) {
-					logger.error(
-						getFunctionsNoRoutesWarning(functionsDirectory, "skipping")
+					logger.warn(
+						`${getFunctionsNoRoutesWarning(functionsDirectory)}. Continuing to serve the last successfully built version of Functions.`
 					);
-				} else if (e instanceof FunctionsBuildError) {
-					logger.error(getFunctionsBuildWarning(functionsDirectory, e.message));
 				} else {
 					/*
 					 * don't break developer flow in watch mode by throwing an error
@@ -701,9 +690,8 @@ export const Handler = async (args: PagesDevArguments) => {
 					 * the last successfully built Functions, and allow developers to
 					 * write their code to completion
 					 */
-					logger.error(
-						`Error while attempting to build the Functions directory ${functionsDirectory}. Skipping to last successfully built version of Functions.\n` +
-							`${e}`
+					logger.warn(
+						`Failed to build Functions at ${functionsDirectory}. Continuing to serve the last successfully built version of Functions.`
 					);
 				}
 			}
@@ -734,8 +722,7 @@ export const Handler = async (args: PagesDevArguments) => {
 				 * issues in the Functions code, and should be addressed before
 				 */
 				throw new FatalError(
-					`Error while attempting to build the Functions directory ${functionsDirectory}\n` +
-						`${e}`
+					`Failed to build Functions at ${functionsDirectory}.`
 				);
 			}
 		}
@@ -747,7 +734,7 @@ export const Handler = async (args: PagesDevArguments) => {
 		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 		scriptReadyResolve!();
 
-		logger.log("No functions. Shimming...");
+		logger.log("No Functions. Shimming...");
 		scriptPath = resolve(getBasePath(), "templates/pages-shim.ts");
 	}
 
--- a/packages/wrangler/src/pages/dev.ts
+++ b/packages/wrangler/src/pages/dev.ts
@@ -1,6 +1,6 @@
 import { execSync, spawn } from "node:child_process";
 import { existsSync, lstatSync, readFileSync } from "node:fs";
-import { join, normalize, resolve } from "node:path";
+import { dirname, join, normalize, resolve } from "node:path";
 import { watch } from "chokidar";
 import * as esbuild from "esbuild";
 import { unstable_dev } from "../api";
@@ -402,7 +402,20 @@ export const Handler = async (args: PagesDevArguments) => {
 			}
 		});
 	} else if (usingWorkerScript) {
+		/*
+		 * Delegate watching for file changes to chokidar entirely. This gives
+		 * us a bit more flexibility and control as opposed to esbuild watch
+		 * mode, and keeps things consistent with the Functions implementation
+		 */
+		// always watch _worker.js
+		const watcher = watch([workerScriptPath], {
+			persistent: true,
+			ignoreInitial: true,
+		});
+		let watchedBundleDependencies: string[] = [];
+
 		scriptPath = workerScriptPath;
+
 		let runBuild = async () => {
 			await checkRawWorker(workerScriptPath, nodejsCompatMode, () =>
 				scriptReadyResolve()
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/pages/functions/buildWorker.ts
```diff
@@ -367,6 +372,39 @@ export function buildNotifierPlugin(onEnd: () => void): Plugin {
 	};
 }
 
+/**
+ *
```
andyjessop, line 376: Is this missing a comment?

packages/wrangler/src/pages/dev.ts
```diff
@@ -428,26 +429,23 @@ export const Handler = async (args: PagesDevArguments) => {
...
 						defineNavigatorUserAgent,
 					});
 				} catch (e: unknown) {
-					logger.warn("Failed to bundle _worker.js.", e);
+					/*
+					 * do not start the `pages dev` session if we encounter errors
+					 * while attempting to build the Worker. These flag underlying
+					 * issues in the _worker.js code, and should be addressed before
+					 */
+					throw new FatalError("Failed to bundle _worker.js\n" + `${e}`, 1);
```
andyjessop, line 443: 👍 Fail fast makes a lot of sense here.

packages/wrangler/src/pages/functions/buildWorker.ts
```diff
@@ -367,6 +372,39 @@ export function buildNotifierPlugin(onEnd: () => void): Plugin {
...
+ */
+export function additionalModulesWriterPlugin(
+	modules: CfModule[],
+	entry: Entry,
+	onEnd: () => void
+): Plugin {
+	return {
+		name: "wrangler-additional-modules-writer",
```
andyjessop, line 384: Nit: could this be `wrangler-dependent-modules-writer`? The word `additional` might be misleading.

andyjessop (approved):
Looks great! Just some minor comments, but nothing blocking.

packages/wrangler/src/pages/functions/buildWorker.ts
CarmenPopoviciu: oh yeah...I was meaning to explain why we needed this plugin to begin with. Thank you for catching that!

packages/wrangler/src/pages/functions/buildWorker.ts
CarmenPopoviciu: oh, much better than the name I went for! I like it! I like it a lot!

packages/wrangler/src/pages/dev.ts
```diff
@@ -402,7 +402,19 @@ export const Handler = async (args: PagesDevArguments) => {
 			}
 		});
 	} else if (usingWorkerScript) {
+		/*
```
CarmenPopoviciu, line 405: TODO @CarmenPopoviciu add comment

packages/wrangler/src/pages/dev.ts
```diff
@@ -417,37 +429,126 @@ export const Handler = async (args: PagesDevArguments) => {
...
+		};
+
+		try {
 			await runBuild();
-		});
+
+			watcher.on("all", async (eventName, path) => {
+				logger.debug(`🌀 "${eventName}" event detected at ${path}.`);
+
+				// if "_worker.js" was deleted wait for 5s before re-building. It
```
CarmenPopoviciu, line 534: are we OK with this approach? and if yes, what's a realistic value for `delayMS`. I have abs no feel for it

packages/wrangler/src/pages/dev.ts
```diff
@@ -402,7 +402,19 @@ export const Handler = async (args: PagesDevArguments) => {
...
+		/*
+		 *
+		 */
+		// always watch _worker.js
+		const watcher = watch([workerScriptPath], {
+			persistent: true,
+			ignoreInitial: true,
+		});
+		let currentBundleDependencies: string[] = [];
+		let watchedBundleDependencies: string[] = [];
```
penalosa, line 414: Is there a reason to have these as arrays given the functions dir implementation uses `Set`s?

packages/wrangler/src/pages/dev.ts
penalosa, line 1: Can we generalise any of this logic across functions and _worker.js?

packages/wrangler/src/pages/dev.ts
```diff
@@ -417,37 +429,126 @@ export const Handler = async (args: PagesDevArguments) => {
...
+		 * "Debouncing ensures that exactly one signal is sent for an
+		 * event that may be happening several times — or even several
+		 * hundreds of times over an extended period. As long as the
+		 * events are occurring fast enough to happen at least once in
+		 * every detection period, the signal will not be sent!"
+		 * (http://unscriptable.com/2009/03/20/debouncing-javascript-methods/)
+		 *
+		 * This is great for the typing experience for example, when ideally
+		 * we want to re-build the Worker script once, when the user is done
+		 * typing, as opposed to once per each typed character.
```
penalosa, line 504: Why would this otherwise build once per typed character? Surely this would only trigger on file save?

packages/wrangler/src/pages/dev.ts
penalosa: This feels a bit off to me, and I think framework users would probably be surprised that that their rebuilt changes take 5 seconds to show up. What about instead watching the parent directory of `_worker.js`, so we can immediately react to the creation event for `_worker.js`?

packages/wrangler/src/pages/dev.ts
```diff
@@ -417,37 +429,126 @@ export const Handler = async (args: PagesDevArguments) => {
...
+				// re-added immediately
+				const delayMS =
+					eventName === "unlink" && path.includes(workerScriptPath) ? 5000 : 50;
+				debouncedRunBuild(delayMS);
+			});
+		} catch (e: unknown) {
+			/*
+			 * fail early if we encounter errors while attempting to build the
+			 * Worker. These flag underlying issues in the _worker.js code, and
+			 * should be addressed before
```
penalosa, line 546: Is this sentence complete?

packages/wrangler/src/pages/dev.ts
```diff
@@ -417,37 +429,126 @@ export const Handler = async (args: PagesDevArguments) => {
...
+			 */
+			throw new FatalError(
+				`Error while attempting to build ${singleWorkerScriptPath}\n` + `${e}`
```
penalosa, line 549: Could we lean on esbuild's error formatting here?

packages/wrangler/src/pages/dev.ts
andyjessop: Why rebuild at all, if they have deleted? Seems like this behaviour would be unexpected.

CarmenPopoviciu: then perhaps the better question here is do we want to rebuild if `_worker.js` has been deleted? For me it was unexpected that nothing happened when I deleted the file. But maybe that's just me?

CarmenPopoviciu: > users would probably be surprised that that their rebuilt changes take 5 seconds to show up
to clarify this will only happen when the `_worker.js` file is deleted, **not** on every change

> What about instead watching the parent directory of _worker.js
I might be missing what you have in mind, but how would this solve the problem at hand? The issue is not reacting to the adding of the file, which this implementation already does, it's the adding it "immediately" (not sure what that actually looks like in terms of `ms`) after it was deleted, and either debouncing the `unlink` event handler, or bypassing it completely.

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: this was me trying to simplify what I had in the `Functions` implementation.  That one is a tad more verbose because of the `if...else` clauses, as opposed to this one, which does a lot of array ops chaining, which is why sets were no longer necessary. 

I hesitated writing the code this, because IMO the `Functions` one is a tad more readable...but then again this whole branching of `isUsingWorkerjs` | `isUsingFunctions` | `isUsingWorkerjsDirectory` is getting harder and harder to read :/

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: there was a failed attempt, but reverted because ultimately it felt like the extra layer of indirection adds more complexity than it solves. But let me look at this again and see what I can do

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: indeed, that was not very accurate of me to say. I'll rephrase

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: oeps 😅

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: ✅

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: ✅

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: ✅

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: as I was looking into this, I realised that the `buildNotifierPlugin` already does some logging for us if there were any build errors/warnings. Re-throwing the error here only duplicates that message in a less aesthetic format :/ 👇 
Screenshot 2024-07-03 at 13 44 24

So I'm wondering if it makes sense to output the error again. Would there be any use case where the build err would **not** go through the `buildNotifier` plugin? I'd imagine not. In which case, should we limit this error to a custom message, smth like "failed to build _worker.js", and defer to the notifier plugin to do the error printing?

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: refactored Functions to be in line with this PR here: https://github.com/cloudflare/workers-sdk/pull/6150/commits/536836fcd5791a58c3f9b7297b36eb391cdd528c

packages/wrangler/src/pages/dev.ts
penalosa: Maybe I'd misunderstand, but I thought the problem this was solving was that the `_worker.js` is deleted by some frameworks on every reload?

CarmenPopoviciu: that is correct yes. and because said frameworks also add the file back, we'd ideally want to either:
- delay building the worker until after the file is added back (which is what this PR does)
- or by-pass rebuilding the worker completely whenever worker file gets deleted (which is how it currently works)

your proposal to react immediately to the `addFile` event (by watching the parent folder), is something that is already accomplished in the crr PR by watching the file itself. The problem in both cases would remain that we'd get all sorts of compile errors when attempting to build the worker after the file was deleted, unless we do one of the 👆

penalosa: What I'm concerned about is:
- Framework deletes `_worker.js` file
- We delay building for 5 seconds
- Framework re-adds `_worker.js` file as part of the build process
- User is waiting for 5 seconds for the build to trigger

Would that not happen?

packages/wrangler/src/pages/dev.ts
penalosa: Makes sense, let's keep things as-is!

penalosa (approved):

penalosa:
Approved pending https://github.com/cloudflare/workers-sdk/pull/6150#discussion_r1664328291, which I'm likely misunderstanding

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: yes that can happen if the framework adds the file back before those 5s have elapsed. So say file is deleted, then added after 1s...the user needs to wait for 4 more seconds before we trigger a rebuild. But this is why I was saying that I have zero feels abt how long this delay should be. We can set it to whatever we want.

CarmenPopoviciu: At the same time, if the framework re-adds the file after 7s, we trigger a rebuild after 5s, which will correctly error because the file does not exist, then trigger another build after 2s when the file was re-added, which, assuming all's well, will compile successfully. So in this case, even with a debounce in place, we'd still run into rebuilding a worker file that does not exist, which is what users don't want.

CarmenPopoviciu: So perhaps, instead of trying to get that delay right, which I'm guessing is impossible, cause the frameworks build process duration is relative to the project size, maybe keep the current behaviour and nor rebuild whenever the file is deleted? I still find the lack of feedback odd, but it might possibly be just me

penalosa: > maybe keep the current behaviour and not rebuild whenever the file is deleted?
I think this might be best, even if it is a bit confusing, since frameworks seem to rely on this behaviour. Maybe we could log a small warning to the terminal?

packages/wrangler/src/pages/dev.ts
CarmenPopoviciu: ✅

Issue #4944, opened 2024-02-07
🐛 BUG: `account_id` is not passed to the `/zones` endpoint


ichernetsky-cf:
### Which Cloudflare product(s) does this pertain to?

Other

### What version(s) of the tool(s) are you using?

Latest Wrangler

### What operating system and version are you using?

Linux

### Observed behavior

Wrangler gets the error below when deploying a Worker.

```
Listing zones across more than 500 accounts is currently not allowed, please apply the 'account.id' filter
```

### Expected behavior

As the error message suggests, [gertZoneIdFromHost](https://github.com/cloudflare/workers-sdk/blob/2b71e22f86be35068eac3713b7e6eefbf04a0970/packages/wrangler/src/zones.ts#L89) should pass `account_id` to the `/zones` endpoint.

taylorlee:
The issue is [here](https://github.com/cloudflare/workers-sdk/blob/2b71e22f86be35068eac3713b7e6eefbf04a0970/packages/wrangler/src/zones.ts#L89). We need to add 'account.id' to URLSearchParams, which presumably means passing the account down through all the calls above getZoneIdFromHost.

PR #6164, opened 2024-06-28
fix: use account id for listing zones

threepointone:
Fixes https://github.com/cloudflare/workers-sdk/issues/4944

Trying to fetch `/zones` fails when it spans more than 500 zones. The fix to use an account id when doing so. This patch passes the account id to the zones call, threading it through all the functions that require it.

- Tests
  - [ ] TODO (before merge)
  - [x] Included - updated
  - [ ] Not necessary because: already covered 
  - [ ] I don't know
  - [ ] Not required because: already covered 
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/dev.tsx
+++ b/packages/wrangler/src/dev.tsx
@@ -537,6 +537,9 @@ export async function startDev(args: StartDevOptions) {
 			args.config ||
 			(args.script && findWranglerToml(path.dirname(args.script)));
 
+		const projectRoot = configPath && path.dirname(configPath);
+		let config = readConfig(configPath, args);
+
 		const devEnv = new DevEnv();
 
 		if (args.experimentalDevEnv) {
@@ -646,7 +649,8 @@ export async function startDev(args: StartDevOptions) {
 				dev: {
 					auth: async () => {
 						return {
-							accountId: args.accountId ?? (await getAccountId()),
+							accountId:
+								args.accountId ?? config.account_id ?? (await getAccountId()),
 							apiToken: requireApiToken(),
 						};
 					},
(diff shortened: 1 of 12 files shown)


packages/wrangler/src/dev.tsx
```diff
@@ -912,7 +916,10 @@ export async function validateDevServerSettings(
 	// we want it to exit the Wrangler process early to allow the user to fix it. Calling it here forces
 	// the error to be thrown where it will correctly exit the Wrangler process
 	if (args.remote) {
-		await getZoneIdForPreview(host, routes);
+		const accountId =
+			args.accountId ?? config.account_id ?? (await getAccountId());
```
threepointone, line 920: @RamIdeas does this look right to you?

RamIdeas: Looks about right. I just checked on the `--x-dev-env` code path and it passes accountId through correctly so this will continue to work when we flip that flag on by default

packages/wrangler/src/dev.tsx
```diff
@@ -537,6 +537,9 @@ export async function startDev(args: StartDevOptions) {
 			args.config ||
 			(args.script && findWranglerToml(path.dirname(args.script)));
 
+		const projectRoot = configPath && path.dirname(configPath);
```
threepointone, line 540: moved this up so I can reference `config` at line 653

RamIdeas (approved):

PR #6167, opened 2024-06-28
feature: alias modules in the worker

threepointone:
Sometimes, users want to replace modules with other modules. This commonly happens inside a third party dependency itself. As an example, a user might have imported `node-fetch`, which will probably never work in workerd. You can use the alias config to replace any of these imports with a module of your choice.

Let's say you make a `fetch-nolyfill.js`
```ts
export default fetch; // all this does is export the standard fetch function`
```

You can then configure `wrangler.toml` like so:
```toml
[alias]
"node-fetch": "./fetch-nolyfill"
```

So any calls to `import fetch from 'node-fetch';` will simply use our nolyfilled version.

You can also pass aliases in the cli (for both `dev` and `deploy`). Like:
```bash
npx wrangler dev --alias node-fetch:./fetch-nolyfill
```

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: vibes
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:


RamIdeas (approved):
🎉

irvinebroque:
Is the reason this only works for Workers but not Pages...

- just that Pages is a different codepath, and viable to make it work for Pages?
- that there's something fundamental about module aliasing that doesn't (at least easily) work a pre-bundled project (i.e. I run `remix vite:build` and then `wrangler pages deploy` with aliasing enabled in my `wrangler.toml`)
- something else

Similar to https://github.com/cloudflare/workers-sdk/pull/6511, trying to make sure we don't add to the set of difference between Workers/Pages where we can avoid it.

And with Workers Assets — will be people doing something akin to `remix vite:build && wrangler deploy`, like they do with `wrangler pages deploy` today. So if there are tricky cases when we try to do aliasing on code that's already gone through this build/bundling step, trying to think ahead to that.

PR #6187, opened 2024-07-02
Multiline hotkeys

RamIdeas:

This PR contains two implementations for multiline hotkey instructions which will display when the boxed, single-line hotkey instructions do not fit on the screen without wrapping.

First implementation (checkout commit e138fd3508f13f0805ffeb5048595bec9be3eba2 to test):

image

Second implementation (use PR prerelease to test):

image

Multiline unboxed 

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because:

### DIFF
--- a/packages/wrangler/src/cli-hotkeys.ts
+++ b/packages/wrangler/src/cli-hotkeys.ts
@@ -29,29 +29,21 @@ export default function (
 			.filter((option) => !unwrapHook(option.disabled))
 			.map(({ keys, label }) => `[${keys[0]}] ${unwrapHook(label)}`);
 
-		let stringifiedInstructions = instructions.join(", ");
+		const stringifiedInstructions = instructions.join(", ");
 
 		const ADDITIONAL_CHARS = 6; // 3 chars on each side of the instructions for the box and spacing ("│  " and "  │")
 		const willWrap =
 			stringifiedInstructions.length + ADDITIONAL_CHARS >
 			process.stdout.columns;
 		if (willWrap) {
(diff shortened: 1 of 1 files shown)


petebacondarwin (approved):
I think I prefer the boxed version to keep it distinct from any other logs that come in.

As an extra bonus follow-up, it would be cool to listen to width changes on the window, if that is possible.

RamIdeas:
> I think I prefer the boxed version to keep it distinct from any other logs that come in.
> 
> As an extra bonus follow-up, it would be cool to listen to width changes on the window, if that is possible.

@Meahaa agrees the boxed version is better (or a subheading for the unboxed version but we decided to go with boxed)

This PR prerelease build now contains the boxed hotkeys instructions

I will follow up with responsive rerendering: will need to manually count the overflow lines to console.clear (should be a case of `linesToClear = previousInstructions.split('\n').map(line => Math.ceil(line / stdout.columns).reduce(sum, 0))`)

petebacondarwin:
Ship it!

petebacondarwin:
Oh no! I Merged this without tests!
I'll create some now.

petebacondarwin:
Here is the test: https://github.com/cloudflare/workers-sdk/pull/6278

PR #6192, opened 2024-07-03
fix: do not report D1 user errors to Sentry

petebacondarwin:

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: D1 has no e2e tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: no user facing change

### DIFF
--- a/packages/wrangler/src/d1/execute.tsx
+++ b/packages/wrangler/src/d1/execute.tsx
@@ -13,7 +13,7 @@ import { fetchResult } from "../cfetch";
 import { readConfig } from "../config";
 import { getLocalPersistencePath } from "../dev/get-local-persistence-path";
 import { confirm } from "../dialogs";
-import { JsonFriendlyFatalError, UserError } from "../errors";
+import { createFatalError, JsonFriendlyFatalError, UserError } from "../errors";
 import { logger } from "../logger";
 import { APIError, readFileSync } from "../parse";
 import { readableRelative } from "../paths";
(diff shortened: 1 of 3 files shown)


penalosa (approved):

packages/wrangler/src/d1/execute.tsx
```diff
@@ -552,17 +555,26 @@ async function d1ApiPost<T>(
...
+				method: "POST",
+				headers: {
+					"Content-Type": "application/json",
+					...(db.internal_env ? { "x-d1-internal-env": db.internal_env } : {}),
+				},
+				body: JSON.stringify(body),
+			}
+		);
+	} catch (x) {
+		if (x instanceof APIError) {
```
rozenmd, line 571: fwiw the API endpoints now correctly throw a `400 Bad Request` when a user error occurs - might be worth specifically looking for that type of response

petebacondarwin: Unfortunately, the way the cfetch code works it drops the HTTP status code by the time it generates an `APIError`. All we get back is:

```
  "result": [],
  "success": false,
  "errors": [
    {
      "code": 7500,
      "message": "no such table: Customers"
    }
  ],
  "messages": []
}
```

When creating the APIError, we do grab the error code from the first error in the response. In the example above 7500.

We could consider fixing the cfetch to throw an APIError whenever we don't get a 2xx response but I worry that this might break other APIs.

PR #6194, opened 2024-07-03
chore: add upload duration and sourcemap size to metrics

zebp:

Adds `durationMs` and `sourceMapSize` properties to the `deploy worker script` metrics event as well as sending the event after the deploy request. This PR can be tested by deploying a Worker using wrangler built from this PR and checking internal tools.

Fixes WP-996

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Other metrics aren't tested in code and can instead be spot checked with internal tools
  - [ ] I don't know
  - [x] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: (as far as I can tell) Wrangler metrics aren't publicly documented

### DIFF
--- a/.changeset/fair-actors-appear.md
+++ b/.changeset/fair-actors-appear.md
@@ -1,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+chore: Add duration and sourcemap size to upload metrics event
+
+Wrangler will now send the duration and the total size of any sourcemaps uploaded with your Worker to Cloudflare if you have metrics enabled.
--- a/packages/wrangler/src/deploy/index.ts
+++ b/packages/wrangler/src/deploy/index.ts
@@ -231,15 +231,6 @@ export async function deployHandler(
 	const projectRoot = configPath && path.dirname(configPath);
 	const config = readConfig(configPath, args);
 	const entry = await getEntry(args, config, "deploy");
-	await metrics.sendMetricsEvent(
-		"deploy worker script",
-		{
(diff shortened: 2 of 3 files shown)


packages/wrangler/src/deploy/index.ts
```diff
@@ -217,6 +217,7 @@ export function deployOptions(yargs: CommonYargsArgv) {
 export async function deployHandler(
 	args: StrictYargsOptionsToInterface<typeof deployOptions>
 ) {
+	const beforeUpload = Date.now();
```
penalosa, line 220: This is probably not the right place to put this—the wrangler banner might fetch an updated Wrangler version from npm, which would influence this timing

zebp: I've moved this line to right before the call to `deploy`, I figured that `requireAuth` and `standardPricingWarning` shouldn't be considered for this metric.

penalosa (approved):

.changeset/fair-actors-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Add duration and sourcemap size to upload metrics event
```
penalosa, line 5: This should follow the changeset format: https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets

PR #6198, opened 2024-07-04
Less checks on PRs

penalosa:

_Description incoming_

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: this is a change to CI
  - [ ] I don't know
  - [x] Not required because: CI change
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: not public facing
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: not public facing

### DIFF
--- a/.github/workflows/test-and-check.yml
+++ b/.github/workflows/test-and-check.yml
@@ -14,7 +14,8 @@ jobs:
     steps:
       - run: curl -X POST https://devprod-status-bot.devprod.workers.dev/pr-project/workers-sdk/${{ github.event.number }}
 
-  check:
+  check-all:
+    if: github.ref == 'refs/heads/changeset-release/main'
     timeout-minutes: 30
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-checks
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-check.yml
```diff
@@ -14,7 +14,8 @@ jobs:
     steps:
       - run: curl -X POST https://devprod-status-bot.devprod.workers.dev/pr-project/workers-sdk/${{ github.event.number }}
 
-  check:
+  check-all:
```
petebacondarwin, line 17: Be aware that we can't make this a required check since it would never pass on any PRs.
So this would only be for information only.

It should not run on `ubuntu-latest` since the `Checks` is a required check to land the Version Packages PR.

.github/workflows/test-and-check.yml
```diff
@@ -53,6 +54,42 @@ jobs:
       - name: Check the changesets
         run: node -r esbuild-register tools/deployments/validate-changesets.ts
 
+  check:
+    if: github.ref != 'refs/heads/changeset-release/main'
```
petebacondarwin, line 58: This will need to run on all PRs since it is now a required check.

petebacondarwin (changes requested):

.github/workflows/test-and-check.yml
penalosa: Ah—I'd been hoping that the name based matching would make this work. I'll fix that

petebacondarwin (approved):

PR #6201, opened 2024-07-04
Granular e2e caching

penalosa:

Turbo only supports caching on the individual task level, but for Wrangler's e2e tests we want to support caching on a more granular basis—at the file level.

As such, this PR constructs multiple turbo tasks—one per e2e test file, ensuring that each file's tests can be cached individually.

Because Turbo ignores the cache when `package.json` changes, we can't generate these `package.json` scripts on the fly in CI, and so we require them to be commited to the repository. Since this is easy to mess up, this PR adds a script to lint `turbo.json`, the root `package.json`, and Wrangler's `package.json` to ensure everything is set up as intended.

The intended flow here is that CI will run `pnpm test:e2e:wrangler`, which will trigger turbo to run an individual task for each Wrangler e2e test file. These tasks are defined in the root `turbo.json`, and correspond to scripts in the Wrangler package. We ensure there's no accidental task name collision across packages by adding `--filter wrangler` to the turbo command, to make it only run tasks in the Wrangler package.

An example of the benefit this gives us are these two runs:
- https://github.com/cloudflare/workers-sdk/actions/runs/9797436298/job/27054030668?pr=6201
  This e2e test run failed in `dev-with-resources.test.ts`, but passed `dev-env.test.ts
- https://github.com/cloudflare/workers-sdk/actions/runs/9797436298/job/27054188910?pr=6201#step:6:197
  On a re-run, the previous `dev-env.test.ts` success was cached, speeding up the test run

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: modification to how tests are run
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: not user facing
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: not user facing


petebacondarwin (changes requested):
Unfortunately this breaks on Windows due to its inability to have `:` colons in filepaths.

Is it not possible to make this more dynamic by using an ENV var to set which e2e tests to run and then running turbo multiple times for each env var?

tools/e2e/runIndividualE2EFiles.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Turbo only supports caching on the individual task level, but for Wrangler's
+ * e2e tests we want to support caching on a more granular basis—at the file level.
+ *
+ * As such, we run the `test:e2e` turbo task multiple times—once per e2e test file
+ * with different arguments, ensuring that each file's tests can be cached individually.
```
petebacondarwin, line 6: 🎉

petebacondarwin (approved):

tools/e2e/runIndividualE2EFiles.ts
```diff
@@ -0,0 +1,32 @@
...
+// Get a list of e2e test files, each of which should have an associated script
+const e2eTests = readdirSync("packages/wrangler/e2e");
+
+const tasks = new Set<string>();
+
+for (const file of e2eTests) {
+	// Ignore other files in the e2e directory (the README, for instance)
+	if (file.endsWith(".test.ts")) {
+		tasks.add(
+			`pnpm test:e2e --log-order=stream --summarize --filter wrangler --concurrency 1 -- run ./e2e/${file}`
```
petebacondarwin, line 23: Perhaps we could run the `build` task once at at the start of the run, and then use the `--only` flag here to avoid dumping all the build output out each time (even though it is full turbo)?

PR #6216, opened 2024-07-07
No-c3-framework-deploy-tests

petebacondarwin:

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:


packages/create-cloudflare/e2e-tests/helpers.ts
```diff
@@ -41,7 +43,7 @@ const testEnv = {
 	// do not use the same global cache and accidentally hit race conditions.
 	YARN_CACHE_FOLDER: "./.yarn/cache",
 	YARN_ENABLE_GLOBAL_CACHE: "false",
-	PNPM_HOME: "./.pnpm",
+	// PNPM_HOME: "./.pnpm",
```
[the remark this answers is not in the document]
petebacondarwin: Ah that was from my previous hackings. I can't remember why. Let me revert

packages/create-cloudflare/e2e-tests/cli.test.ts
```diff
@@ -71,11 +48,12 @@ describe.skipIf(experimental || frameworkToTest || isQuarantineMode())(
 			expect(output).toEqual(version);
 		});
 
-		test.skipIf(process.platform === "win32")(
+		test({ experimental }).skipIf(process.platform === "win32")(
 			"Using arrow keys + enter",
-			async () => {
+			async ({ logStream, project }) => {
+				console.log(project);
```
RamIdeas, line 54: Noticed this pop up the the output https://github.com/cloudflare/workers-sdk/actions/runs/11072563792/job/30767063710?pr=6216#step:4:190

Was it intentional?

Reminds me: can we log in the job output what pm+framework failed so we can more easily distinguish flakes from repeated real failures without opening the logs file?

petebacondarwin: Again hacking/debugging leftovers.

But yeah I agree we should make the error cases more easy to diagnose.

packages/create-cloudflare/e2e-tests/helpers.ts
```diff
@@ -340,36 +347,38 @@ const getLogPath = (opts: { experimental: boolean }, suite: Suite) => {
...
 
 	const randomSuffix = crypto.randomBytes(4).toString("hex");
 	const baseProjectName = `${C3_E2E_PREFIX}${randomSuffix}`;
 
-	const getName = (suffix: string) => `${baseProjectName}-${suffix}`;
-	const getPath = (suffix: string) => path.join(tmpDirPath, getName(suffix));
-	const clean = (suffix: string) => {
+	const getName = () =>
+		// Worker project names cannot be longer than 58 characters
+		`${baseProjectName}-${test.substring(0, 57 - baseProjectName.length)}`;
```
RamIdeas, line 372: If 2 tests have the same prefix, this will cause them the clash. How likely is that to happen?

RamIdeas (commented):
How much of this is necessary for the goal of running against local dev vs deploying the worker?

What's the benefit of these changes vs just setting `deploy: false` and hitting the local url instead of the deployed url?

packages/create-cloudflare/e2e-tests/helpers.ts
petebacondarwin: The `baseProjectName` already has a random suffix, which makes these unique in general.

petebacondarwin:
> How much of this is necessary for the goal of running against local dev vs deploying the worker?
> 
> What's the benefit of these changes vs just setting `deploy: false` and hitting the local url instead of the deployed url?

I started out just trying to "not deploy" these tests but it became clear that we would need to change how we run the tests, since the current setup expects them to be deployed with remote URLs.  As I started fixing that up these other refactorings and changes came to light as sensible improvements.

Is there a concern over these other changes? What would you prefer to exclude from this PR?

packages/create-cloudflare/e2e-tests/helpers.ts
RamIdeas: Ah ok I assumed projectName was shared across the suite but I see now it's not

RamIdeas (approved):

PR #6229, opened 2024-07-08
Clarify what C3 is in bug template

irvinebroque:
More clearly spell out `npm create cloudflare` — the command people typed — rather than just referring to this as "C3"

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-template.yaml
+++ b/.github/ISSUE_TEMPLATE/bug-template.yaml
@@ -15,7 +15,7 @@ body:
       label: Which Cloudflare product(s) does this pertain to?
       multiple: true
       options:
-        - C3
+        - C3 (npm create cloudflare)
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -25,7 +25,7 @@ body:
         - Workers for Platforms
         - Constellation
```
penalosa, line 26: We should probably rename this to Workers AI

penalosa (approved):

.github/ISSUE_TEMPLATE/bug-template.yaml
irvinebroque:
```suggestion
        - Workers AI
```

penalosa (approved):

PR #6245, opened 2024-07-10
Add support for creating Hyperdrive over Access configs

OilyLime:

Fixes SQC-259

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/15559
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/hyperdrive/client.ts
+++ b/packages/wrangler/src/hyperdrive/client.ts
@@ -16,11 +16,11 @@ export type PublicOrigin = {
 	database?: string;
 	user?: string;
 	access_client_id?: string;
-	access_client_secret?: string;
 };
 
-export type OriginWithPassword = PublicOrigin & {
+export type OriginWithSecrets = PublicOrigin & {
 	password?: string;
+	access_client_secret?: string;
 };
 
 export type CachingOptions = {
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/hyperdrive/client.ts
```diff
@@ -9,15 +9,14 @@ export type HyperdriveConfig = {
...
 	host?: string;
 	port?: number;
-};
-
-export type PublicOrigin = Origin & {
 	scheme?: string;
 	database?: string;
 	user?: string;
+	access_client_id?: string;
+	access_client_secret?: string;
```
ReppCodes, line 19: This one section is a bit odd to me.  We exclude `password` here, presumably because it's not meant to be public?  Should we also give `access_client_secret` the same treatment?

OilyLime: You're right. Moved it to the other type alongside the password.

petebacondarwin (approved):

PR #6252, opened 2024-07-11
VS-197: Enable wrangler to interact with Vectorize V2 indexes

garvit-gupta:

Enables users to operate on Vectorize V2 operations via wrangler.

Fixes #[VS-197].

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [] TODO (before merge)
  - [x ] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/15916
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/thick-hairs-notice.md
+++ b/.changeset/thick-hairs-notice.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feat: Enable Wrangler to operate on Vectorize V2 indexes
--- a/packages/wrangler/src/__tests__/vectorize/vectorize.test.ts
+++ b/packages/wrangler/src/__tests__/vectorize/vectorize.test.ts
@@ -1,4 +1,5 @@
 import { http, HttpResponse } from "msw";
+import { validateQueryFilter } from "../../vectorize/query";
 import { endEventLoop } from "../helpers/end-event-loop";
 import { mockAccountId, mockApiToken } from "../helpers/mock-account-id";
 import { mockConsoleMethods } from "../helpers/mock-console";
@@ -7,6 +8,7 @@ import { useMockIsTTY } from "../helpers/mock-istty";
 import { createFetchResult, msw } from "../helpers/msw";
 import { runInTempDir } from "../helpers/run-in-tmp";
 import { runWrangler } from "../helpers/run-wrangler";
+import type { VectorizeQueryOptions } from "../../vectorize/types";
 
 describe("vectorize help", () => {
 	const std = mockConsoleMethods();
@@ -17,31 +19,40 @@ describe("vectorize help", () => {
 		await endEventLoop();
 
 		expect(std.out).toMatchInlineSnapshot(`
-			"wrangler vectorize
+				"wrangler vectorize
 
-			🧮 Manage Vectorize indexes [open beta]
+				🧮 Manage Vectorize indexes [open beta]
 
-			COMMANDS
-			  wrangler vectorize create <name>  Create a Vectorize index
-			  wrangler vectorize delete <name>  Delete a Vectorize index
-			  wrangler vectorize get <name>     Get a Vectorize index by name
-			  wrangler vectorize list           List your Vectorize indexes
-			  wrangler vectorize insert <name>  Insert vectors into a Vectorize index
+				COMMANDS
+				  wrangler vectorize create <name>                 Create a Vectorize index
+				  wrangler vectorize delete <name>                 Delete a Vectorize index
+				  wrangler vectorize get <name>                    Get a Vectorize index by name
+				  wrangler vectorize list                          List your Vectorize indexes
+				  wrangler vectorize query <name>                  Query a Vectorize index
+				  wrangler vectorize insert <name>                 Insert vectors into a Vectorize index
+				  wrangler vectorize upsert <name>                 Upsert vectors into a Vectorize index
+				  wrangler vectorize get-vectors <name>            Get vectors from a Vectorize index
+				  wrangler vectorize delete-vectors <name>         Delete vectors in a Vectorize index
+				  wrangler vectorize info <name>                   Get additional details about the index
+				  wrangler vectorize create-metadata-index <name>  Enable metadata filtering on the specified property
+				  wrangler vectorize list-metadata-index <name>    List metadata properties on which metadata filtering is enabled
+				  wrangler vectorize delete-metadata-index <name>  Delete metadata indexes
 
-			GLOBAL FLAGS
-			  -j, --experimental-json-config  Experimental: support wrangler.json  [boolean]
-			  -c, --config                    Path to .toml configuration file  [string]
-			  -e, --env                       Environment to use for operations and .env files  [string]
-			  -h, --help                      Show help  [boolean]
-			  -v, --version                   Show version number  [boolean]
+				GLOBAL FLAGS
+				  -j, --experimental-json-config  Experimental: support wrangler.json  [boolean]
+				  -c, --config                    Path to .toml configuration file  [string]
+				  -e, --env                       Environment to use for operations and .env files  [string]
+				  -h, --help                      Show help  [boolean]
+				  -v, --version                   Show version number  [boolean]
 
-			--------------------
-			📣 Vectorize is currently in open beta
-			📣 See the Vectorize docs for how to get started and known issues: https://developers.cloudflare.com/vectorize
-			📣 Please report any bugs to https://github.com/cloudflare/workers-sdk/issues/new/choose
-			📣 To give feedback, visit https://discord.cloudflare.com/
-			--------------------"
-		`);
+				--------------------
+				📣 Vectorize is currently in open beta
+				📣 Please use the '--deprecated-v1' flag to create, get, list, delete and insert vectors into legacy Vectorize indexes
+				📣 See the Vectorize docs for how to get started and known issues: https://developers.cloudflare.com/vectorize
+				📣 Please report any bugs to https://github.com/cloudflare/workers-sdk/issues/new/choose
+				📣 To give feedback, visit https://discord.cloudflare.com/
+				--------------------"
+			`);
 	});
 
 	it("should show help when an invalid argument is passed", async () => {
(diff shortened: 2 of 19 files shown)


ndisidore (approved):
Looks great!

packages/wrangler/src/__tests__/vectorize/vectorize.test.ts
```diff
@@ -110,7 +110,8 @@ describe("vectorize help", () => {
 			  -v, --version                   Show version number  [boolean]
 
 			OPTIONS
-			      --json  return output as clean JSON  [boolean] [default: false]
+			      --json   return output as clean JSON  [boolean] [default: false]
+			      --useV2  Use the v2 version of Vectorize index operations.  [boolean] [default: false]
```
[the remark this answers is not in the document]
garvit-gupta: Fixed!

packages/wrangler/src/__tests__/vectorize/vectorize.test.ts
```diff
@@ -186,6 +219,18 @@ describe("vectorize commands", () => {
 	`);
 	});
 
+	it("should handle a get on a vectorize V2 index", async () => {
+		mockVectorizeV2Request();
+		await runWrangler("vectorize get test-index --use-v2=true");
```
petebacondarwin, line 224: What does v2 mean when it comes to indexes?
What would happen if test-index was not v2?

petebacondarwin (commented):
I was a bit confused by the use of v2 at first.
I think what you are saying is that the v2 flag effectively provides a brand new product from v1, right?
For example it seems that you can create v1 and v2 indexes with the same name and then interact with them separately?
If not then what would it mean to create a v1 index and then use v2 commands to interact with it?

Can you clarify this in the command docs?

.changeset/thick-hairs-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
```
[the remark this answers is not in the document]
garvit-gupta: Vectorize V1 is an open-beta product, and Vectorize V2 is a re-architectured system that will soon go beta and eventually GA (at which point we will deprecate Vectorize v1).

Vectorize V2 is essentially a different product. V2 operations can only work on a v2 index (created using the use-v2 flag) and the same applies for V1. However, we cannot create a v2 index if a v1 index of the same name exists and vice versa.

packages/wrangler/src/__tests__/vectorize/vectorize.test.ts
garvit-gupta: If we try a v2 operation (say get index) on a v1 index, the wrangler client would receive a 403 response from Vectorize. Same goes if we try a v1 operation on a v2 index.

petebacondarwin: What would be the error message. Will it be clear to the user what went wrong and what to do to fix it?
Also what is the roll off plan for deprecating and removing v1?

petebacondarwin: For example, for how long will you support `use-v2` as a command line flag?
In general we are moving towards feature flagging things in Wrangler with the `experimental`/`x` prefix. So in this case it would be 

```
wrangler vectorize get test-index --experimental-v2
```
or
```
wrangler vectorize get test-index --x-v2
```

And for stuff you are removing prefixing with `legacy`/`deprecated`. For example:

```
wrangler vectorize get test-index --legacy-v1
```
or
```
wrangler vectorize get test-index --deprecated-v1
```

garvit-gupta: The error message in the apiv4 response format would be like `Index: "${index_name}" with version "${index_version}" called with api version "${operation_version}"` along with a 403 response.

I would let Yevgen address the deprecation plan.

netgusto: cc @sejoker

.changeset/thick-hairs-notice.md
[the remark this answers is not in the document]
garvit-gupta: Updated the changeset patch to minor, and made the v2 option description in the commands more descriptive. @petebacondarwin I hope that would be sufficient.

threepointone:
Is there a command that converts a v1 index to a v2 index?

sejoker:
> Is there a command that converts a v1 index to a v2 index?

This will be worked on later in the quarter, definitely something we will provide before GA.

packages/wrangler/src/__tests__/vectorize/vectorize.test.ts
sejoker: > For example, for how long will you support `use-v2` as a command line flag? In general we are moving towards feature flagging things in Wrangler with the `experimental`/`x` prefix. So in this case it would be
> 
> ```
> wrangler vectorize get test-index --experimental-v2
> ```
> 
> or
> 
> ```
> wrangler vectorize get test-index --x-v2
> ```
> 
> And for stuff you are removing prefixing with `legacy`/`deprecated`. For example:
> 
> ```
> wrangler vectorize get test-index --legacy-v1
> ```
> 
> or
> 
> ```
> wrangler vectorize get test-index --deprecated-v1
> ```

@petebacondarwin the plan to start with private beta this month, once we will be confident to open it up for public beta sometime in August we will stop allowing V1 index creation and V2 only will be used for new indexes.

sejoker: I like the idea of using `--deprecated-v1` for dealing with existing V1 indexes once V2 goes to public beta.

garvit-gupta: We discussed this within our team, and decided to with an `--experimental-v2` flag with a default value of false until we put v1 indexes on the deprecation path. At that time we will switch to `--deprecated-v1`. 

Made the change to update the commands and the related docs.

CarmenPopoviciu:
@garvit-gupta is my understanding correct that some backend work needs to happen before we can merge & release this PR? Are any further wrangler changes required or is the PR in its current state ready to go?

CarmenPopoviciu:
spoke async with the team, and this is now ready for re-review

.changeset/thick-hairs-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Enable Wrangler to operate on Vectorize V2 indexes
```
CarmenPopoviciu, line 5:
```suggestion
feat: Enable Wrangler to operate on Vectorize V2 indexes
```

packages/wrangler/src/vectorize/insert.ts
```diff
@@ -50,10 +57,17 @@ export async function handler(
...
 
-	if (Number(args.batchSize) > VECTORIZE_MAX_BATCH_SIZE) {
+	if (args.deprecatedV1 && Number(args.batchSize) > VECTORIZE_MAX_BATCH_SIZE) {
 		logger.error(
 			`🚨 Vectorize currently limits upload batches to ${VECTORIZE_MAX_BATCH_SIZE} records at a time.`
 		);
+	} else if (
+		!args.deprecatedV1 &&
+		Number(args.batchSize) > VECTORIZE_V2_MAX_BATCH_SIZE
+	) {
```
sejoker, line 67: let's put a warning for V2 with batch size > 1 and < 5000:
`The global rate limit for the Cloudflare API of 1200 requests per five minutes. Please use a maximum of 5000 vectors per embeddings.ndjson file to stay within the service limits`

.changeset/thick-hairs-notice.md
garvit-gupta: Updated!

packages/wrangler/src/vectorize/insert.ts
garvit-gupta: Updated!

packages/wrangler/src/__tests__/vectorize/vectorize.test.ts
petebacondarwin: > We discussed this within our team, and decided to with an `--experimental-v2` flag with a default value of false until we put v1 indexes on the deprecation path. At that time we will switch to `--deprecated-v1`.
> 
> Made the change to update the commands and the related docs.

This is not what I see in the current code for this PR.
I see that there is no --experimental-v2 flag here only the --deprecated-v1 flag, which defaults to false.

I would assume that for the private beta the default needs to switch to true and the --experimental-v2 flag needs to be implemented?

Also, could I call out that it would be much nicer for the user if these commands inferred the version of the index from its type (via a REST request) rather than requiring the user to keep track of which version they are using. The only place one really needs the v1/v2 flag is at index creation time, no?

petebacondarwin:
As I understand it, this PR should only land when we V2 goes into public beta.

packages/wrangler/src/vectorize/client.ts
```diff
@@ -2,36 +2,32 @@ import { fetchListResult, fetchResult } from "../cfetch";
...
-
 export async function createIndex(
 	config: Config,
-	body: object
-): Promise<VectorizeIndexResult> {
+	body: object,
+	deprecatedV1: boolean
+): Promise<VectorizeIndex> {
 	const accountId = await requireAuth(config);
+	const version_param = deprecatedV1 ? `` : `/v2`;
```
ndisidore, line 27: super nit (that's not worth a commit needing a rebuild): in js, camelCase is preferred

ndisidore (commented):
This is looking great! I know there's some work you want to do around the query parse function. Happy to approve once we look at that

packages/wrangler/src/vectorize/client.ts
garvit-gupta: Fixed it along with some other changes!

garvit-gupta:
> This is looking great! I know there's some work you want to do around the query parse function. Happy to approve once we look at that

The query filter parse and validate logic looks cleaner now. Ready for review!

packages/wrangler/src/vectorize/query.ts
```diff
@@ -16,18 +22,56 @@ export function options(yargs: CommonYargsArgv) {
...
+				type: "boolean",
+				default: false,
+				describe:
+					"Specify if the vector values should be included in the results.",
+			},
+			"return-metadata": {
+				type: "string",
+				default: "none",
+				describe:
+					"Specify if the vector metadata should be included in the results. Should be either ",
```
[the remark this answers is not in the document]
garvit-gupta: Great catch! Fixed it.

sejoker:
@petebacondarwin @CarmenPopoviciu could you please confirm this PR is ready to be merged from your perspective? Confirmed, we want to release this change once Vectorize V2 goes into public beta.

petebacondarwin (approved):
What a massive PR! From what I can see this looks great.
Happy for this to be merged when it is unblocked for product release.
cc @RamIdeas - you might be interested in this contribution.

packages/wrangler/src/vectorize/query.ts
```diff
@@ -16,18 +24,57 @@ export function options(yargs: CommonYargsArgv) {
 			description: "The name of the Vectorize index.",
 		})
 		.options({
-			query: {
-				type: "string",
+			vector: {
```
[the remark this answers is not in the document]
garvit-gupta: The `query` operation was never enabled for Vectorize V1 in Wrangler (if you check index.ts, you'll see that it was enabled as a part of this PR). So it is not a breaking change. `--vector` makes this param consistent with the ApiV4 route params and would be the preferable option.

packages/wrangler/src/vectorize/common.ts
```diff
@@ -1,7 +1,35 @@
...
+) {
+	let batch: string[] = [];
+	for await (const line of rl) {
+		if (batch.push(line) >= batchSize) {
+			yield batch;
+			batch = [];
+		}
+	}
+
+	yield batch;
```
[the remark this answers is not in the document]
garvit-gupta: That's a good idea. Included this change with the latest update.

ndisidore (approved):

sejoker (approved):

sejoker:
@RamIdeas @petebacondarwin we are happy to merge the change and release it next week.

PR #6255, opened 2024-07-12
[D1] teach init --from-dash about d1

### LINKED ISSUE #6253: 🐛 BUG: init --from-dash doesn't know about D1

Wrangler

latest

latest

Mac

### Describe the Bug

.

_No response_

_No response_

rozenmd:
Run `wrangler init --from-dash <my-d1-worker-here>`, check if the wrangler.toml contains unsafe bindings

Fixes #6253

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/init.ts
+++ b/packages/wrangler/src/init.ts
@@ -8,6 +8,7 @@ import { version as wranglerVersion } from "../package.json";
 import { fetchResult } from "./cfetch";
 import { fetchWorker } from "./cfetch/internal";
 import { readConfig } from "./config";
+import { getDatabaseInfoFromId } from "./d1/utils";
 import { confirm, select } from "./dialogs";
 import { getC3CommandFromEnv } from "./environment-variables/misc-variables";
 import { UserError } from "./errors";
@@ -27,6 +28,7 @@ import type {
 	TailConsumer,
 	ZoneNameRoute,
 } from "./config/environment";
+import type { DatabaseInfo } from "./d1/types";
 import type {
 	WorkerMetadata,
 	WorkerMetadataBinding,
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/init.ts
```diff
@@ -1035,6 +1052,18 @@ export function mapBindings(bindings: WorkerMetadataBinding[]): RawConfig {
...
 						}
 						break;
+					case "d1":
+						{
+							configObj.d1_databases = [
+								...(configObj.d1_databases ?? []),
+								{
+									binding: binding.name,
+									database_id: binding.id,
+									database_name: d1BindingsWithInfo[binding.id].name,
```
rozenmd, line 1062: You might be tempted to ask why we don't just teach EWC about `database_name` - the thing is, EWC doesn't store or care about the name of the DB, only D1's API knows it, so we need to fetch it if we want Workers initialised with `--from-dash` to be capable of running `wrangler d1` commands.

penalosa (commented):
Looks mostly good, but could you add a test in `init.test.ts` (or expand an existing test to include D1)?

threepointone (approved):
the code looks fine, stamping. unfortunate that this will make projects with d1 bindings a little slower to pull down, but that's not something we can fix here anyway.

petebacondarwin (approved):

petebacondarwin:
Just running e2e tests in case this breaks one of them.

PR #6259, opened 2024-07-13
Import Durable Object classes in type generation

ottomated:

Adds type safety to Durable Objects in type generation.

Old:
```ts
interface Env {
  DO: DurableObjectNamespace;
}
```

New:
```ts
interface Env {
  DO: DurableObjectNamespace<import("./src/index").MyDurableObject>;
}
```

Import specifier is generated relative from the output file to the worker entrypoint.

- Tests
  - [x] Included
  - [x] I don't know
  - [x] Included
- Public documentation
  - [x] Not necessary because: doesn't require any user action to benefit


packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -112,6 +131,7 @@ const bindingsConfigMock: Omit<
 		bindings: [
 			{ name: "DURABLE_TEST1", class_name: "Durability1" },
 			{ name: "DURABLE_TEST2", class_name: "Durability2" },
+			{ name: "DURABLE_TEST3", class_name: "Durability3" },
```
petebacondarwin, line 134: What about a Durable Object defined in another Worker?
Can you add a test to ensure that this works correctly?
Particularly in the case where there is a matching file in the local Worker which may be picked up incorrectly.

petebacondarwin (changes requested):
Thanks for this. Please can you check and test the external DO use case.

packages/wrangler/src/__tests__/type-generation.test.ts
ottomated: Test added!

packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -286,8 +319,10 @@ describe("generateTypes()", () => {
 			ANOTHER: \\"thing\\";
 			\\"some-other-var\\": \\"some-other-value\\";
 			OBJECT_VAR: {\\"enterprise\\":\\"1701-D\\",\\"activeDuty\\":true,\\"captian\\":\\"Picard\\"};
-			DURABLE_TEST1: DurableObjectNamespace;
-			DURABLE_TEST2: DurableObjectNamespace;
+			DURABLE_DIRECT_EXPORT: DurableObjectNamespace<import(\\"./index\\").DurableDirect>;
+			DURABLE_RE_EXPORT: DurableObjectNamespace<import(\\"./index\\").DurableReexport>;
+			DURABLE_NO_EXPORT: DurableObjectNamespace /* DurableNoexport */;
+			DURABLE_EXTERNAL: DurableObjectNamespace /* DurableExternal from external-worker */;
```
petebacondarwin, line 325: 🎉

petebacondarwin (approved):

PR #6273, opened 2024-07-16
Add 240/ ranges to Miniflare

penalosa:

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/packages/miniflare/src/plugins/core/index.ts
+++ b/packages/miniflare/src/plugins/core/index.ts
@@ -790,7 +790,7 @@ export function getGlobalServices({
 			network: {
 				// Allow access to private/public addresses:
 				// https://github.com/cloudflare/miniflare/issues/412
-				allow: ["public", "private"],
+				allow: ["public", "private", "240.0.0.0/4"],
 				deny: [],
 				tlsOptions: {
 					trustBrowserCas: true,


irvinebroque:
@kentonv - any thoughts on just enabling this for everyone vs. needing to make it opt-in via some way of configuring Wrangler?

kentonv (commented):
I guess this is fine, given that "private" is already listed.

But I'm a little worried about the fact that we allow "private" here. This makes `fetch()` in test behave differently than in production, and also makes `wrangler dev` unsafe to expose to the internet (which hopefully people aren't doing anyway, but someone might).

The _right_ way for people to point their worker at test backends running on localhost is to represent them as bindings. A binding can point anywhere, while still keeping global `fetch()` restricted to the internet. Plus then in production, you can have the binding point at a private origin server or Argo Tunnel or something. (We, uh, don't support all those options today in production, but we should.)

Similarly, the right way to support `240.0.0.0` would be via a binding. (`workerd` supports such bindings today -- you'd define a separate network binding with a different allow list.)

But this all requires a fair amount of user education. Maybe given we already list "private" here, the best thing to do for now is punt on this?

irvinebroque:
Makes sense, think need to punt on that but I hear you about the "should be bindings" bit.

Think clearest path is to ship this as-is. We know that it solves the problem. No extra config.

RamIdeas (approved):

PR #6290, opened 2024-07-19
Remove warning about 1mb+ Worker size

irvinebroque:


irvinebroque:
@petebacondarwin — assume there might be a bit more here to get this over the line, but just opening to start the ball rolling. I see people trip up on this, made sense with previous constraints but no longer.

packages/wrangler/src/deployment-bundle/bundle-reporter.ts
```diff
@@ -40,12 +40,6 @@ export async function printBundleSize(
 				: chalk.green(bundleReport);
 
 	logger.log(`Total Upload: ${colorizedReport}`);
-
-	if (gzipSize > ALLOWED_INITIAL_MAX && !process.env.NO_SCRIPT_SIZE_WARNING) {
```
CarmenPopoviciu, line 44: @irvinebroque I can take over this PR as on-call, but I need to understand a bit better what it is that we're trying to do. Is it just the copy that became irrelevant here? The 1MB Worker size is retained in `ALLOWED_INITIAL_MAX`, which is used to calculate the colorized report just above (see `logger.log(`Total Upload: ${colorizedReport}`);`). Is that value of 1MB outdated as well? If yes, do we have a new max?

CarmenPopoviciu: perhaps there's an internal ticket/external GH issue you can point me to for more details? 🙏

petebacondarwin:
Closing in favour of https://github.com/cloudflare/workers-sdk/pull/6381 which contains this.

PR #2074, opened 2022-10-24
Generate DTS `wrangler-overrides.d.ts` from config

JacobMGEvans:
The type command aggregates the bindings from config and the rules for modules, then generates a DTS file for both service workers `declare global { ... }` or the preferred module workers `interface Env { ... }`

The rules for modules generates `declare module` based on the module type, the 3 types `Text`, `Data` & `CompiledWasm`
Module Example Outputs:

**CompiledWasm**
```ts
declare module "**/*.wasm" {
const value: WebAssembly.Module;
export default value;
}
``` 

**Data** 
```ts
declare module "**/*.webp" {
const value: ArrayBuffer;
export default value;
}
``` 
**Text**
```ts
declare module "**/*.text" {
const value: string;
export default value;
}
``` 

resolves #2034 
resolves #2033

PR #6295, opened 2024-07-19
feat: add runtime type generation to wrangler types command

andyjessop:
This PR introduces dynamic runtime type generation, according to a user's project configuration.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: This is an experimental feature, so will be quietly released through community channels.


packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Constructs a comprehensive log message for the user after generating runtime types.
+ */
+export function logRuntimeTypesMessage(outFile: string, tsconfigPath: string) {
+	const existingTypes = readExistingTypes(tsconfigPath);
+	const isWorkersTypesInstalled = workersTypesEntryExists(existingTypes);
+	const updatedTypesString = buildUpdatedTypesString(existingTypes, outFile);
+
+	const message = isWorkersTypesInstalled
+		? `\n📣 Replace the existing "@cloudflare/workers-types" entry with the generated types path:`
```
andyjessop, line 13: I'm not sure what is the convention for using emojis here. Without them it's a little dull and harder to read, but maybe we have another way to do this?

packages/wrangler/src/type-generation/index.ts
```diff
@@ -468,6 +487,7 @@ function writeDTSFile({
 				combinedTypeStrings,
 			].join("\n")
 		);
+		logger.log(`Generating project types...\n`);
```
andyjessop, line 490: I referred to this as "project types", but open to suggestions of course.

fixtures/runtime-types/package.json
```diff
@@ -0,0 +1,22 @@
...
+	"version": "0.0.0",
+	"private": true,
+	"scripts": {
+		"deploy": "wrangler deploy",
+		"dev": "wrangler dev",
+		"test": "vitest",
+		"cf-typegen": "wrangler types"
+	},
+	"devDependencies": {
+		"@cloudflare/vitest-pool-workers": "^0.4.5",
```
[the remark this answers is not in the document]
andyjessop: I've removed the fixture, because I don't believe we need one for this change.

packages/wrangler/scripts/deps.ts
```diff
@@ -21,6 +21,7 @@ export const EXTERNAL_DEPENDENCIES = [
 	// unenv must be external because it contains unenv/runtime code which needs to be resolved
 	// and read when we are bundling the worker application
 	"unenv",
+	"workerd/worker.mjs",
```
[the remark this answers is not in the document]
andyjessop: There is a console warning if you `requre.resolve` the module:

```
> node -r esbuild-register scripts/bundle.ts

▲ [WARNING] "workerd/worker.mjs" should be marked as external for use with "require.resolve" [require-resolve-not-external]

    src/type-generation/runtime/index.ts:77:45:
      77 │   const worker = readFileSync(require.resolve("workerd/worker.mjs")).toString();
         ╵                                               ~~~~~~~~~~~~~~~~~~~~

```

packages/wrangler/e2e/types.test.ts
```diff
@@ -0,0 +1,89 @@
...
+		);
+		expect(output.stdout).toContain(
+			`"types": ["./.wrangler/types/runtime.d.ts"]`
+		);
+		expect(output.stdout).toContain(
+			`📣 To get Node.js typings, install with "npm i --save-dev @types/node".`
+		);
+		expect(output.stdout).toContain(
+			`Remember to run 'wrangler types --x-with-runtime' again if you change 'compatibility_date' or 'compatibility_flags' in your wrangler.toml.`
+		);
```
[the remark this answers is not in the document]
andyjessop: The formatting is a pain to get right, and makes it brittle to any changes. This way, we're only asserting the important stuff (that the warnings exist), and not the unimportant stuff (whether it's separated by tabs/spaces, or if the layout changes).

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -0,0 +1,100 @@
+import { readFileSync } from "fs";
+import { writeFile } from "fs/promises";
+import { Miniflare } from "miniflare";
+import { ensureDirectoryExists } from "../../utils/log-file";
```
penalosa, line 4: This file doesn't really have anything to do with writing a log-file, I'm not sure it should necessarily be importing from here. It's a pretty simple utility, so maybe it could just be inlined? Or extracted out to it's own file independent of `log-file`?

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -0,0 +1,100 @@
...
+ * - The generated types are written to a file specified by DEFAULT_OUTFILE_RELATIVE_PATH.
+ * - This could be improved by hashing the compat date and flags to avoid unnecessary regeneration.
+ */
+export async function generateRuntimeTypes({
+	config,
+	generateFn = generate,
+	outFile = DEFAULT_OUTFILE_RELATIVE_PATH,
+}: {
+	config: Config;
+	generateFn?: typeof generate;
```
penalosa, line 43: It doesn't look like this option is used by callers of this function—what's the purpose of it?

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -0,0 +1,100 @@
...
+	outFile?: string;
+}) {
+	const { compatibility_date, compatibility_flags = [] } = config;
```
penalosa, line 46: nit, but these could be destructured in the function arguments above

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+	if (!compatibility_date) {
+		throw new Error("Config must have a compatability date.");
```
penalosa, line 49:
```suggestion
		throw new Error("Config must have a compatibility date.");
```

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -0,0 +1,100 @@
...
+	const types = await generateFn({
+		compatibilityDate: compatibility_date,
+		compatibilityFlags: compatibility_flags,
+	});
+
+	await writeFile(outFile, types, "utf8");
+
+	return {
+		outFile,
+	};
```
penalosa, line 63: Why is this an object rather than just a string?

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
penalosa: Emojis are fun, so this seems fine to me? 🤷

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -0,0 +1,79 @@
...
+		: "📣 Add the generated types to the types array in your tsconfig.json:";
+
+	logger.info(`
+✨ Runtime types written to ${outFile}
+${message}
+
+{
+	"compilerOptions": {
+		"types": ${updatedTypesString}
+	}
```
penalosa, line 23:
```suggestion
	"compilerOptions": {
	  ...
		"types": ${updatedTypesString}
		...
	}
```

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+    `);
+	if (isWorkersTypesInstalled) {
+		logger.info('📣 You can now uninstall "@cloudflare/workers-types".');
+	}
+	logger.info(
+		'📣 To get Node.js typings, install with "npm i --save-dev @types/node".'
+	);
```
penalosa, line 31: Can we check whether they have a node related compat flag turned on here?

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Attempts to read the tsconfig.json at the current path.
+ */
+function readExistingTypes(tsconfigPath: string): string[] {
+	if (!existsSync(tsconfigPath)) {
+		return [];
+	}
+
+	try {
+		const tsconfig = JSON.parse(readFileSync(tsconfigPath, "utf-8"));
```
penalosa, line 65: tsconfig files are JSONC, so this may not work. You probably want the [`parseJSONC` utility function](https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/src/parse.ts#L179)

packages/wrangler/src/type-generation/runtime/index.ts
andyjessop: The reason I did this was:
- if we need to extend it by returning the content too, then this way it will be backwards-compatible - i.e. require changes in fewer places, and
- it's more explicit (and therefore readable) what is actually being returned to the calling function.

I can remove it if you think it's overkill though.

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
andyjessop: Good idea, we can do this 👍

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
andyjessop: Thanks, good call.

packages/wrangler/src/type-generation/runtime/index.ts
andyjessop: 🙈

packages/wrangler/src/type-generation/runtime/index.ts
andyjessop: Good spot - that was left in from previous testing.

packages/wrangler/src/type-generation/runtime/index.ts
andyjessop: I'll move it to a new `packageswrangler/src/utils/filesystem.ts` util file.

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
andyjessop: Ok, question here. Do we only want to display this if they have node compat? I'm wondering if for the sake of clarity, our users will be better educated if we leave this in even if it's not strictly necessary.

packages/wrangler/src/type-generation/runtime/index.ts
penalosa: Makes sense!

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
penalosa: We're still in a bit of a weird place with node compat (until the v2 hybrid mode stabilises), and so I'm not sure this is the right place to educate users (I can see users reading this and thinking all they need to do to get Node.js APIs working is to add `@types/node`, when in reality it's a lot more complex and project specific).

Separately, we should make sure to encourage users to use the [right version of `@types/node` ](https://developers.cloudflare.com/workers/languages/typescript/#transitive-loading-of-typesnode-overrides-cloudflareworkers-types). Versions after `20.8.3` will cause unexpected clashes with workers types like `Request` and `Response`

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -0,0 +1,94 @@
...
+✨ Runtime types written to ${outFile}
+${message}
+
+{
+	"compilerOptions": {
+	    ...
+		"types": ${updatedTypesString}
+		...
+	}
+}
```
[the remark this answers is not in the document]
penalosa: Sorry, this was my bad in the previous suggestion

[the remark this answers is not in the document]
andyjessop: Not your fault. I added these manually, but used tabs instead of spaces and it messed up the formatting.

packages/wrangler/src/type-generation/index.ts
```diff
@@ -29,6 +31,11 @@ export function typesOptions(yargs: CommonYargsArgv) {
 			default: "Env",
 			describe: "The name of the generated environment interface",
 			requiresArg: true,
+		})
+		.option("x-with-runtime", {
```
RamIdeas, line 35:
```suggestion
		.option("experimental-with-runtime", {
			alias: "x-with-runtime",
```

packages/wrangler/src/type-generation/index.ts
RamIdeas: Should it match the --x-with-runtime arg and say "Generating runtime types..."

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -0,0 +1,96 @@
...
+ *
+ * @remarks
+ * - The generated types are written to a file specified by DEFAULT_OUTFILE_RELATIVE_PATH.
+ * - This could be improved by hashing the compat date and flags to avoid unnecessary regeneration.
+ */
+export async function generateRuntimeTypes({
+	config: { compatibility_date, compatibility_flags = [] },
+	outFile = DEFAULT_OUTFILE_RELATIVE_PATH,
+}: {
+	config: Config;
```
RamIdeas, line 41:
```suggestion
	config: Pick<Config, "compatibility_date" | "compatibility_flags">;
```

packages/wrangler/src/type-generation/index.ts
andyjessop: This is not for the runtime types, but for the `Env` types, so the wording is for distinguishing between the two.

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -92,14 +94,34 @@ export function validateNodeCompat({
...
-		return "v1";
+		mode = "v1";
 	}
-	if (legacyNodeCompat) {
-		return "legacy";
+	if (node_compat) {
+		mode = "legacy";
 	}
-	return null;
+	mode = null;
```
RamIdeas, line 119: is this a mistake?

andyjessop: Yep...fixing now.

penalosa (approved):
Looks good to me! One additional thing that would be good to see is some sort of comment header at the top of the generated types file that explains what compatibility data was used in its generation, for future optimisations. Something like `// Auto generated runtime types: workerd@1.20240101.0 /20230102+global_navigator`

fixtures/worker-app/worker-configuration.d.ts
```diff
@@ -1,5 +1,5 @@
-// Generated by Wrangler on Thu Mar 28 2024 11:27:36 GMT+0000 (Greenwich Mean Time)
-// by running `wrangler types`
+// Generated by Wrangler on Mon Jul 22 2024 10:41:44 GMT+0200 (Central European Summer Time)
+// by running `wrangler types --x-with-runtime`
```
RamIdeas, line 2: Would be great if this file was reproducible without a diff each time `wrangler types --with-runtime` was run.

Is the timestamp necessary? I think @penalosa's idea of using the workerd version and compat date+flags would work better

RamIdeas (approved):
LGTM! One note about the flag name: `–x-with-runtime`. Rather than "with" can we go with "include": `--x-include-runtime`?

andyjessop:
> LGTM! One note about the flag name: `–x-with-runtime`. Rather than "with" can we go with "include": `--x-include-runtime`?

Can do! Is there any precent for either option, or are we just choosing something that makes sense here?

fixtures/worker-app/worker-configuration.d.ts
andyjessop: Looking back through the history, and the [original PR](https://github.com/cloudflare/workers-sdk/pull/2074/files#diff-3ffb2d8628f77f1a82a17c8d3d13371329bd2a1846cff95da89c41acfa5a3edfR152) that introduced this, there doesn't seem to be a specific reason to keep it.

andyjessop: @RamIdeas @penalosa can I suggest doing this in a follow-up PR? It requires some refactoring so might be better of as an atomic set of changes.

RamIdeas:
> > LGTM! One note about the flag name: `–x-with-runtime`. Rather than "with" can we go with "include": `--x-include-runtime`?
> 
> Can do! Is there any precedent for either option, or are we just choosing something that makes sense here?

Just thinking about this becoming non-experimental and then on-by-default and then using this flag to disable the runtime types `--no-with-runtime` vs `--no-include-runtime`

petebacondarwin:
> > > LGTM! One note about the flag name: `–x-with-runtime`. Rather than "with" can we go with "include": `--x-include-runtime`?
> > 
> > 
> > Can do! Is there any precedent for either option, or are we just choosing something that makes sense here?
> 
> Just thinking about this becoming non-experimental and then on-by-default and then using this flag to disable the runtime types `--no-with-runtime` vs `--no-include-runtime`

Just one thing to consider. I would naturally avoid encouraging people to use the `--no-` prefix option for turning off most flags. Instead I would suggest `--include-runtime=false`/`--with-runtime=false`, which works for either wording.

PR #6306, opened 2024-07-22
fix(create-cloudflare): cloning non-existent repository URL

edmundhung:

When users try to create an application from a remote template using c3, the process can get stuck at the cloning stage if git prompt for credentials (e.g. non-existent / private repository). This happens because the spinner implementation suppresses the prompts from the internal git clone command, preventing users from seeing the prompt. This PR replaces it with static messages instead. 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


threepointone (approved):

packages/create-cloudflare/src/templates.ts
```diff
@@ -410,11 +415,10 @@ const downloadRemoteTemplate = async (src: string) => {
 
 		const tmpDir = await mkdtemp(join(tmpdir(), "c3-template"));
 		await emitter.clone(tmpDir);
-		s.stop(`${brandColor("template")} ${dim("cloned and validated")}`);
```
CarmenPopoviciu, line 413: do we need to update the status here as well?

CarmenPopoviciu: oh, I think you're doing that [here](https://github.com/cloudflare/workers-sdk/pull/6306/files#diff-0717b6ed1b4a85c41d325f24981721e346254c785696159aec9c0e444d768efeR330) instead

CarmenPopoviciu (approved):

packages/create-cloudflare/src/templates.ts
edmundhung: Yep. I moved it out because it wasn't validated yet.

CarmenPopoviciu:
@edmundhung don't forget to add the changeset pls 🙏

edmundhung:
> @edmundhung don't forget to add the changeset pls 🙏

Done

packages/create-cloudflare/src/__tests__/templates.test.ts
```diff
@@ -246,3 +249,32 @@ describe("addWranglerToGitIgnore", () => {
...
+
+	test("should download template using degit", async () => {
+		const mock = mockDegit();
+
+		await downloadRemoteTemplate("cloudflare/workers-sdk");
+
+		expect(mock).toBeCalled();
+	});
+
+	test("should not use a spinner", async () => {
```
RamIdeas, line 273: This needs a better test name or a comment block explaining why.

This test is highly coupled to the code so clarifying it is a regression test and how the regression occurred and how not using spinner avoids it in the future will be more helpful.

Think about this test failing in the future and the dev wondering why they should not use a spinner

RamIdeas: Notably, this test has made it into this PR for fixing "cloning non-existent repository URL". But it also seems this PR is fixing "cloning a private repo" which requires prompting for username/password. That should also be included in the changeset

RamIdeas (changes requested):
LGTM except 2 things:
1. the test for not using `spinner()` needs additional explanation
2. the changeset should mention fixing cloning private repos

.changeset/red-rockets-joke.md
```diff
@@ -2,4 +2,4 @@
 "create-cloudflare": patch
 ---
 
-fix: cloning non-existent or private repository should fail gracefully
+fix: creating an application from a non-existent or private repository will no longer cause the CLI to get stuck in the cloning stage
```
RamIdeas, line 5: Thanks for updating! Can you also add what the expected behaviour now is?

"Now, when cloning a non-existent repo the CLI will exit with the error message '...' and when cloning a private repo you will be prompted for the username/password."

edmundhung: done.

RamIdeas (approved):

PR #6310, opened 2024-07-22
fix(create-cloudflare):  support GitHub URL with subdirectory

edmundhung:

At the moment, users are not able to create an application with a GitHub URL if it is inside a subdirectory, e.g. https://github.com/cloudflare/workers-sdk/tree/main/templates/worker-router

This PR fixes it by transforming the URL to the shorthand format. Note: This only address templates on the main branch as a branch name might includes slashes that span multiple segments in the URL.  

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/thick-moons-call.md
+++ b/.changeset/thick-moons-call.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+fix(create-cloudflare): support GitHub URL with subdirectory
--- a/.changeset/thick-moons-call.md
+++ b/.changeset/thick-moons-call.md
@@ -3,3 +3,5 @@
 ---
 
 fix: support GitHub URL with subdirectory
+
+You can now copy+paste the URL directly from your address bar when you find a template within a (mono)repo on github.com.
--- a/.changeset/thick-moons-call.md
+++ b/.changeset/thick-moons-call.md
@@ -2,4 +2,4 @@
 "create-cloudflare": patch
 ---
 
-fix(create-cloudflare): support GitHub URL with subdirectory
+fix: support GitHub URL with subdirectory
--- a/packages/create-cloudflare/e2e-tests/cli.test.ts
+++ b/packages/create-cloudflare/e2e-tests/cli.test.ts
@@ -166,5 +166,29 @@ describe.skipIf(frameworkToTest || isQuarantineMode())(
 				expect(output).toContain(`no deploy`);
 			},
 		);
+
+		test.skipIf(process.platform === "win32")(
+			"Cloning remote template with full GitHub URL",
+			async () => {
+				const { output } = await runC3(
+					[
+						projectPath,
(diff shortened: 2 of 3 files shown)


.changeset/thick-moons-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+fix(create-cloudflare): support GitHub URL with subdirectory
```
CarmenPopoviciu, line 5:
```suggestion
fix: support GitHub URL with subdirectory
```

packages/create-cloudflare/e2e-tests/cli.test.ts
```diff
@@ -166,5 +166,29 @@ describe.skipIf(frameworkToTest || isQuarantineMode())(
 				expect(output).toContain(`no deploy`);
 			},
 		);
+
+		test.skipIf(process.platform === "win32")(
```
CarmenPopoviciu, line 170: why are we skipping for `win32`?

edmundhung: I was just following the patterns we had on the test file as it looks like we always skip `win32` when running c3 e2e. Only tests that ran with `--version` are not skipped.

Happy to remove the skip check if you think this is unnecessary.

packages/create-cloudflare/src/templates.ts
```diff
@@ -323,7 +323,18 @@ export const processRemoteTemplate = async (args: Partial<C3Args>) => {
 		defaultValue: C3_DEFAULTS.template,
 	});
 
-	const path = await downloadRemoteTemplate(templateUrl);
+	let src = templateUrl;
+
+	// GitHub URL with subdirectory is not supported by degit and has to be transformed.
+	// This only address templates on the main branch as a branch name
+	// might includes slashes that span multiple segments in the URL.
```
RamIdeas, line 330:
```suggestion
	// GitHub URL with subdirectory is not supported by degit and has to be transformed.
	// This only addresses input template URLs on the main branch as a branch name
	// might includes slashes that span multiple segments in the URL and cannot be
	// reliably differentiated from the subdirectory path.
```

.changeset/thick-moons-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+fix: support GitHub URL with subdirectory
```
RamIdeas, line 5:
```suggestion
fix: support GitHub URL with subdirectory

You can now copy+paste the URL directly from your address bar when you find a template within a (mono)repo on github.com.
```

RamIdeas (approved):

PR #6312, opened 2024-07-22
[wrangler] Add experimental assets config and flag

emily-shen:

Add an experimental asset config key and flag.
Doesn't really do anything yet but should have appropriate validation.
See [WC-2389](https://jira.cfdata.org/browse/WC-2389)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: doesn't affect existing e2e tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: we don't want this on the docs yet

### DIFF
--- a/packages/wrangler/src/__tests__/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/configuration.test.ts
@@ -1736,6 +1736,58 @@ describe("normalizeAndValidateConfig()", () => {
 			});
 		});
 
+		describe("[experimental_assets]", () => {
+			it("should override `experimental_assets` config defaults with provided values", () => {
+				const expectedConfig: RawConfig = {
+					experimental_assets: [
+						{
+							directory: "public/",
+							binding: "ASSETS",
+						},
+					],
+				};
+
+				const { config, diagnostics } = normalizeAndValidateConfig(
+					expectedConfig,
+					undefined,
+					{ env: undefined }
+				);
+
+				expect(config).toEqual(expect.objectContaining(expectedConfig));
+				expect(diagnostics.hasErrors()).toBe(false);
+				expect(diagnostics.hasWarnings()).toBe(false);
+			});
+
+			it("should error on invalid `experimental_assets` values", () => {
+				const expectedConfig = {
+					experimental_assets: [
+						{
+							binding: 2,
+							notAField: "boop",
+						},
+					],
+				};
+
+				const { config, diagnostics } = normalizeAndValidateConfig(
+					expectedConfig as unknown as RawConfig,
+					undefined,
+					{ env: undefined }
+				);
+
+				expect(config).toEqual(expect.objectContaining(expectedConfig));
+				expect(diagnostics.hasWarnings()).toBe(true);
+				expect(diagnostics.renderWarnings()).toMatchInlineSnapshot(`
+					"Processing wrangler configuration:
+					  - Unexpected fields found in \\"experimental_assets[0]\\" field: \\"notAField\\""
+				`);
+				expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+					"Processing wrangler configuration:
+					  - \\"experimental_assets[0]\\" should have a string \\"directory\\" field but got {\\"binding\\":2,\\"notAField\\":\\"boop\\"}.
+					  - \\"experimental_assets[0]\\" should, optionally, have a string \\"binding\\" field, but got {\\"binding\\":2,\\"notAField\\":\\"boop\\"}."
+				`);
+			});
+		});
+
 		describe("[browser]", () => {
 			it("should error if browser is an array", () => {
 				const { diagnostics } = normalizeAndValidateConfig(
--- a/packages/wrangler/src/__tests__/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/configuration.test.ts
@@ -1739,12 +1739,10 @@ describe("normalizeAndValidateConfig()", () => {
 		describe("[experimental_assets]", () => {
 			it("should override `experimental_assets` config defaults with provided values", () => {
 				const expectedConfig: RawConfig = {
-					experimental_assets: [
-						{
-							directory: "public/",
-							binding: "ASSETS",
-						},
-					],
+					experimental_assets: {
+						directory: "public/",
+						binding: "ASSETS",
+					},
 				};
 
 				const { config, diagnostics } = normalizeAndValidateConfig(
@@ -1760,12 +1758,10 @@ describe("normalizeAndValidateConfig()", () => {
 
 			it("should error on invalid `experimental_assets` values", () => {
 				const expectedConfig = {
-					experimental_assets: [
-						{
-							binding: 2,
-							notAField: "boop",
-						},
-					],
+					experimental_assets: {
+						binding: 2,
+						notAField: "boop",
+					},
 				};
 
 				const { config, diagnostics } = normalizeAndValidateConfig(
@@ -1778,12 +1774,12 @@ describe("normalizeAndValidateConfig()", () => {
 				expect(diagnostics.hasWarnings()).toBe(true);
 				expect(diagnostics.renderWarnings()).toMatchInlineSnapshot(`
 					"Processing wrangler configuration:
-					  - Unexpected fields found in \\"experimental_assets[0]\\" field: \\"notAField\\""
+					  - Unexpected fields found in experimental_assets field: \\"notAField\\""
 				`);
 				expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
 					"Processing wrangler configuration:
-					  - \\"experimental_assets[0]\\" should have a string \\"directory\\" field but got {\\"binding\\":2,\\"notAField\\":\\"boop\\"}.
-					  - \\"experimental_assets[0]\\" should, optionally, have a string \\"binding\\" field, but got {\\"binding\\":2,\\"notAField\\":\\"boop\\"}."
+					  - \\"experimental_assets.directory\\" is a required field.
+					  - Expected \\"experimental_assets.binding\\" to be of type string but got 2."
 				`);
 			});
 		});
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -34,6 +34,7 @@ import type {
 	DeprecatedUpload,
 	DispatchNamespaceOutbound,
 	Environment,
+	ExperimentalAssets,
 	RawEnvironment,
 	Rule,
 	TailConsumer,
@@ -2050,49 +2051,43 @@ const validateAssetsConfig: ValidatorFn = (diagnostics, field, value) => {
 		return true;
 	}
 
-	if (!Array.isArray(value) || value === null) {
+	if (typeof value !== "object" || value === null) {
 		diagnostics.errors.push(
-			`The field "${field}" should be an array, but got value ${JSON.stringify(
-				value
-			)} of type ${typeof value}.`
+			`"${field}" should be an object, but got value ${JSON.stringify(
+				field
+			)} of type ${typeof value}`
 		);
 		return false;
 	}
 
 	let isValid = true;
-	for (let i = 0; i < value.length; i++) {
-		const assetConfig = value[i];
 
-		if (typeof assetConfig !== "object" || assetConfig === null) {
-			diagnostics.errors.push(
-				`"${field}[${i}]" should be an object, but got value ${JSON.stringify(
-					assetConfig
-				)} of type ${typeof assetConfig}`
-			);
-			isValid = false;
-		}
-
-		if (!isRequiredProperty(assetConfig, "directory", "string")) {
-			diagnostics.errors.push(
-				`"${field}[${i}]" should have a string "directory" field but got ${JSON.stringify(assetConfig)}.`
-			);
-			isValid = false;
-		}
-
-		if (!isOptionalProperty(assetConfig, "binding", "string")) {
-			diagnostics.errors.push(
-				`"${field}[${i}]" should, optionally, have a string "binding" field, but got ${JSON.stringify(assetConfig)}.`
-			);
-			isValid = false;
-		}
-
-		validateAdditionalProperties(
+	// ensure we validate all props before we show the validation errors
+	// this way users have all the necessary info to fix all errors in one go
+	isValid =
+		validateRequiredProperty(
 			diagnostics,
-			`"${field}[${i}]"`,
-			Object.keys(assetConfig),
-			["directory", "binding"]
-		);
-	}
+			field,
+			"directory",
+			(value as ExperimentalAssets).directory,
+			"string"
+		) && isValid;
+
+	isValid =
+		validateOptionalProperty(
+			diagnostics,
+			field,
+			"binding",
+			(value as ExperimentalAssets).binding,
+			"string"
+		) && isValid;
+
+	isValid =
+		validateAdditionalProperties(diagnostics, field, Object.keys(value), [
+			"directory",
+			"binding",
+		]) && isValid;
+
 	return isValid;
 };
 
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -1222,6 +1222,14 @@ function normalizeAndValidateEnvironment(
 			isObjectWith("crons"),
 			{ crons: [] }
 		),
+		experimental_assets: inheritable(
+			diagnostics,
+			topLevelEnv,
+			rawEnv,
+			"experimental_assets",
+			validateAssetsConfig,
+			undefined
+		),
 		usage_model: inheritable(
 			diagnostics,
 			topLevelEnv,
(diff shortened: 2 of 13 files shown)


packages/wrangler/src/deploy/index.ts
```diff
@@ -108,6 +109,12 @@ export function deployOptions(yargs: CommonYargsArgv) {
 				requiresArg: true,
 				hidden: true,
 			})
+			.option("experimental-assets", {
+				describe: "(Experimental) Static assets to be served",
```
CarmenPopoviciu, line 113:
```suggestion
				describe: "Static assets to be served",
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2037,6 +2045,40 @@ const validateCflogfwdrBinding: ValidatorFn = (diagnostics, field, value) => {
...
+		return true;
+	}
+	if (!Array.isArray(value) || value === null) {
+		return false;
+	}
+	let isValid = true;
+	for (const assetConfig of value) {
+		if (typeof assetConfig !== "object") {
+			diagnostics.errors.push(
+				`Expected the \`experimental_assets\` field to be an object, but got ${typeof value}.`
```
CarmenPopoviciu, line 2059:
```suggestion
				`The field \`experimental_assets\` should be an object, but got ${typeof value}.`
```

CarmenPopoviciu: to keep things consistent with other validation errors

CarmenPopoviciu: though, reading through the code further, there doesn't seem to be much consistency across validation copy :(

packages/wrangler/src/config/validation.ts
```diff
@@ -2037,6 +2045,40 @@ const validateCflogfwdrBinding: ValidatorFn = (diagnostics, field, value) => {
...
+			isValid = false;
+		}
+		const { directory, binding, ...rest } = assetConfig;
+		if (typeof directory !== "string") {
+			diagnostics.errors.push(
+				`\`experimental_assets\` should have a string "directory" field.`
+			);
+			isValid = false;
+		}
+		if (binding && typeof binding !== "string") {
```
CarmenPopoviciu, line 2070:
```suggestion
		if (!isOptionalProperty(value, "binding", "string")) {
```

packages/wrangler/src/deploy/index.ts
```diff
@@ -270,14 +277,43 @@ export async function deployHandler(
...
+		throw new UserError(
+			`The directory provided for the experimental-assets feature does not exist:\n` +
+				`${path.resolve(process.cwd(), args.experimentalAssets)}`
+		);
+	} else {
+		config.experimental_assets?.forEach((x) => {
+			const assetDir = path.resolve(process.cwd(), x.directory);
+			if (!existsSync(assetDir)) {
+				throw new UserError(
+					`The directory provided for the experimental-assets feature does not exist:\n` +
```
CarmenPopoviciu, line 304:
```suggestion
					`The directory specified by the "experimental_assets" field in your configuration file does not exist:\n` +
```

packages/wrangler/src/deploy/index.ts
```diff
@@ -270,14 +277,43 @@ export async function deployHandler(
...
 			"Cannot use Assets and Workers Sites in the same Worker."
 		);
 	}
 
+	if (
+		args.experimentalAssets &&
+		!existsSync(path.resolve(process.cwd(), args.experimentalAssets))
+	) {
+		throw new UserError(
+			`The directory provided for the experimental-assets feature does not exist:\n` +
```
CarmenPopoviciu, line 296:
```suggestion
			`The directory specified by the "--experimental-assets" command line argument does not exist:\n` +
```

packages/wrangler/src/dev.tsx
```diff
@@ -765,6 +772,28 @@ export async function startDev(args: StartDevOptions) {
...
 		}
 
+		if (args.experimentalAssets || config.experimental_assets) {
+			args.forceLocal = true;
+			if (
+				args.experimentalAssets &&
+				!existsSync(path.resolve(process.cwd(), args.experimentalAssets))
+			) {
+				throw new UserError(
+					`The directory provided for the experimental-assets feature does not exist:\n` +
```
CarmenPopoviciu, line 782:
```suggestion
					`The directory specified by the "--experimental-assets" command line argument does not exist:\n` +
```

packages/wrangler/src/dev.tsx
```diff
@@ -765,6 +772,28 @@ export async function startDev(args: StartDevOptions) {
...
+						`${path.resolve(process.cwd(), args.experimentalAssets)}`
+				);
+			} else {
+				config.experimental_assets?.forEach((x) => {
+					if (!existsSync(path.resolve(process.cwd(), x.directory))) {
+						throw new UserError(
+							`The directory provided for the experimental-assets feature does not exist:\n` +
```
CarmenPopoviciu, line 789:
```suggestion
							`The directory specified by the "experimental_assets" field in your configuration file does not exist:\n` + +
```

CarmenPopoviciu (approved):
Approving Emily's changes on behalf of Workers + Assets team. We should get another review though, since I added some small fixes myself

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -76,6 +76,7 @@ type Props = {
 	compatibilityDate: string | undefined;
 	compatibilityFlags: string[] | undefined;
 	assetPaths: AssetPaths | undefined;
+	experimentalAssets: string | undefined;
```
RamIdeas, line 79: This prop has been added but there are no changes to the implementation of `deploy()`? Is that correct?

RamIdeas: Ah I see this PR is just adding the flag and validation

packages/wrangler/src/deploy/index.ts
```diff
@@ -300,6 +336,10 @@ export async function deployHandler(
 					args.siteExclude
 				);
 
+	//TODO: decide what to do with multiple directories
```
RamIdeas, line 339: If these multiple directories are separate bindings, why are we passing them through as one directory? Better the single directory from args gets added as an entry of this array (with a default binding name eg ASSETS?) and this whole array is passed through

RamIdeas (approved):

packages/wrangler/src/deploy/deploy.ts
CarmenPopoviciu: yes that is correct. This just preps the boilerplate for us, to when we're ready to start implementing the new behaviour

packages/wrangler/src/deploy/index.ts
CarmenPopoviciu: ah thx for catching this!! I meant to add a comment as well when I reviewed Emily's changes!

TBH I don't think this is something we want to support in the first iteration...not as I understood it at least. For now we want to support the simplest and most straightforward use case, aka 
`wrangler de --experimental-assets <directory>` and `wrangler deploy --experimental-assets <directory>`.

I'll remove this comment!

packages/wrangler/src/config/validation.ts
```diff
@@ -2037,6 +2045,57 @@ const validateCflogfwdrBinding: ValidatorFn = (diagnostics, field, value) => {
...
+			);
+			isValid = false;
+		}
+
+		if (!isRequiredProperty(assetConfig, "directory", "string")) {
+			diagnostics.errors.push(
+				`"${field}[${i}]" should have a string "directory" field but got ${JSON.stringify(assetConfig)}.`
+			);
+			isValid = false;
+		}
```
petebacondarwin, line 2080: NIT I think you could just do:

```
valid &&= validateRequiredProperty(diagnostics, `${field}[${i}]`, "directory", value, "string");
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2037,6 +2045,57 @@ const validateCflogfwdrBinding: ValidatorFn = (diagnostics, field, value) => {
...
+
+		if (!isOptionalProperty(assetConfig, "binding", "string")) {
+			diagnostics.errors.push(
+				`"${field}[${i}]" should, optionally, have a string "binding" field, but got ${JSON.stringify(assetConfig)}.`
+			);
+			isValid = false;
+		}
+
+		validateAdditionalProperties(
```
petebacondarwin, line 2089:
```suggestion
		isValid &&= validateAdditionalProperties(
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2037,6 +2045,57 @@ const validateCflogfwdrBinding: ValidatorFn = (diagnostics, field, value) => {
...
+			);
+			isValid = false;
+		}
+
+		if (!isOptionalProperty(assetConfig, "binding", "string")) {
+			diagnostics.errors.push(
+				`"${field}[${i}]" should, optionally, have a string "binding" field, but got ${JSON.stringify(assetConfig)}.`
+			);
+			isValid = false;
+		}
```
petebacondarwin, line 2087: I think you can do:

```
valid &&= validateOptionalProperty(diagnostics, `${field}[${i}]`, "binding", value, "string");
```

packages/wrangler/src/deploy/index.ts
```diff
@@ -270,14 +277,43 @@ export async function deployHandler(
...
 		(args.site || config.site)
 	) {
 		throw new UserError(
 			"Cannot use Assets and Workers Sites in the same Worker."
 		);
 	}
 
+	if (
+		args.experimentalAssets &&
+		!existsSync(path.resolve(process.cwd(), args.experimentalAssets))
```
petebacondarwin, line 293: Are asset directory paths relative to the cwd or to the wrangler.toml?

petebacondarwin: The legacy assets stuff appears to use cwd if `--assets` is defined in the command line, and the wrangler.toml directory if using the `assets` property in the config file.

petebacondarwin: See `getAssetPaths()` helper for more info.

petebacondarwin (changes requested):

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: is there added value in doing this given that we still need the if clause in order to handle the diagnostics?

packages/wrangler/src/deploy/index.ts
CarmenPopoviciu: indeed, this was an open question that I mean to bring up today. I'll have a look, at `getAssetsPath`

CarmenPopoviciu: fixed in https://github.com/cloudflare/workers-sdk/pull/6312/commits/7b204ca6084bfc276ce67200198aff951661430a#diff-99329bc31323333c32855181ed3b738e17d682a5f16c825ec49be45d30e54082R8

packages/wrangler/src/config/validation.ts
petebacondarwin: Note that I changed it from `isRequiredProperty()` to `validateRequiredProperty()`. This does handle the diagnostics and keeps the error messages DRY.

packages/wrangler/src/config/validation.ts
petebacondarwin: We need this change otherwise we might return `true` even if there are additional unexpected properties.

packages/wrangler/src/dev.tsx
```diff
@@ -765,6 +772,29 @@ export async function startDev(args: StartDevOptions) {
...
+				args.experimentalAssets &&
+				!existsSync(path.resolve(process.cwd(), args.experimentalAssets))
+			) {
+				throw new UserError(
+					`The directory specified by the "--experimental-assets" command line argument does not exist:\n` +
+						`${path.resolve(process.cwd(), args.experimentalAssets)}`
+				);
+			} else {
+				config.experimental_assets?.forEach((x) => {
+					const assetDir = path.resolve(process.cwd(), x.directory);
```
petebacondarwin, line 787: I think this should also be relative to the wrangler.toml

packages/wrangler/src/deploy/index.ts
```diff
@@ -108,6 +110,12 @@ export function deployOptions(yargs: CommonYargsArgv) {
 				requiresArg: true,
 				hidden: true,
 			})
+			.option("experimental-assets", {
```
RamIdeas, line 113: alias: "x-assets",

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: for posterity....

instead of using the logical AND assignment operator, I instead did
```
isValid = validateRequiredProperty(diagnostics, `${field}[${i}]`, "directory", value, "string") && isValid;
```
This way we ensure that **all** props are validated **before** we show the validation errors. Otherwise, we're forcing our users into an infinite fix-then-revalidate loop, which is a very annoying user experience. As a user, I just want to see everything that's wrong with my config so I can fix it all in one go and move on. I added a comment to reflect this.

AFAICT this is not the de facto state of things when it comes to config validation, but IMHO, it should be

packages/wrangler/src/deploy/index.ts
```diff
@@ -289,35 +290,29 @@ export async function deployHandler(
...
-	if (
-		args.experimentalAssets &&
-		!existsSync(
-			path.resolve(experimentalAssetsBasePath, args.experimentalAssets)
-		)
-	) {
-		throw new UserError(
-			`The directory specified by the "--experimental-assets" command line argument does not exist:\n` +
-				`${path.resolve(experimentalAssetsBasePath, args.experimentalAssets)}`
+	const experimentalAssets = args.experimentalAssets
```
CarmenPopoviciu, line 293: this is merging args and config into one `ExperimentalAssets` shape for convenience. I also expect this will come in handy once we start to implement the dev/deploy behaviour, Same is done in `dev`

packages/wrangler/src/__tests__/configuration.test.ts
```diff
@@ -1739,12 +1739,10 @@ describe("normalizeAndValidateConfig()", () => {
 		describe("[experimental_assets]", () => {
 			it("should override `experimental_assets` config defaults with provided values", () => {
 				const expectedConfig: RawConfig = {
-					experimental_assets: [
-						{
-							directory: "public/",
-							binding: "ASSETS",
-						},
-					],
+					experimental_assets: {
```
CarmenPopoviciu, line 1742: as per our sync call today, moving away from array support, as this is not something for this iteration

packages/wrangler/src/deploy/index.ts
```diff
@@ -375,7 +367,7 @@ export async function deployHandler(
 		jsxFragment: args.jsxFragment,
 		tsconfig: args.tsconfig,
 		routes: args.routes,
-		experimentalAssets: experimentalAssetsDirectory,
+		experimentalAssets: experimentalAssets?.directory,
```
petebacondarwin, line 370: I think this needs to be the `resolvedExperimentalAssetsPath`, right?

packages/wrangler/src/deployment-bundle/entry.ts
```diff
@@ -59,7 +59,7 @@ export async function getEntry(
 			args.legacyAssets ||
 			config.legacy_assets ||
 			args.experimentalAssets ||
-			config.experimental_assets?.some((x) => x.directory)
+			config.experimental_assets?.directory
```
petebacondarwin, line 62: Should this be just `config.experimental_assets`? Once you have this object the `directory` cannot be undefined (and I think we should have validation that it cannot be empty).

packages/wrangler/src/dev.tsx
```diff
@@ -772,27 +774,31 @@ export async function startDev(args: StartDevOptions) {
...
+			if (!existsSync(resolvedExperimentalAssetsPath)) {
+				const sourceOfTruthMessage = args.experimentalAssets
+					? '"--experimental-assets" command line argument'
+					: '"experimental_assets.directory" field in your configuration file';
+
 				throw new UserError(
-					`The directory specified by the "--experimental-assets" command line argument does not exist:\n` +
-						`${path.resolve(process.cwd(), args.experimentalAssets)}`
+					`The directory specified by the ${sourceOfTruthMessage} does not exist:\n` +
+						`${resolvedExperimentalAssetsPath}`
```
petebacondarwin, line 797: I feel a bit bad that this is a duplicate of the same code in the deploy command...

packages/wrangler/src/deploy/index.ts
CarmenPopoviciu: ah yes, thank you for catching that!

packages/wrangler/src/dev.tsx
CarmenPopoviciu: yeah, I'm aware of the code duplication. I can abstract the user error logic into a separate fn if that helps. I don't have strong opinions either way, other than that we should not abstract all of this code block away, as that might become unreadable. Plus we might see some differences as we move further into the implementation.

packages/wrangler/src/deployment-bundle/entry.ts
CarmenPopoviciu: true. your comment however triggered smth else tho, and made me wonder what happens if `directory` is an empty string....and sure enough, `directory = ""` passes validation just fine. To be clear, this seems to be a general thing, not the `experimental_assets` in particular. Feels like we should handle this use case one way or the other, either by failing validation, or resolving to `cwd`? Would that even be expected behaviour...the resolving to `cwd` I mean

packages/wrangler/src/config/validation.ts
petebacondarwin: Very good point! I agree this is how it should work. We should change other validation to follow suit.

CarmenPopoviciu: 💯 this!!! raise here: https://github.com/cloudflare/workers-sdk/issues/6337

packages/wrangler/src/experimental-assets.ts
```diff
@@ -0,0 +1,15 @@
+import path from "node:path";
+import type { Config } from "./config";
+
+/**
+ * Returns the base path of the experimental assets to upload.
+ *
+ */
+export function getExperimentalAssetsBasePath(
+	config: Config,
+	experimentalAssetsDirectory: string | undefined
```
petebacondarwin, line 10: NIT - it would be clearer if this was called something like `experimentalAssetsCommandLineArg`.

petebacondarwin (approved):

packages/wrangler/src/deployment-bundle/entry.ts
CarmenPopoviciu: addressed here: https://github.com/cloudflare/workers-sdk/pull/6346

packages/wrangler/src/experimental-assets.ts
CarmenPopoviciu: addressed here: https://github.com/cloudflare/workers-sdk/pull/6346

PR #6318, opened 2024-07-23
Log worker script startup time

danlapid:

Adds a log for worker startup time as per EW-8590.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: we just add a log line to give users more information

### DIFF
--- a/packages/wrangler/src/deploy/deploy.ts
+++ b/packages/wrangler/src/deploy/deploy.ts
@@ -707,10 +707,11 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			}
 		}
 
-		printBindings({ ...withoutStaticAssets, vars: maskedVars });
-
-		if (!props.dryRun) {
+		if (props.dryRun) {
+			printBindings({ ...withoutStaticAssets, vars: maskedVars });
+		} else {
 			await ensureQueuesExistByConfig(config);
+			let bindingsPrinted = false;
 
 			// Upload the script so it has time to propagate.
 			// We can also now tell whether available_on_subdomain is set
@@ -722,6 +723,7 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 					pipeline_hash: string | null;
 					mutable_pipeline_id: string | null;
 					deployment_id: string | null;
+					startup_time_ms: number;
 				}>(
 					workerUrl,
 					{
@@ -736,6 +738,10 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 						excludeScript: "true",
 					})
 				);
+
+				logger.log("Worker Startup Time:", result.startup_time_ms, "ms");
+				bindingsPrinted = true;
+				printBindings({ ...withoutStaticAssets, vars: maskedVars });
 
 				deploymentId = addHyphens(result.deployment_id) ?? result.deployment_id;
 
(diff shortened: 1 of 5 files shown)


kflansburg:
@taylorlee is this number reliable?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -737,6 +738,8 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 					})
 				);
 
+				logger.log("Script Startup Time: ", result.startup_time_ms, "ms");
```
irvinebroque, line 741: We should say "Worker" rather than "Script"

irvinebroque:
Would like to preview before we merge and release

packages/wrangler/src/deploy/deploy.ts
danlapid: Switched it

tanushree-sharma:
I'd prefer the startup time to be logged ahead of the bindings that we report on upload. Fits naturally to be the next metric after Worker size too!

tanushree-sharma:
Thanks @danlapid 

Before:
image


After:
image

irvinebroque (commented):
LGTM will let someone on Authoring team ✅

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -737,6 +738,9 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 					})
 				);
 
+				logger.log("Worker Startup Time: ", result.startup_time_ms, "ms");
```
petebacondarwin, line 741: In all the tests this value is `undefined`. That doesn't seem right?
Are those tests providing a mock response for the deployment?
Does that mock need to be updated with the startup_time_ms?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -760,6 +764,7 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 					}
 				}
 			} catch (err) {
+				printBindings({ ...withoutStaticAssets, vars: maskedVars });
```
petebacondarwin, line 767: Having this in the catch... could it result in printing the bindings twice? What happens if an error is thrown after the bindings have already been printed?

penalosa:
@danlapid I've re-added the PR template—would you be able to fill it out? In particular, this change will probably need a changeset

packages/wrangler/src/deploy/deploy.ts
danlapid: It seemed to me like the code after `printBindings` isn't fallible as it's just a few prints, I guess that could change in the future.
I could add a flag to check if the bindings have already been printed if you feel like it's more appropriate.

packages/wrangler/src/deploy/deploy.ts
danlapid: Updated.

packages/wrangler/src/deploy/deploy.ts
danlapid: Added a flag.

petebacondarwin (approved):

PR #6322, opened 2024-07-24
Add cloudchamber scopes to existing scopes instead of replacing them

IRCody:
When using any cloudchamber command this adds the cloudchamber scope, if not present, in addition to the existing scopes instead of replacing them.

Before this PR users had to re-auth when switching between cloudchamber/non-cloudchamber commands. This PR makes it so cloudchamber users only have to auth one additional time to add cloudchamber the cloudchamber scopes instead of every time they switch between cloudchamber/non-cloudchamber commands.

If the user isn't logged in and they run a cloudchamber command this also adds the defaultScopes + the cloudchamber scope as part of  that login.

- Tests
  - [ ] TODO (before merge): 
  - [ ] Included
  - [x] Not necessary because: changes which scopes are present only.
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because: 
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: This should be mostly transparent to users except they will no longer need to re-authenticate as often because cloudchamber was replacing existing scopes.

### DIFF
--- a/.changeset/gorgeous-readers-move.md
+++ b/.changeset/gorgeous-readers-move.md
@@ -1,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+Add cloudchamber scope to existing scopes instead of replacing them.
+
+When using any cloudchamber command the cloudchamber scope will now be added to the existing scopes instead of replacing them.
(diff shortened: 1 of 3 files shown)


mikenomitch (approved):
Looks generally good to me, but I'd get an eng to review too!

petebacondarwin (approved):

petebacondarwin:
The tests for cloudchamber mock out the auth flow, so there is no way to test this change.
Given that the scope of the change is limited to the cloudchamber commands, I feel comfortable approving this.

PR #6329, opened 2024-07-24
Add header + caching to runtime type generation

### LINKED ISSUE #6575: 🚀 Feature Request: Add a command to check that the generated types are in sync
### Describe the solution

In my CI job, I'd like to have a check that the `worker-configuration.d.ts` file is actually in sync with the worker configuration (i.e. that devs changing the configuration don't forget to run `npm run cf-typegen`) so that typescript type checking can be relied on.

I attempted to create a GHA job running `npm run cf-typegen` to regenerate the file and then using https://github.com/marketplace/actions/verify-changed-files to make the job fail if the file is modified. However, this fails because the output of `wrangler types` is not reproducible. The output contains a comment with the timestamp of the generation run so the file would be modified all the time in my CI job.

I see 2 ways to solve that:
- provide a dedicated check mode in `wrangler types` (for instance `wrangler types --check`) which would compare the existing file content with the generated content with special logic to ignore changes in the timestamp, and report differences by a failure exit code
- provide a way to make the `wrangler types` output reproducible (either always or with a dedicated flag) by removing the timestamp, which would allow to use existing tooling checking for modified files after running it again.

penalosa:

Fixes #6575 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: internal optimisation


petebacondarwin:
Converting to draft because this feels like it is not yet ready for review.

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -55,7 +71,7 @@ export async function generateRuntimeTypes({
 		),
 	});
 
-	await writeFile(outFile, types, "utf8");
+	await writeFile(outFile, header + "\n" + types, "utf8");
```
andyjessop, line 74:
```suggestion
	await writeFile(outFile, `${header}\n${types}`, "utf8");
```
I thought we had an ESLint rule for this?

andyjessop (approved):
LGTM! One very minor comment.

PR #6330, opened 2024-07-24
Add --x-dev-env to fixtures tests

RamIdeas:

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/fixtures/external-durable-objects-app/tests/index.test.ts
+++ b/fixtures/external-durable-objects-app/tests/index.test.ts
@@ -30,6 +30,7 @@ describe(
 				experimental: {
 					fileBasedRegistry: true,
 					disableExperimentalWarning: true,
+					devEnv: true,
 				},
 			});
 			await setTimeout(1000);
@@ -38,6 +39,7 @@ describe(
 				experimental: {
 					fileBasedRegistry: true,
 					disableExperimentalWarning: true,
+					devEnv: true,
 				},
 			});
 			await setTimeout(1000);
@@ -46,6 +48,7 @@ describe(
 				experimental: {
 					fileBasedRegistry: true,
 					disableExperimentalWarning: true,
+					devEnv: true,
 				},
 			});
 			await setTimeout(1000);
--- a/fixtures/external-durable-objects-app/tests/index.test.ts
+++ b/fixtures/external-durable-objects-app/tests/index.test.ts
@@ -1,10 +1,9 @@
-import { fork } from "child_process";
 import * as path from "path";
 import { setTimeout } from "timers/promises";
 import { fetch } from "undici";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { unstable_dev } from "wrangler";
-import type { ChildProcess } from "child_process";
+import { runWranglerPagesDev } from "../../shared/src/run-wrangler-long-lived";
 import type { UnstableDevWorker } from "wrangler";
 
 // TODO: reenable when https://github.com/cloudflare/workers-sdk/pull/4241 lands
@@ -15,14 +14,7 @@ describe(
 		let a: UnstableDevWorker;
 		let b: UnstableDevWorker;
 		let c: UnstableDevWorker;
-
-		let dWranglerProcess: ChildProcess;
-		let dIP: string;
-		let dPort: number;
-		let dResolveReadyPromise: (value: unknown) => void;
-		const dReadyPromise = new Promise((resolve) => {
-			dResolveReadyPromise = resolve;
-		});
+		let d: Awaited<ReturnType<typeof runWranglerPagesDev>>;
 
 		beforeAll(async () => {
 			a = await unstable_dev(path.join(__dirname, "../a/index.ts"), {
@@ -49,56 +41,26 @@ describe(
 				},
 			});
 			await setTimeout(1000);
-			dWranglerProcess = fork(
-				path.join(
-					"..",
-					"..",
-					"..",
-					"packages",
-					"wrangler",
-					"bin",
-					"wrangler.js"
-				),
+
+			d = await runWranglerPagesDev(
+				path.resolve(__dirname, "..", "d"),
+				"public",
 				[
-					"pages",
-					"dev",
-					"public",
 					"--compatibility-date=2024-03-04",
 					"--do=PAGES_REFERENCED_DO=MyDurableObject@a",
-				],
-				{
-					stdio: ["ignore", "ignore", "ignore", "ipc"],
-					cwd: path.resolve(__dirname, "..", "d"),
-				}
-			).on("message", (message) => {
-				const parsedMessage = JSON.parse(message.toString());
-				dIP = parsedMessage.ip;
-				dPort = parsedMessage.port;
-				dResolveReadyPromise(undefined);
-			});
-			await setTimeout(1000);
+					"--port=0",
+				]
+			);
 		});
 
 		afterAll(async () => {
-			await dReadyPromise;
 			await a.stop();
 			await b.stop();
 			await c.stop();
-
-			await new Promise((resolve, reject) => {
-				dWranglerProcess.once("exit", (code) => {
-					if (!code) {
-						resolve(code);
-					} else {
-						reject(code);
-					}
-				});
-				dWranglerProcess.kill("SIGTERM");
-			});
+			await d.stop();
 		});
 
 		it("connects up Durable Objects and keeps state across wrangler instances", async () => {
-			await dReadyPromise;
 			await setTimeout(1000);
 
 			const responseA = await a.fetch(`/`, {
@@ -114,7 +76,7 @@ describe(
 			const responseC = await c.fetch(`/`);
 			const dataC = (await responseC.json()) as { count: number; id: string };
 			expect(dataC.count).toEqual(3);
-			const responseD = await fetch(`http://${dIP}:${dPort}/`);
+			const responseD = await fetch(`http://${d.ip}:${d.port}/`);
 			const dataD = (await responseD.json()) as { count: number; id: string };
 			expect(dataD.count).toEqual(4);
 			const responseA2 = await a.fetch(`/`);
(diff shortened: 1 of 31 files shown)


fixtures/external-durable-objects-app/tests/index.test.ts
```diff
@@ -28,7 +28,6 @@ describe(
 			a = await unstable_dev(path.join(__dirname, "../a/index.ts"), {
 				config: path.join(__dirname, "../a/wrangler.toml"),
 				experimental: {
-					fileBasedRegistry: true,
```
RamIdeas, line 31: Had to remove this to get this test suite to pass cc @penalosa

fixtures/external-durable-objects-app/tests/index.test.ts
```diff
@@ -65,7 +62,6 @@ describe(
 				[
 					"pages",
 					"dev",
-					"--x-registry",
```
RamIdeas, line 65: Had to remove this to get this test suite to pass cc @penalosa

packages/wrangler/src/dev/start-server.ts
```diff
@@ -131,26 +143,62 @@ export async function startDevServer(
...
+			once(devEnv.runtimes[1], "reloadComplete"),
+			once(devEnv.runtimes[0], "error").then((err) => Promise.reject(err)),
+			once(devEnv.runtimes[1], "error").then((err) => Promise.reject(err)),
+		]);
+
+		return {
+			stop: async () => {
+				await Promise.allSettled([stopWorkerRegistry(), devEnv.teardown()]);
+			},
+		};
```
RamIdeas, line 189: Made this function clearer by returning early when props.experimentalDevEnv is set. The rest of the function is for when it is not set

fixtures/shared/src/run-wrangler-long-lived.ts
```diff
@@ -101,10 +107,25 @@ async function runLongLivedWrangler(
...
+						wranglerProcess.pid,
+						e
+					);
+				}
+				// fallthrough to resolve() because either the process is already dead
+				// or don't have permission to kill it or some other reason?
+				// either way, there is nothing we can do and we don't want to fail the test because of this
+				resolve();
+			});
+		});
```
RamIdeas, line 128: This replaces the fixture's runLongLivedWranglers stop implementation with the same implementation from our e2e tests which is more stable (read: actually reliable) than node's child_process.kill('SIGTERM') which sometimes does nothing (especially on windows: always does nothing) causing our tests to hang during cleanup even if they're assertions passed

fixtures/external-durable-objects-app/tests/index.test.ts
```diff
@@ -1,10 +1,9 @@
-import { fork } from "child_process";
 import * as path from "path";
 import { setTimeout } from "timers/promises";
 import { fetch } from "undici";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { unstable_dev } from "wrangler";
-import type { ChildProcess } from "child_process";
+import { runWranglerPagesDev } from "../../shared/src/run-wrangler-long-lived";
```
RamIdeas, line 6: Refactored this test suite to use runWranglerPagesDev as it was doing pretty much the same thing with fork

packages/wrangler/src/dev/start-server.ts
```diff
@@ -131,26 +143,75 @@ export async function startDevServer(
...
+			//   - only resolve when _perceived_ UserWorker is ready
+			//   - throw if _perceived_ UserWorker fails to start
+			// to the eyeball, the ProxyWorker is the _perceived_ UserWorker
+			Promise.race([
+				devEnv.proxy.ready.promise,
+				once(devEnv.proxy, "error").then((err) => Promise.reject(err)),
+			]),
+		]);
+
+		return {
```
RamIdeas, line 198: Made this function clearer by consolidating into one if-block and returning early when props.experimentalDevEnv is set. The rest of the function is for when it is not set

packages/wrangler/src/dev/start-server.ts
```diff
@@ -94,7 +98,15 @@ export async function startDevServer(
 	const devEnv = props.devEnv;
 	const startDevWorkerOptions: StartDevWorkerInput = {
```
RamIdeas, line 99: Had to map more unstable_dev props to StartDevWorkerOptions

packages/wrangler/src/dev/start-server.ts
```diff
@@ -131,26 +143,75 @@ export async function startDevServer(
...
-	// temp: fake these events by calling the handler directly
-	if (!props.experimentalDevEnv) {
-		devEnv.proxy.onConfigUpdate({
-			type: "configUpdate",
-			config: fakeResolvedInput(startDevWorkerOptions),
-		});
-		devEnv.proxy.onBundleStart({
-			type: "bundleStart",
-			config: fakeResolvedInput(startDevWorkerOptions),
+	if (props.experimentalDevEnv) {
```
RamIdeas, line 165: Made this function clearer by consolidating into one if-block and returning early when props.experimentalDevEnv is set. The rest of the function is for when it is not set

packages/wrangler/src/dev/start-server.ts
```diff
@@ -131,26 +143,75 @@ export async function startDevServer(
...
+				const { proxyWorker } = await devEnv.proxy.ready.promise;
+				const url = await proxyWorker.ready;
+
+				props.onReady?.(url.hostname, parseInt(url.port), ev.proxyData);
+			});
 		});
+
+		await devEnv.config.set(startDevWorkerOptions);
+
+		await Promise.all([
```
RamIdeas, line 178: This adheres to unstable_dev's existing behaviour for its returned promise

packages/wrangler/templates/startDevWorker/ProxyWorker.ts
```diff
@@ -124,8 +124,10 @@ export class ProxyWorker implements DurableObject {
 			Object.assign(userWorkerUrl, proxyData.userWorkerUrl);
 
 			// set request.url in the UserWorker
-			const innerUrl = new URL(request.url);
-			Object.assign(innerUrl, proxyData.userWorkerInnerUrlOverrides);
+			const innerUrl = urlFromParts(
+				proxyData.userWorkerInnerUrlOverrides ?? {},
+				request.url
+			);
```
RamIdeas, line 130: This is a refactor to make clearer what is going on in one step rather than two. No change in behaviour.

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -197,6 +197,7 @@ export class LocalRuntimeController extends RuntimeController {
 					userWorkerInnerUrlOverrides: {
 						protocol: data.config?.dev?.origin?.secure ? "https:" : "http:",
 						hostname: data.config?.dev?.origin?.hostname,
+						port: data.config?.dev?.origin?.hostname ? "" : undefined,
```
petebacondarwin, line 200: This looks like a bug fix that deserves a changeset?

packages/wrangler/src/api/startDevWorker/ProxyController.ts
```diff
@@ -282,14 +282,14 @@ export class ProxyController extends Controller<ProxyControllerEventMap> {
 
 		try {
 			await this.runtimeMessageMutex.runWith(async () => {
-				assert(this.proxyWorker, "proxyWorker should already be instantiated");
+				const { proxyWorker } = await this.ready.promise;
```
petebacondarwin, line 285: Another bug fix?

petebacondarwin (changes requested):
I think this shows that there is value in our fixture and running their tests, since it appears that they uncovered some bugs in our new work, right?
Can we add appropriate changesets for the fixes, please?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
RamIdeas: Changeset added in https://github.com/cloudflare/workers-sdk/pull/6330/commits/9f1a6650b23e8c707209ec122adfa53256950c94

packages/wrangler/src/api/startDevWorker/ProxyController.ts
RamIdeas: Not a bug fix. Tests pass without this. But leads to better error messages when other tests fail.

RamIdeas:
Can't reply to [comment](https://github.com/cloudflare/workers-sdk/pull/6330#discussion_r1691826075) because it is on a removed line
> What was the error? 

Basically, it couldn't find MyDurableObject in the dev registry.

It was an obscure error because the test wants calls response.json() but Miniflare's EXTERNAL_SERVICE_WORKER_SCRIPT returns this text (non-json) response:
https://github.com/cloudflare/workers-sdk/blob/fixtures-x-dev-env/packages/wrangler/src/dev/miniflare.ts#L111

fixtures/external-durable-objects-app/tests/index.test.ts
RamIdeas: Basically, it couldn't find MyDurableObject in the dev registry.

It was an obscure error because the test wants calls response.json() but Miniflare's EXTERNAL_SERVICE_WORKER_SCRIPT returns this text (non-json) response:
https://github.com/cloudflare/workers-sdk/blob/fixtures-x-dev-env/packages/wrangler/src/dev/miniflare.ts#L111

@penalosa and I agreed that since we're trying to default-on --x-dev-env before we default-on --x-registry that we should remove this flag in these tests for now and address independently

petebacondarwin (approved):

PR #6331, opened 2024-07-24
fix: only warn about miniflare feature support (ai, vectorize, cron) once

threepointone:
image

We have some warnings in local mode dev when trying to use ai bindings / vectorize / cron, but they are printed every time the worker is started. This PR changes the warning to only be printed once per worker start.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: a simple dx tweak
  - [ ] I don't know
  - [x] Not required because: no functional change
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/lemon-items-type.md
+++ b/.changeset/lemon-items-type.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: only warn about miniflare feature support (ai, vectorize, cron) once
+
+We have some warnings in local mode dev when trying to use ai bindings / vectorize / cron, but they are printed every time the worker is started. This PR changes the warning to only be printed once per worker start.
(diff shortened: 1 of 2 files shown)


penalosa (approved):

penalosa:
Might be nice to add a helper to the global logger for this? `logger.warnOnce` or something?

RamIdeas:
> Might be nice to add a helper to the global logger for this? `logger.warnOnce` or something?

Was thinking the same. It would encourage contributions to not fall into the same ux trap

threepointone:
It's a good idea. I'll leave that up to you folks!

PR #6336, opened 2024-07-25
fix(wrangler): Rename more `legacy-assets` things

CarmenPopoviciu:

[Workers + Assets]

This PR renames a whole bunch of `assets` things to `legacy-assets`

🚨 Pls review carefully as this touches many parts of the codebase

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: our current tests should cover us
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: just naming changes
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: just naming changes


RamIdeas (approved):
I don't see any logic changes in this PR so I can confidently say if typescript checks pass then this LGTM!

packages/wrangler/src/dev/local.tsx
```diff
@@ -89,7 +89,8 @@ export async function localPropsToConfigBundle(
 		inspectorPort: props.runtimeInspectorPort,
 		bindings: props.bindings,
 		workerDefinitions: props.workerDefinitions,
-		assetPaths: props.assetPaths,
+		// TODO @Carmen can we rename here?
```
petebacondarwin, line 92: What is the answer here?

petebacondarwin (commented):
Should this have a changeset? I get that it is an internal refactor, so perhaps not?

packages/wrangler/src/dev/local.tsx
CarmenPopoviciu: the answer was yes, and `assetsPath` was renamed to `legacyAssetsPath`, but thank you for spotting the todo comment. I cleaned that up

CarmenPopoviciu:
> Should this have a changeset? I get that it is an internal refactor, so perhaps not?

This was my reasoning as well. These are not user facing changes, since all we're doing is renaming some internal things. The user-facing nomenclature already happened in a separate PR

PR #6346, opened 2024-07-26
fix(wrangler): Add extra `experimental_assets` validation

CarmenPopoviciu:

This commit adds an extra validation rule for the `experimental_assets` config key, that should prevent `directory` brom being an empty string.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: config validation change
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: config validation change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: config validation change

### DIFF
--- a/packages/wrangler/src/config/validation-helpers.ts
+++ b/packages/wrangler/src/config/validation-helpers.ts
@@ -232,6 +232,27 @@ export const isString: ValidatorFn = (diagnostics, field, value) => {
 };
 
 /**
+ * Validate that the field is a non-empty string.
+ */
+export const isNonEmptyString: ValidatorFn = (diagnostics, field, value) => {
+	if (value !== undefined && typeof value !== "string") {
+		diagnostics.errors.push(
+			`Expected "${field}" to be of type string but got ${JSON.stringify(
+				value
+			)}.`
+		);
+		return false;
+	}
+
+	if (value?.trim() === "") {
+		diagnostics.errors.push(`Expected "${field}" cannot be an empty string.`);
+		return false;
+	}
+
+	return true;
+};
+
+/**
  * Validate that the `name` field is compliant with EWC constraints.
  */
 export const isValidName: ValidatorFn = (diagnostics, field, value) => {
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/config/validation-helpers.ts
```diff
@@ -231,6 +231,27 @@ export const isString: ValidatorFn = (diagnostics, field, value) => {
...
+ */
+export const isNonEmptyString: ValidatorFn = (diagnostics, field, value) => {
+	if (value !== undefined && typeof value !== "string") {
+		diagnostics.errors.push(
+			`Expected "${field}" to be of type string but got ${JSON.stringify(
+				value
+			)}.`
+		);
+		return false;
+	}
```
penalosa, line 245: Could this call the existing `isString()` validation function? It seems like the same logic

packages/wrangler/src/config/validation-helpers.ts
```diff
@@ -231,6 +231,27 @@ export const isString: ValidatorFn = (diagnostics, field, value) => {
...
+
+	if (value?.trim() === "") {
+		diagnostics.errors.push(`Expected "${field}" cannot be an empty string.`);
```
penalosa, line 248: maybe: 
```suggestion
		diagnostics.errors.push(`Expected "${field}" to be a non-empty string.`);
```

penalosa (approved):
nit + wording suggestion, but otherwise LGTM

petebacondarwin (approved):
I agree with @penalosa's suggestions.

packages/wrangler/src/config/validation-helpers.ts
CarmenPopoviciu: ✅

CarmenPopoviciu:
thank you for reviewing this. Addressed all the comments. Merging

PR #6370, opened 2024-07-30
feat: Create very basic Asset Worker and plumb it into `wrangler dev`

CarmenPopoviciu:
**[Workers + Assets]**
This PR includes work needed to add Assets support for Workers 🚀 

**Architecture Diagram**
Screenshot 2024-07-31 at 21 04 33

Based on the above diagram, this PR does the following:

- creates a new package called `workers-shared` that will host the Asset Server Worker and Router Worker
- scaffolds the Asset Server Worker in some very basic form, with basic configuration. Further behaviour implementation will follow in a subsequent PR
- does the ground work of plumbing ASW into miniflare

 These changes are supported by both `wrangler dev` and `wrangler dev --x-dev-env` 💅 ...

Fixes WC-2423

🐙 Please review per commit 🐙

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: we're not there yet

### DIFF
--- a/.changeset/orange-icons-build.md
+++ b/.changeset/orange-icons-build.md
@@ -1,0 +1,12 @@
+---
+"@cloudflare/workers-shared": minor
+"wrangler": minor
+---
+
+feat: Create very basic Asset Server Worker and plumb it into `wrangler dev`
+
+These changes do the ground work needed in order to add Assets support for Workers in `wrangler dev`. They implement the following:
+
+- it creates a new package called `workers-shared` that hosts the `Asset Server Worker`, and the `Router Worker`in the future
+- it scaffolds the `Asset Server Worker` in some very basic form, with basic configuration. Further behaviour implementation will follow in a subsequent PR
+- it does the ground work of plumbing ASW into Miniflare
--- a/fixtures/workers-with-assets/README.md
+++ b/fixtures/workers-with-assets/README.md
@@ -1,0 +1,23 @@
+# workers-with-assets
+
+`workers-with-assets` is a test fixture that showcases Workers with Assets. This particular fixture sets up an assets-only Workers project.
+
+## dev
+
+To start a dev session you can run
+
+```
+wrangler dev
+```
+
+or
+
+```
+wrangler dev --experimental-assets=./public
+```
+
+## Run tests
+
+```
+npm run test
+```
--- a/fixtures/workers-with-assets/wrangler.toml
+++ b/fixtures/workers-with-assets/wrangler.toml
@@ -1,0 +1,5 @@
+name = "assets-worker"
+compatibility_date = "2024-01-01"
+
+[experimental_assets]
+directory = "./public"
--- a/packages/workers-shared/README.md
+++ b/packages/workers-shared/README.md
@@ -1,0 +1,13 @@
+# `@cloudflare/workers-shared`
+
+This is a package that is used at Cloudflare to power some internal features of [Cloudflare Workers](https://developers.cloudflare.com/workers/), as well as their open-source equivalents here in workers-sdk and Wrangler.
+
+## `asset-server`
+
+The Asset Server Worker.
+
+For more details please refer to the dedicated README file.
+
+> [!NOTE]
+> Since code in this package is used by the Workers infrastructure, it is important that PRs are given careful review with regards to how they could cause a failure in production.
+> Ideally, there should be comprehensive tests for changes being made to give extra confidence about the behavior.
--- a/packages/workers-shared/asset-server-worker/wrangler.toml
+++ b/packages/workers-shared/asset-server-worker/wrangler.toml
@@ -1,0 +1,12 @@
+##
+# Configuration file for the Asset Server Worker
+#
+# Please note that wrangler has a dependency on this file, and will
+# attempt to read it as part of setting up a new Miniflare instance
+# in developemnt mode. We should ensure that any configuration changes
+# to this file are persisted in wrangler as well, when necessary.
+# (see packages/wrangler/src/dev/miniflare.ts -> buildMiniflareOptions())
+##
+name = "asset-server"
+main = "src/index.ts"
+compatibility_date = "2024-07-31"
--- a/packages/wrangler/scripts/deps.ts
+++ b/packages/wrangler/scripts/deps.ts
@@ -22,6 +22,7 @@ export const EXTERNAL_DEPENDENCIES = [
 	// and read when we are bundling the worker application
 	"unenv",
 	"workerd/worker.mjs",
+	"@cloudflare/workers-shared",
 ];
 
 const pathToPackageJson = path.resolve(__dirname, "..", "package.json");
--- a/packages/wrangler/src/api/dev.ts
+++ b/packages/wrangler/src/api/dev.ts
@@ -189,7 +189,7 @@ export async function unstable_dev(
 		localProtocol: options?.localProtocol,
 		httpsKeyPath: options?.httpsKeyPath,
 		httpsCertPath: options?.httpsCertPath,
-		experimentalAssets: options?.experimentalAssets,
+		experimentalAssets: undefined,
 		legacyAssets: options?.legacyAssets,
 		site: options?.site, // Root folder of static assets for Workers Sites
 		siteInclude: options?.siteInclude, // Array of .gitignore-style patterns that match file or directory names from the sites directory. Only matched items will be uploaded.
--- a/packages/wrangler/src/api/startDevWorker/ConfigController.ts
+++ b/packages/wrangler/src/api/startDevWorker/ConfigController.ts
@@ -204,6 +204,10 @@ async function resolveConfig(
 			legacyAssets: Boolean(legacyAssets),
 			script: input.entrypoint,
 			moduleRoot: input.build?.moduleRoot,
+			// getEntry only needs to know if experimental_assets was specified.
+			// The actualy value is not relevant here, which is why not passing
+			// the entire ExperimentalAssets object is fine.
+			experimentalAssets: input?.experimental?.assets?.directory,
 		},
 		config,
 		"dev"
@@ -256,6 +260,9 @@ async function resolveConfig(
 		unsafe: {
 			capnp: input.unsafe?.capnp ?? unsafe?.capnp,
 			metadata: input.unsafe?.metadata ?? unsafe?.metadata,
+		},
+		experimental: {
+			assets: input?.experimental?.assets,
 		},
 	} satisfies StartDevWorkerOptions;
 
--- a/packages/wrangler/src/api/startDevWorker/types.ts
+++ b/packages/wrangler/src/api/startDevWorker/types.ts
@@ -1,6 +1,7 @@
 import type { Config } from "../../config";
 import type {
 	CustomDomainRoute,
+	ExperimentalAssets,
 	Rule,
 	ZoneIdRoute,
 	ZoneNameRoute,
@@ -161,6 +162,9 @@ export interface StartDevWorkerInput {
 		enableServiceEnvironments?: boolean;
 	};
 	unsafe?: Omit<CfUnsafe, "bindings">;
+	experimental?: {
+		assets?: Omit<ExperimentalAssets, "bindings">;
+	};
 }
 
 export type StartDevWorkerOptions = StartDevWorkerInput & {
@@ -258,7 +262,8 @@ export type Binding =
 	| ({ type: "dispatch_namespace" } & Omit<CfDispatchNamespace, "binding">)
 	| ({ type: "mtls_certificate" } & Omit<CfMTlsCertificate, "binding">)
 	| ({ type: "logfwdr" } & Omit<CfLogfwdrBinding, "name">)
-	| { type: `unsafe_${string}` };
+	| { type: `unsafe_${string}` }
+	| { type: "assets" };
 
 export type ServiceFetch = (request: Request) => Promise<Response> | Response;
 
--- a/packages/wrangler/src/deployment-bundle/entry.ts
+++ b/packages/wrangler/src/deployment-bundle/entry.ts
@@ -38,13 +38,14 @@ export async function getEntry(
 		format?: CfScriptFormat | undefined;
 		legacyAssets?: string | undefined | boolean;
 		moduleRoot?: string;
-		experimentalAssets?: string;
+		experimentalAssets?: string | undefined;
 	},
 	config: Config,
 	command: "dev" | "deploy" | "versions upload" | "types"
 ): Promise<Entry> {
 	let file: string;
 	let directory = process.cwd();
+
 	if (args.script) {
 		// If the script name comes from the command line it is relative to the current working directory.
 		file = path.resolve(args.script);
@@ -55,13 +56,10 @@ export async function getEntry(
 				? path.resolve(config.site?.["entry-point"])
 				: // site.entry-point could be a directory
 					path.resolve(config.site?.["entry-point"], "index.js");
-		} else if (
-			args.legacyAssets ||
-			config.legacy_assets ||
-			args.experimentalAssets ||
-			config.experimental_assets
-		) {
+		} else if (args.legacyAssets || config.legacy_assets) {
 			file = path.resolve(getBasePath(), "templates/no-op-worker.js");
+		} else if (args.experimentalAssets || config.experimental_assets) {
+			file = path.resolve(getBasePath(), "templates/no-op-assets-worker.ts");
 		} else {
 			throw new UserError(
 				`Missing entry-point: The entry-point should be specified via the command line (e.g. \`wrangler ${command} path/to/script\`) or the \`main\` config field.`
--- a/packages/wrangler/src/dev.tsx
+++ b/packages/wrangler/src/dev.tsx
@@ -562,6 +562,7 @@ export async function startDev(args: StartDevOptions) {
 				"--local is no longer required and will be removed in a future version.\n`wrangler dev` now uses the local Cloudflare Workers runtime by default. 🎉"
 			);
 		}
+
 		if (args.experimentalLocal) {
 			logger.warn(
 				"--experimental-local is no longer required and will be removed in a future version.\n`wrangler dev` now uses the local Cloudflare Workers runtime by default. 🎉"
@@ -574,6 +575,7 @@ export async function startDev(args: StartDevOptions) {
 				"Passing --inspect is unnecessary, now you can always connect to devtools."
 			);
 		}
+
 		if (args.experimentalPublic) {
 			throw new UserError(
 				"The --experimental-public field has been deprecated, try --legacy-assets instead."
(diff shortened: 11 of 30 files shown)


packages/wrangler/scripts/deps.ts
```diff
@@ -22,6 +22,7 @@ export const EXTERNAL_DEPENDENCIES = [
 	// and read when we are bundling the worker application
 	"unenv",
 	"workerd/worker.mjs",
+	"@cloudflare/workers-shared/dist",
```
CarmenPopoviciu, line 25: marking `@cloudflare/workers-shared/dist` in esbuild as [external](https://esbuild.github.io/api/#external) means that "instead of being bundled, the import will be preserved and will be evaluated at run time instead.". This unfortunately comes with the side effect that `@cloudflare/workers-shared/dist` won't be considered a dependency, and will therefore be ignored in watch mode. This means that when in `dev` mode, any changes to the `asset-server-worker` code (or any code inside `@cloudflare/workers-shared` for that matter) won't trigger a wrangler re-bundle, which we'd ideally want, since `miniflare` references the bundled ASW [here](https://github.com/cloudflare/workers-sdk/pull/6370/files#diff-ca7f0876a94f845d3a9323b32cba14fbfb3add60d26b7a0d86a5fc68beb3ca3aR846).

While this will most likely not impact end-users, since they shouldn't be fiddling with `workers-shared/ASW` anyway, it does create some friction for us as the wrangler team, and our dev experience.

One possible solution I played with, that seemed to do the trick, was to factor the rebuilding of ASW in the wrangler dev process, meaning 👇 (with `pnpm -C ../workers-shared run bundle:asset-server --watch` being the bit of interest here)
```
# packages/wrangler/package.json

"dev": "pnpm run clean && concurrently -c black,blue --kill-others-on-fail false \"pnpm -C ../workers-shared run bundle:asset-server --watch\" \"pnpm run bundle --watch\" \"pnpm run check:type --watch --preserveWatchOutput\""
```

But tbh, I'm not sure how I feel abt that, and since I feel like I'd like to have validation of this work **before** I spend more time on this, I left it out of my PR for the time being

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -837,6 +842,42 @@ export async function buildMiniflareOptions(
...
+	);
+	const assetServerConfigPath = require.resolve(
+		"@cloudflare/workers-shared/asset-server-worker/wrangler.toml"
+	);
+	let assetServerConfig: Config | undefined;
+
+	try {
+		assetServerConfig = readConfig(assetServerConfigPath, {});
+	} catch (err) {
+		//  ignore and continue with some predefined default config
```
CarmenPopoviciu, line 856: dbl check: is this what we want to do here?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -837,6 +842,42 @@ export async function buildMiniflareOptions(
...
+	}
+
+	const assetServerWorker: WorkerOptions | undefined = config.experimentalAssets
+		? {
+				name: assetServerConfig?.name ?? "asset-server",
```
CarmenPopoviciu, line 861: dbl-check: are these sensible defaults?

packages/wrangler/src/deployment-bundle/entry.ts
```diff
@@ -55,13 +55,10 @@ export async function getEntry(
...
 					path.resolve(config.site?.["entry-point"], "index.js");
-		} else if (
-			args.legacyAssets ||
-			config.legacy_assets ||
-			args.experimentalAssets ||
-			config.experimental_assets
-		) {
+		} else if (args.legacyAssets || config.legacy_assets) {
 			file = path.resolve(getBasePath(), "templates/no-op-worker.js");
+		} else if (args.experimentalAssets || config.experimental_assets) {
```
CarmenPopoviciu, line 60: IMHO this should be a no-op Worker or the User Worker itself, but never the Asset Server Worker (ASW) or the Router Worker (RW). Let me explain...

The entry-point Worker will be bundled by `wrangler` via esbuild, with some very wrangler-specific esbuild configuration. In the process, `wrangler` will apply a bunch of middlewares that don't seem relevant to either ASW or RW. `wrangler`'s esbuild configuration was devised for bundling User Workers specifically, and while ASW and RW are Workers themselves, they are external to `wrangler` as a package, are self contained, and will come with their own bundling opinions and requirements. Therefore conflating bundling between the two packages, without an actual need for it, feels very wrong to me. If we want ASW and RW to be common entities to both wrangler and EWC, and to emulate production in `dev` as much as possible, ASW & RW should IMO own their own bundling + configuration, and any other devices needed for them to be deployed. To that effect, if wrangler is to have a dependency on these Workers and spin them up in a new Miniflare instance, then it should be on the already bundled versions of these Worker.

This is at least how I see the ideal world, though it remains to be seen if we can pull this off completely. But for now, this is kind of what I'm striving towards. 

Happy to hear other thoughts though ^.^

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -879,6 +920,9 @@ export async function buildMiniflareOptions(
 					proxy: true,
 				})),
 			},
+			...(config.experimentalAssets
```
CarmenPopoviciu, line 923: technically, for assets-only Workers, we should create a miniflare instance with just the AssetServerWorker here. Removing the entry Worker however broke things, and unfortunately I was kinds clueless as to why. BUT...I might have had a breakthrough in the meanwhile, I just need a bit more time to verify my assumptions & polish the code. I would like to ask that we don't block this PR on that 🙏

packages/workers-shared/README.md
```diff
@@ -0,0 +1,13 @@
...
+This is a package that is used at Cloudflare to power some internal features of [Cloudflare Workers](https://developers.cloudflare.com/workers/), as well as their open-source equivalents here in workers-sdk and Wrangler.
+
+## `asset-server`
+
+The Asset Server Worker.
+
+For more details please refer to the dedicated README file.
+
+> [!NOTE]
+> Since code in this package is used by the Pages infrastructure, it is important that PRs are given careful review with regards to how they could cause a failure in production.
```
GregBrimble, line 12:
```suggestion
> Since code in this package is used by the Workers infrastructure, it is important that PRs are given careful review with regards to how they could cause a failure in production.
```

fixtures/workers-with-assets/wrangler.toml
```diff
@@ -0,0 +1,5 @@
+name = "assets-worker"
+compatibility_date = "2024-01-01"
+
+[experimental_assets]
+directory = "./public"
```
GregBrimble, line 5:
```suggestion
directory = "./public"

```

packages/workers-shared/asset-server-worker/wrangler.toml
```diff
@@ -0,0 +1,3 @@
+name = "asset-server"
+main = "src/index.ts"
+compatibility_date = "2021-11-02" # Oldest version of the runtime
```
GregBrimble, line 3:
```suggestion
compatibility_date = "2024-07-31"

```

Pages' ASW is currently a bit stuck on old stuff, so let's start off with our best foot forwards with this new one if we can.

packages/wrangler/src/api/dev.ts
```diff
@@ -189,7 +189,7 @@ export async function unstable_dev(
 		localProtocol: options?.localProtocol,
 		httpsKeyPath: options?.httpsKeyPath,
 		httpsCertPath: options?.httpsCertPath,
-		experimentalAssets: options?.experimentalAssets,
+		experimentalAssets: undefined,
```
GregBrimble, line 192: Are we tracking unstable_dev support elsewhere? If not, worth leaving a TODO?

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -204,6 +204,10 @@ async function resolveConfig(
 			legacyAssets: Boolean(legacyAssets),
 			script: input.entrypoint,
 			moduleRoot: input.build?.moduleRoot,
+			// getEntry only needs to know if experimental_assets was specified.
```
GregBrimble, line 207: If that's true, might I suggest flattening this to just a boolean?

packages/wrangler/src/dev.tsx
```diff
@@ -1044,6 +1046,7 @@ export async function startApiDev(args: StartDevOptions) {
 				args.accountId ?? configParam.account_id ?? getAccountFromCache()?.id,
 			legacyAssetPaths: legacyAssetPaths,
 			legacyAssetsConfig: configParam.legacy_assets,
+			experimentalAssets: undefined,
```
GregBrimble, line 1049: Same here. Tracked elsewhere? Worth a TODO?

packages/wrangler/src/dev.tsx
```diff
@@ -1181,6 +1184,18 @@ export async function validateDevServerSettings(
...
 	config: Config
 ) {
+	if (
+		(args.legacyAssets ||
+			config.legacy_assets ||
+			args.experimentalAssets ||
+			config.experimental_assets) &&
+		(args.site || config.site)
+	) {
+		throw new UserError(
```
GregBrimble, line 1194: Do we need a similar error preventing `legacyAssets` to be used in conjunction with `experimentalAssets`?

packages/wrangler/templates/no-op-assets-worker.ts
```diff
@@ -0,0 +1,34 @@
...
+ * a User Worker.
+ *
+ * ```bash
+ * wrangler dev --experimental-assets dist
+ * wrangler deploy --experimental-assets dist
+ * ```
+ *
+ * ```toml
+ * name = "assets-only-worker"
+ * compatibility_date = "2024-01-01"
```
GregBrimble, line 17:
```suggestion
 * compatibility_date = "2024-01-01"
 * experimental_assets = "dist"
```

packages/wrangler/templates/no-op-assets-worker.ts
```diff
@@ -0,0 +1,34 @@
...
+ * this Worker is.
+ */
+type Env = {
+	ASSET_SERVER: Fetcher;
+};
+
+export default {
+	async fetch(request: Request, env: Env) {
+		const response = await env.ASSET_SERVER.fetch(request.url, request);
+		return new Response(response.body, response);
```
GregBrimble, line 32:
```suggestion
		return env.ASSET_SERVER.fetch(request);
```

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -408,7 +411,9 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 	// Setup service bindings to external services
 	const serviceBindings: NonNullable<WorkerOptions["serviceBindings"]> = {
 		...config.serviceBindings,
+		...(config.experimentalAssets ? { ASSET_SERVER: "asset-server" } : {}),
```
GregBrimble, line 414: We'll only want to do this when there is no user worker (i.e. when we're using the no-op-assets-worker as our entrypoint). Can we add that check in now? Or do we want to tackle that later when we add in custom user-specified bindings from user workers to ASW2?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -837,6 +842,42 @@ export async function buildMiniflareOptions(
 		}
 	}
 
+	const assetServerModulePath = require.resolve(
+		"@cloudflare/workers-shared/dist/asset-server-worker.mjs"
+	);
+	const assetServerConfigPath = require.resolve(
```
GregBrimble, line 848: nice — cool that we can manage that out in ASW2's directory :)

packages/wrangler/src/dev/miniflare.ts
GregBrimble: Let's just blow up and error? This should never happen, right? We'll ship with that toml, so it should be available. If there's ever an issue there, I'd be happy to have that version of Wrangler or ASW2 package crash out and for us to fix it in a patch.

packages/wrangler/src/dev/miniflare.ts
GregBrimble: Again, I'd just be tempted to write a `requireConfig()` typeguard sort of function and blow up if we fail to validate that before giving to miniflare. What do you think?

packages/wrangler/src/dev/miniflare.ts
GregBrimble: Yup, fine by me. I think we'll eventually get to a place where we're _always_ at least attaching a Router Worker when assets are present, so don't spend too much time getting this running with ASW2 alone.

packages/wrangler/src/deployment-bundle/entry.ts
GregBrimble: Totally agree. I was originally thinking we'd just `import ASW from "@cloudflare/workers-shared";` from within a template and let Wrangler do it's thing, but you're completely correct — ASW2 should get it's own bundling and _that_ is what we should boot up in miniflare, with separation between this and any user worker stuff.

GregBrimble (approved):

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: indeed, I agree with you. I feel this should never happen since we own the Worker and should in theory never publish it with an invalid toml file. But even so, having this extra sanity check `try{} catch()` is not a bad  thing IMO. I think it's reasonable to fail here, like you said. Will add that

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: totes agree! I should have prefixed this question with, if we don't throw on config read... I'll remove

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: indeed, that's kind of the architecture diagram I have in my head as well. The same principle will apply tho, since we'll need to figure out how to make the RW the entry point in the "miniflare pipeline". As it stands right now, that entry point seems to always be the entry-point Worker (which basically wraps the User Worker), so I'll need to have that figured out either way. But again, this can be done as a separate task

packages/workers-shared/asset-server-worker/wrangler.toml
CarmenPopoviciu: yeah, I'm glad you caught this! I meant to ask if this is indeed the campat_date we want to commit to

packages/wrangler/src/api/dev.ts
CarmenPopoviciu: we talked about this in our sync meeting yesterday, and we agreed that there's not much value to add support in `unstable_dev` since:

- the long-term plan is to move away from it
- `unstable_dev` is only used by Pages and fixture tests

But if you have other thoughts happy to hear them

packages/wrangler/src/api/startDevWorker/ConfigController.ts
CarmenPopoviciu: that's very fair! fixed ✅

packages/wrangler/src/dev.tsx
CarmenPopoviciu: same reply as in https://github.com/cloudflare/workers-sdk/pull/6370#discussion_r1699567506

packages/wrangler/src/dev.tsx
CarmenPopoviciu: we totally do, and I was totes convinced we had that in place 😅 . Thx for catching this! Fixed ✅

packages/wrangler/templates/no-op-assets-worker.ts
CarmenPopoviciu: technically, `experimental_assets` doesn't need to be defined here since we won't be doing anything with it. This no-op Worker is meant to proxy any incoming request to ASW, and was as a temp solution until I figure out how to have miniflare instantiate ASW without an entry-point Worker. I expect this template to be removed in a following PR, once that bit is figured out, so leaving as is for now

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: ✅

fixtures/workers-with-assets/README.md
```diff
@@ -0,0 +1,23 @@
...
+To start a dev session you can run
+
+```
+wrangler dev
+```
+
+or
+
+```
+wrangler dev --name=assets-worker --experimental-assets=public
```
RamIdeas, line 16: nit: can we use `--experimental-assets=./public` in our examples to clearly indicate the value is a filepath?

packages/wrangler/src/api/dev.ts
RamIdeas: startWorker will supersede unstable_dev and looks to be considered in this PR – no need to add new features to unstable_dev

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -184,6 +186,7 @@ export type StartDevWorkerOptions = StartDevWorkerInput & {
 		persist: string;
 	};
 	entrypoint: string;
+	experimentalAssets?: ExperimentalAssets;
```
RamIdeas, line 189: This will already be on this type via the `StartDevWorkerInput &` intersection

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -161,6 +162,7 @@ export interface StartDevWorkerInput {
 		enableServiceEnvironments?: boolean;
 	};
 	unsafe?: Omit<CfUnsafe, "bindings">;
+	experimentalAssets?: ExperimentalAssets;
```
RamIdeas, line 165: startWorker's types moved away from conflating resource + binding config. This property should only contain the directory. And the `Binding` union type below should get an additional member: `{ type: 'assets' }`

RamIdeas: Can you also move this to `experimental.assets` instead please

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: ✅

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: I'll be handling this as part of WC-2435

fixtures/workers-with-assets/README.md
CarmenPopoviciu: good callout! fixed ✅

packages/wrangler/src/api/startDevWorker/types.ts
CarmenPopoviciu: ✅

packages/wrangler/src/api/startDevWorker/types.ts
CarmenPopoviciu: ✅

RamIdeas (approved):

.changeset/orange-icons-build.md
```diff
@@ -0,0 +1,41 @@
...
+
+<!-- prettier-ignore-start -->
+                   |
+              wrangler dev
+                   |              workerd
++------------------+---------------------+
+|                  |                     |
+|      +-----------v--------------+      |
+|      |                          |      |
+|      |     NOOP Entry Worker    |      |
```
andyjessop, line 18: Is this the `Router Worker`?

CarmenPopoviciu: not yet. We are introducing that in a separate iteration, as there are still some open questions around it

petebacondarwin: Strictly this is not "noop" it is actually a passthrough Worker. 😄

CarmenPopoviciu: fair!

PR #6373, opened 2024-07-30
fix(C3): make sure to always warn developers trying to use git when their git instance is not properly configured

dario-piotrowicz:

Currently in C3 if someone has git installed but not properly configured, C3 will simply skip
the git initialization, a more clear workflow would be to actually still offer git but warn (and skip it)
in case they don't have it properly configured, that's what this PR is introducing:
Screenshot 2024-07-30 at 16 42 43

> [!NOTE]
> This is quite an edge case for people not too accustomed with git, and it unlikely to effect most C3 users

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because: we don't currently test the git integration in C3
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: not something that needs documentation

### DIFF
--- a/.changeset/slimy-mangos-punch.md
+++ b/.changeset/slimy-mangos-punch.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+fix: make sure to always warn developers trying to use git when their git instance is not properly configured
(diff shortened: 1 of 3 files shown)


jculvey (approved):
Looks like you'll need to adjust some of the tests, but LGTM

dario-piotrowicz:
> Looks like you'll need to adjust some of the tests, but LGTM

ah yeah, I wasn't sure if there were any that needed updating 🙈👍

PR #6374, opened 2024-07-30
[wrangler] wrangler deploy - upload assets 

emily-shen:

[WC-2409](https://jira.cfdata.org/browse/WC-2409) - upload assets
[WC-2428](https://jira.cfdata.org/browse/WC-2428) - allow empty worker upload

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: todo when relevant APIs are done
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: feature is still in progress
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: not for release yet


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -26,6 +26,7 @@ import { addHyphens } from "../deployments";
 import { confirm } from "../dialogs";
 import { getMigrationsToUpload } from "../durable";
 import { UserError } from "../errors";
+import { syncAssets } from "../experimental-assets";
```
RamIdeas, line 29: Should we rename this `syncExperimentalAssets`?

It would conform to the `legacyAssets` / `experimentalAssets` naming convention and reduce confusion between the two.

When we stabilise this stuff, we can rename all experimental assets occurrences to just assets

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -588,7 +589,15 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 				})
 			: undefined;
 
-		const assets = await syncAssets(
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
RamIdeas, line 592: Does the underscore prefix not remove the need for this comment?

packages/wrangler/src/sites.ts
```diff
@@ -118,7 +118,7 @@ function pluralise(count: number) {
  * @returns a promise for an object mapping the relative paths of the assets to the key of that
  * asset in the KV namespace.
  */
-export async function syncAssets(
+export async function syncSitesAssets(
```
RamIdeas, line 121: To match naming conventions:

```suggestion
export async function syncLegacyAssets(
```

The fact this is located in the `sites.ts` module is enough to make the connection

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4422,6 +4407,231 @@ addEventListener('fetch', event => {});`
...
+					async ({ request }) => {
+						uploadRequests.push(request as StrictRequest<UploadPayloadFile[]>);
+						const body = (await request.text())
+							.split("\n")
+							.map((x) => JSON.parse(x)) as UploadPayloadFile[];
+
+						bodies.push(body);
+						expect(request.headers.get("Authorization")).toBe(
+							"Bearer <<aus-token>>"
+						);
```
RamIdeas, line 4558: I know this already exists around the codebase but I don't love this usage of `expect` inside the mock endpoints. When the test fails (it is a network error to the `fetch` call) it makes debugging incredibly difficult.

We should be writing our tests so that they fail in a way that helps us find the cause easily. (Contrived example: using `expect(string).toContain(substring)` vs expect(string.includes(substring)).toBe(true)` results in a better output from vitest that shows both strings and a diff instead of "expected false but got true")

Instead, can we store an array of these Authorization header values in the test function scope and push to it here (like `bodies`) and assert with all the other assertions at the end of the test. (Looks like we already have `uploadedRequests` so use that further down)

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4352,23 +4354,6 @@ addEventListener('fetch', event => {});`
...
-				experimental_assets: { directory: "assets" },
-			});
-			writeWorkerSource();
-			mockUploadWorkerRequest({
-				expectedMainModule: "no-op-worker.js",
-			});
-			mockSubDomainRequest();
-			await runWrangler("deploy");
-		});
-
```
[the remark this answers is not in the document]
RamIdeas: Spoke in-person and it was intentional

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+	metadata: {
+		contentType: string;
+	};
+};
+
+type UploadResponse = {
+	jwt?: string;
+};
+
+export const syncAssets = async (
```
RamIdeas, line 44: To keep the two distinct til we land it as stable:

```suggestion
export const syncExperimentalAssets = async (
```

RamIdeas (changes requested):

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: Ah I agree completely, I meant to fix this yesterday and just forgot 🤦 - will change now

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+	accountId: string | undefined,
+	scriptName: string,
+	assetDirectory: string | undefined,
+	dryRun: boolean | undefined
+): Promise<{
+	manifest: AssetManifest | undefined;
+	jwt: string | undefined;
+}> => {
+	if (assetDirectory === undefined) {
```
CarmenPopoviciu, line 53: can `directory` be undefined at this point? and if yes, what is the value of running `syncAssets` if there is technically no path to said assets?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+		return { manifest: undefined, jwt: undefined };
+	}
+	if (dryRun) {
```
CarmenPopoviciu, line 56: is this meant to do something in the future, or can we remove it?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -588,7 +589,15 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
+		const _assetManifest = await syncAssets(
+			accountId,
+			scriptName,
+			props.experimentalAssets,
```
CarmenPopoviciu, line 596: are we not passing the whole `ExperimentalAssets` object here? I'm assuming `experimental_assets.binding` will be needed at some point

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+		logger.log("(Note: doing a dry run, not uploading or deleting anything.)");
+		return { manifest: undefined, jwt: undefined };
+	}
+	assert(accountId, "Missing accountId");
+
+	const directory = path.resolve(assetDirectory);
```
CarmenPopoviciu, line 62: technically the `directory` path should already be resolved here. That's done in `getExperimentalAssetsBasePath` which should have been called somewhere in `deploy` I reckon. That was in fact the whole idea there, that `getExperimentalAssetsBasePath` gives us everything we need, so we don't have to worry about resolving paths and what not further down the stack. I wonder if we need to make that somewhat more clear 🤔 ?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+
+	// 1. generate asset manifest
```
CarmenPopoviciu, line 64: 🙏 <3

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+	const start = Date.now();
+	let assetLogCount = 0;
+	let bucketUploadCount = 0;
+	let completionJwt = "";
+
+	for (const bucket of initializeAssetsResponse.buckets) {
+		attempts = 0;
+		let gatewayErrors = 0;
+		const doUpload = async (): Promise<UploadResponse> => {
+			// Populate the payload only when actually uploading (this is limited to 3 concurrent uploads at 50 MiB per bucket meaning we'd only load in a max of ~150 MiB)
```
CarmenPopoviciu, line 97: sanity check: do we need to dbl check these upload size limitations with EWC folks?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+					{
+						method: "POST",
+						headers: {
+							"Content-Type": "application/x-ndjson",
+							Authorization: `Bearer ${initializeAssetsResponse.jwt}`,
+						},
+						body: payload.map((x) => JSON.stringify(x)).join("\n"),
+					}
+				);
+				logger.debug("result:", res);
```
CarmenPopoviciu, line 142: can we pls make this a little more helpful for our users? If they're at the point of needing debugging logs, we should make an effort to help them figure out what is going on. Not sure what the API returns here, but IMHO it would be good to know which bucket was uploaded (`bucket x out of n`), how many files it included, and perhaps smth more maeningful than `result` :)

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+				return res;
+			} catch (e) {
+				if (attempts < MAX_UPLOAD_ATTEMPTS) {
+					logger.debug("failed:", e, "retrying...");
```
CarmenPopoviciu, line 146: similar to https://github.com/cloudflare/workers-sdk/pull/6374/files#r1701809610

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+					// Exponential backoff, 1 second first time, then 2 second, then 4 second etc.
+					await new Promise((resolvePromise) =>
+						setTimeout(resolvePromise, Math.pow(2, attempts) * 1000)
+					);
+					// TODO: handle other errors e.g. jwt expired here
```
CarmenPopoviciu, line 151: TODO in this iteration?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+						// only count as a failed attempt after a few initial gateway errors
+						if (gatewayErrors >= MAX_UPLOAD_GATEWAY_ERRORS) {
+							attempts++;
+						}
+					} else {
+						attempts++;
+					}
+					return doUpload();
+				} else {
+					logger.debug("failed:", e);
```
CarmenPopoviciu, line 170: similar to https://github.com/cloudflare/workers-sdk/pull/6374/files#r1701809610

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+				}
+			)
+		);
+	}
+	await queue.onIdle();
+	if (!completionJwt) {
+		throw new FatalError("Failed to upload all files. Please try again.", 1);
+	} else if (bucketUploadCount !== initializeAssetsResponse.buckets.length) {
+		throw new FatalError(
+			"Completion signal received unexpectedly early - we cannot confirm if all files were successfully uploaded. Please try again.",
```
CarmenPopoviciu, line 200:
```suggestion
			"Upload completion signal received unexpectedly early - we cannot confirm if all files were successfully uploaded. Please try again.",
```

maybe?

packages/wrangler/src/deploy/deploy.ts
emily-shen: I don't think the binding will be needed at this point? this particular function just uploads assets so all it needs is the assets directory as far as I can tell

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+			const filestat = await stat(filepath);
+
+			if (filestat.isSymbolicLink()) {
+				return;
+			}
+
+			if (filestat.isDirectory()) {
+				manifest = await walk(filepath, manifest, startingDir);
+			} else {
+				if (counter >= MAX_ASSET_COUNT) {
```
CarmenPopoviciu, line 240: sanity check: are we using the correct `MAX_ASSET_COUNT` value here, as per spec?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+					throw new UserError(
+						`You cannot have more than ${MAX_ASSET_COUNT.toLocaleString()} files in a deployment. Ensure you have specified your build output directory correctly.`
+					);
+				}
+
+				const name = urlSafe(relativeFilepath);
+				if (filestat.size > MAX_ASSET_SIZE) {
```
CarmenPopoviciu, line 247: sanity check: are we using the correct `MAX_ASSET_SIZE` value here, as per spec?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+					throw new FatalError(
```
CarmenPopoviciu, line 248:
```suggestion
					throw new UserError(
```

packages/wrangler/src/experimental-assets.ts
emily-shen: good point - directory is undefined when experimental assets is not being used and I was handling that inside syncAssets (following the pattern of syncSitesAssets). Happy to move that bit of logic outside into deploy.ts if you think that reads better.

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+			if (filestat.isSymbolicLink()) {
+				return;
+			}
+
+			if (filestat.isDirectory()) {
+				manifest = await walk(filepath, manifest, startingDir);
+			} else {
+				if (counter >= MAX_ASSET_COUNT) {
+					throw new UserError(
+						`You cannot have more than ${MAX_ASSET_COUNT.toLocaleString()} files in a deployment. Ensure you have specified your build output directory correctly.`
```
CarmenPopoviciu, line 242: can we shape this err message more in the lines of https://github.com/cloudflare/workers-sdk/blob/main/STYLEGUIDE.md#general-error 🙏 ?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+					);
+				}
+
+				const name = urlSafe(relativeFilepath);
+				if (filestat.size > MAX_ASSET_SIZE) {
+					throw new FatalError(
+						`Max file size is ${prettyBytes(MAX_ASSET_SIZE, {
```
CarmenPopoviciu, line 249: similar to https://github.com/cloudflare/workers-sdk/pull/6374/files#r1701820626

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+		let gatewayErrors = 0;
+		const doUpload = async (): Promise<UploadResponse> => {
+			// Populate the payload only when actually uploading (this is limited to 3 concurrent uploads at 50 MiB per bucket meaning we'd only load in a max of ~150 MiB)
+			// This is so we don't run out of memory trying to upload the files.
+			const payload: UploadPayloadFile[] = await Promise.all(
+				bucket.map(async (fileHash) => {
+					const manifestEntryIndex = filteredFiles.findIndex(
+						(file) => file[1].hash === fileHash
+					);
+					// not sure if this is really a user error - what should i use instead if so?
```
CarmenPopoviciu, line 104: `FatalError` is perfect here

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+	);
+	return manifest;
+};
+
+/**
+ * Convert a filePath to be safe to use as a relative URL.
+ *
+ * Primarily this involves converting Windows backslashes to forward slashes.
+ */
+function urlSafe(filePath: string): string {
```
CarmenPopoviciu, line 273: are any of these fns common between Pages and Workers. Anything we can extract into a shared thing to avoid code duplication?

packages/wrangler/src/deploy/deploy.ts
CarmenPopoviciu: +1

packages/wrangler/src/deploy/deploy.ts
CarmenPopoviciu: right. What I meant was, shouldn't `props.experimentalAssets` be of type `ExperimentalAssets`? in which case we'd need to pass `props.experimentalAssets.directory` here

emily-shen: I'm fairly sure the binding is not needed at the point, but also just realised the ExperimentalAssets object was not properly passed in so will fix that

packages/wrangler/src/__tests__/deploy.test.ts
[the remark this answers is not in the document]
emily-shen: Yes initially, but have brought it back upon reflection (and now it actually tests what it says it does...)

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4422,6 +4407,231 @@ addEventListener('fetch', event => {});`
...
+					}
+				)
+			);
+			mockSubDomainRequest();
+			mockUploadWorkerRequest({
+				expectedMainModule: "no-op-worker.js",
+			});
+			await runWrangler("deploy");
+			expect(uploadRequests.length).toBe(mockBuckets.length);
+			expect(bodies.map((b) => b.length).sort()).toEqual([1, 1, 1, 2]);
```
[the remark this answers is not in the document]
emily-shen: good point, removed

packages/wrangler/src/experimental-assets.ts
emily-shen: good point, just an oversight on my part :)

packages/wrangler/src/experimental-assets.ts
emily-shen: Yep the spec has it at 20000 per manifest, same with pages so I just pulled it from there. Will set the constants in the file so they're not dependent on pages though

packages/wrangler/src/experimental-assets.ts
emily-shen: the urlSafe helper func is from workers sites, will import instead. While the bucket upload is pretty much ripped from pages, its sufficiently different that it can't be shared I think.

packages/wrangler/src/experimental-assets.ts
emily-shen: Removing this one as I realised debug mode shows them the api request + result anyway. Going to log successful bucket upload instead :)

RamIdeas (approved):

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4422,6 +4442,243 @@ addEventListener('fetch', event => {});`
...
+					},
+					"/boop/file-2.txt": {
+						hash: "7574a8cd3094a050388ac9663af1c1d6",
+						size: 17,
+					},
+				},
+			});
+		});
+
+		it("should upload assets in the requested buckets", async () => {
```
GregBrimble, line 4547: brilliant and comprehensive test 😃

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4422,6 +4442,243 @@ addEventListener('fetch', event => {});`
...
+			const assets = [
+				{ filePath: "file-1.txt", content: "Content of file-1" },
+				{ filePath: "boop/file-2.txt", content: "Content of file-2" },
+				{ filePath: "boop/file-3.txt", content: "Content of file-3" },
+				{ filePath: "file-4.txt", content: "Content of file-4" },
+				{ filePath: "beep/file-5.txt", content: "Content of file-5" },
+				{ filePath: "beep/file-6.txt", content: "Content of file-1" },
```
GregBrimble, line 4554: Realizing that we haven't yet defined behavior for when there are wonky characters in file names yet: spaces, question marks, etc.

My gut says we should essentially only support things which equal themselves when we `decodeURIComponent(new URL(`http://fakehost/${filePath}`).pathname))` and error for anything else. Can we add that check in to start with, and then revisit once we've got more of ASW2 in place?

petebacondarwin: Please can we capture these rules in the SPEC.

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4352,21 +4353,40 @@ addEventListener('fetch', event => {});`
...
 			writeAssets(assets);
-			writeWranglerToml({
-				experimental_assets: { directory: "assets" },
-			});
-			writeWorkerSource();
+			// note that wrangler.toml and worker source are not created
+			// pass expected formBody in
 			mockUploadWorkerRequest({
-				expectedMainModule: "no-op-assets-worker.js",
+				expectedExperimentalAssetsForm: [
```
GregBrimble, line 4365: Just a nit about naming/shapes:

This looks like a bit of a weird shape to me, but that's maybe just me not being familiar with other test util things here. `assets` is always in `metadata` and the key is always `assets`, so I'd either expect the `expectedExperimentalAssetsForm` thing to have that built in, _or_ for this to be a more generic utility that helps you validate the contents of `metadata` (e.g. `expectedMetadata`). It's not super clear in it's current state if the test would pass if there was other unexpected stuff in metadata, for example.

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4422,6 +4442,243 @@ addEventListener('fetch', event => {});`
 				)
 			);
 		});
+
+		it("should upload an asset manifest of the files in the directory specified by --experimental-assets", async () => {
```
GregBrimble, line 4446: it's not super clear how this test and the previous one differ. Feels like they could be combined into a singe test which validates that both:
- `wrangler deploy` works when there is no worker source
- the manifest is uploaded

To me, they don't mean much on their own.

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4422,6 +4442,243 @@ addEventListener('fetch', event => {});`
...
+					},
+					"/boop/file-2.txt": {
+						hash: "7574a8cd3094a050388ac9663af1c1d6",
+						size: 17,
+					},
+				},
+			});
+		});
+
+		it("should upload an asset manifest of the files in the directory specified by [experimental_assets] config", async () => {
```
GregBrimble, line 4496: can I get one more test which validates behavior when there is both a CLI arg and the assets being defined in the config file? It would be good to confirm that the arg wins, and will set us up for answering questions about how we merge(?) config in the future e.g. `--assets=./foo` and `assets = { directory = "./bar", binding = "BAZ" }`.

GregBrimble, line 4496: And can I also get another which proves out the relative resolution? That the CLI arg is relative to cwd, and that the config is relative to the config file location?

packages/wrangler/src/__tests__/helpers/mock-upload-worker.ts
```diff
@@ -57,7 +58,10 @@ export function mockUploadWorkerRequest(
 		const metadata = JSON.parse(
 			await toString(formBody.get("metadata"))
 		) as WorkerMetadata;
-		if (expectedType === "esm") {
+		if (expectedExperimentalAssetsForm) {
```
GregBrimble, line 61: Yeah, I think this is also a bit of a symptom of my earlier comment. To me, it doesn't feel like this belongs to this if/else. The if/else here validates the two possible ways of specifying an entrypoint to the Worker. Independently of that, we may or may not have presence of an `assets` key in the metadata that we want to validate.

packages/wrangler/src/deployment-bundle/entry.ts
```diff
@@ -112,6 +113,9 @@ export async function getEntry(
 		directory,
 		format,
 		moduleRoot: args.moduleRoot ?? config.base_dir ?? path.dirname(file),
+		staticAssetsOnly:
+			!!(args.experimentalAssets || config.experimental_assets) &&
+			!(args.script && config.main),
```
GregBrimble, line 118: In `packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts`, you have a check for if `staticAssetsOnly`, and if yes, only ever send the assets piece. Fair enough!

But here, you're also saying that `staticAssetsOnly` is not true if any user worker script is set.

I think you either want to loosen this to say that if there is experimentalAssets/experimental_assets is defined, then that just always means `staticAssetsOnly` (forget about the presence of script/main).

```suggestion
```

*OR* you want to just always try attaching the `assets` piece in metadata and let the API fail if it fails. We'll get there eventually, but I have a slight personal preference of the former, just to reduce the work you need to do and keep it consistent with all the tests you've already written :)

GregBrimble: On that note, might be nice to have another test which *does* directly reference a user Worker (e.g. with `main` config) and to see how that collapses.

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+		return;
+	}
+	assert(accountId, "Missing accountId");
+
+	// 1. generate asset manifest
+	logger.info("🌀 Building list of assets...");
+	const manifest = await walk(assetDirectory, {});
+
+	// 2. fetch buckets w/ hashes
+	logger.info("🌀 Uploading asset manifest...");
```
GregBrimble, line 67: Users don't need to know about manifests :)

```suggestion
	logger.info("🌀 Starting asset upload...");
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+		`/accounts/${accountId}/workers/scripts/${scriptName}/assets-upload-session`,
+		{
+			headers: { "Content-Type": "application/json" },
+			method: "POST",
+			body: JSON.stringify({ manifest: manifest }),
+		}
+	);
+
+	// if nothing to upload, return
+	if (initializeAssetsResponse.buckets.flat().length === 0) {
```
GregBrimble, line 78: _Ideally_ we'd be looking for a 201 here to decide that we can proceed, but I see that's actually tucked away behind all these fetcher utils, so fair enough :)

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+		logger.info(`✨ Success! (No files to upload)`);
+		return initializeAssetsResponse.jwt;
```
GregBrimble, line 80: Should we verify that there is indeed a `jwt` here first though? Throw an error if not, rather than blindly proceeding without the assets bit.

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+			const manifestEntryIndex = manifestLookup.findIndex(
+				(file) => file[1].hash === fileHash
+			);
+			if (manifestEntryIndex === -1) {
+				throw new FatalError(
+					`A file was requested that does not appear to exist.`,
+					1
+				);
+			}
+			const manifestEntry = manifestLookup.splice(manifestEntryIndex, 1)[0];
```
GregBrimble, line 102:
```suggestion
			const manifestEntry = manifestLookup.find(
				(file) => file[1].hash === fileHash
			);
			if (manifestEntry === undefined) {
				throw new FatalError(
					`A file was requested that does not appear to exist.`,
					1
				);
			}
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+			const payload: UploadPayloadFile[] = await Promise.all(
+				bucketSkeleton.map(async (manifestEntry) => {
+					const absFilePath = path.join(assetDirectory, manifestEntry[0]);
+					return {
+						name: manifestEntry[0],
+						hash: manifestEntry[1].hash,
+						contents: (await readFile(absFilePath)).toString("base64"),
+						metadata: {
+							contentType: getType(absFilePath) || "application/octet-stream",
+						},
```
GregBrimble, line 131: "base64": true,
    "key": manifestEntry[1].hash,
    "metadata": {
      "contentType": getType(absFilePath) || "application/octet-stream",
    },
    "value": (await readFile(absFilePath)).toString("base64"),

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+						if (gatewayErrors >= MAX_UPLOAD_GATEWAY_ERRORS) {
+							attempts++;
+						}
+					} else {
+						attempts++;
+					}
+					return doUpload();
+				} else if (isJwtExpired(initializeAssetsResponse.jwt)) {
+					throw new FatalError(
+						`Asset upload took too long on bucket ${bucketIndex + 1}/${initializeAssetsResponse.buckets.length}. Please try again.`
```
GregBrimble, line 181: Idk, maybe worth trying to say something like this? Take it, edit it, or leave it. I'm just trying to say that they don't need to worry about having to start from scratch again to indeed encourage trying again. If that just confuses the message, feel free to not bother.

```suggestion
						`Asset upload took too long on bucket ${bucketIndex + 1}/${initializeAssetsResponse.buckets.length}. Please try again — assets already uploaded have been saved, so the next attempt will resume from this point.`
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+		logger.error(error.message);
+		throw error;
+	});
+	await queue.onIdle();
+
+	// if queue finishes without receiving JWT from asset upload service (AUS)
+	// AUS only returns this in the final bucket upload response
+	if (!completionJwt) {
+		throw new FatalError(
+			"Failed to receive completion signal for file upload. Please try again.",
```
GregBrimble, line 206: Again, users don't need to know about "completion signals" — this is something we can word a little better for them.

```suggestion
			"Failed to complete asset upload. Please try again.",
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+			1
+		);
+	} else if (bucketUploadCount !== initializeAssetsResponse.buckets.length) {
+		throw new FatalError(
+			"Upload completion signal received unexpectedly early - we cannot confirm if all files were successfully uploaded. Please try again.",
+			1
+		);
+	}
```
GregBrimble, line 214: I think because you iterate over the full set of buckets and just update the value of completionJwt as you go rather than early exiting, this will never actually fire. That's fine :) Just flagging in case you didn't realize.
```suggestion
	} else if (bucketUploadCount !== initializeAssetsResponse.buckets.length) {
		throw new FatalError(
			"Asset uploading finished unexpectedly early — we cannot confirm if all files were successfully uploaded. Please try again.",
			1
		);
	}
```

packages/wrangler/src/experimental-assets.ts
GregBrimble: Oh, that's a good point. I had hoped to move these values to EWC so we could change them on the fly for different customers or whatever. But I guess we'll always need at least some sort of sensible upper-bound in Wrangler so it doesn't crawl the world for a misconfigured app. Stick with what you've done here for now. We can address this in the future if and when we ever raise those limits :)

packages/wrangler/src/pages/upload.tsx
```diff
@@ -353,11 +353,13 @@ export const upload = async (
 };
 
 // Decode and check that the current JWT has not expired
-function isJwtExpired(token: string): boolean | undefined {
+export const isJwtExpired = (token: string): boolean | undefined => {
```
GregBrimble, line 356: ideally this gets pulled out into some common place, but nbd.

packages/wrangler/src/sites.ts
```diff
@@ -434,7 +434,7 @@ function validateAssetKey(assetKey: string) {
  *
  * Primarily this involves converting Windows backslashes to forward slashes.
  */
-function urlSafe(filePath: string): string {
+export function urlSafe(filePath: string): string {
```
GregBrimble, line 437: same here, but yeah, nbd

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,301 @@
...
+			} else {
+				if (counter >= MAX_ASSET_COUNT) {
+					throw new UserError(
+						`Max asset count exceeded: ${counter.toLocaleString()} files found.
+						You cannot have more than ${MAX_ASSET_COUNT.toLocaleString()} files in a deployment.
+						Ensure you have specified your build output directory correctly (currently set as ${startingDir}).`
+					);
+				}
+
+				const name = urlSafe(relativeFilepath);
```
GregBrimble, line 257: ah ha! This kinda already starts to touch on [my earlier comment](https://github.com/cloudflare/workers-sdk/pull/6374/files#r1705794267). Let's handle this in a chat with Carmen/me/you and figure out exactly what we want to do here.

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: Good point, we should definitely validate this - I might leave this out of this PR for now and add a ticket to milestone 2 to follow up on this when we come to a decision.

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: Yeah I agree its a bit dodgy - a generic expectedMetadata seems like a better idea. will add that in instead

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: ah whoops the first no worker source tests is redundant, removed it

packages/wrangler/src/__tests__/helpers/mock-upload-worker.ts
emily-shen: added a third 'none' entry type for the static assets only case

packages/wrangler/src/experimental-assets.ts
emily-shen: whoops that was silly, thanks

packages/wrangler/src/deployment-bundle/entry.ts
emily-shen: I might be misunderstanding your comment, but the idea _was_ to have `staticAssetsOnly` as a flag for when `experimental-assets` was used without a user worker. This was to make it easier to pass through to `create-worker-upload-form` to make sure the `no-op-asset-worker` wasn't uploaded. Admittedly this was a bit of a janky temp solution, since I think this wont be necessary once we can get rid of the `no-op-asset-worker` that `wrangler dev` needs. 

Upon reflection I think `Entry` does seem an inappropriate place for this, so I've moved it to the `experimentalAssets` options that gets passed through to `create-worker-upload-form`,

packages/wrangler/src/pages/upload.tsx
CarmenPopoviciu: +1

packages/wrangler/src/sites.ts
CarmenPopoviciu: indeed. Can we pls keep track of this follow-up somewhere, for both this and prev comment 🙏 ?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+
+	// if nothing to upload, return
+	if (initializeAssetsResponse.buckets.flat().length === 0) {
+		if (!initializeAssetsResponse.jwt) {
+			throw new FatalError(
+				"Could not find assets information to attach to deployment. Please try again.",
+				1
+			);
+		}
+		logger.info(`✨ Success! (No files to upload)`);
```
CarmenPopoviciu, line 85:
```suggestion
		logger.info(`No files to upload. Proceeding with deployment`);
```
or smth along the lines of 👆 . IMHO the `✨ Success!` copy is more relevant when files were actually uploaded. This made me wonder what's the sucess in this case, if no files need be uploaded

packages/wrangler/src/experimental-assets.ts
CarmenPopoviciu: I see this was addressed here: https://github.com/cloudflare/workers-sdk/pull/6374/files#diff-99329bc31323333c32855181ed3b738e17d682a5f16c825ec49be45d30e54082R79 but shouldn't that check be performed independent of whether `initializeAssetsResponse.buckets.flat().length === 0`?

so basically
```
if (!initializeAssetsResponse.jwt) {
	throw new FatalError(
		"Could not find assets information to attach to deployment. Please try again.",
		1
	);
}

if (initializeAssetsResponse.buckets.flat().length === 0) {
	logger.info(`✨ Success! (No files to upload)`);
	return initializeAssetsResponse.jwt;
}
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+		return initializeAssetsResponse.jwt;
+	}
+
+	// 3. fill buckets and upload assets
+	const numberFilesToUpload = initializeAssetsResponse.buckets.flat().length;
+	logger.info(`🌀 Uploading ${numberFilesToUpload} file(s)...`);
```
CarmenPopoviciu, line 91: since we are already inside the `logger.info("🌀 Starting asset upload...");` scope, this message feels a bit duplicative to me. How about `logger.info(`🌀 Found ${numberFilesToUpload} file${numberFilesToUpload > 1 ? "s" : ""} to upload. Continuing...`);`

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+
+	// Create the buckets outside of doUpload so we can retry without losing track of potential duplicate files
+	// But don't add the actual content until uploading so we don't run out of memory
+	const manifestLookup = Object.entries(manifest);
+	let assetLogCount = 0;
+	const bucketSkeletons = initializeAssetsResponse.buckets.map((bucket) => {
```
CarmenPopoviciu, line 97: can we specify the type for `bucketSkeletons` here as opposed to have it inferred?

CarmenPopoviciu: better yet, maybe have a dedicated type if cleaner

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+				(file) => file[1].hash === fileHash
+			);
+			if (manifestEntry === undefined) {
+				throw new FatalError(
+					`A file was requested that does not appear to exist.`,
+					1
+				);
+			}
+			// just logging file uploads at the moment...
+			// unsure how to log deletion vs unchanged file ignored/if we want to log this
```
CarmenPopoviciu, line 109: smth to figure out now or later? if for later, can we pls track this somewhere :pray:?

packages/wrangler/src/experimental-assets.ts
CarmenPopoviciu: also, can we name these simply `buckets` or `assetBuckets`? `skeletons` doesn't tell me much?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+	const manifestLookup = Object.entries(manifest);
+	let assetLogCount = 0;
+	const bucketSkeletons = initializeAssetsResponse.buckets.map((bucket) => {
+		return bucket.map((fileHash) => {
+			const manifestEntry = manifestLookup.find(
+				(file) => file[1].hash === fileHash
+			);
+			if (manifestEntry === undefined) {
+				throw new FatalError(
+					`A file was requested that does not appear to exist.`,
```
CarmenPopoviciu, line 104: what's the action users can take here? Also, is it worth specifying the fileHash that was requested but we couldn't find?

```suggestion
					`Could not find file with fileHash "${fileHahs}".\n` +
          `We attempted to upload file with hash "${fileHash}" but could not find it in the file system. Please <insert-action-here>`,
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+						if (gatewayErrors >= MAX_UPLOAD_GATEWAY_ERRORS) {
+							attempts++;
+						}
+					} else {
+						attempts++;
+					}
+					return doUpload();
+				} else if (isJwtExpired(initializeAssetsResponse.jwt)) {
+					throw new FatalError(
+						`Asset upload took too long on bucket ${bucketIndex + 1}/${initializeAssetsResponse.buckets.length}. Please try again.
```
CarmenPopoviciu, line 185: I noticed that we're using `assets` / `files` interchangeably in our copies. Perhaps let's settle on one for consistency purposes?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+						// only count as a failed attempt after a few initial gateway errors
+						if (gatewayErrors >= MAX_UPLOAD_GATEWAY_ERRORS) {
+							attempts++;
+						}
+					} else {
+						attempts++;
+					}
+					return doUpload();
+				} else if (isJwtExpired(initializeAssetsResponse.jwt)) {
+					throw new FatalError(
```
CarmenPopoviciu, line 184:
```suggestion
					throw new FatalError(
						`Upload took too long.\n` +
            `Asset upload took too long on bucket ${bucketIndex + 1}/${initializeAssetsResponse.buckets.length}. Please try again.
						Assets already uploaded have been saved, so the next attempt will automatically resume from this point.`
          );
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+			1
+		);
+	}
+
+	const uploadMs = Date.now() - start;
+	const skipped = Object.keys(manifest).length - numberFilesToUpload;
+	const skippedMessage = skipped > 0 ? `(${skipped} already uploaded) ` : "";
+
+	logger.log(
+		`✨ Success! Uploaded ${numberFilesToUpload} file(s) ${skippedMessage}${formatTime(uploadMs)}\n`
```
CarmenPopoviciu, line 220:
```suggestion
		`✨ Success! Uploaded ${numberFilesToUpload} file${numberFilesToUpload > 1 ? "s" : ""} ${skippedMessage}${formatTime(uploadMs)}\n`
```

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+			}
+
+			if (filestat.isDirectory()) {
+				manifest = await walk(filepath, manifest, startingDir);
+			} else {
+				if (counter >= MAX_ASSET_COUNT) {
+					throw new UserError(
+						`Max asset count exceeded: ${counter.toLocaleString()} files found.
+						You cannot have more than ${MAX_ASSET_COUNT.toLocaleString()} files in a deployment.
+						Ensure you have specified your build output directory correctly (currently set as ${startingDir}).`
```
CarmenPopoviciu, line 252: > `Ensure you have specified your build output directory correctly (currently set as ${startingDir}).`

the action we are recommending here feels a bit misleading to me. We're pointing the problem at the assets directory, when in fact it's the number of files that's the problem. I expect in most cases the assets directory specififed by users will be the correct one, and the action they need to take is to restrict the number of assets in said directory. To that effect, I wonder if a better action here would be


```suggestion
					throw new UserError(
						`Maximum number of assets exceeded.
						Cloudflare Workers supports up to ${MAX_ASSET_COUNT.toLocaleString()} assets in a deployment. We found ${counter.toLocaleString()} files in the specified assets directory "${startingDir}".
						Ensure your assets directory contains a maximum of ${MAX_ASSET_COUNT.toLocaleString()} files, or that you have specified your assets directory correctly.`
```

packages/wrangler/src/experimental-assets.ts
CarmenPopoviciu: have we figured this out? and if not, can this PR be merged without having those details in place?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -1,8 +1,299 @@
...
+					);
+				}
+
+				const name = urlSafe(relativeFilepath);
+				if (filestat.size > MAX_ASSET_SIZE) {
+					throw new UserError(
```
CarmenPopoviciu, line 258:
```suggestion
					throw new UserError(
            `Asset too large.\n` + 
            `Cloudflare Workers supports assets with sizes up to ${prettyBytes(MAX_ASSET_SIZE, {
							binary: true,
						})}. We found file ${filepath} with a size of ${prettyBytes(filestat.size, {
							binary: true,
						})}.\n`
            `Ensure all assets in your assets directory "${startDir}" conform with the Workers maximum size requirement.`
					);
```

packages/wrangler/src/experimental-assets.ts
emily-shen: I think it should be okay to leave for now, and I've added a [ticket](https://jira.cfdata.org/browse/WC-2455) to keep track.

packages/wrangler/src/experimental-assets.ts
emily-shen: If the asset manifest shouldn't user facing info as per Greg's earlier comment, I'm not sure the hashes should be either. Also not sure what we can really suggest to solve this. Made a [followup ticket](https://jira.cfdata.org/browse/RM-19435) to track all the suggestions to do with copy so I can make sure everyone approves all the wording.

packages/wrangler/src/experimental-assets.ts
emily-shen: [followup ticket](https://jira.cfdata.org/browse/RM-19435)

packages/wrangler/src/experimental-assets.ts
emily-shen: It doesn't make it that far down - it returns immediately.

emily-shen:
As all the logging/error messages wont be public-facing yet, I'd prefer to leave it for another PR tracked by [this ticket](https://jira.cfdata.org/browse/WC-2464). Will leave those comments unresolved.

packages/wrangler/src/deployment-bundle/entry.ts
GregBrimble: Yeah, tad janky but you've improved it a bit and like you say, it's temporary anyway and we'll be in amongst here plenty in future PRs for user worker pieces when they're etc., so won't block!

packages/wrangler/src/experimental-assets.ts
GregBrimble: yeah, I think you're good with this in it's current state. You do a similar check later on too :)

GregBrimble (approved):

packages/wrangler/src/experimental-assets.ts
emily-shen: tracking this in [this ticket](https://jira.cfdata.org/browse/WC-2464)

packages/wrangler/src/experimental-assets.ts
CarmenPopoviciu: SGTM

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -678,6 +690,10 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			placement,
 			tail_consumers: config.tail_consumers,
 			limits: config.limits,
+			experimental_assets: {
```
petebacondarwin, line 693: Should this whole property be undefined if there are no assets?
This could be conditional upon whether `experimentalAssetsJwt` is defined or perhaps on whether `props.experimentalAssets` is defined?

packages/wrangler/src/deployment-bundle/entry.ts
petebacondarwin: Is there still a no-op asset worker passed to `wrangler dev` as an entry?
It feels like this is no longer needed and we could just say that the entry is undefined when there are only assets?

packages/wrangler/src/deployment-bundle/worker.ts
```diff
@@ -336,6 +336,10 @@ export interface CfWorkerInit {
 	tail_consumers: CfTailConsumer[] | undefined;
 	limits: CfUserLimits | undefined;
 	annotations?: Record<string, string | undefined>;
+	experimental_assets?: {
+		jwt: string | undefined;
+		staticAssetsOnly: boolean | undefined;
```
petebacondarwin, line 341: I probably missed the discussion but could we not have `main` be optionally undefined. So if there is no `main` but there is `experimental_assets` then one can infer that this is static assets only?

There would be an assertion that at least on of `main` and `experimental_assets` is defined.

packages/wrangler/src/deployment-bundle/entry.ts
petebacondarwin: Oh I see that there is a `no-op-asset-worker` in the templates. Do we really need this? Isn't the work that Carmen has done avoid the need for this?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -588,7 +592,15 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
 			: undefined;
 
-		const assets = await syncAssets(
+		// Upload assets if experimental assets is being used
+		const experimentalAssetsJwt = await syncExperimentalAssets(
+			accountId,
+			scriptName,
+			props.experimentalAssets?.directory,
+			props.dryRun
+		);
```
petebacondarwin, line 601: Can we explicitly gate this behind the experimental_assets flag, to avoid running any new code at all when the flag is not enabled.

packages/wrangler/src/deploy/deploy.ts
petebacondarwin: Moreover, to keep this new stuff as segregated from the normal modes, we should try to avoid adding any new properties to this object if the new flag is not enabled.

petebacondarwin (commented):
I think there is some additional refactoring we can do to remove the no-op-asset-worker and simplify the config somewhat. But this looks fine for now.

packages/wrangler/src/deploy/deploy.ts
emily-shen: Yes it should be, was a bit of a temp solution to let me only touch the `deploy` code path in this PR because it was getting a bit large. Will be doing a fast followup today for `versions upload` and other bits that should clean this up.

packages/wrangler/src/deployment-bundle/worker.ts
emily-shen: This is the end goal but waiting for the no-op-asset-worker to be removed first :)

PR #6379, opened 2024-07-30
Clearer error message when trying to use Workers Sites with `wrangler versions upload`

RamIdeas:

Workers Sites is a legacy feature which we do not support in Gradual Deployments. This change makes the user aware.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:


packages/wrangler/src/deploy/utils.ts
```diff
@@ -0,0 +1,44 @@
+import { existsSync } from "node:fs";
```
CarmenPopoviciu, line 1: I'd rather this lived [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/experimental-assets.ts) as it is not deploy specific. We do smth similar in `[dev](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/dev.tsx#L807-L833)` as well. Careful tho with https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/dev.tsx#L832 which `dev` needs, but is not relevant to `deploy`

packages/wrangler/src/versions/upload.ts
```diff
@@ -50,6 +50,7 @@ type Props = {
 	env: string | undefined;
 	compatibilityDate: string | undefined;
 	compatibilityFlags: string[] | undefined;
+	experimentalAssets: string | undefined;
```
CarmenPopoviciu, line 53:
```suggestion
	experimentalAssets: ExperimentalAssets | undefined;
```
since `wrangler.toml` supports
```
[experimental_assets]
directory = "dist"
binding = "binding_name"
```

RamIdeas: I copied this from `https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/deploy/deploy.ts#L79` and it's passed just the `.directory` property on https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/deploy/index.ts#L372

Happy to change it in both places if you'd prefer?

packages/wrangler/src/deploy/utils.ts
RamIdeas: Good catch – thanks for letting me know about dev.tsx!

Moved in https://github.com/cloudflare/workers-sdk/pull/6379/commits/d16520472ba5fbf2805e654308ed16cac79c5aac

packages/wrangler/src/versions/index.ts
```diff
@@ -191,6 +215,21 @@ export async function versionsUploadHandler(
...
+	if (args.site || config.site) {
+		throw new UserError(
+			"Workers Sites are not supported in Gradual Deployments."
+		);
+	}
+	if (args.legacyAssets || config.legacy_assets) {
+		throw new UserError(
+			"Legacy Assets are not supported in Gradual Deployments."
+		);
+	}
```
RamIdeas, line 229: cc @irvinebroque @tanushree-sharma will this copy suffice?

packages/wrangler/src/deploy/utils.ts
CarmenPopoviciu: thank you Rahul!

packages/wrangler/src/versions/upload.ts
petebacondarwin: Since this is the new experimental assets, there is nothing to be done with the binding name yet. So I think it is OK to match deploy.ts here and @emily-shen can update it when she implements the new assets support.

petebacondarwin (approved):

packages/wrangler/src/versions/upload.ts
CarmenPopoviciu: ah, I see. I think `string` was temporarily used there as a placeholder, until the type of `experimental _assets` was fully settled on (there's been some back and forth around it). We've definitely settled on `ExperimentalAssets` since then. The reason why `deploy` only gets `experimentalAssets.directory` passed in, is probably because of a similar reason. The code around experimental assets is very much in flux atm, which is why you're seeing these inconsistencies & rough edges. I've already worked through these for `dev` and Emily is currently busy with the `deploy` bit, so it'll all be smoothed out soon. Fantastic catch tho!

So to tl;dr:
- use the `ExperimentalAssets` type for all experimental assets things (the **ONLY** exception to that is `getEntry()`, which takes in `experimentalAssets: string` see my comment [here](https://github.com/cloudflare/workers-sdk/pull/6370/files#diff-524c0b6c44e4e63168f1849abe9420fbb7c44f64c3a6166f44515fdd00b2118dR207))
- pass the entire object to `deploy` (Emily is taking care of this)/`versions` etc

packages/wrangler/src/versions/index.ts
tanushree-sharma: I would modify to "Workers Sites does not support uploading versions through "wrangler versions upload". You must use [wrangler deploy](https://developers.cloudflare.com/workers/wrangler/commands/#deploy) instead."

tanushree-sharma: Gah sorry, commenting after you've merged.

RamIdeas: Thanks for the copy. I'll make another PR to update

PR #6383, opened 2024-07-31
feat: support outputing ND-JSON files via an environment variable

petebacondarwin:

Adds a feature needed for the Workers CI/CD team.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because: no e2e changes
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: internal usage for now


packages/wrangler/src/output.ts
```diff
@@ -0,0 +1,95 @@
...
+			outputFileDirectoryFromEnv,
+			`wrangler-output-${date}-${randomBytes(3).toString("hex")}.json`
+		);
+	}
+
+	return null;
+}
+
+interface OutputEntryBase<T extends string> {
+	version: number;
```
jrf0110, line 61: Might want to constrain the type here to `version: 1` to prevent any accidental mis-versioning

petebacondarwin: I was just thinking that to myself!

petebacondarwin: a981459e670155e2d623f997f44f5264cc81c76f

petebacondarwin: I set it at the individual entry level rather than in the base, since they may want to evolve independently.

packages/wrangler/src/__tests__/output.test.ts
```diff
@@ -0,0 +1,229 @@
...
+
+expect.extend({
+	toContainEntries(received: string, expected: OutputEntry[]) {
+		const outputEntries = received
+			.trim()
+			.split("\n")
+			.map((line) => JSON.parse(line))
+			.map(({ timestamp, ...entry }) => {
+				expect(typeof timestamp).toBe("string");
+				console.log(entry);
```
RamIdeas, line 210: Has this been left intentionally?

packages/wrangler/src/__tests__/output.test.ts
```diff
@@ -0,0 +1,229 @@
...
+});
+
+expect.extend({
+	toContainEntries(received: string, expected: OutputEntry[]) {
+		const outputEntries = received
+			.trim()
+			.split("\n")
+			.map((line) => JSON.parse(line))
+			.map(({ timestamp, ...entry }) => {
+				expect(typeof timestamp).toBe("string");
```
RamIdeas, line 209: I think this makes this utility too specific to the ndjson format for this use-case.

Instead, you can add `timestamp: expect.any(String)` to each OutputEntry which would avoid hiding the entire shape of the object from anyone reading these tests

packages/wrangler/src/deploy/index.ts
```diff
@@ -361,6 +363,15 @@ export async function deployHandler(
 		experimentalVersions: args.experimentalVersions,
 	});
 
+	if (deploymentId) {
+		writeOutput({
+			type: "deployment",
+			version: 1,
+			worker_id: name,
```
RamIdeas, line 370: 👍

packages/wrangler/src/__tests__/output.test.ts
petebacondarwin: Oops!

packages/wrangler/src/__tests__/output.test.ts
petebacondarwin: Yeah I used exactly `expect.any(String)` originally and it didn't seem to work.
I'll check again in case I was doing something wrong.

packages/wrangler/src/environment-variables/factory.ts
```diff
@@ -17,7 +17,9 @@ type VariableNames =
 	| "WRANGLER_LOG_SANITIZE"
 	| "WRANGLER_REVOKE_URL"
 	| "WRANGLER_SEND_METRICS"
-	| "WRANGLER_TOKEN_URL";
+	| "WRANGLER_TOKEN_URL"
+	| "WRANGLER_OUTPUT_FILE_DIRECTORY"
+	| "WRANGLER_OUTPUT_FILE_PATH";
```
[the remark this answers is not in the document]
petebacondarwin: That feels somewhat magical? The surface is basically the same it is just that the user needs to know that specifying `.json` on the end makes a difference to the behaviour. What if they decide they want to use `.ndjson` or something instead?

packages/wrangler/src/index.ts
```diff
@@ -253,6 +254,15 @@ export function createCLIParser(argv: string[]) {
...
 				}
 			}
+
+			// Write a session entry to the output file (if there is one).
+			writeOutput({
+				type: "wrangler-session",
+				version: 1,
+				wrangler_version: wranglerVersion,
+				command_line_args: argv,
+			});
```
RamIdeas, line 264: Great default entry with args for context! 👏 

Can you also add the log file path as well

packages/wrangler/src/versions/upload.ts
```diff
@@ -108,9 +108,13 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
 	return false;
 }
 
-export default async function versionsUpload(props: Props): Promise<void> {
+export default async function versionsUpload(
+	props: Props
+): Promise<{ versionId: string | null }> {
```
RamIdeas, line 113: Is this used anywhere? Should there also be an output entry for version upload? in versions/index.ts

packages/wrangler/src/deploy/index.ts
```diff
@@ -361,6 +363,15 @@ export async function deployHandler(
...
+			deployment_id: deploymentId,
+		});
+	}
```
RamIdeas, line 373: I think you intended to have something like this for versions upload in versions/index.ts (which matches this deploy/deploy.ts + deploy/index.ts file structure)

packages/wrangler/src/__tests__/output.test.ts
```diff
@@ -0,0 +1,229 @@
...
+				WRANGLER_OUTPUT_FILE_DIRECTORY: ".",
+			};
+			writeOutput({
+				type: "wrangler-session",
+				version: 1,
+				wrangler_version: "0.0.0.0",
+				command_line_args: ["--help"],
+			});
+			const outputFile = readFileSync(WRANGLER_OUTPUT_FILE_PATH, "utf8");
+			expect(outputFile).toContainEntries([
```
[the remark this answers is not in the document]
RamIdeas: I think custom matchers are great for common use-cases to a whole codebase and should be introduced globally so they can be defined clearly, learnt and relied upon. Implementing them as one-offs per suite I think is obfuscating

packages/wrangler/src/__tests__/output.test.ts
petebacondarwin: I was indeed doing it wrong. Thanks for the nudge!

packages/wrangler/src/__tests__/output.test.ts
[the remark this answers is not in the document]
petebacondarwin: It just a helper function to avoid duplicating the same logic throughout and keeping the tests themselves nice and readable.

It is only used in this file (where it is defined) and it makes no sense outside of this file. It is nicer than using a vanilla function because Vitest treats it specially, for example it will format the errors nicely and do diffing of the actual and expected results.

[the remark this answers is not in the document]
petebacondarwin: If you Cmd-Click on the function it takes you to the type definition below, which is exactly next to the implementation. So I don't think it is too hard to understand.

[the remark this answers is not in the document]
petebacondarwin: Somewhat ironically, when you create global custom matchers like this it is common to put the type in a `.d.ts` file - e.g. https://github.com/cloudflare/workers-sdk/blob/f6300a71adb83e5f92cd9c9ddd56a85132e03f98/packages/create-cloudflare/vitest.d.ts

In this case it is actually much harder to find the original implementation of the matcher. For example the `toExist()` matcher implementation is over here: https://github.com/cloudflare/workers-sdk/blob/cc335ee1c1dbd58009714c4b49e23e9dcf121d1e/packages/create-cloudflare/e2e-tests/setup.ts

packages/wrangler/src/versions/upload.ts
petebacondarwin: Added

packages/wrangler/src/__tests__/output.test.ts
petebacondarwin: Removed: https://github.com/cloudflare/workers-sdk/pull/6383/commits/08d006463afca5c3d68ae93b76e0da3aba58375c

packages/wrangler/src/__tests__/output.test.ts
petebacondarwin: Fixed: https://github.com/cloudflare/workers-sdk/pull/6383/commits/08d006463afca5c3d68ae93b76e0da3aba58375c

packages/wrangler/src/index.ts
petebacondarwin: Added : https://github.com/cloudflare/workers-sdk/pull/6383/commits/34ac024023326edd5b132fb7981eab968c4864eb

packages/wrangler/src/versions/upload.ts
petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/6383/commits/8483a680c7e0d023c987f7d98f1df73a9c319fc8

packages/wrangler/src/versions/index.ts
```diff
@@ -275,6 +277,15 @@ export async function versionsUploadHandler(
 		tag: args.tag,
 		message: args.message,
 	});
+
+	if (versionId && name) {
```
RamIdeas, line 281: Why does this need `&& name` but the same code in deploy/deploy.ts doesn't? Not sure which is correct but I think they should be the same, right?

RamIdeas (approved):

packages/wrangler/src/versions/index.ts
petebacondarwin: This is because the check for whether the name is defined or not (causing an error to be thrown) is deep inside the `versionsUpload()` function. So it should always be defined by the time it gets here. But I felt it was too large a refactor to pull that out.

packages/wrangler/src/deploy/index.ts
jrf0110, line 370: Something I was just chatting with the team about - It would be a lot better for us to get a stable ID for the worker. Ideally we'd use the `script_tag` - Would that be possible?

petebacondarwin: The `script_tag` is internal to EWC and not made available to Wrangler.

packages/wrangler/src/versions/index.ts
RamIdeas: Isn't that the same for deploy? Both call `getScriptName`. Both should be consistent.

packages/wrangler/src/deploy/index.ts
petebacondarwin: All the APIs that Wrangler hits use the Worker name and not the ID. AFAIK.

packages/wrangler/src/deploy/index.ts
```diff
@@ -361,6 +363,15 @@ export async function deployHandler(
 		experimentalVersions: args.experimentalVersions,
 	});
 
+	if (deploymentId) {
```
RamIdeas, line 366: https://github.com/cloudflare/workers-sdk/pull/6383/files#r1705821768

packages/wrangler/src/deploy/index.ts
jrf0110: Wrangler does use script_tag for fetching deployments https://github.com/cloudflare/workers-sdk/blob/362bf3482da81d28d4e8821535e9b09319ceabe2/packages/wrangler/src/deployments.ts#L71 The problem with using `name` for us is that users can update the name in dash and then all of our internal references would be out-of-sync

packages/wrangler/src/deploy/index.ts
petebacondarwin: Ahah. So in that case I made the entry allow `worker_id` to be undefined!

petebacondarwin: I can do that too for version uploads I guess.

packages/wrangler/src/deploy/index.ts
petebacondarwin: So basically someone has to make this call:

```
fetchResult<ServiceMetadataRes>(`/accounts/${accountId}/workers/services/${scriptName}`)).default_environment.script.tag;
```

I guess that Wrangler could do it since it will also have the authentication tokens etc.

I'll update to get that.

RamIdeas: deploy() already makes that call https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/deploy/deploy.ts so you can add it to the returned object

petebacondarwin: Thanks @RamIdeas

PR #6399, opened 2024-08-01
Create manually triggered Github Action for running CI on behalf of forks

RamIdeas:
This PR introduces a GIthub Action which can be triggered manually to clone a fork to a new branch and open a draft PR with the "e2e" label which will (hopefully) run the e2e jobs.

Inputs:
- PR number `type: string` (the PR to clone from) 
- Reviewed `type: boolean` (to confirm the user has reviewed the PR for usage/leakage of secrets)

Also:
- The user triggering the action is enforced to be on the `wrangler` team

### DIFF
--- a/.github/workflows/run-ci-for-external-forks.yml
+++ b/.github/workflows/run-ci-for-external-forks.yml
@@ -1,0 +1,25 @@
+# This workflow cleans up any leftover projects created by e2e runs.
+
+name: Run CI on behalf of External Forks
+on:
+  workflow_dispatch:
+    inputs:
+      pr-number:
+        description: "The PR number to run CI on behalf of"
+        required: true
+      reviewed:
+        description: "Confirm that the PR has been reviewed for use/leakage of secrets"
+        type: boolean
+        required: true
+jobs:
+  create-draft-pr:
+    if: ${{ inputs.reviewed == true }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout PR"
+        run: gh pr checkout ${{ inputs.pr-number }}
+      - name: "Create Draft PR"
+        run: git checkout -b run-ci-on-hehalf-of-${{ inputs.pr-number }}
+          gh pr create --draft --title "Run CI on behalf of \#${{ inputs.pr-number }}" --body "This PR is created to run CI on behalf of \#${{ inputs.pr-number }}. It can be closed after the CI run is complete."
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
--- a/.github/workflows/run-ci-for-external-forks.yml
+++ b/.github/workflows/run-ci-for-external-forks.yml
@@ -16,10 +16,24 @@ jobs:
     if: ${{ inputs.reviewed == true }}
     runs-on: ubuntu-latest
     steps:
+      - name: Check user for team affiliation
+        uses: tspascoal/get-user-teams-membership@v2
+        id: teamAffiliation
+        with:
+          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+          username: ${{ github.actor }}
+          team: wrangler
+      - name: Stop workflow if user is no member
+        if: ${{ steps.teamAffiliation.outputs.isTeamMember == false }}
+        run: |
+          echo "You have must be on the "wrangler" team to trigger this job."
+          exit 1
+
       - name: "Checkout PR"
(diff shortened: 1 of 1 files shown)


.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check user for team affiliation
+        uses: tspascoal/get-user-teams-membership@v2
+        id: teamAffiliation
+        with:
+          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+          username: ${{ github.actor }}
+          team: wrangler
+      - name: Stop workflow if user is no member
```
andyjessop, line 26:
```suggestion
      - name: Stop workflow if user is not a "wrangler" team member
```
?

.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -0,0 +1,39 @@
...
+        if: ${{ steps.teamAffiliation.outputs.isTeamMember == false }}
+        run: |
+          echo "You have must be on the "wrangler" team to trigger this job."
+          exit 1
+
+      - name: "Checkout PR"
+        run: gh pr checkout ${{ inputs.pr-number }}
+
+      - name: "Create Draft PR"
+        run: git checkout -b run-ci-on-hehalf-of-${{ inputs.pr-number }}
```
andyjessop, line 36: Should we run a `branch -D` on this branch first? Unlikely conflict, but possible.

.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -0,0 +1,39 @@
+# This workflow cleans up any leftover projects created by e2e runs.
+
+name: Run CI on behalf of External Forks
+on:
+  workflow_dispatch:
+    inputs:
+      pr-number:
+        description: "The PR number to run CI on behalf of"
```
andyjessop, line 8: ~~The PR number of which behalf to run CI on~~
~~The PR number on behalf which of to run CI~~
~~The PR number half behon to which CI~~

Nah, you're good.

.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  create-draft-pr:
+    if: ${{ inputs.reviewed == true }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check user for team affiliation
+        uses: tspascoal/get-user-teams-membership@v2
+        id: teamAffiliation
+        with:
+          GITHUB_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
```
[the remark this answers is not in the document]
andyjessop: This is a good point. The source code is "obfuscated" in some sense because `dist` is built and committed to the repo, so there is a security issue there.

We might look instead at copy/pasting the source into our own source code and calling it directly.

andyjessop:
One more thing we might consider is to update [this section of `CONTRIBUTING.md`](tspascoal/get-user-teams-membership), explaining that some tests may fail and that it requires a wrangler team member to run the action.

.github/workflows/run-ci-for-external-forks.yml
RamIdeas: 🤣

.github/workflows/run-ci-for-external-forks.yml
RamIdeas: Yeah I was considering this needs to be repeatable. I was thinking a force push or something once I confirmed this all worked but `branch -D` would also work – do we need to delete the remote branch too? and close the existing PR?

.github/workflows/run-ci-for-external-forks.yml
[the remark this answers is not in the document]
RamIdeas: We could also create a read:org only token https://github.com/tspascoal/get-user-teams-membership/blob/master/README.md?plain=1#L23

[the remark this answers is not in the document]
RamIdeas: But forking that repo into our org might be the safest option

.github/workflows/run-ci-for-external-forks.yml
andyjessop: If that's possible then it seems like a good idea, to avoid needing any potential manual cleanup.

penalosa (approved):

andyjessop (approved):

PR #6403, opened 2024-08-02
feat(miniflare): Implement KV/Assets plugin and Workers Assets simulator

CarmenPopoviciu:

This commit extends Miniflare's KV plugin's behaviour to support Workers assets, and therefore enables the emulation of Workers with assets in local development.

Fixes WC-2394

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because:

### DIFF
--- a/.changeset/eight-terms-beg.md
+++ b/.changeset/eight-terms-beg.md
@@ -1,0 +1,8 @@
+---
+"miniflare": minor
+"@cloudflare/workers-shared": minor
+---
+
+feat: Extend KV plugin behaviour to support Workers assets
+
+This commit extends Miniflare's KV plugin's behaviour to support Workers assets, and therefore enables the emulation of Workers with assets in local development.
--- a/fixtures/workers-with-assets/tests/index.test.ts
+++ b/fixtures/workers-with-assets/tests/index.test.ts
@@ -17,10 +17,22 @@ describe("[Workers + Assets] `wrangler dev`", () => {
 		await stop?.();
 	});
 
-	it("renders ", async ({ expect }) => {
-		const response = await fetch(`http://${ip}:${port}/`);
-		const text = await response.text();
+	it("should respond with static asset content", async ({ expect }) => {
+		let response = await fetch(`http://${ip}:${port}/index.html`);
+		let text = await response.text();
 		expect(response.status).toBe(200);
-		expect(text).toContain(`Hello from Asset Server Worker 🚀`);
+		expect(text).toContain(`<h1>Hello Workers + Assets World 🚀!</h1>`);
+
+		response = await fetch(`http://${ip}:${port}/about/index.html`);
+		text = await response.text();
+		expect(response.status).toBe(200);
+		expect(text).toContain(`<p>Learn more about Workers with Assets soon!</p>`);
+	});
+
+	it("should not resolve '/' to '/index.html' ", async ({ expect }) => {
+		let response = await fetch(`http://${ip}:${port}/`);
+		let text = await response.text();
+		expect(response.status).toBe(404);
+		expect(text).toContain("Not Found");
 	});
 });
--- a/packages/miniflare/README.md
+++ b/packages/miniflare/README.md
@@ -528,6 +528,16 @@ parameter in module format Workers.
   If set, only files with paths _not_ matching these glob patterns will be
   served.
 
+  - `assetsPath?: string`
+
+  Path to serve Workers assets from.
+
+  - `assetsKVBindingName?: string`
+    Name of the binding to the KV namespace that the assets are in. If `assetsPath` is set, this binding will be injected into this Worker.
+
+  - `assetsManifestBindingName?: string`
+    Name of the binding to an `ArrayBuffer` containing the binary-encoded assets manifest. If `assetsPath` is set, this binding will be injected into this Worker.
+
 #### R2
 
 - `r2Buckets?: Record<string, string> | string[]`
--- a/packages/miniflare/src/plugins/kv/assets.ts
+++ b/packages/miniflare/src/plugins/kv/assets.ts
@@ -1,0 +1,99 @@
+import { KVOptionsSchema } from "miniflare";
+import SCRIPT_KV_ASSETS from "worker:kv/assets";
+import { z } from "zod";
+import { Service, Worker_Binding } from "../../runtime";
+import { getAssetsBindingsNames, SharedBindings } from "../../workers";
+import { kProxyNodeBinding } from "../shared";
+import { KV_PLUGIN_NAME } from "./constants";
+
+export interface AssetsOptions {
+	assetsPath: string;
+	assetsKVBindingName?: string;
+	assetsManifestBindingName?: string;
+}
+
+export function isWorkersWithAssets(
+	options: z.infer<typeof KVOptionsSchema>
+): options is AssetsOptions {
+	return options.assetsPath !== undefined;
+}
+
+const SERVICE_NAMESPACE_ASSET = `${KV_PLUGIN_NAME}:asset`;
+
+export function buildAssetsManifest(): Uint8Array {
+	const buffer = new ArrayBuffer(20);
+	const assetManifest = new Uint8Array(buffer); // [0, 0, 0, ..., 0, 0]
+	// this will signal to Asset Server Worker that its running in a
+	// local dev "context"
+	assetManifest.set([1], 0); // [1, 0, 0, ..., 0, 0]
+
+	return assetManifest;
+}
+
+export async function getAssetsBindings(
+	options: AssetsOptions
+): Promise<Worker_Binding[]> {
+	const assetsBindings = getAssetsBindingsNames(
+		options?.assetsKVBindingName,
+		options?.assetsManifestBindingName
+	);
+
+	const assetsManifest = buildAssetsManifest();
+
+	return [
+		{
+			// this is the binding to the KV namespace that the assets are in.
+			name: assetsBindings.ASSETS_KV_NAMESPACE,
+			kvNamespace: { name: SERVICE_NAMESPACE_ASSET },
+		},
+		{
+			// this is the binding to an ArrayBuffer containing the binary-encoded
+			// assets manifest.
+			name: assetsBindings.ASSETS_MANIFEST,
+			data: assetsManifest,
+		},
+	];
+}
+
+export async function getAssetsNodeBindings(
+	options: AssetsOptions
+): Promise<Record<string, unknown>> {
+	const assetsManifest = buildAssetsManifest();
+	const assetsBindings = getAssetsBindingsNames(
+		options?.assetsKVBindingName,
+		options?.assetsManifestBindingName
+	);
+
+	return {
+		[assetsBindings.ASSETS_KV_NAMESPACE]: kProxyNodeBinding,
+		[assetsBindings.ASSETS_MANIFEST]: assetsManifest,
+	};
+}
+
+export function getAssetsServices(options: AssetsOptions): Service[] {
+	const storageServiceName = `${SERVICE_NAMESPACE_ASSET}:storage`;
+	const storageService: Service = {
+		name: storageServiceName,
+		disk: { path: options.assetsPath, writable: true },
+	};
+	const namespaceService: Service = {
+		name: SERVICE_NAMESPACE_ASSET,
+		worker: {
+			compatibilityDate: "2023-07-24",
+			compatibilityFlags: ["nodejs_compat"],
+			modules: [
+				{
+					name: "assets.worker.js",
+					esModule: SCRIPT_KV_ASSETS(),
+				},
+			],
+			bindings: [
+				{
+					name: SharedBindings.MAYBE_SERVICE_BLOBS,
+					service: { name: storageServiceName },
+				},
+			],
+		},
+	};
+	return [storageService, namespaceService];
+}
--- a/packages/miniflare/src/plugins/kv/index.ts
+++ b/packages/miniflare/src/plugins/kv/index.ts
@@ -20,6 +20,12 @@ import {
 	Plugin,
 	SERVICE_LOOPBACK,
 } from "../shared";
+import {
+	getAssetsBindings,
+	getAssetsNodeBindings,
+	getAssetsServices,
+	isWorkersWithAssets,
+} from "./assets";
 import { KV_PLUGIN_NAME } from "./constants";
 import {
 	getSitesBindings,
@@ -30,6 +36,11 @@ import {
 
 export const KVOptionsSchema = z.object({
 	kvNamespaces: z.union([z.record(z.string()), z.string().array()]).optional(),
+
+	// Workers + Assets
+	assetsPath: PathSchema.optional(),
+	assetsKVBindingName: z.string().optional(),
+	assetsManifestBindingName: z.string().optional(),
 
 	// Workers Sites
 	sitePath: PathSchema.optional(),
@@ -71,18 +82,29 @@ export const KV_PLUGIN: Plugin<
 			bindings.push(...(await getSitesBindings(options)));
 		}
 
+		if (isWorkersWithAssets(options)) {
+			bindings.push(...(await getAssetsBindings(options)));
+		}
+
 		return bindings;
 	},
+
 	async getNodeBindings(options) {
 		const namespaces = namespaceKeys(options.kvNamespaces);
 		const bindings = Object.fromEntries(
 			namespaces.map((name) => [name, kProxyNodeBinding])
 		);
+
 		if (isWorkersSitesEnabled(options)) {
 			Object.assign(bindings, await getSitesNodeBindings(options));
 		}
+
+		if (isWorkersWithAssets(options)) {
+			Object.assign(bindings, await getAssetsNodeBindings(options));
+		}
 		return bindings;
 	},
+
 	async getServices({
 		options,
 		sharedOptions,
@@ -151,8 +173,13 @@ export const KV_PLUGIN: Plugin<
 			services.push(...getSitesServices(options));
 		}
 
+		if (isWorkersWithAssets(options)) {
+			services.push(...getAssetsServices(options));
+		}
+
 		return services;
 	},
+
 	getPersistPath({ kvPersist }, tmpPath) {
 		return getPersistPath(KV_PLUGIN_NAME, tmpPath, kvPersist);
 	},
--- a/packages/miniflare/src/workers/kv/assets.worker.ts
+++ b/packages/miniflare/src/workers/kv/assets.worker.ts
@@ -1,0 +1,33 @@
+import { SharedBindings } from "miniflare:shared";
+import { KVParams } from "./constants";
+
+interface Env {
+	[SharedBindings.MAYBE_SERVICE_BLOBS]: Fetcher;
+}
+
+export default <ExportedHandler<Env>>{
+	async fetch(request, env) {
+		// Only permit reads
+		if (request.method !== "GET") {
+			const message = `Cannot ${request.method.toLowerCase()}() with Workers Assets namespace`;
+			return new Response(message, { status: 405, statusText: message });
+		}
+
+		// Decode key
+		const url = new URL(request.url);
+		let key = url.pathname.substring(1); // Strip leading "/"
+
+		if (url.searchParams.get(KVParams.URL_ENCODED)?.toLowerCase() === "true") {
+			key = decodeURIComponent(key);
+		}
+
+		const blobsService = env[SharedBindings.MAYBE_SERVICE_BLOBS];
+		if (key === "" || key === "/") {
+			return new Response("Not Found", {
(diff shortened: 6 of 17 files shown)


packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -0,0 +1,116 @@
+import fs from "fs/promises";
```
CarmenPopoviciu, line 1: this is mostly a copy of `/plugin/kv/sites.ts`, but slightly adapted to our Workers + Assets needs

packages/miniflare/src/plugins/kv/index.ts
```diff
@@ -71,18 +80,29 @@ export const KV_PLUGIN: Plugin<
 			bindings.push(...(await getSitesBindings(options)));
 		}
 
+		if (isWorkersWithAssets(options)) {
```
CarmenPopoviciu, line 83: decided to keep the `assets` and `sites` code paths separate so we don't have to worry on the long term about breaking the sites behaviour

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -0,0 +1,142 @@
+import { base64Decode, base64Encode, SharedBindings } from "miniflare:shared";
```
CarmenPopoviciu, line 1: this is a copy of `/kv/sites.worker.ts` but adapted to our needs

packages/miniflare/test/fixtures/assets/modules.ts
```diff
@@ -0,0 +1,35 @@
+import manifestJSON from "__STATIC_CONTENT_MANIFEST";
```
CarmenPopoviciu, line 1: copy of `/fixtures/sites/modules.ts`

packages/miniflare/src/plugins/kv/index.ts
```diff
@@ -71,18 +80,29 @@ export const KV_PLUGIN: Plugin<
...
 		const namespaces = namespaceKeys(options.kvNamespaces);
 		const bindings = Object.fromEntries(
 			namespaces.map((name) => [name, kProxyNodeBinding])
 		);
+
 		if (isWorkersSitesEnabled(options)) {
 			Object.assign(bindings, await getSitesNodeBindings(options));
 		}
+
+		if (isWorkersWithAssets(options)) {
```
CarmenPopoviciu, line 100: do we want this to be an `if...else if` situation instead, and allow declaring only one of `sitePath` or `assetsPath`?

packages/miniflare/src/workers/kv/constants.ts
```diff
@@ -30,6 +30,11 @@ export const SiteBindings = {
 	JSON_SITE_FILTER: "MINIFLARE_SITE_FILTER",
 } as const;
 
+export const AssetsBindings = {
+	KV_NAMESPACE_ASSET: "__STATIC_CONTENT",
+	JSON_ASSET_MANIFEST: "__STATIC_CONTENT_MANIFEST",
+} as const;
```
petebacondarwin, line 36: Should we make this configurable?

packages/miniflare/test/fixtures/assets/modules.ts
petebacondarwin, line 1: I don't think we should support importing this manifest. It should only be available on the `env` object. This was only here for back compatibility with how Workers Sites worked.

packages/miniflare/test/fixtures/assets/modules.ts
```diff
@@ -0,0 +1,35 @@
...
+import { getAssetFromKV } from "@cloudflare/kv-asset-handler";
```
petebacondarwin, line 2: We should not be using `kv-asset-handler` in this fixture.

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,5 +1,9 @@
 export default {
 	async fetch(request, env) {
-		return new Response("Hello from Asset Server Worker 🚀");
+		const url = new URL(request.url);
+		let { pathname } = url;
+
+		const content = await env.__STATIC_CONTENT.get(pathname);
```
petebacondarwin, line 6: If the asset does not exist what does `get()` return? `null`?
If so we would need to deal with that specially to return a 404 response IMO.

packages/miniflare/test/plugins/kv/assets.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+const FIXTURES_PATH = path.resolve(
+	__dirname,
+	"..",
+	"..",
+	"..",
+	"..",
+	"test",
+	"fixtures",
+	"assets"
+);
```
petebacondarwin, line 18: NIT: I prefer this to be:

```suggestion
const FIXTURES_PATH = path.resolve(__dirname, "../../../../test/fixtures/assets");
```

packages/miniflare/test/plugins/kv/assets.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+			...options,
+			scriptPath: t.context.modulesPath,
+			modules: true,
+			assetsPath: tmp,
+		});
+		t.teardown(() => mf.dispose());
+
+		for (const [route, expectedContents] of Object.entries(routeContents)) {
+			const res = await mf.dispatchFetch(`http://localhost:8787${route}`);
+			const expected = expectedRoutes.has(route as Route);
```
petebacondarwin, line 68: I can't see any tests where this would be true?

petebacondarwin (changes requested):

.changeset/eight-terms-beg.md
```diff
@@ -0,0 +1,7 @@
+---
+"miniflare": patch
+---
+
+feat: Implement KV/Assets plugin and Workers Assets simulator
+
+This commit adds a new KV/Assets plugin and an Assets Worker in Miniflare, that are meant to emulate Workers with Assets in local development.
```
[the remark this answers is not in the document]
CarmenPopoviciu: ✅

fixtures/workers-with-assets/tests/index.test.ts
```diff
@@ -18,9 +18,23 @@ describe("[Workers + Assets] `wrangler dev`", () => {
...
+
+		response = await fetch(`http://${ip}:${port}/about/index.html`);
+		text = await response.text();
+		expect(response.status).toBe(200);
+		expect(text).toContain(`<p>Learn more about Workers with Assets soon!</p>`);
+	});
+
+	it("does not resolve '/' to '/index.html' ", async ({ expect }) => {
+		let response = await fetch(`http://${ip}:${port}/`);
+		expect(response.status).toBe(200);
```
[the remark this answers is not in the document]
CarmenPopoviciu: this seems to happen only for Windows but not other OSs. While testing this on MacOS + Chrome, the request returns what is asserted in the test. I need to look into it and understand what is going on. I also need to dbl check what the KV plugin returns if an asset is not found. Maybe smth dubious happening there

packages/miniflare/README.md
```diff
@@ -528,6 +528,14 @@ parameter in module format Workers.
...
   served.
 
+  - `assetsPath?: string`
+
+  Path to serve Workers assets from. If set, `__STATIC_CONTENT` and
+  `__STATIC_CONTENT_MANIFEST` bindings will be injected into this Worker. In
+  modules mode, `__STATIC_CONTENT_MANIFEST` will also be exposed as a module
+  with a `string`-typed `default` export, containing the JSON-stringified
+  manifest. Note Workers assets files are never cached in Miniflare.
+
```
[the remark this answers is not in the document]
CarmenPopoviciu: which would allow us to not enforce `__STATIC_CONTENT` as a binding name on ASW2 I presume? If so, indeed, this is a very good callout! Updated!

packages/workers-shared/asset-server-worker/src/index.ts
CarmenPopoviciu: This is supposed to be handled in WC-2434, unless we strongly object to that

packages/miniflare/src/plugins/kv/index.ts
CarmenPopoviciu: as per our sync meeting today, we should allow for both to work side-by-side. Will look into making that happen

CarmenPopoviciu: ✅

packages/miniflare/test/plugins/kv/assets.spec.ts
CarmenPopoviciu: I might be misunderstanding what you're asking, but isn't [this test](https://github.com/cloudflare/workers-sdk/pull/6403/files#diff-9721850ebe796f312beb43aaba5cc547fd069855bb449e9db21dea1ce82e920dR76-R81) confirming that?

packages/miniflare/test/plugins/kv/assets.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+		assetsPath: tmp,
+	});
+	t.teardown(() => mf.dispose());
+	const res = await mf.dispatchFetch("http://localhost:8787/ń.txt");
+	t.is(await res.text(), "test");
+});
+
+const isWindows = process.platform === "win32";
+const unixTest = isWindows ? test.skip : test;
+unixTest("static content namespace supports listing keys", async (t) => {
```
CarmenPopoviciu, line 148: I don't think listing keys is relevant for Workers + Assets, therefore removing...

fixtures/workers-with-assets/tests/index.test.ts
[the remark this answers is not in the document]
petebacondarwin: Is it a backslash thing?

packages/workers-shared/asset-server-worker/src/index.ts
GregBrimble: Yeah, _lots_ still to add here — totally fine if that's being handled in another PR.

packages/miniflare/test/fixtures/assets/modules.ts
CarmenPopoviciu: indeed. The whole manifest thing kinda had me very confused. I hadn't realised this was specific to Workers Sites and the `kv-asset-handler`. Removed!

packages/miniflare/test/fixtures/assets/modules.ts
CarmenPopoviciu: indeed. This was a misunderstanding on my part wrt to how Workers Sites are set up. `kv-asset-handler` is not relevant to our assets implementation. Removed

packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -18,68 +19,53 @@ export function isWorkersWithAssets(
...
-	// Build __STATIC_CONTENT_MANIFEST contents
-	const staticContentManifest: Record<string, string> = {};
-	for await (const key of listKeysInDirectory(assetsPath)) {
-		staticContentManifest[key] = encodeAssetsKey(key);
-	}
-	return staticContentManifest;
+export function buildAssetsManifest(): Uint8Array {
+	const buffer = new ArrayBuffer(20);
+	const assetManifest = new Uint8Array(buffer); // [0, 0, 0, ..., 0, 0]
+	assetManifest.set([1], 0); // [1, 0, 0, ..., 0, 0]
```
CarmenPopoviciu, line 27: @GregBrimble will this be enough?

packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -18,68 +19,53 @@ export function isWorkersWithAssets(
...
 		},
 	];
 }
 
 export async function getAssetsNodeBindings(
 	options: AssetsOptions
 ): Promise<Record<string, unknown>> {
-	const __STATIC_CONTENT_MANIFEST = await buildStaticContentManifest(
-		options.assetsPath
+	const assetsManifest = await buildAssetsManifest();
```
CarmenPopoviciu, line 60: I'm not 100% sure if `getAssetsNodeBindings` is relevant for our work :/

petebacondarwin: "Node bindings" are to support `getPlatformProxy()` which is used by framework dev tools, such as Next.js.

packages/miniflare/README.md
```diff
@@ -530,11 +530,14 @@ parameter in module format Workers.
...
 
-  Path to serve Workers assets from. If set, `__STATIC_CONTENT` and
-  `__STATIC_CONTENT_MANIFEST` bindings will be injected into this Worker. In
-  modules mode, `__STATIC_CONTENT_MANIFEST` will also be exposed as a module
-  with a `string`-typed `default` export, containing the JSON-stringified
-  manifest. Note Workers assets files are never cached in Miniflare.
+  Path to serve Workers assets from.
+  Note Workers assets files are never cached in Miniflare.
+
+  - `assetsKVBindingName?: string`
```
CarmenPopoviciu, line 536: TODO @CarmenPopoviciu a bit more detail for `assetsKVBindingName` and `assetsManifestKVBindingName`

CarmenPopoviciu: ✅

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -931,6 +931,8 @@ function getAssetServerWorker(
 					port: 0,
 				},
 			],
+			assetsPath: config.experimentalAssets.directory,
+			assetsKVBindingName: "ASSETS_KV_NAMESPACE",
```
petebacondarwin, line 935: Should we conifigure `assetsManifestBindingName` here too?

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,5 +1,14 @@
...
+}
+
 export default {
-	async fetch() {
-		return new Response("Hello from Asset Server Worker 🚀");
+	async fetch(request: Request, env: Env) {
+		const url = new URL(request.url);
+		const { pathname } = url;
+
+		const content = await env.ASSETS_KV_NAMESPACE.get(pathname);
```
petebacondarwin, line 11: So in this worker we need to make use of the manifest binding.
For this PR I think it is enough to assert that the binding is of the special form that indicates it is dev mode pass through, and then to create a class that has a `get()` function that is the identity.

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -0,0 +1,34 @@
...
+
+		// Decode key
+		const url = new URL(request.url);
+		let key = url.pathname.substring(1); // Strip leading "/"
+		if (url.searchParams.get(KVParams.URL_ENCODED)?.toLowerCase() === "true") {
+			key = decodeURIComponent(key);
+		}
+
+		// Strip ASSETS_NO_CACHE_PREFIX
+		key = decodeAssetsKey(key);
```
petebacondarwin, line 24: Remove?
Or not if we feel that we should leave this cache busting in place

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -0,0 +1,34 @@
...
+			const message = `Cannot ${request.method.toLowerCase()}() with Workers Assets namespace`;
+			return new Response(message, { status: 405, statusText: message });
+		}
+
+		// Decode key
+		const url = new URL(request.url);
+		let key = url.pathname.substring(1); // Strip leading "/"
+		if (url.searchParams.get(KVParams.URL_ENCODED)?.toLowerCase() === "true") {
+			key = decodeURIComponent(key);
+		}
```
petebacondarwin, line 21: Is this possible in the case of assets (compared to sites, etc)

packages/miniflare/src/workers/kv/constants.ts
```diff
@@ -116,3 +117,38 @@ export function testSiteRegExps(
...
+export function decodeAssetsKey(key: string): string {
+	return key.startsWith(ASSETS_NO_CACHE_PREFIX)
+		? decodeURIComponent(key.substring(ASSETS_NO_CACHE_PREFIX.length))
+		: key;
+}
+
+export function isAssetsRequest(request: { url: string }) {
+	const url = new URL(request.url);
+	return url.pathname.startsWith(`/${ASSETS_NO_CACHE_PREFIX}`);
+}
```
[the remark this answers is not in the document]
CarmenPopoviciu: indeed. I left it in because I was busy last night digging into the whole caching situation, trying to understand how it works. I was hoping there's a quick solution to it all, but doesn't seem like that.

If I'm understanding the code correctly, it would seem that while Miniflare does provide us with a caching primitive via the Cache Plugin, `PUT`ing/`GET`ing from cache is not done automatically for us? I guess that was my expectation, that by "just" having the Cache plugin in place, requests would automatically be intercepted and paths added to or read from cache. But for example, in the case of Workers Sites, it would seem that it's the `kv-assets-handler` that's responsible for handling caching, both adding to cache [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/kv-asset-handler/src/index.ts#L307-L320) and reading from cache [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/kv-asset-handler/src/index.ts#L263-L265).

So by the loos of 👆  I'd gather we'd need a similar mechanism in place, where the actual cache calls are done in ASW2? Unless I'm misunderstanding how this works :/

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: I left it out on purpose as ASW2 wasn't doing anything with, but I see you left a comment asking to add some basic handling of that binding in ASW. Will do that

CarmenPopoviciu: ✅

packages/miniflare/src/workers/kv/constants.ts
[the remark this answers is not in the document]
CarmenPopoviciu: will go ahead and remove any assets caching things out of this PR and will track caching specifically in WC-2508

packages/miniflare/src/workers/kv/assets.worker.ts
CarmenPopoviciu: ✅

packages/miniflare/src/workers/kv/assets.worker.ts
CarmenPopoviciu: tested, and yes, we do need to decode the key
Screenshot 2024-08-13 at 19 28 21

fixtures/workers-with-assets/tests/index.test.ts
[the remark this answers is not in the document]
CarmenPopoviciu: alright, so given that this test was asserting on a result that we would never encounter in prod, given that if no asset is found we'd return a 404, I've updated ASW2 to actually return a 404 in such use cases. This behaviour was not in place, pending proper implementation of ASW, but seems like effort to fix this test was not worth it, only to then have to change the assertion anyway. [This change](https://github.com/cloudflare/workers-sdk/pull/6403/commits/9af15dccf7e295dd02aefa70c2e1b794311e08b7#diff-67a90fb405dc23a85bdeb2be30130464cbfcc7c742e4f48cab05bee45f8d36dbR32) should fix this test

CarmenPopoviciu:
@petebacondarwin this is ready to re-review

packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -0,0 +1,102 @@
...
+	return {
+		[assetsBindings.ASSETS_KV_NAMESPACE]: kProxyNodeBinding,
+		[assetsBindings.ASSETS_MANIFEST]: assetsManifest,
+	};
+}
+
+export function getAssetsServices(options: AssetsOptions): Service[] {
+	// Use unsanitised file storage to ensure file names containing e.g. dots
+	// resolve correctly.
+	const persist = path.resolve(options.assetsPath);
```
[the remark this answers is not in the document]
petebacondarwin: Yeah if this is already absolute then this can be removed.
The reason it would make it worse on Windows is that a Windows absolute path will look like:

```
C:\users\carmen\projects\foo
```

and resolving that with the posix version will do something like:

```
/current/working/directory/C:\users\carmen\projects\foo
```

😭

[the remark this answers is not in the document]
petebacondarwin: It only makes sense to use `posix.resolve()` on a relative path.

petebacondarwin (approved):
Let's get this in and iterate next week.

PR #6408, opened 2024-08-02
on by default: –x-dev-env

RamIdeas:
## What this PR solves / how to test

Fixes [DEVX-1298](https://jira.cfdata.org/browse/DEVX-1298)

This PR updates the default value for `--x-dev-env` for `wrangler dev` (not yet for `wrangler pages dev`).

It also updates the e2e tests (which assumed the default value was false) to explicitly test with the flag on + off (TODO: need to update the snapshots because this changed the test names)

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Not necessary because:
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [x] Changeset included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: experimental flag

PR #6409, opened 2024-08-02
Update e2e tests to use startWorker directly

RamIdeas:

Update e2e tests to use startWorker directly instead of DevEnv.startWorker

startWorker is the public interface whereas DevEnv provides access to internals (controllers) for advanced use-cases with APIs that we will iterate on and do not want to commit to yet.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: test change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/e2e/startWorker.test.ts
+++ b/packages/wrangler/e2e/startWorker.test.ts
@@ -1,5 +1,6 @@
 import assert from "assert";
 import events from "events";
+import { afterEach } from "node:test";
 import path from "path";
 import { setTimeout } from "timers/promises";
 import getPort from "get-port";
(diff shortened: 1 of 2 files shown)


petebacondarwin:
We should get this green and landed before we flip the --x-dev to default to on. cc @RamIdeas

RamIdeas:
> We should get this green and landed before we flip the --x-dev to default to on. cc @RamIdeas

We can and will because the only fail is a flake (albeit repeatedly). But this is unrelated to flipping the default and [that PR](https://github.com/cloudflare/workers-sdk/pull/6408) will need some test rewrites before landing

petebacondarwin:
Cool. But also before we actually flip the default I would like to understand (and ideally fix) why the file based registry is not compatible with some of these fixtures when this flag is on.

packages/wrangler/e2e/startWorker.test.ts
```diff
@@ -1,5 +1,6 @@
 import assert from "assert";
 import events from "events";
+import { afterEach } from "node:test";
```
petebacondarwin, line 3: This looks wong...

packages/wrangler/e2e/startWorker.test.ts
```diff
@@ -30,16 +31,17 @@ function waitForMessageContaining<T>(ws: WebSocket, value: string): Promise<T> {
...
 	let helper: WranglerE2ETestHelper;
 	let wrangler: Wrangler;
-	let DevEnv: Wrangler["unstable_DevEnv"];
+	let startWorker: Wrangler["unstable_startWorker"];
 	beforeEach(async () => {
 		helper = new WranglerE2ETestHelper();
 		wrangler = await helper.importWrangler();
-		DevEnv = wrangler.unstable_DevEnv;
+		startWorker = wrangler.unstable_startWorker;
+		startWorker = wrangler.unstable_startWorker;
```
petebacondarwin, line 39: duplicate?

petebacondarwin (approved):

PR #6412, opened 2024-08-02
style(create-cloudflare): update welcome and summary message

edmundhung:

This PR updates both the styles of our welcome message and summary.

Before:
Screenshot 2024-08-02 at 21 12 09
Screenshot 2024-08-02 at 21 12 55

After: 
Screenshot 2024-08-06 at 18 18 31
Screenshot 2024-08-06 at 18 18 51
Screenshot 2024-08-06 at 18 19 00

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because: Style updates
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Style updates

### DIFF
--- a/packages/create-cloudflare/src/__tests__/dialog.test.ts
+++ b/packages/create-cloudflare/src/__tests__/dialog.test.ts
@@ -1,0 +1,19 @@
+import { describe, expect, test } from "vitest";
+import { collectCLIOutput, normalizeOutput } from "../../../cli/test-util";
+import { printWelcomeMessage } from "../dialog";
+
+describe("dialog helpers", () => {
+	const std = collectCLIOutput();
+
+	test("printWelcomeMessage", () => {
+		printWelcomeMessage("0.0.0");
+
+		expect(normalizeOutput(std.out)).toMatchInlineSnapshot(`
+			" ╭──────────────────────────────────────────────────────────────╮
+			 │ ☁️  Welcome to create-cloudflare v0.0.0!                      │
+			 │ 🧡 Let's get started.                                        │
+			 ╰──────────────────────────────────────────────────────────────╯
+			"
+		`);
+	});
+});
--- a/packages/create-cloudflare/src/__tests__/dialog.test.ts
+++ b/packages/create-cloudflare/src/__tests__/dialog.test.ts
@@ -63,8 +63,8 @@ describe("dialog helpers", () => {
 				 │    Dash: https://dash.cloudflare.com/?to=/:account/workers/services/view/test-project │
 				 │                                                                                       │
 				 │ 💻 Continue Developing                                                                │
-				 │    Start dev server: npm run start                                                   │
-				 │    Deploy again: npm run deploy                                                      │
+				 │    Start dev server: pnpm run start                                                   │
+				 │    Deploy again: pnpm run deploy                                                      │
 				 │                                                                                       │
 				 │ 📖 Explore Documentation                                                              │
 				 │    https://developers.cloudflare.com/workers                                          │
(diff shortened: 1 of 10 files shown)


packages/create-cloudflare/src/dialog.ts
```diff
@@ -0,0 +1,169 @@
...
+			: []),
+		`   ${gray("Start dev server:")} ${blue(devServerCommand)}`,
+		`   ${gray(ctx.deployment.url ? `Deploy again:` : "Deploy:")} ${blue(deployCommand)}`,
+		``,
+		`📖 Explore Documentation`,
+		`   ${blue.underline(hyperlink(documentationUrl))}`,
+		``,
+		`💬 Join our Community`,
+		`   ${blue.underline(hyperlink(discordUrl))}`,
+		``,
```
jculvey, line 159: I think we should have equal top and bottom margins in this dialog, it feels a little lopsided having the extra newline at the bottom:
image

jculvey: Also, I like the typography and icon changes here, but having a dialog in the summary feels a little weird since it's breaking from the convention we're using everywhere else of using `├` for headers and `│` for body text.

jculvey (commented):
Since we're touching the summary, it would probably also be a good time to add an item in next steps for `npm run cf-typegen` since this command is now pretty much standard across all the templates

packages/create-cloudflare/src/dialog.ts
edmundhung: Thanks for pointing this out! We have moved the dialog out of section 3. I have also updated the screenshots to reflect the changes.

jculvey (approved):

packages/create-cloudflare/src/__tests__/dialog.test.ts
```diff
@@ -0,0 +1,137 @@
...
+				 │                                                                                       │
+				 │ 💻 Continue Developing                                                                │
+				 │    Start dev server: npm run start                                                   │
+				 │    Deploy again: npm run deploy                                                      │
+				 │                                                                                       │
+				 │ 📖 Explore Documentation                                                              │
+				 │    https://developers.cloudflare.com/workers                                          │
+				 │                                                                                       │
+				 │ 💬 Join our Community                                                                 │
+				 │    https://discord.cloudflare.com                                                     │
```
RamIdeas, line 73: Is the alignment of the right hand side border correct here?

I understand emoji rendering on github may affect certain line widths but it looks like some other lines are off by a bit as well

Should we add an additional assertion that all lines must be the same number of chars? Might not be worth if if we have to use the Segmenter and special-case emoji width replacements

packages/create-cloudflare/src/__tests__/dialog.test.ts
```diff
@@ -0,0 +1,137 @@
...
+				 │                                                              │
+				 │ 💬 Join our Community                                        │
+				 │    https://discord.cloudflare.com                            │
+				 ╰──────────────────────────────────────────────────────────────╯
+				"
+			`);
+		});
+
+		test("with lines truncated", async () => {
+			process.stdout.columns = 40;
```
RamIdeas, line 111: This should be unset in an afterEach to not affect any following tests (there aren't any now but additional tests may be added in the future and it shouldn't be on the next dev to figure that out)

packages/create-cloudflare/src/cli.ts
```diff
@@ -127,8 +126,10 @@ const runTemplate = async (ctx: C3Context) => {
 	await create(ctx);
 	await configure(ctx);
 	await deploy(ctx);
-
 	await printSummary(ctx);
+
+	logRaw("");
+	process.exit(0);
```
RamIdeas, line 132: Is this line needed?

packages/create-cloudflare/src/__tests__/dialog.test.ts
edmundhung: It indeed has a different number of chars on lines without emoji even though it looks totally aligned on the terminal 😅 Looking into it why...

edmundhung: It was because we normalized the package manager (from `pnpm` to `npm` ). So those 2 lines with the script will always be one characters less 😅 

I am gonna stop normalizing the package manager as we always run the unit tests in pnpm.

RamIdeas (approved):

PR #6418, opened 2024-08-05
style(create-cloudflare): add description support to select prompt

edmundhung:

This PR adds description to each categories and templates with proper text wrapping logic.

Screenshot 2024-08-05 at 12 35 05
Screenshot 2024-08-05 at 12 35 22
Screenshot 2024-08-05 at 12 35 45
Screenshot 2024-08-05 at 12 36 02

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Styles update

### DIFF
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -214,13 +214,13 @@ export const selectTemplate = async (args: Partial<C3Args>) => {
 			{
 				label: "Framework Starter",
 				value: "web-framework",
-				description:
-					"Select from a range of starter applications using various Cloudflare products",
+				description: "Select from the most popular full-stack web frameworks",
 			},
 			{
 				label: "Demo application",
 				value: "demo",
-				description: "Select from the most popular full-stack web frameworks",
+				description:
+					"Select from a range of starter applications using various Cloudflare products",
 			},
 			{
 				label: "Template from a Github repo",
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -209,22 +209,23 @@ export const selectTemplate = async (args: Partial<C3Args>) => {
 				label: "Hello World example",
 				value: "hello-world",
 				description:
-					"[Recommended for beginners] Build an empty starter to configure project structure and bare-bones",
+					"Select from barebones examples to get started with Workers",
 			},
 			{
 				label: "Framework Starter",
 				value: "web-framework",
-				description: "Create a website with a JavaScript framework",
+				description:
+					"Select from a range of starter applications using various Cloudflare products",
 			},
 			{
 				label: "Demo application",
 				value: "demo",
-				description: "Build a web application using Cloudflare Workers",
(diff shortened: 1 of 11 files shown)


packages/create-cloudflare/templates/hello-world-durable-object/c3.ts
```diff
@@ -2,6 +2,8 @@ export default {
 	configVersion: 1,
 	id: "hello-world-durable-object",
 	displayName: "Hello World Worker Using Durable Objects",
+	description:
```
jculvey, line 5: The template seems a little more bare bones than this to me. Where are the multiplayer application parts in the template?

packages/create-cloudflare/src/templates.ts
```diff
@@ -203,10 +205,27 @@ export const selectTemplate = async (args: Partial<C3Args>) => {
...
 		options: [
-			{ label: "Hello World example", value: "hello-world" },
-			{ label: "Framework Starter", value: "web-framework" },
-			{ label: "Demo application", value: "demo" },
-			{ label: "Template from a Github repo", value: "remote-template" },
+			{
+				label: "Hello World example",
+				value: "hello-world",
+				description:
+					"[Recommended for beginners] Build an empty starter to configure project structure and bare-bones",
```
jculvey, line 212: I think the copy on these descriptions needs a little more refinement. `Bare-bones starter projects for Workers` or something along those lines might be better here.

packages/create-cloudflare/src/templates.ts
```diff
@@ -203,10 +205,27 @@ export const selectTemplate = async (args: Partial<C3Args>) => {
...
+			},
+			{
+				label: "Framework Starter",
+				value: "web-framework",
+				description: "Create a website with a JavaScript framework",
```
jculvey, line 217: Perhaps something like `Create an application with a full-stack web framework` might be better

packages/create-cloudflare/src/templates.ts
```diff
@@ -203,10 +205,27 @@ export const selectTemplate = async (args: Partial<C3Args>) => {
...
+			},
+			{
+				label: "Demo application",
+				value: "demo",
+				description: "Build a web application using Cloudflare Workers",
```
jculvey, line 222: `Build a web application using Cloudflare Workers` -- not all these examples here are web apps, and this doesn't sufficiently distinguish this category from the `hello world` templates. Id probably highlight that the examples here are more full-fledged and showcase advanced usage of the developer products.

edmundhung:
Thanks for the feedback @jculvey! Those descriptions do require some more thoughts. I will share this with Nevi and see how we can make it better.

jculvey (approved):
Great work!

RamIdeas (approved):

PR #6427, opened 2024-08-06
miniflare: kill workerd process with tree-kill

RamIdeas:

- Tests
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/packages/miniflare/package.json
+++ b/packages/miniflare/package.json
@@ -49,6 +49,7 @@
 		"exit-hook": "^2.2.1",
 		"glob-to-regexp": "^0.4.1",
 		"stoppable": "^1.1.0",
+		"tree-kill": "^1.2.2",
 		"undici": "^5.28.4",
 		"workerd": "1.20240821.1",
 		"ws": "^8.17.1",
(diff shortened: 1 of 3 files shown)


RamIdeas:
This fix seems to work when running wrangler via npx (public registry and prerelease urls) but not when we run via pnpm workspaces (in fixtures and our test suite). Therefore I'm having trouble writing a regression test.

Since this fix works for real (npx) usage, I'm inclined to merge this without a test for the benefit of users. We can work on  regression tests and finding root causes async but land this fix early.

RamIdeas:
I tried `fkill` instead of `tree-kill` and it's almost like it kills the process _too_ quickly and many of the miniflare tests fail

penalosa:
Closing in favour of https://github.com/cloudflare/workers-sdk/pull/7169

PR #6428, opened 2024-08-06
feat(create-cloudflare): allow users to change prompt answer

edmundhung:

This PRs introduces a "Go back" option to most template related questions so users can go back to the previous question and change the answers.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: No impact to wrangler
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: No changes in behavior.

### DIFF
--- a/packages/create-cloudflare/e2e-tests/cli.test.ts
+++ b/packages/create-cloudflare/e2e-tests/cli.test.ts
@@ -236,8 +236,8 @@ describe.skipIf(frameworkToTest || isQuarantineMode())(
 							matcher: /What would you like to start with\?/,
 							input: {
 								type: "select",
-								searchBy: "description",
-								target:
+								target: "Demo application",
+								assertDescriptionText:
 									"Select from a range of starter applications using various Cloudflare products",
 							},
 						},
@@ -245,8 +245,9 @@ describe.skipIf(frameworkToTest || isQuarantineMode())(
 							matcher: /Which template would you like to use\?/,
 							input: {
 								type: "select",
-								searchBy: "description",
-								target: "Get started building a basic API on Workers",
+								target: "API starter (OpenAPI compliant)",
+								assertDescriptionText:
+									"Get started building a basic API on Workers",
 							},
 						},
 					],
@@ -290,16 +291,16 @@ describe.skipIf(frameworkToTest || isQuarantineMode())(
 							matcher: /Which template would you like to use\?/,
 							input: {
 								type: "select",
-								defaultValue: "Queue consumer & producer Worker",
 								target: "Go back",
+								assertDefaultSelection: "Queue consumer & producer Worker",
 							},
 						},
 						{
 							matcher: /What would you like to start with\?/,
 							input: {
 								type: "select",
-								defaultValue: "Demo application",
 								target: "Framework Starter",
+								assertDefaultSelection: "Demo application",
 							},
 						},
 						{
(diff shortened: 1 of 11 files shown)


packages/create-cloudflare/src/templates.ts
```diff
@@ -159,8 +159,8 @@ export const getTemplateMap = async () => {
 	} as Record<string, TemplateConfig>;
 };
 
-export const inferRelatedArgs = (args: Partial<C3Args>) => {
-	// Infering the type based on the additional arguments provided
+export const deriveCorelatedArgs = (args: Partial<C3Args>) => {
```
Skye-31, line 162: Little typo 🙂 ```suggestion
export const deriveCorrelatedArgs = (args: Partial<C3Args>) => {
```

Comment (edmundhung) [reply packages/create-cloudflare/src/templates.ts]: Oops! Thanks Skye.

File: packages/create-cloudflare/e2e-tests/cli.test.ts [hunk 2]
```diff
@@ -258,5 +258,98 @@ describe.skipIf(frameworkToTest || isQuarantineMode())(
...
+						},
+						{
+							matcher: /Which language do you want to use\?/,
+							input: {
+								type: "select",
+								target: "Go back",
+							},
+						},
+						{
+							matcher: /● Queue consumer & producer Worker/,
```
RamIdeas, line 290: Is this matcher not meant to match a prompt question? What does it mean to match an active option and then select "Go back"?

edmundhung: Yes. It is intentional to confirm that we have the previous selected option as the default value.

packages/create-cloudflare/src/templates.ts
```diff
@@ -200,115 +202,249 @@ export const selectTemplate = async (args: Partial<C3Args>) => {
...
+				newArgs.framework = undefined;
+				args.framework = undefined;
+				break;
+			case "lang":
+				linesPrinted = 12;
+				newArgs.type = undefined;
+				newArgs.lang = undefined;
+				args.lang = undefined;
+				break;
+		}
```
RamIdeas, line 252:
```suggestion
		switch (from) {
			case "type":
				linesPrinted = 9;
				newArgs.category = undefined;
				break;
			case "framework":
				linesPrinted = 9;
				newArgs.category = undefined;
				break;
			case "lang":
				linesPrinted = 12;
				newArgs.type = undefined;
				break;
		}
		
		newArgs[from] = undefined;
		args[from] = undefined;
```

packages/create-cloudflare/e2e-tests/cli.test.ts
RamIdeas: Makes sense. But it's a bit implicit. If this tests fails in the future, I'm not sure the dev would know why this is the matcher.

We should either leave a comment on each of these matchers as-is or extend the PromptHandler config to match multiple regexes so we can test for the question _and_ this active element

edmundhung: I have updated the select prompt handler to support a `defaultValue` option.

Issue #6405, opened 2024-08-02
🐛 BUG: `require('debug')` resolves to `@cloudflare/vitest-pool-workers/dist/worker/lib/debug.mjs` instead of to the correct file


hansottowirtz:
### Which Cloudflare product(s) does this pertain to?

Workers Vitest Integration

### What version(s) of the tool(s) are you using?

0.4.17

### What version of Node are you using?

20

### What operating system and version are you using?

macOS Sonoma 14.5

### Observed behavior
When importing `knex`, a CJS module importing `debug`, an error is thrown: Cannot use require() to import an ES Module.
This is because `require('debug')` resolves to `@cloudflare/vitest-pool-workers/dist/worker/lib/debug.mjs`, due to these lines:

https://github.com/cloudflare/workers-sdk/blob/eb6f249510a9d36a91d459a2c5918f65b74d2c36/packages/vitest-pool-workers/src/pool/module-fallback.ts#L318-L337

This results in a redirect:
```
VITEST-POOL-WORKERS:MODULE-FALLBACK 24525: require("<root>/node_modules/knex/lib/debug") relative to <root>/node_modules/knex/lib/client.js?mf_vitest_no_cjs_esm_shim: redirect: <root>/node_modules/@cloudflare/vitest-pool-workers/dist/worker/lib/debug.mjs?mf_vitest_no_cjs_esm_shim
```

### Expected behavior

The file should resolve to the correct module.

### Steps to reproduce

Make a worker with `import Knex from "knex"`, and use it in a Vitest test.

### Please provide a link to a minimal reproduction

https://github.com/hansottowirtz/workers-sdk-require-debug-repro

edmundhung:
We have just added Vite dependency pre-bundling support on [v0.6.14](https://github.com/cloudflare/workers-sdk/releases/tag/%40cloudflare%2Fvitest-pool-workers%400.6.14) that should help resolving this issue. 

For example, you can get `ext-dep` working with the following config:

```ts
export default defineWorkersConfig({
  test: {
    deps: {
      optimizer: {
        ssr: {
          enabled: true,
          include: ["ext-dep"],
        },
      },
    },
    poolOptions: {
      workers: {
        // ...
      },
    },
  },
});
```

penalosa:
Closing, since dependency pre-bundling resolves this issue

PR #6444, opened 2024-08-08
fix(vitest-pool-workers): add node lib timers implementation

hansottowirtz:

Adds node lib timers implementation, so that the following code works:

```
require('node:timers').setTimeout
```

`setImmediate`, `active`, `unroll`, and `promises` are not implemented.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: they weren't there before, but I'm willing to add some
  - [ ] I don't know
  - [x] Not necessary because: not relevant for this type of fix
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: wasn't there before

### DIFF
--- a/packages/vitest-pool-workers/src/worker/lib/node/timers.ts
+++ b/packages/vitest-pool-workers/src/worker/lib/node/timers.ts
@@ -1,1 +1,9 @@
-export default {};
+// partial implementation of node:timers
+// missing functions are setImmediate, active, enroll, unenroll,
+// and the timers.promises object
+export const setTimeout = globalThis.setTimeout.bind(globalThis);
+export const clearTimeout = globalThis.clearTimeout.bind(globalThis);
+export const setInterval = globalThis.setInterval.bind(globalThis);
+export const clearInterval = globalThis.clearInterval.bind(globalThis);
+const _default = { setTimeout, clearTimeout, setInterval, clearInterval };
+export default _default;


petebacondarwin:
If it is this simple, I wonder why @mrbbot didn't add this in the first place?

hansottowirtz:
Thanks for your quick replies! We're trying to test our worker, and before deployment, we use esbuild to resolve `node:timers`, `node:fs` etc (using https://github.com/bubblydoo/cloudflare-workers-compat).
However, in tests, we preferably just import our worker directly from typescript without a building step (to have faster refreshes etc).
Unfortunately if we do:

```
defineWorkersConfig({
  ...,
  resolve: {
    alias: {
       timers: './aliases/timers.cjs',
       'node:timers': './aliases/timers.cjs',
    }
  }
})
```

Then, the `worker/lib/node/timers.cjs` file will still get priority instead of `./aliases/timers.cjs`.

It would be great if the files in `lib/node` can only be imported by Vitest code. That would also fix #6405.

petebacondarwin:
So if we change the resolution to check the aliases before the built-in modules that would solve it for you?

hansottowirtz:
That would fix this specific issue, but not the others I've opened. Also as mrbbot mentioned the lib files being available in tests is quite a big difference between tests & production.

There are multiple things that can be improved in our situation:

- configuring `bundleDependencies`
- configuring `esbuild` settings
- vitest-specific lib (e.g. fs, timers, debug)
- bundling configuration that's shared between tests & `wrangler deploy`

I will look at the codebase and try to see what I can do, and then I will get back to this PR, maybe with a more holistic solution 🙂

andyjessop:
I'm closing this for now, as `node:timers` has just been implemented in `workerd` (https://github.com/cloudflare/workerd/releases/tag/v1.20250120.0)

PR #6446, opened 2024-08-08
Publish shared utilities for the online editing experience

penalosa:

Scaffolding & initial work for https://jira.cfdata.org/browse/DEVX-1309. This PR exposes a new package `@cloudflare/workers-editor-shared` which will be used to power the Workers Playground and Quick Edit. It also refactors the Workers Playground to depend on this new package.

See https://github.com/cloudflare/wrangler2/blob/penalosa/shared-editor/packages/workers-playground/README.md for instructions on running the Playground locally.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: This code is currently only tested manually
  - [ ] I don't know
  - [x] Not required because: This code is currently only tested manually
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: New package intended for internal use


packages/workers-playground/package.json
```diff
@@ -32,35 +32,26 @@
...
 		"lz-string": "^1.5.0",
-		"react": "^18.2.0",
-		"react-dom": "^18.2.0",
-		"react-router-dom": "^6.15.0",
-		"react-split-pane": "^0.1.92",
-		"react-spring": "^8.0.27",
-		"react-tabs": "^6.0.2",
-		"react-use-websocket": "^3.0.0",
+		"react": "^17.0.2",
+		"react-dom": "^17.0.2",
```
edmundhung, line 40: Why are we downgrading to React 17?

packages/workers-editor-shared/lib/index.ts
```diff
@@ -0,0 +1,6 @@
+export { useInjectSources } from "./useInjectSources";
+export * from "./ipc";
+export { useRefreshableIframe } from "./useRefreshableIframe";
+export { default as SplitPane, DragContext } from "./SplitPane";
+export { default as Frame } from "./Frame";
+export * from "./constants";
```
edmundhung, line 6: Should we make it more explicit what we are exporting from the package?

packages/workers-playground/package.json
penalosa: Because the dashboard is at React 17, and since we're now publishing a shared component library to be used by both, it's easier to use the same react version for both clients. In particular https://github.com/facebook/create-react-app/issues/11769 makes it hard to publish a library for React 17 & 18 because the module reference for `react/jsx-runtime` needs to be `react/jsx-runtime.js` in React 17, but `react/jsx-runtime` in React 18. We're also not really using any React 18 features, and since we control both consumers of the `@cloudflare/workers-editor-shared` package it seemed to make sense to make things simpler for ourselves by only targeting one version of React.

packages/workers-editor-shared/vite.config.ts
```diff
@@ -0,0 +1,49 @@
...
+	};
+}
+
+// https://vitejs.dev/config/
+export default defineConfig({
+	plugins: [
+		pluginRewriteAll(),
+		react(),
+		dts({ rollupTypes: true }),
+		addJsxRuntimeExtension(),
```
andyjessop, line 31: I'm seeing type errors on these, along these lines:

```
Type '{ name: string; configureServer(server: ViteDevServer): () => void; }' is not assignable to type 'PluginOption'.
```

packages/workers-editor-shared/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+	"extends": "@cloudflare/workers-tsconfig/tsconfig.json",
+	"compilerOptions": {
+		"baseUrl": ".",
+		"target": "ES2020",
+		"useDefineForClassFields": true,
+		"lib": ["ES2020", "DOM", "DOM.Iterable"],
+		"module": "ESNext",
+		"skipLibCheck": true,
+		// "preserveSymlinks": true,
```
andyjessop, line 10: I think we can just get rid of this.

packages/workers-editor-shared/lib/useInjectSources.tsx
```diff
@@ -0,0 +1,93 @@
...
+			.decode(modules[entrypoint].contents)
+			.split("\n").length;
+
+		internalLines = totalScriptLength - userScriptLength;
+	}
+
+	return { files, internalLines };
+}
+
+export function useInjectSources(
```
andyjessop, line 50: Given that this is a fairly unique hook, it would be worth just describing what it does and what is its use.

As a kind of nit-addendum, I like to put the main export at the top of the file to make the code more readable. I don't know if this has been discussed elsewhere and a decision has been made, so this is just adding my two euro-cents to the conversation. When I open a file, I like to see immediately:

- what it imports,
- what it exports,
- and a description of what it does.

packages/workers-playground/src/QuickEditor/HTTPTab/ResponseView.tsx
```diff
@@ -42,15 +42,8 @@ function maybeGetLanguage(
...
-		status: string;
-		statusCode: number;
-	} => {
-		return {
-			status: response.headers.get("cf-ew-status") || "",
-			statusCode: window.parseInt(status, 10),
-		};
-	}, [response]);
+	const status = response.headers.get("cf-ew-status") || "";
+	const statusCode = window.parseInt(status, 10);
```
andyjessop, line 46: I think this will return `NaN` in the case of an empty string.

andyjessop (commented):
This is a nice improvement!

packages/workers-editor-shared/package.json
```diff
@@ -0,0 +1,60 @@
+{
+	"name": "@cloudflare/workers-editor-shared",
+	"version": "0.0.15",
```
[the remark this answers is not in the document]
penalosa: At this stage it's pretty arbitrary. Once merged changesets will take over, and in particular the first public release will be `0.1.0` (given the changeset file in this PR).

packages/workers-editor-shared/vite.config.ts
penalosa: Looks like this was because `vite` wasn't a workspace dependency—added

packages/workers-editor-shared/lib/useInjectSources.tsx
penalosa: I've added a description. With regards to the ordering, happy to change to a different order if you'd prefer. I generally tend to stick with exports at the bottom of a file, with all utility functions they depend on above them (so that you can read through a file and see how things are built up), but that's just a mild personal preference rather than anything else.

packages/workers-playground/src/QuickEditor/HTTPTab/ResponseView.tsx
andyjessop: Cool, better safe than sorry!

packages/workers-editor-shared/lib/useInjectSources.tsx
andyjessop: Completely personal preference, I agree. Until someone brings out a paper on it I'm happy to go with whatever everyone else does.

andyjessop (approved):
LGTM!

packages/workers-editor-shared/lib/useInjectSources.tsx
```diff
@@ -0,0 +1,111 @@
...
+ *     potentially some additional lines of code prepended. The calculated `internalLines` number
+ *     represents how many internal lines of code have been prepended, and should be stripped
+ *     before displaying to the user. The prepended lines of code can be arbitrary, but in our context
+ *     usually represent the middleware system: https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/templates/middleware/loader-sw.ts
+ *
+ * this hook will return a callback that can be used to respond to the `load` event of
+ * an iframe which renders the `format-errors` worker. It will construct the `MessageChannel`
+ * & `MessagePort`s appropriately to construct the right messages for inflating the `format-errors`
+ * stack traces: https://github.com/cloudflare/workers-sdk/tree/main/packages/format-errors/README.md#L28
+ */
```
andyjessop, line 67: 👍

PR #6451, opened 2024-08-08
Shows membership information in wrangler whoami

danielrs:

This PR will show membership information in addition to token scopes when a request against an account-scoped resource fails with Authentication Error (code=10000). This is useful because sometimes requests will fail due to membership permissions, even if the token has the appropriate permissions.

For example, here's the last few lines of what we show now if I try to upload a Worker to an account I'm a member of with read-only permissions:

```
...
🔓 Token Permissions: If scopes are missing, you may need to logout and re-login.
Scope (Access)
- account (read)
- user (read)
- workers (write)
- workers_kv (write)
- workers_routes (write)
- workers_scripts (write)
- workers_tail (read)
- d1 (write)
- pages (write)
- zone (read)
- ssl_certs (write)
- ai (write)
- queues (write)
- offline_access
🎢 Membership roles in "Daniel Account": Contact account admin if you need more roles.
- Administrator Read Only
```

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: unit tests are enough.
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: not a new feature, just some extra information in `whoami`.


petebacondarwin (approved):
Nice PR!

petebacondarwin:
Unfortunately CI doesn't agree. Need some work on the tests.

packages/wrangler/src/whoami.ts
```diff
@@ -52,6 +61,40 @@ export async function whoami() {
...
+		);
+		if (!selectedAccount) {
+			return;
+		}
+		const membershipRoles = await fetchMembershipRoles(selectedAccount.id);
+		if (!membershipRoles) {
+			return;
+		}
+		logger.log(
+			`🎢 Membership roles in "${selectedAccount.name}": Contact account admin if you need more roles.`
```
JoshuaYe, line 81: Copy suggestion: 🎢 Membership roles in "${selectedAccount.name}": Contact account admin if you lack permissions for your work.

JoshuaYe (changes requested):
Made small copy suggestion, aside from that looks great Daniel :D

packages/wrangler/src/whoami.ts
tanushree-sharma: Membership roles in "${selectedAccount.name}": Contact your account's super admin to change your permissions.

PR #6455, opened 2024-08-09
fix: ensure that directory imports in commonjs modules get resolved to their index file

### LINKED ISSUE #6406: 🐛 BUG: Node modules `require('./dir')` does not resolve to `require('./dir/index.js')`

Workers Vitest Integration

0.4.17

_No response_

macOS Sonoma 14.5

### Observed & expected behavior

When importing `knex` in a Vitest test, its subdependency `db-errors` will make a `require('./lib/parsers')` call. This should resolve to `./lib/parsers/index.js` but it does not, instead throwing:
```
Error: No such module "<root>/node_modules/db-errors/lib/parsers"
```
This is in contrast to the default Vitest tests, which import it correctly.

### Steps to reproduce

Import `knex` in a test.

https://github.com/hansottowirtz/workers-sdk-index-js-require-repro

```
pnpm i
pnpm test
```

Result will be:

```
Error: No such module "<root>/node_modules/.pnpm/ext-dep-1@file+vendor+ext-dep/node_modules/ext-dep-1/subfolder".
```

_No response_

petebacondarwin:

Fixes #6406

The build from this PR should fix the reproduction here: https://github.com/hansottowirtz/workers-sdk-index-js-require-repro.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/vitest-pool-workers/src/pool/module-fallback.ts
+++ b/packages/vitest-pool-workers/src/pool/module-fallback.ts
@@ -75,6 +75,17 @@ function isFile(filePath: string): boolean {
 	}
 }
 
+function isDirectory(filePath: string): boolean {
+	try {
+		return fs.statSync(filePath).isDirectory();
+	} catch (e) {
+		if (isFileNotFoundError(e)) {
+			return false;
+		}
+		throw e;
+	}
+}
+
 function getParentPaths(filePath: string): string[] {
 	const parentPaths: string[] = [];
 	// eslint-disable-next-line no-constant-condition
@@ -225,6 +236,9 @@ function maybeGetTargetFilePath(target: string): string | undefined {
 	}
 	if (target.endsWith(disableCjsEsmShimSuffix)) {
 		return target;
+	}
+	if (isDirectory(target)) {
+		return maybeGetTargetFilePath(target + "/index");
 	}
 }
 
--- a/.changeset/dry-cobras-smash.md
+++ b/.changeset/dry-cobras-smash.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/vitest-pool-workers": patch
+---
+
+fix: ensure that directory imports in commonjs modules get resolved to their index file
+
+Fixes #6406
--- a/fixtures/vitest-pool-workers-examples/internal-module-resolution/README.md
+++ b/fixtures/vitest-pool-workers-examples/internal-module-resolution/README.md
@@ -1,0 +1,5 @@
+# internal-module-resolution
+
+This simple fixture checks that the Vitest integration import resolution works correctly when a CommonJS packages has a require to a directory rather than a specific file.
+
+There is no Worker defined here but we still need a minimal wrangler.toml that we can pass to the vitest-pool-workers plugin to trigger that plugin to handle import resolution and file loading.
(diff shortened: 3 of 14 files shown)


packages/vitest-pool-workers/src/pool/module-fallback.ts
```diff
@@ -75,6 +75,17 @@ function isFile(filePath: string): boolean {
...
+	try {
+		return fs.statSync(filePath).isDirectory();
+	} catch (e) {
+		if (isFileNotFoundError(e)) {
+			return false;
+		}
+		throw e;
+	}
+}
+
```
andyjessop, line 88: We have basically the same code in several other places in the repo - I wonder if we can consolidate these?

[1](https://github.com/cloudflare/workers-sdk/blob/c2bb42fefa402ece691fd4b034a749d2862ec61a/packages/wrangler/src/global-wrangler-config-path.ts/#L6-L13), [2](https://github.com/cloudflare/workers-sdk/blob/c2bb42fefa402ece691fd4b034a749d2862ec61a/packages/wrangler/src/pages/functions/buildPlugin.ts/#L67-L72), [3](https://github.com/cloudflare/workers-sdk/blob/c2bb42fefa402ece691fd4b034a749d2862ec61a/packages/wrangler/src/pages/functions/buildWorker.ts/#L94-L99), [4](https://github.com/cloudflare/workers-sdk/blob/c2bb42fefa402ece691fd4b034a749d2862ec61a/packages/create-cloudflare/src/helpers/files.ts/#L39-L49). [5](https://github.com/cloudflare/workers-sdk/blob/c2bb42fefa402ece691fd4b034a749d2862ec61a/packages/wrangler/src/pages/dev.ts/#L351-L352), [6](https://github.com/cloudflare/workers-sdk/blob/c2bb42fefa402ece691fd4b034a749d2862ec61a/packages/wrangler/src/pages/functions.ts/#L48-L49)

RamIdeas (approved):

andyjessop (approved):
Looks good! One comment, but more as a case-in-point of something that happens all over the repo. Perhaps it's time to create a utils package as a common place for these types of functions.

packages/vitest-pool-workers/src/pool/module-fallback.ts
petebacondarwin: For sure within a package (such as the 4 versions in wrangler). I am less bothered about sharing little utilities like this between packages.

petebacondarwin: Feel free to follow up with clean up PRs for that sort of thing.

andyjessop: The case for, I would say, is to both standardise and to avoid common errors. This one, for example, will throw an error that TS doesn't pick up:

```
const t = fs.statSync('sdfs').isDirectory();
```

There are other common pitfalls, like `parseInt(couldBeEmptyString, 10) -> Nan`.

One more reason is that we have some nice little gems that might never get discovered because they're hidden away, like [this one from the C3 code base](https://github.com/cloudflare/workers-sdk/blob/c2bb42fefa402ece691fd4b034a749d2862ec61a/packages/create-cloudflare/src/helpers/retry.ts/#L24-L39).

I think a common utility library could make our code more consistent as well as more robust and discoverable.

PR #6474, opened 2024-08-13
feat(wrangler) - make asset server worker the entrypoint for `wrangler dev`

emily-shen:

Fixes [WC-2435](https://jira.cfdata.org/browse/WC-2435)

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: I don't _think_ this change needs additional testing beyond passing existing workers-with-assets fixture...?
  - [ ] I don't know
  - [ ] Not required because: 
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: no public facing behaviour changes
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/deployment-bundle/entry.ts
```diff
@@ -57,10 +55,13 @@ export async function getEntry(
...
-		} else if (args.legacyAssets || config.legacy_assets) {
+		} else if (
+			args.legacyAssets ||
+			config.legacy_assets ||
+			args.experimentalAssets ||
+			config.experimental_assets
+		) {
 			file = path.resolve(getBasePath(), "templates/no-op-worker.js");
-		} else if (args.experimentalAssets || config.experimental_assets) {
-			file = path.resolve(getBasePath(), "templates/no-op-assets-worker.ts");
```
CarmenPopoviciu, line 63: we should prolly delete `templates/no-op-assets-worker.ts` if we're not going to use it anymore

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -867,6 +870,41 @@ export async function buildMiniflareOptions(
...
+	internalObjects: CfDurableObject[];
+	externalWorkers: WorkerOptions[];
+}> {
+	const { bindingOptions, internalObjects, externalWorkers } =
+		buildMiniflareBindingOptions(config);
+	// if static assets only, no user worker:
+	if (
+		config.experimentalAssets &&
+		config.bundle.entry.file ===
+			path.resolve(getBasePath(), "templates/no-op-worker.js")
```
petebacondarwin, line 893: Is this PR only a temporary step?
If not, why doe we still have a `bundle.entry` at all if  it is assets-only?
It looks like we are still creating this entry but marking it with the `no-op-worker.js` to indicate it is not a real one?

CarmenPopoviciu: my concern was that not having an entry worker would require us to put all sorts of `if..else`s in place to by-pass the different layers of bundling we currently have in place. Felt rather risky too, since we'd need to track down all layers where bundling is happening, and was worried we might break existing things, introduce too much complexity, etc. So my suggestion was to use a no-op worker that we can just ignor in assets-only mode.

emily-shen: Yes I originally wanted to never use the no-op worker as an entry at all, but the problem is *a lot* of wrangler assumes there is a user worker. Talked with Carmen and we thought it might be better to keep the changes for this more localised and just remove the no-op worker at the last minute. 

But I'm very open to other solutions, or if you think its fine to let the entry/main module be undefined and add in the appropriate checks around the codebase, then I'm happy to do that :)

emily-shen:
superseded by https://github.com/cloudflare/workers-sdk/pull/6564

emily-shen:
superseded by https://github.com/cloudflare/workers-sdk/pull/6564

PR #6482, opened 2024-08-13
reimplement module aliasing with esbuild plugin

### LINKED ISSUE #6448: user-defined aliases should take precedence over unenv polyfills
We recently introduced module aliasing in https://github.com/cloudflare/workers-sdk/pull/6167

The `experimental:nodejs_compat_v2` flag introduces unenv polyfills via an esbuild plugin onResolve hook. This plugin takes precedence over the user-defined aliases.

A simple repro can be found here https://github.com/RamIdeas/esbuild-alias-wrangler-repro/blob/main/build.mjs

I believe the solution is to reimplement the user-defined aliases via our own plugin which needs to be earlier in the plugin array. Creating this issue to track.

RamIdeas:

Fixes #6448

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/tidy-berries-type.md
+++ b/.changeset/tidy-berries-type.md
@@ -2,4 +2,4 @@
 "wrangler": patch
 ---
 
-feat: reimplement module aliasing so user-defined aliases take precedence over other plugins (eg unenv node.js polyfills)
+fix: reimplement module aliasing so user-defined aliases take precedence over other plugins (eg unenv node.js polyfills)
--- a/.changeset/tidy-berries-type.md
+++ b/.changeset/tidy-berries-type.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+feat: reimplement module aliasing so user-defined aliases take precedence over other plugins (eg unenv node.js polyfills)
--- a/packages/wrangler/src/deployment-bundle/bundle.ts
+++ b/packages/wrangler/src/deployment-bundle/bundle.ts
@@ -301,8 +301,8 @@ export async function bundleWorker(
 			// filter the hook calls to only those that match the alias keys
 			// this should avoid slowing down builds which don't use aliasing
 			const filter = new RegExp(
-				Object.entries(alias)
-					.map(([key, _value]) => escapeRegex(key))
+				Object.keys(alias)
+					.map((key) => escapeRegex(key))
 					.join("|")
 			);
 
--- a/packages/wrangler/src/deployment-bundle/bundle.ts
+++ b/packages/wrangler/src/deployment-bundle/bundle.ts
@@ -27,6 +27,15 @@ import type { ModuleCollector } from "./module-collection";
 import type { NodeJSCompatMode } from "./node-compat";
 import type { CfModule, CfModuleType } from "./worker";
 
+// Taken from https://stackoverflow.com/a/3561711
+// which is everything from the tc39 proposal, plus the following two characters: ^/
+// It's also everything included in the URLPattern escape (https://wicg.github.io/urlpattern/#escape-a-regexp-string), plus the following: -
+// As the answer says, there's no downside to escaping these extra characters, so better safe than sorry
+const ESCAPE_REGEX_CHARACTERS = /[-/\\^$*+?.()|[\]{}]/g;
+const escapeRegex = (str: string) => {
+	return str.replace(ESCAPE_REGEX_CHARACTERS, "\\$&");
+};
+
 export const COMMON_ESBUILD_OPTIONS = {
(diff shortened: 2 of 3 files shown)


.changeset/tidy-berries-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+feat: reimplement module aliasing so user-defined aliases take precedence over other plugins (eg unenv node.js polyfills)
```
petebacondarwin, line 5:
```suggestion
fix: reimplement module aliasing so user-defined aliases take precedence over other plugins (eg unenv node.js polyfills)
```

packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -278,6 +287,42 @@ export async function bundleWorker(
...
+
+			// reimplement module aliasing as an esbuild plugin onResolve hook
+			build.onResolve({ filter }, (args) => {
+				const aliasPath = alias[args.path];
+				if (aliasPath) {
+					return {
+						// resolve with node resolution
+						path: require.resolve(aliasPath, {
+							// From the esbuild alias docs: "Note that when an import path is substituted using an alias, the resulting import path is resolved in the working directory instead of in the directory containing the source file with the import path."
+							// https://esbuild.github.io/api/#alias:~:text=Note%20that%20when%20an%20import%20path%20is%20substituted%20using%20an%20alias%2C%20the%20resulting%20import%20path%20is%20resolved%20in%20the%20working%20directory%20instead%20of%20in%20the%20directory%20containing%20the%20source%20file%20with%20the%20import%20path.
```
CarmenPopoviciu, line 317: nit: perhaps just link to https://esbuild.github.io/api/#alias? We already have the relevant text in the comment above this one

packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -278,6 +287,42 @@ export async function bundleWorker(
...
+				return;
+			}
+
+			// filter the hook calls to only those that match the alias keys
+			// this should avoid slowing down builds which don't use aliasing
+			const filter = new RegExp(
+				Object.entries(alias)
+					.map(([key, _value]) => escapeRegex(key))
+					.join("|")
+			);
```
petebacondarwin, line 307: If we don't need the values then can we just use:

```suggestion
				Object.keys(alias)
					.map((key) => escapeRegex(key))
					.join("|")
			);
```

petebacondarwin (approved):

packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -317,12 +362,12 @@ export async function bundleWorker(
 				...define,
 			},
 		}),
-		alias,
 		loader: {
 			...COMMON_ESBUILD_OPTIONS.loader,
 			...(loader || {}),
 		},
 		plugins: [
+			aliasPlugin,
```
CarmenPopoviciu, line 370: what's the order in which these plugins are executed? Is first to last in the `plugins` array, or the other way around? I can't find that mentioned in the esbuild docs

RamIdeas: First to last. I confirmed by playing with my repro linked in the issue.

CarmenPopoviciu: in that case, what happens if `aliasPlugin` and the unenv one are trying to alias the same module? In that case, won't unenv override the user alias?

RamIdeas: Good question. esbuild runs the onResolve hooks on each plugin in order **until** one does not return undefined. It doesn't run anymore onResolve hooks once it receives a path.

CarmenPopoviciu: TIL! thx for clarifying that 🙏

CarmenPopoviciu (approved):

packages/wrangler/src/deployment-bundle/bundle.ts
CarmenPopoviciu: not sure if we want to capture this use case in a test, I'll leave that up to you

IgorMinar:
Thank you @RamIdeas!

PR #6487, opened 2024-08-13
feat(wrangler) - add assets binding to wrangler deploy

emily-shen:

[WC-2506](https://jira.cfdata.org/browse/WC-2506) - add assets binding

and tacking on:
[WC-2455](https://jira.cfdata.org/browse/WC-2455) - encode file path

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: Part of incomplete/unreleased feature
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/api/startDevWorker/utils.ts
+++ b/packages/wrangler/src/api/startDevWorker/utils.ts
@@ -240,6 +240,11 @@ export function convertCfWorkerInitBindingstoBindings(
 				}
 				break;
 			}
+			case "experimental_assets": {
+				const { binding, ...x } = info;
+				output[binding] = { type: "assets", ...x };
+				break;
+			}
 			default: {
 				assertNever(type);
 			}
@@ -277,6 +282,7 @@ export async function convertBindingsToCfWorkerInitBindings(
 		mtls_certificates: undefined,
 		logfwdr: undefined,
 		unsafe: undefined,
+		experimental_assets: undefined,
 	};
 
 	const fetchers: Record<string, ServiceFetch> = {};
--- a/packages/wrangler/src/api/startDevWorker/utils.ts
+++ b/packages/wrangler/src/api/startDevWorker/utils.ts
@@ -241,8 +241,7 @@ export function convertCfWorkerInitBindingstoBindings(
 				break;
 			}
 			case "experimental_assets": {
-				const { binding, ...x } = info;
-				output[binding] = { type: "assets", ...x };
+				output[info["binding"]] = { type: "assets" };
 				break;
 			}
 			default: {
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -2201,6 +2201,7 @@ const validateUnsafeBinding: ValidatorFn = (diagnostics, field, value) => {
 			"service",
 			"logfwdr",
 			"mtls_certificate",
+			"assets",
 		];
 
 		if (safeBindings.includes(value.type)) {
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -2201,7 +2201,6 @@ const validateUnsafeBinding: ValidatorFn = (diagnostics, field, value) => {
 			"service",
 			"logfwdr",
 			"mtls_certificate",
-			"assets",
 		];
 
 		if (safeBindings.includes(value.type)) {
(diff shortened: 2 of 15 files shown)


packages/wrangler/src/api/startDevWorker/utils.ts
```diff
@@ -240,6 +240,11 @@ export function convertCfWorkerInitBindingstoBindings(
 				}
 				break;
 			}
+			case "experimental_assets": {
+				const { binding, ...x } = info;
```
CarmenPopoviciu, line 244: `x` is not very descriptive here. can we pls use smth more meaningful 🙏 ?

packages/wrangler/src/config/validation.ts
```diff
@@ -2201,6 +2201,7 @@ const validateUnsafeBinding: ValidatorFn = (diagnostics, field, value) => {
 			"service",
 			"logfwdr",
 			"mtls_certificate",
+			"assets",
```
CarmenPopoviciu, line 2204: is this `assets` or `experimental_assets`?

emily-shen: That's the experimental assets binding - since EWC will be expecting the binding type to be 'assets', I've just used that immediately. Is a bit confusing though sorry!

CarmenPopoviciu: but the type of `experimental_assets.binding` is not exposed to end users. The shape we send to EWC is something entirely internal. Furthermore, this validation is part of `validateUnsafeBinding`, which is called as part of `validateUnsafeSettings`, which doesn't seem to be related to our work? AFAICT `validateUnsafeSettings` validates the non-inheritable `unsafe` config key, which does indeed expose the `type`, but is something entirely different
```
unsafe: {
	bindings?: {
		name: string;
		type: string;
		[key: string]: unknown;
	}[];
        ...
}
```

What this code here does, is basically to ensure that if users are trying to configure any non-inheritable bindings via `unsafe`, but that `wrangler.toml` supports directly, we error with the appropriate message. Since `experimental_assets` is inheritable, I don't think we should be adding it to this list

packages/wrangler/src/experimental-assets.ts
```diff
@@ -343,3 +341,11 @@ export function processExperimentalAssetsArg(
 
 	return experimentalAssets;
 }
+
+const validateAndEncodeFilePath = (filePath: string) => {
+	const encodedPath = filePath
+		.split(path.sep)
+		.map((segment) => encodeURIComponent(segment))
+		.join("/");
+	return path.join("/", encodedPath);
```
petebacondarwin: 😱

packages/wrangler/src/config/validation.ts
emily-shen: whoops yeah I think I didn't dig through that enough and misunderstood - fixed

packages/wrangler/src/deploy/index.ts
```diff
@@ -353,7 +345,7 @@ export async function deployHandler(
 		jsxFragment: args.jsxFragment,
 		tsconfig: args.tsconfig,
 		routes: args.routes,
-		experimentalAssets: experimentalAssetsOptions,
+		experimentalAssets: experimentalAssets,
```
CarmenPopoviciu, line 348:
```suggestion
		experimentalAssets,
```

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -171,14 +172,14 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
 		tail_consumers,
 		limits,
 		annotations,
-		experimental_assets,
+		experimental_assets_jwt,
 	} = worker;
 
 	// short circuit if static assets upload only
-	if (experimental_assets?.staticAssetsOnly) {
+	if (main.name === "no-op-assets-worker.js") {
```
CarmenPopoviciu, line 179: this makes me a bit nervous, but I can't think of a better solution rn. I also feel the way we will solve https://github.com/cloudflare/workers-sdk/pull/6474#discussion_r1716794688 will highly influence how we want to perform this detection

emily-shen: Actually I should've put it behind the experimental assets flag as well - that should make it a bit more safe for now til/if we come up with a better solution

CarmenPopoviciu: raised in WC-2534

CarmenPopoviciu (approved):

PR #6497, opened 2024-08-14
GA versions/deployments commands

WalshyDev:

All the commands guarded by `--x-versions` are now un-guarded. Any command that had output without the flag has had it's output updated, we have tried to keep compatibility where possible however.

We have also updated `wrangler deploy` to use the new APIs, this will still fallback to the old script upload API with any of the following conditions:
* User passes `--no-x-versions`
* The Worker does not exist - the versions upload requires a Worker to exist
* It isn't a legacy service env deploy - they have their own path
* It isn't a service Worker - we do not allow SW uploads
* It doesn't contain a DO migration - we cannot yet handle DO migrations (this will be removed in the future)
* It isn't an FPW - these may work but it isn't tested and much less risky to keep on the old for now

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/16373
  - [ ] Not necessary because:


packages/wrangler/src/versions/rollback/index.ts
```diff
@@ -151,6 +133,9 @@ export async function versionsRollbackHandler(args: VersionsRollbackArgs) {
 	cli.success(
 		`Worker Version ${versionId} has been deployed to 100% of traffic.`
 	);
+
+	// Keep compatibility with the way users grep for the version ID today
+	console.log("\nCurrent Version ID: " + versionId);
```
RamIdeas, line 138:
```suggestion
	logger.log("\nCurrent Version ID: " + versionId);
```

packages/wrangler/src/versions/list.ts
```diff
@@ -57,11 +57,9 @@ export async function versionsListHandler(args: VersionsListArgs) {
...
 
 	const versionCache: VersionCache = new Map();
-	const versions = await fetchDeployableVersions(
-		accountId,
-		workerName,
-		versionCache
-	);
+	const versions = (
+		await fetchDeployableVersions(accountId, workerName, versionCache)
+	).sort((a, b) => a.number - b.number);
```
RamIdeas, line 62: What's `.number`? Should we not explicitly sort by `.metadata.created_on`?

packages/wrangler/src/__tests__/versions/deployments/deployments.list.test.ts
```diff
@@ -39,63 +39,63 @@ describe("deployments list", () => {
...
+				Message:     -
+				Version(s):  (10%) 10000000-0000-0000-0000-000000000000
+				                 Created:  2021-01-01T00:00:00.000Z
+				                     Tag:  -
+				                 Message:  -
+
+				             (90%) 20000000-0000-0000-0000-000000000000
+				                 Created:  2021-01-01T00:00:00.000Z
+				                     Tag:  -
+				                 Message:  -
```
RamIdeas, line 68: Why have these 2 deployments been switched in order without a change to the test? Not sure this snapshot should have updated

RamIdeas:
This all looks fine but IMO it would be a lot safer to leave the other code paths around for a while and simply flip the flag default. This would allow users experiencing issue to set the inverse flag `--no-x-versions` (or `--x-versions=false` as a temporary workaround while they report the issue and allow us time to fix.

packages/wrangler/src/__tests__/versions/deployments/deployments.list.test.ts
WalshyDev: We reversed the ordering in the list command - newest is now always at the bottom (in this case 4th is newer than the 1st)

packages/wrangler/e2e/versions.test.ts
```diff
@@ -76,15 +76,15 @@ describe("versions deploy", { timeout: TIMEOUT }, () => {
...
+			Message:     -
 			Version ID:  00000000-0000-0000-0000-000000000000
 			Created:     TIMESTAMP
 			Author:      person@example.com
-			Source:      Upload
-			Tag:         -
-			Message:     -"
+			Source:      Unknown (version_upload)
+			Tag:         e2e-upload
+			Message:     Upload via e2e test"
```
RamIdeas, line 87: Looks like this order has been reversed. Was that intentional?

WalshyDev: Yes, version & deployment list are reversed - newest last (at the bottom)

We looked at the usage and it's very low so we're taking advantage of that + beta to make this change. This will make it much more inline with what users expect and easier to see the latest in terminals.

packages/wrangler/src/__tests__/versions/versions.list.test.ts
```diff
@@ -41,36 +41,36 @@ describe("versions list", () => {
...
+				Message:     Rolled back for this version
 
-			Version ID:  10000000-0000-0000-0000-000000000000
-			Created:     2021-01-04T00:00:00.000Z
-			Author:      Jean-Luc-Picard@federation.org
-			Source:      Rollback
-			Tag:         -
-			Message:     -
+				Version ID:  20000000-0000-0000-0000-000000000000
+				Created:     2021-02-03T00:00:00.000Z
```
RamIdeas, line 66: These snapshots look out of order. Even considering the reversal

WalshyDev: The fake UUIDs are just done based on how the old order used to be, if you looked at the created it's actually all correct.

I'll update the fake UUIDs so it's less confusing

RamIdeas: Argh I was only looking at the days not the months 🤦

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -221,13 +213,9 @@ describe("deploy", () => {
 			- Vars:
 			  - xyz: 123
 			Uploaded test-worker (TIMINGS)
-			Published test-worker (TIMINGS)
+			Deployed test-worker triggers (TIMINGS)
```
RamIdeas, line 216: Good catch!

packages/wrangler/src/__tests__/deployments.test.ts
```diff
@@ -82,173 +79,151 @@ describe("deployments", () => {
 			it("should log deployments", async () => {
 				writeWranglerToml();
 
-				await runWrangler("deployments list");
+				await runWrangler("deployments list --no-x-versions");
 				expect(std.out).toMatchInlineSnapshot(`
-			"🚧\`wrangler deployments\` is a beta command. Please report any issues to https://github.com/cloudflare/workers-sdk/issues/new/choose
-
-
-			Deployment ID: Constitution-Class-tag:test-name
```
RamIdeas, line 90: I feel like we could keep this line when `--no-x-versions` is set? It's probably not a big deal in this instance since we've logged a warning about the future removal of this line in wrangler for months. But one of the ideas of flipping feature flag defaults is to have an escape-hatch for users to go back to the old behaviour. This doesn't preserve the old behaviour.

RamIdeas (approved):

PR #6502, opened 2024-08-15
VS-197: Fix Vectorize List MetadataIndex Http Method

garvit-gupta:

Fixes VS-197. Fix Vectorize List MetadataIndex Http Method. Currently POST, changed it to GET.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:


packages/wrangler/src/vectorize/query.ts
```diff
@@ -28,7 +28,7 @@ export function options(yargs: CommonYargsArgv) {
 				type: "array",
 				demandOption: true,
 				describe:
-					"Vector to query the Vectorize Index. Example: `--vector 1 2 3 0.5 1.25 6`",
+					"Vector to query the Vectorize Index. Example: `--vector 1 2 3 0.5 1.25 6`. To read from a json file that contains data in the format [1, 2, 3], use `--vector $(jq -r '.[]' data.json | xargs)`",
```
threepointone, line 31: this expects the user to have jq and xargs on their system. shouldn't we do this in wrangler itself?

garvit-gupta: That's a good idea. I have created a task for this in our backlog. I have also updated this command to be more of a suggestion than an instruction. I hope that would be sufficient for this change.

.changeset/stale-moles-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Fix Vectorize List MetadataIndex Http Method
```
CarmenPopoviciu, line 5:
```suggestion
fix: Fix Vectorize List MetadataIndex Http Method
```

packages/wrangler/src/vectorize/query.ts
```diff
@@ -28,7 +28,7 @@ export function options(yargs: CommonYargsArgv) {
 				type: "array",
 				demandOption: true,
 				describe:
-					"Vector to query the Vectorize Index. Example: `--vector 1 2 3 0.5 1.25 6`",
+					"Vector to query the Vectorize Index. Example: `--vector 1 2 3 0.5 1.25 6`. To read from a json file that contains data in the format [1, 2, 3], you could use a command like `--vector $(jq -r '.[]' data.json | xargs)`",
```
CarmenPopoviciu, line 31: nit: I'm a little worried about the verbosity here. The scope of the added copy feels more like docs scope rather than CLI help scope. Looking at the `wrangler vectorize query help` output, it's looking like 👇 , which feels a bit difficult to read IMHO. Was this considered, and are we sure this is what we want here?

Screenshot 2024-08-16 at 08 37 51

garvit-gupta: Well the idea was to provide some additional guidance to the customers and make it easier for them to use the query operation. But it is unrelated to the fix in this PR, and I can remove it to unblock the fix if the verbosity is a cause for concern.

CarmenPopoviciu (approved):

packages/wrangler/src/vectorize/query.ts
CarmenPopoviciu: I have an idea how we can fix this! Stay tuned...

CarmenPopoviciu: addressed here: https://github.com/cloudflare/workers-sdk/pull/6513

Issue #5908, opened 2024-05-24
🐛 BUG: Producer sending message to queue throwing Error: Network connection lost in Miniflare


Soviut:
### Which Cloudflare product(s) does this pertain to?

Miniflare

### What version(s) of the tool(s) are you using?

3.20240512.0 [miniflare]

### What version of Node are you using?

20.9.0

### What operating system and version are you using?

Unbutu 22 under WSL2 on Windows 11

### Describe the Bug

My producer can't send messages to my queue in miniflare.

### Observed behavior

I'm testing a basic miniflare configuration with two workers and a queue. One worker is the producer, the other is the consumer.

Trying to send any messages to the queue result in the following error.

> Error: Network connection lost.

It happens when you call the producer's route (which sends a message to the queue)

```typescript
const res = await mf.dispatchFetch('http://localhost:5000/producer')
```

And it also happens when you get the producer directly from the miniflare instance and try to send a message.

```typescript
const queue = await mf.getQueueProducer('CANDIDATES_QUEUE', 'producer')
queue.send({ ding: 'Hello from Miniflare!' })
```
### Expected behavior

I expect the producer to be able to send a message to the queue and return a response.
I expect the consumer to consume the message from the queue and log out what it found.

### Steps to reproduce

run the following code with `npx tsx .`

```typescript
import { Miniflare } from 'miniflare'

const mf = new Miniflare({
  // host: '0.0.0.0',
  port: 5000,

  workers: [
    {
      name: 'producer',
      modules: true,
      routes: ['http://localhost/producer*'],
      script: `
        export default {
          async fetch(req, env, ctx) {
            console.log('fetch')
            await env.CANDIDATES_QUEUE.send({
              testing: 'Hello from Miniflare!',
            })
            return new Response('Producer sent message to queue');
          }
        }
      `,

      queueProducers: {
        CANDIDATES_QUEUE: 'candidates-queue',
      },
    },
    {
      name: 'consumer',
      modules: true,
      routes: ['http://localhost/queue*'],
      script: `
        export default {
          async queue(batch, env) {
            console.log('received ' + batch.messages.length)
            for (let message of batch.messages) {
              console.log(
                'message ' + message.id + ' processed: ' + JSON.stringify(message.body)
              )
            }
          }
        }
      `,

      queueConsumers: ['candidates-queue'],
    },
  ],
})

// this throws the network error
const res = await mf.dispatchFetch('http://localhost:5000/producer')
console.log(res.text())

[...]

PR #6514, opened 2024-08-16
fix(miniflare): Fix regression introduced in #5570

CarmenPopoviciu:

PR #5570 introduced a regression in Miniflare, namely that declaring Queue Producers like 
```
queueProducers: { "MY_QUEUE": "my-queue" }
```
no longer works. This PR restores balance in the Miniflare + Queues Universe 🪐 

Possibly addresses  #5908

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: regression fix

### DIFF
--- a/.changeset/cool-beans-applaud.md
+++ b/.changeset/cool-beans-applaud.md
@@ -1,0 +1,9 @@
+---
+"miniflare": minor
+---
+
+fix: Fix Miniflare regression introduced in #5570
+
+PR #5570 introduced a regression in Miniflare, namely that declaring Queue Producers like `queueProducers: { "MY_QUEUE": "my-queue" }` no longer works. This commit fixes the issue.
+
+Fixes #5908
--- a/.changeset/cool-beans-applaud.md
+++ b/.changeset/cool-beans-applaud.md
@@ -1,5 +1,5 @@
 ---
-"miniflare": minor
+"miniflare": patch
 ---
 
 fix: Fix Miniflare regression introduced in #5570
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -420,6 +420,7 @@ function getQueueProducers(
 		if (workerProducers !== undefined) {
 			// De-sugar array consumer options to record mapping to empty options
 			if (Array.isArray(workerProducers)) {
+				// queueProducers: ["MY_QUEUE"]
 				workerProducers = Object.fromEntries(
 					workerProducers.map((bindingName) => [
 						bindingName,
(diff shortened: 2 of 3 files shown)


threepointone (approved):

packages/miniflare/src/index.ts
```diff
@@ -409,8 +409,22 @@ function getQueueProducers(
 				);
 			}
 
-			for (const [bindingName, opts] of Object.entries(workerProducers)) {
-				queueProducers.set(bindingName, { workerName, ...opts });
+			type Entries<T> = { [K in keyof T]: [K, T[K]] }[keyof T][];
+			type ProducersIterable = Entries<typeof workerProducers>;
+			const producersIterable = Object.entries(
+				workerProducers
+			) as ProducersIterable;
```
petebacondarwin, line 416: I think this can just be:

```suggestion
			const producersIterable = Object.entries<typeof workerProducers>(
				workerProducers
			);
```

petebacondarwin (approved):
Interested whether the Queue team has a view on this?
I seem to remember that they were making a breaking change with this on purpose?

packages/miniflare/src/index.ts
```diff
@@ -409,8 +409,22 @@ function getQueueProducers(
...
+
+			for (const [bindingName, opts] of producersIterable) {
+				if (typeof opts === "string") {
+					// queueProducers: { "MY_QUEUE": "my-queue" }
+					queueProducers.set(bindingName, { workerName, queueName: opts });
+				} else {
+					opts;
+					// queueProducers: ["MY_QUEUE"] or
```
RamIdeas, line 424: This case is handled in the `if (Array.isArray(workerProducers)) {` block further up

CarmenPopoviciu: I can see why the comment is misleading here. Fixed

packages/miniflare/src/index.ts
CarmenPopoviciu: the type of `workersProducers` here is
```
Record<string, string> | Record<string, {
    queueName: string;
    deliveryDelay?: number | undefined;
}>
```

Calling `Object.entries(workersProducers)` will return an array, where each item is an array of type either `[string, string]` or `[string,  {queueName: string; deliveryDelay?: number | undefined;}`.

`Object.entries<typeof workerProducers>(workerProducers)` alone won't give us the typing we need here, as it results in
```
string, Record<string, string> | Record<string, {
    queueName: string;
    deliveryDelay?: number | undefined;
}>][]
```

So we need that extra `Entries` to get the right typing

CarmenPopoviciu:
> Interested whether the Queue team has a view on this?
I've reached out to the queues team and waiting for them to get back to us on this. 

> I seem to remember that they were making a breaking change with this on purpose?
Sure, but given it's a breaking change, shouldn't it then have landed in a `major` version update rather than `minor`. And shouldn't docs have then be updated as well?

packages/miniflare/src/index.ts
```diff
@@ -409,8 +410,21 @@ function getQueueProducers(
...
+			const producersIterable = Object.entries(
+				workerProducers
+			) as ProducersIterable;
+
+			for (const [bindingName, opts] of producersIterable) {
+				if (typeof opts === "string") {
+					// queueProducers: { "MY_QUEUE": "my-queue" }
+					queueProducers.set(bindingName, { workerName, queueName: opts });
+				} else {
+					opts;
```
penalosa, line 424:
```suggestion
```

CarmenPopoviciu: I'm so confused what that line even is supposed to mean. Thx for spotting it 🙏

.changeset/cool-beans-applaud.md
```diff
@@ -0,0 +1,9 @@
+---
+"miniflare": minor
```
edmundhung, line 2: should this be a patch?
```suggestion
"miniflare": patch
```

edmundhung (approved):

PR #6525, opened 2024-08-19
feat(wrangler) add asset manifest in dev and use in ASW

emily-shen:

[WC-2533](https://jira.cfdata.org/browse/WC-2533)
1. generate an asset manifest in miniflare
2. use asset manifest to lookup path/KV key in ASW

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: just needs to pass existing tests?
  - [ ] I don't know
  - [x] Not required because: no relevant e2e test yet
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: incomplete unreleased feature
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: incomplete unreleased feature


packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -20,15 +23,14 @@ export function isWorkersWithAssets(
...
-	assetManifest.set([1], 0); // [1, 0, 0, ..., 0, 0]
-
-	return assetManifest;
-}
+export const buildAssetsManifest = async (dir: string) => {
+	const assetManifest = await walk(dir);
+	const sortedAssetManifest = sortManifest(assetManifest);
+	const encodedAssetManifest = encodeManifest(sortedAssetManifest);
+	// to indicate it is in dev:
+	encodedAssetManifest.set([1], 0);
```
petebacondarwin, line 31: This should not be needed.

packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -97,3 +98,122 @@ export function getAssetsServices(options: AssetsOptions): Service[] {
...
+					throw new Error(
+						`Asset too large.\n` +
+							`Cloudflare Workers supports assets with sizes of up to ${MAX_ASSET_SIZE}. We found a file ${filepath} with a size of ${filestat.size}.\n` +
+							`Ensure all assets in your assets directory "${dir}" conform with the Workers maximum size requirement.`
+					);
+				}
+
+				// not the same as deploy but more similar to what is stored in gcs
+				manifest.push({
+					contentHash: fakeHashFile(),
```
petebacondarwin, line 149: Perhaps just this?

```suggestion
					contentHash: relativeFilepath,
```

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -21,13 +20,17 @@ export default {
...
+		if (!assetEntry) {
+			return new Response("Not Found", { status: 404 });
+		}
 
-		const content = await ASSETS_KV_NAMESPACE.get(assetEntry);
+		const isLocalDevContext = new Uint8Array(ASSETS_MANIFEST).at(0) === 1;
+		const assetKey = isLocalDevContext
+			? assetEntry.path
+			: assetEntry.contentHash;
+		const content = await ASSETS_KV_NAMESPACE.get(assetKey);
```
petebacondarwin, line 33: We should avoid any dev specific code in this Worker.
It should just do the same as in production.

petebacondarwin (changes requested):

packages/workers-shared/asset-server-worker/src/index.ts
emily-shen: Not sure that's possible (unless I'm missing something obvious!)

- We need to pass into `ASSETS_KV_NAMESPACE.get()` the pathname in dev, but the content hash in prod.
- We also want to search the manifest in the same way in prod and dev, so therefore the manifest should be the same shape in prod and dev
- In the production manifest, each content hash gets 16 bytes

If we want the ASW to follow exactly the same code in prod and dev, I think we would have to reversibly encode any filepaths into those 16 bytes. I'm not sure that's possible, so unfortunately I do think we need to have a minor divergence here... :(

petebacondarwin: Urgh!!! So we only have 16 bytes to play with... 😢  That is very sad.

packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -97,3 +98,122 @@ export function getAssetsServices(options: AssetsOptions): Service[] {
 	};
 	return [storageService, namespaceService];
 }
+
+const MAX_ASSET_COUNT = 20_000;
```
CarmenPopoviciu, line 102: could we please extract all the assets manifest related logic in a separate file 🙏 ?

packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -97,3 +98,122 @@ export function getAssetsServices(options: AssetsOptions): Service[] {
...
+	await Promise.all(
+		files.map(async (file) => {
+			const filepath = path.join(dir, file);
+			const relativeFilepath = path.relative(dir, filepath);
+			const filestat = await fs.stat(filepath);
+
+			if (filestat.isSymbolicLink() || filestat.isDirectory()) {
+				return;
+			} else {
+				if (counter >= MAX_ASSET_COUNT) {
```
CarmenPopoviciu, line 131: yeah, I thought about this too. In my original attempt to crack the assets manifest, I passed both `MAX_ASSET_COUNT` and `MAX_ASSET_SIZE` as options, alongside `assetsPath`. The thinking there was that we'd have "one" source of truth, and we'd de-risk running out of sync between wrangler and miniflare. But in the light of us working towards making ASW a Miniflare primitive, I think it's OK for these to be internal to Miniflare. We "just" need to remember we have to update these values in two places, if we ever need to update them

packages/miniflare/src/plugins/kv/assets.ts
emily-shen: Hmm actually I would prefer to keep this all in this file to prevent it getting too fragmented. But I've added some nice big comment headers to segment out the file in a hopefully sensible way.

packages/miniflare/src/plugins/kv/assets.ts
emily-shen: i'll add a comment to the corresponding wrangler bit for posterity.

packages/miniflare/src/plugins/kv/assets.ts
emily-shen: see https://github.com/cloudflare/workers-sdk/pull/6525#discussion_r1722039312

packages/miniflare/src/plugins/kv/assets.ts
emily-shen: see https://github.com/cloudflare/workers-sdk/pull/6525#discussion_r1722039312

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4579,7 +4579,10 @@ addEventListener('fetch', event => {});`
 				{ filePath: "boop/file-3.txt", content: "Content of file-3" },
 				{ filePath: "file-4.txt", content: "Content of file-4" },
 				{ filePath: "beep/file-5.txt", content: "Content of file-5" },
-				{ filePath: "beep/file-6.txt", content: "Content of file-1" },
+				{
+					filePath: "beep/boop/beep/boop/file-6.txt",
+					content: "Content of file-1",
```
GregBrimble, line 4584:
```suggestion
					content: "Content of file-6",
```

packages/workers-shared/asset-server-worker/src/assets-manifest.ts
```diff
@@ -6,6 +12,82 @@ export class AssetsManifest {
...
 }
+
+const hashPath = async (path: string) => {
+	const encoder = new TextEncoder();
+	const data = encoder.encode(path);
+	const hashBuffer = await crypto.subtle.digest("SHA-256", data.buffer);
+	return new Uint8Array(hashBuffer, 0, PATH_HASH_SIZE);
+};
+
+const binarySearch = (
```
GregBrimble, line 32: I always wanted to swap this out for an interpolation search which should be slightly more efficient that binary because our search space is sorted and uniformly distributed, but I never got around to it. Now that it's here and available so easily for testing in `wrangler dev`, it'll be much easier to swap out in the future :)

GregBrimble (approved):

packages/miniflare/src/plugins/kv/assets.ts
CarmenPopoviciu: my understanding from that comment is that we can remove this?

packages/workers-shared/asset-server-worker/src/assets-manifest.ts
emily-shen: if we have time at the end (lol), happy to add that in for you :)

packages/miniflare/src/plugins/kv/index.ts
```diff
@@ -41,6 +41,7 @@ export const KVOptionsSchema = z.object({
 	assetsPath: PathSchema.optional(),
 	assetsKVBindingName: z.string().optional(),
 	assetsManifestBindingName: z.string().optional(),
+	assetsReverseMapBindingName: z.string().optional(),
```
petebacondarwin, line 44: Not needed?

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -1,10 +1,14 @@
 import { SharedBindings } from "miniflare:shared";
-import { KVParams } from "./constants";
 
 interface Env {
 	[SharedBindings.MAYBE_SERVICE_BLOBS]: Fetcher;
+	__STATIC_ASSETS_REVERSE_MAP: ArrayBuffer;
```
petebacondarwin, line 5: There should be no need for this to be an array buffer. It can just be a JSON binding to an object that maps hashes to paths and content type.

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -13,21 +17,26 @@ export default <ExportedHandler<Env>>{
...
+		const decoder = new TextDecoder();
+		const reverseMap: AssetReverseMap = JSON.parse(
+			decoder.decode(env.__STATIC_ASSETS_REVERSE_MAP)
+		);
 
-		if (url.searchParams.get(KVParams.URL_ENCODED)?.toLowerCase() === "true") {
-			key = decodeURIComponent(key);
-		}
+		// don't uri decode pathname, because we encode the filepath before hashing
+		const key = new URL(request.url).pathname.substring(1);
```
petebacondarwin, line 26: Let's be very explicit...

```suggestion
		const pathHash = new URL(request.url).pathname.substring(1);
```

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -13,21 +17,26 @@ export default <ExportedHandler<Env>>{
...
+
+		const entry = reverseMap[key] ?? {};
```
petebacondarwin, line 28:
```suggestion
		const { filePath, contentType } = reverseMap[key] ?? {};
```

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -13,21 +17,26 @@ export default <ExportedHandler<Env>>{
...
+		const filePath = entry["filePath"] ?? "";
+		// falls back to application/octet-stream
+		// const contentType = entry["contentType"];
 
 		const blobsService = env[SharedBindings.MAYBE_SERVICE_BLOBS];
-		if (key === "" || key === "/") {
+		if (filePath === "" || filePath === "/") {
```
petebacondarwin, line 34:
```suggestion
		if (filePath === undefined || filePath === "/") {
```

petebacondarwin: I would expect `""` to be an invalid entry in the reverseMap ...
Also, when would the reverseMap contain `"/"`?

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -13,21 +17,26 @@ export default <ExportedHandler<Env>>{
...
 			return new Response("Not Found", {
 				status: 404,
 			});
```
petebacondarwin, line 37: Actually how about:

```ts
const entry = reverseMap[key];
if ( entry === undefined ) {
  return new Response("No Found", { status: 404 });
}
const { filePath, contentType } = entry;
assert(filePath, "Invalid `filePath` in `reverseMap`");
```

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -21,14 +20,13 @@ export default {
...
-		const content = await ASSETS_KV_NAMESPACE.get(assetEntry);
+		const assetsManifest = new AssetsManifest(ASSETS_MANIFEST);
+		const assetKey = await assetsManifest.get(pathname);
+		if (!assetKey) {
+			return new Response("Not Found", { status: 404 });
+		}
 
+		const content = await ASSETS_KV_NAMESPACE.get(assetKey);
 		if (!content) {
 			return new Response("Not Found", { status: 404 });
```
petebacondarwin, line 31: Could even go so far as to 500 error here, since we should never have an assetKey from the manifest that doesn't already exist in KV, right? @GregBrimble

packages/wrangler/src/experimental-assets.ts
```diff
@@ -39,10 +39,11 @@ type UploadResponse = {
 
 // constants same as Pages for now
 const BULK_UPLOAD_CONCURRENCY = 3;
-const MAX_ASSET_COUNT = 20_000;
-const MAX_ASSET_SIZE = 25 * 1024 * 1024;
 const MAX_UPLOAD_ATTEMPTS = 5;
 const MAX_UPLOAD_GATEWAY_ERRORS = 5;
+// NB also used in miniflare in plugins/kv/assets.ts, so please update there too.
+const MAX_ASSET_COUNT = 20_000;
+const MAX_ASSET_SIZE = 25 * 1024 * 1024;
```
petebacondarwin, line 46: Strawman proposal: export these from `workers-shared` and add that as a devDep here.

packages/wrangler/src/experimental-assets.ts
```diff
@@ -218,32 +219,23 @@ export const syncExperimentalAssets = async (
...
 
-// modified from /pages/validate.tsx
-const walk = async (
-	dir: string,
-	manifest: AssetManifest,
-	startingDir: string = dir
-) => {
-	const files = await readdir(dir);
-
+export const buildAssetsManifest = async (dir: string) => {
```
petebacondarwin, line 222: Also consider exporting this from `workers-shared` so it is kept near the algorithm for reading the manifest.

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,5 +1,4 @@
 import { AssetsManifest } from "./assets-manifest";
-import { NoopAssetsManifest } from "./assets-manifest-no-op";
 
 interface Env {
 	ASSETS_MANIFEST: ArrayBuffer;
```
petebacondarwin, line 4: Consider putting the documentation comments on here (as JSDoc) rather than on the const variable definitions inside the fetch handler.

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -21,14 +20,13 @@ export default {
...
 		const { pathname } = url;
 
-		const isLocalDevContext = new Uint8Array(ASSETS_MANIFEST).at(0) === 1;
-		const assetsManifest = isLocalDevContext
-			? new NoopAssetsManifest()
-			: new AssetsManifest(ASSETS_MANIFEST);
-		const assetEntry = await assetsManifest.get(pathname);
-
-		const content = await ASSETS_KV_NAMESPACE.get(assetEntry);
+		const assetsManifest = new AssetsManifest(ASSETS_MANIFEST);
```
petebacondarwin, line 23: Should `env.ASSETS_MANIFEST` ever be undefined? In the const variable declaration above it is defaulting to `new ArrayBuffer(0)` but I think that is outdated code, since we should "always" be providing a manifest now, right?

packages/miniflare/src/plugins/kv/assets.ts
```diff
@@ -92,8 +99,184 @@ export function getAssetsServices(options: AssetsOptions): Service[] {
...
+								filestat.size,
+								{
+									binary: true,
+								}
+							)}.\n` +
+							`Ensure all assets in your assets directory "${dir}" conform with the Workers maximum size requirement.`
+					);
+				}
+
+				manifest.push(await hashPath(encodeFilePath(relativeFilepath)));
```
petebacondarwin, line 175: A manifest entry consist of a path hash (16 bytes) and a content hash (16 bytes), right?
In the case of Miniflare they are going to be the same thing, since we are not bothering to hash the contents of the file.
So I would expect this to be:

```suggestion
        const pathHash = await hashPath(encodeFilePath(relativeFilepath));
				manifest.push(pathHash, pathHash);
```

petebacondarwin (changes requested):

packages/miniflare/src/plugins/kv/index.ts
CarmenPopoviciu: agreed. This is an internal implementation detail

packages/workers-shared/asset-server-worker/src/index.ts
CarmenPopoviciu: I am handling that [here](https://github.com/cloudflare/workers-sdk/pull/6539/files#diff-67a90fb405dc23a85bdeb2be30130464cbfcc7c742e4f48cab05bee45f8d36dbR42)

packages/miniflare/src/plugins/kv/index.ts
emily-shen: oops I meant to remove that

packages/miniflare/src/plugins/kv/assets.ts
emily-shen: Think I'm doing essentially the same thing when I insert in the pathHash twice while encoding here: https://github.com/cloudflare/workers-sdk/blob/1b9f58546fb7165882c7c6ba551f1e0a8d739397/packages/miniflare/src/plugins/kv/assets.ts#L230

packages/miniflare/src/workers/kv/assets.worker.ts
emily-shen: Yeah I thought that was a bit funny, should've flagged it - before i remove it, @CarmenPopoviciu was there a particular reason this was checking for "/"?

packages/miniflare/src/workers/kv/assets.worker.ts
emily-shen: i don't think it even needs to check for filepath right, since it shouldn't be possible for an entry to exist without a filepath/this should error earlier when constructing the reverse map

packages/miniflare/src/workers/kv/assets.worker.ts
CarmenPopoviciu: this from before any sort of assetManifest impl. days, and is prolly not relevant anymore

packages/wrangler/src/experimental-assets.ts
emily-shen: went a bit overboard with that here... 
https://github.com/cloudflare/workers-sdk/commit/9f3bf38bd769a29b080435a26aa5b638eb8d48aa 
(doesn't build and I think I've gotten confused, but could this work?)

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4579,7 +4579,10 @@ addEventListener('fetch', event => {});`
 				{ filePath: "boop/file-3.txt", content: "Content of file-3" },
 				{ filePath: "file-4.txt", content: "Content of file-4" },
 				{ filePath: "beep/file-5.txt", content: "Content of file-5" },
-				{ filePath: "beep/file-6.txt", content: "Content of file-1" },
+				{
+					filePath: "beep/boop/beep/boop/file-6.txt",
```
petebacondarwin, line 4583: Is the idea here to have a file whose path is longer than 16 bytes?

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -1,10 +1,14 @@
 import { SharedBindings } from "miniflare:shared";
-import { KVParams } from "./constants";
 
 interface Env {
 	[SharedBindings.MAYBE_SERVICE_BLOBS]: Fetcher;
+	__STATIC_ASSETS_REVERSE_MAP: AssetReverseMap;
```
petebacondarwin, line 5: By the way there is no need to use special naming here (e.g. `__` prefix) since this binding is already private to this fake KV namespace service.

petebacondarwin (approved):
Looking good!
I would love to avoid the duplication of the walking algorithm between `walk()` and `getReverseMap()` functions if possible but not a blocker here.

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: Not specifically, I just through I should probably chuck in a more nested test case in there

PR #6526, opened 2024-08-19
defineCommand util

RamIdeas:

Fixes [DEVX-1377](https://jira.cfdata.org/browse/DEVX-1377) [DEVX-1367](https://jira.cfdata.org/browse/DEVX-1367)

This PR introduces a thin abstraction around yargs to improve wrangler's CLI command registration with 3 utilities:

1. `defineCommand`: define an absolute command (`wrangler [..namespace] subcommand`) – satisfying the typescript compiler will ensure all necessary info is provided (description, status, owner, args, handler, etc...)
2. `defineNamespace`: define an absolute command namespace (`wrangler <..namespace>`) – satisfying the typescript compiler will ensure all necessary info is provided (description, status, owner, etc...)
3. `defineAlias`: define an alias from any absolute command to any other absolute command. Optionally, override metadata (eg description, status, deprecated, hidden, etc...).

*Implementation details:*
These 3 core utils are in `src/core/define-commands`. They simply push the definition into a flat array of all definitions.

In `src/core/register-commands`, a tree is built from the flat list and is validated, and exposes utils to register commands with yargs: `register.registerAllCommands`, `register.registerNamespace` (the latter of which is needed to register commands in the same order as the existing implementation).

Before registering the commands, the definitions are wrapped with `src/core/wrap-commands` to provide centralised behaviour and transforms the definition into something more friendly to pass to yargs.

Addressed:
- current contrib experience starts with copy/paste
- leading to inconsistencies
- required knowledge of yargs (not obvious)

Wins:
- ensure subtleties of command generation are handled consistently, centrally
  - subhelp by default
  - deprecated messaging handled
  - status (experimental, alpha, beta, etc) messaging handled
  - easier definition of args + sharing arg definitions (as plain objects)
- less random yargs builder functions + importing them across multiple files
  - removed alias complexity
- less complex yargs generic types
- less naming fatigue + export/import chains
  - move code where it makes sense, not where yargs makes you
- declarative interface with types such that any missing info is caught by type checking
  - PR reviews can focus on implementation and not usage of yargs
- definitions separated from the registration
  - module graph simplified
  - enabling test improvements in future

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: internal change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/versions/deploy.ts
```diff
@@ -44,61 +41,73 @@ const EPSILON = 0.001; // used to avoid floating-point errors. Comparions to a v
...
->;
+export type VersionsDeployArgs = typeof command.args;
 
 type OptionalPercentage = number | null; // null means automatically assign (evenly distribute remaining traffic)
 
-export function versionsDeployOptions(yargs: CommonYargsArgv) {
-	return yargs
-		.option("name", {
+const command = defineCommand({
+	command: "wrangler versions deploy",
```
andyjessop, line 49: Do we have a plan as to how sub-commands are defined? Is defineCommand composable (or could it be)?

RamIdeas: All commands + subcommands are expected to be defined fully and independently. This example is a subcommand itself.

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+        command: "wrangler kv",
+        metadata: {
+            description: "Commands for interacting with Workers KV",
+            status: "stable",
+        },
+    });
+    // Every level of namespaces must be defined
+    // eg "wrangler kv key" in "wrangler kv key put"
+    defineNamespace({
+        command: "wrangler kv",
```
andyjessop, line 32: Should this be `wrangler kv key`?

packages/wrangler/src/__tests__/core/command-registration.test.ts
```diff
@@ -0,0 +1,477 @@
...
+import { mockConsoleMethods } from "../helpers/mock-console";
+import { runInTempDir } from "../helpers/run-in-tmp";
+import { runWrangler } from "../helpers/run-wrangler";
+
+describe("Command Registration", () => {
+	runInTempDir();
+	const std = mockConsoleMethods();
+
+	beforeEach(() => {
+		COMMAND_DEFINITIONS.length = 0; // clears commands definitions so tests do not conflict with eachother
```
edmundhung, line 17: Would it be better to provide a util to clears commands?

RamIdeas: No it's only necessary for testing itself. No other test file should need to do this.

packages/wrangler/CONTRIBUTING.md
penalosa, line 1: Stylistic, I know, but I think this doc should use capital letters at the start of sentences

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+## What you will learn
+
+- How to add a new command to Wrangler
+- How to specify arguments for a command
+- How to use experimental flags
+- How to read from the config
+- How to implement a command
+- How to test a command
+
+## Defining your new command to Wrangler
```
penalosa, line 12:
```suggestion
## Defining your new command in Wrangler
```

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+1. define the command with the util defineCommand
```
penalosa, line 14:
```suggestion
1. Define the command structure with the utils `defineNamespace()` & `defineCommand()`:
```

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+		// implementation here
+	},
+});
+```
+
+2. global vs shared vs command specific (named + positional) args
+
+- Command-specific args are defined in the `args` field of the command definition. Command handlers receive these as a typed object automatically. To make any of these positional, add the key to the `positionalArgs` array.
+- You can share args between commands by declaring a separate object and spreading it into the `args` field. Feel free to import from another file.
+- Global args are shared across all commands and defined in `src/commands/global-args.ts` (same schema as command-specific args). They are passed to every command handler.
```
penalosa, line 75:
```suggestion
2. global vs shared vs command specific (named + positional) args

- Command-specific args are defined in the `args` field of the command definition. Command handlers receive these as a typed object automatically. To make any of these positional, add the key to the `positionalArgs` array.
- You can share args between commands by declaring a separate object and spreading it into the `args` field. Feel free to import from another file.
- Global args are shared across all commands and defined in `src/commands/global-args.ts` (same schema as command-specific args). They are available in every command handler.
```

I think these lists should probably be in the same order (the `vs` list and the bullet list)

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+3. (get a type for the args)
```
penalosa, line 77: Why the brackets?

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+await fetchResult(
+	`/accounts/${accountId}/workers/services/${scriptName}`,
+	{ method: "DELETE" },
+	new URLSearchParams({ force: needsForceDelete.toString() })
+);
+```
+
+- logging
+
+Do not use `console.*` methods to log. You must import and use the `logger` singleton.
```
penalosa, line 99: This should probably link to the file to import to access the logger singleton

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+You may want to pass your args to other functions. These functions will need to be typed. To get a type of your args, you can use `typeof command.args`.
+
+4. implement the command handler
+
+A command handler is just a function that receives the args as the first param. This is where you will want to do API calls, I/O, logging, etc.
+
+- api calls
+
+Define API response type. Use `fetchResult` to make API calls. `fetchResult` will throw an error if the response is not 2xx.
```
penalosa, line 87: This should link to where to import `fetchResult()` from, and should mention that it makes authed requests to the Cloudflare API.

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+	{ method: "DELETE" },
+	new URLSearchParams({ force: needsForceDelete.toString() })
+);
+```
+
+- logging
+
+Do not use `console.*` methods to log. You must import and use the `logger` singleton.
+
+- error handling - UserError vs Error
```
penalosa, line 101: This should have an example, and a link to where `UserError` is defined

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+Throw `UserError` for errors _caused_ by the user -- these are not sent to Sentry whereas regular `Error` are and show be used for unexpected exceptions.
```
penalosa, line 103: An example of what types of errors are caused by users would be helpful (i.e. D1 crashed vs invalid SQL statement)

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+Errors are caught at the top-level and formatted for the console.
```
penalosa, line 105: Could you include a screenshot, maybe?

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Best Practices
+
+### Status / Deprecation
+
+Status can be alpha, private-beta, open-beta, or stable. Breaking changes can freely be made in alpha or private-beta. Try avoid breaking changes in open-beta but are acceptable and should be called out in changeset.
```
penalosa, line 111: This should link/reference the changesets guidelines around breaking changes for non stable products (`patch` bump)

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+Stable commands should never have breaking changes.
+
+### Changesets
+
+Run `npx changesets` from the top of the repo. New commands warrant a "minor" bump. Please explain the functionality with examples.
```
penalosa, line 117: This should link/reference the changesets guidelines in the top-level contributing guide (i.e. `feat:` etc...). It might also be nice to include an example of a good changeset for adding a new command

packages/wrangler/CONTRIBUTING.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Experimental Flags
+
+If you have a stable command, new features should be added behind an experimental flag. By convention, these are named `--experimental-<feature-name>` and have an alias `--x-<feature-name>`. These should be boolean, defaulting to false (off by default).
+
+To stabilise a feature, flip the default to true while keeping the flag to allow users to disable the feature.
+
+After a bedding period, you can mark the flag as deprecated and hidden. And remove all code paths using the flag.
```
penalosa, line 125: Is the phrase "bedding period" common?
```suggestion
After a bedding period, you can mark the flag as deprecated and hidden, and remove all code paths using the flag.
```

packages/wrangler/src/core/register-commands.ts
```diff
@@ -0,0 +1,236 @@
...
+		node.definition = def;
+
+		if ("aliasOf" in def) {
+			aliases.add(def);
+		}
+	}
+
+	// STEP 2: Resolve all aliases to their real definitions
+
+	const MAX_HOPS = 5; // reloop to allow aliases of aliases (to avoid infinite loop, limit to 5 hops)
```
penalosa, line 94: Do we have any aliases of aliases in the codebase today? That seems like a DX anti-pattern we'd want to avoid

packages/wrangler/src/core/register-commands.ts
```diff
@@ -0,0 +1,236 @@
...
+function createCommandTree() {
+	const root: DefinitionTreeNode = { subtree: new Map() };
+	const aliases = new Set<AliasDefinition>();
+
+	// STEP 1: Create tree from flat definitions array
+
+	for (const def of COMMAND_DEFINITIONS) {
+		const node = createNodeFor(def.command, root);
+
+		if (node.definition) {
```
penalosa, line 80: How will `node.definition` be defined if it's just been created with `createNodeFor()`?

packages/wrangler/src/core/register-commands.ts
```diff
@@ -0,0 +1,236 @@
...
+		throw new CommandRegistrationError(
+			`Missing namespace definition for '${segment}'`
+		);
+	}
+
+	// convert our definition into something we can pass to yargs.command
+	const def = wrapCommandDefinition(definition);
+
+	// register command
+	yargs.command(
```
penalosa, line 161:
```suggestion
	return yargs.command(
```

packages/wrangler/src/core/wrap-command.ts
```diff
@@ -0,0 +1,117 @@
...
+	BaseNamedArgDefinitions,
+	CommandDefinition,
+	HandlerArgs,
+	NamespaceDefinition,
+} from "./define-command";
+
+const betaCmdColor = "#BD5B08";
+
+export function wrapCommandDefinition(
+	def: CommandDefinition | NamespaceDefinition
```
penalosa, line 17: I'm a bit confused by the naming here. Why does a function called `wrapCommandDefinition()` take a `NamespaceDefinition`?

packages/wrangler/src/core/wrap-command.ts
```diff
@@ -0,0 +1,117 @@
...
+		| undefined
+		| ((args: HandlerArgs<BaseNamedArgDefinitions>) => Promise<void>) =
+		undefined;
+
+	if (def.metadata.status !== "stable") {
+		description += chalk.hex(betaCmdColor)(` [${def.metadata.status}]`);
+
+		statusMessage =
+			def.metadata.statusMessage ??
+			`🚧 \`${def.command}\` is a ${def.metadata.status} command. Please report any issues to https://github.com/cloudflare/workers-sdk/issues/new/choose`;
```
penalosa, line 38: Nit, but `a` won't always be appropriate here (e.g. `is a alpha command`)

packages/wrangler/src/core/register-commands.ts
RamIdeas: The node has been created, or returned if already exists. The definition is undefined it it was just created.

packages/wrangler/src/core/register-commands.ts
RamIdeas: Doesn't seem like an anti-pattern to me. Devs shouldn't need to consider whether another command is an alias to alias to it. 5 seemed a reasonable limit to allow aliases to **any** command but no harm in increasing higher even.

packages/wrangler/src/core/define-command.ts
```diff
@@ -0,0 +1,163 @@
...
+export function defineCommand<NamedArgs extends BaseNamedArgDefinitions>(
+	definition: CommandDefinition<NamedArgs>
+) {
+	COMMAND_DEFINITIONS.push(definition as unknown as CommandDefinition);
+
+	return {
+		definition,
+		get args(): HandlerArgs<NamedArgs> {
+			throw new Error();
+		},
```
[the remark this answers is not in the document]
RamIdeas: To allow `typeof command.args` but open to changing to a generic like `ArgsOf<typeof command>`

packages/wrangler/src/core/define-command.ts
```diff
@@ -0,0 +1,163 @@
...
+	 */
+	handler: (
+		args: HandlerArgs<NamedArgs>,
+		ctx: HandlerContext
+	) => void | Promise<void>;
+};
+
+export const COMMAND_DEFINITIONS: Array<
+	CommandDefinition | NamespaceDefinition | AliasDefinition
+> = [];
```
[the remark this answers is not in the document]
RamIdeas: Yeah I did and it seemed like naming fatigue and lots of import/export for not much gain. Open to changing in the future but if they're only used in one place (`main()`) then this pattern makes sense.

packages/wrangler/src/core/define-command.ts
```diff
@@ -0,0 +1,163 @@
...
+		 */
+		sendMetrics?: boolean;
+
+		/**
+		 * By default, wrangler's version banner will be printed before the handler is executed.
+		 * Set this value to `false` to skip printing the banner.
+		 *
+		 * @default true
+		 */
+		printBanner?: boolean;
```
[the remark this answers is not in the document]
RamIdeas: One of the KV commands prints json unconditionally and a bunch of other commands don't print the banner (for reasons we can address after the refactor, now that it's obvious). So I added a way to turn off the banner. The abstraction shouldn't impede anyone – but divergence from the normal path should be obvious for us to notice in PRs and/or fix later.

packages/wrangler/src/core/define-command.ts
```diff
@@ -0,0 +1,163 @@
...
+	deprecated?: boolean;
+	deprecatedMessage?: string;
+	hidden?: boolean;
+	owner: Teams;
+};
+
+export type ArgDefinition = PositionalOptions & Pick<Options, "hidden">;
+export type BaseNamedArgDefinitions = { [key: string]: ArgDefinition };
+type StringKeyOf<T> = Extract<keyof T, string>;
+export type HandlerArgs<Args extends BaseNamedArgDefinitions> = OnlyCamelCase<
```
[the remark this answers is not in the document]
RamIdeas: I want to replace all usage of that helper. These utils will allow defining args and using the types (eg `typeof command.args` or otherwise)

packages/wrangler/src/core/define-command.ts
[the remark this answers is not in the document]
RamIdeas: I'll comment on this

packages/wrangler/src/index.ts
```diff
@@ -522,9 +524,7 @@ export function createCLIParser(argv: string[]) {
 
 	/******************** CMD GROUP ***********************/
 	// kv
-	wrangler.command("kv", `🗂️  Manage Workers KV Namespaces`, (kvYargs) => {
-		return registerKvSubcommands(kvYargs, subHelp);
-	});
+	register.registerNamespace("kv");
```
andyjessop, line 527: Would this be better named `commands`?:

```
commands.registerNamespace("kv");
```

is more descriptive.

packages/wrangler/src/core/define-command.ts
```diff
@@ -0,0 +1,165 @@
...
+	}>;
+export function defineCommand<NamedArgs extends BaseNamedArgDefinitions>(
+	definition: CommandDefinition<NamedArgs>
+): DefineCommandResult<NamedArgs>;
+export function defineCommand(
+	definition: CommandDefinition
+): DefineCommandResult<BaseNamedArgDefinitions> {
+	COMMAND_DEFINITIONS.push(definition as unknown as CommandDefinition);
+
+	// @ts-ignore return type is used for type inference only
```
RamIdeas, line 146:
```suggestion
	// @ts-expect-error return type is used for type inference only
```

packages/wrangler/src/core/define-command.ts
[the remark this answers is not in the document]
andyjessop: I was wondering this too. What if they're used in multiple places, like in separate isolated packages?

To avoid any potential issues with shared state, and to make it more portable, we could create an index file here and export functions with closed state: 

```
export const { defineCommand, defineNamespace, defineAlias } = createCommandDefineFns(); // terrible name...
```

This could then be created in each package where we want to isolate this behaviour, or could be used in the main wrangler package just as we're doing here in `kv`.

packages/wrangler/src/core/register-commands.ts
```diff
@@ -0,0 +1,227 @@
...
+}
+
+type DefinitionTreeNode = {
+	definition?: CommandDefinition | NamespaceDefinition | AliasDefinition;
+	subtree: DefinitionTree;
+};
+type DefinitionTree = Map<string, DefinitionTreeNode>;
+
+type ResolvedDefinitionTreeNode = {
+	definition?: CommandDefinition | NamespaceDefinition;
```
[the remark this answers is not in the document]
RamIdeas: aliases are now resolved during registration, so this type is no longer necessary

packages/wrangler/src/core/register-commands.ts
```diff
@@ -0,0 +1,227 @@
...
+
+		if ("aliasOf" in def) {
+			aliases.add(def);
+		}
+	}
+
+	// STEP 2: Resolve all aliases to their real definitions
+
+	const MAX_HOPS = 5; // reloop to allow aliases of aliases (to avoid infinite loop, limit to 5 hops)
+	for (let hops = 0; hops < MAX_HOPS && aliases.size > 0; hops++) {
```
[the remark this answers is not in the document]
RamIdeas: circular refs are now detected and clearly messaged

packages/wrangler/src/core/register-commands.ts
```diff
@@ -0,0 +1,227 @@
...
+		for (const def of aliases) {
+			const realNode = findNodeFor(def.aliasOf, root);
+			const real = realNode?.definition;
+			if (!real || "aliasOf" in real) {
```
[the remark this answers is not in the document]
RamIdeas: this was giving a chance for the aliases of aliases to be resolved but after refactoring the whole chain of aliases are resolved

andyjessop (approved):
This is great work. I love the way that commands are searchable by their string value, and the args validation is extracted from the handler.

The only thing missing here is tests. The code itself is not too complex, but it is difficult to grok without reading through studiously. Tests would shine a light on the API for any devs coming to this in the future. Perhaps this could be addressed in a separate PR.

RamIdeas:
rebased to address conflict

PR #6539, opened 2024-08-20
feat(workers-shared): Add Asset Server Worker behaviour

CarmenPopoviciu:

[Workers + Assets]

This PR adds a proper-ish implementation of a **very** un-opinionated Asset Server Worker. Proper-ish because it handles the essential things needed for ASW to work, but I expect some other details we'll be needed as Workers + Assets work progresses further.

Fixes WC-2434

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because: no e2e tests capture ASW2
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: we're not there yet


packages/workers-shared/asset-server-worker/src/utils/headers.ts
```diff
@@ -0,0 +1,54 @@
...
+	assetMetadata: AssetMetadata | null,
+	request: Request
+) {
+	let contentType = assetMetadata?.contentType ?? "application/octet-stream";
+	if (contentType.startsWith("text/") && !contentType.includes("charset")) {
+		contentType = `${contentType}; charset=utf-8`;
+	}
+
+	const headers = new Headers({
+		"Access-Control-Allow-Origin": "*",
```
GregBrimble, line 38: As discussed, these may go.

packages/workers-shared/asset-server-worker/src/global.d.ts
```diff
@@ -0,0 +1,32 @@
...
+	ENVIRONMENT: Environment;
+};
+
+type BodyEncoding = "manual" | "automatic";
+
+interface ResponseInit {
+	encodeBody?: BodyEncoding;
+}
+
+interface KVNamespace {
```
GregBrimble, line 15: Why can't we use `@cloudflare/workers-types` here? That would save manually typing these things and potentially making mistakes/getting out of sync.

GregBrimble (approved):
Approved with additional small pieces here: https://github.com/cloudflare/workers-sdk/pull/6542

packages/workers-shared/asset-server-worker/src/global.d.ts
CarmenPopoviciu: you are totes right! removed!

CarmenPopoviciu:
@GregBrimble thanks for reviewing this. I've incorporated the changes you suggested in [this commit](https://github.com/cloudflare/workers-sdk/pull/6539/commits/db338a248671adf7cbbdcb865924fb73b3a46d82)

packages/workers-shared/asset-server-worker/src/global.d.ts
```diff
@@ -0,0 +1,13 @@
...
+type Env = {
+	// ASSETS_MANIFEST is a pipeline binding to an ArrayBuffer containing the
+	// binary-encoded site manifest
+	ASSETS_MANIFEST: ArrayBuffer;
+
+	// ASSETS_KV_NAMESPACE is a pipeline binding to the KV namespace that the
+	// assets are in.
+	ASSETS_KV_NAMESPACE: KVNamespace;
+
+	ENVIRONMENT: Environment;
```
petebacondarwin, line 12: Not needed?

packages/workers-shared/asset-server-worker/wrangler.toml
```diff
@@ -7,6 +7,6 @@
 # to this file are persisted in wrangler as well, when necessary.
 # (see packages/wrangler/src/dev/miniflare.ts -> buildMiniflareOptions())
 ##
-name = "asset-server"
+name = "asset-server-worker"
```
petebacondarwin, line 10:
```suggestion
name = "asset-worker"
```

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -411,7 +411,9 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 	// Setup service bindings to external services
 	const serviceBindings: NonNullable<WorkerOptions["serviceBindings"]> = {
 		...config.serviceBindings,
-		...(config.experimentalAssets ? { ASSET_SERVER: "asset-server" } : {}),
+		...(config.experimentalAssets
+			? { ASSET_SERVER: "asset-server-worker" }
```
petebacondarwin, line 415:
```suggestion
			? { ASSET_SERVER: "asset-worker" }
```

petebacondarwin: Can we stabilize on the same name everywhere `asset-worker` (and also change the directory name where it lives).

packages/workers-shared/package.json
```diff
@@ -23,7 +23,7 @@
 	],
 	"scripts": {
 		"build": "pnpm run clean && pnpm run bundle:asset-server:prod",
-		"bundle:asset-server": "esbuild asset-server-worker/src/index.ts --format=esm --bundle --outfile=dist/asset-server-worker.mjs --sourcemap=external",
+		"bundle:asset-server": "esbuild asset-server-worker/src/index.ts --format=esm --bundle --outfile=dist/asset-server-worker.mjs --sourcemap=external --external:cloudflare:*",
```
petebacondarwin, line 26: I wonder if we could use `wrangler deploy --dry-run --outdir=dist/asset-worker.mjs` to do this so that we don't have to keep the esbuild settings in sync?

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,38 +1,66 @@
...
+import { getAdditionalHeaders, getMergedHeaders } from "./utils/headers";
+import { getAssetWithMetadataFromKV } from "./utils/kv";
 
-interface Env {
-	ASSETS_MANIFEST: ArrayBuffer;
-	ASSETS_KV_NAMESPACE: KVNamespace;
-}
+export default class extends WorkerEntrypoint<Env> {
+	async fetch(request: Request) {
+		if (!request.method.match(/^(get)$/i)) {
```
petebacondarwin, line 14: This seems a bit complicated... why not:

```suggestion
		if (request.method.toLowerCase() !== "get") {
```

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,38 +1,66 @@
...
+		}
+	}
 
-			// ASSETS_KV_NAMESPACE is a pipeline binding to the KV namespace that the
-			// assets are in.
-			ASSETS_KV_NAMESPACE,
-		} = env;
+	async hasResponseForRequest(request: Request) {
+		return !!(await this.getAssetEntry(request));
+	}
```
petebacondarwin, line 27: Still not keen on this.

petebacondarwin: How about we leave it out now and add it when it becomes necessary?

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,38 +1,66 @@
...
-		const isLocalDevContext = new Uint8Array(ASSETS_MANIFEST).at(0) === 1;
-		const assetsManifest = isLocalDevContext
-			? new NoopAssetsManifest()
-			: new AssetsManifest(ASSETS_MANIFEST);
-		const assetEntry = await assetsManifest.get(pathname);
+		const assetResponse = await getAssetWithMetadataFromKV(
+			this.env.ASSETS_KV_NAMESPACE,
+			assetEntry
+		);
+		console.log(assetResponse);
```
petebacondarwin, line 39: Remove?

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,38 +1,66 @@
...
+		if (!assetResponse || !assetResponse.value) {
```
petebacondarwin, line 40: It feels like `assetResponse.value` being falsy is a programming error on our part?

packages/workers-shared/asset-server-worker/src/responses.ts
```diff
@@ -0,0 +1,52 @@
...
+	}
+}
+
+export class InternalServerErrorResponse extends Response {
+	constructor(err: Error, init?: ResponseInit) {
+		let body: string | undefined = undefined;
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		if ((globalThis as any).DEBUG) {
+			body = `${err.message}\n\n${err.stack}`;
+		}
```
petebacondarwin, line 36: How would this DEBUG be set? Via an esbuild `define`?
Would this appear in production?

petebacondarwin (commented):
Looking good. I left a bunch of comments.
I'd really really like to see some tests of this worker.

packages/workers-shared/asset-server-worker/src/global.d.ts
CarmenPopoviciu: indeed. I meant to remove it and got distracted. will remove

packages/workers-shared/asset-server-worker/src/responses.ts
CarmenPopoviciu: ah yes, good catch. Removing for now

fixtures/workers-with-assets/tests/index.test.ts
```diff
@@ -32,7 +32,20 @@ describe("[Workers + Assets] `wrangler dev`", () => {
 	it("should not resolve '/' to '/index.html' ", async ({ expect }) => {
 		let response = await fetch(`http://${ip}:${port}/`);
 		let text = await response.text();
-		expect(response.status).toBe(404);
-		expect(text).toContain("Not Found");
+		expect(response.status).toBe(500);
+		expect(text).toContain(
+			"Requested asset &#x2F; exists in the asset manifest but not in the KV namespace."
+		);
```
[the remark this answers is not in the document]
CarmenPopoviciu: My understanding was that if the key is in the manifest but KV returns nul, smth must have gone wrong somewhere, which is why we return 500

packages/workers-shared/asset-server-worker/src/index.ts
CarmenPopoviciu: removed

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: yes, I agree! I will do that in a separate PR and link here

fixtures/workers-with-assets/tests/index.test.ts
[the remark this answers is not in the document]
CarmenPopoviciu: actually...nvm me...I was just being silly. You were right, this does return a 404

packages/miniflare/src/workers/kv/assets.worker.ts
```diff
@@ -27,6 +27,15 @@ export default <ExportedHandler<Env>>{
...
 		const { filePath, contentType } = entry;
 		const blobsService = env[SharedBindings.MAYBE_SERVICE_BLOBS];
-		return blobsService.fetch(new URL(filePath, "http://placeholder"));
+		const response = await blobsService.fetch(
+			new URL(filePath, "http://placeholder")
+		);
+		const newResponse = new Response(response.body, response);
+		// ensure the runtime will return the metadata we need
+		newResponse.headers.append(
+			"cf-kv-metadata",
```
CarmenPopoviciu, line 36: see https://github.com/cloudflare/workerd/blob/a20cb7eb98309b0cd38be1c294dcdf894308a2d1/src/workerd/api/kv.c%2B%2B#L218-L220

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,38 +1,62 @@
...
-	 */
-	ASSETS_KV_NAMESPACE: KVNamespace;
-}
-
-export default {
-	async fetch(request: Request, env: Env) {
-		const { ASSETS_MANIFEST, ASSETS_KV_NAMESPACE } = env;
+export default class extends WorkerEntrypoint<Env> {
+	async fetch(request: Request) {
+		if (request.method.toLowerCase() !== "get") {
```
CarmenPopoviciu, line 14: todo @CarmenPopoviciu add fixture test for this

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: tracking in WC-2595

packages/workers-shared/asset-server-worker/wrangler.toml
CarmenPopoviciu: tracking in WC-2595

packages/workers-shared/tsconfig.json
```diff
@@ -5,8 +5,8 @@
 		"sourceMap": true,
 		"forceConsistentCasingInFileNames": true,
 		"useUnknownInCatchVariables": false,
-		"types": ["@cloudflare/workers-types"]
+		"types": ["@cloudflare/workers-types/experimental", "vitest/globals"]
```
CarmenPopoviciu, line 8: does this look reasonable? alternatively I could have a separate `tsconfig.json` for just the test files, but that felt unnecessary?

packages/workers-shared/vitest.config.mts
```diff
@@ -0,0 +1,12 @@
+import { defineProject, mergeConfig } from "vitest/config";
```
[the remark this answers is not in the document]
CarmenPopoviciu: added [here](https://github.com/cloudflare/workers-sdk/pull/6539/commits/98ab8e48594cd6de89127e95794b061cd0621380)

packages/workers-shared/asset-server-worker/src/index.ts
CarmenPopoviciu: added [here](https://github.com/cloudflare/workers-sdk/pull/6539/commits/98ab8e48594cd6de89127e95794b061cd0621380)

packages/workers-shared/asset-server-worker/src/index.ts
CarmenPopoviciu: technically this should be just a sanity check, but felt like we should check nonetheless. I'll ask the EWC folks if we want to throw a different error here

packages/workers-shared/tsconfig.json
Skye-31: It's usually good practice to have different tsconfigs for directories which use different types - you'll see it in the fixtures directory a lot, for example

CarmenPopoviciu: kk, thank you for that input <3! I'll adjust

andyjessop: I think the way this is donw in `wrangler` is to have both the Vitest config and a separate `tsconfig.json` inside the tests folder. I don't much like having a separate tests folder, but one major benefit is that you can isolate all the test-related config.

CarmenPopoviciu: wrangler has a separate `tsconfig.json` file inside the `__tests__` folder, but `vitest.config.mts` lives in the root. `workers-shared` will most likely share the vitest config between the AssetWorker and RouterWorker, in which case I guess it makes sense for it to live in the root?

packages/workers-shared/asset-server-worker/src/responses.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export class NotModifiedResponse extends Response {
+	constructor(...[_body, _init]: ConstructorParameters<typeof Response>) {
+		super(undefined, {
+			status: 304,
+			statusText: "Not Modified",
+		});
+	}
+}
```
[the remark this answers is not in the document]
CarmenPopoviciu: if RW needs to make use of these then yes, I abs agree! We just haven't gotten there yet

packages/workers-shared/tsconfig.json
CarmenPopoviciu: I went ahead and created a separate `tsconfig.json` file inside the tests folder, which includes the types relevant for tests, and removed `vitest/globals` from the root `tsconfig.json`. I did keep vitest.config in root though for the reasons I explained above. Are we 👌  with that?

CarmenPopoviciu: here we go: https://github.com/cloudflare/workers-sdk/pull/6539/commits/1154d6399a0a989c010fd4df1402ea2b1244e1ce

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -1,38 +1,62 @@
...
+
+		return new OkResponse(assetContent, { headers });
+	}
+
+	private async getAssetEntry(request: Request) {
+		const url = new URL(request.url);
+		let { pathname } = url;
+
+		const assetsManifest = new AssetsManifest(this.env.ASSETS_MANIFEST);
+		pathname = globalThis.decodeURIComponent(pathname);
```
emily-shen, line 58: not sure if this should be decoded because the assetManifest uses encoded pathnames?
Might be worth adding a test for url encoding with a funky pathname. Not high priority though.

packages/workers-shared/asset-server-worker/src/responses.ts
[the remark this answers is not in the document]
emily-shen: Will absolutely be doing this in a refactor PR incoming :)

andyjessop (approved):
Happy with this - very nice work!

packages/workers-shared/asset-server-worker/src/assets-manifest.ts
```diff
@@ -4,6 +4,11 @@ const CONTENT_HASH_SIZE = 16;
 const TAIL_SIZE = 8;
 const ENTRY_SIZE = PATH_HASH_SIZE + CONTENT_HASH_SIZE + TAIL_SIZE;
 
+export type AssetEntry = {
```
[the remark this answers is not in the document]
CarmenPopoviciu: oh right. must be some leftover from all the iterations 😅 .removing

packages/workers-shared/asset-server-worker/src/index.ts
CarmenPopoviciu: ok, so funky path names def don't work, with or without this decoding. Let's have a look together, and I'll add the proper fix/tests in a follow-up PR. Tracking in WC-2597

PR #6542, opened 2024-08-21
Configure GitHub Actions to deploy Asset Worker

GregBrimble:

Configure GitHub Actions to deploy Asset Worker

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: GH action change
  - [ ] I don't know
  - [x] Not required because: GH action change
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: internal facing change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: internal facing change

### DIFF
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -64,6 +64,7 @@ jobs:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           WORKERS_NEW_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.WORKERS_NEW_CLOUDFLARE_ACCOUNT_ID }}
           WORKERS_NEW_CLOUDFLARE_API_TOKEN: ${{ secrets.WORKERS_NEW_CLOUDFLARE_API_TOKEN }}
+          WORKERS_DEPLOY_AND_CONFIG_CLOUDFLARE_API_TOKEN: ${{ secrets.WORKERS_DEPLOY_AND_CONFIG_CLOUDFLARE_API_TOKEN }}
 
       - name: Create C3 Diffs PR
         if: contains(steps.changesets.outputs.publishedPackages, '"create-cloudflare"')
--- a/packages/workers-shared/package.json
+++ b/packages/workers-shared/package.json
@@ -31,6 +31,8 @@
 		"check:type": "pnpm run check:type:tests && tsc",
 		"check:type:tests": "tsc -p ./asset-worker/tests/tsconfig.json",
 		"clean": "rimraf dist",
+		"deploy": "pnpm run deploy:asset-worker",
+		"deploy:asset-worker": "CLOUDFLARE_API_TOKEN=$WORKERS_NEW_CLOUDFLARE_API_TOKEN wrangler versions upload --experimental-versions -c asset-worker/wrangler.toml",
 		"dev": "pnpm run clean && concurrently -n bundle:asset-worker,bundle:router-worker -c blue,magenta \"pnpm run bundle:asset-worker --watch\" \"pnpm run bundle:router-worker --watch\"",
 		"test": "vitest",
 		"test:ci": "pnpm run test run"
--- a/packages/workers-shared/asset-worker/README.md
+++ b/packages/workers-shared/asset-worker/README.md
@@ -1,3 +1,3 @@
 # `asset-worker`
 
-The Asset Worker is a [Cloudflare Worker](https://developers.cloudflare.com/workers/) that is responsible of serving assets for Workers deployed on the edge, that contain static assets as well.
+The Asset Worker is a [Cloudflare Worker](https://developers.cloudflare.com/workers/) that is responsible of serving static assets. It is deployed in production and is also used in local `wrangler dev`.
--- a/packages/workers-shared/asset-worker/wrangler.toml
+++ b/packages/workers-shared/asset-worker/wrangler.toml
@@ -8,5 +8,7 @@
 # (see packages/wrangler/src/dev/miniflare.ts -> buildMiniflareOptions())
 ##
 name = "asset-worker"
+account_id = "0f1b8aa119a907021f659042f95ea9ba"
+workers_dev = false
 main = "src/index.ts"
 compatibility_date = "2024-07-31"


packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -60,6 +65,6 @@ export default {
 		);
 		const headers = getMergedHeaders(request.headers, additionalHeaders);
 
-		return new OkResponse(assetContent, { headers, encodeBody: "automatic" });
-	},
-};
+		return new OkResponse(assetContent, { headers });
```
GregBrimble, line 68: `encodeBody` is `"automatic"` by default, so don't think we need to specify it here.

packages/workers-shared/asset-server-worker/src/index.ts
```diff
@@ -8,48 +9,52 @@ import {
...
-		const {
-			// ASSETS_MANIFEST is a pipeline binding to an ArrayBuffer containing the
-			// binary-encoded site manifest
-			ASSETS_MANIFEST = new ArrayBuffer(0),
-
-			// ASSETS_KV_NAMESPACE is a pipeline binding to the KV namespace that the
-			// assets are in.
-			ASSETS_KV_NAMESPACE,
-		} = env;
+	async hasResponseForRequest(request: Request) {
```
GregBrimble, line 25: `router-worker` will be able to ask `env.assets.hasResponseForRequest(request)` if we want to now, rather than doing the `response.statusCode != 404` sort of thing.

packages/workers-shared/asset-server-worker/src/utils/kv.ts
```diff
@@ -18,7 +18,7 @@ export async function getAssetWithMetadataFromKV(
 		} catch (err) {
 			if (attempts >= retries) {
 				throw new Error(
-					`Requested asset ${assetKey} could not be found in KV namespace.`
+					`Requested asset ${assetKey} could not be fetched from KV namespace.`
```
GregBrimble, line 21: I _think_, (but I haven't actually tested), that `getWithMetadata` would only throw in the case of an error in fetching, rather than a non-existent record?

CarmenPopoviciu: that's a good callout! Docs were a bit scarce on this sort of details. I'll try to test and get some clarity

packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+		try {
+			return this.handleRequest(request);
+		} catch (err) {
+			return new InternalServerErrorResponse(err);
+		}
+	}
+
+	async hasResponseForRequest(request: Request) {
+		return !!(await this.getAssetEntry(request));
+	}
```
petebacondarwin, line 27: This seems like it's adding a premature complication.
The requirements for the second milestone does not require that the Router Worker makes two requests to the Asset Worker.
The logic is that the Router Worker tries to get the asset from the Asset Worker. If it is there then it returns it, otherwise it tries the User Worker. The simplest pseudo code would be:

**router-worker/index.ts**
```ts
export default {
  async fetch(request: Request, env: Env) {
    const assetResponse = await env.ASSETS.fetch(request);
    if (assetResponse.status === 200) {
      return assetResponse;
    } else {
      return env.USER.fetch(request);
    }
  }
}
```

In this case the Asset Worker does not need to be a `WorkerEntrypoint` and it doesn't need to have this extra public method.

We should be optimising for simplicity rather than optimistic future proofing.
Converting the Asset Worker to have this architecture is easy enough to add in the future if that is needed; and if it is not needed we won't have added unnecessary complexity.

petebacondarwin (changes requested):

packages/workers-shared/asset-worker/src/index.ts
GregBrimble: I won't die on this hill, but I'd personally disagree. Yes, it's not strictly required, but this is just three extra lines and I'd argue actually simplifies the contract to something explicit rather than implicit magic based on status codes.

petebacondarwin: But the contract does not involve even asking the question "does this asset exist?" since the Router Worker will always respond with the asset if it exists. So this function does not need to be called.

Using standard HTTP response codes is not really magic. It is a well defined and well understood contract.

I would argue that adding in secondary entry-points means that the maintainer needs to understand how JSRPC works, which adds mental overhead. If we do want to go all in on JSRPC then we should in fact drop the standard `fetch()` function and rename it something like `getAssetForRequest()`.

CarmenPopoviciu (approved):

edmundhung (approved):

.github/workflows/changesets.yml
```diff
@@ -64,6 +64,7 @@ jobs:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           WORKERS_NEW_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.WORKERS_NEW_CLOUDFLARE_ACCOUNT_ID }}
           WORKERS_NEW_CLOUDFLARE_API_TOKEN: ${{ secrets.WORKERS_NEW_CLOUDFLARE_API_TOKEN }}
+          WORKERS_DEPLOY_AND_CONFIG_CLOUDFLARE_API_TOKEN: ${{ secrets.WORKERS_DEPLOY_AND_CONFIG_CLOUDFLARE_API_TOKEN }}
```
petebacondarwin, line 67: Have we now added this secret to GH?

petebacondarwin (approved):

packages/workers-shared/package.json
```diff
@@ -31,6 +31,8 @@
 		"check:type": "pnpm run check:type:tests && tsc",
 		"check:type:tests": "tsc -p ./asset-worker/tests/tsconfig.json",
 		"clean": "rimraf dist",
+		"deploy": "pnpm run deploy:asset-worker",
+		"deploy:asset-worker": "CLOUDFLARE_API_TOKEN=$WORKERS_NEW_CLOUDFLARE_API_TOKEN wrangler versions upload --experimental-versions -c asset-server-worker/wrangler.toml",
```
penalosa, line 35: I don't think `assets-server-worker` is the right path--isn't it just `asset-worker`?

penalosa (changes requested):

packages/workers-shared/package.json
CarmenPopoviciu: ah, such a good catch! I totes missed that old reference

packages/workers-shared/asset-worker/src/index.ts
CarmenPopoviciu: I am going to resolve this conversation since it is not relevant to this PR + it was already addressed in other PRs

.github/workflows/changesets.yml
GregBrimble: Yeah

PR #6545, opened 2024-08-21
BLOCKED fix: remove `experimental:` prefix requirement for nodejs_compat_v2

petebacondarwin:
See https://jira.cfdata.org/browse/DEVDASH-218

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/green-lies-crash.md
+++ b/.changeset/green-lies-crash.md
@@ -4,4 +4,4 @@
 
 fix: remove `experimental:` prefix requirement for nodejs_compat_v2
 
-See https://jira.cfdata.org/bro
+See https://jira.cfdata.org/browse/DEVDASH-218
--- a/.changeset/green-lies-crash.md
+++ b/.changeset/green-lies-crash.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: remove `experimental:` prefix requirement for nodejs_compat_v2
+
+See https://jira.cfdata.org/bro
--- a/packages/wrangler/src/deployment-bundle/node-compat.ts
+++ b/packages/wrangler/src/deployment-bundle/node-compat.ts
@@ -27,25 +27,22 @@ export type NodeJSCompatMode = "legacy" | "v1" | "v2" | null;
  * We assert that you must prefix v2 mode with `experimental`.
  * We warn if using legacy or v2 mode.
  */
-export function validateNodeCompat({
-	legacyNodeCompat,
-	compatibilityFlags,
-	noBundle,
-}: {
-	legacyNodeCompat: boolean;
-	/* mutate */ compatibilityFlags: string[];
-	noBundle: boolean;
-}): NodeJSCompatMode {
-	if (legacyNodeCompat) {
-		logger.warn(
-			"Enabling Wrangler compile-time Node.js compatibility polyfill mode for builtins and globals. This is experimental and has serious tradeoffs."
+export function validateNodeCompat(
+	config: Pick<Config, "compatibility_flags" | "node_compat" | "no_bundle">
+): NodeJSCompatMode {
+	const {
+		mode,
+		nodejsCompat,
+		nodejsCompatV2,
+		experimentalNodejsCompatV2,
+		legacy,
+	} = getNodeCompatMode(config);
+
+	if (experimentalNodejsCompatV2) {
+		throw new UserError(
+			"The `experimental:` prefix on `nodejs_compat_v2` is no longer valid. Please remove it and try again."
 		);
 	}
-
-	const { mode, nodejsCompat, nodejsCompatV2 } = getNodeCompatMode({
-		compatibility_flags: compatibilityFlags,
-		node_compat: legacyNodeCompat,
-	});
 
 	if (nodejsCompat && nodejsCompatV2) {
 		throw new UserError(
@@ -53,21 +50,27 @@ export function validateNodeCompat({
 		);
 	}
 
-	if (legacyNodeCompat && (nodejsCompat || nodejsCompatV2)) {
+	if (legacy && (nodejsCompat || nodejsCompatV2)) {
 		throw new UserError(
 			`The ${nodejsCompat ? "`nodejs_compat`" : "`nodejs_compat_v2`"} compatibility flag cannot be used in conjunction with the legacy \`--node-compat\` flag. If you want to use the Workers ${nodejsCompat ? "`nodejs_compat`" : "`nodejs_compat_v2`"} compatibility flag, please remove the \`--node-compat\` argument from your CLI command or \`node_compat = true\` from your config file.`
 		);
 	}
 
-	if (noBundle && legacyNodeCompat) {
+	if (config.no_bundle && legacy) {
 		logger.warn(
 			"`--node-compat` and `--no-bundle` can't be used together. If you want to polyfill Node.js built-ins and disable Wrangler's bundling, please polyfill as part of your own bundling process."
 		);
 	}
 
-	if (noBundle && nodejsCompatV2) {
+	if (config.no_bundle && nodejsCompatV2) {
 		logger.warn(
 			"`nodejs_compat_v2` compatibility flag and `--no-bundle` can't be used together. If you want to polyfill Node.js built-ins and disable Wrangler's bundling, please polyfill as part of your own bundling process."
+		);
+	}
+
+	if (mode === "legacy") {
+		logger.warn(
+			"Enabling Wrangler compile-time Node.js compatibility polyfill mode for builtins and globals. This is experimental and has serious tradeoffs."
 		);
 	}
 
@@ -81,6 +84,9 @@ export function getNodeCompatMode({
 	const legacy = node_compat === true;
 	const nodejsCompat = compatibility_flags.includes("nodejs_compat");
 	const nodejsCompatV2 = compatibility_flags.includes("nodejs_compat_v2");
+	const experimentalNodejsCompatV2 = compatibility_flags.includes(
+		"experimental:nodejs_compat_v2"
+	);
 
 	let mode: NodeJSCompatMode;
 	if (nodejsCompatV2) {
@@ -98,5 +104,6 @@ export function getNodeCompatMode({
 		mode,
 		nodejsCompat,
 		nodejsCompatV2,
+		experimentalNodejsCompatV2,
 	};
 }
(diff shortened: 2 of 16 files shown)


.changeset/green-lies-crash.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: remove `experimental:` prefix requirement for nodejs_compat_v2
+
+See https://jira.cfdata.org/bro
```
vicb, line 7: Bro is a dead link

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -20,9 +20,8 @@ export type NodeJSCompatMode = "legacy" | "v1" | "v2" | null;
  *  - "v2": experimental nodejs_compat_v2 flag
  *  - null: no Node.js compatibility
  *
- * Currently we require that the v2 mode is configured via `experimental:nodejs_compat_v2` compat flag,
- * where the `experimental:` prefix is stripped before being passed to the runtime since that does not
- * understand this prefix.
+ * Currently we require that the v2 mode is configured via `nodejs_compat_v2` compat flag.
```
vicb, line 23: "Currently" -> "As of September 2024" ?

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -95,9 +79,7 @@ export function getNodeCompatMode({
 	node_compat,
 }: Pick<Config, "compatibility_flags" | "node_compat">) {
 	const nodejsCompat = compatibility_flags.includes("nodejs_compat");
-	const nodejsCompatV2 = compatibility_flags.includes(
-		"experimental:nodejs_compat_v2"
-	);
+	const nodejsCompatV2 = compatibility_flags.includes("nodejs_compat_v2");
```
vicb, line 82: `nodejsCompatV2 === true` implies  `nodejsCompat === true` because of the common prefix - this is probably a bug?

vicb: A few more nits - not directly related to this PR.

- What about renaming `node_compat` to `nodeCompatLegacy`?
- Could setting legacy use `mode` instead of `node_compat`?
- Any reason we are not using enums for flag names and mode?

vicb (changes requested):

packages/wrangler/src/deployment-bundle/node-compat.ts
petebacondarwin: > `nodejsCompatV2 === true` implies `nodejsCompat === true` because of the common prefix - this is probably a bug?

`compatibility_flags` is an array of strings, so these `include()` calls are correct.
E.g.

```
["nodejs_compat_v2"].includes("nodejs_compat") === false
```

petebacondarwin: > What about renaming node_compat to nodeCompatLegacy?

Sure happy to rename.

petebacondarwin: > Any reason we are not using enums for flag names and mode?

We don't tend to use enums much in this code base. It is nice to have strings when debugging.
Any reason to use enums?

petebacondarwin: > Could setting legacy use mode instead of node_compat?

Not sure what you mean here?

To be clear, what this function does is compute the `mode`, but also return flags telling us if any of these properties were turned on, so we can check the validaty of combinations.

vicb: > Any reason to use enums?

Avoid repeating a string with the risk of a typo.
enum 
- can be strings `enum Mode { V1 = "v1",  Legacy="legacy"}`
- support autocomplete
- support exhaustive switch (i.e. warn/error when adding a value and not updating a switch)
- are easier to search in the code (arguably)

petebacondarwin: > > Any reason to use enums?
> 
> Avoid repeating a string with the risk of a typo. enum
> 
> * can be strings `enum Mode { V1 = "v1",  Legacy="legacy"}`
> * support autocomplete
> * support exhaustive switch (i.e. warn/error when adding a value and not updating a switch)
> * are easier to search in the code (arguably)

Since the type is a narrow collection of const strings we get all those already: https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcByEAmwBSBlAYQgFswBDGAWS1QF44AiAG2AHMyBjRBuAH0YBuARh78GAgEyi4AOwCuTJgG4AUCo4QZAZ3hkAXOhr4ipCtWxx6zNp26qN2+ACMDGbMZLkqNS4JH3NHTgOVyNCTzMfK0kGAMc4TFD3cNNvC3p5RTigpzJEw2STL3M6RgAzCAgGNWwOJjIoVDK5GQ4YAEtNOFZgNOAACgBKJNwU4ppVFS0Ad3aYDgALOH6evqHBuABvFThd4LItVGt2LgY9Hb3Lp0ayAGtVS44Do+Ezi8vd6+A7h72nw8EUnOHz2Xx+732AIYuUwbxBnxu9wh-1QmSYwPhYKRAF8gA

petebacondarwin: > > Not sure what you mean here?
> 
> `legacy: node_compat === true,` -> `legacy: mode === "legacy",`
> 
> Just to avoid repeating logic as repeating might lead to out of sync code

Ah I see! I can create an extra var to DRY that.

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -94,39 +78,25 @@ export function getNodeCompatMode({
 	compatibility_flags,
 	node_compat,
 }: Pick<Config, "compatibility_flags" | "node_compat">) {
+	const legacy = node_compat === true;
```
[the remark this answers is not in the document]
petebacondarwin: Legacy is a special pre-`nodejs_compat` mode.
After we got `nodejs_compat` we go into versioned mode, e.g. v1, v2, etc.
V1 is not legacy it is just an older version of the same mode.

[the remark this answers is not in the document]
vicb: Oh, it reminds me one nice thing with enum: you can see the JSDoc when hovering on a value

Screen Shot 2024-08-21 at 20 13 56 PM

petebacondarwin:
> one more thing to consider: do we want to issue a nice warning to users that are currently using the experimental prefix? could we error and tell them that the flag is no longer experimental and they should update their config rather than error with message saying that the flag doesn't exist, which I think will happen once this change is merged.

Very good point. I'll add that.

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -104,29 +94,17 @@ export function getNodeCompatMode({
...
-	} else if (node_compat) {
+	} else if (legacy) {
 		mode = "legacy";
 	} else {
 		mode = null;
 	}
 
 	return {
-		legacy: node_compat === true,
+		legacy,
```
vicb, line 104: I would love this function to return a single value for clarity.

It looks to me that `legacy`, `nodejsCompat`, and `nodejsCompatv2` can be collapsed into mode.

`experimentalNodejsCompatV2` might be handled in a different functions.

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -17,85 +16,76 @@ export type NodeJSCompatMode = "legacy" | "v1" | "v2" | null;
...
+): NodeJSCompatMode {
+	const {
+		mode,
+		nodejsCompat,
+		nodejsCompatV2,
+		experimentalNodejsCompatV2,
+		legacy,
+	} = getNodeCompatMode(compatibilityFlags, nodeCompat);
+
+	if (experimentalNodejsCompatV2) {
```
vicb, line 42: Maybe throw inside of `getNodeCompatMode`so that there is no need to return `experimentalNodejsCompatV2`

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -17,85 +16,76 @@ export type NodeJSCompatMode = "legacy" | "v1" | "v2" | null;
...
 
-	const { mode, nodejsCompat, nodejsCompatV2 } = getNodeCompatMode({
-		compatibility_flags: compatibilityFlags,
-		node_compat: legacyNodeCompat,
-	});
-
-	const nodejsCompatV2NotExperimental =
-		compatibilityFlags.includes("nodejs_compat_v2");
-
 	if (nodejsCompat && nodejsCompatV2) {
```
vicb, line 48: Same here, I think we should throw in `getNodeCompatMode`

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -17,85 +16,76 @@ export type NodeJSCompatMode = "legacy" | "v1" | "v2" | null;
...
 		throw new UserError(
 			"The `nodejs_compat` and `nodejs_compat_v2` compatibility flags cannot be used in together. Please select just one."
 		);
 	}
 
-	if (legacyNodeCompat && (nodejsCompat || nodejsCompatV2)) {
+	if (legacy && (nodejsCompat || nodejsCompatV2)) {
```
vicb, line 54: Same here, I think we should throw in getNodeCompatMode

packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -17,85 +16,76 @@ export type NodeJSCompatMode = "legacy" | "v1" | "v2" | null;
...
 	}
 
-	if (nodejsCompatV2NotExperimental) {
-		throw new UserError(
-			`The \`nodejs_compat_v2\` compatibility flag is experimental and must be prefixed with \`experimental:\`. Use \`experimental:nodejs_compat_v2\` flag instead.`
-		);
-	}
-
-	if (nodejsCompatV2) {
+	if (noBundle && legacy) {
```
vicb, line 60: Same here, I think we should throw in getNodeCompatMode

RamIdeas:
Closing in favour of #6593

PR #6546, opened 2024-08-21
feat(quick-edit): add "search.exclude" support to quick editor

edmundhung:

Fixes [DEVX-1384](https://jira.cfdata.org/browse/DEVX-1384)

This PR adds `wrangler.toml` and `package.json` to the `files.exclude` setting to reduce user confusion and make sure these files are not searchable.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: No feature change


packages/quick-edit/functions/_middleware.ts
```diff
@@ -19,6 +19,8 @@ export const onRequest = async ({
 				"files.exclude": {
 					"*.d.ts": true,
 					"jsconfig.json": true,
+					"**/package.json": true,
+					"**/wrangler.toml": true,
```
edmundhung, line 23: @penalosa We are hidding the `.d.ts` and `jsconfig.json` files on root only. Do we want to make it hidden on any directories as well? 

```suggestion
					"**/*.d.ts": true,
					"**/jsconfig.json": true,
					"**/package.json": true,
					"**/wrangler.toml": true,
```

penalosa: No—these settings are to catch files we inject at the root

packages/quick-edit-extension/src/cfs.ts
```diff
@@ -527,19 +527,49 @@ declare module "*.bin" {
...
+						)
+					);
+
+					patterns.push(filenamePattern);
+				}
+
+				return patterns;
+			},
+			[]
+		);
```
penalosa, line 565: I think this would probably read more clearly as a filter -> map chain? (or just a map)

packages/quick-edit-extension/src/cfs.ts
```diff
@@ -527,19 +527,49 @@ declare module "*.bin" {
...
+
 		for (const file of files) {
-			if (!pattern || pattern.exec(file.name)) {
+			if (
+				(!pattern || pattern.exec(file.name)) &&
+				// Ensure the file name is not excluded
+				excludePatterns.every((regex) => !regex.exec(file.name))
```
penalosa, line 571: Personal preference, but I would find this clearer as `!excludePatterns.some(...)`

packages/quick-edit/functions/_middleware.ts
penalosa: Although to be honest we also only need to catch wrangler.toml and package.json at the root, so maybe those don't need wildcards?

penalosa (commented):
Looking good!

packages/quick-edit/functions/_middleware.ts
edmundhung: Fixed on https://github.com/cloudflare/workers-sdk/pull/6546/commits/ff46d68cbf4d9f555b2b18350264c230cb15feab.

packages/quick-edit-extension/src/cfs.ts
edmundhung: I simplified how exclude are matched against each file on https://github.com/cloudflare/workers-sdk/pull/6546/commits/35b63ae1b6c2a8ee1cfa41b3e7bf8cbc9d694e33

penalosa (approved):

PR #6550, opened 2024-08-21
feat(wrangler): add ratelimit type generation

### LINKED ISSUE #6529: 🚀 Feature Request: Wrangler type generator `RateLimit` type
### Describe the solution

Hello,

Now that [this PR](https://github.com/cloudflare/workerd/pull/2170) adding the type definition for the beta `RateLimit` binding is merged, is it worth adding it in the wrangler type generator? I would like to use it, but it gets overwritten by `any` when I generate the types.

Since it's still an unsafe binding, I assume the checks would still have to me made [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/type-generation/index.ts#L347) for now and later on removed to become its own env field binding. Is there anywhere else it needs to be added?

If this is okay to add, I can open a PR with the type + tests.

Pedr0Rocha:
Adds the type [RateLimit](https://github.com/cloudflare/workerd/blob/main/types/defines/ratelimit.d.ts) to wrangler type generation for the beta rate limiter unsafe binding.

Fixes #6529 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/type-generation.test.ts
+++ b/packages/wrangler/src/__tests__/type-generation.test.ts
@@ -198,7 +198,10 @@ const bindingsConfigMock: Omit<
 	},
 	wasm_modules: { MODULE1: "module1.wasm", MODULE2: "module2.wasm" },
 	unsafe: {
-		bindings: [{ name: "testing_unsafe", type: "plain_text" }],
+		bindings: [
+			{ name: "testing_unsafe", type: "plain_text" },
+			{ name: "UNSAFE_RATELIMIT", type: "ratelimit" },
+		],
 		metadata: { some_key: "some_value" },
 	},
 	rules: [
@@ -342,6 +345,7 @@ describe("generateTypes()", () => {
 			SOME_TEXT_BLOB1: string;
 			SOME_TEXT_BLOB2: string;
 			testing_unsafe: any;
+			UNSAFE_RATELIMIT: RateLimit;
 			TEST_QUEUE_BINDING: Queue;
 			SEND_EMAIL_BINDING: SendEmail;
 			VECTORIZE_BINDING: VectorizeIndex;
@@ -482,6 +486,7 @@ describe("generateTypes()", () => {
 		export {};
 		declare global {
 			const testing_unsafe: any;
+			const UNSAFE_RATELIMIT: RateLimit;
 		}
 		"
 	`);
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -482,6 +486,7 @@ describe("generateTypes()", () => {
 		export {};
```
Pedr0Rocha, line 486: Should I update this to ES Module or there's a reason we have Service Worker syntax in this test?

packages/wrangler/src/type-generation/index.ts
```diff
@@ -344,7 +344,11 @@ async function generateTypes(
 
 	if (configToDTS.unsafe?.bindings) {
 		for (const unsafe of configToDTS.unsafe.bindings) {
-			envTypeStructure.push(constructType(unsafe.name, "any"));
+			if (unsafe.type === "ratelimit") {
+				envTypeStructure.push(constructType(unsafe.name, "RateLimit"));
+			} else {
+				envTypeStructure.push(constructType(unsafe.name, "any"));
```
Skye-31, line 350: I'm not sure this is a great idea to fallback to any for other unsafe bindings - people may be defining the unsafe binding types manually which this will conflict with

Pedr0Rocha: I might be missing something, but isn't this the current behaviour?
This PR is supposed to fix this problem only for the rate limiter unsafe binding

Skye-31: You're right - I shouldn't be reviewing code late at night 😅

packages/wrangler/src/__tests__/type-generation.test.ts
penalosa: This is intended, to ensure we properly generate the global interface

penalosa (approved):
Looks good to me!

PR #6554, opened 2024-08-22
fix: ignore nodejs_compat flags in runtime type generation

### LINKED ISSUE #6540: 🐛 BUG: wrangler types --x-include-runtime  => No such compatibility flag: experimental:nodejs_compat_v2

Wrangler

3.72.0

21.3

Sonoma latest

### Describe the Bug

try it on https://github.com/irvinebroque/fs-stub-override

```
fs-stub-override % npx wrangler@latest types --x-include-runtime
Need to install the following packages:
wrangler@3.72.1
Ok to proceed? (y) y


 ⛅️ wrangler 3.72.1
-------------------

Generating runtime types...

✘ [ERROR] Error: No such compatibility flag: experimental:nodejs_compat_v2

      at Object.fetch (script:0:4049:980)
```

_No response_

_No response_

andyjessop:

Fixes #6540.

There is currently no way to dynamically generate `nodejs_compat` runtime types, so here they are being filtered out before making the call to the type-generating Worker.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Bug fix


packages/wrangler/src/deployment-bundle/node-compat.ts
```diff
@@ -118,7 +118,7 @@ export function getNodeCompatMode({
 	}
 
 	return {
-		legacy: node_compat === true,
+		legacy: mode === "legacy",
```
petebacondarwin, line 121: why this change?

andyjessop: Sorry, should have mentioned. This is just a really minor nit that I had which was because it was duplicating some logic from above, whereas this change derives the property from the above if/else.

andyjessop: I've added some tests here because I believe that was actually fixing a bug: https://github.com/cloudflare/workers-sdk/pull/6554/commits/9d8e1a72a2ea74c43b75441583011d08303a0e83

Unless I'm mistaken and it should return `legacy` no matter what flags are overriding the `node_compat = true` setting.

packages/wrangler/index.js
```diff
@@ -0,0 +1,6 @@
+import { foo } from "./another";
```
andyjessop, line 1: No idea how this file got in there. I'll remove it now.

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -18,10 +18,10 @@ import {
 } from "../ai/fetcher";
 import { readConfig } from "../config";
 import { ModuleTypeToRuleType } from "../deployment-bundle/module-collection";
-import { stripExperimentalPrefixes } from "../deployment-bundle/node-compat";
 import { withSourceURLs } from "../deployment-bundle/source-url";
 import { UserError } from "../errors";
 import { logger } from "../logger";
+import { stripExperimentalPrefixes } from "../node-compat";
```
andyjessop, line 24: These changes are because the node_compat helpers have been moved to a location where it makes more sense (outside the `deployment-bundle` folder).

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -49,7 +49,10 @@ export async function generateRuntimeTypes({
 
 	const types = await generate({
 		compatibilityDate: compatibility_date,
-		compatibilityFlags: compatibility_flags,
+		// Ignore nodejs compat flags as there is currently no mechanism to generate these dynamically.
+		compatibilityFlags: compatibility_flags.filter(
+			(flag) => !flag.includes("nodejs_compat")
```
andyjessop, line 54: --> This is the actual bug fix <---

packages/wrangler/index.js
petebacondarwin: See https://github.com/cloudflare/workers-sdk/pull/6547

packages/wrangler/src/deployment-bundle/node-compat.ts
petebacondarwin: I have some refactoring of this code in here: https://github.com/cloudflare/workers-sdk/pull/6545

andyjessop: Ah I see. Coincidence there. Would it be best if I just removed all the refactoring I've done here and leave the original change (just the bug fix)?

petebacondarwin: Up to you. That other PR won't land for a while.
So no doubt there will be other conflicts to tidy up then.

andyjessop: I'll just remove it - they both tackle the same issue.

CarmenPopoviciu (approved):

PR #6564, opened 2024-08-23
feat:`wrangler dev` with dynamic sites + miniflare assets plugin

emily-shen:

This creates a new `assets` miniflare plugin. The Asset Worker and Router Worker are imported from `workers-shared` and used as miniflare services. If assets are detected, the miniflare entry worker will always point towards the Router Worker instead of the User Worker like it usually does.  

We also add in a config object that can be passed to Router Worker. Currently this only contains `{hasUserWorker: boolean}`, but will be used later for other routing config.

User Worker + Assets diagram. If there is no UW, this remains the same except wrangler provides a no-op worker that cannot be reached.

diagram

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [x] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/miniflare/scripts/build.mjs
```diff
@@ -102,7 +102,7 @@ const embedWorkersPlugin = {
 					bundle: true,
 					sourcemap: true,
 					sourcesContent: false,
-					external: ["miniflare:shared", "miniflare:zod"],
+					external: ["miniflare:shared", "miniflare:zod", "cloudflare:workers"],
```
CarmenPopoviciu, line 105: why is `cloudflare:workers` marked as an external dependency here?

packages/miniflare/src/plugins/assets/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const ASSETS_PLUGIN_NAME = "assets";
+export const ASSETS_SERVICE_NAME = `${ASSETS_PLUGIN_NAME}:assets-service`;
+export const ROUTER_SERVICE_NAME = `${ASSETS_PLUGIN_NAME}:router`;
+export const FAKE_KV_SERVICE_NAME = `${ASSETS_PLUGIN_NAME}:kv`;
```
CarmenPopoviciu, line 4: personally I'd rather stick with the existing naming convention that we have in Miniflare, which doesn't seem to use prefixing such as `FAKE_`. I don't see any added value to that prefix. I think `KV_SERVICE_NAME` or `ASSETS_KV_SERVICE_NAME` should do the trick

packages/miniflare/src/plugins/assets/schema.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from "zod";
+import { PathSchema } from "../../shared";
+
+export const AssetsOptionsSchema = z.object({
+	workerName: z.string().optional(),
+	assetsPath: PathSchema.optional(),
```
CarmenPopoviciu, line 10: as per the interface definition below, `assetsPath` is mandatory. So this should be `z.string()` instead. This will ensure that incorrect asset options will fail zod validation and will result in an error, which is what we want. So basically AssetOptions is optional on the Miniflare options type def, but `assetsPath` should be mandatory, if those options are passed

CarmenPopoviciu: the validation err we'll see once `assetsPath` is marked as non-optional is this
Screenshot 2024-08-29 at 10 08 17

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -1,95 +1,84 @@
...
-export interface AssetsOptions {
-	assetsPath: string;
-	assetsKVBindingName?: string;
-	assetsManifestBindingName?: string;
-}
+export const ASSETS_PLUGIN: Plugin<typeof AssetsOptionsSchema> = {
+	options: AssetsOptionsSchema,
+	async getBindings(options) {
+		const bindings: Worker_Binding[] = [];
+		if (isWorkersWithAssets(options)) {
```
CarmenPopoviciu, line 29: I don't think the extra `if (isWorkersWithAssets(options))` check is necessary here. Technically, if the asset options were not passed into the Miniflare options, `options.assetsBindingName` will be `undefined`, ergo `if (options.assetsBindingName)` will be falsy and getBindings will return an empty array. The underlying issue here I believe is that the zod schema marks `assetsPath` as optional, when it in fact should be mandatory, see my comment [here](https://github.com/cloudflare/workers-sdk/pull/6564/commits/522bae675f9b31193de03c20cfa5fd24a8f74509#r1735762668), so I'm assuming the extra check comes from that interface/zod schema misalignment. Once we fix the zod schema, this will all work as expected, and that extra check won't be necessary anymore

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -1,95 +1,84 @@
...
+	},
 
-export const buildAssetsManifest = async (dir: string) => {
-	const manifest = await walk(dir);
-	const sortedAssetManifest = sortManifest(manifest);
-	const encodedAssetManifest = encodeManifest(sortedAssetManifest);
-	return encodedAssetManifest;
+	async getServices({ options }) {
+		const services: Service[] = [];
+		if (isWorkersWithAssets(options)) {
```
CarmenPopoviciu, line 55: based on my comments above, this can be `if (options.assetsPath)`

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -1,95 +1,84 @@
...
 
-	const storageServiceName = `${SERVICE_NAMESPACE_ASSET}:storage`;
+	const storageServiceName = `${ASSETS_PLUGIN_NAME}:storage`;
 	const storageService: Service = {
 		name: storageServiceName,
 		disk: { path: options.assetsPath, writable: true },
 	};
 	const namespaceService: Service = {
-		name: SERVICE_NAMESPACE_ASSET,
+		name: FAKE_KV_SERVICE_NAME,
```
CarmenPopoviciu, line 74: same comment as [here](https://github.com/cloudflare/workers-sdk/pull/6564/commits/522bae675f9b31193de03c20cfa5fd24a8f74509#r1735684791)

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -98,16 +87,64 @@ export async function getAssetsServices(
...
 			],
 		},
 	};
-	return [storageService, namespaceService];
+
+	const assetsManifest = await buildAssetsManifest(options.assetsPath);
+	const assetService: Service = {
+		name: ASSETS_SERVICE_NAME,
+		worker: {
+			compatibilityDate: "2024-08-01",
```
CarmenPopoviciu, line 101: is there a way to defer to the compat date & other relevant config specified by the AW's `wrangler.toml`?

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -98,16 +87,64 @@ export async function getAssetsServices(
...
+					data: assetsManifest,
+				},
+			],
+		},
+	};
+
+	const routerService: Service = {
+		name: ROUTER_SERVICE_NAME,
+		worker: {
+			compatibilityDate: "2024-08-01",
```
CarmenPopoviciu, line 124: same here. Ideally we want AW and RW to be the source of truth for these things..as much as possible and makes sense ofc

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -98,16 +87,64 @@ export async function getAssetsServices(
...
+			modules: [
+				{
+					name: "router-worker.mjs",
+					esModule: SCRIPT_ROUTER(),
+				},
+			],
+			bindings: [
```
CarmenPopoviciu, line 131: nice! very clean and readable! 😻

packages/miniflare/src/plugins/assets/schema.ts
```diff
@@ -0,0 +1,20 @@
...
+	assetsBindingName: z.string().optional(),
+});
+
+export interface AssetsOptions {
+	workerName?: string;
+	assetsPath: string;
+	assetsBindingName?: string;
+}
+
+export function isWorkersWithAssets(
```
CarmenPopoviciu, line 23: I think this won't be necessary anymore, if we fix the zod schema

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -712,7 +716,11 @@ export function getGlobalServices({
 		{ name: CoreBindings.JSON_LOG_LEVEL, json: JSON.stringify(log.level) },
 		{
 			name: CoreBindings.SERVICE_USER_FALLBACK,
-			service: { name: getUserServiceName(fallbackWorkerName) },
+			service: {
```
CarmenPopoviciu, line 719: IMHO this decision should be moved to where `getGlobalServices` is [called](https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/src/index.ts#L1249) and should not be an implementation detail of this fn

packages/miniflare/test/fixtures/assets/asset-test-helpers.ts
emily-shen, line 1: for context, these tests were removed because we are now importing the asset worker in `assets.worker.ts` from `workers-shared`, which has its own tests

packages/miniflare/scripts/build.mjs
CarmenPopoviciu: because `cloudflare:workers` is a runtime thing, and therefore esbuild cannot resolve it at build time. Marking this as external is the correct way to go

packages/miniflare/src/plugins/assets/constants.ts
emily-shen: ASSETS_KV_SERVICE_NAME sounds good to me :)

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -137,6 +137,10 @@ export async function getAssetService(
 					name: "USER_WORKER",
 					service: { name: getUserServiceName(options.workerName) },
 				},
+				{
+					name: "CONFIG",
```
CarmenPopoviciu, line 141: nit: would it be more expressive if we called this `ROUTING_CONFIG`?

packages/miniflare/src/plugins/assets/constants.ts
CarmenPopoviciu: much awesomeness!

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -680,6 +678,19 @@ export function buildPersistOptions(
 	}
 }
 
+function buildAssetOptions(config: Omit<ConfigBundle, "rules">) {
+	if (config.experimentalAssets) {
+		return {
+			/** If undefined, this is a static assets only project and there is no user worker */
+			workerName: config.name,
```
CarmenPopoviciu, line 684: is there no way for Miniflare to extract the worker name without us passing it as an assets option? Feels like assets options shouldn't necessarily be aware of/have to deal with the user worker name?

fixtures/workers-with-assets/public/binding.html
```diff
@@ -0,0 +1 @@
+<h1>✨This is from a user worker binding✨</h1>
```
CarmenPopoviciu, line 1: So sorry, this should be a suggestion, but GH doesn't support that? when reviewing per commit :/

`<h1>✨This is from a user Worker binding✨</h1>`

fixtures/workers-with-assets/package.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "assets-worker",
+	"name": "workers-assets-with-worker",
```
CarmenPopoviciu, line 2: nit: `workers-assets-with-user-worker` or `dynamic-workers-with-assets` and explain what that means in the readme

fixtures/workers-with-assets/tests/index.test.ts
```diff
@@ -29,22 +29,26 @@ describe("[Workers + Assets] `wrangler dev`", () => {
 		expect(text).toContain(`<p>Learn more about Workers with Assets soon!</p>`);
 	});
 
-	it("should not resolve '/' to '/index.html' ", async ({ expect }) => {
-		let response = await fetch(`http://${ip}:${port}/`);
-		expect(response.status).toBe(404);
-	});
-
-	it("should 404 if asset is not found in the asset manifest", async ({
+	it("should fallback to the user worker if there are no assets at a given path ", async ({
```
CarmenPopoviciu, line 32: `user Worker`

fixtures/workers-with-assets/tests/index.test.ts
```diff
@@ -29,22 +29,26 @@ describe("[Workers + Assets] `wrangler dev`", () => {
...
 		expect,
 	}) => {
-		let response = await fetch(`http://${ip}:${port}/hello.html`);
-		expect(response.status).toBe(404);
-
-		response = await fetch(`http://${ip}:${port}/hello.txt`);
-		expect(response.status).toBe(404);
+		// Requests should hit the asset worker *first*, then try the user worker
```
CarmenPopoviciu, line 35: `Asset Worker`

CarmenPopoviciu: basically what I'm trying to say is that worker should be `Worker` when referring to CF Workers

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -1,21 +1,17 @@
...
 		const result = await env.ASSET_WORKER.fetch(request);
-		if (!result.ok) {
-			if (result.status === 404) {
-				return await env.USER_WORKER.fetch(request);
-			}
-			// return failed response on non-404 errors
-			return new Response(`Failed to fetch content: ${result.statusText}`, {
-				status: result.status,
-			});
+		// 404 = asset not in manifest
```
CarmenPopoviciu, line 13: ♥️

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -1,21 +1,17 @@
-import type { Fetcher, Request } from "@cloudflare/workers-types";
-
 interface Env {
 	ASSET_WORKER: Fetcher;
 	USER_WORKER: Fetcher;
 }
+
 export default {
 	async fetch(request: Request, env: Env) {
+		const maybeUserRequest = request.clone();
```
CarmenPopoviciu, line 11: I'll need to come back to this, but just as a reminder that I have a questions here... do we want to `clone()` here or create a new req via the `Request` constructor?

fixtures/workers-with-assets/tests/index.test.ts
```diff
@@ -29,22 +29,26 @@ describe("[Workers + Assets] `wrangler dev`", () => {
...
+		let response = await fetch(`http://${ip}:${port}/no-assets-here`);
+		let text = await response.text();
+		expect(text).toContain(
+			"There were no assets at this route! Hello from the user worker instead!"
+		);
+		// `/` should not resolve to `/index.html`
```
CarmenPopoviciu, line 41: can we have this in a separate test pls 🙏 ?

fixtures/workers-with-assets-only/tests/index.test.ts
```diff
@@ -0,0 +1,155 @@
+import { resolve } from "node:path";
+import { fetch } from "undici";
+import { afterAll, beforeAll, describe, it } from "vitest";
+import { runWranglerDev } from "../../shared/src/run-wrangler-long-lived";
+
+describe("[Workers + Assets] `wrangler dev`", () => {
```
CarmenPopoviciu, line 6: `"[Workers + Assets] static-only site"`

packages/workers-shared/utils/types.ts
```diff
@@ -0,0 +1,10 @@
+// import { z } from "zod";
+
+// export const RoutingConfigSchema = z.object({
+// 	hasUserWorker: z.boolean(),
+// });
+
+// export type RoutingConfig = z.infer<typeof RoutingConfigSchema>;
+export type RoutingConfig = {
+	hasUserWorker: boolean;
```
CarmenPopoviciu, line 9: I'm a bit confused. Is this the same routing config that is supposed to config things like `spaMode`/`trailingSlashes`/etc? Cause if so, adding `hasUserWorker` to it doesn't feel right to me. It's two very diff concerns

fixtures/workers-with-assets/tests/index.test.ts
```diff
@@ -73,44 +77,103 @@ describe("[Workers + Assets] `wrangler dev`", () => {
...
+		});
+		expect(response.status).toBe(200);
+
+		response = await fetch(`http://${ip}:${port}/no-assets-here`, {
+			method: "PATCH",
+		});
+		expect(response.status).toBe(200);
+	});
+
+	it("should be able to use an ASSETS binding", async ({ expect }) => {
```
CarmenPopoviciu, line 170: we should also test the use case when an asset is actually found in the asset manifest

fixtures/workers-with-assets/src/index.ts
```diff
@@ -0,0 +1,18 @@
+export interface Env {
+	// Example binding to a Service. Learn more at https://developers.cloudflare.com/workers/runtime-apis/service-bindings/
+	ASSETS: Fetcher;
+}
+
+export default {
```
CarmenPopoviciu, line 6: I'm a a little confused by this fixture setup. I'll come back to it, or ping you async with questions

packages/miniflare/src/plugins/assets/schema.ts
emily-shen: whoops yeah good point - I've nested the options now so the whole assetOptions object can be optional, but some of the properties required

packages/miniflare/src/plugins/assets/index.ts
emily-shen: sorted it out now alongside - https://github.com/cloudflare/workers-sdk/pull/6564#discussion_r1736038376

fixtures/workers-with-assets/package.json
RamIdeas: prefer the uber explicit `workers-assets-with-user-worker`

packages/miniflare/src/plugins/assets/index.ts
emily-shen: Since it was quite painful to import + read a `wrangler.toml` file from outside `workers-shared` last time we tried, I might add this to our list of followups?

fixtures/workers-with-assets/tests/index.test.ts
RamIdeas: Seems reasonable to keep this in this test. It's a slight variation of the same assertion: fallback user worker on arbitrary paths _including the root path_

packages/miniflare/src/plugins/assets/index.ts
RamIdeas: I agree. "fake" doesn't add value. Try to identify all these assets with their feature like "assets"

packages/miniflare/src/plugins/assets/index.ts
RamIdeas: This is a great callout! Thinking this through, reading a wrangler.toml from miniflare is going to be hacky. I think the best option for now, and simplest, is to add a comment to the compatibility_date line in each of the wrangler.toml's of the AssetWorker and RouterWorker that says it needs to be kept in sync with these lines in miniflare

fixtures/workers-with-assets/tests/index.test.ts
CarmenPopoviciu: kk

packages/workers-shared/utils/types.ts
emily-shen: Yep its the same routing config. Having the user Worker there does change the behaviour of Router Worker so it makes sense to me (whether to pass onto UW or return 404). It is _also_ used for other things, but ehhh I like reusing it more than creating another separate option. I think EWC is set up for it to be like that already 🫠

fixtures/workers-with-assets/src/index.ts
CarmenPopoviciu: nvm my brain was just tired yesterday. Keep calm and carry on and ignore me :)

packages/miniflare/src/plugins/assets/index.ts
CarmenPopoviciu: this is tracked in WC-2636

packages/wrangler/src/dev/miniflare.ts
CarmenPopoviciu: let's just leave as it is for now

fixtures/workers-with-assets/tests/index.test.ts
CarmenPopoviciu: turns out that's already tested. moving on

packages/miniflare/src/plugins/assets/index.ts
CarmenPopoviciu: for posterity...we need to stick to `CONFIG` since that's what EWC does

packages/miniflare/src/plugins/assets/index.ts
CarmenPopoviciu: this is tracked in WC-2636

fixtures/workers-with-assets/README.md
```diff
@@ -1,6 +1,6 @@
-# workers-with-assets
+# workers-assets-with-user-worker"
 
-`workers-with-assets` is a test fixture that showcases Workers with Assets. This particular fixture sets up an assets-only Workers project.
+`workers-assets-with-user-worker"` is a test fixture that showcases Workers with Assets. This particular fixture sets up a User Worker, assets, and a binding from the user Worker to the assets.
```
CarmenPopoviciu, line 3: :meouw-heart:

CarmenPopoviciu (approved):

packages/miniflare/src/index.ts
```diff
@@ -1250,6 +1250,7 @@ export class Miniflare {
 			loopbackPort,
 			log: this.#log,
 			proxyBindings,
+			hasWorkerAssets: !!this.#workerOpts[0].assets.assets,
```
RamIdeas, line 1253: This is a fun assignment 😆

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -132,6 +171,13 @@ const TAIL_RESERVED_SIZE = 8;
 
 const ENTRY_SIZE = PATH_HASH_SIZE + CONTENT_HASH_SIZE + TAIL_RESERVED_SIZE;
 
+export const buildAssetsManifest = async (dir: string) => {
```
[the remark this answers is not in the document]
RamIdeas: Big fan!

packages/miniflare/src/index.ts
CarmenPopoviciu: hmmm...that is true

packages/miniflare/src/plugins/core/index.ts
RamIdeas: I don't think this comment has been addressed. Marking as unresolved

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -1,113 +1,139 @@
...
-};
+export const ASSETS_PLUGIN: Plugin<typeof AssetsOptionsSchema> = {
+	options: AssetsOptionsSchema,
+	async getBindings(options: z.infer<typeof AssetsOptionsSchema>) {
+		if (!options.assets?.bindingName) {
+			return [];
+		}
+		return [
+			{
+				// binding between User Worker and Asset Worke
```
RamIdeas, line 30:
```suggestion
				// binding between User Worker and Asset Worker
```

RamIdeas (approved):

PR #6565, opened 2024-08-23
Update Svelte display name to SvelteKit

### LINKED ISSUE #3263: 🐛 BUG: C3's Svelte option is really SvelteKit - confusing?

penalosa:

Per https://github.com/cloudflare/workers-sdk/issues/3263#issuecomment-1888034131, this is what the Svelte team recommends. The Svelte framework template runs `create-svelte` which scaffolds a SvelteKit project, so this should make it clearer to users what type of project they're actually going to get. 

Fixes #3263

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: this doesn;t have any tests
  - [ ] I don't know
  - [x] Not required because: e2e doesn't apply to c3
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: The Svelte guide skips this step with a CLI arg: https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-site/

### DIFF
--- a/.changeset/giant-pens-explain.md
+++ b/.changeset/giant-pens-explain.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+fix: Rename Svelte to SvelteKit
--- a/packages/create-cloudflare/templates/svelte/c3.ts
+++ b/packages/create-cloudflare/templates/svelte/c3.ts
@@ -97,7 +97,7 @@ const updateTypeDefinitions = (ctx: C3Context) => {
 const config: TemplateConfig = {
 	configVersion: 1,
 	id: "svelte",
-	displayName: "Svelte",
+	displayName: "SvelteKit",
 	platform: "pages",
 	copyFiles: {
 		variants: {


CarmenPopoviciu (approved):

penalosa:
Closing in favour of https://github.com/cloudflare/workers-sdk/pull/7109

PR #6573, opened 2024-08-23
 [C3] Use create-vite instead of create-react-app for React projects

penalosa:

[React's documentation](https://react.dev/learn/start-a-new-react-project#can-i-use-react-without-a-framework) now recommends using `create-vite` over `create-react-app`, as the latter has been deprecated. To align with these best practices, React projects created with C3 will now use [Vite](https://vitejs.dev/).

With this change, the default development command will switch from using `create-react-app` to `create-vite`, providing a more modern and efficient development experience.

Additionally, selection of variants for React projects created by Vite has been added, allowing users to choose different configurations based on their needs. The test suite has also been adapted to accommodate the new configuration, ensuring all tests run without errors.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Not necessary because: Vite already has detailed [documentation](https://vitejs.dev/guide/).


dario-piotrowicz (approved):
Looks great to me! 😄

Thanks a lot for getting this over the line @penalosa! ❤️ 

And thank you so very much for the original PR @santraez, it is truly appreciated! 🫶  sorry we didn't notice and merge it for months! 🙇

IgorMinar (approved):
I'm supportive of this move, but I also want to bring up that this change was originally proposed in https://github.com/cloudflare/workers-sdk/pull/5891 by @santraez and we didn't process it in a timely manner before @santraez closed the PR.

I suggest we still credit @santraez in the changelog given that he was right about this all along, we just didn't have the capacity to process his suggestion.

penalosa:
> I'm supportive of this move, but I also want to bring up that this change was originally proposed in #5891 by @santraez and we didn't process it in a timely manner before @santraez closed the PR.
> 
> I suggest we still credit @santraez in the changelog given that he was right about this all along, we just didn't have the capacity to process his suggestion.

Absolutely! And for the record, the reason this is a new PR rather than a rebase of @santraez's original PR is that the fork was deleted. I'll modify the changelog to make authorship clear.

PR #6574, opened 2024-08-23
feat(wrangler): watch mode for Workers + Assets

CarmenPopoviciu:

This PR implements watch mode for Workers with assets, more specifically:
- it watches the static assets directory for changes (add/delete files or directories)
- it reacts to config file changes and ensures we're still watching for the relevant asset directory

Fixes WC-2554

Clean commit history is caring 🫶  so pls review per commit

## How this was tested

Beside e2e tests, these changes were also manually tested in the following scenarios:

**[`wrangler dev` | `wrangler dev --x-dev-env`]**
- add/remove files & directories in the assets directory
Screenshot 2024-08-28 at 12 37 21

- change the assets folder in `wrangler.toml` + file changes in new assets directory + file changes in old assets directory to ensure it is not watched anymore
Screenshot 2024-08-28 at 12 38 12

**[`wrangler dev --x-assets="dist"` | `wrangler dev --x-dev-env --x-assets="dist"`]**
- add/remove files & directories in the assets directory
- change the assets folder in `wrangler.toml` + file changes in that assets directory to ensure we are still watching the assets dir specified by `--x-assets`
Screenshot 2024-08-28 at 12 46 40

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Workers + Assets is not documented yet

### DIFF
--- a/packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
+++ b/packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
@@ -2,7 +2,7 @@ import assert from "node:assert";
 import crypto from "node:crypto";
 import { onTestFinished } from "vitest";
 import { generateResourceName } from "./generate-resource-name";
-import { makeRoot, seed } from "./setup";
+import { makeRoot, remove, seed } from "./setup";
 import {
 	runWrangler,
 	WRANGLER_IMPORT,
@@ -19,6 +19,10 @@ export class WranglerE2ETestHelper {
 
 	async seed(files: Record<string, string | Uint8Array>) {
 		await seed(this.tmpPath, files);
+	}
+
+	async remove(files: string[]) {
+		await remove(this.tmpPath, files);
 	}
 
 	// eslint-disable-next-line @typescript-eslint/consistent-type-imports
--- a/packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
+++ b/packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
@@ -2,7 +2,7 @@ import assert from "node:assert";
 import crypto from "node:crypto";
 import { onTestFinished } from "vitest";
 import { generateResourceName } from "./generate-resource-name";
-import { makeRoot, remove, seed } from "./setup";
+import { makeRoot, removeFiles, seed } from "./setup";
 import {
 	runWrangler,
 	WRANGLER_IMPORT,
@@ -21,8 +21,8 @@ export class WranglerE2ETestHelper {
 		await seed(this.tmpPath, files);
 	}
 
-	async remove(files: string[]) {
-		await remove(this.tmpPath, files);
+	async removeFiles(files: string[]) {
+		await removeFiles(this.tmpPath, files);
 	}
 
 	// eslint-disable-next-line @typescript-eslint/consistent-type-imports
--- a/packages/wrangler/e2e/helpers/setup.ts
+++ b/packages/wrangler/e2e/helpers/setup.ts
@@ -1,5 +1,5 @@
 import { mkdtempSync } from "node:fs";
-import { mkdir, writeFile } from "node:fs/promises";
+import { mkdir, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 
@@ -18,5 +18,15 @@ export async function seed(
 		const filePath = path.resolve(root, name);
 		await mkdir(path.dirname(filePath), { recursive: true });
 		await writeFile(filePath, contents);
+	}
+}
+
+// Removes the given files from the `root` directory on the file system.
+export async function remove(root: string, files: string[]) {
+	for (const name of files) {
+		const filePath = path.resolve(root, name);
+		if (filePath) {
+			await rm(filePath);
+		}
 	}
 }
--- a/packages/wrangler/e2e/helpers/setup.ts
+++ b/packages/wrangler/e2e/helpers/setup.ts
@@ -22,7 +22,7 @@ export async function seed(
 }
 
 // Removes the given files from the `root` directory on the file system.
-export async function remove(root: string, files: string[]) {
+export async function removeFiles(root: string, files: string[]) {
 	for (const name of files) {
 		const filePath = path.resolve(root, name);
 		if (filePath) {
--- a/packages/wrangler/src/api/startDevWorker/ConfigController.ts
+++ b/packages/wrangler/src/api/startDevWorker/ConfigController.ts
@@ -322,6 +322,7 @@ export class ConfigController extends Controller<ConfigControllerEventMap> {
 		if (configPath) {
 			this.#configWatcher = watch(configPath, {
 				persistent: true,
+				ignoreInitial: true,
 			}).on("change", async (_event) => {
 				logger.log(`${path.basename(configPath)} changed...`);
 				assert(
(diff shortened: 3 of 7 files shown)


packages/wrangler/src/dev.tsx
```diff
@@ -811,15 +818,52 @@ export async function startDev(args: StartDevOptions) {
...
 		if (config.configPath && !args.experimentalDevEnv) {
-			watcher = watch(config.configPath, {
+			configFileWatcher = watch(config.configPath, {
 				persistent: true,
 			}).on("change", async (_event) => {
-				// TODO: Do we need to handle different `_event` types differently?
-				//       e.g. what if the file is deleted, or added?
-				config = readConfig(configPath, args);
-				if (config.configPath) {
+				try {
```
CarmenPopoviciu, line 824: The two fns that can potentially throw here are `readConfig` and `processExperimentalAssetsArg`. Since we're in watch mode, we ideally don't want to error and exit every time there's a config file read error or we couldn't properly resolve the assets directory path. The better user experience IMO is to log the error and allow users to fix as they go, without having them re-start the dev process every single time. This brings the watch mode experience in line with `pages dev`, which we recently refactored for precisely the same reason.

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -352,33 +374,46 @@ export class ConfigController extends Controller<ConfigControllerEventMap> {
...
-					: input.dev?.server?.secure
-						? "https"
-						: "http",
-		});
-		void this.#ensureWatchingConfig(fileConfig.configPath);
-
-		const resolvedConfig = await resolveConfig(fileConfig, input);
-		if (signal.aborted) {
-			return;
+		try {
```
CarmenPopoviciu, line 377: basically for the same reason as in https://github.com/cloudflare/workers-sdk/pull/6574/files#r1733250847. But I would like some sanity check on whether this gets executed only in watch mode. If not, we might need to rethink the logic here, in case throwing and exiting the dev process is expected to happen in some use cases

packages/wrangler/src/dev.tsx
```diff
@@ -811,15 +818,52 @@ export async function startDev(args: StartDevOptions) {
...
+					if (!config.configPath) {
+						return;
+					}
+
 					logger.log(`${path.basename(config.configPath)} changed...`);
+
+					/*
+					 * Handle experimental assets watching on config file changes
+					 *
+					 * 1. if experimental assets was specified via CLI args, config file
```
CarmenPopoviciu, line 837: I wonder if for this use case we shouldn't show a warning or log message reminding users that they are currently running `wrangler dev` with the `-x--assets`/`--experimental-assets` flag, and therefore config file changes related to `experimental_assets` will not be reflected in the current dev process 🤔

CarmenPopoviciu: as per our async convo, I will put smth in place here

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -388,6 +423,7 @@ export class ConfigController extends Controller<ConfigControllerEventMap> {
 	async teardown() {
 		logger.debug("ConfigController teardown beginning...");
 		await this.#configWatcher?.close();
+		await this.#assetsWatcher?.close();
```
RamIdeas, line 426: If the first call rejects, the 2nd won't be called.

We should wrap these in Promise.allSettled

packages/wrangler/src/dev.tsx
```diff
@@ -411,12 +414,15 @@ This is currently not supported 😭, but we think that we'll get it to work soo
...
+			configFileWatcher = devInstance.configFileWatcher;
+			assetsWatcher = devInstance.assetsWatcher;
+
 			const { waitUntilExit } = devInstance.devReactElement;
 			await waitUntilExit();
 		}
 	} finally {
-		await watcher?.close();
+		await configFileWatcher?.close();
+		await assetsWatcher?.close();
```
RamIdeas, line 425: If the first call rejects, the 2nd won't be called.

We should wrap these in Promise.allSettled

packages/wrangler/src/dev.tsx
```diff
@@ -932,18 +976,33 @@ export async function startDev(args: StartDevOptions) {
...
 				devReactElement.unmount();
-				await watcher?.close();
+				await configFileWatcher?.close();
+				await assetsWatcher?.close();
 			},
 		};
 	} catch (e) {
-		await watcher?.close();
+		await configFileWatcher?.close();
+		await assetsWatcher?.close();
```
RamIdeas, line 1005: If the first call rejects, the 2nd won't be called.

We should wrap these in Promise.allSettled

packages/wrangler/src/dev.tsx
```diff
@@ -932,18 +976,33 @@ export async function startDev(args: StartDevOptions) {
...
 		return {
 			devReactElement,
-			watcher,
+			configFileWatcher,
+			assetsWatcher,
 			stop: async () => {
 				devReactElement.unmount();
-				await watcher?.close();
+				await configFileWatcher?.close();
+				await assetsWatcher?.close();
```
RamIdeas, line 1000: If the first call rejects, the 2nd won't be called.

We should wrap these in Promise.allSettled

packages/wrangler/src/api/startDevWorker/ConfigController.ts
CarmenPopoviciu: ah yes, you're totes right

packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
```diff
@@ -21,6 +21,10 @@ export class WranglerE2ETestHelper {
 		await seed(this.tmpPath, files);
 	}
 
+	async remove(files: string[]) {
```
RamIdeas, line 24: nit: `removeFiles`

packages/wrangler/e2e/helpers/setup.ts
```diff
@@ -20,3 +20,13 @@ export async function seed(
 		await writeFile(filePath, contents);
 	}
 }
+
+// Removes the given files from the `root` directory on the file system.
+export async function remove(root: string, files: string[]) {
```
RamIdeas, line 25: nit: `removeFiles`

packages/wrangler/src/dev.tsx
CarmenPopoviciu: ✅

packages/wrangler/src/dev.tsx
CarmenPopoviciu: ✅

packages/wrangler/src/dev.tsx
CarmenPopoviciu: ✅

packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
CarmenPopoviciu: ✅

RamIdeas (approved):

PR #6598, opened 2024-08-29
fix: allow encoded characters in paths for workers + assets

emily-shen:

~Waiting on EWC work, **will break existing deployments!!**~

Fixes #[WC-2597](https://jira.cfdata.org/browse/WC-2597)
Out of scope for this PR: moving duplicated code into shared folder. 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: no relevant e2e tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: not a public facing change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: unreleased feat

### DIFF
--- a/packages/miniflare/src/workers/assets/assets-kv.worker.ts
+++ b/packages/miniflare/src/workers/assets/assets-kv.worker.ts
@@ -17,9 +17,7 @@ export default <ExportedHandler<Env>>{
 			return new Response(message, { status: 405, statusText: message });
 		}
 
-		// don't uri decode pathname, because we encode the filepath before hashing
 		const pathHash = new URL(request.url).pathname.substring(1);
-
 		const entry = env.ASSETS_REVERSE_MAP[pathHash];
 		if (entry === undefined) {
 			return new Response("Not Found", { status: 404 });
--- a/packages/workers-shared/asset-worker/src/assets-manifest.ts
+++ b/packages/workers-shared/asset-worker/src/assets-manifest.ts
@@ -17,7 +17,6 @@ export class AssetsManifest {
 	}
 
 	async get(pathname: string) {
-		// encode pathname to match manifest encoding
 		const pathHash = await hashPath(pathname);
 		const entry = binarySearch(
 			new Uint8Array(this.data, HEADER_SIZE),
--- a/fixtures/workers-with-assets/public/about/[fünky].txt
+++ b/fixtures/workers-with-assets/public/about/[fünky].txt
@@ -1,0 +1,1 @@
+This should work.
--- a/fixtures/workers-with-assets/tests/index.test.ts
+++ b/fixtures/workers-with-assets/tests/index.test.ts
@@ -125,6 +125,15 @@ describe("[Workers + Assets] dynamic site", () => {
 		expect(response.statusText).toBe("Method Not Allowed");
 	});
 
+	it("should work with encoded path names", async ({ expect }) => {
+		let response = await fetch(
+			`http://${ip}:${port}/about/%5Bf%C3%BCnky%5D.txt`
+		);
+		let text = await response.text();
+		expect(response.status).toBe(200);
+		expect(text).toContain(`This should work.`);
+	});
+
 	it("should forward all request types to the user Worker if there are *not* assets on that route", async ({
 		expect,
 	}) => {
--- a/packages/workers-shared/asset-worker/src/index.ts
+++ b/packages/workers-shared/asset-worker/src/index.ts
@@ -50,11 +50,7 @@ export default class extends WorkerEntrypoint<Env> {
 
 	private async getAssetEntry(request: Request) {
 		const url = new URL(request.url);
-		let { pathname } = url;
-
 		const assetsManifest = new AssetsManifest(this.env.ASSETS_MANIFEST);
-		pathname = globalThis.decodeURIComponent(pathname);
-
-		return await assetsManifest.get(pathname);
+		return await assetsManifest.get(url.pathname);
 	}
 }
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -4260,35 +4260,51 @@ addEventListener('fetch', event => {});`
 			);
 		});
 
-		it("should encode file paths", async () => {
+		it("should be able to upload files with special characters in filepaths", async () => {
 			// NB windows will disallow these characters in file paths anyway < > : " / \ | ? *
 			const assets = [
 				{ filePath: "file-1.txt", content: "Content of file-1" },
-				{ filePath: "boop/file#1.txt", content: "Content of file-1" },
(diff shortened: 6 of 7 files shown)


packages/workers-shared/asset-worker/src/assets-manifest.ts
```diff
@@ -96,3 +96,11 @@ const contentHashToKey = (buffer: Uint8Array) => {
 	).slice(0, CONTENT_HASH_SIZE);
 	return [...contentHash].map((b) => b.toString(16).padStart(2, "0")).join("");
 };
+
+const encodeFilePath = (filePath: string) => {
+	const encodedPath = filePath
+		.split("/")
+		.map((segment) => encodeURIComponent(segment))
+		.join("/");
```
penalosa, line 104: Why do we need to split on `/`? Does calling `encodeURIComponent` on the whole path not work?

RamIdeas:
```
> encodeURI('/about/fünky:path.txt')
'/about/f%C3%BCnky:path.txt'
> encodeURIComponent('/about/fünky:path.txt')
'%2Fabout%2Ff%C3%BCnky%3Apath.txt'
> '/about/fünky:path.txt'.split('/').map(x => encodeURIComponent(x)).join('/')
'/about/f%C3%BCnky%3Apath.txt'
> '/about/fünky:path.txt'.split('/').map(x => encodeURIComponent(x)).join('/') === encodeURI('/about/fünky:path.txt')
false
```

RamIdeas (approved):

packages/workers-shared/asset-worker/src/assets-manifest.ts
RamIdeas: Considering this needs to match how the asset manifest is constructed, it would be great if this could be colocated with that

penalosa: Sorry, I don't think I was super clear—I know it's different, but I'm wondering why we can't just encode the whole thing?

penalosa: As in, just `encodeURIComponent(path)` and only storing the encoded paths. Why do we need slashes specifically to be preserved?

RamIdeas: Seems like this manifest is sent to an API endpoint so I imagine the endpoint itself has made this decision https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/experimental-assets.ts#L65

If we want to change this, we would need to do so before users are uploading assets because the blob has these keys persisted remotely

penalosa: Hopefully we still can, since we're pre release. Cc @GregBrimble

emily-shen: Currently EWC is checking for a `/`at the start of manifest keys but Greg says he’s happy to change that? So I’ll tidy this PR up in a bit and encode the whole thing as suggested :)

penalosa (approved):

packages/miniflare/src/workers/assets/assets-kv.worker.ts
```diff
@@ -17,9 +17,7 @@ export default <ExportedHandler<Env>>{
 			return new Response(message, { status: 405, statusText: message });
 		}
 
-		// don't uri decode pathname, because we encode the filepath before hashing
 		const pathHash = new URL(request.url).pathname.substring(1);
-
 		const entry = env.ASSETS_REVERSE_MAP[pathHash];
```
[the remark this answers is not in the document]
emily-shen: Yep, this is after the asset worker has gotten the pathname **with** leading /, and hashed it. In dev, miniflare sets up the `assets-kv-worker` which gets passed this pathHash and look it up in the reverse map to get it from the filesystem instead of an actual kv namespace. And the leading '/' in this file is not actually part of the pathHash so its discarded.


/some-path -> asdfasddsfdasf-pathhash -> /asdfasddsfdasf-pathhash -> asdfasddsfdasf-pathhash

packages/workers-shared/asset-worker/src/assets-manifest.ts
RamIdeas: One more. Is there a reason we can't use `encodeURI` instead of `encodeURIComponent`? Might make reading the manifest by a human slightly easier if the slashes aren't also encoded eg `/file-1.txt` instead of `%2Ffile-1.txt`

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -221,11 +221,7 @@ const hashPath = async (path: string) => {
 };
 
 const encodeFilePath = (filePath: string) => {
-	const encodedPath = filePath
-		.split(path.sep)
-		.map((segment) => encodeURIComponent(segment))
-		.join("/");
-	return "/" + encodedPath;
+	return encodeURIComponent("/" + filePath.split(path.sep).join("/"));
```
RamIdeas, line 224: Can this be encodeURl now since it isn't operating on each segment anymore?

```suggestion
	return encodeURI("/" + filePath.split(path.sep).join("/"));
```

packages/workers-shared/asset-worker/src/assets-manifest.ts
RamIdeas: Another question: why do we need to encode these paths before we hash them?

packages/miniflare/src/plugins/assets/index.ts
emily-shen: encodeURI doesn't encode characters like '/' and '#' while encodeURIComponent does

packages/workers-shared/asset-worker/src/assets-manifest.ts
emily-shen: EncodeURI doesn't encode characters like '?' and '#' while encodeURIComponent does. We decided to allow users to upload filenames with those characters if they really wanted to do that - (e.g. the file `foo?bar.txt` would be accessible at `/foo%3Fbar.txt`). To allow that, we need use encodeURIComponent and also encode before hashing so that the hashed path corresponds to `/foo%3Fbar.txt`.

RamIdeas:
> Sorry for going around in circles - going back to (basically) the original fix because:
> 
> 1. frameworks will request something like`/about/%5Bpage%5D` not `%2Fabout%2F%5Bpage%5D` and encoding each segment means we can handle this without asset worker having to do anything to an incoming request path.
> 2. we want to allow filepaths with `?` or `#` as long as theyre encoded and not doing encodeURIComponent on each segment doesn't do this properly.

I think this is the right decision 👍  Please document this reasoning in a comment with the (shared) encode/decode functions

PR #6603, opened 2024-08-29
[wrangler] log version preview url when previews exist

taylorlee:

New feature - log version preview url when previews exist. This feature is currently restricted, but I'm adding wrangler support now to enable internal testing.

- Tests
  - [x] Included
  - [x] Not required because: e2e account will not have previews yet, so current upload tests are correct in not printing previews.
  - [x] Included
- Public documentation
  - [x] Not necessary because: Docs will be added when feature is released externally.

### DIFF
--- a/.changeset/funny-comics-lick.md
+++ b/.changeset/funny-comics-lick.md
@@ -1,0 +1,13 @@
+---
+"wrangler": minor
+---
+
+feature: log version preview url when previews exist
+
+The version upload API returns a field indicating whether
+a preview exists for that version. If a preview exists and
+workers.dev is enabled, wrangler will now log the full
+URL on version upload.
+
+This does not impact wrangler deploy, which only prints the
+workers.dev route of the latest deployment.
--- a/.changeset/modern-scissors-relax.md
+++ b/.changeset/modern-scissors-relax.md
@@ -1,0 +1,18 @@
+---
+"wrangler": patch
+---
+
+chore: fix version upload log order
+
+Previously deploy prints:
+upload timings
+deploy timings
+current version id
+
+while version upload prints:
+worker version id
+upload timings
+
+This change makes version upload more similar to deploy by printing
+version id after upload, which also makes more sense, as version ID can
+only be known after upload has finished.
--- a/packages/wrangler/e2e/versions.test.ts
+++ b/packages/wrangler/e2e/versions.test.ts
@@ -61,8 +61,8 @@ describe("versions deploy", { timeout: TIMEOUT }, () => {
 		expect(normalize(upload.stdout)).toMatchInlineSnapshot(`
 			"Total Upload: xx KiB / gzip: xx KiB
 			Worker Startup Time: (TIMINGS)
-			Worker Version ID: 00000000-0000-0000-0000-000000000000
 			Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
+			Worker Version ID: 00000000-0000-0000-0000-000000000000
 			To deploy this version to production traffic use the command wrangler versions deploy
 			Changes to non-versioned settings (config properties 'logpush' or 'tail_consumers') take effect after your next deployment using the command wrangler versions deploy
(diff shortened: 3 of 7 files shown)


packages/wrangler/src/versions/upload.ts
```diff
@@ -538,6 +543,14 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 
 	logger.log("Worker Version ID:", versionId);
 
+	if (versionId && config.workers_dev && hasPreview) {
+		const userSubdomain = await getWorkersDevSubdomain(accountId);
```
taylorlee, line 547: When thinking more about this - I think I should also be checking the subdomain status of the worker. Version upload doesn't apply the workers_dev setting from the wrangler.toml, so using that as a source of truth could be misleading.

taylorlee: Fixed!

taylorlee: We may end up wanting to add another warning if workers_dev is out of sync the current output note about "triggers deploy" accurate covers this scenario as well.

taylorlee: "Changes to triggers (routes, custom domains, cron schedules, etc) must be applied with the command wrangler triggers deploy"

Though maybe we should specifically call out workers_dev rather than letting it fall under "etc"?

packages/wrangler/e2e/versions.test.ts
```diff
@@ -61,8 +61,8 @@ describe("versions deploy", { timeout: TIMEOUT }, () => {
 		expect(normalize(upload.stdout)).toMatchInlineSnapshot(`
 			"Total Upload: xx KiB / gzip: xx KiB
 			Worker Startup Time: (TIMINGS)
-			Worker Version ID: 00000000-0000-0000-0000-000000000000
 			Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
+			Worker Version ID: 00000000-0000-0000-0000-000000000000
```
WalshyDev, line 65: Never realised that version upload had this before -- deploy is correct (uploaded then version)

WalshyDev (approved):

.changeset/funny-comics-lick.md
```diff
@@ -0,0 +1,13 @@
+---
+"wrangler": minor
+---
+
+feature: log version preview url when previews exist
+
+The version upload API returns a field indicating whether
+a preview exists for that version. If a preview exists and
+workers.dev is enabled, wrnagler will now log the full
```
RamIdeas, line 9: typo
```suggestion
workers.dev is enabled, wrangler will now log the full
```

.changeset/modern-scissors-relax.md
```diff
@@ -0,0 +1,18 @@
...
+upload timings
+deploy timings
+current version id
+
+while version upload prints:
+worker version id
+upload timings
+
+This change makes version upload more similar to deploy by printing
+version after upload, which also makes more sense, as version ID can
```
RamIdeas, line 17:
```suggestion
version id after upload, which also makes more sense, as version ID can
```

packages/wrangler/src/__tests__/versions/versions.upload.test.ts
```diff
@@ -98,8 +111,74 @@ describe("versions upload", () => {
...
+	});
+
+	test("should print preview url if version has preview", async () => {
+		mockGetScript();
+		mockUploadVersion(true);
+		mockGetWorkerSubdomain(true);
+		mockSubDomainRequest();
+
+		// Setup
+		fs.mkdirSync("./versions-upload-test-worker", { recursive: true });
```
RamIdeas, line 126: what's this for? there is already a `runInTmpDir()` for these tests

RamIdeas (approved):

packages/wrangler/src/__tests__/versions/versions.upload.test.ts
taylorlee: I removed the redundant mkdirSync and absorbed it into the test refactor commit.

PR #6605, opened 2024-08-30
Update non-versioned script settings on deploy

WalshyDev:

Fixes tail consumer + logpush not updating in new deploy path

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: No change to output or anything
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: no change

### DIFF
--- a/packages/wrangler/src/deploy/deploy.ts
+++ b/packages/wrangler/src/deploy/deploy.ts
@@ -46,6 +46,7 @@ import {
 	maybeRetrieveFileSourceMap,
 } from "../sourcemap";
 import triggersDeploy from "../triggers/deploy";
+import { patchNonVersionedScriptSettings } from "../versions/api";
 import { confirmLatestDeploymentOverwrite } from "../versions/deploy";
 import { getZoneForRoute } from "../zones";
 import type { Config } from "../config";
@@ -742,6 +743,8 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 		if (props.dryRun) {
 			printBindings({ ...withoutStaticAssets, vars: maskedVars });
 		} else {
+			assert(accountId, "Missing accountId");
+
 			await ensureQueuesExistByConfig(config);
 			let bindingsPrinted = false;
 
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -742,6 +743,8 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 		if (props.dryRun) {
 			printBindings({ ...withoutStaticAssets, vars: maskedVars });
 		} else {
+			assert(accountId, "Missing accountId");
```
RamIdeas, line 746: This shouldn't be an assertion. If a user can get themselves in this situation, throw a UserError. An AssertionError is a mistake in our logic and will be sent to Sentry whereas a UserError won't.

WalshyDev: The assertion was just moved from the bottom to fix the type - it probably is something we want in Sentry tbf. Users _shouldn't_ be able to get to the deploy path without an account ID -- we should be prompting or accepting the config value

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -786,6 +791,12 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 						}
 					);
 
+					// Update tail consumers and logpush settings
+					await patchNonVersionedScriptSettings(accountId, scriptName, {
```
RamIdeas, line 795: Can/should we run the deploy and patch requests in parallel?

RamIdeas (approved):

packages/wrangler/src/deploy/deploy.ts
RamIdeas: I see, sorry I thought you introduced it. It should've been an UserError already. The config is already considered. It can only be undefined if not specified in args OR config.

packages/wrangler/src/deploy/deploy.ts
WalshyDev: Since this does a deploy we'd have concurrency issues sadly where they'd try to block on eachother and probably more often than not timeout

PR #6606, opened 2024-08-30
Add new_sqlite_classes migration for Durable Objects

joshthoward:

- Tests
  - [ ] TODO (before merge)
  - [X] Included
  - [ ] Not necessary because:
  - [X] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Included
  - [X] Not necessary because: Will release documentation with birthday week announcement. We are still gating access in EWC.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Not necessary because: Will release documentation with birthday week announcement. We are still gating access in EWC.

### DIFF
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -3144,8 +3144,14 @@ const validateMigrations: ValidatorFn = (diagnostics, field, value) => {
 
 	let valid = true;
 	for (let i = 0; i < rawMigrations.length; i++) {
-		const { tag, new_classes, renamed_classes, deleted_classes, ...rest } =
-			rawMigrations[i];
+		const {
+			tag,
+			new_classes,
+			new_sqlite_classes,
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/config/validation.ts
```diff
@@ -3172,6 +3178,14 @@ const validateMigrations: ValidatorFn = (diagnostics, field, value) => {
 				"string"
 			) && valid;
 
+		valid =
```
andyjessop, line 3181: Presumably we also need to validate that every string in `new_sqlite_classes` matches an existing durable object name?

joshthoward: This is validated in EWC. I am copying the existing validation that is present for new_classes so that they have the same behavior.

andyjessop (approved):

PR #6612, opened 2024-09-01
enable hyperdrive bindings in `getPlatformProxy`

dario-piotrowicz:

This PR simply adds hyperdrive bindings support to Miniflare's `getBindings` and consequently to `getPlatformProxy`.

The hyperdrive proxies have overrides to make them usable within node.js, since the hyperdrive values can only work inside the workerd process itself.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/16706
  - [ ] Not necessary because

### DIFF
--- a/packages/miniflare/src/plugins/hyperdrive/index.ts
+++ b/packages/miniflare/src/plugins/hyperdrive/index.ts
@@ -1,7 +1,7 @@
 import assert from "node:assert";
 import { z } from "zod";
 import { Service, Worker_Binding } from "../../runtime";
-import { Plugin } from "../shared";
+import { Plugin, ProxyNodeBinding } from "../shared";
 
 export const HYPERDRIVE_PLUGIN_NAME = "hyperdrive";
 
@@ -90,8 +90,24 @@ export const HYPERDRIVE_PLUGIN: Plugin<typeof HyperdriveInputOptionsSchema> = {
 			}
 		);
 	},
-	getNodeBindings() {
-		return {};
+	getNodeBindings(options) {
+		return Object.fromEntries(
+			Object.entries(options.hyperdrives ?? {}).map(([name, url]) => {
+				const connectionOverrides: Record<string | symbol, string | number> = {
+					connectionString: `${url}`,
+					port: Number.parseInt(url.port),
+					host: url.hostname,
+				};
+				const proxyNodeBinding = new ProxyNodeBinding({
+					get(target, prop) {
+						return prop in connectionOverrides
+							? connectionOverrides[prop]
+							: target[prop];
+					},
+				});
+				return [name, proxyNodeBinding];
+			})
+		);
 	},
 	async getServices({ options }) {
 		return Object.entries(options.hyperdrives ?? {}).map<Service>(
--- a/packages/miniflare/test/plugins/hyperdrive/index.spec.ts
+++ b/packages/miniflare/test/plugins/hyperdrive/index.spec.ts
@@ -1,5 +1,6 @@
 import test from "ava";
 import { Miniflare, MiniflareOptions } from "miniflare";
+import type { Hyperdrive } from "@cloudflare/workers-types/experimental";
 
 test("fields match expected", async (t) => {
 	const connectionString = `postgresql://user:password@localhost:5432/database`;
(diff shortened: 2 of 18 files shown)


packages/miniflare/src/plugins/hyperdrive/index.ts
```diff
@@ -90,8 +90,24 @@ export const HYPERDRIVE_PLUGIN: Plugin<typeof HyperdriveInputOptionsSchema> = {
...
+				const proxyNodeBinding = new ProxyNodeBinding({
+					get(target, prop) {
+						return prop in connectionOverrides
+							? connectionOverrides[prop]
+							: target[prop];
+					},
+				});
+				return [name, proxyNodeBinding];
+			})
+		);
```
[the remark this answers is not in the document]
penalosa: However, I don't really understand why we need this new `ProxyNodeBinding()` class. What about just:

```ts
getNodeBindings(options) {
		return Object.fromEntries(
			Object.entries(options.hyperdrives ?? {}).map(([name, url]) => {
				return [name, {
					connectionString: url.href,
					port: url.port,
					host: url.host,
				}];
			})
		);
```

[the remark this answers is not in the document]
dario-piotrowicz: The point of `ProxyNodeBinding` is that I wanted to add a new way that can generally be applied to all bindings to allow the overriding of some property accessors.

In this specific case we use the class to override `connectionString`, `port` and `host` but do allow everything else to pass through, which, here is just `connect` method (which I haven't tested yet to be honest).

I think this could be quite nice as it will enable us to simply override any proxy behavior in the future as (/if) the need arises.

packages/miniflare/test/plugins/hyperdrive/index.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { Hyperdrive } from "@cloudflare/workers-types/experimental";
```
dario-piotrowicz, line 1:
```suggestion
import type { Hyperdrive } from "@cloudflare/workers-types/experimental";
```

packages/miniflare/test/plugins/hyperdrive/index.spec.ts
```diff
@@ -34,6 +35,30 @@ test("fields match expected", async (t) => {
 	t.is(hyperdrive.port, 5432);
 });
 
+test("fields in binding proxy match expected", async (t) => {
+	const connectionString = `postgresql://user:password@localhost:5432/database`;
```
dario-piotrowicz, line 39:
```suggestion
	const connectionString = 'postgresql://user:password@localhost:5432/database';
```

packages/miniflare/test/plugins/hyperdrive/index.spec.ts
```diff
@@ -34,6 +35,30 @@ test("fields match expected", async (t) => {
...
+	const mf = new Miniflare({
+		modules: true,
+		script: `export default { fetch() {} }`,
```
dario-piotrowicz, line 42:
```suggestion
		script: 'export default { fetch() {} }',
```

packages/wrangler/e2e/get-platform-proxy.test.ts
```diff
@@ -1,57 +1,134 @@
...
+						compatibility_date = "2024-08-20"
+						compatibility_flags = ["nodejs_compat"]
+
+						[[hyperdrive]]
+						binding = "HYPERDRIVE"
+						id = "hyperdrive_id"
+						localConnectionString = "postgresql://user:%21pass@127.0.0.1:${port}/some_db"
+				`,
+				"index.mjs": dedent/*javascript*/ `
+						import { getPlatformProxy } from "${WRANGLER_IMPORT}"
```
dario-piotrowicz, line 90:
```suggestion
						import { getPlatformProxy } from "${WRANGLER_IMPORT}";
```

packages/wrangler/e2e/get-platform-proxy.test.ts
```diff
@@ -1,57 +1,134 @@
...
+				const socketDataMsgPromise = new Promise<string>((resolve, _) => {
+					server.on("connection", (sock) => {
+						sock.on("data", (data) => {
+							resolve(new TextDecoder().decode(data));
+							server.close();
+						});
+					});
+				});
+
+				execSync(`node index.mjs`, {
```
dario-piotrowicz, line 124:
```suggestion
				execSync('node index.mjs', {
```

packages/wrangler/e2e/get-platform-proxy.test.ts
```diff
@@ -1,57 +1,134 @@
...
+							"name": "hyperdrive-app",
+							"version": "0.0.0",
+							"private": true
+						}
+						`,
+			});
+		});
+
+		it.skipIf(
+			// in c3 this test fails for windows because of ECONNRESET issues
```
penalosa, line 110: I don't understand this comment—why is this test using C3?

dario-piotrowicz: sorry.... it's just a typo, I meant `ci`! 🤦

dario-piotrowicz: PS: I'm not completely sure why it fails in CI, but given the amount of flakiness I get when I run the e2es (unrelated to the changes here) plus the fact that I feel that windows in GitHub CI is always a bit error prone I didn't really explore this too much and I figured just skipping this for windows is the simplest path here, hopefully that is fine 🙂

penalosa (approved):

PR #6615, opened 2024-09-02
avoid potential double-install of create-cloudflare

RamIdeas:

When `wrangler init` delegates to C3, it did so via `npm create cloudflare@2.5.0`. C3's v2.5.0 was the first to include auto-update support to avoid `npx`'s potentially stale cache. But this also guaranteed a double install for users who do not have 2.5.0 cached. Now, wrangler delegates via `npm create cloudflare@^2.5.0` which should use the latest version cached on the user's system or install and use the latest v2.x.x.

For more context, see https://github.com/cloudflare/workers-sdk/pull/6415#discussion_r1739312919

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [x] Not required because: covered by other tests + the e2e tests delegate to c3 in a custom way
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: optimisation

### DIFF
--- a/.changeset/friendly-brooms-brush.md
+++ b/.changeset/friendly-brooms-brush.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+chore: avoid potential double-install of create-cloudflare
+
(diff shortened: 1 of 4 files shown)


andyjessop (approved):

Jamesgrac30 (commented):
500.0$

PR #6616, opened 2024-09-02
feat: `wrangler --dotenv` / `dotenv = true` populates `process.env` / `import.meta.env` / `env`

threepointone:
- `wrangler --dotenv` should read from `.env` and populate `process.env`/ `import.meta.env`/ `env` with the values.
- `wrangler --dotenv --env-file some/path/to/.env` should read from `some/path/to/.env` and populate `process.env`/ `import.meta.env`/ `env` with the values.
- `wrangler --dotenv --env xyz` should read from `.env.xyz` and populate `process.env`/ `import.meta.env`/ `env` with the values.
- `wrangler --penv KEY` should read `$KEY` from the actual environment, and set `process.env.KEY`/ `import.meta.env.KEY`/ `env.KEY` with that value. This will override values in the `.env` file.
- `wrangler --penv KEY=value` should set `process.env.KEY`/ `import.meta.env.KEY`/ `env.KEY` with that value. This will override values in the `.env` file.
- This will mean that we can also write `wrangler --penv KEY=$SOMETHING` and it will work as expected; reading the value of $SOMETHING from the environment, and setting it to `process.env.KEY`/ `import.meta.env.KEY`/ `env.KEY`.

Fixes https://github.com/cloudflare/workers-sdk/issues/7516

An example of code that gets simplified with this PR https://github.com/threepointone/remixflare/pull/2

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because: no changes to e2e behaviour 
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:


packages/wrangler/src/__tests__/dotenv.test.ts
```diff
@@ -0,0 +1,135 @@
...
+});
+
+describe("dotenv", () => {
+	it("should set process.env and import.meta.env when enabled in wrangler.toml", async () => {
+		writeWranglerToml({
+			main: "index.js",
+			dotenv: true,
+		});
+		fs.writeFileSync("index.js", `export default {};`);
+		fs.writeFileSync(".env", "TEST_VAR=test_value");
```
threepointone, line 25: hmm I need to verify whether we should automatically add quotes here.

threepointone: ok tested that we match dotenv behaviour correctly, they have to all be coerced into strings

packages/wrangler/src/dev.tsx
```diff
@@ -930,7 +958,7 @@ export async function startDev(args: StartDevOptions) {
 					minify={args.minify ?? configParam.minify}
 					nodejsCompatMode={nodejsCompatMode}
 					build={configParam.build || {}}
-					define={{ ...configParam.define, ...cliDefines }}
+					define={{ ...processEnvValues, ...configParam.define, ...cliDefines }}
```
RamIdeas, line 961: This also needs to be added to the`--x-dev-env` codepath

https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/api/startDevWorker/ConfigController.ts#L250

threepointone: thanks! updated the PR with support for `--x-dev-env`, and e2e tests

southpolesteve:
Thanks everyone following this thread! Special shoutout to @threepointone for pioneering this work and proving out the concept.

After internal discussions about environment variable handling, we've recognized the need for a more comprehensive platform-wide solution. While bundling env vars offers immediate benefits, it introduces potential confusion and edge cases since these variables are set at **build time** rather than **runtime** (which differs from typical env var behavior).

We want to implement this properly, so we'll be looking at a more integrated solution in 2025. As always, we welcome your feedback and insights.

vladshcherbin:
@southpolesteve was any solution found to have build time env variables?

southpolesteve:
@vladshcherbin We added support for process.env into the runtime. Does that fit your use case? https://developers.cloudflare.com/changelog/2025-03-11-process-env-support/

We're also starting to look at .env in local dev. Have an RFC for that here https://github.com/cloudflare/workers-sdk/discussions/9394

vladshcherbin:
@southpolesteve somewhat, thank you.

Need to test locally. I was using both `.env` and `.dev.vars` with flustration - hopefully it's improved. I also just recently discovered `import { env } from 'cloudflare:workers'` - go figure out what to use now 🙈

The main question I have, can wrangler strip functions during build time according to this variables?

```ts
// should be removed everywhere except staging
if (process.env.ENVIRONMENT === 'staging') {
  doSomeCrazyStruff()
}
```

Meanwhile I've tried `vite` with [import.meta.env](https://vite.dev/guide/env-and-mode.html#built-in-constants) with esbuild bundler and it removed this code. I also don't know what's suggested to use now, pure wrangler or vite wrapper 🙈

southpolesteve:
Yeah I realize its a bit confusing but vite and Wrangler both have some advantages. We have some docs that address it: https://developers.cloudflare.com/workers/development-testing/wrangler-vs-vite/

PR #6620, opened 2024-09-03
fix: don't warn about `node:async_hooks` if `nodejs_als` is set

### LINKED ISSUE #6011: 🐛 BUG: wrangler does not mark `node:async_hooks` as external when `nodejs_als` compatibility flag is used

Workers Runtime, Wrangler core

3.60.2

18

macOS 14.5

### Observed behavior
Please describe.

`wrangler.toml`
```toml
#: schema ../node_modules/wrangler/config-schema.json

compatibility_date = "2024-05-24"
main = "src/worker.ts"
compatibility_flags = ["nodejs_als"]
```

`src/worker.ts`
```ts
import { AsyncLocalStorage } from 'node:async_hooks'
```

### Expected behavior

No warning is generated, since the `"nodejs_als"` compatibility flag is specified in `wrangler.toml`.

### Actual behavior

```
▲ [WARNING] The package "node:async_hooks" wasn't found on the file system but is built into node.
```

from [kian](https://ptb.discord.com/channels/595317990191398933/1214571685424930846/1250341639939952701) on Cloudflare's Discord:
> Wrangler has no idea about that flag so it doesn’t know what to do with it. I'd open an issue

petebacondarwin:

Fixes #6011

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/workers-shared/package.json
+++ b/packages/workers-shared/package.json
@@ -33,7 +33,7 @@
 		"clean": "rimraf dist",
 		"dev": "pnpm run clean && concurrently -n bundle:asset-worker,bundle:router-worker -c blue,magenta \"pnpm run bundle:asset-worker --watch\" \"pnpm run bundle:router-worker --watch\"",
 		"test": "vitest",
-		"test:ci": "pnpm run test"
+		"test:ci": "pnpm run test run"
 	},
 	"devDependencies": {
 		"@cloudflare/eslint-config-worker": "workspace:*",
--- a/packages/wrangler/src/deployment-bundle/esbuild-plugins/als-external.ts
+++ b/packages/wrangler/src/deployment-bundle/esbuild-plugins/als-external.ts
@@ -1,0 +1,13 @@
+import type { Plugin } from "esbuild";
+
+/**
+ * An esbuild plugin that will mark `node:async_hooks` imports as external.
+ */
+export const asyncLocalStoragePlugin: Plugin = {
+	name: "Mark async local storage imports as external plugin",
+	setup(pluginBuild) {
+		pluginBuild.onResolve({ filter: /^node:async_hooks/ }, () => {
+			return { external: true };
+		});
+	},
+};
--- a/packages/wrangler/src/deployment-bundle/esbuild-plugins/als-external.ts
+++ b/packages/wrangler/src/deployment-bundle/esbuild-plugins/als-external.ts
@@ -6,7 +6,7 @@ import type { Plugin } from "esbuild";
 export const asyncLocalStoragePlugin: Plugin = {
 	name: "Mark async local storage imports as external plugin",
 	setup(pluginBuild) {
-		pluginBuild.onResolve({ filter: /^node:async_hooks/ }, () => {
+		pluginBuild.onResolve({ filter: /^node:async_hooks(\/|$)/ }, () => {
 			return { external: true };
 		});
 	},
--- a/.changeset/silly-ways-brake.md
+++ b/.changeset/silly-ways-brake.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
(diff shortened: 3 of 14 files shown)


packages/wrangler/src/deployment-bundle/esbuild-plugins/als-external.ts
```diff
@@ -0,0 +1,13 @@
+import type { Plugin } from "esbuild";
+
+/**
+ * An esbuild plugin that will mark `node:async_hooks` imports as external.
+ */
+export const asyncLocalStoragePlugin: Plugin = {
+	name: "Mark async local storage imports as external plugin",
+	setup(pluginBuild) {
+		pluginBuild.onResolve({ filter: /^node:async_hooks/ }, () => {
```
RamIdeas, line 9:
```suggestion
		pluginBuild.onResolve({ filter: /^node:async_hooks$/ }, () => {
```

RamIdeas (approved):

packages/wrangler/src/deployment-bundle/esbuild-plugins/als-external.ts
petebacondarwin: I did this for the potential future case of there being a sub entry point to `async_hooks`. Since we would also want this plugin to kick in for imports like:

```ts
import {...} from "node:async_hooks/blah";
```

Do you feel that we should not support this @RamIdeas ?

petebacondarwin: I guess there is the fear that a new module called `node:async_hooks_2` would also match this...

I could tighten this up a bit...?

RamIdeas: I was thinking the latter but I guess the former is equally likely. Since we're only marking as external, I think we can leave it as-is 👍

petebacondarwin: Doh! I updated it now 😆

packages/workers-shared/package.json
```diff
@@ -33,7 +33,7 @@
 		"clean": "rimraf dist",
 		"dev": "pnpm run clean && concurrently -n bundle:asset-worker,bundle:router-worker -c blue,magenta \"pnpm run bundle:asset-worker --watch\" \"pnpm run bundle:router-worker --watch\"",
 		"test": "vitest",
-		"test:ci": "pnpm run test"
+		"test:ci": "pnpm run test run"
```
RamIdeas, line 36: mistake?

packages/wrangler/src/deployment-bundle/esbuild-plugins/als-external.ts
```diff
@@ -6,7 +6,7 @@ import type { Plugin } from "esbuild";
 export const asyncLocalStoragePlugin: Plugin = {
 	name: "Mark async local storage imports as external plugin",
 	setup(pluginBuild) {
-		pluginBuild.onResolve({ filter: /^node:async_hooks/ }, () => {
+		pluginBuild.onResolve({ filter: /^node:async_hooks(\/|$)/ }, () => {
```
RamIdeas, line 9: 👍

packages/workers-shared/package.json
petebacondarwin: This is not a mistake as such because the original script is wrong; not adding `run` causes local runs of `test:ci` to hang since this one goes into interactive mode.

PR #6621, opened 2024-09-03
fix(wrangler): validate routes in dev and deploy

emily-shen:

Fixes [WC-2484](https://jira.cfdata.org/browse/WC-2484).
Also now errors if invalid route config is present in `wrangler dev`. 

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: no behaviour change

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -1455,11 +1455,12 @@ Update them to point to this script instead?`,
 					routes: [{ pattern: "*.example.com", custom_domain: true }],
 				});
 				writeWorkerSource();
-				await expect(
-					runWrangler("deploy ./index")
-				).rejects.toThrowErrorMatchingInlineSnapshot(
-					`[Error: Cannot use "*.example.com" as a Custom Domain; wildcard operators (*) are not allowed]`
-				);
+				await expect(runWrangler("deploy ./index")).rejects
+					.toThrowErrorMatchingInlineSnapshot(`
+					[Error: Invalid Routes:
+					*.example.com:
+					Wildcard operators (*) are not allowed in Custom Domains]
+				`);
 
 				writeWranglerToml({
 					routes: [
@@ -1469,11 +1470,12 @@ Update them to point to this script instead?`,
 				writeWorkerSource();
 				mockServiceScriptData({});
 
-				await expect(
-					runWrangler("deploy ./index")
-				).rejects.toThrowErrorMatchingInlineSnapshot(
-					`[Error: Cannot use "api.example.com/at/a/path" as a Custom Domain; paths are not allowed]`
-				);
+				await expect(runWrangler("deploy ./index")).rejects
+					.toThrowErrorMatchingInlineSnapshot(`
+					[Error: Invalid Routes:
+					api.example.com/at/a/path:
+					Paths are not allowed in Custom Domains]
+				`);
 			});
 
 			it("should not continue with publishing an override if user does not confirm", async () => {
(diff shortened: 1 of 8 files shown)


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -168,6 +168,60 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
...
+	hasExperimentalAssets: boolean
+) => {
+	const invalidRoutes: { [route: string]: string[] } = {};
+	for (const route of routes) {
+		if (typeof route !== "string" && route.custom_domain) {
+			if (route.pattern.includes("*")) {
+				invalidRoutes[route.pattern] = [
+					...(invalidRoutes[route.pattern] ?? []),
+					`Wildcard operators (*) are not allowed in Custom Domains`,
+				];
```
RamIdeas, line 182: nit
```suggestion
				invalidRoutes[route.pattern] ??= [];
				invalidRoutes[route.pattern].push(
					`Wildcard operators (*) are not allowed in Custom Domains`,
				);
```

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -168,6 +168,60 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
...
+				typeof route === "object" &&
+				hasExperimentalAssets &&
+				route.pattern.includes("/") &&
+				!route.pattern.endsWith("/*")
+			) {
+				invalidRoutes[route.pattern] = [
+					...(invalidRoutes[route.pattern] ?? []),
+					`Routes with paths (except for the wildcard /*) are not allowed in Workers + Assets projects`,
+				];
+			}
```
RamIdeas, line 212: Is this handling routes with a path AND ending in `/*` correctly? eg `/some/path/*` – is that allowed or not?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -168,6 +168,60 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
...
+				typeof route === "string" &&
+				hasExperimentalAssets &&
+				route.includes("/") &&
+				!route.endsWith("/*")
+			) {
+				invalidRoutes[route] = [
+					...(invalidRoutes[route] ?? []),
+					`Routes with paths (except for the wildcard /*) are not allowed in Workers + Assets projects`,
+				];
+			}
```
RamIdeas, line 201: This if-block is duplicated

packages/wrangler/src/deploy/deploy.ts
emily-shen: nope that should be allowed whoops - fixed

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -1514,6 +1516,41 @@ Update them to point to this script instead?`,
 			});
 		});
 
+		it("should error on routes with paths if experimental assets are present", async () => {
+			writeWranglerToml({
+				routes: [
+					"simple.co.uk/path",
+					"simple.co.uk/*",
+					"simple.co.uk",
```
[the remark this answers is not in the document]
emily-shen: whoops thanks, didn't realise that about bare domains. added some extra validation.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -168,6 +168,53 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
...
+		} else if (hasExperimentalAssets) {
+			const pattern = typeof route === "string" ? route : route.pattern;
+			const components = pattern.split("/");
+
+			if (
+				components.length === 1 ||
+				(components.length === 2 && components[1] === "")
+			) {
+				invalidRoutes[pattern] ??= [];
+				invalidRoutes[pattern].push(`Update the route to end with /*`);
```
CarmenPopoviciu, line 199: Having just the action stated here might be a bit too cryptic for users. Can we also tell them what routing rule they're actually breaking here? like we do in all other places.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -168,6 +168,53 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
 	return false;
 }
 
+export const validateRoutes = (
+	routes: Route[],
+	hasExperimentalAssets: boolean
+) => {
+	const invalidRoutes: { [route: string]: string[] } = {};
+	for (const route of routes) {
+		if (typeof route !== "string" && route.custom_domain) {
```
CarmenPopoviciu, line 177: do these apply for Workers with assets as well?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -168,6 +168,53 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
...
+					`Wildcard operators (*) are not allowed in Custom Domains`
+				);
+			}
+			if (route.pattern.includes("/")) {
+				invalidRoutes[route.pattern] ??= [];
+				invalidRoutes[route.pattern].push(
+					`Paths are not allowed in Custom Domains`
+				);
+			}
+		} else if (hasExperimentalAssets) {
```
CarmenPopoviciu, line 190: I'm wondering...if the purpose of this function to surface all route validation errors? Ideally IMHO we should aim for that, because surfacing these one at a time is rather frustrating. To that end, should this be an `if..else` block, or rather a series of `if` blocks?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -1514,6 +1516,63 @@ Update them to point to this script instead?`,
 			});
 		});
 
+		it("should error on routes with paths if experimental assets are present", async () => {
```
CarmenPopoviciu, line 1519: thank you for these :meouw-heart:

packages/wrangler/src/deploy/deploy.ts
emily-shen: the custom domain ones apply for everything, including workers + assets

packages/wrangler/src/deploy/deploy.ts
emily-shen: This should be surfacing everything at once already (see the tests) - unless i've misunderstood this comment...? 😅

packages/wrangler/src/deploy/deploy.ts
emily-shen: Ah I'm not sure what the phrasing should be - i stole this one from dash. tacking copy changes/discussion onto this https://jira.cfdata.org/browse/WC-2675

packages/wrangler/src/deploy/deploy.ts
CarmenPopoviciu: oh yeah you're right. gotta love tests. carry on 😋

packages/wrangler/src/deploy/deploy.ts
penalosa: Could we preserve the `Workers which have static assets cannot be routed on a URL which has a path component.` prefix here? The dash has a pattern of hiding the extra informational text when displaying errors (which is why it's so terse), but at the point that error is shown in dash that user should have been able to read the informational text.

penalosa (approved):

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -1514,6 +1516,63 @@ Update them to point to this script instead?`,
...
+				.toThrowErrorMatchingInlineSnapshot(`
+				[Error: Invalid Routes:
+				simple.co.uk/path:
+				Workers which have static assets cannot be routed on a URL which has a path component. Update the route to replace /path with /*
+
+				simple.co.uk/path/*:
+				Workers which have static assets cannot be routed on a URL which has a path component. Update the route to replace /path/* with /*
+
+				simple.co.uk/:
+				Update the route to end with /*
```
petebacondarwin, line 1550: I'm slightly worried about this error message. In non-assets Workers this is a valid route, right?
So the message should be similar to those above, no?
Currently I would be a bit confused to see this message.

packages/wrangler/src/deploy/deploy.ts
petebacondarwin: Ah yes, I just commented on this above. I agree it is not ideal copy.

petebacondarwin (approved):

packages/wrangler/src/__tests__/deploy.test.ts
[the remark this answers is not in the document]
emily-shen: added one without assets ;)

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -168,6 +168,55 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
...
+				invalidRoutes[pattern] ??= [];
+				invalidRoutes[pattern].push(
+					`Workers which have static assets must end with a wildcard path. Update the route to end with /*`
+				);
+			} else if (!(components.length === 2 && components[1] === "*")) {
+				invalidRoutes[pattern] ??= [];
+				invalidRoutes[pattern].push(
+					`Workers which have static assets cannot be routed on a URL which has a path component. Update the route to replace /${components.slice(1).join("/")} with /*`
+				);
+			}
```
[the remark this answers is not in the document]
RamIdeas: We agreed on a call to keep the logic as-is but add comments to each condition

RamIdeas (approved):

petebacondarwin (approved):

RamIdeas (approved):

PR #6627, opened 2024-09-04
Fix max asset count message

GregBrimble:

Fixes reporting back just the 20k limit again.

image

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: untested code
  - [ ] I don't know
  - [x] Not required because: untested
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/miniflare/src/plugins/assets/index.ts
+++ b/packages/miniflare/src/plugins/assets/index.ts
@@ -202,7 +202,7 @@ const walk = async (dir: string) => {
 			}
 		})
 	);
-	if (counter >= MAX_ASSET_COUNT) {
+	if (counter > MAX_ASSET_COUNT) {
 		throw new Error(
 			`Maximum number of assets exceeded.\n` +
 				`Cloudflare Workers supports up to ${MAX_ASSET_COUNT.toLocaleString()} assets in a version. We found ${counter.toLocaleString()} files in the specified assets directory "${dir}".\n` +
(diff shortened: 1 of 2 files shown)


CarmenPopoviciu (approved):

emily-shen:
I'm a bit confused - doesn't seem like any of the changed code is affecting the 20k limit constant?  Was the error not surfacing properly from inside the promise loop? D:

Does the same change also need to be done to this file? (yes I really need to move the duplicated code to a shared file somewhere 🥲 ) https://github.com/cloudflare/workers-sdk/blob/8d1d464f2b549dc7d7020fd45f025cd7c8671ce9/packages/wrangler/src/experimental-assets.ts#L240

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -210,6 +202,13 @@ const walk = async (dir: string) => {
 			}
 		})
 	);
+	if (counter >= MAX_ASSET_COUNT) {
```
petebacondarwin, line 205: Should this be 

```suggestion
	if (counter > MAX_ASSET_COUNT) {
```

petebacondarwin:
> > Fixes reporting back just the 20k limit again.
> 
> I'm a bit confused - doesn't seem like any of the changed code is affecting the 20k limit constant? Was the error not surfacing properly from inside the promise loop? D:
> 
> Does the same change also need to be done to this file? (yes I really need to move the duplicated code to a shared file somewhere 🥲 )
> 
> https://github.com/cloudflare/workers-sdk/blob/8d1d464f2b549dc7d7020fd45f025cd7c8671ce9/packages/wrangler/src/experimental-assets.ts#L240

I think the point is that if we throw the error as we find the 20,000th file then we will always say that we found 20,000 files.
Whereas if we wait until the loop is complete, it will report the actual number of files found.

PR #6628, opened 2024-09-04
Improve asset uploading messaging

GregBrimble:

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [x] Not required because: no e2e test coverage
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/experimental-assets.ts
```diff
@@ -149,18 +150,14 @@ export const syncExperimentalAssets = async (
...
 				return res;
 			} catch (e) {
 				if (attempts < MAX_UPLOAD_ATTEMPTS) {
-					logger.info(
-						chalk.dim(
-							`Bucket ${bucketIndex + 1}/${initializeAssetsResponse.buckets.length} upload failed. Retrying...\n`,
-							e
-						)
-					);
+					logger.info(chalk.dim(`Asset upload failed. Retrying...\n`, e));
```
CarmenPopoviciu, line 160: we sure prev copy is not more useful for users?

emily-shen: Eh I like these changes - I don't think buckets are really helping anyone meaningfully track any problem down. Plus they were a bit noisy because sometimes they were printing out of order, and I think sometimes the buckets can be empty.

No. of assets successfully uploaded seems more useful for users.

GregBrimble: Confirmed, we're going with the messaging I have here in this PR.

PR #6629, opened 2024-09-04
Add Cloudflare Pipelines commands

oliy:

Adds new Cloudflare Pipelines commands: PIPE-27, PIPE-96, PIPE-110

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/16639
  - [ ] Not necessary because: PM writing docs separately, but needs working wrangler to test.

### DIFF
--- a/.changeset/polite-tigers-end.md
+++ b/.changeset/polite-tigers-end.md
@@ -1,0 +1,20 @@
+---
+"wrangler": minor
+---
+
+feature: Integrate the Cloudflare Pipelines product into wrangler.
+
+Cloudflare Pipelines is a product that handles the ingest of event streams 
+into R2.  This feature integrates various forms of managing pipelines.
+
+Usage:
+  wrangler pipelines create <pipeline>  Create a new pipeline
+  wrangler pipelines list               List current pipelines
+  wrangler pipelines show <pipeline>    Show a pipeline configuration
+  wrangler pipelines update <pipeline>  Update a pipeline
+  wrangler pipelines delete <pipeline>  Delete a pipeline
+
+Examples:
+  wrangler pipelines create my-pipeline --r2 MY_BUCKET --access-key-id "my-key" --secret-access-key "my-secret"
+  wrangler pipelines show my-pipeline
+  wrangler pipelines delete my-pipline
--- a/packages/wrangler/src/__tests__/pipelines.test.ts
+++ b/packages/wrangler/src/__tests__/pipelines.test.ts
@@ -1,0 +1,571 @@
+import { http, HttpResponse } from "msw";
+import { __testSkipDelays } from "../pipelines";
+import { endEventLoop } from "./helpers/end-event-loop";
+import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
+import { mockConsoleMethods } from "./helpers/mock-console";
+import { msw } from "./helpers/msw";
+import { runInTempDir } from "./helpers/run-in-tmp";
+import { runWrangler } from "./helpers/run-wrangler";
+import type { Pipeline, PipelineEntry } from "../pipelines/client";
+
+describe("pipelines", () => {
+	const std = mockConsoleMethods();
+	mockAccountId();
+	mockApiToken();
+	runInTempDir();
+
+	const samplePipeline = {
+		currentVersion: 1,
+		id: "0001",
+		name: "my-pipeline",
+		metadata: {},
+		source: [
+			{
+				type: "binding",
+				format: "json",
+			},
+		],
+		transforms: [],
+		destination: {
+			type: "json",
+			batch: {},
+			compression: {
+				type: "none",
+			},
+			format: "json",
+			path: {
+				bucket: "bucket",
+			},
+		},
+		endpoint: "https://0001.pipelines.cloudflarestorage.com",
+	};
+
+	function mockCreateR2Token(bucket: string) {
+		const requests = { count: 0 };
+		msw.use(
+			http.get(
+				"*/accounts/:accountId/r2/buckets/:bucket",
+				async ({ params }) => {
+					expect(params.accountId).toEqual("some-account-id");
+					expect(params.bucket).toEqual(bucket);
+					requests.count++;
+					return HttpResponse.json(
+						{
+							success: true,
+							errors: [],
+							messages: [],
+							result: null,
+						},
+						{ status: 200 }
+					);
+				},
+				{ once: true }
+			),
+			http.get(
+				"*/user/tokens/permission_groups",
+				async () => {
+					requests.count++;
+					return HttpResponse.json(
+						{
+							success: true,
+							errors: [],
+							messages: [],
+							result: [
+								{
+									id: "2efd5506f9c8494dacb1fa10a3e7d5b6",
+									name: "Workers R2 Storage Bucket Item Write",
+									description:
+										"Grants write access to Cloudflare R2 Bucket Scoped Storage",
+									scopes: ["com.cloudflare.edge.r2.bucket"],
+								},
+							],
+						},
+						{ status: 200 }
+					);
+				},
+				{ once: true }
+			),
+			http.post(
+				"*/user/tokens",
+				async () => {
+					requests.count++;
+					return HttpResponse.json(
+						{
+							success: true,
+							errors: [],
+							messages: [],
+							result: {
+								id: "service-token-id",
+								name: "my-service-token",
+								value: "my-secret-value",
+							},
+						},
+						{ status: 200 }
+					);
+				},
+				{ once: true }
+			)
+		);
+		return requests;
+	}
+
+	function mockCreeatR2TokenFailure(bucket: string) {
+		const requests = { count: 0 };
+		msw.use(
+			http.get(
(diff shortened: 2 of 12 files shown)


andyjessop:
@oliy I'm seeing a fair amount of linting errors here. Can you run ESLint locally and push fixes?

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+
+				if (args.filepath) {
+					pipelineConfig.destination.path.filepath = args.filepath;
+				}
+				if (args.filename) {
+					pipelineConfig.destination.path.filename = args.filename;
+				}
+
+				logger.log(`🌀 Creating pipeline named "${name}"`);
+				const pipeline = await createPipeline(accountId, pipelineConfig);
```
andyjessop, line 243: What should happen if this fails?

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+		.command(
+			"list",
+			"List current pipelines",
+			(yargs) => yargs,
+			async (args) => {
+				const config = readConfig(args.config, args);
+				const accountId = await requireAuth(config);
+
+				// TODO: we should show bindings & transforms if they exist for given ids
+				const pipelines = await listPipelines(accountId);
```
andyjessop, line 266: As above, we should handle the situation where remote calls fail.

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+				}
+
+				logger.log(`🌀 Creating pipeline named "${name}"`);
+				const pipeline = await createPipeline(accountId, pipelineConfig);
+				await metrics.sendMetricsEvent("create pipeline", {
+					sendMetrics: config.send_metrics,
+				});
+
+				logger.log(
+					`✅ Successfully created pipeline "${pipeline.name}" with ID ${pipeline.id}`
```
andyjessop, line 249: Suggestion for making the output more consistent.

```suggestion
					`✅ Successfully created pipeline "${pipeline.name}" with id "${pipeline.id}"`
```

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+			},
+			async (args) => {
+				await printWranglerBanner();
+
+				const name = args.pipeline;
+				// only the fields set will be updated - other fields will use the existing config
+				const config = readConfig(args.config, args);
+				const accountId = await requireAuth(config);
+
+				const pipelineConfig = await getPipeline(accountId, name);
```
andyjessop, line 333: What should happen if this pipeline doesn't exist?

packages/wrangler/src/__tests__/pipelines.test.ts
```diff
@@ -0,0 +1,438 @@
...
+		await runWrangler("pipelines delete foo");
+
+		expect(std.err).toMatchInlineSnapshot(`""`);
+		expect(std.out).toMatchInlineSnapshot(`
+			"Deleting pipeline foo.
+			Deleted pipeline foo."
+		`);
+		expect(requests.count).toEqual(1);
+	});
+});
```
andyjessop, line 438: We could add unhappy path tests here too. Things like:
- try to update a pipeline that doesn't exist
- delete a pipeline that doesn't exist
- user has invalid credentials
- bucket doesn't exist
- any other reason why create/update/delete operations might fail

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+					demandOption: true,
+				});
+			},
+			async (args) => {
+				await printWranglerBanner();
+				const config = readConfig(args.config, args);
+				const accountId = await requireAuth(config);
+				const name = args.pipeline;
+
+				if (!name.match(/^[a-zA-Z0-9-]+$/)) {
```
andyjessop, line 447: As this logic is repeated, it could be extracted to a shared function. This is good practice in general for Regex anyway as it's so notoriously unreadable (although this example is simple).

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+							entrypoint: split[1],
+						});
+					} else if (split.length === 1) {
+						pipelineConfig.transforms.push({
+							script: split[0],
+							entrypoint: "Transform",
+						});
+					} else {
+						throw new Error(
+							"invalid transform: required syntax script.entrypoint"
```
andyjessop, line 230:
```suggestion
							"Invalid transform: required syntax script.entrypoint"
```

andyjessop (changes requested):

.changeset/polite-tigers-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Add Pipelines commands to Wrangler
```
penalosa, line 5: Could you modify this to follow the changeset format? Additionally, since it's a big feature a couple examples of how to use the commands would be nice!

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+	const serviceToken = await generateR2ServiceToken(
+		`Service token for Pipeline ${name}`,
+		accountId,
+		bucket
+	);
+	const access_key_id = serviceToken.id;
+	const secret_access_key = sha256(serviceToken.value);
+
+	// wait for token to settle/propagate
+	await sleep(3000);
```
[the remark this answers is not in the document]
oliy: I agree this isn't ideal, but the alternative seemed rougher.  The confirmation would require the use of the S3 API, which would be either:
1. much more complication, since we'd have to generate a signature ourselves: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html
2. include a large package that implements the s3 standard, just to get endpoint verification.

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+				if (args.authentication !== undefined && args.authentication === true) {
+					pipelineConfig.source = [
+						{
+							type: "binding",
+							format: "json",
+						},
+					];
+				}
+
+				if (args.transform !== undefined) {
```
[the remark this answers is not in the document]
oliy: I added comment and simplified.

packages/wrangler/src/pipelines/index.ts
oliy: It throws an exception with the appropriate error message.  Is there a better mechanism for errors?  This seemed to be the common pattern among other handlers.

packages/wrangler/src/pipelines/index.ts
oliy: Exception gets thrown that shows the error message that the pipeline doesn't exist:

```
Pipeline does not exist [code: 1000]
```

.github/workflows/prerelease-pipelines.yml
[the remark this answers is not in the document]
oliy: can remove.  this was added as part of copying the pattern from a different change.

packages/wrangler/src/pipelines/client.ts
```diff
@@ -0,0 +1,192 @@
...
+
+	return await fetchResult<ServiceToken>(`/user/tokens`, {
+		method: "POST",
+		headers: API_HEADERS,
+		body,
+	});
+}
+
+// Get R2 bucket information from v4 API
+export async function getR2Bucket(
```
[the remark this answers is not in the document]
oliy: Surprisingly, no.

packages/wrangler/src/pipelines/index.ts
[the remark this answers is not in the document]
oliy: To address the delay in tests, I've added a delay bypass for tests.

packages/wrangler/src/pipelines/index.ts
```diff
@@ -0,0 +1,459 @@
...
+						},
+					},
+				};
+				const destination = pipelineConfig.destination;
+				if (
+					!destination.credentials.access_key_id &&
+					!destination.credentials.secret_access_key
+				) {
+					// auto-generate a service token
+					await authorizeR2Bucket(name, accountId, pipelineConfig.destination);
```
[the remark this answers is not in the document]
oliy: changed.

packages/wrangler/src/pipelines/index.ts
oliy: should only fail if the account was invalid, which will give an API error that should have the corresponding error message.

andyjessop:
Closing because this was superseded by this PR: https://github.com/cloudflare/workers-sdk/pull/6649

PR #6641, opened 2024-09-05
Add [[pipelines]] binding in wrangler.

oliy:

PIPE-116 Implement regular [[pipelines]] bindings

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [x] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/16639
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/configuration.test.ts
@@ -130,6 +130,7 @@ describe("normalizeAndValidateConfig()", () => {
 			upload_source_maps: undefined,
 			placement: undefined,
 			tail_consumers: undefined,
+			pipelines: [],
 		});
 		expect(diagnostics.hasErrors()).toBe(false);
 		expect(diagnostics.hasWarnings()).toBe(false);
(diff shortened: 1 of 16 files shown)


packages/wrangler/src/__tests__/configuration.test.ts
```diff
@@ -10,6 +10,8 @@ import type {
 	RawDevConfig,
 	RawEnvironment,
 } from "../config";
+import { pipeline } from "node:stream";
```
andyjessop, line 13: These imports are not used. I think you will need to run ESLint on this branch.

packages/wrangler/src/__tests__/configuration.test.ts
```diff
@@ -3181,6 +3184,104 @@ describe("normalizeAndValidateConfig()", () => {
...
+				expect(diagnostics.hasWarnings()).toBe(false);
+				expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Processing wrangler configuration:
+			  - The field \\"pipelines\\" should be an array but got 999."
+		`);
+			});
+
+			it("should error if pipelines is null", () => {
+				const { diagnostics } = normalizeAndValidateConfig(
+					{ pipelines: null } as unknown as RawConfig,
```
andyjessop, line 3232: Nit: one thing you could do here is to ignore the error instead of asserting it:

```
const { diagnostics } = normalizeAndValidateConfig(
  // @ts-expect-error purposely passing an invalid config
  { pipelines: null },
  undefined,
	{ env: undefined }
);
```

It's very minor, and you can do it either way, but ignoring the error here is being more explicit about what's going on.

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -109,8 +109,9 @@ export type WorkerMetadataBinding =
 				params?: { name: string }[];
 			};
 	  }
-	| { type: "mtls_certificate"; name: string; certificate_id: string }
-	| {
+	  | { type: "mtls_certificate"; name: string; certificate_id: string }
+	  | { type: "pipelines"; name: string; id: string }
```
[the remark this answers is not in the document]
oliy: unfortunately, this is the schema that already exists in EWC and runtime capnprotos.  There would be a lot more PRs to complete to change this schema.

packages/wrangler/src/config/validation.ts
```diff
@@ -3103,6 +3114,44 @@ const validateConsumer: ValidatorFn = (diagnostics, field, value, _config) => {
...
+				value
+			)}`
+		);
+		return false;
+	}
+	let isValid = true;
+	// Pipelin bindings must have a binding and a pipeline.
+	if (!isRequiredProperty(value, "binding", "string")) {
+		diagnostics.errors.push(
+			`"${field}" bindings should have a string "binding" field but got ${JSON.stringify(
```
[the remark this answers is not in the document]
oliy: This pattern is used through this file.  I think the end result is that string interpolation of undefined it will show a message:
```
... binding should have a string "binding" field but got undefined
```
Which seems still like a reasonable outcome.

packages/wrangler/src/__tests__/configuration.test.ts
oliy: Resolved all check:lint errors.

andyjessop:
Closing this as the changes have been added here: https://github.com/cloudflare/workers-sdk/pull/6674

PR #6631, opened 2024-09-04
feat: asset serving config for workers + assets

emily-shen:
## What this PR solves / how to test
Please review by commit <3
[add AW tests to new fixture 'asset-config' (wip)](https://github.com/cloudflare/workers-sdk/pull/6631/commits/bb8c12e0a646bf64dae60db04ca12756ec973363) - this one is a wip so don't 'review' but imo is very useful for understanding behaviour of [add asset config behaviour to AW (this is the big one)](https://github.com/cloudflare/workers-sdk/pull/6631/commits/f9c71359e848791d613ff703949130ec4afaccb5)

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required / Maybe required
  - [ ] Not required because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: not public facing yet

PR #6670, opened 2024-09-10
Rename asset-worker RPC methods as unstable

GregBrimble:

Just a tiny thing. Do not merge before getting in #6631.

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: untested
  - [ ] I don't know
  - [x] Not required because: untested
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/popular-snakes-report.md
+++ b/.changeset/popular-snakes-report.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/workers-shared": patch
+---
+
+chore: Rename asset-worker RPC methods as unstable\_\*
+
+The Asset Worker is currently set up as a `WorkerEntrypoint` class so that it is able to accept RPC calls to any of its public methods. There are currently four such public methods defined on this Worker: `canFetch`, `getByETag`, `getByPathname` and `exists`. While we are stabilising the implementation details of these methods, we would like to prevent developers from having their Workers call these methods directly. To that end, we are adopting the `unstable_<method_name>` naming convention for all of the aforementioned methods, to indicate that they are still in flux and that they are not an established API contract.
--- a/packages/workers-shared/router-worker/wrangler.toml
+++ b/packages/workers-shared/router-worker/wrangler.toml
@@ -8,5 +8,20 @@
 # (see packages/wrangler/src/dev/miniflare.ts -> buildMiniflareOptions())
 ##
 name = "router-worker"
+account_id = "0f1b8aa119a907021f659042f95ea9ba"
+workers_dev = false
 main = "src/index.ts"
 compatibility_date = "2024-07-31"
+
+[[unsafe.bindings]]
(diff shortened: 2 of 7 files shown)


petebacondarwin (commented):
I think there might be some references in tests too?

.changeset/popular-snakes-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/workers-shared": patch
+---
+
+chore: Rename asset-worker RPC methods as unstable\_\*
```
CarmenPopoviciu, line 5: what is the purpose we are doing this for? what does `unstable` mean in this context? does it express that these APIs are subject to change?

packages/workers-shared/router-worker/wrangler.toml
```diff
@@ -8,5 +8,20 @@
...
+type = "internal_assets"
+fetcherApi = "fetcher"
+
+[[unsafe.bindings]]
+name = "USER_WORKER"
+type = "origin"
+
+[unsafe.metadata.build_options]
+stable_id = "cloudflare/cf_router_worker"
+networks = ["cf","jdc"]
```
petebacondarwin, line 27: How come these were not already added in the assets PR?

.changeset/popular-snakes-report.md
petebacondarwin: Because we have set up the Asset Worker as an EntryPoint class it is able to accept RPC calls to any of its public methods. Greg doesn't want developers in their User Worker to call these methods directly and become part of our public API contract, so he is naming them with the `unstable_` prefix to mark them as not part of that contradt.

I agree with the sentiment although the correct approach is for these not to be on the public API in the first place, which we could achieve by having a separate named entry-point to this Worker that only the Router Worker gets access to.

In the meantime this is a hack to avoid the public API contract, which I am happy to go with, but with the intention of a fast follow up later this autumn.

petebacondarwin (approved):

.changeset/popular-snakes-report.md
CarmenPopoviciu: yeah, I figured as much. Do we need to explicitly call that out maybe? As a comment or co?

CarmenPopoviciu (approved):

.changeset/popular-snakes-report.md
CarmenPopoviciu: in fact, lemme do that real quick so we can merge this in

CarmenPopoviciu: ok, done ✅

CarmenPopoviciu (approved):

PR #6680, opened 2024-09-11
Fix Pages duplicating hash in redirects

WalshyDev:

Fixes CUSTESC-44240

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: no new changes, just fixing behaviour

### DIFF
--- a/packages/pages-shared/__tests__/asset-server/handler.test.ts
+++ b/packages/pages-shared/__tests__/asset-server/handler.test.ts
@@ -390,40 +390,6 @@ describe("asset-server handler", () => {
 		});
 	}
 
-	// test("Returns a redirect without duplicating the hash component", async () => {
-	// 	const { response, spies } = await getTestResponse({
-	// 		request: "https://foo.com/bar",
-	// 		metadata: createMetadataObjectWithRedirects([
-	// 			{ from: "/bar", to: "https://foobar.com/##heading-7", status: 301 },
-	// 		]),
-	// 	});
-
-	// 	expect(spies.fetchAsset).toBe(0);
-	// 	expect(spies.findAssetEntryForPath).toBe(0);
-	// 	expect(spies.getAssetKey).toBe(0);
-	// 	expect(spies.negotiateContent).toBe(0);
-	// 	expect(response.status).toBe(301);
-	// 	expect(response.headers.get("Location")).toBe(
-	// 		"https://foobar.com/##heading-7"
-	// 	);
-	// });
-
-	test("it should redirect uri-encoded paths", async () => {
-		const { response, spies } = await getTestResponse({
-			request: "https://foo.com/some%20page",
-			metadata: createMetadataObjectWithRedirects([
-				{ from: "/some%20page", to: "/home", status: 301 },
-			]),
-		});
-
-		expect(spies.fetchAsset).toBe(0);
-		expect(spies.findAssetEntryForPath).toBe(0);
-		expect(spies.getAssetKey).toBe(0);
-		expect(spies.negotiateContent).toBe(0);
-		expect(response.status).toBe(301);
-		expect(response.headers.get("Location")).toBe("/home");
-	});
-
 	// 	test("getResponseFromMatch - same origin paths specified as root-relative", () => {
 	// 		const res = getResponseFromMatch(
 	// 			{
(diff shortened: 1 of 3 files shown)


packages/pages-shared/__tests__/asset-server/handler.test.ts
```diff
@@ -920,6 +886,102 @@ describe("asset-server handler", () => {
...
+			const { response } = await getTestResponse({
+				request: "https://foo.com/bar",
+				metadata: createMetadataObjectWithRedirects([
+					{ from: "/bar", to: "/?test=abc#def", status: 301 },
+				]),
+			});
+
+			expect(response.status).toBe(301);
+			expect(response.headers.get("Location")).toBe("/?test=abc#def");
+		});
```
GregBrimble, line 969:
```suggestion
		test("redirects to a query string and hash same-origin", async () => {
			const { response } = await getTestResponse({
				request: "https://foo.com/bar",
				metadata: createMetadataObjectWithRedirects([
					{ from: "/bar", to: "/?test=abc#def", status: 301 },
				]),
			});

			expect(response.status).toBe(301);
			expect(response.headers.get("Location")).toBe("/?test=abc#def");
		});
		
		test("redirects to a incoming query string and hash same-origin", async () => {
			const { response } = await getTestResponse({
				request: "https://foo.com/bar?test=abc",
				metadata: createMetadataObjectWithRedirects([
					{ from: "/bar", to: "/#def", status: 301 },
				]),
			});

			expect(response.status).toBe(301);
			expect(response.headers.get("Location")).toBe("/?test=abc#def");
		});
```

I think that's going to fail with `/#def?test=abc`.

GregBrimble: But also, that's not the result of your change here. That's just existing behavior, I think...

GregBrimble (approved):
So tentatively approved, but with a fairly serious dollop of caution.

packages/pages-shared/__tests__/asset-server/handler.test.ts
WalshyDev: Yeah I looked at this and per rfc, query string must be first - https://www.rfc-editor.org/rfc/rfc3986#section-4.1
Everything after hash is treated as a hash so I think this is actually ok

```
test("redirects to a query string and hash against rfc", async () => {
			const { response } = await getTestResponse({
				request: "https://foo.com/bar",
				metadata: createMetadataObjectWithRedirects([
					{ from: "/bar", to: "https://foobar.com/#def?test=abc", status: 301 },
				]),
			});

			expect(response.status).toBe(301);
			expect(response.headers.get("Location")).toBe(
				"https://foobar.com/#def?test=abc"
			);
		});

Comment (GregBrimble) [reply packages/pages-shared/__tests__/asset-server/handler.test.ts:969]: That test you have written there is totally valid behavior. It's not against the RFC because, like you say, it's being (correctly) treated as part of the hash.

Mine was exposing a separate bug where we're incorrectly appending `.search` on the end of `.href`. In the case of `.href` containing anchor (`.hash`) parts and there are incoming `.search` parts, then the link we construct will be wrong. The anchor won't match up to anything on that redirected page and the `.search` parts will be dropped.

But yeah, separate bug which will be unaffected by this PR, best I can tell.

Comment (WalshyDev) [reply packages/pages-shared/__tests__/asset-server/handler.test.ts:969]: Oh yeah, that's been the behaviour for a long time... I was trying to test this but I got 404, it never actually hit the redirect and I stopped looking at that point

Should probably look into it at some point though

File: packages/pages-shared/asset-server/handler.ts [hunk 2]
```diff
@@ -227,9 +227,8 @@ export async function generateHandler<
 						? `${destination.pathname}${destination.search || search}${
 								destination.hash
 							}`
-						: `${destination.href}${destination.search ? "" : search}${
-								destination.hash
-							}`;
+						: `${destination.href}${destination.search ? "" : search}`;
```
GregBrimble, line 230: How about

```suggestion
						: `${destination.origin}${destination.search ? "" : search}${destination.hash}`;
```

WalshyDev: Yeah that's good, ty!

PR #6693, opened 2024-09-12
Fix wrangler version secrets with Assets

GregBrimble:

Persists Workers Assets across `wrangler versions secrets put/bulk`.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/versions/secrets/index.ts
+++ b/packages/wrangler/src/versions/secrets/index.ts
@@ -209,6 +209,7 @@ export async function copyWorkerVersionWithNewSecrets({
 			"workers/message": versionMessage,
 			"workers/tag": versionTag,
 		},
+		keep_assets: true,
 		experimental_assets: undefined,
 		observability: scriptSettings.observability,
 	};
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/versions/secrets/index.ts
```diff
@@ -209,6 +209,7 @@ export async function copyWorkerVersionWithNewSecrets({
 			"workers/message": versionMessage,
 			"workers/tag": versionTag,
 		},
+		keep_assets: true,
```
penalosa, line 212: Oh I see! I really don't like this, there should be a versions secrets API endpoint that Wrangler can hit

penalosa (approved):

packages/wrangler/src/versions/secrets/index.ts
GregBrimble: Totally agree.

PR #6696, opened 2024-09-12
Support `WRANGLER_CI_MATCH_TAG` environment variable

penalosa:

Fixes [DEVX-1399](https://jira.cfdata.org/browse/DEVX-1399)

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
  - [ ] I don't know
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: Internal detail for CI use


packages/wrangler/src/deploy/index.ts
```diff
@@ -326,6 +328,15 @@ export async function deployHandler(
 
 	const beforeUpload = Date.now();
 	const name = getScriptName(args, config);
+	assert(
```
CarmenPopoviciu, line 331: what is the diff between using `assert` over `logger.error()` here?

packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -89,3 +89,12 @@ export const getOutputFileDirectoryFromEnv = getEnvironmentVariableFactory({
 export const getOutputFilePathFromEnv = getEnvironmentVariableFactory({
 	variableName: "WRANGLER_OUTPUT_FILE_PATH",
 });
+
+/**
+ * `WRANGLER_CI_MATCH_TAG` specifies a worker tag
+ *
+ * If this is set, Wrangler will ensure the worker being targeted has this tag
```
CarmenPopoviciu, line 96:
```suggestion
 * If this is set, Wrangler will ensure the Worker being targeted has this tag
```

packages/wrangler/src/match-tag.ts
```diff
@@ -0,0 +1,46 @@
...
+		const worker = await fetchResult<ServiceMetadataRes>(
+			`/accounts/${accountId}/workers/services/${workerName}`
+		);
+		tag = worker.default_environment.script.tag;
+	} catch (e) {
+		logger.debug(e);
+		// code: 10090, message: workers.api.error.service_not_found
+		if ((e as { code?: number }).code === 10090) {
+			throw new FatalError(
+				`Your Worker's name (${workerName}) does not match what is expected by the CI system`
```
CarmenPopoviciu, line 30: is there an action we can recommend to users here?

packages/wrangler/src/match-tag.ts
```diff
@@ -0,0 +1,46 @@
...
+			throw new FatalError(
+				"Wrangler cannot validate that your Worker name matches what is expected by the CI system"
+			);
+		}
+	}
+	if (tag !== matchTag) {
+		logger.debug(
+			`Failed to match Worker tag. The API returned "${tag}", but the CI system expected "${matchTag}"`
+		);
+		throw new FatalError(
```
CarmenPopoviciu, line 42: is it worth merging these two error messages into one, and surfacing everything in the `FatalError`?

CarmenPopoviciu: IMHO I shouldn't be in debug mode to get that kind of error details

packages/wrangler/src/versions/index.ts
```diff
@@ -244,7 +246,19 @@ export async function versionsUploadHandler(
 	const accountId = args.dryRun ? undefined : await requireAuth(config);
 	const name = getScriptName(args, config);
 
-	await standardPricingWarning(config);
+	assert(
```
CarmenPopoviciu, line 249: same `assert` vs `logger.error` question here

packages/wrangler/src/versions/index.ts
```diff
@@ -244,7 +246,19 @@ export async function versionsUploadHandler(
...
+		name,
+		'You need to provide a name when publishing a worker. Either pass it as a cli arg with `--name <name>` or in your config file as `name = "<name>"`'
+	);
+
+	if (!args.dryRun) {
```
CarmenPopoviciu, line 254: it look like the two `if` blocks can be merged into one?

packages/wrangler/src/versions/index.ts
```diff
@@ -244,7 +246,19 @@ export async function versionsUploadHandler(
 	const accountId = args.dryRun ? undefined : await requireAuth(config);
 	const name = getScriptName(args, config);
 
-	await standardPricingWarning(config);
+	assert(
+		name,
+		'You need to provide a name when publishing a worker. Either pass it as a cli arg with `--name <name>` or in your config file as `name = "<name>"`'
```
CarmenPopoviciu, line 251:
```suggestion
		'You need to provide a name when publishing a Worker. Either pass it as a cli arg with `--name <name>` or in your config file as `name = "<name>"`'
```

CarmenPopoviciu (approved):

packages/wrangler/src/match-tag.ts
jahands: We do have some error messaging written up [here](https://jira.cfdata.org/browse/BANDA-411). cc @aninibread

PR #6697, opened 2024-09-13
Allow CI deploying non-private packages

GregBrimble:

workers-shared is the only non-private and deployable package. And it was skipped in the last release because of this check.

- Tests
  - [ ] TODO (before merge)
  - [x] Included
  - [ ] Not necessary because: 
  - [ ] I don't know
  - [ ] TODO (before merge)
  - [ ] Included
  - [x] Not necessary because: internal tooling change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Not necessary because: internal tooling change

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
 	"author": "wrangler@cloudflare.com",
 	"scripts": {
 		"build": "dotenv -- turbo build",
-		"check": "pnpm check:fixtures && node lint-turbo.mjs && dotenv -- turbo check:lint check:type check:format type:tests",
+		"check": "pnpm check:fixtures && pnpm check:deployments && node lint-turbo.mjs && dotenv -- turbo check:lint check:type check:format type:tests",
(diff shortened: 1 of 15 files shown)


petebacondarwin (changes requested):
I'm not keen on this change. If someone adds a deploy script to an npm package changeset will attempt to deploy it twice.

What if we change the name of the script that triggers the non-npm deployment to something like `deployWorkers`?

GregBrimble:
Happy to go with whatever devprod want here. If they want `deployWorkers`, that's easily changed :)

petebacondarwin:
Actually, given we already have a workers-sdk section in packages (in order to trigger prerelease previews) (e.g. https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/package.json#L106). How about we just add some config there they says when a package should be also deployed outside of the npm deployments?

GregBrimble:
Yup, I like that. Can do!

petebacondarwin (approved):

PR #6705, opened 2024-09-13
fix: upload compat date for static-asset only projects

emily-shen:

Fixes [WC-2722](https://jira.cfdata.org/browse/WC-2722)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
+++ b/packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
@@ -553,12 +553,13 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
 		not_found_handling: experimental_assets?.assetConfig?.not_found_handling,
 	};
 
-	const staticAssetOnly = experimental_assets?.routingConfig?.has_user_worker === false;
-	let mainModuleName = {}
-	if(main.type !== "commonjs" && !staticAssetOnly) {
-		mainModuleName = { main_module: main.name }
+	const staticAssetOnly =
+		experimental_assets?.routingConfig?.has_user_worker === false;
+	let mainModuleName = {};
+	if (main.type !== "commonjs" && !staticAssetOnly) {
+		mainModuleName = { main_module: main.name };
 	} else if (!staticAssetOnly) {
-		mainModuleName = { body_part: main.name }
+		mainModuleName = { body_part: main.name };
 	}
 
 	const metadata: WorkerMetadata = {
@@ -600,7 +601,7 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
 		);
 	}
 
-	if (!staticAssetOnly){
+	if (!staticAssetOnly) {
 		for (const module of [main].concat(modules || [])) {
 			formData.set(
 				module.name,
--- a/packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
+++ b/packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
@@ -202,6 +202,7 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
 					config: assetConfig,
 				},
 				...(compatibility_date && { compatibility_date }),
+				...(compatibility_flags && { compatibility_flags }),
 			})
 		);
 		return formData;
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -569,10 +548,21 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
...
+	} else if (!staticAssetOnly) {
+		mainModuleName = { body_part: main.name }
+	}
+
 	const metadata: WorkerMetadata = {
-		...(main.type !== "commonjs"
-			? { main_module: main.name }
-			: { body_part: main.name }),
+		...mainModuleName,
 		bindings: metadataBindings,
```
emily-shen, line 566: Someone could add bindings etc. to a static-asset only project and this probably would not error. It also wouldn't have any affect because EWC won't save anything besides compat date/flags.
Is this something we want to warn users about?

emily-shen: already being tracked here for after BW https://jira.cfdata.org/browse/WC-2562

petebacondarwin (commented):
I couldn't find a test that showed that compat_date gets included in the upload...?

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -569,10 +548,22 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
...
 
+	const assetConfig = {
+		html_handling: experimental_assets?.assetConfig?.html_handling,
+		not_found_handling: experimental_assets?.assetConfig?.not_found_handling,
+	};
+
+	const staticAssetOnly =
+		experimental_assets?.routingConfig?.has_user_worker === false;
+	let mainModuleName = {};
+	if (main.type !== "commonjs" && !staticAssetOnly) {
```
CarmenPopoviciu, line 559: nit
```suggestion
	if (!staticAssetOnly) {
	  mainModuleName = main.type !== "commonjs"
			? { main_module: main.name }
			: { body_part: main.name }
	}
```

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -184,27 +184,6 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
 		observability,
 	} = worker;
 
-	const assetConfig = {
-		html_handling: experimental_assets?.assetConfig?.html_handling,
-		not_found_handling: experimental_assets?.assetConfig?.not_found_handling,
-	};
-
-	// short circuit if static assets upload only
```
CarmenPopoviciu, line 192: I might be missing smth, so just for me to understand...why can't we keep this short-circuit and attach the `compatibility_date` to the metadata object? Aka

```
formData.set(
			"metadata",
			JSON.stringify({
                                 ...(compatibility_date && { compatibility_date }),
				assets: {
					jwt: experimental_assets.jwt,
					config: assetConfig,
				},
			})
		);
```

emily-shen: It would work with the short circuit still, but IMO its actually cleaner to keep everything in one path - open to being persuaded otherwise though!

CarmenPopoviciu: but not short-circuit-ing results in having to add all those `if(!assetsOnly)` clauses, which add to the cognitive load IMO. I mean, it's not a whole lotta of extra load, but why not exit early if we can?

CarmenPopoviciu: basically this PR would become a one line change + tests no?

petebacondarwin: Short circuit here feels simpler?
Otherwise there are more extra if statements below...

emily-shen: yep sounds good i'll revert these then :)

CarmenPopoviciu (approved):

GregBrimble (changes requested):
Can we add flags too please? Sorry, should have made that clear in the ticket. Compat date is always accompanied with flags (they work as a pair in the runtime where flags can override the defaults of the date).

GregBrimble (approved):
perfect, thank you!

PR #6719, opened 2024-09-14
fix(queues): Respect delivery delays during local dev

sdnts:

Fixes [MQ-744](https://jira.cfdata.org/browse/MQ-744)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: already tested in unit tests
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this is a bugfix

@jbwcloudflare


sdnts:
I also realized that there was a test for this bug, but the test was written incorrectly, so it was passing.

I moved delay tests to their own file and rewrote some of them so they were accurate and tested what they claimed they were testing.

packages/miniflare/src/workers/queues/broker.worker.ts
```diff
@@ -212,7 +212,9 @@ export class QueueBrokerObject extends MiniflareDurableObject<QueueBrokerObjectE
 	}
 
 	get #maybeProducer() {
-		return this.#producers[this.name];
+		return Object.values(this.#producers).find(
+			(p) => p?.queueName === this.name
+		);
```
sdnts, line 217: This is the fix. Producers are keyed by the binding name, not the worker name.

jbwcloudflare:
LGTM! thanks for the fix

penalosa (approved):

PR #6728, opened 2024-09-16
fix: try to serve assets from unencoded and encoded paths

emily-shen:

Fixes [WC-2693](https://jira.cfdata.org/browse/WC-2693)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no relevant e2es yet :(
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


fixtures/asset-config/test-cases/encoding-test-cases.ts
```diff
@@ -0,0 +1,335 @@
...
+				matchedFile: "/[boop].html",
+				finalPath: "/[boop]",
+			},
+			{
+				title: "/%5Bboop%5D -> 200 (with /%5Bboop%5D.html)",
+				files: ["/%5Bboop%5D.html", "/[boop].html"],
+				requestPath: "/%5Bboop%5D",
+				matchedFile: "/%5Bboop%5D.html",
+				finalPath: "/%5Bboop%5D",
+			},
```
emily-shen, line 71: Square brackets can be not encoded and so both versions could be uploaded and accessed. However, a user could upload two versions of something with characters that must be encoded, would not be able to access the unencoded file name.
e.g. a user tries to upload /beep/bòop.html and /beep/b%C3%B2op.html. Only /beep/b%C3%B2op.html can ever be accessed

GregBrimble: Similarly, `/beep/bòop.html` would be available at `/beep/b%C3%B2op` and `/beep/b%C3%B2op.html` would be available at `/beep/b%25C3%25B2op`. Totally daft, but they're being weird by uploading us encoded paths, so they asked for it.

GregBrimble (approved):

fixtures/asset-config/test-cases/encoding-test-cases.ts
```diff
@@ -0,0 +1,335 @@
...
+				title: "/beep/[b%C3%B2op] -> /beep/%5Bb%C3%B2op%5D 307",
+				files: ["/beep/[bòop].html"],
+				requestPath: "/beep/[b%C3%B2op]",
+				matchedFile: "/beep/[bòop].html",
+				finalPath: "/beep/%5Bb%C3%B2op%5D",
+			},
+			// exact matches should win
+			{
+				title: "/[boop] -> 200 (with /[boop].html)",
+				files: ["/%5Bboop%5D.html", "/[boop].html"],
```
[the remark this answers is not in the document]
emily-shen: > If you have time, maybe cook up another version which does "/%5Bboop%5D.html at /%255Bboop%255D only" and we can evaluate if we want to make that change separately? I think it might allow you to simplify the asset serving logic a fair bit — it makes us more consistent and deterministic. 

I actually tried this out the other day, and indeed to it does make the logic a lot simpler. I've chucked that on a branch here: https://github.com/cloudflare/workers-sdk/tree/double-encode

packages/workers-shared/utils/helpers.ts
```diff
@@ -1,22 +1,11 @@
 import { getType } from "mime";
 
-/** normalises sep for windows, and encodes each segment */
-export const encodeFilePath = (filePath: string, sep: string) => {
-	const encodedPath = filePath
-		.split(sep)
-		.map((segment) => encodeURIComponent(segment))
-		.join("/");
+/** normalises sep for windows */
+export const normalizeFilePath = (filePath: string, sep: string) => {
```
CarmenPopoviciu, line 4: if this fn exists solely for the purpose of path normalization in Windows, would it make sense to then call it `normalizeFilePathInWin`?

petebacondarwin: I'd push back on that name change. No need to leak the fact that this is needed for windows to the caller of the code.

Also there is the addition of a `/` prefix to all paths whether on Windows or not.

packages/workers-shared/utils/helpers.ts
```diff
@@ -1,22 +1,11 @@
...
+	const encodedPath = filePath.split(sep).join("/");
```
petebacondarwin, line 5: Do we need to concern ourselves over Windows drives? E.g. `C:\foo`?
Or is the filePath guaranteed to be relative?

The fact that we always prefix with `/` suggests that is the case?
Would it be worth asserting that is true in case some one starts calling this function with an absolute path in the future?

packages/workers-shared/asset-worker/src/handler.ts
```diff
@@ -18,7 +18,12 @@ export const handleRequest = async (
 ) => {
 	const { pathname, search } = new URL(request.url);
 
-	const intent = await getIntent(pathname, configuration, exists);
+	const decoded = pathname
```
CarmenPopoviciu, line 21:
```suggestion
	const decodedPathname = pathname
```

packages/workers-shared/asset-worker/src/handler.ts
```diff
@@ -18,7 +18,12 @@ export const handleRequest = async (
...
+		.split("/")
+		.map((x) => decodeURIComponent(x))
+		.join("/");
+	const intent = await getIntent(decoded, configuration, exists);
```
CarmenPopoviciu, line 25:
```suggestion
	const intent = await getIntent(decodedPathname, configuration, exists);
```

petebacondarwin (approved):
Do we need to be careful to release these changes to Wrangler and workers-shared/asset-worker at the same time since we are changing the contract between them for these corner cases?

CarmenPopoviciu (approved):
pending on Pete's comments and changeset

packages/workers-shared/utils/helpers.ts
emily-shen: It's only ever used on relative paths currently, but ive added a check in case :)

fixtures/asset-config/test-cases/encoding-test-cases.ts
```diff
@@ -0,0 +1,631 @@
...
+			},
+			{
+				title: "/%5Bboop%5D -> /%5Bboop%5D/ 307 (with /[boop]/index.html)",
+				files: ["/[boop].html", "/[boop]/index.html"],
+				requestPath: "/%5Bboop%5D",
+				matchedFile: "/[boop]/index.html",
+				finalPath: "/%5Bboop%5D/",
+			},
+			{
+				title: "/[boop].html -> /%5Bboop%5D.html 307 (with /[boop].html)",
```
GregBrimble, line 413: Good test, asserting the execution order!

fixtures/asset-config/test-cases/encoding-test-cases.ts
```diff
@@ -0,0 +1,631 @@
...
+				title: "/[boop] -> 404",
+				files: ["/[boop].html"],
+				requestPath: "/[boop]",
+			},
+			{
+				title: "/%5Bboop%5D -> 404",
+				files: ["/[boop].html"],
+				requestPath: "/%5Bboop%5D",
+			},
+			// encoding still operates when html_handling is set to 'none'
```
GregBrimble, line 551: yup, excellent

.changeset/shaggy-peas-notice.md
```diff
@@ -0,0 +1,11 @@
+---
+"@cloudflare/workers-shared": patch
+"miniflare": patch
+"wrangler": patch
+---
+
+fix: remove filepath encoding on asset upload and handle sometimes-encoded characters
+
+Some characters like [ ] @ are encoded by encodeURIComponent() but are often requested at an unencoded filepath.
```
petebacondarwin, line 9:
```suggestion
Some characters like [ ] @ are encoded by encodeURIComponent() but are often requested at an unencoded URL path.
```

.changeset/shaggy-peas-notice.md
```diff
@@ -0,0 +1,11 @@
...
+This change will make assets with filenames with these characters accessible at both the encoded and unencoded paths,
+but use the encoded path as the canonical one.
```
petebacondarwin, line 11:
```suggestion
but to use the encoded path as the canonical one, and to redirect requests to the canonical path if necessary.
```

packages/workers-shared/utils/helpers.ts
```diff
@@ -1,22 +1,15 @@
...
 
-/** normalises sep for windows, and encodes each segment */
-export const encodeFilePath = (filePath: string, sep: string) => {
-	const encodedPath = filePath
-		.split(sep)
-		.map((segment) => encodeURIComponent(segment))
-		.join("/");
+/** normalises sep for windows and prefix with `/` */
+export const normalizeFilePath = (relativeFilepath: string) => {
+	if (!isAbsolute(relativeFilepath)) {
```
GregBrimble, line 6: "if not isAbsolute, thow expected relative path error"

i.e.

"if is relative, throw expected relative path error"

is that backwards?

fixtures/asset-config/test-cases/html-handling-test-cases.ts
```diff
@@ -0,0 +1,788 @@
...
+				finalPath: "/bin/",
+			},
+			{
+				title: "/bin.html -> /bin/ 307 (with /bin/index.html)",
+				files: ["/bin%2F", "/bin/index.html"],
+				requestPath: "/bin.html",
+				matchedFile: "/bin/index.html",
+				finalPath: "/bin/",
+			},
+			// This is a bit rogue, but then so is the test case
```
GregBrimble, line 155: 👍

petebacondarwin (approved):

GregBrimble (approved):
Just that one question about `normalizeFilePath`!

packages/workers-shared/utils/helpers.ts
petebacondarwin: 😱  - good spot!

emily-shen: yes!! very backwards! currently fixing and have uncovered more issues when there's a user worker 😱

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -239,10 +239,9 @@ const walk = async (dir: string) => {
 				 */
 
 				const [pathHash, contentHash] = await Promise.all([
-					hashPath(encodeFilePath(relativeFilepath, path.sep)),
-					hashPath(
-						encodeFilePath(filepath, path.sep) + filestat.mtimeMs.toString()
-					),
+					hashPath(normalizeFilePath(relativeFilepath)),
+					// used absolute filepath here so that changes to the enclosing asset folder will be registered
+					hashPath(filepath + filestat.mtimeMs.toString()),
```
emily-shen, line 244: Note what goes into contentHash doesn't actually matter, as long as it is uniquely associated with that filepath and updates with file changes. Just removed the filepath encoding because it was unnecessary and because normalizeFilePath() expects a relative filepath now.

petebacondarwin (approved):

edmundhung (approved):

PR #6731, opened 2024-09-17
Allow setting readOnly mode in Quick Editor

WalshyDev:

Fixes [WC-2199](https://jira.cfdata.org/browse/WC-2199)

image

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: There are no tests but have deployed this change and it works on the dash
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e tests
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no public changes

### DIFF
--- a/packages/quick-edit-extension/src/cfs.ts
+++ b/packages/quick-edit-extension/src/cfs.ts
@@ -198,6 +198,7 @@ declare module "*.bin" {
 				create: true,
 				overwrite: true,
 				suppressChannelUpdate: true,
+				readOnly: files.readOnly,
 			});
 		}
 		if (this.readRoot !== null) {
--- a/.changeset/tall-needles-marry.md
+++ b/.changeset/tall-needles-marry.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-editor-shared": patch
+---
+
+chore: add readOnly property to WorkerLoaded event allowing us to put the editor into a read only state when loading a Worker.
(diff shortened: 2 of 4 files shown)


packages/quick-edit-extension/src/cfs.ts
```diff
@@ -198,6 +198,7 @@ declare module "*.bin" {
 				create: true,
 				overwrite: true,
 				suppressChannelUpdate: true,
+				readOnly: files.readOnly,
```
CarmenPopoviciu, line 201: propably not the case, but just to be sure...to we need to `readOnly: files.readOnly ?? false` here? Or will the editor handle a possible `readOnly: undefined`?

WalshyDev: The property didn't exist before and was fine so it should handle undefined without issue (and does seem to from my tests)
I didn't modify this object structure, just the loaded one

CarmenPopoviciu: all good. was merely a sanity check than anything else

CarmenPopoviciu (approved):

edmundhung (approved):

PR #6736, opened 2024-09-17
fix: update manifest/eTags on asset file content change

emily-shen:

Fixes [WC-2713](https://jira.cfdata.org/browse/WC-2713)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [x] TODO (before merge)
  - [ ] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/miniflare/src/plugins/assets/index.ts
+++ b/packages/miniflare/src/plugins/assets/index.ts
@@ -164,9 +164,14 @@ export const buildAssetsManifest = async (dir: string) => {
 	return encodedAssetManifest;
 };
 
+type ManifestEntry = {
+	pathHash: Uint8Array;
+	contentHash: Uint8Array;
+};
+
 const walk = async (dir: string) => {
 	const files = await fs.readdir(dir, { recursive: true });
-	const manifest: Uint8Array[] = [];
+	const manifest: ManifestEntry[] = [];
 	let counter = 0;
 	await Promise.all(
 		files.map(async (file) => {
@@ -195,10 +200,19 @@ const walk = async (dir: string) => {
 							`Ensure all assets in your assets directory "${dir}" conform with the Workers maximum size requirement.`
 					);
 				}
+				const modifiedTime = (await fs.stat(filepath)).mtimeMs;
 
-				manifest.push(
-					await hashPath(encodeFilePath(relativeFilepath, path.sep))
-				);
+				manifest.push({
+					pathHash: await hashPath(encodeFilePath(relativeFilepath, path.sep)),
+					// In prod the structure of each manifest entry is [header, pathHash, contentHash].
+					// In dev we want to avoid reading and hashing all file contents every time we
+					// reload the dev server, so in dev the 'content hash' is hash(path + modifiedTime).
+					// This way the 'content hash' still updates on file change, so the eTag also updates,
+					// and we don't incorrectly get a 304.
+					contentHash: await hashPath(
+						encodeFilePath(relativeFilepath, path.sep) + modifiedTime.toString()
+					),
+				});
 				counter++;
 			}
 		})
@@ -213,40 +227,42 @@ const walk = async (dir: string) => {
 	return manifest;
 };
 // sorts ascending by path hash
-const sortManifest = (manifest: Uint8Array[]) => {
+const sortManifest = (manifest: ManifestEntry[]) => {
 	return manifest.sort(comparisonFn);
 };
 
-const comparisonFn = (a: Uint8Array, b: Uint8Array) => {
+const comparisonFn = (a: ManifestEntry, b: ManifestEntry) => {
 	// i don't see why this would ever be the case
-	if (a.length < b.length) {
+	if (a.pathHash.length < b.pathHash.length) {
 		return -1;
 	}
-	if (a.length > b.length) {
+	if (a.pathHash.length > b.pathHash.length) {
 		return 1;
 	}
-	for (const [i, v] of a.entries()) {
-		if (v < b[i]) {
+	for (const [i, v] of a.pathHash.entries()) {
+		if (v < b.pathHash[i]) {
 			return -1;
 		}
-		if (v > b[i]) {
+		if (v > b.pathHash[i]) {
 			return 1;
 		}
 	}
 	return 1;
 };
 
-const encodeManifest = (manifest: Uint8Array[]) => {
+const encodeManifest = (manifest: ManifestEntry[]) => {
 	const assetManifestBytes = new Uint8Array(
 		HEADER_SIZE + manifest.length * ENTRY_SIZE
 	);
+
 	for (const [i, entry] of manifest.entries()) {
 		const entryOffset = HEADER_SIZE + i * ENTRY_SIZE;
-		// NB: PATH_HASH_OFFSET = 0
-		// set the path hash:
-		assetManifestBytes.set(entry, entryOffset + PATH_HASH_OFFSET);
-		// set the content hash, which happens to be the same as the path hash in dev:
-		assetManifestBytes.set(entry, entryOffset + CONTENT_HASH_OFFSET);
+		assetManifestBytes.set(entry.pathHash, entryOffset + PATH_HASH_OFFSET);
+		// NB content hash in dev is hash(path + modifiedTime)
+		assetManifestBytes.set(
+			entry.contentHash,
+			entryOffset + CONTENT_HASH_OFFSET
+		);
 	}
 	return assetManifestBytes;
 };
@@ -278,7 +294,10 @@ const createReverseMap = async (dir: string) => {
 				return;
 			} else {
 				const pathHash = bytesToHex(
-					await hashPath(encodeFilePath(relativeFilepath, path.sep))
+					await hashPath(
+						encodeFilePath(relativeFilepath, path.sep) +
+							filestat.mtimeMs.toString()
+					)
 				);
 
 				assetsReverseMap[pathHash] = {
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -7,6 +7,7 @@ import { fetch } from "undici";
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
 import { fetchText } from "./helpers/fetch-text";
+import { fetchWithETag } from "./helpers/fetch-with-etag";
 import { generateResourceName } from "./helpers/generate-resource-name";
 import { retry } from "./helpers/retry";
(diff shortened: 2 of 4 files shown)


packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -195,10 +200,19 @@ const walk = async (dir: string) => {
...
-				);
+				manifest.push({
+					pathHash: await hashPath(encodeFilePath(relativeFilepath, path.sep)),
+					// In prod the structure of each manifest entry is [header, pathHash, contentHash].
+					// In dev we want to avoid reading and hashing all file contents every time we
+					// reload the dev server, so in dev the 'content hash' is hash(path + modifiedTime).
+					// This way the 'content hash' still updates on file change, so the eTag also updates,
+					// and we don't incorrectly get a 304.
+					contentHash: await hashPath(
+						encodeFilePath(relativeFilepath, path.sep) + modifiedTime.toString()
```
GregBrimble, line 213: oh clever! I was going to hash the whole thing à la prod, but this saves an expensive computation!

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -278,7 +294,10 @@ const createReverseMap = async (dir: string) => {
 				return;
 			} else {
 				const pathHash = bytesToHex(
-					await hashPath(encodeFilePath(relativeFilepath, path.sep))
+					await hashPath(
+						encodeFilePath(relativeFilepath, path.sep) +
+							filestat.mtimeMs.toString()
```
GregBrimble, line 299: don't _love_ needing to keep this in sync with the other thing above, but it's fine for now.

packages/wrangler/e2e/dev.test.ts
```diff
@@ -901,18 +902,31 @@ describe("watch mode", () => {
 				const worker = helper.runLongLived(cmd);
 				const { url } = await worker.waitForReady();
 
-				let text = await fetchText(`${url}/index.html`);
-				expect(text).toBe("<h1>Hello Workers + Assets</h1>");
+				// let cachedETags: Record<string,string> = {};
```
GregBrimble, line 905:
```suggestion
```

GregBrimble (approved):
image

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -196,9 +212,26 @@ const walk = async (dir: string) => {
...
+				// reload the dev server, so in dev the 'content hash' is hash(path + modifiedTime).
+				// This way the 'content hash' still updates on file change, so the eTag also updates,
+				// and we don't incorrectly get a 304.
+				const pathHash = await hashPath(
+					encodeFilePath(relativeFilepath, path.sep)
+				);
+				const modifiedTime = (await fs.stat(filepath)).mtimeMs;
+				const contentHash = await hashPath(
+					encodeFilePath(relativeFilepath, path.sep) + modifiedTime.toString()
 				);
```
edmundhung, line 226: Nit: if you can reuse the `filestat` above, you can parallelize the two hashPath calls, e.g.
```ts
const [pathHash, contentHash] = await Promise.all(
  hashPath(encodeFilePath(relativeFilepath, path.sep)),
  hashPath(encodeFilePath(relativeFilepath, path.sep) + filestate.mtimeMs.toString())
);
```

edmundhung (approved):

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -57,13 +57,17 @@ export const ASSETS_PLUGIN: Plugin<typeof AssetsOptionsSchema> = {
...
 		}
-		const assetsReverseMap = await createReverseMap(options.assets?.path);
 
 		const storageServiceName = `${ASSETS_PLUGIN_NAME}:storage`;
 		const storageService: Service = {
 			name: storageServiceName,
 			disk: { path: options.assets.path, writable: true },
 		};
+
+		const { encodedAssetManifest, assetsReverseMap } = await buildAssetManifest(
```
CarmenPopoviciu, line 67: what is the reasoning for combining the asset manifest and the reverse map under one fn? What are we optimizing for? Also, how does this play out with our plans of sharing more code between `workers-shared` and `wrangler`/`miniflare`. Does `buildAssetManifest` qualify in that category?

emily-shen: Before it was traversing the asset directory twice, once each to create the reverse map and manifest, which wasn't great. And we can't really share buildAssetManifest because half of it happens in AUS for prod.

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -146,27 +150,39 @@ export const ASSETS_PLUGIN: Plugin<typeof AssetsOptionsSchema> = {
...
-//
-// 2. Sort and binary encode the asset manifest
-// This is available to asset service worker as a binding.
+// The Asset Manifest and Asset Reverse Map are used to map a request path to an asset.
+// 1. Hash path of request
+// 2. Use this path hash to find the manifest entry
+// 3. Get content hash from manifest entry
+// 4a. In prod, use content hash to get asset from KV
+// 4b. In dev, we fake out the KV store and use the file system instead.
+//     Use content hash to get file path from asset reverse map.
```
CarmenPopoviciu, line 160:
```suggestion
/*
 * The Asset Manifest and Asset Reverse Map are used to map a request path to an asset.
 * 1. Hash path of request
 * 2. Use this path hash to find the manifest entry
 * 3. Get content hash from manifest entry
 * 4a. In prod, use content hash to get asset from KV
 * 4b. In dev, we fake out the KV store and use the file system instead.
 *     Use content hash to get file path from asset reverse map.
 */
```

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -146,27 +150,39 @@ export const ASSETS_PLUGIN: Plugin<typeof AssetsOptionsSchema> = {
...
+	contentHash: Uint8Array;
 };
 
+export type AssetReverseMap = {
+	[pathHash: string]: { filePath: string; contentType: string };
+};
+
+/**
+ * Traverses the asset directory to create an asset manifest and asset reverse map.
+ * These are available to asset service worker as a binding.
```
CarmenPopoviciu, line 179:
```suggestion
 * These are available to asset service Worker as a binding.
```

packages/miniflare/src/plugins/assets/index.ts
CarmenPopoviciu: yeah I thought as much, but wanted to sanity check

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -196,9 +212,26 @@ const walk = async (dir: string) => {
...
 				}
 
-				manifest.push(
-					await hashPath(encodeFilePath(relativeFilepath, path.sep))
+				// Each manifest entry is [header, pathHash, contentHash].
+				// In dev we want to avoid reading and hashing every file every time we
+				// reload the dev server, so in dev the 'content hash' is hash(path + modifiedTime).
+				// This way the 'content hash' still updates on file change, so the eTag also updates,
+				// and we don't incorrectly get a 304.
+				const pathHash = await hashPath(
```
CarmenPopoviciu, line 220:
```suggestion
				/*
				 * Each manifest entry is a series of bytes that store data in a [header,
				 * pathHash, contentHash] format, where:
				 *   - `header` is a fixed ??x?? bytes reserved for ??
				 *   - `pathHash` is the hashed file path
				 *   - `contentHash` is the hashed file content
				 *
				 * The `contentHash` of a file is determined by reading the contents of
				 * the file, and applying a hash function on the read data. In local
				 * development, performing this operation for each asset file would
				 * become very expensive very quickly, as it would have to be performed
				 * every time a dev server reload is trigerred. In watch mode, depending
				 * on the user's setup, this could potentially be on every file change.
				 *
				 * To avoid this from becoming a performance bottleneck, we're doing
				 * things a bit differently for dev, and implementing the `contentHash`
				 * as a hash function of the file path and the modified timestamp.
				 * reload the dev server, so in dev the 'content hash' is hash(path + modifiedTime).
				 * This way a file's corresponding 'contentHash' will always update
				 * if the file changes, and `wrangler dev` will serve the updated asset
				 * files instead of incorrectly returning 304s.
				 */
```

packages/wrangler/e2e/dev.test.ts
```diff
@@ -943,20 +960,26 @@ describe("watch mode", () => {
...
+				));
+				expect(await response.text()).toBe("About Workers + Assets");
+
+				({ response, cachedETags } = await fetchWithETag(
+					`${url}/workers/index.html`,
+					cachedETags
+				));
+				expect(await response.text()).toBe("Cloudflare Workers!");
+
+				// expect 304 for the original asset as the content has not changed
```
CarmenPopoviciu, line 977: great!

CarmenPopoviciu (approved):

packages/miniflare/src/plugins/assets/index.ts
```diff
@@ -212,18 +214,36 @@ const walk = async (dir: string) => {
...
-					encodeFilePath(relativeFilepath, path.sep)
-				);
-				const modifiedTime = (await fs.stat(filepath)).mtimeMs;
-				const contentHash = await hashPath(
-					encodeFilePath(relativeFilepath, path.sep) + modifiedTime.toString()
-				);
+				/*
+				 * Each manifest entry is a series of bytes that store data in a [header,
+				 * pathHash, contentHash] format, where:
+				 *   - `header` is a fixed ??x?? bytes reserved for ??
```
CarmenPopoviciu, line 220: I didn't have the exact data here, so you'll need to fill in `??x?? bytes` and what they are reserved for. Sorry abt that

emily-shen: oops yes will do

packages/wrangler/e2e/dev.test.ts
```diff
@@ -1044,12 +1044,13 @@ describe("watch mode", () => {
 						`,
 					"public/index.html": dedent`
 								<h1>Hello Workers + Assets</h1>`,
+				});
+				await helper.seed({
 					"public2/index.html": dedent`
 								<h1>Hola Workers + Assets</h1>`,
 					"public2/about/index.html": dedent`
 								<h1>Read more about Workers + Assets</h1>`,
 				});
```
edmundhung, line 1053: I am curious why a separate seed helps?

emily-shen: ah tbh its not necessary given we're using the absolute filepath now (it would have given it a different file creation/modified time and thus different hash)

PR #6749, opened 2024-09-18
fix: Throw error when attempting to configure Workers with assets and tail consumers

CarmenPopoviciu:
Tail Workers are currently not supported for Workers with assets. This commit ensures we throw the corresponding error if users are attempting to configure `tail_consumers` via their configuration file, for a Worker with assets. This validation is applied for all `wrangler dev`, `wrangler deploy`, `wrangler versions upload`.

Fixes [WC-2711](https://jira.cfdata.org/browse/WC-2711)

Screenshot 2024-09-18 at 13 33 57

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not docummented

### DIFF
--- a/packages/wrangler/src/deploy/index.ts
+++ b/packages/wrangler/src/deploy/index.ts
@@ -3,7 +3,10 @@ import path from "node:path";
 import { findWranglerToml, readConfig } from "../config";
 import { getEntry } from "../deployment-bundle/entry";
 import { UserError } from "../errors";
-import { processExperimentalAssetsArg } from "../experimental-assets";
+import {
+	processExperimentalAssetsArg,
+	verifyMutuallyExclusiveAssetsArgsOrConfig,
+} from "../experimental-assets";
 import {
 	getRules,
 	getScriptName,
@@ -231,9 +234,9 @@ export function deployOptions(yargs: CommonYargsArgv) {
 	);
 }
 
-export async function deployHandler(
-	args: StrictYargsOptionsToInterface<typeof deployOptions>
-) {
+export type DeployArgs = StrictYargsOptionsToInterface<typeof deployOptions>;
+
+export async function deployHandler(args: DeployArgs) {
 	await printWranglerBanner();
 
 	// Check for deprecated `wrangler publish` command
@@ -281,22 +284,15 @@ export async function deployHandler(
 	}
 
 	if (
-		(args.legacyAssets ||
-			config.legacy_assets ||
-			args.experimentalAssets ||
-			config.experimental_assets) &&
+		(args.legacyAssets || config.legacy_assets) &&
 		(args.site || config.site)
 	) {
 		throw new UserError(
-			"Cannot use Assets and Workers Sites in the same Worker."
+			"Cannot use Legacy Assets and Workers Sites in the same Worker."
 		);
 	}
 
-	if (args.assets) {
-		logger.warn(
-			"The --assets argument is experimental and may change or break at any time"
-		);
-	}
+	verifyMutuallyExclusiveAssetsArgsOrConfig(args, config);
 
 	const experimentalAssetsOptions = processExperimentalAssetsArg(args, config);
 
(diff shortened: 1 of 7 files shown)


packages/wrangler/src/deploy/index.ts
```diff
@@ -281,22 +284,15 @@ export async function deployHandler(
...
+		(args.legacyAssets || config.legacy_assets) &&
 		(args.site || config.site)
 	) {
 		throw new UserError(
-			"Cannot use Assets and Workers Sites in the same Worker."
+			"Cannot use Legacy Assets and Workers Sites in the same Worker."
 		);
 	}
 
-	if (args.assets) {
```
CarmenPopoviciu, line 295: removing this because duplicate of what we have on line `249` 👇 

```
if (args.assets) {
		logger.warn(
			`The --assets argument is experimental. We are going to be changing the behavior of this experimental command after August 15th.\n` +
				`Releases of wrangler after this date will no longer support current functionality.\n` +
				`Please shift to the --legacy-assets command to preserve the current functionality.`
		);
	}
```

packages/wrangler/src/dev.tsx
```diff
@@ -600,24 +603,22 @@ export async function startDev(args: StartDevOptions) {
...
-			args.forceLocal = true;
-		}
-
-		/*
-		 * - `config.legacy_assets` conflates `legacy_assets` and `assets`
-		 * - `args.legacyAssets` conflates `legacy-assets` and `assets`
-		 */
 		if (
 			(args.legacyAssets || config.legacy_assets) &&
-			experimentalAssetsOptions
```
CarmenPopoviciu, line 614: keeping experimental assets and legacy assets validation separate here, for consistency with the rest of our code base

packages/wrangler/src/dev.tsx
```diff
@@ -600,24 +603,22 @@ export async function startDev(args: StartDevOptions) {
...
+			(args.site || config.site)
 		) {
 			throw new UserError(
-				"Cannot use Legacy Assets and Experimental Assets in the same Worker."
+				"Cannot use Legacy Assets and Workers Sites in the same Worker."
 			);
 		}
 
+		verifyMutuallyExclusiveAssetsArgsOrConfig(args, config);
```
CarmenPopoviciu, line 615: we should always perform this validation **before** we validate that the assets dir exists. IMHO this ordering makes more sense, since we'd want to first establish which of the mutually exclusive config keys/args the user actually meant, and once we're clear on that, we can proceed to perform validation specific to that config/arg

packages/wrangler/src/experimental-assets.ts
```diff
@@ -366,6 +368,51 @@ export function processExperimentalAssetsArg(
...
+
+	if (
+		(args.experimentalAssets || config.experimental_assets) &&
+		(args.site || config.site)
+	) {
+		throw new UserError(
+			"Cannot use Experimental Assets and Workers Sites in the same Worker."
+		);
+	}
+
```
emily-shen, line 405: what about this combo?
```
if (
		(args.legacyAssets || config.legacyAssets) &&
		(args.site || config.site)
	) {
		throw new UserError(
			"Cannot use Legacy Assets and Workers Sites in the same Worker."
		);
	}
```

edmundhung (approved):

packages/wrangler/src/experimental-assets.ts
CarmenPopoviciu: this is handled separately outside this fn (see [this comment](https://github.com/cloudflare/workers-sdk/pull/6749/files#r1764903581)). This fn should only handle experimental assets.

Issue #2805, opened 2023-02-26
🐛 BUG: Wrangler will incorrectly disregard `exports` if `browser`/`module`/`main` are defined.


balupton:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What version of `Wrangler` are you using?

2.11.1

### What operating system are you using?

Mac

### Describe the Bug

Wrangler in practice ignores its documented behaviour at:

- https://developers.cloudflare.com/workers/wrangler/bundling/#conditional-exports
- https://runtime-keys.proposal.wintercg.org/#workerd

If the `module`, `browser`, or `main` fields are defined.

---

I've setup a proof of failure repo at:

https://github.com/balupton/astro-cfpages-esbuild-resolution-bug

```
> npm install
> npx wrangler pages dev public/
🚧 'wrangler pages <command>' is a beta command. Please report any issues to https://github.com/cloudflare/workers-sdk/issues/new/choose
▲ [WARNING] No compatibility_date was specified. Using today's date: 2023-02-27.

  Pass it in your terminal:
  --compatibility-date=2023-02-27
  See https://developers.cloudflare.com/workers/platform/compatibility-dates/ for more information.


Compiling worker to "/var/folders/3v/fjmy4fyx1p9c28v31vzm8j9r0000gn/T/functionsWorker-0.773551694099083.mjs"...
✨ Compiled Worker successfully
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ [b] open a browser, [d] open Devtools, [l] turn off local mode, [c] clear console, [x] to exit                                                                                  │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
Debugger listening on ws://127.0.0.1:55956/5cee4c70-78a6-4b16-be7e-69b1d6b359fd
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
[pages:wrn] Service bindings are experimental. There may be breaking changes in the future.
[pages:err] ReferenceError: window is not defined
    at browser (/private/Users/balupton/Projects/websites/astro-cfpages-esbuild-resolution-bug/conditional-resolution-package/browser.mjs:2:2)
    at /private/Users/balupton/Projects/websites/astro-cfpages-esbuild-resolution-bug/functions/[[catchall]].tsx:3:16
    at SourceTextModule.evaluate (node:internal/vm/module:225:23)
[...]

PR #6752, opened 2024-09-18
refactor: use `require.resolve` and remove `resolve` dependency

bluwy:

Fixes N/A

It looks like we can simplify the resolve using `require.resolve`, similar to how other parts of the codebase that also does so:

https://github.com/cloudflare/workers-sdk/blob/0deb42b2b6b5960d0bd79884471805069c0f29b0/packages/vitest-pool-workers/src/pool/module-fallback.ts#L292

https://github.com/cloudflare/workers-sdk/blob/0deb42b2b6b5960d0bd79884471805069c0f29b0/packages/wrangler/src/deployment-bundle/bundle.ts#L332-L337

This allows the package to drop the `resolve` dependency, which isn't used anywhere else in the `wrangler` prod dependency graph.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: refactors code
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: refactors code
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


bluwy:
I had to remove a test to fix the fail. Looks like it's because the test was importing a non-exported entrypoint here:

https://github.com/cloudflare/workers-sdk/blob/bea65583f72fcf8346ed82d4b9b1176669bcfd23/fixtures/import-wasm-example/src/index.js#L4

It worked with the `resolve` package because it does not support the package `"exports"` field when resolving paths, but node (since v12.7.0) had supported that field when resolving, so it prevents accessing non-exported paths.

Technically it could be a breaking change that it now respects the `"exports"` field when resolving packages, but on the other hand, `"exports"` field had been supported by most tooling for many years now that it could be seen as a fix, but I'll leave y'all to decide. I don't mind closing this PR if we're not ready to change that yet.

penalosa:
@bluwy Thanks for this contribution! Given it requires removing a test I don't think we'd want to merge it as-is—we do want to preserve that test's behaviour, if possible

bluwy:
Alright I'll close this for now then, but I think it's something worth revisiting again in a breaking version to respect the `"exports"` field. https://github.com/cloudflare/workers-sdk/issues/2805 might be related.

PR #6753, opened 2024-09-18
refactor: prevent bundling entire `package.json` in built code

bluwy:

Fixes N/A

I noticed that the `package.json` was bundled in `wrangler/wrangler-dist/cli.js`. I found that we can simplify the `package.json` import to prevent that from happening.

The `package.json` is used by the `check-update` dependency, but it only requires the `name` and `version` to detect updates: https://github.com/vercel/update-check/blob/4338e215c9d089f46c6edff45ba41941c070d39f/index.js#L169-L201. So we only need to import the name and version to satisfy it.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: refactors
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: refactors
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/eight-emus-compete.md
+++ b/.changeset/eight-emus-compete.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+refactor: prevent bundling entire `package.json` in built code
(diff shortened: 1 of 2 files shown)


penalosa (approved):
LGTM! Thanks for this!

PR #6754, opened 2024-09-18
refactor: move `@cloudflare/workers-shared` as dev dependency

bluwy:

Fixes N/A

It looks like `@cloudflare/workers-shared` is being bundled so it can be moved as a dev dependency instead. It was moved as a dependency in https://github.com/cloudflare/workers-sdk/pull/6432, but after the refactors from https://github.com/cloudflare/workers-sdk/pull/6564, it should be safe to move it back as a dev dependency again.

I also noticed that the published code for `@cloudflare/workers-shared` isn't quite right (?) as there's no entrypoint when importing `@cloudflare/workers-shared` directly. So externalizing it may not work.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: refactor
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: refactor
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


CarmenPopoviciu (approved):

CarmenPopoviciu:
thank you @bluwy for your contribution ♥️ . This is merged now :)

PR #6756, opened 2024-09-18
Disable warnings for invalid config in wrangler login & logout

WalshyDev:

Fixes [WO-184](https://jira.cfdata.org/browse/WO-184)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no changes
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no public changes

### DIFF
--- a/.changeset/selfish-readers-exercise.md
+++ b/.changeset/selfish-readers-exercise.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+chore: disable wrangler.toml warnings when doing `wrangler login` & `wrangler logout`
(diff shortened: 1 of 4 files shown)


penalosa:
I'm not sure I understand the intention here—in this case there is an invalid `wrangler.toml` file, and I think we do want to warn. It's not a hard error, so this shouldn't be causing user issues

WalshyDev:
Yeah but warning on an invalid wrangler.toml during login/logout just doesn't make much sense - we read it just for the metrics value but we don't use anything else.

A user seeing a warning after logging in is confusing.

threepointone:
Actually a good idea, I like it

penalosa (approved):

penalosa:
For context, I had originally blocked this because I had thought that config being invalid could potentially cause weird behaviour in login, primarily around the `account_id` field. Having read the PR more closely, since we're only reading config _after_ actually logging in or out that concern is removed. Thanks @WalshyDev for this!

edmundhung (approved):

PR #6760, opened 2024-09-19
fix: add a missing experimental template to C3 for asset-only Worker

petebacondarwin:

Fixes N/A

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: experimental C3 template
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: experimental C3 template
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: experimental C3 template


edmundhung (approved):

RamIdeas (approved):

packages/create-cloudflare/templates-experimental/hello-world-durable-object-with-assets/js/__dot__gitignore
```diff
@@ -0,0 +1,172 @@
+# Logs
+
+logs
+_.log
+npm-debug.log_
+yarn-debug.log*
+yarn-error.log*
+lerna-debug.log*
+.pnpm-debug.log*
+
```
edmundhung, line 10: I just notice that these `__dot_gitignore` file should be ignored. There is already an entry to ignore all of these on `templates` so you probably need to add one more for the `templates-experimental`:

```sh
# Already in .gitignore
templates/**/__dot__gitignore
# I think we should add this
templates-experimental/**/__dot__gitignore
```

We might have it added with other experimental templates too. So it would be great if you can remove them as well. 👍🏼

edmundhung (changes requested):

petebacondarwin:
Oops sorry @edmundhung - I merged it before I saw your review. I'll follow up now with another PR.

PR #6767, opened 2024-09-19
chore: An attempt at a simpler bug report / pull request form

threepointone:
This is an attempt at simplifying our bug report / pull request forms. Some of it might seem controversial, so approach it with an open mind. Let's have a conversation here!

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-template.yaml
+++ b/.github/ISSUE_TEMPLATE/bug-template.yaml
@@ -1,86 +1,28 @@
 name: "\U0001F41B Bug Report"
 description: Report an issue or possible bug
-title: "\U0001F41B BUG:"
+title: "\U0001F41B Bug:"
 labels: ["bug"]
 
 body:
   - type: markdown
     attributes:
       value: |
-        ## Quick Bug Form
-        Thank you for taking the time to file a bug report! Please fill out this form as completely as possible.
-  - type: dropdown
-    id: products
+        ## Please provide as many details as possible regarding your bug.
+  - type: input
     attributes:
-      label: Which Cloudflare product(s) does this pertain to?
-      multiple: true
-      options:
-        - C3 (npm create cloudflare)
-        - D1
-        - KV
-        - Pages
-        - Pub/Sub
-        - Queues
-        - R2
-        - Workers for Platforms
-        - Workers AI
-        - Workers Runtime
-        - Wrangler
-        - Miniflare
-        - KV Asset Handler
-        - Workers Vitest Integration
-        - Other
+      label: What versions are you using?
+      placeholder: 0.0.0 [Wrangler], 0.0.0 [C3], 0.0.0 [Node.js], etc.
     validations:
       required: true
   - type: input
     attributes:
-      label: What version(s) of the tool(s) are you using?
-      placeholder: 0.0.0 [Wrangler], 0.0.0 [C3], etc.
-    validations:
-      required: true
-  - type: input
-    attributes:
-      label: What version of Node are you using?
-      placeholder: 0.0.0
-    validations:
-      required: false
-  - type: input
-    attributes:
-      label: What operating system and version are you using?
+      label: What Operating System are you using?
       placeholder: Mac Sonoma 14.2.1, Windows 11, Linux (or WSL) Debian Bookworm, etc.
     validations:
-      required: true
+      required: false
   - type: textarea
     attributes:
       label: Describe the Bug
-      description: Please describe the bug and include steps to reproduce.
-      value: |
-        ### Observed behavior
-        Please describe.
-
-        ### Expected behavior
-        Please describe.
-
-        ### Steps to reproduce
-        Please provide the following:
-        - A minimal working subset of your worker code
-        - A minimal working subset of your `wrangler.toml`
-        - Commands used to start your local dev server, including custom env and cli args
-        - Steps to be performed in the browser, curl commands, or a test we can run that reliably fails (at least a percent of the time)
-
-        A git repo we can clone and run a test suite on, or which has a README with step-by-step instructions, is even better. In this case, please use the field below to provide a link to the minimal repro.
+      description: Please describe the bug and include steps to reproduce. Although not required, we often request a minimal reproduction to help troubleshoot the issue, so providing this up-front streamlines the process towards resolution. Please include any error logs if possible.
     validations:
       required: true
-  - type: input
-    attributes:
-      label: Please provide a link to a minimal reproduction
-      description: Although not required, we often request a minimal reproduction to help troubleshoot the issue, so providing this up-front streamlines the process towards resolution.
-      placeholder: https://github.com/foobar-user/minimal-repro
-    validations:
-      required: false
-  - type: textarea
-    attributes:
-      label: Please provide any relevant error logs
-      description: Although not required, we often request logs to help troubleshoot the issue, so providing this up-front streamlines the process towards resolution. Please be careful to hide any sensitive information.
-    validations:
-      required: false
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,30 +4,12 @@ Fixes #[insert GH or internal issue number(s)].
 
 ## Author has addressed the following
 
-- Tests
-  - [ ] TODO (before merge)
-  - [ ] Tests included
-  - [ ] Tests not necessary because:
-- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
-  - [ ] I don't know
-  - [ ] Required
-  - [ ] Not required because:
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -1,86 +1,28 @@
...
+title: "\U0001F41B Bug:"
 labels: ["bug"]
 
 body:
   - type: markdown
     attributes:
       value: |
-        ## Quick Bug Form
-        Thank you for taking the time to file a bug report! Please fill out this form as completely as possible.
-  - type: dropdown
```
threepointone, line 12: This is usually wrangler or c3, and when it's not, the user will tell us what the product is. We can make the form less intimidating by removing it.

.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -1,86 +1,28 @@
...
-        - KV Asset Handler
-        - Workers Vitest Integration
-        - Other
-    validations:
-      required: true
+        ## Please provide as many details as possible regarding your bug.
   - type: input
     attributes:
-      label: What version(s) of the tool(s) are you using?
-      placeholder: 0.0.0 [Wrangler], 0.0.0 [C3], etc.
```
threepointone, line 38: Let's have them add the node version here, and make it optional. We aim to support all (modern) versions, and if a bug is for an odler version it's usually immediately apparent.

.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -1,86 +1,28 @@
...
-      placeholder: 0.0.0
-    validations:
-      required: false
-  - type: input
-    attributes:
-      label: What operating system and version are you using?
+      label: What Operating System are you using?
       placeholder: Mac Sonoma 14.2.1, Windows 11, Linux (or WSL) Debian Bookworm, etc.
     validations:
-      required: true
```
threepointone, line 52: We can ask as a followup if it's not obvious from the report.

.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -1,86 +1,28 @@
...
+      required: false
   - type: textarea
     attributes:
       label: Describe the Bug
-      description: Please describe the bug and include steps to reproduce.
-      value: |
-        ### Observed behavior
```
threepointone, line 58: The issue is meant to be a starting point for a conversation, unless it's obvious from the report. We can simplify the wording here.

.github/pull_request_template.md
```diff
@@ -4,30 +4,12 @@ Fixes #[insert GH or internal issue number(s)].
...
+- [ ] E2E Tests (Use "e2e" label to run them on CI, or ask maintainer to run separately)
+- [ ] [Changeset](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets)
+- [ ] Docs
 
 <!--
 Have you read our [Contributing guide](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md)?
-In particular, for non-trivial changes, please always engage on the issue or create a discussion or feature request issue first before writing your code.
 -->
 
-<!--
```
threepointone, line 29: This has never been used. Let's remove it.

.github/pull_request_template.md
```diff
@@ -4,30 +4,12 @@ Fixes #[insert GH or internal issue number(s)].
...
-  - [ ] Cloudflare docs PR(s): <!--e.g. <https://github.com/cloudflare/cloudflare-docs/pull/>...-->
-  - [ ] Documentation not necessary because:
+- [ ] Tests
+- [ ] E2E Tests (Use "e2e" label to run them on CI, or ask maintainer to run separately)
+- [ ] [Changeset](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets)
+- [ ] Docs
 
 <!--
 Have you read our [Contributing guide](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md)?
-In particular, for non-trivial changes, please always engage on the issue or create a discussion or feature request issue first before writing your code.
```
threepointone, line 26: People (like me) might use the PR itsel f as an engagement point (and we should actually encourage it).

.github/pull_request_template.md
```diff
@@ -4,30 +4,12 @@ Fixes #[insert GH or internal issue number(s)].
...
-  - [ ] Changeset included
-  - [ ] Changeset not necessary because:
-- Public documentation
-  - [ ] TODO (before merge)
-  - [ ] Cloudflare docs PR(s): <!--e.g. <https://github.com/cloudflare/cloudflare-docs/pull/>...-->
-  - [ ] Documentation not necessary because:
+- [ ] Tests
+- [ ] E2E Tests (Use "e2e" label to run them on CI, or ask maintainer to run separately)
+- [ ] [Changeset](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets)
+- [ ] Docs
```
penalosa, line 10: What's the main motivation here? There's definitely some simplification in order with this template, but it seems like the main change here is removing the requirement to justify omitting tests/changeset/etc.... Does that present a huge burden to external contributors?

.github/pull_request_template.md
penalosa: One thing I'd be slightly concerned about with this change is an external contributor opening up a PR for a feature we don't want to add for whatever reason, and then us having to close the PR after they've already done a bunch of work (as opposed to closing an issue after just a suggestion of a feature). There probably won't be loads of cases of that though, to be fair, and maybe that's an okay tradeoff?

.github/pull_request_template.md
threepointone: For internal contributors, we can always have higher standards. Those don't have to be reflected on the form.

.github/pull_request_template.md
threepointone: Yup, that's what I think.

.github/pull_request_template.md
RamIdeas: How do we enforce the higher standards on internal contributions?

I'd love to encourage more external contributions but Wrangler is an active product and internal contributions are far more common

.github/ISSUE_TEMPLATE/bug-template.yaml
RamIdeas: Good shout collapsing the versions question into less inputs 👍

.github/pull_request_template.md
threepointone: I guess my point is all the existing checkboxes don't really enforce it valuably enough to outweigh the cons, and we can still write automation around a smaller set of checkboxes.

threepointone: But part of me also wants to make it possible for internal folks to do driveby fixes with the same freedom we give external contributors. We use it the most ourselves so I bet we'd like to quickly send fixes too.

.github/pull_request_template.md
petebacondarwin: I suspect what would actually happen in practice, rather than the external contributor going and creating an issue for discussion first, they would spend ages making the PR, then after pushing it realise that they need an issue, go create the issue, and then we have to close both the PR and issue if it is something we didn't want in the first place.

threepointone: To clarify, that means you're in support of removing this line/requirement right? Because you're describing the situation as it exists right now?

(I also want to make sure we're _not_ optimising to prevent the rare scenario of an external contributor making a big PR without ever talking to us. Has it happened often?)

lrapoport-cf: for some context here, the reason this was added is for exactly the scenario @penalosa cited -- we were finding ourselves having to close PRs that implemented features that hadn't been discussed with the team, but which didn't make sense to include for various reasons. the goal is to save folks from doing work that won't be landable. 

i agree that at the point of someone seeing this it's likely too late to save them doing the work, but i think having some language to signal process can 1) help manage expectations if the PR does end up being closed, and 2) reduce potential future friction

(as a note, while this is specified in the contributing guide [here](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md?plain=1#L9), it can be easily missed)

.github/pull_request_template.md
lrapoport-cf: definitely want to reduce friction where possible. however my concern here is that by removing the multiple choice, there is no signal to our team on whether a checkbox was intentionally left unchecked or just forgotten. and if intentionally left unchecked, whether it was valid to do so. while we can engage in the PR on these items to clarify, it takes more overall back and forth than if the information were provided up front.

lrapoport-cf: i'd also be curious to understand better the extent of the con for external folks. do you know of cases where people don't put up PRs because of the checkboxes?

penalosa:
Thanks for this—we're going to close this PR for now because it's a bit stale, but we'll add some of the bug report template modifications as a follow-up PR

PR #6770, opened 2024-09-19
Test fixups & skip tests for docs changes

### LINKED ISSUE #6685: add support for skipping required checks
for some PRs, running the the full gamut of checks is unnecessary and inefficient. add support to skip required checks for PRs which do not benefit from the entire suite. 

example PRs where this would have been helpful/relevant:

- https://github.com/cloudflare/workers-sdk/pull/6660
- https://github.com/cloudflare/workers-sdk/pull/6710
- https://github.com/cloudflare/workers-sdk/pull/6737
- https://github.com/cloudflare/workers-sdk/pull/6637

reference: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks

penalosa:

Fixes #6685

* Fixup tests to not run E2E tests on forks
* Only run normal tests when a non-markdown file has been changed (example PR: https://github.com/cloudflare/workers-sdk/pull/6774)

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: workflows change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: workflows change
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: this shouldn't trigger a release
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: workflows change


.github/workflows/e2e.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.node }}
       cancel-in-progress: true
     timeout-minutes: 40
-    if: github.repository_owner == 'cloudflare' && (github.event_name != 'pull_request' || (github.event_name == 'pull_request' && contains(github.event.*.labels.*.name, 'e2e' )))
+    if: (github.event_name != 'pull_request' || (github.event_name == 'pull_request' && contains(github.event.*.labels.*.name, 'e2e' ) && github.event.pull_request.head.repo.owner.login == 'cloudflare'))
```
penalosa, line 17: Turns out `repository_owner` is always `cloudflare`, even for forks

.github/workflows/test-and-check.yml
```diff
@@ -1,10 +1,12 @@
...
 
-on: pull_request
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
 
 jobs:
   add-to-project:
-    if: github.ref != 'refs/heads/changeset-release/main'
+    if: github.head_ref != 'changeset-release/main'
```
penalosa, line 9: Turns out `github.ref` for `pull_request` events refers to the pull request ref (e.g. `refs/6770/merge`)

.github/workflows/test-and-check.yml
```diff
@@ -89,6 +91,61 @@ jobs:
...
+      cancel-in-progress: true
+
+    name: "Tests"
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-13]
+        filter: ["./fixtures/*"]
+    runs-on: ${{ matrix.os }}
+    if: github.head_ref == 'changeset-release/main' || contains(github.event.*.labels.*.name, 'fixtures' )
```
penalosa, line 107: This is the pertinent change—only run fixtures if the fixtures label is applied

.github/workflows/test-and-check.yml
```diff
@@ -100,15 +157,23 @@ jobs:
...
         with:
           fetch-depth: 0
 
+      - uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          filters: |
+            everything_but_markdown:
+              - '!**/*.md'
+
```
penalosa, line 174: Only run the remaining tests if a non-markdown file was changed, allowing e.g. README fixes to be light on CI time

.github/workflows/test-and-check.yml
RamIdeas: Can we auto-apply this label under certain circumstances?

eg. when the fixtures dir has been touched, when wrangler/src has been touched, etc

.github/workflows/test-and-check.yml
```diff
@@ -100,15 +157,27 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-13]
-        filter: ["./fixtures/*", "./packages/*", "./tools"]
+        filter: ["./packages/*"]
+        # Things in the tools folder are for running in CI, and so only need to run on linux
+        include:
+          - os: ubuntu-latest
+            filter: "./tools"
```
edmundhung, line 164: 👍🏼

.github/workflows/test-and-check.yml
penalosa: fab3a85f286d7ac32ec605c9490f586d96499a6f

.github/workflows/test-and-check.yml
```diff
@@ -89,6 +91,61 @@ jobs:
...
+        env:
+          TMP_CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          TMP_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          NODE_OPTIONS: "--max_old_space_size=8192"
+          WRANGLER_LOG_PATH: ${{ runner.temp }}/wrangler-debug-logs/
+          TEST_REPORT_PATH: ${{ runner.temp }}/test-report/index.html
+          CI_OS: ${{ runner.os }}
+
+      - name: Run changeset version tests
+        run: node .github/changeset-version.test.js
```
edmundhung, line 133: Do we need to test the changeset version everytime? Maybe it could be a seperate job?

penalosa: Potentially, but I'm not sure it's worth it—it's really fast, and I think the overhead of spinning up another job could be worse

edmundhung: How about moving it to `./tools`?

petebacondarwin:
> Only run fixture tests when the fixtures label is applied (or on Version Packages)

I think this is dangerous. There are a lot of very useful tests in fixtures that actually test significant functionality.
But perhaps what we need to do is pull those tests out of fixtures and put them somewhere else if we are going to not both running the tests on most PRs.

penalosa:
> There are a lot of very useful tests in fixtures that actually test significant functionality.

They're also really flaky, though. In most cases (especially from forks where the turbo cache isn't available) I think any signal in fixtures is lost to the noise of flakes. We'll still be running them on Version Packages, so we won't be releasing with fixture failures. Of course, this might end up in loads of things needing to be reverted/fixed in Version Packages, but if that's the case we can revisit things—it seems like it's worth a try to make opening PRs and contributing smoother?

.github/workflows/test-and-check.yml
penalosa: https://github.com/cloudflare/workers-sdk/pull/6770/commits/240db178440792e84cbe189822d331413f87c1e2

petebacondarwin:
Could we then only skip the flaky ones if this label is not applied?

penalosa:
> Could we then only skip the flaky ones if this label is not applied?

And keep the rest running? That could work

edmundhung (approved):

tools/deployments/__tests__/changeset-version.test.ts
```diff
@@ -1,5 +1,6 @@
-const assert = require("node:assert");
-const { getNextMiniflareVersion } = require("./changeset-version.js");
+import { expect, it } from "vitest";
+// @ts-expect-error This is a JS file
+import { getNextMiniflareVersion } from "../../../.github/changeset-version";
```
RamIdeas, line 3: you should be able to import js from ts

PR #6775, opened 2024-09-19
fix(wrangler): Support switching between static and dynamic Workers

CarmenPopoviciu:

This commit fixes the current behaviour of watch mode for Workers with assets, and adds support for switching between static and dynamic Workers within a single `wrangler dev` session.

Fixes [WC-2721](https://jira.cfdata.org/browse/WC-2721)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -1087,6 +1087,110 @@ describe("watch mode", () => {
 					"<h1>Read more about Workers + Assets</h1>"
 				);
 			});
+
+			it(`supports switching from assets-only to Workers with assets during the current dev session`, async () => {
+				const helper = new WranglerE2ETestHelper();
+				await helper.seed({
+					"wrangler.toml": dedent`
+								name = "${workerName}"
+								compatibility_date = "2023-01-01"
+
+								[experimental_assets]
+								directory = "./public"
+						`,
+					"public/index.html": dedent`
+								<h1>Hello Workers + Assets</h1>`,
+				});
+				const worker = helper.runLongLived(cmd);
+				const { url } = await worker.waitForReady();
+
+				// verify response from Asset Worker
+				let { response } = await fetchWithETag(`${url}/index.html`, {});
+				expect(await response.text()).toBe("<h1>Hello Workers + Assets</h1>");
+
+				// verify no response from route that will be handled by the
+				// User Worker in the future
+				({ response } = await fetchWithETag(`${url}/hey`, {}));
+				expect(await response.status).toBe(404);
+
+				await helper.seed({
+					"wrangler.toml": dedent`
+							name = "${workerName}"
+							main = "src/index.ts"
+							compatibility_date = "2023-01-01"
+
+							[experimental_assets]
+							directory = "./public"
+					`,
+					"src/index.ts": dedent`
+						export default {
+							fetch(request) {
+								return new Response("Hello from user Worker!")
+							}
+						}`,
+				});
+
+				await worker.waitForReload();
+
+				({ response } = await fetchWithETag(`${url}/index.html`, {}));
+				expect(await response.text()).toBe("<h1>Hello Workers + Assets</h1>");
+
+				({ response } = await fetchWithETag(`${url}/hey`, {}));
+				expect(await response.text()).toBe("Hello from user Worker!");
+			});
+
+			it(`supports switching from Workers with assets to assets-only during the current dev session`, async () => {
+				const helper = new WranglerE2ETestHelper();
+				await helper.seed({
+					"wrangler.toml": dedent`
+							name = "${workerName}"
+							main = "src/index.ts"
+							compatibility_date = "2023-01-01"
+
+							[experimental_assets]
+							directory = "./public"
+					`,
+					"public/index.html": dedent`
+							<h1>Hello Workers + Assets</h1>`,
+					"src/index.ts": dedent`
+						export default {
+							fetch(request) {
+								return new Response("Hello from user Worker!")
+							}
+						}`,
+				});
+
+				const worker = helper.runLongLived(cmd);
+				const { url } = await worker.waitForReady();
+
+				// verify response from Asset Worker
+				let { response } = await fetchWithETag(`${url}/index.html`, {});
+				expect(await response.text()).toBe("<h1>Hello Workers + Assets</h1>");
+
+				// verify response from User Worker
+				({ response } = await fetchWithETag(`${url}/hey`, {}));
+				expect(await response.text()).toBe("Hello from user Worker!");
+
+				await helper.seed({
+					"wrangler.toml": dedent`
+								name = "${workerName}"
+								compatibility_date = "2023-01-01"
+
+								[experimental_assets]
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/dev.tsx
```diff
@@ -836,11 +836,39 @@ export async function startDev(args: StartDevOptions) {
...
 					logger.log(`${path.basename(config.configPath)} changed...`);
 
+					/*
+					 * If this is a Worker with assets, we want to enable switching
+					 * from assets only Worker to Worker with assets and vice versa
+					 * during local development. For that, we need to watch for changes
+					 * in the `wrangler.toml` file, specifically the `main` configuration
+					 * key, and re-asses the entry point every time.
+					 */
+					if (experimentalAssetsOptions?.directory) {
```
CarmenPopoviciu, line 846: technically, we could support this for all Workers, not just Workers with assets. Is this something we would like to do?

packages/wrangler/e2e/dev.test.ts
```diff
@@ -1176,5 +1280,102 @@ describe("watch mode", () => {
...
+			});
+			const worker = helper.runLongLived(cmd);
+			const { url } = await worker.waitForReady();
+
+			// verify response from Asset Worker
+			let { response } = await fetchWithETag(`${url}/index.html`, {});
+			expect(await response.text()).toBe("<h1>Hello Workers + Assets</h1>");
+			expect(await response.status).toBe(200);
+
+			// verify nresponse from User Worker
```
penalosa, line 1357:
```suggestion
			// verify response from User Worker
```

penalosa (approved):
LGTM. Does this work with the `--x-dev-env` flow as well?

packages/wrangler/e2e/dev.test.ts
```diff
@@ -1087,6 +1087,110 @@ describe("watch mode", () => {
...
+								directory = "./public"
+						`,
+					"public/index.html": dedent`
+								<h1>Hello Workers + Assets</h1>`,
+				});
+				const worker = helper.runLongLived(cmd);
+				const { url } = await worker.waitForReady();
+
+				// verify response from Asset Worker
+				let { response } = await fetchWithETag(`${url}/index.html`, {});
```
emily-shen, line 1108: sorry the `fetchWithETag` helper is a bit clunky to use - if you don't return the cached ETags and pass it in the next time, it's not really doing anything.

like this:
```
let { response, cachedETags } = await fetchWithETag(`${url}/index.html`, {});
({ response, cachedETags} = await fetchWithETag(`${url}/index.html`, cachedETags));
// then you need should expect just a 304 not 200 with text etc.
expect(response.status).toBe(304)
```

Probably just use the regular fetch here since we're testing the eTag stuff separately.

CarmenPopoviciu: yeah, fair enough! I'll change the usage

CarmenPopoviciu: ✅

emily-shen (approved):

RamIdeas (approved):
I assume this bug does not exist with `--x-dev-env` because config+args are re-resolved correctly each time config is updated?

CarmenPopoviciu:
> I assume this bug does not exist with `--x-dev-env` because config+args are re-resolved correctly each time config is updated?

indeed. I didn't have to fix anything for `-x-dev-env`. Can't wait for that to become the default

GregBrimble:
In my testing, I see the following behaviors work:

- [x] Static to both
- [x] Both to Worker
- [ ] Worker to both
- [ ] Worker to static
- [x] Static to Worker

- [x] default entrypoint RPC with just a Worker
- [x] default entrypoint RPC fetch with just a Worker
- [x] named entrypoint RPC with just a Worker
- [x] named entrypoint RPC fetch with just a Worker

- [x] default entrypoint RPC with both
- [ ] default entrypoint RPC fetch asset with both
- [x] default entrypoint RPC fetch Worker with both
- [x] named entrypoint RPC with both
- [ ] named entrypoint RPC fetch asset with both
- [x] named entrypoint RPC fetch Worker with both

CarmenPopoviciu:
oh boy...this is def more complex than what I thought the expectation was here. My understanding was that this should support switching between assets only and W + A. I'm surprised tbh it works for anything other than. If. we want to support such wide range of scenarios, I'll need to dbl check with the team to get a feel of whether this is smth we do want to support.

GregBrimble:
We don't need RPC for launch, (other than just ensuring we don't break "RPC to a regular Worker"). But getting Worker to both and Worker to static would be nice if possible.

GregBrimble:
I also think this PR improves the current state. It fixes going from "Static to both" and "Static to Worker", so feel free to merge this as-is and we can address "Worker to both" and "Worker to static" in a follow-up ticket.

CarmenPopoviciu:
kk. I'll keep that in mind, but will bring this to the team anyway just so we have a clear picture of the general opinion here

petebacondarwin (commented):
I haven't thought this through but it feels like we could just say that any change to wrangler.toml causes a complete restart of the dev server and that would naturally pick up the change to asset config and worker main config? Avoiding any need to special case this?

CarmenPopoviciu:
technically yes, and I **think** that `--x-dev-env` gives us the behaviour we're looking for, tho I'd need to do more proper testing to confirm. However, the default `wrangler dev`, which uses the `Dev` React component, needs to re-render that component on every `wrangler.toml` change, which in turn will trigger the dev server to restart. The `main` entrypoint is passed in a s a prop to `<Dev>`, so unless we re-compute it on a config file change, any `main` changes won't be reflected even with a dev server restart. So, unless I'm completely misunderstanding the code, we'd need to call `getEntry()` on every toml change to ensure `main` entry point changes are reflected in what dev server serves

packages/wrangler/src/dev.tsx
CarmenPopoviciu: after internal discussions, the answer is yes. `wrangler dev --x-dev-env` already supports this, no reason why `wrangler dev` shouldn't

CarmenPopoviciu:
after discussing with the team, and doing some more testing, I'm going to go head and capture your suggestions @GregBrimble in this PR. 

I've addressed your feedback in https://github.com/cloudflare/workers-sdk/pull/6775/commits/f6b33b7244ba29c47045ea7a06feab402cbe3533. 

Please have a look and let me know how that [test list](https://github.com/cloudflare/workers-sdk/pull/6775#issuecomment-2361512344) is looking. I've tested a bunch of permutations, except RPC.

With these changes, we are bringing the `wrangler dev` and `wrangler dev --x-dev-env` experience on par, AFAICT at least 🚀

GregBrimble (approved):
Updated my test sheet, and sure enough, everything works great! Thanks @CarmenPopoviciu !

edmundhung (approved):

PR #6778, opened 2024-09-19
fix(wrangler): Error if Workers + Assets are run in remote mode

CarmenPopoviciu:

Workers + Assets are currently supported only in local mode. We should throw an error if users attempt to use Workers with assets in remote mode.

Screenshot 2024-09-19 at 20 54 26

Fixes [WC-2743](https://jira.cfdata.org/browse/WC-2743)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/__tests__/dev.test.tsx
+++ b/packages/wrangler/src/__tests__/dev.test.tsx
@@ -1666,6 +1666,27 @@ describe("wrangler dev", () => {
 				`[Error: Cannot use Experimental Assets and tail consumers in the same Worker. Tail Workers are not yet supported for Workers with assets.]`
 			);
 		});
+
+		it("should error if --experimental-assets and --remote are used together", async () => {
+			fs.openSync("public", "w");
+			await expect(
+				runWrangler("dev --experimental-assets public --remote")
+			).rejects.toThrowErrorMatchingInlineSnapshot(
+				`[Error: Cannot use Experimental Assets in remote mode. Workers with assets are only supported in local mode.]`
+			);
+		});
+
+		it("should error if config.experimental_assets and --remote are used together", async () => {
+			writeWranglerToml({
+				experimental_assets: { directory: "./public" },
+			});
+			fs.openSync("public", "w");
+			await expect(
+				runWrangler("dev --remote")
+			).rejects.toThrowErrorMatchingInlineSnapshot(
+				`[Error: Cannot use Experimental Assets in remote mode. Workers with assets are only supported in local mode.]`
+			);
+		});
 	});
 
 	describe("--inspect", () => {
--- a/packages/wrangler/src/__tests__/dev.test.tsx
+++ b/packages/wrangler/src/__tests__/dev.test.tsx
@@ -345,7 +345,7 @@ describe("wrangler dev", () => {
 					directory: "assets",
 				},
 			});
-			fs.openSync("assets", "w");
+			fs.mkdirSync("assets");
 			await expect(runWrangler(`dev`)).rejects
 				.toThrowErrorMatchingInlineSnapshot(`
 				[Error: Invalid Routes:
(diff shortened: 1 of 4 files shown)


GregBrimble (approved):
Looks good to me! We'll just need to remember to rename "Experimental Assets" to "Assets" when we flip that over.

packages/wrangler/src/experimental-assets.ts
```diff
@@ -411,6 +411,15 @@ export function verifyMutuallyExclusiveAssetsArgsOrConfig(
 			"Cannot use Experimental Assets and tail consumers in the same Worker. Tail Workers are not yet supported for Workers with assets."
 		);
 	}
+
+	if (
+		(args.experimentalAssets || config.experimental_assets) &&
+		(args as StartDevOptions).remote
+	) {
+		throw new UserError(
+			"Cannot use Experimental Assets in remote mode. Workers with assets are only supported in local mode."
```
tanushree-sharma, line 420:
```suggestion
			"Cannot use Experimental Assets in remote mode. Workers with assets are only supported in local mode. Please use `wrangler dev`."
```

tanushree-sharma: Making it actionable!

CarmenPopoviciu: brill!

packages/wrangler/src/__tests__/dev.test.tsx
```diff
@@ -1666,6 +1666,27 @@ describe("wrangler dev", () => {
 				`[Error: Cannot use Experimental Assets and tail consumers in the same Worker. Tail Workers are not yet supported for Workers with assets.]`
 			);
 		});
+
+		it("should error if --experimental-assets and --remote are used together", async () => {
+			fs.openSync("public", "w");
```
petebacondarwin, line 1671: Out of curiosity why use this rather than `mkdir()`?

packages/wrangler/src/experimental-assets.ts
```diff
@@ -412,6 +412,15 @@ export function verifyMutuallyExclusiveAssetsArgsOrConfig(
 			"Cannot use Experimental Assets and tail consumers in the same Worker. Tail Workers are not yet supported for Workers with assets."
 		);
 	}
+
+	if (
+		(args.experimentalAssets || config.experimental_assets) &&
+		(args as StartDevOptions).remote
```
petebacondarwin, line 418: Rather than casting here we should ensure that this value gets passed in by changing the signature of `verifyMutuallyExclusiveAssetsArgsOrConfig()`.

petebacondarwin (changes requested):

packages/wrangler/src/experimental-assets.ts
CarmenPopoviciu: I'll move this check out of this fn instead. This is something we only need to check in dev, whereas `verifyMutuallyExclusiveAssetsArgsOrConfig` is called in multiple places

packages/wrangler/src/__tests__/dev.test.tsx
CarmenPopoviciu: no specific reason. that's what existing tests used, so I went along with that. I'll go ahead and switch

petebacondarwin: Cool. I wasn't actually asking for it to change, I was just interested in why this approach, since I had not seen it before.

packages/wrangler/src/experimental-assets.ts
petebacondarwin: Nice!

petebacondarwin (approved):

packages/wrangler/src/__tests__/dev.test.tsx
CarmenPopoviciu: oh yeah, abs! very valid question. always love a good "pushback" ;)

PR #6801, opened 2024-09-23
Retry deployments

RamIdeas:

This PR adds retries within `wrangler deploy` and `wrangler versions upload` to workaround spotty network connections and service flakes

We need to be careful about retrying the whole multi-step process because of some non-atomic steps.

Users have reported flakes at the upload step so we can safely add retries for that step to address those specific failures.

Fixes #0000

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: transparent change

### DIFF
--- a/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
+++ b/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
@@ -37,11 +37,16 @@ describe("versions upload", () => {
 			)
 		);
 	}
-	function mockUploadVersion(has_preview: boolean) {
+	function mockUploadVersion(has_preview: boolean, flakeCount = 1) {
 		msw.use(
 			http.post(
 				`*/accounts/:accountId/workers/scripts/:scriptName/versions`,
 				({ params }) => {
+					if (flakeCount > 0) {
+						flakeCount--;
+						// return HttpResponse.error();
+					}
+
 					expect(params.scriptName).toEqual("test-worker");
 
 					return HttpResponse.json(
@@ -53,8 +58,7 @@ describe("versions upload", () => {
 							},
 						})
 					);
-				},
-				{ once: true }
+				}
 			)
 		);
 	}
--- a/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
+++ b/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
@@ -44,7 +44,7 @@ describe("versions upload", () => {
 				({ params }) => {
 					if (flakeCount > 0) {
 						flakeCount--;
-						// return HttpResponse.error();
+						return HttpResponse.error();
 					}
 
 					expect(params.scriptName).toEqual("test-worker");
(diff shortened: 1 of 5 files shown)


jahands:
Thank you for this!! The flakiness in deploying made my deploy CI job so unreliable (to the point that I added a retry wrapper script around `wrangler deploy`.) This will help a lot <3

andyjessop (approved):
Very nice 👍 

One question I have is do we know this is going to fix the problem, Or will it just make failed runs take longer?

packages/wrangler/src/__tests__/versions/versions.upload.test.ts
```diff
@@ -37,11 +37,16 @@ describe("versions upload", () => {
...
 	}
-	function mockUploadVersion(has_preview: boolean) {
+	function mockUploadVersion(has_preview: boolean, flakeCount = 1) {
 		msw.use(
 			http.post(
 				`*/accounts/:accountId/workers/scripts/:scriptName/versions`,
 				({ params }) => {
+					if (flakeCount > 0) {
+						flakeCount--;
+						// return HttpResponse.error();
```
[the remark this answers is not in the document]
RamIdeas: argh how? what??

Don't know how that got left uncommented. Fixed now.

jahands:
> Very nice 👍
> 
> One question I have is do we know this is going to fix the problem, Or will it just make failed runs take longer?

I added up to 3 retries to my deployments months ago and have never had a flake since. Before that, it seemed like it had a 30% chance of failure due to 5XX errors.

RamIdeas:
> Very nice 👍
> 
> One question I have is do we know this is going to fix the problem, Or will it just make failed runs take longer?

Failures due to flakey internet or backend will be fixed by this. This will improve the happy path for successful deployments. If there is a legit reason for failure, we can iterate on this solution to bail out of the retries – but I think for now we should optimise for the happy path here and take the hit of a few seconds on the sad path.

packages/wrangler/src/utils/retry.ts
```diff
@@ -0,0 +1,18 @@
+import { setTimeout } from "node:timers/promises";
+
+export async function retryOnError<T>(
+	action: () => T | Promise<T>,
+	backoff = 2_000,
```
threepointone, line 5: this signature means you have to pass a `backoff` if you want to define `attempts`, which isn't ideal. Can we put backoff and attempts into an `options` object instead, with both fields being optional?

packages/wrangler/src/utils/retry.ts
```diff
@@ -0,0 +1,18 @@
...
+	attempts = 3
+): Promise<T> {
+	try {
+		return await action();
+	} catch (err) {
+		if (attempts <= 1) {
+			throw err;
+		}
+
+		await setTimeout(backoff);
```
threepointone, line 15: should the default be 0? what would a 2 second wait (by default) accomplish?

PR #6484, opened 2024-08-13
feat(create-cloudflare): telemetry collection

edmundhung:
## What this PR solves / how to test

Fixes [DEVX-1390](https://jira.cfdata.org/browse/DEVX-1390)
Approved as part of [CR-995529](https://jira.cfdata.org/browse/CR-995529)

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: no coverage
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [x] Changeset included
  - [ ] Changeset not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/16718
  - [ ] Documentation not necessary because:

PR #6807, opened 2024-09-23
chore(create-cloudflare): provide SPARROW_SOURCE_KEY on pre-release and publish build 

edmundhung:

Fixes #6484.

We merged #6484 today and thought that the current build setup should be sufficient as both `wrangler` and `create-cloudflare` rely on the same env `SPARROW_SOURCE_KEY`. Unfortunately, wrangler does an additional re-build before publish and the current setup inject the env only on the publish step instead of during the build.

This is a quick fix to enable telemetry in c3. We should removes the need to set the env on both steps in the future.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no feature change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no feature change
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: Already covered in #6484 which is merged today
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: No feature change

### DIFF
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -34,7 +34,8 @@ jobs:
         env:
           CI_OS: ${{ runner.os }}
           # This is the "production" key for sparrow analytics.
-          # This is needed for create-cloudflare
+          # This is needed for create-cloudflare only as wrangler will be rebuilt in the next step
+          # FIXME: We should removes the need to setup the env in multiple steps
           SPARROW_SOURCE_KEY: "50598e014ed44c739ec8074fdc16057c"
 
       - name: Create Version PR or Publish to NPM
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -33,6 +33,9 @@ jobs:
         run: pnpm run build
         env:
           CI_OS: ${{ runner.os }}
+          # This is the "production" key for sparrow analytics.
+          # This is needed for create-cloudflare
+          SPARROW_SOURCE_KEY: "50598e014ed44c739ec8074fdc16057c"
 
       - name: Create Version PR or Publish to NPM
         id: changesets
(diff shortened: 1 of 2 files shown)


.github/workflows/changesets.yml
```diff
@@ -33,6 +33,9 @@ jobs:
         run: pnpm run build
         env:
           CI_OS: ${{ runner.os }}
+          # This is the "production" key for sparrow analytics.
+          # This is needed for create-cloudflare
+          SPARROW_SOURCE_KEY: "50598e014ed44c739ec8074fdc16057c"
```
edmundhung, line 38: Wrangler will rebuild on publish here: https://github.com/cloudflare/workers-sdk/pull/6807/files#diff-877808f9ecc67bb72efb961c23f50d256e613556479d4a5748d6dbf0f61e89fdR40-R56

.github/workflows/prereleases.yml
```diff
@@ -40,6 +40,9 @@ jobs:
         env:
           NODE_ENV: "production"
           CI_OS: ${{ runner.os }}
+          # this is the "test/staging" key for sparrow analytics
+          # This is needed for create-cloudflare
+          SPARROW_SOURCE_KEY: "5adf183f94b3436ba78d67f506965998"
```
edmundhung, line 45: Wrangler will re-build on publish later: https://github.com/cloudflare/workers-sdk/pull/6807/files#diff-d222a98ebbb50d0533a810ea1520cd202df58f31ee3edf09d0e9f407f3ccd08bR52-R60

CarmenPopoviciu (approved):

.github/workflows/prereleases.yml
CarmenPopoviciu: just for my understanding...why is one workflow using prod and the other staging?

edmundhung:
I make a wrong assumption that only `wrangler` is re-built on publish, but it's not as both `create-cloudflare` and `wrangler` will rebuild.

PR #6832, opened 2024-09-26
chore: refactor flag parsing

vicb:

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/miniflare/src/plugins/core/node-compat.ts
+++ b/packages/miniflare/src/plugins/core/node-compat.ts
@@ -33,19 +33,12 @@ export function getNodeCompat(
 		hasNodejsAlsFlag,
 		hasNodejsCompatFlag,
 		hasNodejsCompatV2Flag,
-		hasNoNodejsCompatV2Flag,
 		hasExperimentalNodejsCompatV2Flag,
-	} = parseNodeCompatibilityFlags(compatibilityFlags);
+	} = computeWorkerdCompatibilityFlags(compatibilityFlags, compatibilityDate);
 
-	const nodeCompatSwitchOverDate = "2024-09-23";
 	const legacy = nodeCompat === true;
 	let mode: NodeJSCompatMode = null;
-	if (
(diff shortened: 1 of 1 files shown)


vicb:
@petebacondarwin I have been experimenting with this idea because I think parsing the flags in the same way workerd does could help.

The validation will not work after this because, i.e. because

```
	if (hasNodejsCompatFlag && hasNodejsCompatV2Flag) {
		throw new UserError(
			"The `nodejs_compat` and `nodejs_compat_v2` compatibility flags cannot be used in together. Please select just one."
		);
	}
```

and `hasNodejsCompatV2Flag` could be implied

But 

```
	if (noBundle && hasNodejsCompatV2Flag) {
		logger.warn(
			"`nodejs_compat_v2` compatibility flag and `--no-bundle` can't be used together. If you want to polyfill Node.js built-ins and disable Wrangler's bundling, please polyfill as part of your own bundling process."
		);
	}
```

is wrong because `hasNodejsCompatFalg`  should also error at least after 09-23

Probably all of the logic for mode, parsing flags and validating should live in a single file?

petebacondarwin:
I feel this is over thinking the problem. The only reason we are doing any parsing of flags here is to decide whether to run build-time polyfilling (and which set of polyfills to run). If we didn't do that then we would not need to parse these flags at all.

The validation is only there to help the user avoid any footguns of colliding certain build-time polyfilling that is not really compatible with certain runtime modes.

CarmenPopoviciu:
hey @vicb this PR is almost one year old and quite stale. I'll go ahead and close it, but pls feel free to re-open if you feel this is still an issue you'd like to solve 🙏 ♥️

PR #6834, opened 2024-09-26
use --prod flag in e2e pnpm deploy step

RamIdeas:

The "Deploy a local version of Wrangler" step of the e2e workflow has been flaking on Windows. eg https://github.com/cloudflare/workers-sdk/actions/runs/11051135464/job/30701705503?pr=6819#step:5:29

All it does it is run `pnpm deploy` so I'm not sure why.

It seems to be failing to create files in the node_modules/.../bin directories so it may be something to do with executable files (permissions/chmod issue)

The intention is to create a wrangler package for the e2e tests which doesn't need the dev dependencies so we can and should use the [`–prod` flag](https://pnpm.io/cli/deploy#:~:text=deployed%20project%20name%3E-,%2D%2Dprod%20deploy,-%3Ctarget%20directory%3E) for pnpm deploy which may avoid having any (or as many) executable files

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: workflow change
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: workflow change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: workflow change

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -45,7 +45,7 @@ jobs:
 
       - name: Deploy a local version of Wrangler
         run: |
-          pnpm --filter wrangler deploy ${{ github.workspace}}/temp/wrangler
+          pnpm --filter wrangler --prod deploy ${{ github.workspace}}/temp/wrangler
           rm ${{ github.workspace}}/temp/wrangler/templates/tsconfig.json
         env:
           NODE_ENV: "production"


petebacondarwin (approved):
Worth giving it a try.

PR #6835, opened 2024-09-26
add workers assets to vitest integration

emily-shen:

Fixes [WC-2771](https://jira.cfdata.org/browse/WC-2771)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/17273


packages/miniflare/src/plugins/core/index.ts
```diff
@@ -231,16 +235,20 @@ function getCustomServiceDesignator(
 	workerIndex: number,
 	kind: CustomServiceKind,
 	name: string,
-	service: z.infer<typeof ServiceDesignatorSchema>
+	service: z.infer<typeof ServiceDesignatorSchema>,
+	// override with router worker
+	hasAssetsAndIsVitest: boolean = false
```
penalosa, line 240: Why the `AndIsVitest`? I think we probably want to override even if not in vitest and has assets?

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -231,16 +235,20 @@ function getCustomServiceDesignator(
...
 	let entrypoint: string | undefined;
 	if (typeof service === "function") {
 		// Custom `fetch` function
 		serviceName = getCustomServiceName(workerIndex, kind, name);
 	} else if (typeof service === "object") {
+		// Worker with entrypoint
 		if ("name" in service) {
 			if (service.name === kCurrentWorker) {
+				// self binding e.g. for Vitest
 				serviceName = getUserServiceName(refererName);
```
penalosa, line 252: This case should also cover the router worker switch

packages/miniflare/src/plugins/core/index.ts
emily-shen: This option only gets used if it's in vitest. I could also set it if not using vitest and has assets, but then its just set and doesn't get used, and I don't know if that's any better.

Ideally I would not have this in core plugin options at all and just check if the asset plugin options are set, but I don't think I can refer to the assets plugin options from another plugin. If I'm mistaken though that would make me very happy!

packages/miniflare/src/plugins/core/index.ts
emily-shen: Might've misunderstood something, but this branch is for RPC entrypoints right, which _shouldn't_ get assets...? To reflect prod behaviour.

penalosa: Oh fair point, I had thought that you could define things as `{ name: "service-name" }` with entrypoint being optional—didn't realise this was just for RPC

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -700,9 +700,8 @@ function buildAssetOptions(config: Omit<ConfigBundle, "rules">) {
 	if (config.assets) {
 		return {
 			assets: {
-				workerName: config.name,
```
RamIdeas, line 703: Does setting workerName here not remove the need for https://github.com/cloudflare/workers-sdk/pull/6835/files#diff-9ceb732889eaca4ea99209c2d4988f82a6ed60a853000e856d47fdcb687d67f2R1134-R1139

emily-shen: We would also have to set this inside the vitest integration, which is a massive pain due to how the worker name is sometimes constructed with the test name as well. I think its easier to just set this entirely inside miniflare. Ideally I wouldn't expose this as an option for the asset plugin at all, I just can't figure out a nice way to pass something into a plugin without doing that.

RamIdeas (approved):

packages/miniflare/src/plugins/core/index.ts
petebacondarwin: I think what you mean is that a `fetch()` handler on a `WorkerEntryPoint` type entry point (whether it is the default or a named entry-point) will check assets first (i.e. go via the Router Worker).

But actual RPC invocations (not to `fetch()`) such as `await env.WORKER.foo()` will not check assets first and will go directly to the User Worker.

Right?

petebacondarwin: Also other non-`fetch()` handlers such as `scheduled` and `queue` will not go via Router Worker

emily-shen: Yes sorry that was imprecise - I meant fetch on a worker entrypoint should return assets. Although currently SELF can only hit the default entrypoint (with fetch or RPC) right? So you still couldn't replicate prod behaviour for assets + fetch on a named entrypoints, since we decided importing the worker directly won't return assets.

petebacondarwin: Can't you have service bindings to Worker + Assets in Vitest tests?

petebacondarwin: Right. Sounds right.

PR #6872, opened 2024-10-01
fix: render a helpful build error if a Service Worker mode Worker has imports

### LINKED ISSUE #6648: 🐛 BUG: Dynamic require of "node:stream" is not supported

Wrangler

3.75.0 [Wrangler]

20.14.0

Mac Sonoma 14.6.1

### Observed behavior
When try to run `wrangler deploy` or `wrangler dev` I get

✘ [ERROR] A request to the Cloudflare API (/accounts/44f10c0dd2f1a5b798f0c93766d4dc4c/workers/scripts/api) failed.

  Uncaught Error: Dynamic require of "node:stream" is not supported

### Expected behavior
It should deploy or run the code

### Steps to reproduce
Please provide the following:
- A minimal working subset of your `wrangler.toml`
```
name = "api"
main = "src/index.ts"
compatibility_date = "2024-09-09"
compatibility_flags = ["nodejs_compat_v2"]
```

The error seems to step from `unenv` package which is used as part of `wrangler` which nowhere in the code we use `unenv` or `streams`

_No response_

Uncaught Error: Dynamic require of "node:stream" is not supported
    at null.<anonymous> (worker.js:15:11)
    at null.<anonymous>
  (file:///Users/k/api/node_modules/unenv/runtime/node/console/index.mjs:1:26)
  in node_modules/unenv/runtime/node/console/index.mjs
    at null.<anonymous> (worker.js:18:61) in __init
    at null.<anonymous>
  (file:///Users/k/api/node_modules/unenv/runtime/node/console/$cloudflare.mjs:10:1)
  in node_modules/unenv/runtime/node/console/$cloudflare.mjs
    at null.<anonymous> (worker.js:18:61) in __init
    at null.<anonymous>
  (file:///Users/k/api/node_modules/wrangler/_virtual_unenv_global_polyfill-console.js:1:1)
  in node_modules/wrangler/_virtual_unenv_global_polyfill-console.js
    at null.<anonymous> (worker.js:18:61) in __init
    at null.<anonymous>
  (file:///Users/k/api/node_modules/unenv/runtime/polyfill/performance.mjs:1:1)
  in node_modules/unenv/runtime/polyfill/performance.mjs
    at init_performance (worker.js:18:61) in __init
    at null.<anonymous>
  (file:///Users/k/api/node_modules/wrangler/_virtual_unenv_global_polyfill-performance.js:1:1)
  in node_modules/wrangler/_virtual_unenv_global_polyfill-performance.js
   [code: 10021]

petebacondarwin:
A common mistake is to forget to export from the entry-point of a Worker, which causes
Wrangler to infer that we are in "Service Worker" mode.

In this mode, imports to external modules are not allowed.
Currently this only fails at runtime, because our esbuild step converts these imports to an internal `__require()` call that throws an error.
The error message is misleading and does not help the user identify the cause of the problem.
This is particularly tricky where the external imports are added by a library or our own node.js polyfills.

Fixes #6648

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/deployment-bundle/esbuild-plugins/cloudflare-internal.ts
+++ b/packages/wrangler/src/deployment-bundle/esbuild-plugins/cloudflare-internal.ts
@@ -1,3 +1,4 @@
+import { dedent } from "../../utils/dedent";
 import type { Plugin } from "esbuild";
 
 /**
@@ -6,18 +7,29 @@ import type { Plugin } from "esbuild";
 export const cloudflareInternalPlugin: Plugin = {
 	name: "Cloudflare internal imports plugin",
 	setup(pluginBuild) {
+		const paths = new Set();
+		pluginBuild.onStart(() => paths.clear());
 		pluginBuild.onResolve({ filter: /^cloudflare:.*/ }, (args) => {
-			if (pluginBuild.initialOptions.format === "iife") {
-				//  If we are bundling a "Service Worker" formatted Worker, imports of external modules,
-				//  which won't be inlined/bundled by esbuild, are invalid.
-				//
-				//  Throw an error if we identify `cloudflare:...` external imports.
-				//
+			paths.add(args.path);
+			return { external: true };
+		});
+		pluginBuild.onEnd(() => {
+			if (pluginBuild.initialOptions.format === "iife" && paths.size > 0) {
+				// If we are bundling in "Service Worker" mode,
+				// imports of external modules such as `cloudflare:...`,
+				// which won't be inlined/bundled by esbuild, are invalid.
+				const pathList = new Intl.ListFormat("en-US").format(
+					Array.from(paths.keys()).map((p) => `"${p}"`)
+				);
 				throw new Error(
-					`Unexpected import "${args.path}" which is not valid in a Service Worker format Worker. Are you missing \`export default { ... }\` from your Worker?`
+					dedent`
+						Unexpected external import of ${pathList}. Imports are not valid in a Service Worker format Worker.
+						Did you mean to create a Module Worker?
+						If so, try adding \`export default { ... }\` in your entry-point.
+						See https://developers.cloudflare.com/workers/reference/migrate-to-module-workers/.
+					`
 				);
 			}
-			return { external: true };
 		});
 	},
 };
--- a/packages/wrangler/src/deployment-bundle/esbuild-plugins/cloudflare-internal.ts
+++ b/packages/wrangler/src/deployment-bundle/esbuild-plugins/cloudflare-internal.ts
@@ -6,7 +6,17 @@ import type { Plugin } from "esbuild";
 export const cloudflareInternalPlugin: Plugin = {
 	name: "Cloudflare internal imports plugin",
 	setup(pluginBuild) {
-		pluginBuild.onResolve({ filter: /^cloudflare:.*/ }, () => {
+		pluginBuild.onResolve({ filter: /^cloudflare:.*/ }, (args) => {
+			if (pluginBuild.initialOptions.format === "iife") {
+				//  If we are bundling a "Service Worker" formatted Worker, imports of external modules,
+				//  which won't be inlined/bundled by esbuild, are invalid.
+				//
+				//  Throw an error if we identify `cloudflare:...` external imports.
+				//
+				throw new Error(
+					`Unexpected import "${args.path}" which is not valid in a Service Worker format Worker. Are you missing a default export from your Worker?`
+				);
+			}
 			return { external: true };
 		});
 	},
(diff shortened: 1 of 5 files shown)


threepointone:
If esm is an issue, making that cjs might also do it

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -12,22 +12,39 @@ export const nodejsHybridPlugin: () => Plugin = () => {
...
+ * If we are bundling a "Service Worker" formatted Worker, imports of external modules,
+ * which won't be inlined/bundled by esbuild, are invalid.
+ *
+ * This `onResolve()` handler will error if it identifies node.js external imports.
+ */
+function errorOnServiceWorkerFormat(build: PluginBuild) {
+	build.onResolve({ filter: NODEJS_MODULES_RE }, (args) => {
+		if (build.initialOptions.format === "iife") {
+			throw new Error(
+				`Unexpected import "${args.path}" which is not valid in a Service Worker format Worker. Are you missing a default export from your Worker?`
```
threepointone, line 35:
```suggestion
				`Unexpected import "${args.path}" which is not valid in a Service Worker format Worker. Are you missing a `export default { ... }` from your Worker?`
```

How about this instead?

petebacondarwin: Yeah I like that. I was expecting a bit of back and forth on the best messaging here.
I was wondering whether we should also say something along the line of "Did you intend to create a Server Worker format Worker?" and also whether we just warn rather than error (since for some of these scenarios the Worker could actually operate fine as long as it didn't hit the line that tries to do the dynamic require).

threepointone: So if they're importing something like node: / cloudflare:, then we're pretty sure they meant a module worker (and most new workers etc are module workers today), so I'd say we lean the messaging towards module workers.

threepointone: So also instead of "which is not valid in a Service Worker format Worker" maybe we can say "which would only work in a Modules format Worker"

packages/wrangler/src/deployment-bundle/esbuild-plugins/cloudflare-internal.ts
```diff
@@ -6,8 +7,29 @@ import type { Plugin } from "esbuild";
...
+			if (pluginBuild.initialOptions.format === "iife" && paths.size > 0) {
+				// If we are bundling in "Service Worker" mode,
+				// imports of external modules such as `cloudflare:...`,
+				// which won't be inlined/bundled by esbuild, are invalid.
+				const pathList = new Intl.ListFormat("en-US").format(
+					Array.from(paths.keys()).map((p) => `"${p}"`)
+				);
+				throw new Error(
+					dedent`
+						Unexpected external import of ${pathList}. Imports are not valid in a Service Worker format Worker.
```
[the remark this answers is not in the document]
petebacondarwin: I thought about doing that, like you did with the node.js stuff. But in this case the problem really isn't the imports it is much more likely that the user didn't mean to use Server Worker format.

If you feel strongly that we should I can add it.

packages/wrangler/src/deployment-bundle/esbuild-plugins/cloudflare-internal.ts
```diff
@@ -6,8 +7,29 @@ import type { Plugin } from "esbuild";
...
 	setup(pluginBuild) {
-		pluginBuild.onResolve({ filter: /^cloudflare:.*/ }, () => {
+		const paths = new Set();
+		pluginBuild.onStart(() => paths.clear());
+		pluginBuild.onResolve({ filter: /^cloudflare:.*/ }, (args) => {
+			paths.add(args.path);
 			return { external: true };
 		});
+		pluginBuild.onEnd(() => {
+			if (pluginBuild.initialOptions.format === "iife" && paths.size > 0) {
```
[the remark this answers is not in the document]
petebacondarwin: Isn't this the same thing? The reality is that we only get this problem when the esbuild is trying to output an iife, which just happens to be the case when we are bundling a service worker format worker.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
petebacondarwin: I updated it to make it more ES Module oriented. PTAL

penalosa (approved):

irvinebroque (commented):
🙌 👏 🎉

PR #6873, opened 2024-10-01
[wrangler] stop logging every change in assets.directory in dev

zwily:

When using `wrangler dev` with a worker with an assets dir set, every time a file changed in that dir, it was logged. This causes a lot of noise when that dir is the target of a build (like with remix). This commit just removes that logging.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: trivial
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because: 
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: small change


packages/wrangler/src/dev.tsx
```diff
@@ -866,10 +866,7 @@ export async function startDev(args: StartDevOptions) {
 							assetsWatcher = watch(assetsOptions.directory, {
 								persistent: true,
 								ignoreInitial: true,
-							}).on("all", async (eventName, changedPath) => {
-								const message = getAssetChangeMessage(eventName, changedPath);
-
-								logger.log(`🌀 ${message}...`);
```
edmundhung, line 872: I think we should log the message inside the debounce

zwily: Log something like "x files in assets directory changed"?

zwily:
This change doesn't catch the logging in `--x-dev-env`. Notice that I'm capturing the # of changes now to log, but I don't know how to do that in https://github.com/cloudflare/workers-sdk/blob/3fa846ec205a1f4e91bc1f69640dfd6e0a7b6a77/packages/wrangler/src/api/startDevWorker/ConfigController.ts#L381 because there's no debounce there - (the debouncing is handled elsewhere). Can we just eliminate the logging completely?

emily-shen (changes requested):
Can you continue logging each asset change message but use `logger.debug` instead?  This would reduce noise by default but still have this info accessible if necessary. Thanks so much for all the PRs and issues btw, really appreciate it!

zwily:
> Can you continue logging each asset change message but use `logger.debug` instead? This would reduce noise by default but still have this info accessible if necessary. Thanks so much for all the PRs and issues btw, really appreciate it!

That's a much better idea. I made that change, and also changed the logging in x-dev-env

RamIdeas:
I also did the same in https://github.com/cloudflare/workers-sdk/pull/6881

My rationale was we don't print the source files changes when we reload the dev server so there's no need to do so for assets either

PR #6879, opened 2024-10-02
fix: the docs command should not crash if given search terms

petebacondarwin:

Fixes a regression accidentally introduced by #3735.

Fixes #0000

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e tests check this code
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/cool-wombats-marry.md
+++ b/.changeset/cool-wombats-marry.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: the docs command should not crash if given search terms
+
+Fixes a regression accidentally introduced by #3735.
--- a/packages/wrangler/src/__tests__/docs.test.ts
+++ b/packages/wrangler/src/__tests__/docs.test.ts
@@ -25,12 +25,10 @@ describe("wrangler docs", () => {
 			http.post<Record<string, never>, { params: string | undefined }>(
 				`*/1/indexes/developers-cloudflare2/query`,
 				async ({ request }) => {
-					const json = await request.json();
-					const params = new URLSearchParams(json.params);
 					return HttpResponse.json({
 						hits: [
 							{
-								url: "FAKE_DOCS_URL?" + params.get("query"),
+								url: `FAKE_DOCS_URL:${await request.text()}`,
 							},
 						],
 					});
@@ -97,11 +95,13 @@ describe("wrangler docs", () => {
 			  "debug": "",
 			  "err": "",
 			  "info": "",
-			  "out": "Opening a link in your default browser: FAKE_DOCS_URL?dev",
+			  "out": "Opening a link in your default browser: FAKE_DOCS_URL:{\\"params\\":\\"query=dev&hitsPerPage=1&getRankingInfo=0\\"}",
 			  "warn": "",
 			}
 		`);
-		expect(openInBrowser).toHaveBeenCalledWith("FAKE_DOCS_URL?dev");
+		expect(openInBrowser).toHaveBeenCalledWith(
+			'FAKE_DOCS_URL:{"params":"query=dev&hitsPerPage=1&getRankingInfo=0"}'
+		);
 	});
 
 	test("opens a browser to Cloudflare docs when given multiple search terms", async ({
@@ -113,10 +113,12 @@ describe("wrangler docs", () => {
 			  "debug": "",
 			  "err": "",
 			  "info": "",
-			  "out": "Opening a link in your default browser: FAKE_DOCS_URL?foo bar",
+			  "out": "Opening a link in your default browser: FAKE_DOCS_URL:{\\"params\\":\\"query=foo+bar&hitsPerPage=1&getRankingInfo=0\\"}",
 			  "warn": "",
 			}
 		`);
-		expect(openInBrowser).toHaveBeenCalledWith("FAKE_DOCS_URL?foo bar");
+		expect(openInBrowser).toHaveBeenCalledWith(
+			'FAKE_DOCS_URL:{"params":"query=foo+bar&hitsPerPage=1&getRankingInfo=0"}'
+		);
 	});
 });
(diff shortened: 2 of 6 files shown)


.changeset/cool-wombats-marry.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: the docs command should not crash if given search terms
```
vicb, line 5:
```suggestion
fix: the docs command should not error if given search terms
```

packages/wrangler/src/__tests__/docs.test.ts
```diff
@@ -0,0 +1,122 @@
...
+		expect(std).toMatchInlineSnapshot(`
+			Object {
+			  "debug": "",
+			  "err": "",
+			  "info": "",
+			  "out": "Opening a link in your default browser: FAKE_DOCS_URL?foo bar",
+			  "warn": "",
+			}
+		`);
+		expect(openInBrowser).toHaveBeenCalledWith("FAKE_DOCS_URL?foo bar");
```
vicb, line 120: Is `?foo bar` valid?

packages/wrangler/src/__tests__/docs.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+		await expect(result).resolves.toBeUndefined();
+		expect(std.out).toMatchInlineSnapshot(`
+			"wrangler docs [command..]
+
+			📚 Open Wrangler's command documentation in your browser
+
+
+			POSITIONALS
+			  command  Enter the wrangler command you want to know more about  [array] [default: []]
```
vicb, line 59: Does it makes sense to pass an array?

vicb (approved):
LGTM.
I am wondering what the use case is to pass an array of commands?

.changeset/cool-wombats-marry.md
petebacondarwin: Why this change? It is true that wrangler crashes and fails to do what you want it to do.
The error message you see is certainly not one we would want to show to users.
Are you concerned that crash has some other connotation?

packages/wrangler/src/__tests__/docs.test.ts
petebacondarwin: This string is not actually what is passed to `openInBrowser()` the actual URL will be the first result returned from the Algolia search. This string is just the fake response I cobbled together in the MSW fetch handler so that we could see what is being passed to Algolia.

packages/wrangler/src/__tests__/docs.test.ts
petebacondarwin: The change that was made last year to convert this to a "I'm feeling lucky" search means that you can pass in multiple items. The fact it says array here, is really just an artifact of how yargs trees variadic positional args.

What we could consider doing is changing the docs to reflect the fact that this is no longer just looking for a `command` but search terms.

petebacondarwin:
Oops I forgot to fix up the wrangler help output tests...

.changeset/cool-wombats-marry.md
vicb: You're right. I wasn't I should have added right added it :)

packages/wrangler/src/__tests__/docs.test.ts
vicb: > What we could consider doing is changing the docs to reflect the fact that this is no longer just looking for a command but search terms.

What about doing that in this PR while we are fixing the command?

packages/wrangler/src/__tests__/docs.test.ts
```diff
@@ -0,0 +1,122 @@
...
+		msw.use(
+			http.post<Record<string, never>, { params: string | undefined }>(
+				`*/1/indexes/developers-cloudflare2/query`,
+				async ({ request }) => {
+					const json = await request.json();
+					const params = new URLSearchParams(json.params);
+					return HttpResponse.json({
+						hits: [
+							{
+								url: "FAKE_DOCS_URL?" + params.get("query"),
```
vicb, line 33: maybe url: `FAKE_DOCS_URL?q="${params.get("query") ?? []}"` ?

packages/wrangler/src/docs/index.ts
```diff
@@ -29,12 +30,12 @@ export async function docsHandler(
...
 			? "https://developers.cloudflare.com/workers/wrangler/commands/"
 			: "";
 
-	if (args.command && args.command.length > 0) {
-		const searchTerm = args.command.join(" ");
+	if (args.search && args.search.length > 0) {
+		const searchTerm = args.search.join(" ");
 		const searchResult = await runSearch(searchTerm);
 
 		urlToOpen = searchResult ?? urlToOpen;
```
vicb, line 41: I find the following less convoluted:

```
	// if no query is provided, open the docs homepage
	// or, if a query IS provided, but we can't find anything, open the docs homepage
  let urlToOpen = "https://developers.cloudflare.com/workers/wrangler/commands/";

	if (args.search?.length > 0) {
		const searchTerm = args.search.join(" ");
		const searchResult = await runSearch(searchTerm);

    if (searchResult) {
      urlToOpen = searchResult;
    } else if (!arg.yes) {
      urlToOpen = undefined;   
    }
	} 
```

One nit (because it's not introduced in this PR) is that `runSearch` could be a more descriptive name, i.e. where are we searching (wrangler, cloudflare, .... docs) and what do we return ("url"). It also doesn't have comment

vicb (approved):
Thanks for fixing that

I added 1 suggestion and 1 nit

packages/wrangler/src/docs/index.ts
petebacondarwin: Yeah I generally agree that this code could be a bit cleaner but I didn't think it was too bad and didn't want to churn the code too much here.

RamIdeas (approved):

PR #6881, opened 2024-10-02
fix assets + custom build = infinite loop

### LINKED ISSUE #6876: 🐛 BUG: Infinite building with --x-dev-env and an app with a build

Wrangler

3.79.0 [Wrangler]

20

Man Sequoia

### Observed behavior
When I have a build process (remix, in this case), static assets on, and use `wrangler dev --x-dev-env`, wrangler gets into an infinite loop of rebuilding my app.

Clone github.com:zwily/remix-wrangler-much-restarting and run `npm i && npm run dev`. I also tested this with wrangler HEAD, which includes `--x-dev-env` on by default, and it still happens.

If I comment out `build.command` from `wrangler.toml` it doesn't happen. Should I just make sure I always build before deploy and remove that line?

https://github.com/zwily/remix-wrangler-much-restarting

_No response_

RamIdeas:

Fixes #6876

Previously, if a `build.custom` command output files into the `assets.directory`, it caused the command to run again in an infinite loop.

This was caused by the assets watcher in the ConfigController to trigger a configUpdate event – with the intention of triggering a runtime reload – which caused the BundleController to trigger a full build.

The assets watcher has now been moved to the BundleController and triggers a bundleCompleted event instead – with the intention of triggering a runtime reload where Miniflare itself reads the new assets from the filesystem.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -927,10 +927,15 @@ describe("custom builds", () => {
 		await helper.seed({
 			"public/index.html": "world",
 		});
-		await setTimeout(300);
 
-		const res2 = await fetch(url);
-		await expect(res2.text()).resolves.toBe("world");
+		const resText = await retry(
+			(text) => text === "hello\n",
+			async () => {
+				const res2 = await fetch(url);
+				return res2.text();
+			}
+		);
+		await expect(resText).toBe("world");
 	});
 });
 
--- a/packages/wrangler/e2e/dev.test.ts
+++ b/packages/wrangler/e2e/dev.test.ts
@@ -877,6 +877,41 @@ describe("custom builds", () => {
 		text = await fetchText(url);
 		expect(text).toMatchInlineSnapshot(`"Hello, World!"`);
 	});
+
+	it("does not infinite-loop custom build with assets", async () => {
+		const helper = new WranglerE2ETestHelper();
+		await helper.seed({
+			"wrangler.toml": dedent`
+                    name = "${workerName}"
+                    compatibility_date = "2023-01-01"
+                    main = "src/index.ts"
+                    build.command = "echo 'hello' > ./public/index.html"
+
+                    [assets]
+                    directory = "./public"
+            `,
+			"src/index.ts": dedent`
+                export default {
+                    async fetch(request) {
+                        return new Response("Hello, World!")
+                    }
+                }
+            `,
+			"public/other.html": "ensure ./public exists",
+		});
+		const worker = helper.runLongLived("wrangler dev");
+
+		// first build on startup
+		await worker.readUntil(/Running custom build/, 5_000);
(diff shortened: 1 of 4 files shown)


zwily:
I tried this out... I updated my test app to point to this PR build:

https://github.com/zwily/remix-wrangler-much-restarting

Now when I run `npm run dev` and then edit a file, it does not result in infinite rebuilds. However, HMR appears to be broken. Whenever I change a file it rebuilds, and wrangler reloads, but the browser never updates until I force a refresh.

RamIdeas:
> I tried this out... I updated my test app to point to this PR build:
> 
> https://github.com/zwily/remix-wrangler-much-restarting
> 
> (Note that I did have to set `assets.watch_dir` - without that, wrangler never reloaded.)
> 
> Now when I run `npm run dev` and then edit a file, it does not result in infinite rebuilds. However, HMR appears to be broken. Whenever I change a file it rebuilds, and wrangler reloads, but the browser never updates until I force a refresh.

I think you're looking for the `--live-reload` flag. We don't support HMR yet – instead we restart the local worker upon changes, but this is different from live reload where we tell the browser something changed and trigger a page reload.

packages/wrangler/e2e/dev.test.ts
```diff
@@ -877,6 +877,61 @@ describe("custom builds", () => {
...
+
+		// now check assets are still fetchable, even after updates
+
+		const res = await fetch(url);
+		await expect(res.text()).resolves.toBe("hello\n");
+
+		await helper.seed({
+			"public/index.html": "world",
+		});
+		await setTimeout(300);
```
emily-shen, line 930: would using the retry helper (used e.g. in the watch mode tests below) be more robust than setTimeout?

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -259,6 +260,25 @@ export class BundlerController extends Controller<BundlerControllerEventMap> {
 		);
 	}
 
+	#assetsWatcher?: ReturnType<typeof watch>;
+	async #ensureWatchingAssets(config: StartDevWorkerOptions) {
+		await this.#assetsWatcher?.close();
+		console.log(config.assets?.directory);
```
emily-shen, line 266: Is this console.log left in here intentionally?

emily-shen (commented):
thanks for fixing this!

packages/wrangler/e2e/dev.test.ts
penalosa: ➕ to this

penalosa (approved):

emily-shen (approved):

PR #6884, opened 2024-10-03
chore: update esbuild

### LINKED ISSUE #5222: Upgrade `esbuild` to `0.19.0`+
Following on from #5213, this issue tracks upgrading the `esbuild` version of all tools in `workers-sdk` including `wrangler` to `0.19.0` and above. [`0.19.0`](https://github.com/evanw/esbuild/blob/main/CHANGELOG.md#0190) adds support for bundling import paths containing wildcards (e.g. `await import("./data/${lang}.json")` will bundle all files matching `./data/**/*.json`, then select the correct module at runtime). Wrangler currently allows module [`rules`](https://developers.cloudflare.com/workers/wrangler/configuration/#bundling) to support these kinds of variable-based dynamic imports. Wildcard imports [cannot be disabled](https://github.com/evanw/esbuild/issues/3337), and [do not trigger `onResolve()` plugin hooks](https://github.com/evanw/esbuild/issues/3317) _(if they did, we could mark them as `external` effectively disabling this feature)_.

### Option 1: integrate wildcard imports with custom module rules

In Wrangler's module collector, we would continue to [mark modules matching module rules as `external`](https://github.com/cloudflare/workers-sdk/blob/1235d48fed9f4e348011fd62fce6458006947501/packages/wrangler/src/deployment-bundle/module-collection.ts#L259). Whilst modules imported via wildcard imports don't trigger `onResolve()` plugin hooks, they do trigger `onLoad()` hooks for each matched module. If an `onLoad()` hook was triggered, we'd know the module must've come from a wildcard import as the `onResolve()` hook marks collected modules as `external`, and `external`s don't trigger `onLoad()`. We could add an `onLoad()` hook matching module rules that re-exported the module from an external `import()`. We'd have to be very careful here not to convert dynamic imports into static imports (i.e. we don't want `await import(...)` to become multiple `import ... from "..."`s). Users rely on dynamic imports to lazy-load parts of their code and we don't want to prevent this.

Even without additional module rules, the wildcard import feature allows files `esbuild` natively understands to be bundled. For example, `await import("./data/${lang}.js")` would be bundled without any additional configuration from the user. With our current `esbuild` bundle, this would result in a runtime error as the module wouldn't be included in the Worker upload. With this in mind, the wildcard import feature is providing functionality that wasn't previously supported. However, there is a danger it bundles too many files (e.g. `node_modules` or files that would cause a build time error because of unsupported `node:*` modules). Workers have a strict size limit, and wildcard imports could easily create bundles exceeding this.

If we enabled this feature, we'd likely want to inform affected users. Currently, `esbuild` generates output that looks like this for wildcard imports:

```js
// import("./data/**/*.txt") in src/index.mjs
var globImport_data_txt = __glob({
  "./data/a.txt": () => Promise.resolve().then(() => ...),
  "./data/b.txt": () => Promise.resolve().then(() => ...),
  "./data/c.txt": () => Promise.resolve().then(() => ...)
});

// src/index.mjs
var module = await globImport_data_txt(`./data/${name}.txt`);
```

We could regexp on `// import("...") in ... \n __glob(` in the output to warn users when a wildcard import was bundled.

### Option 2: mark modules matched by custom module rules as top-level external

Instead of implementing the module collector as an `esbuild` plugin, we could run it before `esbuild`, then pass its collected modules to the top-level `external` option. Whilst wildcard imports don't trigger `onResolve()` hooks, they _do_ respect this option. For example, setting `external: ["*.wasm"]`, then including `await import("./data/*.wasm")` in source code results in the following output:

```js
// import("./data/**/*.wasm") in src/index.mjs
var globImport_data_wasm = __glob({
  "./data/a.wasm": () => import("./data/a.wasm"),
  "./data/b.wasm": () => import("./data/b.wasm")
  "./data/c.wasm": () => import("./data/c.wasm")
});

// src/index.mjs
var module = await globImport_data_wasm(`./data/${name}.wasm`);
```

If we did this, we'd need to restart `esbuild` whenever _new_ modules matching `rules` were added. This option still has a danger of bundling too many files like `Option 1`.

### Option 3: submit PR to `esbuild` to allow this feature to controlled

Ideally, a mechanism would be added to `esbuild` to disable wildcard imports. This could either be via a top-level build option or via a plugin hook. A build option would be relatively simple to implement, but having the ability to control this feature via a plugin would be even better. There are a few ways this could be supported:

1. Call `onResolve()` with the glob pattern generated from the template literal/string concatenation as `args.path`, and allow `onResolve()` to return multiple module paths in this case. If the `onResolve()` hook returned `{ external: true }`, the `import()` statement would be bundled verbatim. This would also allow wildcard imports to be supported by `esbuild` running in the browser, which relies on plugins to provide a virtual "file system".
2. Add a new `onGlobResolve()` hook with similar behaviour to the above.
3. Call `onResolve()` with each module matching the glob pattern. Bundling wildcard imports can be thought of as a three step process: first the import is converted to a glob pattern, second the pattern is evaluated and becomes `__glob({ ...: () => import(...), ... })`, third the new `__glob()` code is bundled as usual. This intermediate second stage can be seen when using the top-level `external` option as in `Option 2`. Arguably, `onResolve()` should be called on each of these intermediate `import(...)` statements, which would allow us to mark them as `external`.

### Option 4: consider switching to another bundler

If none of these options work, we could consider another more-flexible bundler like [Rollup](https://rollupjs.org/) or the newer [Rolldown](https://rolldown.rs/) when that matures. We could also consider switching to a "bundle-less" dev approach like [Vite](https://vitejs.dev/), potentially using the module-fallback service in [a similar way to the new Vitest integration](https://github.com/cloudflare/workers-sdk/blob/main/packages/vitest-pool-workers/src/pool/module-fallback.ts). This would move away from `wrangler dev` running pretty much the same code that gets deployed, which is a nice guarantee to have _(though arguably already broken by the middleware system)_.

threepointone:
- A version of this changeset could probably live in the docs

Fixes #5222 

This patch updates esbuild from 0.17.19 to 0.24.0. That's a big bump! Lots has gone into esbuild since May '23. All the details are available at https://github.com/evanw/esbuild/blob/main/CHANGELOG.md / https://github.com/evanw/esbuild/blob/main/CHANGELOG-2023.md.

- We now support all modern JavasScript/TypeScript features suported by esbuild (as of September 2024). New additions include standard decorators, auto-accessors, and the `using` syntax.

- 0.18 introduced wider support for configuration specified via `tsconfig.json` https://github.com/evanw/esbuild/issues/3019. After observing the (lack of) any actual broken code over the last year for this release, we feel comfortable releasing this without considering it a breaking change.

- 0.19.3 introduced support for import attributes

  ```js
  import stuff from './stuff.json' with { type: 'json' }
  ```

  While we don't currently expose the esbuild configuration for developers to add their own plugins to customise how modules with import attributes are bundled, we may introduce new "types" ourselves in the future.

- 0.19.0 introduced support for wildcard imports. Specifics here (https://github.com/evanw/esbuild/blob/main/CHANGELOG-2023.md#0190). tl;dr -

  - These 2 patterns will bundle all files that match the glob pattern into a single file.

    ```js
    const json1 = await import("./data/" + kind + ".json");
    ```

    ```js
    const json2 = await import(`./data/${kind}.json`);
    ```

  - This pattern will NOT bundle any matching patterns:
    ```js
    const path = "./data/" + kind + ".js";
    const json2 = await import(path);
    ```
    You can use `find_additional_modules` to bundle any additional modules that are not referenced in the code but are required by the project.

  Now, this MAY be a breaking change for some. Specifically, if you were previously using the pattern (that will now include all files matching the glob pattern in the bundle), BUT `find_additional_modules` was NOT configured to include some files, those files would now be included in the bundle. For example, consider this code:

  ```js
  // src/index.js
  export default {
  	async fetch() {
  		const name = url.pathname;
  		const value = (await import("." + name)).default;
  		return new Response(value);
  	},
  };
  ```

  Imagine if in that folder, you had these 3 files:

  ```js
  // src/one.js
  export default "one";
  ```

  ```js
  // src/two.js
  export default "two";
  ```

  ```js
  // src/hidden/secret.js
  export default "do not share this secret";
  ```

  And your `wrangler.toml` was:

  ```toml
  name = "my-worker"
  main = "src/index.js
  ```

  Before this update:

  1. A request to anything but `http://localhost:8787/` would error. For example, a request to `http://localhost:8787/one.js` would error with _No such module "one.js"._

  2. Let's configure `wrangler.toml` to include all `.js` files in the `src` folder:

  ```toml
  name = "my-worker"
  main = "src/index.js"

  find_additional_modules = true
  rules = [
    { type = "ESModule", globs = ["*.js"]}
  ]
  ```

  Now, a request to `http://localhost:8787/one.js` would return the contents of `src/one.js`, but a request to `http://localhost:8787/hidden/secret.js` would error with _No such module "hidden/secret.js"._ To include this file, you could expand the `rules` array to be:

  ```toml
  rules = [
    { type = "ESModule", globs = ["**/*.js"]}
  ]
  ```

  Then, a request to `http://localhost:8787/hidden/secret.js` will return the contents of `src/hidden/secret.js`.

  After this update:

  - Let's put the wrangler.toml back to its original configuration:

  ```toml
  name = "my-worker"
  main = "src/index.js
  ```

  - Now, a request to `http://localhost:8787/one.js` will return the contents of `src/one.js`, but a request to `http://localhost:8787/hidden/secret.js` will ALSO return the contents of `src/hidden/secret.js`. THIS MAY NOT BE WHAT YOU WANT. You can "fix" this in 2 ways:

    1. Remove the inline wildcard import:

    ```js
    // src/index.js
    export default {
    	async fetch() {
    		const name = new URL(request.url).pathname;
    		const moduleName = "./" + name;
    		const value = (await import(moduleName)).default;
    		return new Response(value);
    	},
    };
    ```

    Now, no extra modules are included in the bundle, and a request to `http://localhost:8787/hidden/secret.js` will throw an error. You can use the `find_additional_modules` feature to include it again.

    2. Don't use the wildcard import pattern:

    ```js
    // src/index.js
    import one from "./one.js";
    import two from "./two.js";

    export default {
    	async fetch() {
    		const name = new URL(request.url).pathname;
    		switch (name) {
    			case "/one.js":
    				return new Response(one);
    			case "/two.js":
    				return new Response(two);
    			default:
    				return new Response("Not found", { status: 404 });
    		}
    	},
    };
    ```

    Further, there may be some files that aren't modules (js/ts/wasm/text/binary/etc) that are in the folder being included (For example, a `photo.jpg` file). This pattern will now attempt to include them in the bundle, and throw an error. It will look like this:

    `[ERROR] No loader is configured for ".png" files: src/photo.jpg`

    To fix this, simply move the offending file to a different folder.

    In general, we DO NOT recommend using the wildcard import pattern. If done wrong, it can leak files into your bundle that you don't want, or make your worker slightly slower to start. If you must use it (either with a wildcard import pattern or with `find_additional_modules`) you must be diligent to check that your worker is working as expected and that you are not leaking files into your bundle that you don't want. You can configure eslint to disallow dynamic imports like this:

    ```jsonc
    // .eslintrc.js
    {
    	"rules": {
    		"no-restricted-syntax": [
    			"error",
    			{
    				"selector": "ImportExpression[argument.type!='Literal']",
    				"message": "Dynamic imports with non-literal arguments are not allowed.",
    			},
    		],
    	},
    }
    ```

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because: no functional change
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


Cherry (commented):
I absolutely LOVE this! Being stuck with an older esbuild version has bugged me in many projects.

However, the breaking changes caveats seem like a lot of possible concern, not to mention the platforms supported dropped in newer esbuild versions that wrangler still technically supports today with its node.js 16 minimum version req. 

Is this the time to just rip off the band-aid and do more frequent majors, making this part of a v4? 👀

There's a quote from one of the recent esbuild releases that I love, and I hope wrangler can follow too one day:

> This release doesn't contain any deliberately-breaking changes. However, it contains a very complex new feature and while all of esbuild's tests pass, I would not be surprised if an important edge case turns out to be broken. So I'm releasing this as a breaking change release to avoid causing any trouble.

threepointone:
I'm not convinced this should be a major release, but we haven't taken a call on it yet. 
I'm now working to see if we can list the extra modules being included in the terminal, much like find_additional_modules would, when this feature is used.

threepointone:
@Cherry Could you try this release (with `npm install --save-dev https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/prs/6884/npm-package-wrangler-6884`) and see if your own projects are okay?

Cherry:
Sadly almost all of my non-trivial projects use custom builds for things like Sentry releases, etc. but on the few personal projects I've tried this on, it all seems to be working fine.

bawerd:
> @Cherry Could you try this release (with `npm install --save-dev https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/prs/6884/npm-package-wrangler-6884`) and see if your own projects are okay?

Hi and thanks for this! I came in here to file an issue but you were already on it. :)
 
For what it is worth: I tried this on my shopify app project using the latest [cloudflare workers for remix deploy](https://developers.cloudflare.com/workers/frameworks/framework-guides/remix/) method and it now succesfully builds!
Previously I got the `Expected ";" but found "with"` error when the deploy script executed `wrangler pages functions build --outdir build/worker`

threepointone:
awesome. I'm continuing work on this. we've figured out a way to make this opt-in, but it'll take a little effort.

.changeset/proud-otters-bow.md
```diff
@@ -0,0 +1,175 @@
...
+    Further, there may be some files that aren't modules (js/ts/wasm/text/binary/etc) that are in the folder being included (For example, a `photo.jpg` file). This pattern will now attempt to include them in the bundle, and throw an error. It will look like this:
+
+    `[ERROR] No loader is configured for ".png" files: src/photo.jpg`
+
+    To fix this, simply move the offending file to a different folder.
+
+    In general, we DO NOT recommend using the wildcard import pattern. If done wrong, it can leak files into your bundle that you don't want, or make your worker slightly slower to start. If you must use it (either with a wildcard import pattern or with `find_additional_modules`) you must be diligent to check that your worker is working as expected and that you are not leaking files into your bundle that you don't want. You can configure eslint to disallow dynamic imports like this:
+
+    ```jsonc
+    // .eslintrc.js
```
vicb, line 163: What about generating this file in the c3 templates?

Also:

We have `jsonc` type and JSONc syntax vs `.eslintrc.js`

Maybe we should use flat config?

penalosa (approved):

.changeset/proud-otters-bow.md
vicb: Adressed in https://github.com/cloudflare/workers-sdk/pull/6884/commits/290a077301e217adbdbd2f073c359b9bb31afde0

PR #6893, opened 2024-10-04
file based registry fixtures

RamIdeas:

Fixes N/A

This PR removes flaky dev registry fixtures, and shifts them to be run in Wrangler's e2e test suite to improve reliability and reduce the number of times they're run.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: shouldn't trigger a release
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal workflows change


packages/wrangler/e2e/dev.test.ts
```diff
@@ -221,107 +216,6 @@ describe.each([
 	});
 });
 
-describe.each([
```
penalosa, line 224: Moved to `dev-registry.test.ts`

packages/wrangler/e2e/helpers/command.ts
```diff
@@ -40,15 +41,13 @@ export function runCommand(
...
-			}
-			if (stderr.length) {
-				console.error(stderr);
-			}
+		if (stdout.length) {
+			console.log(`[${path.basename(cwd ?? "/unknown")}]`, stdout);
 		}
+		if (stderr.length) {
+			console.error(`[${path.basename(cwd ?? "/unknown")}]`, stderr);
+		}
```
penalosa, line 49: This means our E2E test output will be a _lot_ noisier, but I think we should keep logging enabled by default for now to help us diagnose and fix flakes

packages/wrangler/src/api/startDevWorker/BundlerController.ts
```diff
@@ -247,17 +254,22 @@ export class BundlerController extends Controller<BundlerControllerEventMap> {
...
+				if (!buildAborter.signal.aborted) {
+					this.emitErrorEvent({
+						type: "error",
+						reason: "Failed to construct initial bundle",
+						cause: castErrorCause(err),
+						source: "BundlerController",
+						data: undefined,
+					});
+				}
+			}
```
penalosa, line 272: BundleController had a race condition where old builds would sometimes overwrite newer ones. We'd thought this was handled by esbuild, but it turns out that it's not in the way we need it to. This fixes the intermittent "Failed to construct initial bundle" flakes we were seeing in E2E tests in CI

fixtures/shared/src/run-wrangler-long-lived.ts
```diff
@@ -25,13 +25,28 @@ export async function runWranglerPagesDev(
 ) {
 	if (publicPath) {
 		return runLongLivedWrangler(
-			["pages", "dev", publicPath, "--x-dev-env", "--ip=127.0.0.1", ...options],
+			[
+				"pages",
+				"dev",
+				publicPath,
+				"--x-dev-env",
+				"--x-registry",
```
andyjessop, line 33: 🎆

packages/wrangler/src/api/startDevWorker/BundlerController.ts
andyjessop: Very cool, thanks for that.

packages/wrangler/src/dev.tsx
```diff
@@ -476,6 +476,11 @@ async function updateDevEnvRegistry(
 	devEnv: DevEnv,
 	registry: WorkerRegistry | undefined
 ) {
+	// Make sure we're not patching an empty config
+	if (!devEnv.config.latestConfig) {
```
andyjessop, line 480: What's this change fixing?

andyjessop (approved):
Looks great! One question, but otherwise really happy this is going forward 🚀

packages/wrangler/src/dev.tsx
penalosa: This is just moving this check higher up the function. Previously we got bound workers and _then_ waited for a config update, but since getting bound workers depends on the config we should've been waiting here instead

RamIdeas: Aha! Does this explain the race we were seeing?

packages/wrangler/src/config/index.ts
```diff
@@ -384,12 +395,23 @@ export function printBindings(bindings: CfWorkerInit["bindings"]) {
...
+				if (entrypoint) {
+					value += `#${entrypoint}`;
 				}
 
+				const registryDefinition = registry?.[service];
+				if (
+					registryDefinition &&
+					(entrypoint
+						? registryDefinition.entrypointAddresses?.[entrypoint]
+						: true)
```
RamIdeas, line 409: nit

```suggestion
					(!entrypoint || registryDefinition.entrypointAddresses?.[entrypoint])
```

packages/wrangler/src/dev.tsx
penalosa: One of them, yes! Another was in the bundle controller, and there might still be others lurking

PR #6894, opened 2024-10-04
fix: improve the rendering of build errors when bundling

petebacondarwin:

Fixes #0000

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: doesn't affect e2e use cases
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: polish / bug fix

### DIFF
--- a/packages/wrangler/src/deployment-bundle/bundle.ts
+++ b/packages/wrangler/src/deployment-bundle/bundle.ts
@@ -468,7 +468,11 @@ export async function bundleWorker(
 			await ctx.watch();
 			result = await initialBuildResultPromise;
 			if (result.errors.length > 0) {
-				throw new UserError("Failed to build");
+				throw new BuildFailure(
+					"Initial build failed.",
+					result.errors,
+					result.warnings
+				);
 			}
 
 			stop = async function () {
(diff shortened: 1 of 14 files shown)


packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -467,7 +467,11 @@ export async function bundleWorker(
 			await ctx.watch();
 			result = await initialBuildResultPromise;
 			if (result.errors.length > 0) {
-				throw new UserError("Failed to build");
```
petebacondarwin, line 470: We were just swallowing the errors and warnings here and relying upon them being logged "directly" to stdout by another instance of esbuild that is doing the format sniffing. See below.

packages/wrangler/src/deployment-bundle/guess-worker-format.ts
```diff
@@ -38,6 +38,7 @@ export default async function guessWorkerFormat(
 		metafile: true,
 		bundle: false,
 		write: false,
+		logLevel: "silent",
```
petebacondarwin, line 41: When guessing the worker format, we use esbuild but were forgetting to silence its stdout logging.
This led to duplicate logs and also confused things where we were relying on this run of esbuild to output initial build failures.

irvinebroque:
🙌

packages/wrangler/src/deployment-bundle/guess-worker-format.ts
petebacondarwin: Landed in a previous PR.

emily-shen (approved):

PR #6915, opened 2024-10-07
Bump Miniflare v2 in Wrangler v2

penalosa:


penalosa:
For posterity—merged with some checks failing because of deprecated actions

PR #6916, opened 2024-10-07
VS-286 Support mixed-mode local bindings for Vectorize in Miniflare

garrettgu10:

Fixes VS-286.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: I don't think this limitation is mentioned anywhere in docs and it's generally expected that bindings work in local dev mode.

### DIFF
--- a/packages/wrangler/e2e/dev-with-resources.test.ts
+++ b/packages/wrangler/e2e/dev-with-resources.test.ts
@@ -513,7 +513,6 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 				index_name = "${name}"
 				`,
 			"samples.ndjson": dedent`
-				{"id":"b0daca4a-ffd8-4865-926b-e24800af2a2d","values":[0.2331,1.0125,0.6131,0.9421,0.9661,0.8121,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"metadata":{"text":"She sells seashells by the seashore"}}
 				{"id":"a44706aa-a366-48bc-8cc1-3feffd87d548","values":[0.2321,0.8121,0.6315,0.6151,0.4121,0.1512,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0],"metadata":{"text":"Peter Piper picked a peck of pickled peppers"}}
 			`,
 			"src/index.ts": dedent`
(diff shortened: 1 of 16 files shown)


cloudkite:
What’s the reasoning behind disabling write in local mode? Doesn’t seem very useful without ability to write

garrettgu10:
@cloudkite Both D1 and R2 operate on local instances that are separate from prod. Since we're directly connecting to the prod Vectorize instance instead, we definitely don't want to catch people off guard and make them cause changes in prod while testing "locally".

We're considering re-enabling write operations behind an optional flag, and I think that's probably the best option.

cloudkite:
@garrettgu10 makes sense in terms of accidentally writing to a live db instance. However having trouble imaging how a readonly "local" db will be useful for development/testing purposes?

garrettgu10:
@cloudkite we are not (yet) launching a local simulation database for vectorize. All requests are sent to the prod database in local dev mode.

cloudkite:
@garrettgu10 yup I understood that part. I just meant if I can’t write I’ll still need to use —remote

How about when wrangler is in local mode it refuses to bind unless the index name has suffix of “-local-dev” or similar, that would make it very unlikely for someone to accidentally bind to their live/prod vectorize index

packages/wrangler/src/vectorize/fetcher.ts
```diff
@@ -0,0 +1,84 @@
...
+				{ status: 403 }
+			);
+		}
+
+		const url = request.url.replace(
+			"http://vector-search/",
+			`/accounts/${accountId}/vectorize/${indexVersion}/indexes/${indexId}/`
+		);
+
+		// TODO: v1 endpoints have a different format
```
sejoker, line 57: we won't support v1 endpoints as v1 is EOL.

sejoker (approved):

irvinebroque (commented):
Let's make sure to update these docs when this lands:

https://developers.cloudflare.com/workers/testing/local-development/#supported-resource-bindings-in-different-environments

garrettgu10:
@mrbbot Thanks for your comments, they've been super helpful internally and we've discussed them at length. I agree that a local simulator is the superior solution re. DX and seems fairly simple to implement as a first-pass, but the team doesn't have the bandwidth for maintaining such a solution at the moment, esp. wrt. feature-parity with updates to the service. 

With that in mind, I think the best compromise is to allow users to bind to production indexes through an optional flag such as `--experimental-vectorize-bind-to-production`. That way, users must explicitly opt-in, minimizing possibility for confusion. Then, sessions launched without the flag will have no Vectorize bindings until we're ready to launch a local simulator. That's the approach we're going with for now.

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -598,6 +603,36 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 		};
 	}
 
+	if (bindings.vectorize) {
+		for (const vectorizeBinding of bindings.vectorize) {
+			const bindingName = vectorizeBinding.binding;
+			const indexName = vectorizeBinding.index_name;
+			const indexVersion = "v2"; // TODO: support v1?
```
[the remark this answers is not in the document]
penalosa: Is there any way to warn/error if a user tries to use a v1 binding?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -889,9 +924,8 @@ export async function buildMiniflareOptions(
 	if (config.bindings.vectorize?.length) {
 		if (!didWarnMiniflareVectorizeSupport) {
 			didWarnMiniflareVectorizeSupport = true;
-			// TODO: add local support for Vectorize bindings (https://github.com/cloudflare/workers-sdk/issues/4360)
 			logger.warn(
-				"Vectorize bindings are not currently supported in local mode. Please use --remote if you are working with them."
+				"Using Vectorize always accesses your Cloudflare account in order to run queries and may incur usage charges even in local development. Write operations are also not enabled in local development; use `--remote` if write operations are needed."
```
penalosa, line 928: Why are write operations not enabled?

penalosa:
> With that in mind, I think the best compromise is to allow users to bind to production indexes through an optional flag such as `--experimental-vectorize-bind-to-production`. That way, users must explicitly opt-in, minimizing possibility for confusion. Then, sessions launched without the flag will have no Vectorize bindings until we're ready to launch a local simulator. That's the approach we're going with for now.

@garrettgu10 This feels like a pretty big break from how other bindings are implemented, and I'd be concerned about adding an `--experimental-vectorize-bind-to-production` flag without a larger discussion of how that fits in to other bindings. This PR looks relatively reasonable as a mixed mode implementation for Vectorize (other than the write limitations, which I touched on in another comment), but I'd been assuming that there were fundamental limitations that required mixed mode for Vectorize (similar to Workers AI & Browser bindings, both of which require large downloads/specific hardware).

If there aren't any fundamental limits on getting Vectorize working locally, would it be possible to revisit that discussion? Do you have a link to a decision doc or anything around this?

packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -579,7 +640,6 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 	it.skipIf(isLocal).todo("exposes send email bindings");
 	it.skipIf(isLocal).todo("exposes browser bindings");
 	it.skipIf(isLocal).todo("exposes Workers AI bindings");
-	it.skipIf(isLocal).todo("exposes Vectorize bindings");
```
penalosa, line 582: 🎉

PR #6920, opened 2024-10-08
chore: update unenv dependency version

vicb:

Update unenv - Pulled fixes are listed in the changeset

See https://github.com/unjs/unenv/issues/314

Fixes n/a

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: unenv update
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: unenv update


anonrig (approved):

petebacondarwin:
The error in the nodejs-hybrid-app is:

```
[ERROR] service core:user:nodejs-hybrid-app: Uncaught TypeError: Cannot read properties of undefined (reading 'bind')
```

Building the worker to `dist`, then changing `wrangler.toml` to point to the compiled output and adding `no_bundle = true`gives the actual line that is failing:

```
addEventListener: performance3.addEventListener.bind(performance3),
```

So it appears that `addEventListener` is not defined

Elsewhere in the code we can see:

```
performance3 = globalThis.performance ?? new Performance2();
```

The fake `Performance2` class has `addEventListener()` but it is not used because `globalThis.performance` is defined, and is missing that function.


From the [docs](https://developers.cloudflare.com/workers/runtime-apis/performance/) `performance` is defined and only has `now()` and `timeOrigin()`. Confirmed in workerd [source code](https://github.com/cloudflare/workerd/blob/78774c81ad42b3ca98eeb8e0b4d51814ac43f7e7/src/workerd/api/global-scope.h#L91).

So this polyfill is a bit broken for Cloudflare. I wonder if we need to be more fine grained about polyfilling that global?

vicb:
Great analysis @petebacondarwin 

> So it appears that `addEventListener` is not defined
> 
> Elsewhere in the code we can see:
> 
> ```
> performance3 = globalThis.performance ?? new Performance2();
> ```
> 
> The fake `Performance2` class has `addEventListener()` but it is not used because `globalThis.performance` is defined, and is missing that function.

unenv should inject the polyfill in the globalScope [here](https://github.com/unjs/unenv/blob/a58d5160ad0bb7e949c9fd016996afa695c34f78/src/presets/cloudflare.ts#L72C34-L72C45). But it looks like this is not working.

Also https://github.com/unjs/unenv/issues/318

petebacondarwin (approved):

vicb:
Thanks Pete!

PR #6921, opened 2024-10-08
chore: Drop the "Fixes" check in CI

vicb:

I had the PR description check failed because I typed "Fixes n/a" instead of "Fixes N/A"
It should be ok to make this check case insensitive.

Fixes N/A

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because: trivial change
  - [ ] I don't know
  - [ ] Required
  - [X] Not required because: CI change
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [X] Changeset not necessary because: CI minor tweak
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: CI

### DIFF
--- a/tools/deployments/validate-pr-description.ts
+++ b/tools/deployments/validate-pr-description.ts
@@ -31,16 +31,6 @@ export function validateDescription(
 		return [];
 	}
 
-	if (
-		!/^Fixes (#\d+|N\/A|\[[A-Z]+-\d+\]\(https:\/\/jira\.cfdata\.org\/browse\/[A-Z]+-\d+\))/m.test(
-			body
-		)
-	) {
-		errors.push(
-			"Your PR description must include an issue reference in the format `Fixes #000` (for GitHub issues), `Fixes [AA-000](https://jira.cfdata.org/browse/AA-000)` (for internal Jira ticket references), or `Fixes N/A` if there's no associated issue (and it doesn't make sense to create one)"
-		);
-	}
-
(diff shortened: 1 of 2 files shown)


tools/deployments/validate-pr-description.ts
```diff
@@ -32,7 +32,7 @@ export function validateDescription(
 	}
 
 	if (
-		!/^Fixes (#\d+|N\/A|\[[A-Z]+-\d+\]\(https:\/\/jira\.cfdata\.org\/browse\/[A-Z]+-\d+\))/m.test(
+		!/^Fixes (#\d+|N\/A|\[[A-Z]+-\d+\]\(https:\/\/jira\.cfdata\.org\/browse\/[A-Z]+-\d+\))/im.test(
```
vicb, line 35: Should we also drop anchoring?

emily-shen:
While we're here, should we just remove this check entirely? I'm not sure what the benefit to having it is. 

We can leave `Fixes #etc.`  in the template to nudge people into including it, but the check is just a bit annoying 😅 

cc @penalosa

penalosa:
Yeah, that makes sense—we're ending up with a lot of Fixes N/A in PRs, which isn't all that useful. Let's remove this entirely.

vicb:
Thanks for the feedback, I'll update the PR

penalosa (approved):

PR #6922, opened 2024-10-08
Strip down wrangler.toml templates

vicb:

The `wrangler.toml` templates have 100+ lines of commented out template config.

I think it doesn't help users and potentially could make them think workers are more complex than they really are.

This is only a draft PR to get feedback.

What would be great is first to agree on the wording.

Then we should probably automatically inject the comments in all `wrangler.toml` instead of having to maintain 20 versions of those comments.

@irvinebroque @petebacondarwin thoughts?

Fixes N/A

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/create-cloudflare/templates/analog/templates/wrangler.toml
+++ b/packages/create-cloudflare/templates/analog/templates/wrangler.toml
@@ -3,83 +3,9 @@ name = "<TBD>"
 compatibility_date = "<TBD>"
 pages_build_output_dir = "./dist/analog/public"
 
-# Automatically place your workloads in an optimal location to minimize latency.
-# If you are running back-end logic in a Pages Function, running it closer to your back-end infrastructure
-# rather than the end user may result in better performance.
-# Docs: https://developers.cloudflare.com/pages/functions/smart-placement/#smart-placement
-# [placement]
-# mode = "smart"
-
-# Variable bindings. These are arbitrary, plaintext strings (similar to environment variables)
-# Note: Use secrets to store sensitive data.
-# Docs:
-# - https://developers.cloudflare.com/pages/functions/bindings/#environment-variables
-# - https://developers.cloudflare.com/pages/functions/bindings/#secrets
-# [vars]
-# MY_VARIABLE = "production_value"
-
-# Bind the Workers AI model catalog. Run machine learning models, powered by serverless GPUs, on Cloudflare’s global network
-# Docs: https://developers.cloudflare.com/pages/functions/bindings/#workers-ai
-# [ai]
-# binding = "AI"
-
-# Bind a D1 database. D1 is Cloudflare’s native serverless SQL database.
-# Docs: https://developers.cloudflare.com/pages/functions/bindings/#d1-databases
-# [[d1_databases]]
-# binding = "MY_DB"
-# database_name = "my-database"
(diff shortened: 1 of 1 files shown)


irvinebroque:
Goal of having this here is

- discover what's possible
- show how to configure fast w/examples

Can see how looks inelegant but would want to replace with something we believe genuinely more effective for onboarding and adoption of full platform, not just remove.

@nevikashah wdyt?

vicb:
Noble goal.

My own experience is that this scared me the first time I generated a project.
Every time I start a new project from C3 deleting those lines is the first thing I do - it gives me a feeling of satisfaction!

_For the records I have discussed with a few people and they were +1 to strip this_

petebacondarwin:
I am also put off by this large commented block; moreover it is a pain to maintains since every time someone adds a binding we have to go and update all these templates.

How about we just have a single comment that links to a web page in the docs that shows a fully featured example wrangler.toml wired up to lots of bindings?

vicb:
> How about we just have a single comment that links to a web page in the docs that shows a fully featured example wrangler.toml wired up to lots of bindings?

Wouldn't "a web page in the docs that shows a fully featured example wrangler.toml wired up to lots of bindings" move the exact same problem to a different place?

IIUC the initial intent was to help new comers. Can't we have an example with a few bindings (do we have stats on the most common? probably assets, KV, plus a couple others) and link to that.

For more advanced users we can link the (reference) docs and they'll be able to find what they want there.

petebacondarwin:
> Wouldn't "a web page in the docs that shows a fully featured example wrangler.toml wired up to lots of bindings" move the exact same problem to a different place?

I believe it would be better for two reasons:

1. There would only be one page to update, rather than many templates. I note you only changed the analog one in this PR but there are actually loads and loads of these wrangler.tomls
2. It would still provide the user with just one click to get access to the information we are trying to make accessible by having this big comment in the template

vicb:
> 1. There would only be one page to update, rather than many templates. I note you only changed the analog one in this PR but there are actually loads and loads of these wrangler.tomls

I mentioned that in the PR description:

> Then we should probably automatically inject the comments in all wrangler.toml instead of having to maintain 20 versions of those comments.

_(That still holds if we move to a more reasonably sized comment)_

vicb:
@irvinebroque @nevikashah Can we please move forward here?

I really think we should keep things simple and not overwhelm fist time users with hundreds of lines of commented out configuration lines - in the same spirit, I've proposed #6957.

Providing a short text and a link to the documentation would be as efficient without making workers look more complicated than they really are.

Thanks!

vicb:
As discussed with @irvinebroque I still plan to try to get this in.
I'll work on that once the migration to wrangler.json is over to avoid conflicts.

penalosa:
Converting this to a draft for now, since it won't go in until after the C3 switch to JSON in Q1

vicb:
Closing for now as partially addressed by https://github.com/cloudflare/workers-sdk/pull/7676

PR #6925, opened 2024-10-08
Skip C3 tests on *.md only changes

penalosa:

Fixes N/A

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: workflow change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: workflow change
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: workflow change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: workflow change

### DIFF
--- a/.github/workflows/c3-e2e-experimental.yml
+++ b/.github/workflows/c3-e2e-experimental.yml
@@ -5,11 +5,6 @@ on:
   pull_request:
     paths:
       - packages/create-cloudflare/**
-  push:
-    branches:
-      - main
-    paths:
-      - packages/create-cloudflare/**
 
 jobs:
   e2e:
@@ -20,7 +15,7 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}
       cancel-in-progress: true
     name: ${{ format('Run experimental tests for {0}@{1} on {2}', matrix.pm.name, matrix.pm.version, matrix.os) }}
-    if: github.repository_owner == 'cloudflare' && github.event.pull_request.user.login != 'dependabot[bot]'
+    if: github.event.pull_request.head.repo.owner.login == 'cloudflare' && github.event.pull_request.user.login != 'dependabot[bot]'
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/c3-e2e-experimental.yml
```diff
@@ -5,11 +5,6 @@ on:
   pull_request:
     paths:
       - packages/create-cloudflare/**
-  push:
-    branches:
-      - main
-    paths:
-      - packages/create-cloudflare/**
```
penalosa, line 12: We don't actually need to run C3 E2E tests on main. They'll be run on Version Packages before release, which has the same effect

.github/workflows/c3-e2e.yml
```diff
@@ -20,7 +15,7 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}
       cancel-in-progress: true
     name: ${{ format('Run tests for {0}@{1} on {2}', matrix.pm.name, matrix.pm.version, matrix.os) }}
-    if: github.repository_owner == 'cloudflare' && github.event.pull_request.user.login != 'dependabot[bot]'
+    if: github.event.pull_request.head.repo.owner.login == 'cloudflare' && github.event.pull_request.user.login != 'dependabot[bot]'
```
penalosa, line 18: Same change as https://github.com/cloudflare/workers-sdk/pull/6770

.github/workflows/c3-e2e-experimental.yml
andyjessop: Why do we want to defer the running of these tests? Won't this get us into a situation where the caretaker is having to debug and fix C3 errors because we haven't tested them correctly?

lrapoport-cf: same question as @andyjessop 👍

penalosa: This is the same setup as we have for Wrangler's E2E tests, where we run on Version Packages instead of main. This doesn't actually defer anything, since Version Packages is always in sync with main. Currently we're running C3 E2E tests _twice_ for every commit to main—once on main and once on Version Packages. This PR just removes one of those instances, which should free up some CI capacity.

emily-shen (approved):

PR #6929, opened 2024-10-08
chore: switch to file-based registry by default

andyjessop:

Fixes #000.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] Tests not necessary because: covered
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Documentation not necessary because: not user facing


penalosa:
I think it would be better to default the value of the `--x-registry` flag to true, so that users can still turn this off for the first week or so until we remove the non-file based code.

andyjessop:
@penalosa agreed. This PR was not meant to be merged, it was just for testing in CI. This was just the first place I came across that would definitely remove the server-based registry from all code paths.

PR #6932, opened 2024-10-09
fix: allow `require`ing alias modules

### LINKED ISSUE #6896: 🐛 BUG: Uncaught TypeError: utils.inherits is not a function

Workers Runtime, Wrangler

3.80.0 [Wrangler]

20.17.0

Mac Sonoma 14.6.1

### Observed behavior

I have a worker which imports `ethers@5.7.2` library. When I switched to `compatibility_flags = ["nodejs_compat_v2"]` it started throwing:
```
✘ [ERROR] service core:user:error-example: Uncaught TypeError: utils.inherits is not a function

    at null.<anonymous> (index.js:7544:11) in
  node_modules/.pnpm/hash.js@1.1.7/node_modules/hash.js/lib/hash/sha/1.js
    at null.<anonymous> (index.js:12:50) in __require
    at null.<anonymous> (index.js:8295:20) in
  node_modules/.pnpm/hash.js@1.1.7/node_modules/hash.js/lib/hash/sha.js
    at null.<anonymous> (index.js:12:50) in __require
    at null.<anonymous> (index.js:8812:17) in
  node_modules/.pnpm/hash.js@1.1.7/node_modules/hash.js/lib/hash.js
    at null.<anonymous> (index.js:12:50) in __require
    at null.<anonymous> (index.js:15900:28)
```

### Expected behavior

The worker should start.

### Steps to reproduce

1. Clone the https://github.com/nkuba/cf-workers-error-example repository
2. Run `pnpm install`
3. Run `pnpm dev`

https://github.com/nkuba/cf-workers-error-example

```
> wrangler dev

 ⛅️ wrangler 3.80.0
-------------------

[wrangler:inf] Ready on http://localhost:8787
⎔ Starting local server...
✘ [ERROR] service core:user:error-example: Uncaught TypeError: utils.inherits is not a function

    at null.<anonymous> (index.js:7544:11) in
  node_modules/.pnpm/hash.js@1.1.7/node_modules/hash.js/lib/hash/sha/1.js
    at null.<anonymous> (index.js:12:50) in __require
    at null.<anonymous> (index.js:8295:20) in
  node_modules/.pnpm/hash.js@1.1.7/node_modules/hash.js/lib/hash/sha.js
    at null.<anonymous> (index.js:12:50) in __require
    at null.<anonymous> (index.js:8812:17) in
  node_modules/.pnpm/hash.js@1.1.7/node_modules/hash.js/lib/hash.js
    at null.<anonymous> (index.js:12:50) in __require
    at null.<anonymous> (index.js:15900:28)



✘ [ERROR] The Workers runtime failed to start. There is likely additional logging output above.
```

vicb:

Fixes #6896 

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/nasty-melons-camp.md
+++ b/.changeset/nasty-melons-camp.md
@@ -1,0 +1,10 @@
+---
+"wrangler": minor
+---
+
+fix: allow `require`ing aliased npm packages
+
+Before this PR `require`ing an NPM packages aliased in unenv would fail.
+That's because `require` would load the mjs file.
+
+This PR adds wraps the mjs file in a virtual ES module to allow `require`ing it.
--- a/.changeset/nasty-melons-camp.md
+++ b/.changeset/nasty-melons-camp.md
@@ -4,7 +4,7 @@
 
 fix: allow `require`ing aliased npm packages
 
-Before this PR `require`ing an NPM packages aliased in unenv would fail.
+Before this PR `require`ing NPM packages aliased in unenv would fail.
 That's because `require` would load the mjs file.
 
 This PR adds wraps the mjs file in a virtual ES module to allow `require`ing it.
--- a/.changeset/nasty-melons-camp.md
+++ b/.changeset/nasty-melons-camp.md
@@ -2,9 +2,9 @@
 "wrangler": minor
 ---
 
-fix: allow `require`ing aliased npm packages
+fix: allow `require`ing unenv aliased packages
 
-Before this PR `require`ing NPM packages aliased in unenv would fail.
+Before this PR `require`ing packages aliased in unenv would fail.
 That's because `require` would load the mjs file.
 
 This PR adds wraps the mjs file in a virtual ES module to allow `require`ing it.
--- a/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
+++ b/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
@@ -6,7 +6,7 @@ import { getBasePath } from "../../paths";
 import type { Plugin, PluginBuild } from "esbuild";
 
 const REQUIRED_NODE_BUILT_IN_NAMESPACE = "node-built-in-modules";
-const REQUIRED_NPM_PACKAGE_NAMESPACE = "required-npm-package";
+const REQUIRED_UNENV_ALIAS_NAMESPACE = "required-unenv-alias";
 
 export const nodejsHybridPlugin: () => Plugin = () => {
 	const { alias, inject, external } = env(nodeless, cloudflare);
@@ -109,11 +109,12 @@ function handleUnenvAliasedPackages(
 		// Note: Does not apply to Node.js packages that are handled in `handleRequireCallsToNodeJSBuiltins`
 		if (
 			args.kind === "require-call" &&
-			(unresolvedAlias.startsWith("unenv/runtime/npm/") || unresolvedAlias.startsWith("unenv/runtime/mock/"))
+			(unresolvedAlias.startsWith("unenv/runtime/npm/") ||
+				unresolvedAlias.startsWith("unenv/runtime/mock/"))
 		) {
 			return {
 				path: args.path,
-				namespace: REQUIRED_NPM_PACKAGE_NAMESPACE,
+				namespace: REQUIRED_UNENV_ALIAS_NAMESPACE,
 			};
 		}
 		// Resolve the alias to its absolute path and potentially mark it as external
@@ -123,13 +124,30 @@ function handleUnenvAliasedPackages(
 		};
 	});
 
+	build.initialOptions.banner = { js: "", ...build.initialOptions.banner };
+	build.initialOptions.banner.js += dedent`
+		function __cf_cjs(esm) {
+		  const cjs = 'default' in esm ? esm.default : {};
+			for (const [k, v] of Object.entries(esm)) {
+				if (k !== 'default') {
+					Object.defineProperty(cjs, k, {
+						enumerable: true,
+						value: v,
+					});
+				}
+			}
+			return cjs;
+		}
+		`;
+
 	build.onLoad(
-		{ filter: /.*/, namespace: REQUIRED_NPM_PACKAGE_NAMESPACE },
+		{ filter: /.*/, namespace: REQUIRED_UNENV_ALIAS_NAMESPACE },
 		({ path }) => {
 			return {
 				contents: dedent`
-        import * as all from '${path}';
-        module.exports = 'default' in all ? all.default : all;`,
+        import * as esm from '${path}';
+				module.exports = __cf_cjs(esm);
+				`,
 				loader: "js",
 			};
 		}
--- a/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
+++ b/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
@@ -89,21 +89,17 @@ function handleUnenvAliasedPackages(
 	external: string[]
 ) {
 	// esbuild expects alias paths to be absolute
-	const aliasAbsolute = Object.fromEntries(
-		Object.entries(alias)
-			.map(([key, value]) => {
-				let resolvedAliasPath;
-				try {
-					resolvedAliasPath = require.resolve(value);
-				} catch (e) {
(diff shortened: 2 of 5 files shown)


.changeset/nasty-melons-camp.md
```diff
@@ -0,0 +1,10 @@
+---
+"wrangler": minor
+---
+
+fix: allow `require`ing aliased npm packages
+
+Before this PR `require`ing an NPM packages aliased in unenv would fail.
```
petebacondarwin, line 7:
```suggestion
Before this PR `require`ing NPM packages aliased in unenv would fail.
```

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -81,38 +82,57 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
+	for (const [module, unresolvedAlias] of Object.entries(alias)) {
+		try {
+			aliasAbsolute[module] = require
+				.resolve(unresolvedAlias)
+				.replace(/\.cjs$/, ".mjs");
+		} catch (e) {
+			// this is an alias for package that is not installed in the current app => ignore
+		}
+	}
+
```
petebacondarwin, line 101: This bit is just a refactoring, right?

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -81,38 +82,57 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
 	const UNENV_ALIAS_RE = new RegExp(
 		`^(${Object.keys(aliasAbsolute).join("|")})$`
 	);
 
 	build.onResolve({ filter: UNENV_ALIAS_RE }, (args) => {
+		// Convert `require()` calls for NPM packages to a virtual ES Module that can be imported avoiding the require calls.
+		// Note: Only applies to npm aliases as node package are handled in `handleRequireCallsToNodeJSBuiltins`
+		if (
+			args.kind === "require-call" &&
+			alias[args.path].startsWith("unenv/runtime/npm")
```
petebacondarwin, line 111: Do we need to cover `unenv/runtime/mock` as well?

There are files in there such as `runtime/mock/empty.cjs` and `runtime/mock/proxy.cjs`, which may also be required via an alias?

petebacondarwin: E.g. https://github.com/unjs/unenv/blob/a58d5160ad0bb7e949c9fd016996afa695c34f78/src/presets/nodeless.ts#L80

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -81,38 +82,57 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
 		};
 	});
+
+	build.onLoad(
+		{ filter: /.*/, namespace: REQUIRED_NPM_PACKAGE_NAMESPACE },
+		({ path }) => {
+			return {
+				contents: dedent`
+        import * as all from '${path}';
+        module.exports = 'default' in all ? all.default : all;`,
```
petebacondarwin, line 131: What if there is a default export and named exports?
Or is the expectation that if there is a default it will already "contain" the named exports?

petebacondarwin (changes requested):
A couple of queries...

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
vicb: Good point, I'll add `mock`!

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
vicb: Good point!
Refactor + added a banner helper function.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -81,38 +82,76 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
 		};
 	});
+
+	build.initialOptions.banner = { js: "", ...build.initialOptions.banner };
+	build.initialOptions.banner.js += dedent`
+		function __cf_cjs(esm) {
+		  const cjs = 'default' in esm ? esm.default : {};
+			for (const [k, v] of Object.entries(esm)) {
+				if (k !== 'default') {
+					Object.defineProperty(cjs, k, {
```
pi0, line 133: defining properties on `default` export is not always possible. it can be frozen (modules are) so a try catch is needed.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -81,38 +82,76 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
 			path: aliasAbsolute[args.path],
-			external: external.includes(alias[args.path]),
+			external: external.includes(unresolvedAlias),
 		};
 	});
+
+	build.initialOptions.banner = { js: "", ...build.initialOptions.banner };
+	build.initialOptions.banner.js += dedent`
+		function __cf_cjs(esm) {
+		  const cjs = 'default' in esm ? esm.default : {};
```
pi0, line 130: This can fail because esm can be a primitive value (see [this](https://github.com/unjs/jiti/pull/321))

pi0:
It is a nice approach. I would suggest:

1. See if we can directly make unenv exports better compatible. it is likely we can do this and avoid the overhead and risks of interop at esbuild level
2. If we need interop, i would suggest to use something like [this](https://github.com/unjs/jiti/blob/544247421cce0349d58dc94d58e862a7bc12176a/src/utils.ts#L101) it is much safer. cjs interops can be really tricky.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -81,38 +82,76 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
+			for (const [k, v] of Object.entries(esm)) {
+				if (k !== 'default') {
+					Object.defineProperty(cjs, k, {
+						enumerable: true,
+						value: v,
+					});
+				}
+			}
+			return cjs;
```
pi0, line 139: cjs (default) can potentially miss other named exports. unlike node.js, not every library has identical namespace exports to `default`. Even in some cases (because of interop used in libraries that import them, they might try to access `.default` (from our interoped CJS default)

vicb:
Thanks for super valuable feedback @pi0!

I was hoping that given the restricted scope of this (`/unenv/runtime/{npm,mock}`) a simpler solution would be enough.

> * If we need interop, i would suggest to use something like [this](https://github.com/unjs/jiti/blob/544247421cce0349d58dc94d58e862a7bc12176a/src/utils.ts#L101) it is much safer. cjs interops can be really tricky.

But if you think this is not safe, using `interopDefault` should solve all the problems if I understand correctly? 

> * See if we can directly make unenv npm compat exports more compatible. it is likely we can do this and avoid the overhead and risks of interop at esbuild level

One "issue" is that the unenv update in wrangler is paused for now because of [a different issue](https://github.com/cloudflare/workers-sdk/pull/6920). To make the change in unenv, we would need to update unenv and have a release without the perf_hook latest change. Might take longer than just releasing the change here (that we can revert later if no more needed).

pi0:
Feel free to open a PR in unenv to revert that. (i think a simply uncomment in preset would solve it)

petebacondarwin (approved):
Let's merge this for now and iterate on the concerns that @pi0 raised.

PR #6939, opened 2024-10-10
chore: convert file registry to class and add tests

andyjessop:

Fixes #000.

This adds unit tests to the file-based registry. In order to do that, it converts the file to a class so that it can be easily re-instantiated in each test.

There are no logic changes to the source code - this is just a re-org.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Not user-facing.


packages/wrangler/src/__tests__/file-registry.test.ts
```diff
@@ -0,0 +1,817 @@
...
+
+		// Verify the worker is no longer registered
+		const registeredWorkers = await registry.getRegisteredWorkers();
+		expect(registeredWorkers?.[workerName]).toBeUndefined();
+	});
+
+	// TODO: This is a difficult problem, but could potentially be solved by maintaining a registry that is
+	//       out-of-process of all wrangler instances. It could then guarantee (e.g. by using a mutex) that
+	//       there will be no read/write conflicts.
+	describe.concurrent("Concurrent read/write operations", () => {
```
penalosa, line 794: This is definitely a problem with the current implementation, but I'm not sure it's worth the effort to fix. In most realistic use cases there won't be multiple Wrangler instances with the same worker name running (outside of some of our tests), and even if we used a mutex the _user_-facing behaviour could still be unexpected (I don't think there's a reasonable intuitive outcome for running multiple wrangler dev sessions with the same name).

penalosa (approved):
Some testing nits, but overall LGTM!

andyjessop:
I'm feeling a lot less confident about this PR now, as it seems like there is some inherent flakiness that moving to a class somehow brings out. We now have the package tests passing on Windows, but failing on Linux and Mac. 

My inclination is to close this PR for now, given the recent increase in stability after moving to the file-based registry by default.

PR #6941, opened 2024-10-10
WC-2695 Add better visibility for router-worker

WillTaylorDev:

Fixes #WC-2695

This commit instruments the router worker with analytics such as request time, colo metadata, error, etc, in order for us to have better visibility into the router worker.

These changes were tested using gradual rollouts with a 0% version and a Cloudflare-Workers-Version-Overrides header.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: not a unit-testable change
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: 
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/workers-shared/router-worker/src/analytics.ts
+++ b/packages/workers-shared/router-worker/src/analytics.ts
@@ -1,0 +1,71 @@
+import type { Environment, ReadyAnalytics } from "./types";
+
+// This will allow us to make breaking changes to the analytic schema
+const VERSION = 1;
+
+export enum DISPATCH_TYPE {
+	ASSETS = "asset",
+	WORKER = "worker",
+}
+
+// When adding new columns please update the schema
+type Data = {
+	// -- Doubles --
+	// double1 - The time it takes for the whole request to complete in milliseconds
+	requestTime?: number;
+	// double2 - Colo ID
+	coloId?: number;
+	// double3 - Metal ID
+	metalId?: number;
+	// double4 - Colo tier (e.g. tier 1, tier 2, tier 3)
+	coloTier?: number;
+
+	// -- Blobs --
+	// blob1 - Hostname of the request
+	hostname?: string;
+	// blob2 - Dispatch type - what kind of thing did we dispatch
+	dispatchtype?: DISPATCH_TYPE;
+	// blob3 - Error message
+	error?: string;
+	// blob4 - The current version UUID of router-server
+	version?: string;
+	// blob5 - Region of the colo (e.g. WEUR)
+	coloRegion?: string;
+};
+
+export class Analytics {
+	private data: Data = {};
+
+	setData(newData: Partial<Data>) {
+		this.data = { ...this.data, ...newData };
+	}
+
+	getData(key: keyof Data) {
+		return this.data[key];
+	}
+
+	write(env: Environment, readyAnalytics?: ReadyAnalytics, hostname?: string) {
+		if (!readyAnalytics) {
+			return;
+		}
+
+		readyAnalytics.logEvent({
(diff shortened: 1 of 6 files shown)


packages/workers-shared/router-worker/src/analytics.ts
```diff
@@ -0,0 +1,70 @@
+// This will allow us to make breaking changes to the analytic schema
+const VERSION = 1;
+
+export enum DISPATCH_TYPE {
+	ASSETS = "asset",
+	WORKER = "worker",
+}
+
+// When adding new columns please update the schema
+type Data = {
```
GregBrimble, line 10: Good schema! When we add more complex routing config options, I'll be tempted to add some sort of "routing evaluation time" thing, but we can do that later of course!

packages/workers-shared/router-worker/src/analytics.ts
```diff
@@ -0,0 +1,70 @@
...
+			accountId: 0, // TODO: need to plumb through
+			indexId: hostname,
+			doubles: [
+				this.data.requestTime ?? -1, // double1
+				this.data.coloId ?? -1, // double2
+				this.data.metalId ?? -1, // double3
+				this.data.coloTier ?? -1, // double4
+			],
+			blobs: [
+				this.data.hostname?.substring(0, 256), // blob1 - trim to 256 bytes
```
GregBrimble, line 62: Is 256 bytes RA's limit? Or are we just being conscientiously space-saving here?

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -8,6 +10,11 @@ interface Env {
 	CONFIG: RoutingConfig;
 
 	SENTRY_DSN: string;
+	ENVIRONMENT: Environment;
```
GregBrimble, line 13: Have you tested this in local dev when these values won't be populated?

WillTaylorDev: Good call! We needed to wrap some of the setData calls to ensure the bindings were present first.

packages/workers-shared/router-worker/src/analytics.ts
WillTaylorDev: 256 bytes per field, so technically we could go up to 256*20, but effectively this is just reserving 1/20th of the total available space.

packages/workers-shared/router-worker/src/analytics.ts
WillTaylorDev: Credit goes to Walshy-- this was largely stolen from analytics in pages :)

GregBrimble (approved):

WalshyDev (approved):

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -16,7 +29,9 @@ interface Env {
 export default {
 	async fetch(request: Request, env: Env, ctx: ExecutionContext) {
 		let sentry: ReturnType<typeof setupSentry> | undefined;
-		const maybeSecondRequest = request.clone();
+		const analytics = new Analytics();
```
petebacondarwin, line 32: NIT: similar to the PerformanceTimer, why not pass in the env.ANALYTICS and env.ENVIRONMENT here in the constructor, rather than in the write method?

petebacondarwin (approved):
So I guess this all works in Miniflare (when router worker is used for local development) because each of the new env properties can be undefined?

PR #6943, opened 2024-10-10
fix(queues): Respect consumer `max_retries` and `retry_delay` properly

sdnts:

Fixes #MQ-731.

A user on Discord reported that for large retry delays, local queues were not retrying enough. Turns out the queue DO was being terminated. Thanks Samuel and Michael Hart for the assist!

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: I included a unit test 
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this is bugfix

### DIFF
--- a/packages/miniflare/src/workers/queues/schemas.ts
+++ b/packages/miniflare/src/workers/queues/schemas.ts
@@ -21,15 +21,24 @@ export type QueueProducer = z.infer<typeof QueueProducerSchema>;
 export const QueueProducersSchema =
 	/* @__PURE__ */ z.record(QueueProducerSchema);
 
-export const QueueConsumerOptionsSchema = /* @__PURE__ */ z.object({
-	// https://developers.cloudflare.com/queues/platform/configuration/#consumer
-	// https://developers.cloudflare.com/queues/platform/limits/
-	maxBatchSize: z.number().min(0).max(100).optional(),
-	maxBatchTimeout: z.number().min(0).max(30).optional(), // seconds
-	maxRetires: z.number().min(0).max(100).optional(),
-	deadLetterQueue: z.ostring(),
-	retryDelay: QueueMessageDelaySchema,
-});
+export const QueueConsumerOptionsSchema = /* @__PURE__ */ z
+	.object({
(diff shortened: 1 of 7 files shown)


jbwcloudflare (approved):
thanks Sid!

packages/miniflare/src/workers/queues/schemas.ts
```diff
@@ -26,7 +26,7 @@ export const QueueConsumerOptionsSchema = /* @__PURE__ */ z.object({
 	// https://developers.cloudflare.com/queues/platform/limits/
 	maxBatchSize: z.number().min(0).max(100).optional(),
 	maxBatchTimeout: z.number().min(0).max(30).optional(), // seconds
-	maxRetires: z.number().min(0).max(100).optional(),
+	maxRetries: z.number().min(0).max(100).optional(),
```
penalosa, line 29: I don't think we can actually remove this, since it's part of the Miniflare public API surface. Instead, could you add a new `maxRetries` property and make it fall back to `maxRetires` if not defined?

sdnts: Oh, didn't realize that. Yeah I can do this!

packages/miniflare/src/workers/queues/schemas.ts
```diff
@@ -21,15 +21,24 @@ export type QueueProducer = z.infer<typeof QueueProducerSchema>;
...
+		maxRetries: z.number().min(0).max(100).optional(),
+		deadLetterQueue: z.ostring(),
+		retryDelay: QueueMessageDelaySchema,
+	})
+	.transform((queue) => {
+		if (queue.maxRetires !== undefined) {
+			queue.maxRetries = queue.maxRetires;
+		}
+
+		return queue as Omit<typeof queue, "maxRetires">;
```
sdnts, line 40: @penalosa this should do it

penalosa (approved):

PR #6957, opened 2024-10-14
refactor: simplify tsconfig templates

vicb:

`create-cloudflare` generates unnecessary verbose `tsconfig.json`.

I think it would be better to generate simpler configs not to confuse out users - simpler is better!

The current `tsconfig.json` templates are probably derived from `tsc --init` which does not make a lot of sense because they would be different for different version of TypeScript.

Fixes N/A.

- Tests
  - [ ] TODO (before merge)
  - [] Tests included
  - [X] Tests not necessary because: Remove commented out code in templates
  - [ ] I don't know
  - [X] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [X] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: No functional changes

### DIFF
--- a/.changeset/twenty-grapes-taste.md
+++ b/.changeset/twenty-grapes-taste.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+refactor: simplify tsconfig.json templates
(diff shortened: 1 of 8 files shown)


irvinebroque (commented):
seems reasonable to me

PR #6963, opened 2024-10-14
fix: respect config.account_id in wrangler dev --remote

### LINKED ISSUE #6947: 🐛 BUG: on by default: –x-dev-env causes issue with multiple accounts

Wrangler

0.0.0-3fa846ec2

_No response_

Mac

### Describe the Bug

When I upgraded from 3.79.0 to 3.80.3 I started seeing the following error when running `wrangler dev --remote`.

```
✘ [ERROR] More than one account available but unable to select one in non-interactive mode.

  Please set the appropriate `account_id` in your `wrangler.toml` file.
  Available accounts are (`<name>`: `<account_id>`):
     ...redacted...
```

This didn't make sense because my wrangler.toml does have an account_id specified.

I went to the [versions page](https://www.npmjs.com/package/wrangler/v/0.0.0-3fa846ec2?activeTab=versions) and tried out each prerelease version after 3.79.0 until I arrived at `0.0.0-3fa846ec2` which triggered the issue.

Looking at [the commit](https://github.com/cloudflare/workers-sdk/commit/3fa846ec2), there's this line, which unblocked me:

> If you experience any issues, you can disable the flag with `--x-dev-env=false`. Please also let us know by opening an issue at https://github.com/cloudflare/workers-sdk/issues/new/choose.

Using `--x-dev-env=false` fixes the issue!

_No response_

_No response_

RamIdeas:

Fixes #6947

- Tests
  - [x] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [x] TODO (before merge)
  - [ ] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: fix


packages/wrangler/src/dev.tsx
```diff
@@ -711,11 +711,11 @@ export async function startDev(args: StartDevOptions) {
 						}),
 					},
 					dev: {
-						auth: async () => {
+						auth: async (config) => {
 							let accountId = args.accountId;
 							if (!accountId) {
 								unregisterHotKeys?.();
-								accountId = await requireAuth({});
+								accountId = await requireAuth(config);
```
RamIdeas, line 718: main fix for `wrangler dev --remote`

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -53,14 +53,16 @@ async function resolveDevConfig(
...
-				apiToken: requireApiToken(),
-			};
-		});
+	const auth = async () => {
+		if (input.dev?.auth) {
+			return unwrapHook(input.dev.auth, config);
+		}
+
+		return {
+			accountId: config.account_id ?? (await getAccountId()),
```
RamIdeas, line 62: fix default for `startWorker`

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -129,7 +129,7 @@ export interface StartDevWorkerInput {
 		/** Whether the worker runs on the edge or locally. */
 		remote?: boolean;
 		/** Cloudflare Account credentials. Can be provided upfront or as a function which will be called only when required. */
-		auth?: AsyncHook<CfAccount>;
+		auth?: AsyncHook<CfAccount, [Pick<Config, "account_id">]>; // provide config.account_id as a hook param
```
RamIdeas, line 132: only the ConfigController has access to the wrangler.toml config so enable it as a hook param for `StartDevWorkerInput.dev.auth`

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -187,6 +187,7 @@ export type StartDevWorkerOptions = Omit<StartDevWorkerInput, "assets"> & {
 	};
 	dev: StartDevWorkerInput["dev"] & {
 		persist: string;
+		auth?: AsyncHook<CfAccount>; // redefine without config.account_id hook param (can only be provided by ConfigController with access to wrangler.toml, not by other controllers eg RemoteRuntimeContoller)
```
RamIdeas, line 190: Other controllers won't have access to the wrangler.toml config file so refine this without a hook param on `StartDevWorkerOptions.dev.auth`

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -53,14 +53,16 @@ async function resolveDevConfig(
...
-		input.dev?.auth ??
-		(async () => {
-			return {
-				accountId: await getAccountId(),
-				apiToken: requireApiToken(),
-			};
-		});
+	const auth = async () => {
+		if (input.dev?.auth) {
+			return unwrapHook(input.dev.auth, config);
```
penalosa, line 58: I'm not sure this is the same behaviour. This seems to unconditionally unwrap the hook if provided, even if not in remote mode?

packages/wrangler/src/api/startDevWorker/types.ts
penalosa: How will this work if e.g. the remote runtime controller unwraps the hook? Will that not fail?

packages/wrangler/src/api/startDevWorker/ConfigController.ts
RamIdeas: It doesn't unconditionally unwrap the hook. It unconditionally sets the auth property with its own hook which will unwrap the original property when itself is unwrapped

(This diff answers your other question. https://github.com/cloudflare/workers-sdk/pull/6963/files#r1799613869)

packages/wrangler/src/api/startDevWorker/ConfigController.ts
Skye-31: Does this mean config has a precedent over the env variable?

penalosa: This should probably use the `requireAuth()` helper

penalosa (approved):
Approved pending https://github.com/cloudflare/workers-sdk/pull/6963/files#r1800710064 and the TODO test

packages/wrangler/src/api/startDevWorker/ConfigController.ts
RamIdeas: > This should probably use the `requireAuth()` helper

https://github.com/cloudflare/workers-sdk/pull/6963/commits/5396ab50544827050e2bdb51408368da207c80a5

> Does this mean config has a precedent over the env variable?

Yes. Which matches previous behaviour unless I'm mistaken?

Skye-31: > Yes. Which matches previous behaviour unless I'm mistaken

So it does, how surprising

RamIdeas: I don't think it's surprising. IMO order of precedence: cli args > config file > env vars, because env vars can be set across projects, config file per project, cli args per command – the specificity is clear.

I can understand setting an env var inline and expecting that to take precedence (eg `CLOUDFLARE_ACCOUNT_ID=1234 npx wrangler deploy`) but instead the cli arg should be used for that use-case (eg. `npx wrangler deploy --account-id 1234`) because that env var could've come from the eg ~/.zshrc file

emily-shen: I've never seen `--account-id` before - it doesn't seem to be documented 🙃

RamIdeas: Huh you're right! We should add one to support per command overrides with higher precedence than the config file

PR #6970, opened 2024-10-15
PIPE-37 Add authenticated HTTP source options to Worker Pipelines

oliy:

Fixes [PIPE-37](https://jira.cfdata.org/browse/PIPE-37) (Support Authenticated HTTP Source in Worker Pipelines)

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] TODO (before merge)
  - [X] Changeset included
  - [ ] Changeset not necessary because: Not sure, but this is a small feature on existing functionality
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: New docs are being created by PM for first experience docs.

### DIFF
--- a/packages/wrangler/src/__tests__/pipelines.test.ts
+++ b/packages/wrangler/src/__tests__/pipelines.test.ts
@@ -7,7 +7,7 @@ import { mockConsoleMethods } from "./helpers/mock-console";
 import { msw } from "./helpers/msw";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { runWrangler } from "./helpers/run-wrangler";
-import type { Pipeline, PipelineEntry } from "../pipelines/client";
+import type { HttpSource, Pipeline, PipelineEntry } from "../pipelines/client";
 
 describe("pipelines", () => {
 	const std = mockConsoleMethods();
@@ -39,7 +39,7 @@ describe("pipelines", () => {
 			},
 		},
 		endpoint: "https://0001.pipelines.cloudflarestorage.com",
-	};
+	} satisfies Pipeline;
 
 	function mockCreateR2Token(bucket: string) {
 		const requests = { count: 0 };
@@ -145,7 +145,10 @@ describe("pipelines", () => {
 		status: number = 200,
 		error?: object
 	) {
-		const requests = { count: 0 };
+		const requests: { count: number; body: Pipeline | null } = {
+			count: 0,
+			body: null,
+		};
 		msw.use(
 			http.post(
 				"*/accounts/:accountId/pipelines",
@@ -153,6 +156,7 @@ describe("pipelines", () => {
 					expect(params.accountId).toEqual("some-account-id");
 					const config = (await request.json()) as Pipeline;
 					expect(config.name).toEqual(name);
+					requests.body = config;
 					requests.count++;
 					const pipeline: Pipeline = {
 						...config,
@@ -167,7 +171,7 @@ describe("pipelines", () => {
 							messages: [],
 							result: pipeline,
 						},
-						{ status: status }
+						{ status }
 					);
 				},
 				{ once: true }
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/pipelines/index.ts
```diff
@@ -136,9 +141,20 @@ function addCreateAndUpdateOptions(yargs: Argv<CommonYargsOptions>) {
 			type: "string",
 			demandOption: false,
 		})
+		.option("binding", {
+			describe: "Enable Worker binding to this pipeline\nDefault: true",
+			type: "boolean",
```
CarmenPopoviciu, line 146:
```suggestion
			type: "boolean",
			default: true,
```

packages/wrangler/src/pipelines/index.ts
```diff
@@ -136,9 +141,20 @@ function addCreateAndUpdateOptions(yargs: Argv<CommonYargsOptions>) {
 			type: "string",
 			demandOption: false,
 		})
+		.option("binding", {
+			describe: "Enable Worker binding to this pipeline\nDefault: true",
```
CarmenPopoviciu, line 145: `yargs` comes with a `default` option config that you can set (see my [suggestion](https://github.com/cloudflare/workers-sdk/pull/6970/files#r1804558117) below), which makes explicitly specifying the default in your description unnecessary.

```suggestion
			describe: "Enable Worker binding to this pipeline",
```

CarmenPopoviciu: this is how it'll look like

Screenshot 2024-10-17 at 12 00 28

packages/wrangler/src/pipelines/index.ts
```diff
@@ -136,9 +141,20 @@ function addCreateAndUpdateOptions(yargs: Argv<CommonYargsOptions>) {
...
+			type: "boolean",
+			demandOption: false,
+		})
+		.option("http", {
+			describe:
+				"Enable HTTPS endpoint to send data to this pipeline\nDefault: true",
+			type: "boolean",
```
CarmenPopoviciu, line 152:
```suggestion
			type: "boolean",
			default: true,
```

packages/wrangler/src/pipelines/index.ts
```diff
@@ -136,9 +141,20 @@ function addCreateAndUpdateOptions(yargs: Argv<CommonYargsOptions>) {
...
+			demandOption: false,
+		})
 		.option("authentication", {
 			describe:
-				"Enabling authentication means that data can only be sent to the pipeline via the binding \nDefault: false",
+				"Require authentication (Cloudflare API Token) to send data to the HTTPS endpoint\nDefault: false",
```
CarmenPopoviciu, line 157:
```suggestion
				"Require authentication (Cloudflare API Token) to send data to the HTTPS endpoint",
```

packages/wrangler/src/pipelines/index.ts
```diff
@@ -136,9 +141,20 @@ function addCreateAndUpdateOptions(yargs: Argv<CommonYargsOptions>) {
...
 			type: "boolean",
```
CarmenPopoviciu, line 158:
```suggestion
			type: "boolean",
			default: true,
```

packages/wrangler/src/pipelines/index.ts
```diff
@@ -237,13 +244,27 @@ export function pipelines(pipelineYargs: CommonYargsArgv) {
...
+						format: "json",
+						config: {
+							authentication:
+								args.authentication !== undefined && args.authentication,
 						},
-					];
+					} satisfies HttpSource);
+				}
+				if (pipelineConfig.source.length < 1) {
+					throw new FatalError("At leat one source should be specified");
```
CarmenPopoviciu, line 267: is this a `UserError` or a `FatalError`. The difference is that the latter will be tracked in Sentry, whereas the former won't

CarmenPopoviciu: I also wonder if we could make this err message more actionable on the user side. So for example, can we point out how they should specify a source? either by providing an example or linking to some docs?

packages/wrangler/src/__tests__/pipelines.test.ts
```diff
@@ -167,7 +171,7 @@ describe("pipelines", () => {
 							messages: [],
 							result: pipeline,
 						},
-						{ status: status }
```
CarmenPopoviciu, line 170: 🎉

packages/wrangler/src/__tests__/pipelines.test.ts
```diff
@@ -361,6 +412,32 @@ describe("pipelines", () => {
 		expect(requests.count).toEqual(1);
 	});
 
+	it("create - should create a pipeline with auth", async () => {
```
CarmenPopoviciu, line 415: should we perhaps group all the `create` specs into a dedicated suite? So 👇 

```
describe("create, () => {
  it("should create a pipeline with auth", async () => {});
  ...
})
```

Same for `update`.

I tend to find this kind of grouping much easier to read, download in my brain, and maintain ^.^

CarmenPopoviciu (changes requested):
awesome stuff ♥️ !! Left a few nits that would make this PR even more awesome 😋

CarmenPopoviciu (approved):

packages/wrangler/src/__tests__/pipelines.test.ts
oliy: Done!  (although, i used to do this, too, but have started going back to defining individual. tests, so that it's easier to isolate individual tests by temporarily adding in `.only` for specific tests that i want to iteratively debug.)

CarmenPopoviciu: oh, does that not work as smoothly with `describe.only` -> `it.only()`?

Issue #6353, opened 2024-07-28
🐛 BUG: relative imports from a nested `main` in `base_dir` breaks


GregBrimble:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What version(s) of the tool(s) are you using?

3.67.1 [Wrangler]

### What version of Node are you using?

v20.9.0

### What operating system and version are you using?

macOS 14.5 (23F79)

### Observed behavior
```
- some/
  - base_dir/
    - nested/
      - index.js
    - foo.js
- wrangler.toml
```

<details>
<summary><code>./some/base_dir/nested/index.js</code></summary>
<pre>
import { x } from "../foo.js";

export default {
  fetch() {
    return new Response(x);
  }
}
</pre>
</details>

<details>
<summary><code>./some/base_dir/foo.js</code></summary>
<pre>
export const x = "Hello, world!";
</pre>
</details>

<details>
<summary><code>./wrangler.toml</code></summary>
<pre>
main = "./some/base_dir/nested/index.js"
base_dir = "./some/base_dir"
no_bundle = true
rules = [
  { type = "ESModule", globs = ["**/*.js"] }
]
</pre>
</details>

`npx wrangler dev` collects `foo.js` as an additional module, but errors with

```txt
✘ [ERROR] service core:user:worker: Uncaught Error: internal error


✘ [ERROR] MiniflareCoreError [ERR_RUNTIME_FAILURE]: The Workers runtime failed to start. There is likely additional logging output above.
```

### Expected behavior
I'm able to import from a parent when `base_dir` includes those modules.

I suspect what is happening is that we correctly attach additional modules with their relative path to `base_dir` (e.g. `/foo.js`), but the main entry path is being incorrectly set as `/index.js` rather than `/nested/index.js`, so relative imports from there are busted.

### Steps to reproduce
As above.

Changing the import to `import { x } from "/foo.js";` "fixes" the issue, hence my theory on how we're setting the current module path.

penalosa:
Apologies for the delayed response! Fixing this at this stage would constitue a breaking change, since it would break anyone relying the fact that entrypoint files are never nested.

Going forward, I think we'll want to rethink the entire bundling config setup with `base_dir`/`no_bundle`/etc... as it can get pretty confusing. We might end up something closer to a deploy dir specification, but there's definitely still discussion to be had there.

Since we're unlikely to implement this as-is, I'm going to close the issue for now.

PR #6971, opened 2024-10-15
chore: fix entrypoint logic and refactor

andyjessop:

Fixes #000. This was supposed to fix (but doesn't) #6353. However, it does do a refactor that improves the code base and adds a relevant test case.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: Small refactor
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Bug fix


penalosa:
I'm pretty sure the fix for the original issue is to change https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/src/dev/miniflare.ts#L291-L292 to `const modulesRoot = config.bundle.entry.moduleRoot;`

penalosa (approved):

andyjessop:
> I'm pretty sure the fix for the original issue is to change https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/src/dev/miniflare.ts#L291-L292 to `const modulesRoot = config.bundle.entry.moduleRoot;`

Or better yet, make miniflare respect the config setting? We have `build.moduleRoot` in the config object, we just don't pass it through when generating the miniflare config.

penalosa:
> Or better yet, make miniflare respect the config setting? 

I'm not sure I get the distinction here—`config.bundle.entry` _is_ the config setting. It comes from https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/src/dev/use-esbuild.ts#L233. Once we get rid of the non --x-dev-env flow we can probably pass it through to miniflare.ts from the LocalRuntimeController more cleanly

andyjessop:
@penalosa ah, got it. Thanks for pointing that out 👍 

Config is a bit messy and could do with some consolidation I think.

andyjessop:
> I'm pretty sure the fix for the original issue is to change https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/src/dev/miniflare.ts#L291-L292 to `const modulesRoot = config.bundle.entry.moduleRoot;`

@penalosa, I think this change is causing the E2E tests to fail. From what I can tell, the `config.bundle.entry.moduleRoot` is referring to the worker definition, but the `scriptPath` was referring to the bundled code (located at `.wrangler/tmp/dev-[hash]/index.js`.

PR #6985, opened 2024-10-15
Simplify pull_request_template.md

penalosa:
This PR simplifies the PR template by:
- Removing the comment about highlighting PR reviews
- Removing the changesets sections. Instead, the PR validation has been updated to _require_ changesets, which is the right default for the vast majority of cases. This can be bypassed with the `no-changeset-required` label if needed. This should be really clear to developers, as the changeset bot posts a very visible comment with a warning icon when no changesets are found, and the failing PR validation check has clear language about how to fix things.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: minor tooling change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: self documenting

### DIFF
--- a/tools/deployments/__tests__/validate-pr-description.test.ts
+++ b/tools/deployments/__tests__/validate-pr-description.test.ts
@@ -4,7 +4,7 @@ import { validateDescription } from "../validate-pr-description";
 describe("validateDescription()", () => {
 	it("should skip validation with the `skip-pr-description-validation` label", () => {
 		expect(
-			validateDescription("", "", '["skip-pr-description-validation"]')
+			validateDescription("", "", '["skip-pr-description-validation"]', "[]")
 		).toHaveLength(0);
 	});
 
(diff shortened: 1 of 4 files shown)


tools/deployments/__tests__/validate-pr-description.test.ts
```diff
@@ -26,28 +26,54 @@ Fixes #[insert GH or internal issue number(s)].
...
+				"[]",
 				"[]"
 			)
 		).toMatchInlineSnapshot(`
 			[
 			  "All TODO checkboxes in your PR description must be unchecked before merging",
 			  "Your PR must include tests, or provide justification for why no tests are required",
 			  "Your PR must run E2E tests, or provide justification for why running them is not required",
-			  "Your PR must include a changeset, or provide justification for why no changesets are required",
+			  "Your PR doesn't include a changeset. Either include one (following the instructions in CONTRIBUTING.md) or add the 'no-changeset-required' label to bypass this check. Most PRs should have a changeset, so only bypass this check if your change should not cause a release of any packages.",
```
CarmenPopoviciu, line 42: `Most PRs should have a changeset, so only bypass this check if your change should not cause a release of any packages.`

I wonder if we can be even more specific here as to what constitutes a change that requires a changeset. "should not cause a release of any packages" is still pretty wide IMO, so might be helpful for our external contributors to either narrow it down even more, or maybe give some examples

penalosa: We could maybe link specifically to https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets?

emily-shen: maybe external contributors shouldn't even worry about changesets? i think we might actually save time overall by writing changesets for external contributors ourselves vs explaining our changeset conventions and chasing people up

RamIdeas (approved):

tools/deployments/__tests__/validate-pr-description.test.ts
penalosa: Added a link to the docs here to make this clearer

PR #6990, opened 2024-10-15
Add additional output fields to pages deployments

courtney-sims:
https://github.com/cloudflare/wrangler-action/issues/300

Adds additional outputs to wrangler's artifact mechanism for pages deploys. This allows wrangler-action to pull from wrangler's artifacts to get info about pages deploys, adding parity to the outputs pages-action provides.

Test by installing the wrangler prerelease in a local pages project, setting WRANGLER_OUTPUT_FILE_DIRECTORY in the local environment, running a pages deploy, and checking the output file.

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
  - [X] Not required because: This code is not covered by those tests.
  - [ ] TODO (before merge)
  - [X] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: Internal usage for now

### DIFF
--- a/packages/wrangler/src/pages/deploy.tsx
+++ b/packages/wrangler/src/pages/deploy.tsx
@@ -448,6 +448,16 @@ ${failureMessage}`,
 		url: deploymentResponse.url,
 	});
 
+	writeOutput({
+		type: "pages-deploy-detailed",
+		version: 1,
+		pages_project: deploymentResponse.project_name,
+		deployment_id: deploymentResponse.id,
+		url: deploymentResponse.url,
+		alias,
+		environment: deploymentResponse.environment,
+	});
+
 	await metrics.sendMetricsEvent("create pages deployment");
 };
 
--- a/.changeset/little-cheetahs-arrive.md
+++ b/.changeset/little-cheetahs-arrive.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feat: Adds new detailed pages deployment output type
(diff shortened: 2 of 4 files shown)


packages/wrangler/src/pages/deploy.tsx
```diff
@@ -446,6 +446,8 @@ ${failureMessage}`,
 		pages_project: deploymentResponse.project_name,
 		deployment_id: deploymentResponse.id,
 		url: deploymentResponse.url,
+		alias: alias,
```
petebacondarwin, line 449: NIT - you can just do

```suggestion
		alias,
```

petebacondarwin (commented):
Thanks @courtney-sims.

You'll need a changeset and some tests. See https://github.com/cloudflare/workers-sdk/blob/09b50927a62731f8aa621b9d872d10d1900a60a5/packages/wrangler/src/__tests__/output.test.ts

But also I think it might be safest if you create a new OutputEntry, rather than updating a current one. These entries are a contract between Wrangler and other tools that have an expectation of what is in the entry. Although this is adding new fields so would probably not break a downstream client of the output, you might have a situation where a client expects fields to be there which are not?

Either way, I think that the wrangler-action would need to ensure that the version of Wranger being used was new enough that this output included these fields.

CarmenPopoviciu (approved):

PR #6995, opened 2024-10-16
fix: quarantine C3 analog e2e tests

CarmenPopoviciu:

This PR quarantines the C3 `analog` e2e tests, which started failing since yesterday. After thoroughly debugging together with Pete, we concluded that the breakage is not caused by one particular change in `workers-sdk`, but rather due to some changes in `analog` itself. Specifically, the breakage seems to come from the [latest release](https://github.com/analogjs/analog/blob/v1.9.0/CHANGELOG.md) of analog, aka `v 1.9.0`. Currently C3 has a dependency on analog `^1.8.0`, which ends up pulling `1.9.0`, and results in the failures we're seeing. We attempted to fix the version number of our dependency to `1.8.0`, but unfortunately that still end up pulling ` 1.9.0`, due to how analog internally declares its dependencies. This is something we don't have control over, which is why we ultimately decided to quarantine these tests.

We also attempted to fix forward, based off of https://github.com/cloudflare/workers-sdk/pull/6975, that bumps `create-analog` to `1.9.0`, but unfortunately that broke other things, so here we are.

These tests should be quarantined until we are either able to successfully bump `create-analog` to `1.9.0`, or until a downstream fix comes from the analog team <3

Fixes N/A.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: e2e test quarantine
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [ ] Changeset included
  - [x] Changeset not necessary because: e2e test quarantine
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: e2e test quarantine

### DIFF
--- a/packages/create-cloudflare/e2e-tests/frameworks.test.ts
+++ b/packages/create-cloudflare/e2e-tests/frameworks.test.ts
@@ -131,6 +131,7 @@ function getFrameworkTests(opts: {
 				],
 			},
 			analog: {
+				quarantine: true,
 				testCommitMessage: true,
 				timeout: LONG_TIMEOUT,
 				unsupportedOSs: ["win32"],


petebacondarwin (approved):

CarmenPopoviciu:
cc// @brandonroberts FYI

brandonroberts:
Thanks for the heads up @CarmenPopoviciu. Is C3 using the `latest` template or still pinned to `angular-v17`? I think using the `--template latest` would fix in the install issues, but I will check and see what else can be done to resolve this.

CarmenPopoviciu:
thank you so much @brandonroberts ♥️ !!! From what I could tell, we're still pinned to `angular-v17`. I *think* Pete tried to pin to latest but that didn't work? I'll have to check again.

Kepp me posted please if you find anything, or if I can action anything on our side

PR #7002, opened 2024-10-17
Add more helpful error messages when validating compatibility date

GregBrimble:

Fixes WC-2854.

image

image

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e coverage
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: just adding better errors :)


.changeset/fifty-dots-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+feat: More helpful error messages when validating compatibility date
```
CarmenPopoviciu, line 5:
```suggestion
fix: More helpful error messages when validating compatibility date
```

packages/wrangler/src/config/validation.ts
```diff
@@ -3159,6 +3159,43 @@ const validateConsumer: ValidatorFn = (diagnostics, field, value, _config) => {
 	return isValid;
 };
 
+const isValidDate = (value: string): boolean => {
+	const data = new Date(value);
+	return !isNaN(data.getTime());
+};
+
+const validateCompatibilityDate: ValidatorFn = (diagnostics, field, value) => {
```
CarmenPopoviciu, line 3167: let's make sure we surface all errors on first throw, as opposed to one per cmd run. See my comment [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/config/validation.ts#L2083-L2084). It makes for a much better UX and it's something we're pushing for since recently.

[validateAssetsConfig](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/config/validation.ts#L2067) is a good example of how'd like things to work

packages/wrangler/src/config/validation.ts
```diff
@@ -3159,6 +3159,43 @@ const validateConsumer: ValidatorFn = (diagnostics, field, value, _config) => {
...
+		);
+		return false;
+	}
+
+	if (
+		value.includes("–") || // en-dash
+		value.includes("—") // em-dash
+	) {
+		diagnostics.errors.push(
+			`Hyphens (-) should be used rather than en-dashes (—) or em-dashes (–) in the "${field}" field.`
```
CarmenPopoviciu, line 3184:
```suggestion
			`"${field}" field should use hyphens (-) rather than en-dashes (—) or em-dashes (–).`
```

CarmenPopoviciu (changes requested):

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: ✅

CarmenPopoviciu (approved):

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: ✅

packages/wrangler/src/config/validation.ts
```diff
@@ -3200,6 +3200,47 @@ const validateConsumer: ValidatorFn = (diagnostics, field, value, _config) => {
 	return isValid;
 };
 
+const isValidDate = (value: string): boolean => {
```
andyjessop, line 3203: We have the `validation-helpers` file - could this be moved into there? I know it's not an ideal spot, as this is a really very generic helper and we don't have a generalised "utils" folder, but maybe it will help for discoverability if it's in a dedicated helper file.

packages/wrangler/src/config/validation.ts
```diff
@@ -3200,6 +3200,47 @@ const validateConsumer: ValidatorFn = (diagnostics, field, value, _config) => {
...
+	let isValid = true;
+
+	if (
+		value.includes("–") || // en-dash
+		value.includes("—") // em-dash
+	) {
+		isValid = false;
+
+		diagnostics.errors.push(
+			`"${field}" field should use hyphens (-) rather than en-dashes (—) or em-dashes (–).`
```
andyjessop, line 3229: Is there a better way to demonstrate this, given that all three look identical in monospace fonts? Maybe there isn't, but I thought I would challenge it just because it reads very strangely in the terminal.

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: ✅

packages/wrangler/src/config/validation.ts
CarmenPopoviciu: not that I can tell I'm afraid. I totes get what you mean, and I did look into it, but it doesn't seem to be possible for us to control how the user's terminal will display these things.

andyjessop: NP! Thanks for looking into it.

andyjessop (approved):

PR #7012, opened 2024-10-17
Workflows binding

RamIdeas:

Fixes #0000

This PR adds a `workflows` section to wrangler.toml to enable binding to a Workflow entrypoint.

Following this guide: https://github.com/cloudflare/workers-sdk/blob/workflows-binding/packages/wrangler/docs/how-to/add-a-binding.md

TODO:
- [ ] add e2e test using the binding
- [ ] add support to `wrangler types`

- Tests
  - [x] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [x] TODO (before merge)
  - [ ] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/16220


packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -507,6 +512,18 @@ export async function bundleWorker(
...
 
+	const notExportedWorkflows = workflowBindings
+		.filter((x) => !x.script_name && !entryPoint.exports.includes(x.class_name))
+		.map((x) => x.class_name);
+	if (notExportedWorkflows.length) {
+		const relativePath = path.relative(process.cwd(), entryFile);
+		throw new UserError(
+			`Your Worker depends on the following Workflows, which are not exported in your entrypoint file: ${notExportedWorkflows.join(
+				", "
+			)}.\nYou should export these objects from your entrypoint, ${relativePath}.`
```
Skye-31, line 523: Perhaps worth including something like "or, if defined in another script, add `script_name` to your config"

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -240,6 +240,23 @@ export default async function triggersDeploy(
...
+						script_name: scriptName,
+						class_name: workflow.class_name,
+					}),
+					headers: {
+						"Content-Type": "application/json",
+					},
+				}).then(() => [`workflow: ${workflow.name}`])
+			);
+		}
+	}
```
RamIdeas, line 258: This change affects `wrangler triggers deploy` (which is used internally by `wrangler deploy`, `wrangler versions deploy`).

I was going to wrap these changes in an experimental flag (`–x-workflows`) but it seems redundant since there is already a if-block around it guarding on whether the user has created a `[[workflows]]` section in their wrangler.toml

Happy to add the flag (to all 3 commands) if anyone insists, but I've decided that it isn't necessary

packages/wrangler/src/config/validation.ts
```diff
@@ -2008,6 +2018,63 @@ const validateDurableObjectBinding: ValidatorFn = (
 	return isValid;
 };
 
+/**
+ * Check that the given field is a valid "workflow" binding object.
+ */
+const validateWorkflowBinding: ValidatorFn = (_diagnostics, _field, _value) => {
+	// if (typeof value !== "object" || value === null) {
```
andyjessop, line 2025: Can we just remove all these commented-out lines? Or, it seems like we just return true from this, so maybe we don't need to include this function at all?

packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -507,6 +512,18 @@ export async function bundleWorker(
 		);
 	}
 
+	const notExportedWorkflows = workflowBindings
+		.filter((x) => !x.script_name && !entryPoint.exports.includes(x.class_name))
+		.map((x) => x.class_name);
+	if (notExportedWorkflows.length) {
+		const relativePath = path.relative(process.cwd(), entryFile);
```
andyjessop, line 519: Is this always going to be relative to the cwd? We have access to the `Entry` here too, so we could feasibly use that instead.

fixtures/workflow/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# worker-ts
```
emily-shen, line 1:
```suggestion
# workflow-demo
```

RamIdeas: This file isn't needed at all 😂

emily-shen: that makes *a lot* more sense 😅

packages/wrangler/src/deployment-bundle/bundle.ts
RamIdeas: Good call

packages/wrangler/src/config/validation.ts
RamIdeas: Argh I noticed this and meant to come back and add some validation

.changeset/green-parrots-lick.md
```diff
@@ -0,0 +1,22 @@
...
+name = "my-workflow"
+class_name = "MyDemoWorkflow"
+```
+
+and export an entrypoint (e.g. `MyDemoWorkflow`) in your script:
+
+```typescript
+import { WorkflowEntrypoint } from "cloudflare:workers";
+
+export class MyDemoWorkflow extends WorkflowEntrypoint<{}, Params> {...}
```
edmundhung, line 21: Nit: Should we give the first generic a proper name here instead of `{}`?

RamIdeas: Good call

sidharthachatterjee (approved):
Looks good for now, thanks

packages/wrangler/src/deployment-bundle/bundle.ts
RamIdeas: Seems entryFile _is_ entry.file

packages/wrangler/src/config/validation.ts
RamIdeas: i've deleted the commented code and explicitly returned true with a TODO comment

packages/wrangler/src/triggers/deploy.ts
RamIdeas: cc @lrapoport-cf

petebacondarwin: I think we still need the flag. In fact, I think we should fail `wrangler.toml` validation if there is a `workflows` binding but no opt-in flag.

petebacondarwin: Or else, the section in wrangler.toml should be called `experimental_workflows`...

lrapoport-cf: agree, let's still have the flag 👍 even though it's redundant, i think it still makes sense to keep so that users know they are using a new/separate pathway for workflows. 

after adding, if user has a workflow binding but doesn't specify the flag, will we error or warn? i'd lean towards being conservative and erroring, but happy to discuss

lrapoport-cf: https://github.com/cloudflare/workers-sdk/pull/7047#issuecomment-2429896208

PR #7024, opened 2024-10-18
Refactor hyperdrive create and update commands to match each other's behavior and address bug with individual parameters on creation

xortive:

Fixes SQC-338

```
fix: make individual parameters work for `wrangler hyperdrive create` when not using HoA

`wrangler hyperdrive create` individual parameters were not setting the database name correctly when calling the api.

refactor: use same param parsing code for `wrangler hyperdrive create` and `wrangler hyperdrive update`

ensures that going forward, both commands support the same features and have the same names for config flags

feature: allow using a connection string when updating hyperdrive configs

both `hyperdrive create` and `hyperdrive update` now support updating configs with connection strings.
```

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [X] I don't know
  - [ ] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [X] Changeset included
- Public documentation
  - [X] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Documentation not necessary because: no changes

### DIFF
--- a/packages/wrangler/src/hyperdrive/client.ts
+++ b/packages/wrangler/src/hyperdrive/client.ts
@@ -9,19 +9,45 @@ export type HyperdriveConfig = {
 	caching: CachingOptions;
 };
 
-export type PublicOrigin = {
-	host?: string;
-	port?: number;
-	scheme?: string;
-	database?: string;
-	user?: string;
-	access_client_id?: string;
+export type OriginCommon = {
+	host: string;
+	scheme: string;
+	database: string;
+	user: string;
 };
 
-export type OriginWithSecrets = PublicOrigin & {
-	password?: string;
-	access_client_secret?: string;
+export type OriginCommonWithSecrets = OriginCommon & {
+	password: string;
 };
+
+export type OriginHoA = OriginCommon & {
+	access_client_id: string;
+	access_client_secret?: never;
+	port?: never;
+};
+
+export type OriginHoAWithSecrets = OriginCommonWithSecrets & {
+	access_client_id: string;
+	access_client_secret: string;
+	port?: never;
+};
+
+export type OriginHostAndPort = OriginCommon & {
+	access_client_id?: never;
+	access_client_secret?: never;
+	port: number;
+};
+
+export type OriginHostAndPortWithSecrets = OriginCommonWithSecrets & {
+	access_client_id?: never;
+	access_client_secret?: never;
+	port: number;
+};
+
+export type PublicOrigin = OriginHostAndPort | OriginHoA;
+export type OriginWithSecrets =
+	| OriginHostAndPortWithSecrets
+	| OriginHoAWithSecrets;
 
 export type CachingOptions = {
 	disabled?: boolean;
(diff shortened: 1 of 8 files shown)


packages/wrangler/src/hyperdrive/client.ts
```diff
@@ -9,20 +9,46 @@ export type HyperdriveConfig = {
...
-export type OriginWithSecrets = PublicOrigin & {
-	password?: string;
-	access_client_secret?: string;
+export type OriginCommonWithSecrets = OriginCommon & {
+	password: string;
 };
 
+export type OriginHoA = OriginCommon & {
+	access_client_id: string;
+	access_client_secret?: never;
```
ReppCodes, line 25: Having the base OriginHoA include the secrets, when we have a WithSecrets version, seems wrong to me?

packages/wrangler/src/hyperdrive/client.ts
```diff
@@ -9,20 +9,46 @@ export type HyperdriveConfig = {
...
+	port?: never;
+};
+
+export type OriginHoAWithSecrets = OriginCommonWithSecrets & {
+	access_client_id: string;
+	access_client_secret: string;
+	port?: never;
+};
+
+export type OriginHostAndPort = OriginCommon & {
```
ReppCodes, line 35: In SQC-Config, the "HostAndPort" naming convention is what we use for non-HoA configs. Having the HoA-specific fields here seems wrong?

packages/wrangler/src/hyperdrive/client.ts
xortive: `OriginHoa` is actually enforcing that the secret is not present here, this type is saying "access_client_secret" must never be set.

packages/wrangler/src/hyperdrive/client.ts
xortive: This is doing the same thing as OriginHoOA, instead enforcing that the HoA fields are not present

ReppCodes:
Looks good to me from the Hyperdrive Team side of things.

emily-shen (changes requested):
Tests are also currently failing - think its just test snapshots need to be updated though (run them locally and press `u`)

packages/wrangler/src/hyperdrive/index.ts
```diff
@@ -28,3 +41,265 @@ export function hyperdrive(yargs: CommonYargsArgv) {
...
+			);
+		} else if (url.username === "") {
+			throw new UserError(
+				"You must provide a username - e.g. 'user:password@database.example.com:port/databasename'"
+			);
+		} else if (url.password === "") {
+			throw new UserError(
+				"You must provide a password - e.g. 'user:password@database.example.com:port/databasename' "
+			);
+		}
```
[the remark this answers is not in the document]
xortive: We have some arguments like `--port` where you need to provide `--port` OR `--access-client-id` and `--access-client-secret` that makes doing that complex, and I don't think there will be an improvement in user experience that justifies the work required as a result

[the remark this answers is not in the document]
emily-shen: wont block on this, but for future reference, in the same place where you currently throw you can push the error messages to an array, then join them back up and throw once after the if/else checks :)

emily-shen (approved):

PR #7037, opened 2024-10-21
fix: ask for confirmation before creating a new worker when uploading secrets

emily-shen:
Fixes [WC-2856](https://jira.cfdata.org/browse/WC-2856)
In interactive mode, if someone specifies a worker name that does not yet exist, ask for confirmation before actually creating that worker (e.g. in case it was a typo).

In non-interactive contexts (CI etc.), always create a worker to maintain existing behaviour.
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: trivial interaction change

### DIFF
--- a/packages/wrangler/src/__tests__/secret.test.ts
+++ b/packages/wrangler/src/__tests__/secret.test.ts
@@ -1003,6 +1003,7 @@ describe("wrangler secret", () => {
 		});
 
 		it("should, in interactive mode, ask to create a new worker if no worker is found under the provided name", async () => {
+			setIsTTY(true);
 			writeFileSync(
 				"secret.json",
 				JSON.stringify({
@@ -1024,6 +1025,7 @@ describe("wrangler secret", () => {
 		});
 
 		it("should, in non-interactive mode, create a new worker if no worker is found under the provided name", async () => {
+			setIsTTY(false);
 			writeFileSync(
 				"secret.json",
 				JSON.stringify({
--- a/packages/wrangler/src/__tests__/secret.test.ts
+++ b/packages/wrangler/src/__tests__/secret.test.ts
@@ -43,6 +43,44 @@ export function mockGetMemberships(
 	);
 }
 
+function mockNoWorkerFound(isBulk = false) {
+	if (isBulk) {
+		msw.use(
+			http.get(
+				"*/accounts/:accountId/workers/scripts/:scriptName/settings",
+				async () => {
+					return HttpResponse.json(
+						createFetchResult(null, false, [
+							{
+								code: 10007,
+								message: "This Worker does not exist on your account.",
+							},
+						])
+					);
+				},
+				{ once: true }
+			)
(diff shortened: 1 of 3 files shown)


emily-shen:
cc @tanushree-sharma

tanushree-sharma:
Quick PR! Thanks @emily-shen.

Possible to move the check for a Worker up to before the user is prompted for the secret? Doing the Worker check first is better in the case that a user specified the wrong Worker name and wants to abort. 
 
image

emily-shen:
That would be ideal DX, but we'd have to add an extra "get worker" request to every secret upload and i'm not sure that's worth it. Currently we just directly upload the secret to an endpoint that's something like `/workerName/secrets` so it's just one request most of the time.

packages/wrangler/src/__tests__/secret.test.ts
```diff
@@ -967,6 +1001,60 @@ describe("wrangler secret", () => {
 			expect(std.err).toMatchInlineSnapshot(`""`);
 			expect(std.warn).toMatchInlineSnapshot(`""`);
 		});
+
+		it("should, in interactive mode, ask to create a new worker if no worker is found under the provided name", async () => {
```
CarmenPopoviciu, line 1005:
```suggestion
		it("should, in interactive mode, ask to create a new Worker if no Worker is found under the provided name", async () => {
```

packages/wrangler/src/secret/index.ts
```diff
@@ -57,7 +57,15 @@ async function createDraftWorker({
 	accountId: string;
 	scriptName: string;
 }) {
-	// TODO: log a warning
+	const confirmation = await confirm(
+		`No Worker called "${scriptName}" appears to exist. Do you want to create a new Worker and add secrets to it?`,
```
CarmenPopoviciu, line 61: are we sure this is the behaviour we want here? Should we instead users to chose a name for this new Worker? Prompting that "non-existent-worker" does not exist feels a bit shady to me?

emily-shen: What do you mean by shady haha? Before the behaviour was to silently create a new worker with whatever name you passed in - if you typo'd the name, you'd end up with a new worker with the typo name. But also some people actually did want to create new workers + upload secrets at them same time, which is why this behaviour was added. 

I might tweak the wording tho:

`There doesn't seem to be a worker called "${scriptName}". Do you want to create a new Worker with that name and add secrets to it?`

CarmenPopoviciu: for postery, as per our async convo, turns I missunderstood how this works, and what you have in place here is totes 👌

CarmenPopoviciu (approved):

PR #7038, opened 2024-10-21
fix: only show fetch warning if on old compatibility_date

### LINKED ISSUE #6950: 🚀 Feature Request: Remove checked-fetch & allow fetches to custom ports in wrangler dev
The limitation around fetching custom ports has been removed when deployed, so we should remove the check locally. See https://developers.cloudflare.com/workers/configuration/compatibility-dates/#allow-specifying-a-custom-port-when-making-a-subrequest-with-the-fetch-api for more context

petebacondarwin:
Now that we have the `allow_custom_ports` compatibility flag, we only need to show the fetch warnings when that flag is not enabled.

Fixes #6950

Fixes https://github.com/cloudflare/workerd/issues/2955

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
  - [ ] TODO (before merge)
  - [x] Changeset included
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/shaggy-monkeys-visit.md
+++ b/.changeset/shaggy-monkeys-visit.md
@@ -1,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+fix: only show fetch warning if on old compatibility_date
+
+Now that we have the `allow_custom_ports` compatibility flag, we only need to show the fetch warnings when that flag is not enabled.
+
+Fixes https://github.com/cloudflare/workerd/issues/2955
--- a/packages/wrangler/src/__tests__/check-fetch.test.ts
+++ b/packages/wrangler/src/__tests__/check-fetch.test.ts
@@ -1,0 +1,28 @@
+import "vitest";
+import { shouldCheckFetch } from "../deployment-bundle/bundle";
+
+describe("shouldCheckFetch()", () => {
+	it("should be true for old compat date", () => {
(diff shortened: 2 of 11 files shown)


fixtures/worker-app/src/index.js
penalosa, line 1: Is there a reason to remove everything else in this file?

penalosa (approved):
16 files for such a small change 😢

fixtures/worker-app/src/index.js
petebacondarwin: Oops. That was just local testing (which proved that it doesn't actually work in production right now!!)

ceddybi:
@petebacondarwin @penalosa not just hiding the warning, fetching with custom HTTPS port doesn't work too, looks like the port is wiped somewhere

petebacondarwin:
@ceddybi - good point. It seems that on production this now works correctly if the compat date is new (>2024-09-01) or the compat flag is on (allow_custom_ports). But my local testing seems to show workerd crashing in Miniflare. I need to dig in why that is the case. Notwithstanding that, this PR is still valid, since in production the warning is no longer needed when the compat date/flag is setup correctly.

PR #7039, opened 2024-10-21
Only show dev registry connection status in local dev

### LINKED ISSUE #7009: FEEDBACK: red or green dot on DO services
In PR https://github.com/cloudflare/workers-sdk/pull/6893, a red or green dot was added beside each service when deploying a DO with Wrangler.

I'm not too familiar with Wrangler internals, but it looks like the red vs. green is decided based on whether a service is in the registry.

Per some feedback on [Discord](https://canary.discord.com/channels/595317990191398933/779390076219686943/1296367336462225469), this isn't really clear to users, and in fact the random red dot can be kind of concerning given that there is no explanation on what it means or why your services are suddenly red.

I believe it was added in PR https://github.com/cloudflare/workers-sdk/pull/6893:
https://github.com/cloudflare/workers-sdk/blob/ae4bfead3ad9911823ec154476cb1c49ea3a02bf/packages/wrangler/src/config/index.ts#L268

penalosa:
Fixes #7009

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: small output change

### DIFF
--- a/packages/wrangler/src/config/index.ts
+++ b/packages/wrangler/src/config/index.ts
@@ -264,16 +264,21 @@ export function printBindings(
 				({ name, class_name, script_name }) => {
 					let value = class_name;
 					if (script_name) {
-						const registryDefinition = context.registry?.[script_name];
-						if (
-							registryDefinition &&
-							registryDefinition.durableObjects.some(
-								(d) => d.className === class_name
-							)
-						) {
-							value += ` (defined in 🟢 ${script_name})`;
+						if (context.local) {
+							const registryDefinition = context.registry?.[script_name];
+
+							if (
+								registryDefinition &&
+								registryDefinition.durableObjects.some(
+									(d) => d.className === class_name
+								)
+							) {
+								value += ` (defined in 🟢 ${script_name})`;
+							} else {
+								value += ` (defined in 🔴 ${script_name})`;
+							}
 						} else {
-							value += ` (defined in 🔴 ${script_name})`;
+							value += ` (defined in ${script_name})`;
 						}
 					}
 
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/config/index.ts
```diff
@@ -264,16 +264,21 @@ export function printBindings(
...
+									(d) => d.className === class_name
+								)
+							) {
+								value += ` (defined in 🟢 ${script_name})`;
+							} else {
+								value += ` (defined in 🔴 ${script_name})`;
+							}
 						} else {
-							value += ` (defined in 🔴 ${script_name})`;
+							value += ` (defined in ${script_name})`;
```
CarmenPopoviciu, line 281: what's the logic based on which we apply these red/green bullets? Cause it seems we apply them in some places but no in others?

penalosa: Before this PR we applied them to all DOs & service bindings based on whether they were in the dev registry. After this PR we apply to all DOs & service bindings based on whether they were in the dev registry, but hide them if not in local dev

packages/wrangler/src/config/index.ts
```diff
@@ -264,16 +264,21 @@ export function printBindings(
...
+
+							if (
+								registryDefinition &&
+								registryDefinition.durableObjects.some(
+									(d) => d.className === class_name
+								)
+							) {
+								value += ` (defined in 🟢 ${script_name})`;
+							} else {
+								value += ` (defined in 🔴 ${script_name})`;
```
CarmenPopoviciu, line 278: now that I've read the actual issue 😅 , I'm seeing that our users' confusion is around using the red bullets. Do we still want to use them in this place in that case?

penalosa: I think this is a good first step—it removes them from the dploy flow, which is where I think most confusion was coming from

CarmenPopoviciu: kk

PR #7040, opened 2024-10-21
Try and ensure the `d.ts` types are definitely the ones loaded in `quick-edit-extension`

penalosa:
Make sure we're loading the .d.ts and not the .ts version of the types. Occasionally the wrong version has been loaded in CI, causing spurious typing errors for users in the playground & quick edit. This PR simplifies the loading, and adds an assertion to make sure builds will fail if the wrong version is loaded. 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: tooling change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: tooling change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal tooling change

### DIFF
--- a/packages/quick-edit-extension/src/cfs.ts
+++ b/packages/quick-edit-extension/src/cfs.ts
@@ -4,7 +4,7 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import workersTypes from "raw:@cloudflare/workers-types/experimental/index.d.ts";
+import workersTypes from "raw:workers-types";
 import {
 	Disposable,
 	EventEmitter,
--- a/.changeset/pink-geckos-peel.md
+++ b/.changeset/pink-geckos-peel.md
@@ -1,5 +1,4 @@
 ---
-"quick-edit-extension": patch
 "@cloudflare/quick-edit": patch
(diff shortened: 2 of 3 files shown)


packages/quick-edit-extension/src/cfs.ts
```diff
@@ -4,7 +4,7 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-import workersTypes from "raw:@cloudflare/workers-types/experimental/index.d.ts";
+import workersTypes from "raw:workers-types";
```
CarmenPopoviciu, line 7: for my understanding, what's the diff between the two?

CarmenPopoviciu: is this possible because we're now resolving the path to `node_modules/@cloudflare/workers-types/experimental/index.d.ts`?

CarmenPopoviciu (approved):

packages/quick-edit-extension/src/cfs.ts
penalosa: Yeah—we now resolve the path in the esbuild plugin rather than from the import statement

PR #7047, opened 2024-10-21
require --x-workflows flag to deploy a workflow

RamIdeas:
Fixes #0000. Addresses https://github.com/cloudflare/workers-sdk/pull/7012#discussion_r1809165316

This PR implements the `–x-workflows` flag on `wrangler deploy`, `wrangler versions upload`, `wrangler versions deploy` and `wrangler trigger deploy`.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/rare-planes-warn.md
+++ b/.changeset/rare-planes-warn.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+require --x-workflows flag to deploy a workflow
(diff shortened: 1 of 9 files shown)


threepointone:
hey, imo this should work without the flag, and just log a warning that it's experimental? why have a flag for this?

threepointone:
The --x flag would be useful if it was triggering al alternate to an existing command/flag, but I don't think that's the case here?

lrapoport-cf:
we can have a clear warning in place of the flag

PR #7067, opened 2024-10-23
feat: change trigger command to comply with new endpoint

LuisDuarte1:
Fixes WOR-280

Because the body for the trigger command changed, we also need to change it here. As a consequence we also add an optional `--id` to let users to optionally customize the new instance ID.


RamIdeas (approved):

sidharthachatterjee:
Let's stick to the PR template form henceforth please, we should follow guidelines for each repo @LuisDuarte1

PR #7068, opened 2024-10-23
Workflows open beta warning on deploy

RamIdeas:
Supersedes https://github.com/cloudflare/workers-sdk/pull/7047

Fixes #7047

Adds a warning reminding the user that Workflows is in open beta – if their config file contains any workflows – in `wrangler deploy`, `wrangler versions upload`, `wrangler versions deploy` and `wrangler trigger deploy`.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/rare-planes-warn.md
+++ b/.changeset/rare-planes-warn.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+log warning of Workflows open-beta status when running deploying a Worker that contains a Workflow binding
--- a/packages/wrangler/src/__tests__/logger.test.ts
+++ b/packages/wrangler/src/__tests__/logger.test.ts
@@ -189,16 +189,35 @@ describe("logger", () => {
 			expect(std.debug).toMatchInlineSnapshot(`""`);
 			expect(std.out).toMatchInlineSnapshot(`"This is a log message"`);
 			expect(std.warn).toMatchInlineSnapshot(`
-        "Unrecognised WRANGLER_LOG value \\"everything\\", expected \\"none\\" | \\"error\\" | \\"warn\\" | \\"info\\" | \\"log\\" | \\"debug\\", defaulting to \\"log\\"...
-        [33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mThis is a warn message[0m
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mUnrecognised WRANGLER_LOG value \\"everything\\", expected \\"none\\" | \\"error\\" | \\"warn\\" | \\"info\\" | \\"log\\" | \\"debug\\", defaulting to \\"log\\"...[0m
 
-        "
-      `);
+
+				[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mThis is a warn message[0m
+
+				"
+			`);
 			expect(std.err).toMatchInlineSnapshot(`
         "[31mX [41;31m[[41;97mERROR[41;31m][0m [1mThis is a error message[0m
 
         "
       `);
+		});
+	});
+
+	describe("warnOnce", () => {
+		it("should only log the same message once", () => {
+			const logger = new Logger();
+			logger.warnOnce("This is a warnOnce message");
+			logger.warnOnce("This is a warnOnce message");
+			logger.warnOnce("This is a warnOnce message");
+			logger.warnOnce("This is a warnOnce message");
+			logger.warnOnce("This is a warnOnce message");
+
+			expect(std.warn).toMatchInlineSnapshot(`
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mThis is a warnOnce message[0m
+
+				"
+			`);
 		});
 	});
 });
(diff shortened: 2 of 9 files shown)


packages/wrangler/src/logger.ts
```diff
@@ -31,17 +31,6 @@ const getLogLevelFromEnv = getEnvironmentVariableFactory({
 	variableName: "WRANGLER_LOG",
 });
 
-const warnOnce = (() => {
-	let logged = false;
-
-	return (...args: Parameters<typeof console.warn>) => {
-		if (!logged) {
-			console.warn(...args);
```
RamIdeas, line 39: This previously used console.warn instead of logger.warn

This helper has been replaced by logger.warnOnce which will use logger.warn

packages/wrangler/src/__tests__/logger.test.ts
```diff
@@ -189,16 +189,35 @@ describe("logger", () => {
...
-        [33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mThis is a warn message[0m
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mUnrecognised WRANGLER_LOG value \\"everything\\", expected \\"none\\" | \\"error\\" | \\"warn\\" | \\"info\\" | \\"log\\" | \\"debug\\", defaulting to \\"log\\"...[0m
 
-        "
-      `);
+
+				[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mThis is a warn message[0m
+
+				"
+			`);
```
RamIdeas, line 198: This was previously using `console.warn`. Now using `logger.warn` prepends the ansi colourised "[WARNING]"

packages/wrangler/src/logger.ts
```diff
@@ -51,7 +40,7 @@ function getLoggerLevel(): LoggerLevel {
 		const expected = Object.keys(LOGGER_LEVELS)
 			.map((level) => `"${level}"`)
 			.join(" | ");
-		warnOnce(
+		logger.warnOnce(
```
penalosa, line 43: I'm not sure we should be using `logger` methods inside `getLoggerLevel()` like this—it feels like a bit of a circular dependency

.changeset/rare-planes-warn.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
emily-shen, line 2:
```suggestion
"wrangler": patch
```
since afaik experimental/unreleased feats are patches...?

packages/wrangler/src/logger.ts
```diff
@@ -118,6 +107,17 @@ export class Logger {
 		Logger.#afterLogHook?.();
 	}
 
+	static warnOnceHistory = new Set();
```
penalosa, line 110: Why is this static? It feels like each logger instance should have it's own warnOne set

packages/wrangler/src/logger.ts
RamIdeas: I thought that may have been the reason. But this function is about to override the default log level – since the default log level will not swallow warnings, I think it's okay to use logger.warn

packages/wrangler/src/logger.ts
RamIdeas: If a subclass wants a new set, it can override its own static property i.e. `static warnOnceHistory = new Set()` on the class

I guess if it was an instance property, the inverse could also be said: If a subclass wants to share the set, it can override its own instance property i.e. `this.warnOnceHistory = logger.warnOnceHistory` in the constructor

The former seems simpler but not tied to either tbh

packages/wrangler/src/logger.ts
penalosa: Is it not going to be slightly circular though? `logger.warnOnce()` will eventually call `getLoggerLevel()` which will then call `logger.warnOnce()` again and so on. I might be misunderstanding, but that would be my expectation

packages/wrangler/src/logger.ts
RamIdeas: Ah sorry you said instance not subclass

The logger is currently a singleton so I'm more concerned with subclasses (we do this with Miniflare's Log class to intercept method calls, where we would expect this kind of property to be shared).

From this feature's perspective, I'd want these warnings to be logged once per invocation of wrangler not per instance of the logger

packages/wrangler/src/logger.ts
RamIdeas: Yeah I guess. But luckily logger.warnOnce only calls logger.warn... once 😉

packages/wrangler/src/logger.ts
penalosa: The logger is used as a singleton in Wrangler, but there's no inherit reason it has to be, and in various tests we construct multiple instances. It seems to me that it would be kinda strange for those instances to share this warnOnce set.

In terms of subclasses, maybe we're coming at this from different angles, but I would be really surprised if a method call on an instance of the parent class affected method calls on an instance of a subclass.

PR #7069, opened 2024-10-23
No more React in `wrangler dev` (`--x-dev-env` only)

### LINKED ISSUE #6935: 🐛 BUG: Importing `getPlatformProxy` hijacks certain `process.on` events

Wrangler

3.80.2 [Wrangler]

22.9.0

Mac Sequoia 15.0.1

### Observed behavior
Simply importing `getPlatformProxy` from wrangler automatically adds `process.on('uncaughtException', ...)` and `process.on('unhandledRejection', ...)` listeners that re-throw . This prevents user code from handling those errors.

### Expected behavior
Being able to listen to and handle these errors despite using `getPlatformProxy`

### Explanation
Importing runtime APIs like `getPlatformProxy` leads to importing the code of the cli and some of that code runs on import.

From what I can tell the listeners that hijack these events stem from the `yoga-layout` package used within `ink`. The dependency is precompiled and zipped and lives inside the `vendor/` directory in this repo.

The specific handlers added (wish I could point directly to the source here but it being precompiled and zipped makes that not possible):
```js
// wrangler/wrangler-dist/cli.js:29768
process.on("uncaughtException", function(a) {
  throw a;
});
process.on(
  "unhandledRejection",
  function(a) {
    throw a;
  }
);
```

Re-throwing here makes all listeners added after not trigger.

### Steps to reproduce

Small repro also provided below, but simply given this code:

```ts
import { getPlatformProxy } from 'wrangler';

// This line is not needed, the import alone is enough.
getPlatformProxy().then(proxy => console.log(proxy));

process.on('uncaughtException', error => {
  console.error('Oops! An uncaughtException occurred:', error);
})

process.on('unhandledRejection', error => {
  console.error('Oops! An unhandledRejection occurred:', error);
})

throw Error('This is an error');
```

One expects to see the error handler called. This is not the case.

```
<dir>/node_modules/wrangler/wrangler-dist/cli.js:29768
            throw a;
            ^

Error: This is an error
    at file:///<dir>/wrangler-repro/index.js:14:7
    at ModuleJob.run (node:internal/modules/esm/module_job:262:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:483:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)
```

https://github.com/rognstadragnar/cloudflare-wrangler-bug-repro

_No response_

penalosa:
Removes the non `--x-dev-env` flow from `wrangler dev`

Among other things, fixes #6935 since we no longer load Ink or Yoga

See below for note for posterity regarding considerations for landing this change: https://github.com/cloudflare/workers-sdk/pull/7069#issuecomment-2468149214

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


lrapoport-cf (changes requested):
please capture a note for posterity summarizing when --x-dev-env was flipped and issues observed/time period without issues since then

IgorMinar (commented):
I'm very very excited about this clean up! Well done @penalosa (and the team) for finally getting us here!

fixtures/pages-functions-with-routes-app/tests/index.test.ts
```diff
@@ -12,6 +12,7 @@ describe("Pages Functions with custom _routes.json", () => {
 			"public",
 			["--port=0", "--inspector-port=0"]
 		));
+		console.log(ip, port);
```
edmundhung, line 15: Is this intentional?

penalosa: https://github.com/cloudflare/workers-sdk/pull/7069/commits/93e02541648acb60957b3c93f42b83aeeef2a890

packages/wrangler/src/api/dev.ts
```diff
@@ -84,13 +82,13 @@ export interface UnstableDevOptions {
 		devEnv?: boolean;
 		fileBasedRegistry?: boolean;
 		vectorizeBindToProd?: boolean;
+		enableIpc?: boolean;
```
penalosa, line 85: This is a new flag—it's required to make the Pages use of `unstable_dev` work with the ipc channel setup for fixture tests.

packages/wrangler/src/api/startDevWorker/devtools.ts
```diff
@@ -11,31 +11,3 @@ export type DevToolsEvent<Method extends _EventMethods> = Method extends unknown
 			params: _Params<Protocol.Events[Method]>;
 		}
 	: never;
-
-export type DevToolsEvents = DevToolsEvent<_EventMethods>;
```
penalosa, line 15: This wasn't used anywhere

packages/wrangler/src/api/startDevWorker/types.ts
```diff
@@ -214,26 +214,6 @@ export type File<Contents = string, Path = string> =
...
-
-export type PatternRoute = {
-	pattern: string;
-} & (
-	| { pattern: string; customDomain: true }
-	| { pattern: string; zoneId: string; customDomain?: true; zoneName?: never }
-	| { pattern: string; zoneName: string; customDomain?: true; zoneId?: never }
-);
-export type WorkersDevRoute = { workersDev: true };
-export type Route = PatternRoute | WorkersDevRoute;
```
penalosa, line 236: These weren't used anywhere

package.json
```diff
@@ -80,7 +80,6 @@
 			"@types/node": "$@types/node"
 		},
 		"patchedDependencies": {
-			"ink@3.2.0": "patches/ink@3.2.0.patch",
```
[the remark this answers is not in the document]
CarmenPopoviciu: OMG! long live the queen! 😅 👑 🎉

packages/wrangler/e2e/dev-registry.test.ts
```diff
@@ -27,6 +29,114 @@ async function fetchJson<T>(url: string, info?: RequestInit): Promise<T> {
...
+							configPath: "wrangler.toml",
+							experimental: {
+								disableExperimentalWarning: true,
+							},
+						}
+					);
+
+					await setTimeout(2000)
+
+					console.log(await parentWorker.fetch("/").then(r => r.text()))
```
CarmenPopoviciu, line 120: is this `log` intentional here?

packages/wrangler/src/dev.ts
```diff
@@ -604,456 +561,207 @@ export async function startDev(args: StartDevOptions) {
...
-			const projectRoot = configPath && path.dirname(configPath);
-			let config = readConfig(configPath, args);
-
 			if (
-				(args.legacyAssets || config.legacy_assets) &&
-				(args.site || config.site)
+				(args.enableIpc || !args.onReady) &&
+				process.send &&
+				typeof vitest === "undefined"
 			) {
```
penalosa, line 574: This is a change in behaviour to properly enable `unstable_dev()` with Pages. Previously, the `DEV_SERVER_READY` event was _only_ sent if `process.send()` was defined (and we weren't in tests). Now, there are two ways of reporting the port:
- If the `onReady()` callback is provided, use that
- Otherwise, if `process.send()` is defined, use that.

These two options are _usually_ mutually exclusive (if you're running `unstable_dev()` you don't usually want to send an IPC message to the parent process). However, the Pages use of `unstable_dev` relied on `DEV_SERVER_READY` to make the `workers-sdk` fixture tests run using `runWranglerLongLived()`, and so the `enableIpc` flag turns on _both_ of these port reporting mechanisms. Long-term, we should refactor Pages to use `startWorker()` and our fixture tests to not rely on IPC messages to detect the port (similar to the Wrangler E2E tests), which will allow us to remove this logic.

packages/wrangler/src/dev.ts
```diff
@@ -1071,177 +779,6 @@ export async function startDev(args: StartDevOptions) {
 	}
 }
 
-export async function startApiDev(args: StartDevOptions) {
```
penalosa, line 1074: This has been entirely removed, since we no longer need a non-react equivalent of the dev flow—the normal dev flow is react-free!

packages/wrangler/src/dev.ts
```diff
@@ -1320,124 +857,7 @@ export function getInferredHost(routes: Route[] | undefined) {
 	}
 }
 
-export async function validateDevServerSettings(
```
penalosa, line 1323: This functionality is duplicated in the `ConfigController`

packages/wrangler/src/dev/dev.ts
penalosa, line 1: This is a renamed version of `dev.tsx` with most content stripped—this could do with some better organisation in the future

packages/wrangler/src/dev/dev.tsx
penalosa, line 1: This has actually been renamed, not deleted—see above

packages/wrangler/src/dev/proxy.ts
penalosa, line 1: This file is completely superseded by the `ProxyWorker`

packages/wrangler/src/dev/local.ts
penalosa, line 1: This is renamed from `local.tsx`

packages/wrangler/src/dev/remote.ts
penalosa, line 1: This is renamed from `remote.tsx`

packages/wrangler/src/hyperdrive/client.ts
```diff
@@ -116,18 +116,6 @@ export async function listConfigs(config: Config): Promise<HyperdriveConfig[]> {
 	});
 }
 
-export async function updateConfig(
```
penalosa, line 119: Unrelated to the rest of this PR, but this function was not used

packages/wrangler/src/pages/dev.ts
```diff
@@ -914,7 +919,7 @@ export const Handler = async (args: PagesDevArguments) => {
 			testMode: false,
 			watch: true,
 			fileBasedRegistry: args.experimentalRegistry,
-			devEnv: args.experimentalDevEnv,
+			enableIpc: true,
```
penalosa, line 922: See above for what this new flag does

packages/wrangler/e2e/dev-registry.test.ts
penalosa: In this case, yes. This is a `.js` file that is run using `node` on line 125. The return value of the `runInNode()` function is the `stdout` of the node subprocess, which is used to assert against further down the file in the test cases.

packages/wrangler/src/api/dev.ts
```diff
@@ -223,74 +219,41 @@ export async function unstable_dev(
 		logLevel: options?.logLevel ?? defaultLogLevel,
 		port: options?.port ?? 0,
 		experimentalVersions: undefined,
-		experimentalDevEnv: devEnv,
+		experimentalDevEnv: undefined,
```
CarmenPopoviciu, line 222: could we remove this arg altogether? or would that constitute a breaking change?

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -111,7 +111,6 @@ export async function getPlatformProxy<
 	const miniflareOptions = await run(
 		{
 			FILE_BASED_REGISTRY: Boolean(options.experimentalRegistry),
-			DEV_ENV: false,
```
CarmenPopoviciu, line 114: anything we need to dbl check with the frameworks folks here? A bit surprised this still defaults to `false` here 🤔

packages/wrangler/src/api/dev.ts
penalosa: It wouldn't (since it's experimental), but what we've tended to do in the past is keep around the experimental flag for a bit with a message saying that it's no longer needed, just in case people are still using it. That's what I did in this PR, but absolutely open to changing if you think that'd be better!

packages/wrangler/src/api/startDevWorker/devtools.ts
CarmenPopoviciu: thank you for cleaning all these up ♥️

packages/wrangler/src/api/integrations/platform/index.ts
penalosa: This is just removing dead code—the `DEV_ENV` value here was never used to actually enable/disable SDW (there's a `TODO: actually use` comment in `experimental-flags.ts`)

packages/wrangler/src/dev.ts
```diff
@@ -366,6 +338,12 @@ type DevArguments = StrictYargsOptionsToInterface<typeof devOptions>;
 export async function devHandler(args: DevArguments) {
 	await printWranglerBanner();
 
+	if (args.experimentalDevEnv) {
+		logger.warn(
+			"--x-dev-env is no longer required and will be removed in a future version.\n`wrangler dev` now uses the unified dev server by default. 🎉"
```
CarmenPopoviciu, line 343: will users know what "unified dev server" means, or is it even relevant for them? Maybe specifying that `--x-dev-env` is now on by default is enough?

penalosa: That's a fair point! I'll change that 👍

penalosa: https://github.com/cloudflare/workers-sdk/pull/7069/commits/6ffd3efafa05aee5303e4b5f5c5f3833aed290f4

packages/wrangler/e2e/c3-integration.test.ts
```diff
@@ -1,13 +1,14 @@
...
+import { readFileSync } from "node:fs";
 import path from "node:path";
 import { fetch } from "undici";
 import { beforeAll, describe, expect, it } from "vitest";
 import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
 import { generateResourceName } from "./helpers/generate-resource-name";
 
-describe("c3 integration", () => {
+// TODO: Investigate why this is really flaky on windows
+describe.runIf(process.platform !== "win32")("c3 integration", () => {
```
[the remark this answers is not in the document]
edmundhung: Nit: As this PR is a pretty big change, would it be safer to skip the test only after this PR? We might wanna know how this PR might impacts c3 integration on Window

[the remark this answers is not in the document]
penalosa: We could, but I'm not sure how much additional confidence that would give us. This PR makes no changes to the init flow, and so c3 on windows shouldn't be affected

packages/wrangler/src/pages/dev.ts
```diff
@@ -269,6 +268,12 @@ export const Handler = async (args: PagesDevArguments) => {
 
 	await printWranglerBanner();
 
+	if (args.experimentalDevEnv) {
+		logger.warn(
+			"--x-dev-env is no longer required and will be removed in a future version.\n`wrangler dev` now uses the unified dev server by default. 🎉"
```
CarmenPopoviciu: same copy comment regarding "unified dev server"

packages/wrangler/e2e/dev-registry.test.ts
CarmenPopoviciu: yeah, I suspected this might be the reason why it's there, but making sure

packages/wrangler/src/api/dev.ts
CarmenPopoviciu: no no, let's stay consistent in how we do things 👍

CarmenPopoviciu (approved):
whoooohooo! exciting 🎉

penalosa:
For posterity (and summarising internal discussion): we've seen a trickle of issues come in related to `startWorker()` since we turned it on by default a month ago, but all of the issues we've seen have been minor and fix-forwardeable. Users always have the option of staying at a specific version of Wrangler as a workaround until we fix an issue if a new one is reported, and so we've decided to push ahead with removing the old code path.

lrapoport-cf (approved):

PR #7074, opened 2024-10-23
[build] Upgrade CI runners from macOS 13 => 15

fhanau:
Update to the latest OS version, which should speed up CI based on using Apple Silicon CPUs.

This will help us with runner exhaustion encountered before we have a long-term solution for that. As noted below, E2E tests must run on this to confirm that things still work on the new image version.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Changing CI itself, running existing tests will show whether this works as intended
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: CI-only change

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, windows-2022, ubuntu-22.04]
+        os: [macos-15, windows-2022, ubuntu-22.04]
         node: ["18"]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


fhanau:
Makes sense. What do you need the memory for? It seems to work well for the workerd build, even though bazel can be resource hungry at times.

penalosa:
> Makes sense. What do you need the memory for? It seems to work well for the workerd build, even though bazel can be resource hungry at times.

We could probably find a way around this, but I seem to recall last time we tried it made things like our E2E tests more flaky cc @petebacondarwin do you remember the specifics?

penalosa:
Closing based on https://github.com/cloudflare/workers-sdk/pull/7074#issuecomment-2433045040. We can re-evaluate in future once GH (hopefully 🤞) bumps the available memory

PR #7091, opened 2024-10-24
Synchronize observability settings during versions deploy

taylorlee:
In addition to logpush and tail_consumers, patch observability settings
on version deploy. This still doesn't quite match the behavior of
"wrangler deploy", which will disable observability if it is not defined
in wrangler.toml (as of #6776), but at least this is more correct for now.

- Tests
  - [x] Tests included
  - [x] Required
- Public documentation
  - [x] Documentation not necessary because: already exists: https://developers.cloudflare.com/workers/wrangler/configuration/#observability

### DIFF
--- a/.changeset/chilled-apes-repeat.md
+++ b/.changeset/chilled-apes-repeat.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: synchronize observability settings during `wrangler versions deploy`
+
+When running `wrangler versions deploy`, Wrangler will now update `observability` settings in addition to `logpush` and `tail_consumers`. Unlike `wrangler deploy`, it will not disable observability when `observability` is undefined in `wrangler.toml`.
--- a/packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
+++ b/packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
@@ -704,17 +704,78 @@ describe("versions deploy", () => {
 				│
 				│ Synced non-versioned settings:
 				│            logpush:  true
+				│      observability:  <skipped>
 				│     tail_consumers:  <skipped>
 				│
 				╰  SUCCESS  Deployed test-name version 00000000-0000-0000-0000-000000000000 at 100% (TIMINGS)"
 			`);
-
-			expect(normalizeOutput(std.out)).toContain("logpush:  true");
 		});
 
-		test("with logpush and tail_consumers in wrangler.toml", async () => {
+		test("with observability disabled in wrangler.toml", async () => {
+			writeWranglerToml({
+				observability: {
+					enabled: false,
+				},
+			});
+
+			const result = runWrangler(
+				"versions deploy 10000000-0000-0000-0000-000000000000 --yes --experimental-gradual-rollouts"
+			);
+
+			await expect(result).resolves.toBeUndefined();
+
+			expect(normalizeOutput(std.out)).toMatchInlineSnapshot(`
+				"╭ Deploy Worker Versions by splitting traffic between multiple versions
+				│
+				├ Fetching latest deployment
+				│
+				├ Your current deployment has 2 version(s):
+				│
+				│ (10%) 00000000-0000-0000-0000-000000000000
+				│       Created:  TIMESTAMP
+				│           Tag:  -
+				│       Message:  -
+				│
+				│ (90%) 00000000-0000-0000-0000-000000000000
+				│       Created:  TIMESTAMP
+				│           Tag:  -
+				│       Message:  -
+				│
+				├ Fetching deployable versions
+				│
+				├ Which version(s) do you want to deploy?
+				├ 1 Worker Version(s) selected
+				│
+				├     Worker Version 1:  00000000-0000-0000-0000-000000000000
+				│              Created:  TIMESTAMP
+				│                  Tag:  -
+				│              Message:  -
+				│
+				├ What percentage of traffic should Worker Version 1 receive?
+				├ 100% of traffic
+				├
+				├ Add a deployment message (skipped)
+				│
+				├ Deploying 1 version(s)
+				│
+				├ Syncing non-versioned settings
+				│
+				│ Synced non-versioned settings:
+				│            logpush:  <skipped>
+				│      observability:  enabled:  false
+				│     tail_consumers:  <skipped>
+				│
+				╰  SUCCESS  Deployed test-name version 00000000-0000-0000-0000-000000000000 at 100% (TIMINGS)"
+			`);
+		});
+
+		test("with logpush, tail_consumers, and observability in wrangler.toml", async () => {
 			writeWranglerToml({
 				logpush: false,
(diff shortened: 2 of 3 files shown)


.changeset/chilled-apes-repeat.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: synchronize observability settings during versions deploy
```
GregBrimble, line 5:
```suggestion
fix: synchronize observability settings during `wrangler versions deploy`
```

.changeset/chilled-apes-repeat.md
```diff
@@ -0,0 +1,7 @@
...
+
+When running `wrangler versions deploy`, wrangler will now update observability settings in addition to logpush and `tail_consumers`. Unlike `wrangler deploy`, it will not disable observability when observability is undefined in wrangler.toml
```
GregBrimble, line 7:
```suggestion
When running `wrangler versions deploy`, Wrangler will now update `observability` settings in addition to `logpush` and `tail_consumers`. Unlike `wrangler deploy`, it will not disable observability when `observability` is undefined in `wrangler.toml`.
```

packages/wrangler/src/versions/deploy.ts
```diff
@@ -558,11 +558,12 @@ async function promptPercentages(
 async function maybePatchSettings(
 	accountId: string,
 	workerName: string,
-	config: Pick<Config, "logpush" | "tail_consumers">
+	config: Pick<Config, "logpush" | "tail_consumers" | "observability">
 ) {
 	const maybeUndefinedSettings = {
 		logpush: config.logpush,
 		tail_consumers: config.tail_consumers,
+		observability: config.observability, // TODO reconcile with how regular deploy handles empty state
```
GregBrimble, line 566: What's the desired end state here when it's empty?

GregBrimble (approved):

packages/wrangler/src/versions/deploy.ts
taylorlee: I'm not sure. When @RamIdeas added versions-deploy, any undefined settings are not synchronized. When @zebp added observability to regular deploy, it defaults to false when undefined.

It would be unfortunate to need to always patch settings to false for undefined observabililty, but that is the simplest way to reconcile them.

packages/wrangler/src/versions/deploy.ts
```diff
@@ -587,9 +588,19 @@ async function maybePatchSettings(
 		},
 	});
 
+	let observability = [];
```
taylorlee, line 591:
```suggestion
	const observability = [];
```

packages/wrangler/src/versions/deploy.ts
```diff
@@ -587,9 +588,20 @@ async function maybePatchSettings(
...
+			observability.push(
+				`head_sampling_rate: ${patchedSettings.observability.head_sampling_rate}`
+			);
+		}
+	}
 	const formattedSettings = formatLabelledValues(
 		{
 			logpush: String(patchedSettings.logpush ?? "<skipped>"),
+			observability:
+				observability.length > 0 ? observability.join("\n") : "<skipped>",
```
RamIdeas, line 604: if `observability` was a Record<string, string> rather than Array<string> you could nest calls to `formatLabelledValues` with the `indentationCount` options to get nice formatting

RamIdeas: If you have time. If not, no need to block this PR. I can come back and clean it up later.

taylorlee: Done!

packages/wrangler/src/versions/deploy.ts
RamIdeas: If the default is false then how about we remove the default (make it undefined, i.e. falsey) so this just works™️

packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
```diff
@@ -704,17 +704,78 @@ describe("versions deploy", () => {
...
+				├
+				├ Add a deployment message (skipped)
+				│
+				├ Deploying 1 version(s)
+				│
+				├ Syncing non-versioned settings
+				│
+				│ Synced non-versioned settings:
+				│            logpush:  <skipped>
+				│      observability:  enabled:  false
```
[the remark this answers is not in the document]
taylorlee: I'm happy to go with whatever you suggest, but I would personally find that confusing. `logpush: <skipped>` means that wrangler left the remote state unchanged. Omitting head_sampling_rate does not skip synchronizing that field, but causes the default behavior of 100% sampling, which may be different from the prior setting.

packages/wrangler/src/versions/deploy.ts
taylorlee: @RamIdeas I'm not sure I follow. The default does seem to be undefined/falsey, but the difference between deploy and versions-deploy is that deploy always patches the full NonVersionedScriptSettings whereas versions-deploy only does so if at least one field is defined.

So deploy with undefined observability will disable it, whereas versions-deploy will leave it unchanged. Should we always patch during versions-deploy? Or conditionally patch during deploy? Or leave them slightly different?

WalshyDev: I don't think we should block on this. The behaviour here comes down to `wrangler deploy` doing something different for OBS compared to logpush/tail consumers, it will disable if it is not in the config. Meanwhile, logpush/tail will stay with their current values.

If we should do the explicit disable here or if deploy should not do that is a whole topic. The fact that OBS is already consistent isn't great, I'm happy with it at least being consistent here. I think there's a larger discussion that needs to happen (cc: @tanushree-sharma @nevikashah ) on what the behaviour should actually be in deploy & here.

I think part of this whole "user expectation" behaviour comes to the fact that Wrangler is already kinda inconsistent with these things. If you remove a binding from the config, it isn't inherited, it's removed. However, you need to explicitly disable workers.dev, it isn't disabled on not being there.

Imo, we need to figure out what the behaviour actually should be and be as consistent across the board as possible.

taylorlee:
@RamIdeas any qualms with merging this? I think there is more to discuss on how `wrangler deploy` and `wrangler versions deploy` treat undefined settings differently, but I don't think that should block making `observability` handling more correct.

RamIdeas (approved):

packages/wrangler/src/versions/deploy.ts
RamIdeas: Yep no need to block on this. Sounds like there's a backend inconsistency re the observability field? Either way, I agree we need to decide the desired behaviour cc @tanushree-sharma @nevikashah

taylorlee: The backend API is consistent, the inconsistency is whether it's called or not by wrangler. `wrangler deploy` always calls it, `wrangler versions deploy` only calls if if one of the 3 observability settings are defined in wrangler.toml

PR #7098, opened 2024-10-25
Support multi-worker Durable Object RPC

### LINKED ISSUE #5918: 🚀 Feature Request:RPC Support for Durable Objects in `wrangler dev`
### Describe the solution

When binding a durable object using RPC that was defined in a different worker, it fails with:

```
Cannot access `MyDurableObject#myMethod` as Durable Object RPC is not yet supported between multiple `wrangler dev` sessions.
```

This is really useful when using durable objects in combination with `wrangler pages dev` since, as far as I understand, you cannot deploy a worker from a Pages project.

RamIdeas:
Fixes #5918

Support Durable Object RPC across multiple wrangler instances

Only supports RPC method calls right now. RPC property access is coming shortly (in this PR)

The entire implementation is in src/dev/miniflare.ts

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/dev/miniflare.ts
+++ b/packages/wrangler/src/dev/miniflare.ts
@@ -38,7 +38,6 @@ import type {
 	CfWorkflow,
 } from "../deployment-bundle/worker";
 import type {
-	WorkerDefinition,
 	WorkerEntrypointsDefinition,
 	WorkerRegistry,
 } from "../dev-registry";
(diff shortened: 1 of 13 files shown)


packages/wrangler/src/dev/miniflare.ts
```diff
@@ -136,25 +143,18 @@ function createNotFoundWorkerEntrypointClass({ service }) {
...
+\n;; ${/* this script needs to be concatenated with the user's script :( -- use this to avoid ASI syntax errors*/ ""}
+import { WorkerEntrypoint as __Wrangler__LocalOnly__WorkerEntrypoint } from "cloudflare:workers";
+export class ${EXTERNAL_SERVICE_RECEIVER_NAME} extends __Wrangler__LocalOnly__WorkerEntrypoint {
+    async proxyMethod({ bindingName, doId, method, args }) {
+        const ns = this.env[bindingName];
+        const stub = ns.get(ns.idFromString(doId));
+        return stub[method](...args);
 	}
 }
 `;
```
RamIdeas, line 160: This script is added to the user's script – not a separate service as the previous implementation did – so we can bind to it as a regular external entrypoint

This may sound scary but I've taken care to use long+clear names to avoid clashing with user variables in scope and that explain their purpose so if a user ever does discover them they will understand its: (a) injected by wrangler and (b) in local code only, not deployed code

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -634,23 +656,41 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
...
+				);
+				return [
+					bindingName,
+					{
+						className: class_name,
+						useSQLite,
+						script_name,
 					},
 				];
 			}),
```
RamIdeas, line 684: These bindings are also added to the user's worker and on _their_ env object but, again, I've taken care to make long+clear names for the same reasons as above

vitest.shared.ts
```diff
@@ -19,6 +19,6 @@ export default defineConfig({
 		restoreMocks: true,
 		// A lot of the fixture tests are extremely flaky because of the dev registry
 		// Retry tests by default so that only real errors are reported
-		retry: 2,
+		retry: 0,
```
petebacondarwin, line 22: Is this temporary? If not then perhaps update the comment?

RamIdeas: Temporary. Will undo

penalosa:
Closing in favour of https://github.com/cloudflare/workers-sdk/pull/7251

PR #7104, opened 2024-10-26
update unenv to latest version

anonrig:
Updates unenv to latest version. This removes `assert/strict`, `sys`, `path/win32` and `path/posix` aliases from unenv.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: all changes are already tested.
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: it updates internals.

### DIFF
--- a/packages/wrangler/package.json
+++ b/packages/wrangler/package.json
@@ -85,7 +85,7 @@
 		"resolve.exports": "^2.0.2",
 		"selfsigned": "^2.0.1",
 		"source-map": "^0.6.1",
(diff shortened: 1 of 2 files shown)


vicb (changes requested):
I added comments on the unenv PR.

I think the sys change must be reverted and path changes need to be double checked

PR #7109, opened 2024-10-28
Rename the framework name

jsparkdev:
The name currently displayed in the CLI is `Svelte`. However, all of the options actually create a `SvelteKit` app, and the name `SvelteKit` is also used in the [documentation](https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-site/).

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: only display name change
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: it matches existing features with documentation.

### DIFF
--- a/.changeset/happy-items-buy.md
+++ b/.changeset/happy-items-buy.md
@@ -1,0 +1,7 @@
+---
+"create-cloudflare": patch
+---
+
+Rename the framework as it appears in the CLI
+
+The name currently displayed in the CLI is `Svelte`. However, all of the options actually create a `SvelteKit` app, and the name `SvelteKit` is also used in the [documentation](https://developers.cloudflare.com/pages/framework-guides/deploy-a-svelte-site/).
--- a/.changeset/happy-items-buy.md
+++ b/.changeset/happy-items-buy.md
@@ -2,6 +2,6 @@
 "create-cloudflare": patch
 ---
 
-Rename the framework as it appears in the CLI
+Rename the framework name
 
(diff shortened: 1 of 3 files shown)


dario-piotrowicz (commented):
Thanks a bunch for the PR @jsparkdev 🙂 

However I think that this is a duplicate of https://github.com/cloudflare/workers-sdk/pull/6565

Also this has been an issue for a while: https://github.com/cloudflare/workers-sdk/issues/3263 I think we should definitely clarify things 😕 (personally I think we should indeed go with `SvelteKit` instead of `Svelte`, but I am not sure if I can make the call all by myself... it's probably something I need to check with the team)

cc. @penalosa

jsparkdev:
@dario-piotrowicz 

Thank you! That's a lot of discussion!

So, do you think it's a good idea to close this PR?

dario-piotrowicz:
Hey @jsparkdev 🙂 

yeah closing this PR is what I was thinking but @penalosa's PR is a bit outdated now and I think they will not be available in the next few days... so might be worth it to keep going with this PR? 🤔 

Let me just check with the team, if there are no objections regarding the renaming I'd go forward with this one maybe 🤔

jsparkdev:
@dario-piotrowicz Thanks for the kind explanation. I'll keep this PR then :)

IgorMinar (approved):
lgtm

PR #7121, opened 2024-10-29
Added pause and resume commands to manage Workflows

bruxodasilva:
Fixes WOR-313

Added pause and resume commands to manage Workflows

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/17855


packages/wrangler/src/workflows/commands/instances/pause.ts
```diff
@@ -0,0 +1,68 @@
...
+	positionalArgs: ["name", "id"],
+	args: {
+		name: {
+			describe: "Name of the workflow",
+			type: "string",
+			demandOption: true,
+		},
+		id: {
+			describe:
+				"ID of the instance - instead of an UUID you can type 'latest' to get the latest instance and describe it",
```
emily-shen, line 25:
```suggestion
				"ID of the instance - instead of an UUID you can type 'latest' to get the latest instance and pause it",
```

packages/wrangler/src/workflows/commands/instances/resume.ts
```diff
@@ -0,0 +1,68 @@
...
+	positionalArgs: ["name", "id"],
+	args: {
+		name: {
+			describe: "Name of the workflow",
+			type: "string",
+			demandOption: true,
+		},
+		id: {
+			describe:
+				"ID of the instance - instead of an UUID you can type 'latest' to get the latest instance and describe it",
```
emily-shen, line 25:
```suggestion
				"ID of the instance - instead of an UUID you can type 'latest' to get the latest instance and resume it",
```

emily-shen:
Hey, since there isn't a hard deadline anymore for workflows, new PRs should have tests, even if this is still a beta feature -can you add tests for this PR? Let me know if you've got any issues with that, or if there's anything i can do to help :)

packages/wrangler/src/workflows/commands/delete.ts
```diff
@@ -21,13 +19,8 @@ defineCommand({
...
-		const accountId = await requireAuth(config);
-
-		await fetchResult(`/accounts/${accountId}/workflows/${args.name}`, {
-			method: "DELETE",
-		});
-
-		logger.info(`Workflow "${args.name}" was successfully removed`);
+	async handler(args) {
+		logger.info("🚫 delete command not yet implement");
+		logger.log(`🚫 Workflow "${args.name}" NOT removed`);
```
Skye-31, line 24: Let's not do this - our API should return a good error to the user that deletion is yet to be supported, so that when we implement it, this doesn't need to be reversed.

emily-shen: 😱 - we should just comment out this command for now, so no one can try and use it.

bruxodasilva: The feedback I got from my team was "The command is already public, let's not remove it. Just show an error message." 😅

bruxodasilva: Which one do you prefer?
1. Implement the API, throw an error on the API and gracefully manage it in Wrangler?
2. Delete the command from Wrangler all together until we have the API and functionality properly implemented?

emily-shen: Even if we don't remove it from wrangler, I think we should hide it so it doesn't show up on the help message etc.

LuisDuarte1: > Even if we don't remove it, I think we should hide it so it doesn't show up on the help message etc.

I can live with this, I think it's fine to leave the command because removing would semantically suggest a breaking change in the cli

bruxodasilva: I go for 2

Skye-31: Hiding is fine, but can the other change be reverted & we add a better error to the API in the meantime?

.changeset/violet-zoos-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Added pause and resume commands to manage Workflows
```
emily-shen, line 5: can you add a note here about hiding the unimplemented delete command?

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -0,0 +1,434 @@
...
+							result: workflows,
+						});
+					},
+					{ once: true }
+				)
+			);
+		};
+
+		it("should get the list of workflows", async () => {
+			writeWranglerToml();
```
emily-shen, line 151: there's no reason for people to set up a workflow without configuring a binding in wrangler.toml right?  should we be erroring (or warning) if people are running workflows commands without workflows configured? (not necessarily for this PR, but just checking)

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -0,0 +1,435 @@
...
+				status: "running",
+			},
+		];
+
+		it("should get the list of instances given a name", async () => {
+			writeWranglerToml();
+			await mockGetInstances(mockInstances);
+
+			await runWrangler(`workflows instances list some-workflow`);
+			console.log(std);
```
emily-shen, line 195:
```suggestion

```

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -0,0 +1,434 @@
...
```
emily-shen, line 195:
```suggestion

```

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -0,0 +1,434 @@
...
+├─────┼─────────┼───────────────────────┼───────────────────────┼───────────┤
+│ foo │ c       │ 1/1/2021, 12:00:00 AM │ 1/1/2021, 12:00:00 AM │ ▶ Running │
+└─────┴─────────┴───────────────────────┴───────────────────────┴───────────┘
+{
+  debug: [Getter],
+  out: [Getter],
+  info: [Getter],
+  err: [Getter],
+  warn: [Getter]
+}"
```
emily-shen, line 214:
```suggestion
"
```

emily-shen (approved):
Approved pending changes.

packages/wrangler/src/__tests__/workflows.test.ts
bruxodasilva: Yep, there's no reason for people to set up a workflow without configuring a binding in wrangler.toml.

packages/wrangler/src/workflows/commands/delete.ts
bruxodasilva: Once we update the API, we will add an improved error message to the it and update the Wrangler command accordingly.

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -0,0 +1,426 @@
...
+			);
+		};
+
+		it("should trigger a workflow given a name", async () => {
+			writeWranglerToml();
+			await mockTriggerWorkflow();
+
+			await runWrangler(`workflows trigger some-workflow`);
+			expect(std.info).toMatchInlineSnapshot(
+				`"🚀 Workflow instance \\"undefined\\" has been queued successfully"`
```
edmundhung, line 408: Can we provide a mock response result so the log message will reflect how it uses the response result (i.e. `response.id`) instead of printing `undefined` here?

edmundhung (approved):
Thank you!

Issue #6913, opened 2024-10-07
🐛 BUG: recently added wrangler retries doesn't seem to help with flakiness


jiri-prokop-pb:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What version(s) of the tool(s) are you using?

3.79.0 [Wrangler]

### What version of Node are you using?

20.16.0

### What operating system and version are you using?

macOS Sequoia 15.0 (24A335) & Ubuntu Jammy (22.04.5 LTS) on CI

### Observed behavior
We had retries on CI level for `wrangler`-related tasks for a while but we recently noticed that "native" retries were added in v3.79.0 with https://github.com/cloudflare/workers-sdk/pull/6801. We jumped in, upgraded to the supported version and removed our CI retry-logic only to find out that our CI is again flaky due to `wrangler`-related errors.

We reverted back to our CI retry logic but it's not ideal. We would welcome this to work as it will simplify our configuration.

### Expected behavior
Flakiness will be low with "native"/built-in retries. Ideally we won't get random failures related to `wrangler` at all if everything is correct on our side.

It would be also great to have some logs for the retries. Right now it's hard to guess if the logic actually does anything. On our side it just looks like it ran once and failed immediately.

### Steps to reproduce
N/A

It's random error that happens semi-regularly on our CI if we don't do retries on our side.

### Please provide any relevant error logs

```
  > wrangler deploy
  
  
   ⛅️ wrangler 3.79.0
  -------------------
  
  Total Upload: 296.15 KiB / gzip: 70.73 KiB
  Your worker has access to the following bindings:
  - Vars:
    - ENVIRONMENT: "dev"
  
  ✘ [ERROR] A request to the Cloudflare API (/accounts/xxxyyyzzz/workers/scripts/aaabbbccc/deployments) failed.
  
    workers.api.error.unknown [code: 10013]
```

jiri-prokop-pb:
@emily-shen thanks and ok, got it; my main problem was that it's not clear if there are actually any retries being done and I believe it would be helpful to see at least some basic info printed by default but anyway, we can try with `WRANGLER_LOG="debug"` as well.

Regarding account id, we don't want to share it publicly here so my colleague will contact you over your official support channel.

Do you have any ETA about when we can expect more stable deploys? Once it's done we could plan some code simplification on our side as right now we have some workarounds in place for stable CI.

lrapoport-cf:
[...]

PR #7122, opened 2024-10-29
Adds a HTTP header to indicate retried requests

GregBrimble:
Firms up the retry logic a bit, and adds a HTTP header (`X-Attempt-Count`) to retried requests for better tracking of when our APIs are flaking. We'll add instrumentation for this on the API side.

Callouts:

1. This introduces `AsyncLocalStorage` to this codebase! A bit controversial to introduce a context thing like this, I know! But I figured it was okay for this metrics-y use-case. I had considered a few other options:

  - prop drilling the attempt count from the caller all the way down into our cfFetch stuff.

      - rejected because this was insanely messy. To get all the way down to the `performApiFetch()`, this would be need to be plumbed through loads and loads of functions, which felt like overkill for this (quite rarely used) utility.

  -  a global variable and keying off of some fetch prop (e.g. URL and method).

      - felt like this would go wrong more often than right. Better to just let the client tell you when it's doing retries since it has all that logic for it already anyway.

2. I also had considered modifying the user agent rather than adding a new header, and although that would have made analysis quick and easy, figured it was more messy.

3. This is always-on, for everyone, unlike the error reporting. Figured there's no sensitive information here, so we're okay to just turn this on for everyone.

I'm very much not sold on this approach though. It's pretty quick and dirty, and we could probably piece this together behind-the-scenes without this at all. This was just a thing I could quickly add to make a start on tracking this problem. If someone objects, or if anyone has any better ideas for implementation, I'm all ears!

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: undocumented

### DIFF
--- a/.changeset/clean-brooms-exercise.md
+++ b/.changeset/clean-brooms-exercise.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+chore: Add instrumentation for retries
--- a/packages/wrangler/src/__tests__/utils/retry.test.ts
+++ b/packages/wrangler/src/__tests__/utils/retry.test.ts
@@ -1,0 +1,113 @@
+import { setTimeout } from "node:timers/promises";
+import {
+	alsAttemptCounter,
+	ExponentialBackoff,
+	LinearBackoff,
+	retryOnError,
+} from "../../utils/retry";
+
+const countTo = (n: number): number[] =>
+	Array.from({ length: n }, (_, i) => i + 1); // [1, 2, ..., n]
+
+describe("retryOnError", () => {
+	beforeEach(() => {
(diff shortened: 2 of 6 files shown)


lrapoport-cf:
reminder to follow up on https://github.com/cloudflare/workers-sdk/issues/6913 after this lands :)

GregBrimble:
Nah, actually, I'm backing out of this. I don't like it enough myself. Asset upload has just been made much much more resilient. I'm going to instrument these other APIs from the other end manually.

PR #4748, opened 2024-01-14
fix: add support for node-like module resolution

Cherry:
Fixes #4726

**What this PR solves / how to test:**

Similar to https://github.com/cloudflare/workers-sdk/pull/4135, this extends the module collection plugin to support standard node-like resolution of file paths.

This allows you to do things like:
```js
import JPEG_DEC_WASM from '@jsquash/jpeg/codec/dec/mozjpeg_dec.wasm';
```
even when the package doesn't explicitly export that file in its `exports` map.

This uses [`resolve`](https://www.npmjs.com/package/resolve), a package that was already in the dependency tree, and used throughout the JS ecosystem everywhere such as in `vite`, `jest`, `eslint-plugin-import`, etc.

**Author has addressed the following:**

- Tests
  - [x] Included
  - [ ] Not necessary because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [x] Included
  - [ ] Not necessary because:
- Associated docs
  - [ ] Issue(s)/PR(s):
  - [ ] Not necessary because:

**Note for PR author:**

We want to celebrate and highlight awesome PR review! If you think this PR received a particularly high-caliber review, please assign it the label `highlight pr review` so future reviewers can take inspiration and learn from it.

PR #4135, opened 2023-10-08
feat: support npm resolution for file imports

Cherry:
Fixes #4098.

**What this PR solves / how to test:**

Previously when importing static files (such as wasm) from an npm package, you had to path directly to it, such as :
```javascript
import wasm from '../../node_modules/svg2png-wasm/svg2png_wasm_bg.wasm'
```
While this works file, the DX isn't ideal, so this PR adds proper package resolution for these statements, allowing you to do:
```javascript
import wasm from 'svg2png-wasm/svg2png_wasm_bg.wasm'
```

This uses [`resolve.exports`](https://www.npmjs.com/package/resolve.exports) to look at and analyse the package's `exports` field, and resolve to the applicable file. This does add an additional dependency, but this is a very common one used by so much build tooling including `vite`, `jest`, `svelte`, and more.

If for any reason this fails, it falls back to the previous behaviour and will just do a path reference, and then end up throwing an `ENOENT` as per previous functionality.

Notes:

- My detection for "is this possibly an npm package" is very dumb right now, checking for presence of `/` and not starting with `.`. It seems to work remarkebly well in practice across the many projects I've tested this against, but I'm happy to improve this if someone has a cleaner idea.
- While I've added tests for the npm package name extraction to ensure it handles scoped packages, I've not added any specific tests for handling of the imports. To test that, it'd need a fixture with a real `node_modules` folder, and dummy package, which I don't think there's any prior art for in the repo today. Happy to add these if someone can guide me on the best way to approach this.

**Author has included the following, where applicable:**

- [x] Tests
- [x] Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))

**Reviewer is to perform the following, as applicable:**

- Checked for inclusion of relevant tests
- Checked for inclusion of a relevant changeset
- Checked for creation of associated docs updates
- Manually pulled down the changes and spot-tested

**Note for PR author:**

We want to celebrate and highlight awesome PR review! If you think this PR received a particularly high-caliber review, please assign it the label `highlight pr review` so future reviewers can take inspiration and learn from it.

PR #7130, opened 2024-10-30
Fix wrangler module import under npm monorepos

### LINKED ISSUE #7129: 🐛 BUG: wrangler fails to resolve wasm module in npm monorepo

Wrangler

3.83.0

18.20.2

Ubuntu 24.04

### Observed behavior
Wrangler is only be able to resolve modules in a relative `node_modules` directory and not the workspace root `node_modules` directory. And will otherwise throw the following errors:

```plain
ENOENT: no such file or directory, open
  <an absolute path from the local node_modules>
The plugin "wrangler-module-collector" was triggered by this import
```

### Expected behavior
Wrangler resolution matches bundler resolution.

### Steps to reproduce

```bash
git clone -b intent https://github.com/nickbabcock/cloud-functions
npm install
npm run --workspace src/highwayhasher build
```

_No response_

_No response_

nickbabcock:
This PR updates import resolution for files and package exports

In an npm workspace environment, wrangler will now be able to successfully resolve package exports.

Previously, wrangler would only be able to resolve modules in a relative `node_modules` directory and not workspace root `node_modules` directory.

Installing locally fixes #7129

--- 

Soft ping to @Cherry who originally implemented this feature, some extra eyes would be appreciated 👀 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: passing on https://github.com/cloudflare/workers-sdk/pull/7283
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: as this implements the behavior implied by documentation.


Cherry:
Thanks for the ping!

Do the tests added for https://github.com/cloudflare/workers-sdk/pull/4135 and https://github.com/cloudflare/workers-sdk/pull/4748, with those use-cases, still work as expected here? I had issues getting `require.resolve` to play nicely unfortunately.

nickbabcock:
That was my concern too, but (unless I'm missing something) they pass locally when I execute:
```
pnpm run --filter import-wasm-example test:watch
```

nickbabcock:
Ah, I should have been executing the following to know if I introduced regressions:
```
pnpm run test:ci --concurrency 1 --filter=./fixtures/import-wasm-example
```
Turns out I did, and reintroducing `resolve` for non-exported members fixed the issue.

I still need to add a test before this PR is good to go.

nickbabcock:
Alright, added npm workspace fixture which demonstrates the PR fixing the bundling issue.

✅ PR now ready for further review.

packages/wrangler/src/deployment-bundle/module-collection.ts
```diff
@@ -284,46 +263,13 @@ export function createModuleCollector(props: {
...
-											);
-										}
-									} catch (e) {
-										// We tried, now it'll just fall-through to the previous behaviour
-										// and ENOENT if the absolute file path doesn't exist.
-									}
+								try {
+									filePath = require.resolve(args.path, {
+										paths: [args.resolveDir],
+									});
```
penalosa, line 269: I think this might be better using esbuild's resolution algorithm rather than `require.resolve()`? Something along the lines of:

```ts
const resolved = await build.resolve(args.path, {
  kind: "import-statement",
  resolveDir: args.resolveDir,
  pluginData: {
	  skip: true,
  },
  });
  if (resolved.path) {
  filePath = resolved.path;
  }
```

with a check for `pluginData.skip` at the top of the `onResolve()` callback to prevent infinite loops.

nickbabcock: Updated to use esbuild's plugin. All the tests seem to pass. Is there a particular reason why you prefer the esbuild plugin vs require.resolve?

penalosa (approved):

emily-shen:
thanks so much for the contribution!

PR #7141, opened 2024-10-31
fix: add better error for user binding called ASSETS in pages projects

### LINKED ISSUE #6266: 🐛 BUG: No error handling for failed deploy due to binding with the name 'assets'.

Pages

Wrangler 3.64.0

20.10.0

Windows 11

### Observed behavior
When deploying a Hono application in Wrangler, if a binding uses the name 'ASSETS' the application will fail to deploy.

### Expected behavior
The application failing to deploy is expected as assets is a reserved keyword, however no error message is provided to explain that this is the cause of the failed deployment. An error message that explains that a binding with the name assets exists and that the name needs to be changed is the expected behavior.

### Steps to reproduce
Please provide the following:
- Create a Hono application
- Add a binding with the name 'ASSETS'
- Attempt deploying the application

emily-shen:
Fixes #6266 
The binding name "ASSETS" is used internally by pages projects. If a user had a binding with this name (totally sensible binding name for an r2 bucket for example) it would fail cryptically (see linked issue). 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no changes to e2es
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: error message fix

### DIFF
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -2773,6 +2773,12 @@ const validateBindingsHaveUniqueNames = (
 				bindingsGroupedByName[bindingName] = [];
 			}
 
+			if (bindingName === "ASSETS" && isPagesConfig(config)) {
+				diagnostics.errors.push(
+					`The name 'ASSETS' is reserved in Pages projects. Please use a different name for your ${bindingType} binding.`
+				);
+			}
+
 			bindingsGroupedByName[bindingName].push(bindingType);
 		}
 	}
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -2773,7 +2773,7 @@ const validateBindingsHaveUniqueNames = (
 				bindingsGroupedByName[bindingName] = [];
 			}
 
-			if (bindingName.toUpperCase() === "ASSETS" && isPagesConfig(config)) {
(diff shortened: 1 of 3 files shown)


emily-shen:
a binding named ASSETS is allowed in workers projects though, and there are plenty of existing tests using that so that case should already be covered

packages/wrangler/src/config/validation.ts
```diff
@@ -2773,6 +2773,12 @@ const validateBindingsHaveUniqueNames = (
 				bindingsGroupedByName[bindingName] = [];
 			}
 
+			if (bindingName.toUpperCase() === "ASSETS" && isPagesConfig(config)) {
+				diagnostics.errors.push(
+					`The name 'ASSETS' is reserved in Pages projects. Please use a different name for your ${bindingType} binding.`
```
edmundhung, line 2778: Nit: Is binding name is case sensitive? Might be better to show the original name user provided.
```suggestion
					`The name '${bindingName}' is reserved in Pages projects. Please use a different name for your ${bindingType} binding.`
```

emily-shen: turns out it is case sensitive, so you totally can deploy with 'assets' (lowercase) no issues

PR #6655, opened 2024-09-09
Refactor all commands to use new registration utils

RamIdeas:
## What this PR solves / how to test

WIP
- [ ] wrangler docs
- [ ] wrangler init
- [ ] wrangler dev
- [ ] wrangler deploy
- [ ] wrangler deployments           
- [ ] wrangler rollback
- [ ] wrangler versions              
- [ ] wrangler triggers              
- [ ] wrangler delete
- [ ] wrangler tail
- [x] wrangler secret                
- [ ] wrangler types
- [x] wrangler kv                    
- [x] wrangler queues                
- [x] wrangler r2                    
- [x] wrangler d1                    
- [ ] wrangler vectorize             
- [x] wrangler hyperdrive            
- [ ] wrangler pages                 
- [ ] wrangler mtls-certificate      
- [ ] wrangler pubsub                
- [ ] wrangler dispatch-namespace    
- [x] wrangler ai                    
- [x] wrangler login                 
- [x] wrangler logout                
- [x] wrangler whoami       

Keeping note of things to fix:
- [ ] respect order of alias definitions (in help output)
- [ ] allow alias command description to be overridden with (`Hook`?) access to real definition metadata

## Author has addressed the following

- Tests
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [ ] Required / Maybe required
  - [ ] Not required because:
- Changeset ([Changeset guidelines](https://github.com/cloudflare/workers-sdk/blob/main/CONTRIBUTING.md#changesets))
  - [ ] TODO (before merge)
  - [ ] Included
  - [ ] Not necessary because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Not necessary because:

PR #7150, opened 2024-11-01
chore: refactor `wrangler login/logout/whoami` to use defineCommand

emily-shen:
Fixes #N/A
Landing #6655 in smaller chunks. This one does `wrangler login/logout/whoami` . There should be no functionality change.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/user/commands.ts
```diff
@@ -0,0 +1,101 @@
...
+});
+
+defineCommand({
+	command: "wrangler logout",
+	metadata: {
+		description: "🚪 Logout from Cloudflare",
+		owner: "Workers: Authoring and Testing",
+		status: "stable",
+	},
+	args: {},
```
penalosa, line 70: Can this be omitted entirely with no args?

packages/wrangler/src/user/commands.ts
```diff
@@ -0,0 +1,101 @@
...
+	async handler(args) {
+		await logout();
+		const config = readConfig(undefined, args);
```
penalosa, line 73: I know this is existing code, but this should probably be `readConfig(args.config, args)` as it is in `login`

packages/wrangler/src/user/commands.ts
```diff
@@ -0,0 +1,101 @@
...
+	args: {
+		account: {
+			type: "string",
+			describe:
+				"Show membership information for the given account (id or name).",
+		},
+	},
+	async handler(args) {
+		await whoami(args.account);
+		const config = readConfig(undefined, args);
```
penalosa, line 96: As above

penalosa (approved):
Couple small things, but otherwise ✅

packages/wrangler/src/core/define-command.ts
```diff
@@ -101,13 +101,20 @@ export type CommandDefinition<
 		 * @default true
 		 */
 		printBanner?: boolean;
+
+		/**
+		 * By default, wrangler will print warnings about wrangler.toml configuration.
+		 * Set this value to `false` to skip printing these warnings.
+		 */
+		printConfigWarnings?: boolean;
```
edmundhung, line 109: I was rebasing the changes on #6756 over to this PR and notice the same changes doesn't work here. The reason was that `defineCommand` will read the config for you and there was no way to disable the warning.  You can find the full changes that enable this option in https://github.com/cloudflare/workers-sdk/pull/7150/commits/ec6d4cd46074138990400971432afa90566229f6

cc: @penalosa @emily-shen

PR #7166, opened 2024-11-05
Add skipIfNoneMatchHeaderHandling parameter to asset-worker handler

GregBrimble:
Frameworks folks have asked for this API with the option to suppress sending back 304s. Alternatively, we could suppress sending ETag headers altogether, but that seems unnecessary?

@jamesopstad , this is just two seconds of work if you do need anything different though, so just shout. Sorry I took so long. I've been trying to change up the structure of this package so it's exported nicely for you, but I'll do that in a different PR so we can keep this one focused exclusively just on the change you actually requested. If you use it before that lands, I'd ask you pay particular attention to the fact that this package is 0.X.X. Please pin an exact version! We'll make changes here reasonably frequently as we work to make Workers Assets generally available.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: undocumented package


packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -0,0 +1,128 @@
...
+		const eTag = "some-etag";
+		const exists = vi.fn().mockReturnValue(eTag);
+		const getByETag = vi.fn().mockReturnValue({
+			readableStream: new ReadableStream(),
+			contentType: "text/html",
+		});
+
+		const response = await handleRequest(
+			new Request("https://example.com/", {
+				headers: { "If-None-Match": eTag },
```
jamesopstad, line 43: I think the `eTag` might need to be included in quotes here.

```suggestion
				headers: { "If-None-Match": `"${eTag}"` },
```

jamesopstad:
Thanks Greg! Made one small suggestion but otherwise looks good.

packages/workers-shared/asset-worker/tests/handler.test.ts
GregBrimble: Good catch, thank you!

jamesopstad:
I had a chat with @petebacondarwin about this PR and he suggested that we remove the `If-None-Match` header on our end so that adding the `skipIfNoneMatchHeaderHandling` boolean is not necessary. In that case this PR won't be necessary. Sorry to have created extra work.

We'll still need the Router Worker, Asset Worker and `handleRequest` to be exported from `@cloudflare/workers-shared` with types.

petebacondarwin (approved):
Tests! Love it.

PR #7172, opened 2024-11-05
Cleanup dev registry messaging

penalosa:
Cleans up & clarifies dev registry messaging (mostly in response to https://github.com/cloudflare/workers-sdk/issues/7009#issuecomment-2427012598) cc @isaac-mcfadyen 

- Shift from inscrutable red & green dots to `connected`/`not connected` with an explanation logged the first time
Screenshot 2024-11-05 at 13 50 49

- Remove the "Live workers" warning in local mode, where it doesn't apply
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/17994

### DIFF
--- a/packages/wrangler/src/config/index.ts
+++ b/packages/wrangler/src/config/index.ts
@@ -306,9 +306,9 @@ export function printBindings(
 									(d) => d.className === class_name
 								)
 							) {
-								value += ` (defined in ${script_name} ${chalk.green("connected")})`;
+								value += ` (defined in ${script_name} ${chalk.green("[connected]")})`;
 							} else {
-								value += ` (defined in ${script_name} ${chalk.red("not connected")})`;
+								value += ` (defined in ${script_name} ${chalk.red("[not connected]")})`;
 							}
 						} else {
 							value += ` (defined in ${script_name})`;
@@ -473,9 +473,9 @@ export function printBindings(
 						(!entrypoint ||
 							registryDefinition.entrypointAddresses?.[entrypoint])
 					) {
-						value = value + " " + chalk.green("connected");
+						value = value + " " + chalk.green("[connected]");
 					} else {
-						value = value + " " + chalk.red("not connected");
+						value = value + " " + chalk.red("[not connected]");
 					}
 				}
 				return {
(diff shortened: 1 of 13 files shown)


packages/wrangler/src/config/index.ts
```diff
@@ -236,6 +237,7 @@ export function printBindings(
 		local?: boolean;
 	} = {}
 ) {
+	let someConnectionStatus = false;
```
CarmenPopoviciu, line 240: nit?
```suggestion
	let hasConnectionStatus = false;
```

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -241,7 +241,7 @@ export default async function triggersDeploy(
 	}
 
 	if (config.workflows?.length) {
-		logger.warnOnce("Workflows is currently in open beta.");
+		logger.once.warn("Workflows is currently in open beta.");
```
CarmenPopoviciu, line 244: do we need to dbl check this change with the Workflows team?

edmundhung:
Why do we need the concept on connected here? Shouldn't we just make it an error if it is not connected?

packages/wrangler/src/triggers/deploy.ts
penalosa: No—this is the same behaviour, it's just a change in how the `warnOnce()` function is defined

penalosa:
> Why do we need the concept of connected here? Shouldn't we just make it an error if it is not connected?

I'm not sure I understand what you mean here—what should be an error?

packages/wrangler/src/config/index.ts
```diff
@@ -463,14 +466,16 @@ export function printBindings(
...
 					const registryDefinition = context.registry?.[service];
+					hasConnectionStatus = true;
+
 					if (
 						registryDefinition &&
 						(!entrypoint ||
 							registryDefinition.entrypointAddresses?.[entrypoint])
 					) {
-						value = `🟢 ` + value;
+						value = value + " " + chalk.green("connected");
```
CarmenPopoviciu, line 476: nit: do you think smth like 👇 would better differentiate between the binding and its status?
```suggestion
						value = value + " " + chalk.green("[connected]");
```

CarmenPopoviciu (approved):

isaac-mcfadyen:
> Cleans up & clarifies dev registry messaging (mostly in response to [#7009 (comment)](https://github.com/cloudflare/workers-sdk/issues/7009#issuecomment-2427012598)) cc @isaac-mcfadyen

Awesome, thank you @penalosa! 😄

packages/wrangler/src/config/index.ts
isaac-mcfadyen: +1 on this especially in scenarios when terminal colors are turned off (CI, certain restricted SSH connections or older terminals, etc)

PR #7173, opened 2024-11-05
WO-250 add workers-observability logs configuration

Ankcorn:
Fixes #[WO-250].

Adds a nested logs configuration 
---

Adds nested logs configuration to workers observability, this lets us add extra settings just for logs like toggling invocation logs.
```toml
[observability.logs]
enabled = true
invocation_logs = false
```

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18032


packages/wrangler/src/config/validation-helpers.ts
```diff
@@ -434,6 +434,44 @@ export const validateRequiredProperty = (
...
+export const validateAtLeastOnePropertyRequired = (
+	diagnostics: Diagnostics,
+	container: string,
+	properties: {
+		key: string;
+		value: unknown;
+		type: TypeofType;
+	}[]
+): boolean => {
+	if (container) {
```
andyjessop, line 449: We'd like to generally avoid mutating function arguments as it makes changes more difficult to track and test. Could this be refactored so that the `container` variable used inside this function has a different name? e.g.

```
const containerPath = container ? `${container}.` : '';
```

andyjessop (approved):
This looks good, just one minor comment, but I won't block merging.

PR #7176, opened 2024-11-05
Remove incorrect logic in Asset Worker

jamesopstad:
When I was looking through the Asset Worker code I spotted an error:

https://github.com/cloudflare/workers-sdk/blob/afe9850931b7730989e005a612ea74887f5d5e72/packages/workers-shared/asset-worker/src/index.ts#L132-L135

This currently does the opposite of what it should and returns the `MethodNotAllowedResponse` for `GET` and `HEAD` methods. The only reason that the Router Worker is still functioning correctly is because it treats the `MethodNotAllowedResponse` as truthy.

https://github.com/cloudflare/workers-sdk/blob/afe9850931b7730989e005a612ea74887f5d5e72/packages/workers-shared/router-worker/src/index.ts#L60

I've removed method handling from `unstable_canFetch` altogether as I feel that it should only return a boolean. The Asset Worker already returns a `MethodNotAllowedResponse` for methods other than `GET` and `HEAD`.

https://github.com/cloudflare/workers-sdk/blob/afe9850931b7730989e005a612ea74887f5d5e72/packages/workers-shared/asset-worker/src/handler.ts#L32-L34

If this also needs to occur earlier then the request method could be checked in the Router Worker.

I haven't added tests as there aren't any others for the Asset Worker beyond the KV store at present.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: not currently implemented
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/workers-shared/asset-worker/src/index.ts
+++ b/packages/workers-shared/asset-worker/src/index.ts
@@ -117,9 +117,8 @@ export default class extends WorkerEntrypoint<Env> {
 		}
 	}
 
-	async unstable_canFetch(request: Request): Promise<boolean | Response> {
+	async unstable_canFetch(request: Request): Promise<boolean> {
 		const url = new URL(request.url);
-		const method = request.method.toUpperCase();
 		const decodedPathname = decodePath(url.pathname);
 		const intent = await getIntent(
 			decodedPathname,
@@ -129,10 +128,6 @@ export default class extends WorkerEntrypoint<Env> {
 			},
 			this.unstable_exists.bind(this)
 		);
-		// if asset exists but non GET/HEAD method, 405
-		if (intent && ["GET", "HEAD"].includes(method)) {
-			return new MethodNotAllowedResponse();
-		}
 		if (intent === null) {
 			return false;
 		}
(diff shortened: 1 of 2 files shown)


packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -129,10 +125,6 @@ export default class extends WorkerEntrypoint<Env> {
 			},
 			this.unstable_exists.bind(this)
 		);
-		// if asset exists but non GET/HEAD method, 405
```
WillTaylorDev, line 132: We might still want this check in router worker, since we would rather return the user worker instead of a 405. I'm totally in agreance this makes sense as just a bool response!

petebacondarwin (approved):
I agree that this makes sense as a boolean; and also that we need to change Router Worker to handle the 405 case.
Also, we could do with more tests on this worker. Currently there are scant. But perhaps unfair on @jamesopstad to be the one to write those?

emily-shen:
There are some tests scattered across the fixtures  - that should probably be documented somewhere whoops. There actually are some tests for the 405 (in `worker-with-assets` I think) but that was clearly a bit too e2e… there’s also some unit tests in `asset-config` that could be built on.

GregBrimble (approved):

CarmenPopoviciu:
@jamesopstad can you please include a changeset to your PR 🙏 ?

packages/workers-shared/asset-worker/src/index.ts
GregBrimble: Can't change that behavior now without a compat flag.

CarmenPopoviciu (approved):

PR #7193, opened 2024-11-07
[R2] Output suggested wrangler.toml changes after creating a bucket

sdnts:
Fixes R2-000

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: this is an output-only change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this is an output-only change

### DIFF
--- a/packages/wrangler/src/r2/create.ts
+++ b/packages/wrangler/src/r2/create.ts
@@ -4,6 +4,7 @@ import { UserError } from "../errors";
 import { logger } from "../logger";
 import * as metrics from "../metrics";
 import { requireAuth } from "../user";
+import { getValidBindingName } from "../utils/getValidBindingName";
 import { LOCATION_CHOICES } from "./constants";
 import { createR2Bucket, isValidR2BucketName } from "./helpers";
 import type {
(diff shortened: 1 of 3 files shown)


toddmantell (approved):

packages/wrangler/src/r2/create.ts
```diff
@@ -68,8 +69,14 @@ export async function Handler(args: HandlerOptions) {
 	logger.log(
 		`✅ Created bucket '${fullBucketName}' with${
 			location ? ` location hint ${location} and` : ``
-		} default storage class of ${storageClass ? storageClass : `Standard`}.`
+		} default storage class of ${storageClass ? storageClass : `Standard`}.\n`
 	);
+
+	logger.log("Configure your Worker to write objects to this bucket:\n");
```
CarmenPopoviciu, line 75: unless there is a specific reason to have these as separate calls, I'd suggest 👇 

```suggestion
	logger.log(
	  "Configure your Worker to write objects to this bucket:\n" +
	  "[[r2_buckets]]\n" +
	  `bucket_name = "${args.name}"\n` +
	  `binding = "${getValidBindingName(args.name, "r2")}"`
	);
```

sdnts: Done!

CarmenPopoviciu (approved):

CarmenPopoviciu (approved):

PR #7197, opened 2024-11-07
[D1] fix console output for `wrangler d1 migrations create`

michelheusschen:
Fixes wrong console output after using `wrangler d1 migrations create`

Before: /home/user/path/{newMigrationName}
After: /home/user/path/0001_migration.sql

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no existing tests, just log output so not worth adding
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no existing e2e tests print this log message I think
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: messaging change

### DIFF
--- a/.changeset/new-adults-float.md
+++ b/.changeset/new-adults-float.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix console output for `wrangler d1 migrations create`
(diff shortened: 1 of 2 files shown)


emily-shen (changes requested):
Actually, can you add a changeset please? you can run `npx changeset` to generate one. This should be a patch. Thanks again!

michelheusschen:
@emily-shen I've added a changeset

emily-shen (approved):

PR #7203, opened 2024-11-08
fix(wrangler): fix failing R2 e2e test

CarmenPopoviciu:
Fixes n/a.

https://github.com/cloudflare/workers-sdk/commit/ad51d1d77483bf0b4dc73fd392f5cdefe4ddf5d8 broke a related R2 e2e test, which we didn't see due to the fact that e2es were not run on that PR. This PR fixes the breakage.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: e2e test fix
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: e2e test fix

### DIFF
--- a/packages/wrangler/e2e/helpers/normalize.ts
+++ b/packages/wrangler/e2e/helpers/normalize.ts
@@ -11,6 +11,7 @@ export function normalizeOutput(
 		removeWorkersDev,
 		removeWorkerPreviewUrl,
 		removeUUID,
+		removeBinding,
 		normalizeErrorMarkers,
 		replaceByte,
 		stripTrailingWhitespace,
@@ -62,10 +63,17 @@ function removeTimestamp(str: string) {
 		.replace(/\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d\.\d+?Z/g, "TIMESTAMP")
 		.replace(/\d\d:\d\d:\d\d/g, "TIMESTAMP");
 }
+
 function removeUUID(str: string) {
 	return str.replace(
 		/\w{8}-\w{4}-\w{4}-\w{4}-\w{12}/g,
 		"00000000-0000-0000-0000-000000000000"
+	);
+}
+function removeBinding(str: string) {
+	return str.replace(
+		/\w{8}_\w{4}_\w{4}_\w{4}_\w{12}/g,
+		"00000000_0000_0000_0000_000000000000"
 	);
 }
 
--- a/packages/wrangler/e2e/r2.test.ts
+++ b/packages/wrangler/e2e/r2.test.ts
@@ -21,7 +21,11 @@ describe("r2", () => {
 
 		expect(normalize(output.stdout)).toMatchInlineSnapshot(`
 			"Creating bucket 'tmp-e2e-r2-00000000-0000-0000-0000-000000000000'...
-			✅ Created bucket 'tmp-e2e-r2-00000000-0000-0000-0000-000000000000' with default storage class of Standard."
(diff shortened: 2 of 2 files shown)


packages/wrangler/e2e/helpers/normalize.ts
```diff
@@ -62,12 +63,19 @@ function removeTimestamp(str: string) {
...
 		.replace(/\d\d:\d\d:\d\d/g, "TIMESTAMP");
 }
+
 function removeUUID(str: string) {
 	return str.replace(
 		/\w{8}-\w{4}-\w{4}-\w{4}-\w{12}/g,
 		"00000000-0000-0000-0000-000000000000"
 	);
 }
+function removeBinding(str: string) {
```
CarmenPopoviciu, line 73: I'd like some sanity check on this fn pls 🙏

CarmenPopoviciu: here is how we create that binding name: https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/utils/getValidBindingName.ts

CarmenPopoviciu: the UUID contains dashes `-`, whereas the binding name contains underscores `_`, so `removeUUID` won't have any effect on the binding name :(

CarmenPopoviciu: looking at https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/utils/getValidBindingName.ts#L11 and https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/utils/getValidBindingName.ts#L13 particularly, I wonder if my implementation is actually correct, since my assumption on the number of caracters separated by dashes might not be correct

penalosa (approved):

edmundhung (approved):

packages/wrangler/e2e/helpers/normalize.ts
```diff
@@ -62,12 +63,19 @@ function removeTimestamp(str: string) {
...
+	return str.replace(
+		/\w{8}_\w{4}_\w{4}_\w{4}_\w{12}/g,
+		"00000000-0000-0000-0000-000000000000"
```
petebacondarwin, line 76:
```suggestion
		"00000000_0000_0000_0000_000000000000"
```

packages/wrangler/e2e/helpers/normalize.ts
petebacondarwin: I think the point is that R2 is actually using a UUID in their name but we are converting it to a valid JS identifier, so it does have the same number of characters as a UUID, but the dashes are converted to underscores.

And also this is just for tests, so not too deep; if it breaks in the future we can just fix it again.

petebacondarwin (approved):

packages/wrangler/e2e/helpers/normalize.ts
petebacondarwin, line 1: Thanks for landing my suggestion, but I think that the e2e tests will now break again 😞 and need to be updated too

packages/wrangler/e2e/helpers/normalize.ts
CarmenPopoviciu: I did double check and this is indeed used only in the e2e tests. Thank you for reviewing Pete <3

packages/wrangler/e2e/helpers/normalize.ts
CarmenPopoviciu: I had those fixed locally and we should be good now

PR #7210, opened 2024-11-09
improve(wrangler): fix subdomain tests

taylorlee:
Many tests had mocks that weren't called, or had false assertions about not updating subdomain routes when they actually were being updated.

Additionally, available_on_subdomain is now unecessary during script-upload, as it is not read there, due to being handled entirely within triggers-deploy. This also saves a double call to get /subdomain in the new-api path.

The get/update subdomain helpers are refactored from deploy/version tests, and unified to standardize default workername and legacyEnv.

I wanted to remove the getSubdomain mock from mockUpload entirely, but there were a bit too many callsites... At least every test now that makes assertions about workers.dev deployments does explicitly reset the mock return value.

Fixes N/A

- Tests
  - [x] Tests included
  - [x] Not required because: this pr fixes unit-tests & mocks
- Public documentation

### DIFF
--- a/packages/wrangler/src/__tests__/helpers/mock-upload-worker.ts
+++ b/packages/wrangler/src/__tests__/helpers/mock-upload-worker.ts
@@ -1,4 +1,5 @@
 import { http, HttpResponse } from "msw";
+import { mockGetWorkerSubdomain } from "./mock-workers-subdomain";
 import { createFetchResult, msw } from "./msw";
 import { serialize, toString } from "./serialize-form-data-entry";
 import type { WorkerMetadata } from "../../deployment-bundle/create-worker-upload-form";
@@ -10,7 +11,6 @@ import type { HttpResponseResolver } from "msw";
 /** Create a mock handler for the request to upload a worker script. */
 export function mockUploadWorkerRequest(
 	options: {
-		available_on_subdomain?: boolean;
 		expectedEntry?: string | RegExp | ((entry: string | null) => void);
 		expectedMainModule?: string;
 		expectedType?: "esm" | "sw" | "none";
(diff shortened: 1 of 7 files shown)


WalshyDev (approved):

packages/wrangler/src/__tests__/helpers/mock-upload-worker.ts
```diff
@@ -227,25 +222,11 @@ export function mockUploadWorkerRequest(
...
-					expect(params.envName).toEqual(env);
-				}
-
-				return HttpResponse.json(
-					createFetchResult({ enabled: available_on_subdomain })
-				);
-			}
-		)
-	);
+	// TODO make explicit by callers?
```
CarmenPopoviciu, line 225: TODO in this PR? Otherwise should we track this somewhere?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -885,7 +877,6 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 								headers: await getMetricsUsageHeaders(config.send_metrics),
 							},
 							new URLSearchParams({
-								include_subdomain_availability: "true",
```
CarmenPopoviciu, line 888: should we remove https://github.com/cloudflare/workers-sdk/blob/481445571735978b6af036254a82b3b9ca73f161/packages/wrangler/src/versions/secrets/index.ts#L219 as well?

WalshyDev: I'll not remove for now just to keep the scope a bit smaller but yeah, it isn't needed here either

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -870,7 +863,6 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 				} else {
 					result = await retryOnError(async () =>
 						fetchResult<{
-							available_on_subdomain: boolean;
```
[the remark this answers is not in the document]
WalshyDev: Yeah for scoping I think good to only remove in this path right now. Deploy is what was using it when it wasn't needed and what was cleaned up. Will make a note to cleanup the rest though

PR #7214, opened 2024-11-10
Check if NaN before incrementing highestMigrationNumber

lloydhumphreys:
I ran into the following issue. If your first D1 migration is one you create and name manually with a single word (for example `init.sql`), then Wrangler will name all subsequent migrations `0NaN-...` because there are no `_` characters to split on so it'll take the word as a number.

I added a check for if the value is a number, and return 0 if not. So it'll start incrementing from 0 for new migrations created via wrangler.

---

- Tests
  - Not included. I wasn't sure they were necessary - if they are, I'll figure out how to do it.
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: I think people would presume it works like this already

### DIFF
--- a/.changeset/afraid-singers-pretend.md
+++ b/.changeset/afraid-singers-pretend.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+fix: number d1 migrations properly even if previous migrations aren't prefixed by a number
+
+We expected d1 migration names to be "0001_migration-name.sql", but a user could manually create a migration that in this format like "init.sql". Subsequent migrations would be named "0NaN_migration-name.sql" - this fixes that bug.
--- a/packages/wrangler/src/__tests__/d1/migrate.test.ts
+++ b/packages/wrangler/src/__tests__/d1/migrate.test.ts
@@ -1,3 +1,4 @@
+import * as fs from "fs";
 import { cwd } from "process";
 import { http, HttpResponse } from "msw";
 import { reinitialiseAuthTokens } from "../../user";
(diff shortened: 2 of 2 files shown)


emily-shen:
Hiya, thanks for putting up this PR! lgtm, just a few things to get this in :)
1. could you add a quick regression test? let me know if you run into any issues. Probably here - https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/__tests__/d1/migrate.test.ts 
2. there are some formatting errors, should go away if you run `pnpm fix`
3. can you add a changeset? should be a patch, and you can generate one by running `npx changeset`
thanks again for the contribution :)

emily-shen (changes requested):
Hiya, thanks for putting up this PR! lgtm, just a few things to get this in :)
1. could you add a quick regression test? let me know if you run into any issues. Probably here - https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/__tests__/d1/migrate.test.ts 
2. there are some formatting errors, should go away if you run `pnpm fix`
3. can you add a changeset? should be a patch, and you can generate one by running `npx changeset`
thanks again for the contribution :)

emily-shen (approved):

andyjessop (approved):

penalosa:
I'm going to close this for now, since this has gotten pretty stale. After further discussion, this solution may have an issue with applying migrations on a fresh database, since manually named migrations (like `init.sql`) may not necessarily run in the correct order before numbered migrations.

PR #7220, opened 2024-11-11
chore: defineCommand refactor for `wrangler types` and `tail`

emily-shen:
moving another couple of commands to defineCommand
related #6655 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/tail/index.ts
```diff
@@ -17,225 +17,233 @@ import {
...
+	behaviour: {
+		printBanner: false,
+	},
+	async handler(args, { config }) {
+		if (!process.stdout.isTTY) {
+			args.format = "json";
+		}
+		if (args.format === "pretty") {
+			await printWranglerBanner();
+		}
```
edmundhung, line 100: Might be worth standardizing this behavior in the future?

packages/wrangler/src/type-generation/index.ts
```diff
@@ -2,150 +2,141 @@ import * as fs from "node:fs";
...
-	if (!outputPath.endsWith(".d.ts")) {
-		throw new CommandLineArgsError(
-			`The provided path value ("${outputPath}") does not point to a declaration file (please use the 'd.ts' extension)`
-		);
-	}
+		if (!outputPath.endsWith(".d.ts")) {
+			throw new CommandLineArgsError(
+				`The provided path value ("${outputPath}") does not point to a declaration file (please use the 'd.ts' extension)`
+			);
+		}
```
edmundhung, line 61: The difference you see on the tests seems to be caused by us calling `readConfig` before the handler. Move these logic to `validateArgs` might fix it

emily-shen: ugh it isn't failing locally, that's super weird - but good point

packages/wrangler/src/tail/index.ts
emily-shen: as in make `format=pretty` a global arg?

edmundhung: I thought this is about the wrangler type command and I wasn't sure why we wanna hide the banner. But I guess it makes sense only in this particular case 🤔

packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -246,22 +246,22 @@ describe("generateTypes()", () => {
...
-		expect(std.warn).toMatchInlineSnapshot(`
-		"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mNo config file detected (at hello.toml), aborting[0m
-
-		"
-	`);
+	it("should error when no custom config file is detected", async () => {
+		await expect(
+			runWrangler("types -c hello.toml")
+		).rejects.toThrowErrorMatchingInlineSnapshot(
+			`[ParseError: Could not read file: hello.toml]`
```
penalosa, line 263: This seems like a slightly worse error message—is there any way we can improve it?

PR #7225, opened 2024-11-11
Fix workflows binding to create a workflow without arguments & miss match instance.id between wrangler local dev and production

### LINKED ISSUE #7135: Fix getting error on create
https://github.com/cloudflare/workers-sdk/blob/74ede561863feaf60e3c531c7602a1ca7e5c4e5d/packages/workflows-shared/src/binding.ts#L17

bruxodasilva:
Fixes WOR-317. Fixes #7135 

With this PR, 2 bugs were fixed:
* Fix workflows binding to create a workflow without arguments. If no argument were passed, the wrangler local dev would fail. Now we validate that the arguments exist, and if not, manage the creation of the necessary variables.
* `instance.id` is now available in wrangler local dev synchronously, as in production, which makes it possible to now run the sample workflows starter project

Co-authored-by: Sidhartha Chatterjee <me@sidharthachatterjee.com>

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: workflows-shared is managed by the workflows team
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s):
  - [x] Documentation not necessary because: no modifications to the feature usage


packages/workflows-shared/src/binding.ts
```diff
@@ -13,13 +13,12 @@ type Env = {
...
-		params,
-	}: WorkflowInstanceCreateOptions): Promise<WorkflowInstance> {
-		if (!id) {
-			id = crypto.randomUUID();
-		}
+	public async create(
+		options: WorkflowInstanceCreateOptions
+	): Promise<WorkflowInstance> {
+		const id = options?.id ? options?.id : crypto.randomUUID();
+		const params = options?.params ? options?.params : {};
```
penalosa, line 20: Any reason this can't be:

```suggestion
	public async create(
		{ id = crypto.randomUUID(), params = {}}: WorkflowInstanceCreateOptions
	): Promise<WorkflowInstance> {
```

packages/workflows-shared/src/binding.ts
```diff
@@ -14,12 +14,9 @@ type Env = {
 // this.env.WORKFLOW is WorkflowBinding
 export class WorkflowBinding extends WorkerEntrypoint<Env> implements Workflow {
 	public async create({
-		id,
-		params,
+		id = crypto.randomUUID(),
+		params = {},
 	}: WorkflowInstanceCreateOptions): Promise<WorkflowInstance> {
```
Skye-31, line 19:
```suggestion
		id = crypto.randomUUID(),
		params = {},
	}: WorkflowInstanceCreateOptions = {}): Promise<WorkflowInstance> {
```

I'm pretty sure we're still missing this?

bruxodasilva: well catched!

Skye-31 (commented):
I'm a little confused on what the second change around the id is for

bruxodasilva:
> I'm a little confused on what the second change around the id is for - could you clarify? 🙂

I am sorry for the noise. I identified a miss match between wrangler local dev and production, regarding the way you could access instance.id.
This commits were me trying to debug, until I learned how to do local builds with fixtures.
This last commit address this issue.

Skye-31 (approved):
Can see now how this makes the behaviour match better. Thanks

bruxodasilva:
@penalosa can you take a look at this? This PR is really improve for our customers, as currently wrangler local dev is not working properly.

packages/workflows-shared/src/binding.ts
```diff
@@ -34,22 +31,47 @@ export class WorkflowBinding extends WorkerEntrypoint<Env> implements Workflow {
...
-		return new WorkflowHandle(id, stub);
+		const handle = new WorkflowHandle(id, stub);
+		return {
+			id: id,
+			pause: handle.pause.bind(handle),
+			resume: handle.resume.bind(handle),
+			terminate: handle.terminate.bind(handle),
+			restart: handle.restart.bind(handle),
+			status: handle.status.bind(handle),
+		};
```
penalosa, line 42: Why this change? Is `WorkflowHandle` not an `RpcStub`?

packages/workflows-shared/src/binding.ts
```diff
@@ -34,22 +31,47 @@ export class WorkflowBinding extends WorkerEntrypoint<Env> implements Workflow {
...
-	constructor(
-		public id: string,
-		private stub: DurableObjectStub<Engine>
-	) {
+	public id: string;
+	private stub: DurableObjectStub<Engine>;
+	constructor(id: string, stub: DurableObjectStub<Engine>) {
 		super();
+		this.id = id;
+		this.stub = stub;
```
penalosa, line 74: Why this change? This seems like equivalent code to me

packages/workflows-shared/src/binding.ts
bruxodasilva: The usage of a `RcpStub` make it mandatory to `await` every parameter get,  (such as `instance.id`), which is not the same as in Workflows production deployment. This change allow us to not require the `await` on the gets, allowing customers to run the exact same code on wrangler local dev and Workflows production.
Notice that this `binding.ts` logic is an emulation of what we have in Workflows production environment.

packages/workflows-shared/src/binding.ts
penalosa: I'd lean towards re-introducing it if possible for readability—if it's causing bugs we should investigate

penalosa:
@bruxodasilva could you add a regression test to the workflows fixture test (`fixtures/workflow`)?

packages/workflows-shared/src/binding.ts
sidharthachatterjee: In production, we create a Object in workerd that returns a handle. We started out with an `RpcTarget` here but @bruxodasilva found a bug in that locally that implies every property access is async (which is not true in production). This emulates that without the other services in production.

sidharthachatterjee:
Looks good. Thank you!

penalosa (approved):

PR #7229, opened 2024-11-11
cloudchamber: Start `cloudchamber apply`, the new command to deploy

gabivlj:
container app changes

This command is able to take the [[container-app]] configurations, and deploy them to Cloudchamber.

To render the differences, we are introducing a new dependency with "diff". This was already included in the pnpm lock, however we could consider not rolling out a new dependency into wrangler unless absolutely necessary.

The command is designed to be CI friendly. In the tests there is some example command renders from different kind of configurations.

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: No e2e test for Cloudchamber yet.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: product is private only for certain users

### DIFF
--- a/packages/wrangler/src/cloudchamber/apply.ts
+++ b/packages/wrangler/src/cloudchamber/apply.ts
@@ -1,0 +1,585 @@
+import {
+	cancel,
+	crash,
+	endSection,
+	log,
+	logRaw,
+	shapes,
+	startSection,
+	success,
+	updateStatus,
+} from "@cloudflare/cli";
+import { processArgument } from "@cloudflare/cli/args";
+import { bold, brandColor, dim, green, red } from "@cloudflare/cli/colors";
+import { formatConfigSnippet } from "../config";
+import {
+	ApiError,
+	ApplicationsService,
+	DeploymentMutationError,
+	SchedulingPolicy,
+} from "./client";
+import { promiseSpinner } from "./common";
+import { diffLines } from "./helpers/diff";
+import { wrap } from "./helpers/wrap";
+import type { Config } from "../config";
+import type { ContainerApp } from "../config/environment";
+import type {
+	CommonYargsArgvJSON,
+	StrictYargsOptionsToInterfaceJSON,
+} from "../yargs-types";
+import type {
+	Application,
+	ApplicationID,
+	ApplicationName,
+	CreateApplicationRequest,
+	ModifyApplicationRequestBody,
+	UserDeploymentConfiguration,
+} from "./client";
+import type { JsonMap } from "@iarna/toml";
+
+export function applyCommandOptionalYargs(yargs: CommonYargsArgvJSON) {
+	return yargs.option("skip-defaults", {
+		requiresArg: true,
+		type: "boolean",
+		demandOption: false,
+		describe: "Skips recommended defaults added by apply",
+	});
+}
+
+function createApplicationToModifyApplication(
+	req: CreateApplicationRequest
+): ModifyApplicationRequestBody {
+	return {
+		configuration: req.configuration,
+		instances: req.instances,
+		constraints: req.constraints,
+		affinities: req.affinities,
+		scheduling_policy: req.scheduling_policy,
+	};
+}
+
+function applicationToCreateApplication(
+	application: Application
+): CreateApplicationRequest {
+	return {
+		configuration: application.configuration,
+		constraints: application.constraints,
+		name: application.name,
+		scheduling_policy: application.scheduling_policy,
+		affinities: application.affinities,
+		instances: application.instances,
+		jobs: application.jobs ? true : undefined,
+	};
+}
+
+function containerAppToCreateApplication(
+	containerApp: ContainerApp,
+	skipDefaults = false
+): CreateApplicationRequest {
+	const configuration =
+		containerApp.configuration as UserDeploymentConfiguration;
+	return {
+		...containerApp,
+		configuration,
+		scheduling_policy:
+			(containerApp.scheduling_policy as SchedulingPolicy) ??
+			SchedulingPolicy.REGIONAL,
+		constraints: {
+			...(containerApp.constraints ??
+				(!skipDefaults ? { tier: 1 } : undefined)),
+			cities: containerApp.constraints?.cities?.map((city) =>
+				city.toLowerCase()
+			),
+			regions: containerApp.constraints?.regions?.map((region) =>
+				region.toUpperCase()
+			),
+		},
+	};
+}
+
+function isNumber(c: string | number) {
+	if (typeof c === "number") {
+		return true;
+	}
+	const code = c.charCodeAt(0);
+	const zero = "0".charCodeAt(0);
+	const nine = "9".charCodeAt(0);
+	return code >= zero && code <= nine;
+}
+
+/**
+ * createLine takes a string and goes through each character, rendering possibly syntax highlighting.
+ * Useful to render TOML files.
+ */
+function createLine(el: string, startWith = ""): string {
+	let line = startWith;
+	let lastAdded = 0;
+	const addToLine = (i: number, color = (s: string) => s) => {
+		line += color(el.slice(lastAdded, i));
+		lastAdded = i;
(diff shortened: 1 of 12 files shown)


mikenomitch:
Screenshot 2024-11-18 at 11 00 57 PM

Love that this comes up when you run apply without an app in wrangler.toml.

I think"Deploy a container application deploy changes to your application" is a little confusing though. What's the white part vs the orange part? Maybe drop the white part?

mikenomitch:
Screenshot 2024-11-18 at 11 04 52 PM

Small bug when running the default app then modifying its instance count

mikenomitch:
Not a blocker: I kind of want a "cloudchamber status" code that can tell me what the status of the apps are in my wrangler.toml and if they're up now/what the diff would be.

mikenomitch:
I logged out of my account and then ran "cloudchamber apply" on a different account, but the diff it was giving me was from my old account

Screenshot 2024-11-18 at 11 08 04 PM

This should be fresh per account login right? (happy to be challenged on this)

mikenomitch:
Is there a way to unset tier? Should there be?

mikenomitch:
My dream in using this is to be able to somehow output a wrangler.toml configuration with every possible attribute that I might be able to set and have it add comments for what the possible options are. Docs could do this too, but I don't want to have to find them.

If you hooked up the plumbing, I could provide the template for this.

Not sure what the command would be to do it though "wrangler cloudchamber init --verbose" or something?

Maybe linking to the docs for "wrangler apply" from the "-h" command could be generally useful for describing attrs/showing a verbose toml config?

mikenomitch:
--json doesn't seem to do anything (at least not on noops or errors)

mikenomitch:
Overall, LOVE the overall feel and the workflow, but left some commends - some of which are brainstormy/not blockers.

mikenomitch:
I've now gotten into a state where I've made a bunch of apps via the CLI and I can't delete them via the CLI because they're app-driven, so even if I remove the deployments they'll come back to life (I think?).

We'll probably need either a "cloudchamber destroy" method that reads from the toml and reverts it all, and/or application CRUD from the CLI. Or maybe some special config to remove an application on next "apply" - thoughts?

gabivlj:
> Is there a way to unset tier? Should there be?

there should definitely be a way to skip defaults for sure

mikenomitch:
Screenshot 2024-12-02 at 4 43 56 PM

Small bug, when I apply with --json and this config over and over, it auto-applies the config, but then it seems to not actually take, because on the next apply it says Im going to change tier again.

mikenomitch (approved):

penalosa:
@gabivlj Is this ready for review?

gabivlj:
> @gabivlj Is this ready for review?

Yes. This is up for review. Command designed for heavy Cloudchamber-only users mainly.

My main worry to discuss with the wrangler folks here is the diff dependency we are introducing to detect line changes, we can discuss if we can have another method of detecting these line changes. I also understand that we are moving from TOML to JSON in some cases and this command will need to do that instead of TOML rendering.

packages/wrangler/src/config/environment.ts
```diff
@@ -1,3 +1,4 @@
+import type { CreateApplicationRequest } from "../cloudchamber/client";
```
penalosa, line 1: Types in this file are used to generate the public JSON schema for Wrangler's config, so ideally we want to keep the dependencies minimal. I know it's not ideal, but would you be able to duplicate the relevant `ContainerApp` type in this file, along with doc comments?

packages/wrangler/src/config/environment.ts
```diff
@@ -450,6 +462,17 @@ export interface EnvironmentNonInheritable {
...
+	/**
+	 * Container app configuration
+	 *
+	 * NOTE: This field is not automatically inherited from the top level environment,
+	 * and so must be specified in every named environment.
+	 *
+	 * @default `{}`
+	 * @nonInheritable
+	 */
+	container_app: ContainerApp[];
```
penalosa, line 474: Bikeshedding on the name, but do you have a link to any internal discussion that led to the `container_app` name?

packages/wrangler/src/cloudchamber/apply.ts
penalosa, line 1: I won't review this in detail since it's cloudchamber specific and limited to the cloudchamber commands, but as a not this shouldn't be using `TOML.stringify()` directly, instead preferring the `formatConfigSnippet()` helper to handle JSON config

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -0,0 +1,566 @@
...
+				stripUndefined(applicationToCreateApplication(application))
+			);
+
+			const prev = TOML.stringify({ container_app: [prevApp] });
+			const now = TOML.stringify({
+				container_app: [
+					sortObjectRecursive<CreateApplicationRequest>(appConfig),
+				],
+			});
+			const results = Diff.diffLines(prev, now);
```
penalosa, line 346: This seems slightly overkill, and I'm a bit concerned about adding a ~600kb dependency just for this. What type of data is this likely to be? If it's an objet, have you considered just diffing by key/value? That feels like it'd be sufficient to me, and wouldn't need a text diffing algorithm

packages/wrangler/src/config/environment.ts
mikenomitch: Hey @penalosa, I don't think we've got a link, but this was something that we landed on for the public name of these apps.

I do think this would be annoying to change after the fact though, so I'll kick off a convo with our team just to make sure we're all feeling good here and nobody has a last second veto.

"Application" is pretty set in stone for us on Cloudchamber at least, and then we were adding "container" here to differentiate it from Workers apps. IMO it's the right name for now, but will make sure we've got near-consensus.

packages/wrangler/src/cloudchamber/apply.ts
gabivlj: I can take a look at the dependency we are using and see if we can simplify, as 600kb for line diffing is too much. However, I'd argue that line diffing is the right approach here because we have complex nested objects inside arrays etc. that will make obj diffing approach a bit of a mess to test and render, and I've observed that line diffing to be easier to reason about once you just order the object properties.

packages/wrangler/src/config/environment.ts
gabivlj: There is like a huge amount of types we have to copy for this... And this type specifically gets new added properties and change. Are we sure we can't get away with importing this type?

packages/wrangler/src/config/environment.ts
mikenomitch: We talked about this a bunch as a team, and "[[containers.app]]" is what we want to go with. - App is approachable and we needed a clear name for group of container instances.

packages/wrangler/src/config/environment.ts
penalosa: If possible, writing it out in this file would be much better. It would also allow for handwriting doc comments to explain what each property does. This forms part of Wrangler's public config contract—changing based on changes to autogenerated code doesn't give much oversight over what's ending up in the config file.

Having said that, I wonder if the fact that this type is super detailed is in itself a bit of a red flag? What types of things does it contain? Why is it so big?

packages/wrangler/src/config/environment.ts
```diff
@@ -39,6 +40,17 @@ export type CloudchamberConfig = {
 	ipv4?: boolean;
 };
 
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
```
penalosa, line 43: There's already a TypeScript built-in for this, I believe?

penalosa (approved):
Approved from the Wrangler side pending dropping the `diff` dependency (which I think you've done?) and copying the config type into the config validation file

PR #7230, opened 2024-11-11
Turn on `wrangler.json(c)` support by default

penalosa:
Fixes [DEVX-1466](https://jira.cfdata.org/browse/DEVX-1466)

Turn on the `--experimental-json-config` flag by default, enabling reading of `wrangler.json` & `wrangler.jsonc` config files. It can still be disabled by setting `--experimental-json-config=false`. Since it's on by default, this also unblocks support for Pages (pending release of this PR & bumping of Wrangler in the build image)

This PR _only_ enables it in Wrangler. Follow up PRs will adjust the documentation to call out this support properly as well as updating C3.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18313


packages/wrangler/src/index.ts
```diff
@@ -212,6 +212,7 @@ export function createCLIParser(argv: string[]) {
 			alias: "j",
 			describe: `Experimental: support wrangler.json`,
```
Skye-31, line 213: Is it still experimental?

penalosa: Not after this PR, but the _flag_ is experimental (we'll be entirely removing it soon). The wording could be clarified

petebacondarwin (commented):
Ran out of time so dumping my comments so far.

packages/wrangler/src/config/validation.ts
```diff
@@ -73,7 +73,8 @@ export function isPagesConfig(rawConfig: RawConfig): boolean {
 export function normalizeAndValidateConfig(
 	rawConfig: RawConfig,
 	configPath: string | undefined,
-	args: NormalizeAndValidateConfigArgs
+	args: NormalizeAndValidateConfigArgs,
+	parsedFormat: Config["parsedFormat"] = "toml"
```
petebacondarwin, line 77: This function is internal, so there is no value in making this field optional, IMO. It is better to constrain all usage to provide this. That also helps to ensure we have considered every usage.

Also since this and `configPath` are similar properties (in their usage) I would propose to group this new param as the third one ahead of `args`.

packages/wrangler/src/config/validation.ts
```diff
@@ -1026,41 +1034,9 @@ const validateTailConsumers: ValidatorFn = (diagnostics, field, value) => {
...
-): Environment;
 /**
  * Validate the named environment configuration and return the normalized values.
  */
-function normalizeAndValidateEnvironment(
-	diagnostics: Diagnostics,
-	configPath: string | undefined,
-	rawEnv: RawEnvironment,
-	isDispatchNamespace: boolean,
-	envName: string,
```
petebacondarwin, line 1046: The idea of this overload was that if you specified an `envName` then you "had" to also specify all the other optional params.

packages/wrangler/src/config/validation.ts
```diff
@@ -1069,7 +1045,8 @@ function normalizeAndValidateEnvironment(
 	envName = "top level",
 	topLevelEnv?: Environment | undefined,
 	isLegacyEnv?: boolean,
-	rawConfig?: RawConfig | undefined
+	rawConfig?: RawConfig | undefined,
+	parsedFormat?: Config["parsedFormat"]
```
petebacondarwin, line 1049: Make this non-optional and move it up to below the `configPath` param?

petebacondarwin (changes requested):
I love where this is going but the more I read through the changes the more I questioned whether we need a `parseFormat` property on config.
We have the `configPath` property already, and it would be simple to provide a helper function like:

```ts
function configFormat(configPath: string|undefined): "json"|"toml"|"none" { ... }
```

This means not having to thread this extra parameter around alongside the `configPath`.
Moreover, as you can guess from the signature of the helper function above, there is the scenario where the user has not provided a config file at all, which we need to support. In that case, snippet messaging might be more along the lines of "Create a wrangler.json with this content".

petebacondarwin (approved):
Love it!
There may be some instances missing but I didn't find any.
Ship it!

packages/wrangler/README.md
```diff
@@ -43,7 +43,7 @@ $ npm install wrangler --save-dev
 
 ## Configuration:
 
-Wrangler is configured via a `wrangler.toml` file in the project root. When utilizing the `wrangler init` command, a `wrangler.toml` file will be created for you.
+Wrangler is configured via a `wrangler.toml` or `wrangler.json` file in the project root. When utilizing the `wrangler init` command, a `wrangler.toml` file will be created for you.
```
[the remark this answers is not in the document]
penalosa: I haven't changed the `wrangler init` path since it delegates to C3 (and the non-delegated version is deprecated). I'll change C3 in a month or so once this has had time to settle

PR #7236, opened 2024-11-12
feat: add telemetry commands

emily-shen:
Fixes [DEVX-1480](https://jira.cfdata.org/browse/DEVX-1480)

Adds new `wrangler telemetry disable` and `wrangler telemetry enable` commands which globally toggle telemetry collection across all projects on that machine.

Adds a new `wrangler telemetry status` command which shows the current telemetry setting, resolved to the current project.

With send_metrics set in wrangler.toml
Screenshot 2024-11-15 at 09 44 59

With WRANGLER_SEND_METRICS environment variable set
Screenshot 2024-11-15 at 09 45 20

Without either send_metrics or WRANGLER_SEND_METRICS set
Screenshot 2024-11-15 at 09 44 18

We prioritise env var > wrangler.toml > global setting via command.

NB we're using a feature branch so will add one changeset when we merge that in.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because:  no existing e2e's, adequately covered by unit tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: currently hidden. Will add later.

### DIFF
--- a/.changeset/little-crabs-rule.md
+++ b/.changeset/little-crabs-rule.md
@@ -1,5 +1,5 @@
 ---
-"wrangler": minor
+"wrangler": patch
 ---
 
-feat: add new `wrangler telemetry` commands to configure whether Wrangler collects usage telemetry
+feat: add new `wrangler telemetry` commands to configure whether Wrangler collects usage telemetry. These are hidden for now, and will be exposed in a future release once telemetry collection is fully implemented.
--- a/.changeset/little-crabs-rule.md
+++ b/.changeset/little-crabs-rule.md
@@ -1,5 +1,0 @@
----
-"wrangler": patch
----
-
-feat: add new `wrangler telemetry` commands to configure whether Wrangler collects usage telemetry. These are hidden for now, and will be exposed in a future release once telemetry collection is fully implemented.
--- a/.changeset/little-crabs-rule.md
+++ b/.changeset/little-crabs-rule.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feat: add new `wrangler telemetry` commands to configure whether Wrangler collects usage telemetry
--- a/packages/wrangler/src/__tests__/metrics.test.ts
+++ b/packages/wrangler/src/__tests__/metrics.test.ts
@@ -18,6 +18,7 @@ import { useMockIsTTY } from "./helpers/mock-istty";
 import { msw, mswSuccessOauthHandlers } from "./helpers/msw";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { runWrangler } from "./helpers/run-wrangler";
+import { writeWranglerToml } from "./helpers/write-wrangler-toml";
 import type { MockInstance } from "vitest";
 
 declare const global: { SPARROW_SOURCE_KEY: string | undefined };
@@ -469,6 +470,22 @@ describe("metrics", () => {
 				"Status: Enabled
 
 				Status: Disabled
+				"
+			`);
+		});
+
+		it("prints a global and project status if a wrangler.toml with send_metrics is present", async () => {
+			writeMetricsConfig({
+				permission: {
+					enabled: true,
+					date: new Date(2022, 6, 4),
+				},
+			});
+			writeWranglerToml({ send_metrics: false });
+			await runWrangler("telemetry status");
+			expect(std.out).toMatchInlineSnapshot(`
+				"Global status: Enabled
+				Project status: Disabled
 				"
 			`);
 		});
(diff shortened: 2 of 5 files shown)


packages/wrangler/src/metrics/commands.ts
```diff
@@ -0,0 +1,70 @@
...
+	command: "wrangler telemetry status",
+	metadata: {
+		description: "Check whether Wrangler telemetry collection is enabled",
+		owner: "Workers: Authoring and Testing",
+		status: "stable",
+	},
+	async handler(_, { config }) {
+		const metricsConfig = await getMetricsConfig({
+			sendMetrics: config.send_metrics,
+		});
```
edmundhung, line 60: Would `readMetricsConfig` be enough? `getMetricsConfig` does more to get userId and will re-confirm whether users wanna enable telemetry in some cases which seems unnecessary here.

.changeset/little-crabs-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+feat: add new `wrangler telemetry` commands to configure whether Wrangler collects usage telemetry
```
edmundhung, line 5: Should we make it clear that these commands are currently hidden and will be exposed in a future release once telemetry collection is fully implemented?

packages/wrangler/src/metrics/commands.ts
emily-shen: oh yeah meant to flag this - should `status` be per-project? i.e. should it take into account send_metrics in wrangler.toml at all. If we strictly want to to be about user-level, we should probably clarify that in the messaging. And also we should probably clarify the order of precedence for wrangler.toml vs env variable vs set-by-command.

Also it does re-confirm currently, but I figure we would remove that when we switch to on-by-default.

.changeset/little-crabs-rule.md
emily-shen: yep! also wondering if this is 'experimental' and thus a patch?

edmundhung: Either way works for me 🤔

packages/wrangler/src/metrics/commands.ts
edmundhung: I would expect the status to show the actual stored preference—basically, what the telemetry command is modifying. But yeah, we should make sure the message is clear here.

edmundhung: I feel like we should show two status: Global / Project (which reflects the send_metrics config?)

.changeset/little-crabs-rule.md
emily-shen: d697b4da81c39e31e6d1379950fb32876bc3e895

packages/wrangler/src/__tests__/metrics.test.ts
```diff
@@ -434,6 +435,98 @@ describe("metrics", () => {
...
+			await runWrangler("telemetry disable");
+			expect(std.out).toMatchInlineSnapshot(`
+				"Status: Disabled
+
+				Wrangler is no longer collecting telemetry about your usage.
+				"
+			`);
+			expect(await getMetricsConfig({})).toMatchObject({
+				enabled: false,
+			});
```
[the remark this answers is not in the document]
emily-shen: 9bdd400be8d809750b366a8a11ada603f09ad423

packages/wrangler/src/metrics/commands.ts
emily-shen: what do you think about this - b44e1b5e3f7e67f54a826a502dc550c6a369991d 
And do you think we should explain how global/project is set underneath, or is just the status fine?

edmundhung: Let's get some feedbacks from Nevi and Liz on this.

emily-shen: resolved in 7bc5de1fa7707d6731d8e82d1b417f793f9fcc67 after discussions with Nevi/Liz

packages/wrangler/src/__tests__/metrics.test.ts
```diff
@@ -434,6 +392,191 @@ describe("metrics", () => {
...
+					},
+				});
+				await runWrangler("telemetry status");
+				expect(std.out).toMatchInlineSnapshot(`
+					"Status: Enabled
+
+					To configure telemetry globally on this machine, you can run \`wrangler telemetry disable / enable\`.
+					You can override this for individual projects with the environment variable \`WRANGLER_SEND_METRICS=true/false\`.
+					"
+				`);
```
edmundhung, line 419: I think we can skip this assertion since we will have a full snapshot next anyway.

edmundhung (approved):
LGTM

edmundhung:
For posterity, there is no changeset as this is merged only into a feature branch.

PR #7241, opened 2024-11-12
bump C3's create-next-app to 15.0.3

dario-piotrowicz:
This PR is for updating C3 to create a Next.js 15 app in non-experimental mode

However (see comments below) the bump of create-next-app for non-experimental mode currently applies to both the non-experimental and experimental Next.js templates, forcing us to either make sure that both templates can work with Next.js 15 or that we find a way to have each use a different version of create-next-app

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included (the quarantined e2e Next.js test has been updated)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: I think that the Next.js (non-experimental) e2e test isn't run because it is quarantined
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: it's just a version bump

### DIFF
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -65,16 +65,21 @@ export type TemplateConfig = {
 	 * to handle config version skew between different versions of c3
 	 */
 	configVersion: number;
-	/** The id by which template is referred to internally and keyed in lookup maps*/
+	/** The id by which template is referred to internally and keyed in lookup maps */
 	id: string;
-	/** A string that controls how the template is presented to the user in the selection menu*/
+	/** A string that controls how the template is presented to the user in the selection menu */
 	displayName: string;
-	/** A string that explains what is inside the template, including any resources and how those will be used*/
+	/** A string that explains what is inside the template, including any resources and how those will be used */
 	description?: string;
 	/** The deployment platform for this template */
 	platform: "workers" | "pages";
-	/** The name of the framework cli tool that is used to generate this project or undefined if none. */
+	/** The name of the framework cli tool that is used to generate this project or undefined if none */
 	frameworkCli?: string;
+	/**
+	 * The version of the framework cli tool to use.
+	 * If omitted the cli version is taken from src/frameworks/package.json, which is the default/standard behavior.
+	 */
+	frameworkCliVersion?: string;
 	/** When set to true, hides this template from the selection menu */
 	hidden?: boolean;
 	/** Specifies a set of files that will be copied to the project directory during creation.
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -79,7 +79,7 @@ export type TemplateConfig = {
 	 * A specific version of the framework cli tool to use instead of the standard one taken from the src/frameworks/package.json
 	 * (which gets managed and bumped by dependabot)
 	 */
-	pinFrameworkCli?: string;
+	frameworkCliPinnedVersion?: string;
 	/** When set to true, hides this template from the selection menu */
 	hidden?: boolean;
 	/** Specifies a set of files that will be copied to the project directory during creation.
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -76,10 +76,10 @@ export type TemplateConfig = {
 	/** The name of the framework cli tool that is used to generate this project or undefined if none */
 	frameworkCli?: string;
 	/**
-	 * The version of the framework cli tool to use.
-	 * If omitted the cli version is taken from src/frameworks/package.json, which is the default/standard behavior.
+	 * A specific version of the framework cli tool to use instead of the standard one taken from the src/frameworks/package.json
+	 * (which gets managed and bumped by dependabot)
 	 */
-	frameworkCliVersion?: string;
+	pinFrameworkCli?: string;
 	/** When set to true, hides this template from the selection menu */
 	hidden?: boolean;
 	/** Specifies a set of files that will be copied to the project directory during creation.
--- a/packages/create-cloudflare/templates-experimental/next/c3.ts
+++ b/packages/create-cloudflare/templates-experimental/next/c3.ts
@@ -30,6 +30,10 @@ export default {
 	configVersion: 1,
 	id: "next",
 	frameworkCli: "create-next-app",
+	// TODO: here we need to specify a version of create-next-app which is different from the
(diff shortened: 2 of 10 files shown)


dario-piotrowicz:
> [!Note]
> This PR updates the non-experimental Next.js C3 template, so I think that the `C3 E2E Tests (experimental)` checks (which are failing because of Next.js) can safely be ignored (I can deal with them in a followup PR if that's ok?)

dario-piotrowicz:
> [!Note]
> Currently C3 creates a partially broken Next.js application, in order to fix it we could pin the `vercel` dependency to `39.1.1` or alternatively wait for the next next-on-pages release (which will include these changes: https://github.com/cloudflare/next-on-pages/pull/917)
> I think that that's a separate issue anyways (which effects Next.js 14 as well) so if it's ok I would just do it as a followup (I'd probably wait for the next next-on-pages release)

andyjessop (approved):
Looks great, thanks @dario-piotrowicz

petebacondarwin (changes requested):
> so I think that the C3 E2E Tests (experimental) checks (which are failing because of Next.js) can safely be ignored

I don't think it is fair to break the experimental Next.js support by bumping the Next.js dependency.
If we can't separate them to use different versions of Next.js then I think we need to ensure they are both fixed together.

dario-piotrowicz:
> > so I think that the C3 E2E Tests (experimental) checks (which are failing because of Next.js) can safely be ignored
> 
> I don't think it is fair to break the experimental Next.js support by bumping the Next.js dependency. If we can't separate them to use different versions of Next.js then I think we need to ensure they are both fixed together.

ah yeah for sure! I was under the assumption that the experimental e2e was failing either way I didn't realize the bump was breaking it (but it clearly runs without the bump: [example run](https://github.com/vitejs/vite/blob/ef7a6a35e6827b92445e5a0c2c0022616efc80dd/packages/vite/src/node/plugins/define.ts#L64))

it's quite unfortunate that the create-next-app bump effects both templates 😕 😓

petebacondarwin:
I think @vicb was working up updating opennextjs?

dario-piotrowicz:
I'm looking at the code and I don't think there's a clean way to use different versions of `create-next-app` 😓

We could have a new `package.json` just for the experimental templates... that feels like the cleanest solution, but also very much overkill for just Next.js 😓

Alternatively we could have some very hacky `if` statement that overrides the `create-next-app` version for the experimental Next.js template...

Or yeah just wait for our opennext adapter to support Next 15 😓

packages/create-cloudflare/templates-experimental/next/c3.ts
```diff
@@ -18,7 +18,14 @@ const generate = async (ctx: C3Context) => {
 };
 
 const configure = async () => {
-	const packages = ["@opennextjs/cloudflare", "@cloudflare/workers-types"];
+	const packages = [
+		// Note: this is using the experimental pkg.pr.new prerelease
+		//       we do this because the stable @opennextjs/cloudflare package
+		//       does not yet support Next 15, we should move away from this
+		//       as soon as possible
+		"https://pkg.pr.new/@opennextjs/cloudflare@experimental",
```
vicb, line 26: Haven't looked at the details but the experimental branch should have an open next config file and a different wrangler.toml.
Should they be updated as part of the PR?

Also loading instrumentation will fall on 15.0.3 if not patched.

dario-piotrowicz: ah yeah of course, good callout 😓👍

dario-piotrowicz:
@petebacondarwin what do you think of this temporary workaround?

https://github.com/cloudflare/workers-sdk/pull/7241/commits/d2bd8b16a15ee9cf93119e654ccef757f82c9d53

packages/create-cloudflare/src/templates.ts
```diff
@@ -65,16 +65,21 @@ export type TemplateConfig = {
...
 	/** The deployment platform for this template */
 	platform: "workers" | "pages";
-	/** The name of the framework cli tool that is used to generate this project or undefined if none. */
+	/** The name of the framework cli tool that is used to generate this project or undefined if none */
 	frameworkCli?: string;
+	/**
+	 * The version of the framework cli tool to use.
+	 * If omitted the cli version is taken from src/frameworks/package.json, which is the default/standard behavior.
+	 */
+	frameworkCliVersion?: string;
```
vicb, line 82: Another option would be to allow `frameworkCli` to specify the version (i.e. `create-next-app@14.2.5`)

petebacondarwin: I think this approach is fine. You could even change it to something to make it clear this is exceptional:

```suggestion
	pinFrameworkCliVersion?: string;
```
We use dependabot to keep the CLI versions up to date, which is why they are stored in a package.json. Moving the versions here would mess that up. But in this case we don't want dependabot to do anything with this version since we are explictly pinning it.

petebacondarwin (approved):

packages/create-cloudflare/src/templates.ts
vicb: My proposal is to use either `cli` to use the version from the package.json or `cli@version` to pin the version as an alternative to using the `pinFrameworkCliVersion ` property

dario-piotrowicz: @petebacondarwin I've renamed the field as you suggested 🙂 

I've also updated the comment to make the thing a bit clearer

https://github.com/cloudflare/workers-sdk/pull/7241/commits/c36670d831a9ea7ade96b91af8396474ea821b98

packages/create-cloudflare/src/templates.ts
```diff
@@ -65,16 +65,21 @@ export type TemplateConfig = {
...
 	/** The deployment platform for this template */
 	platform: "workers" | "pages";
-	/** The name of the framework cli tool that is used to generate this project or undefined if none. */
+	/** The name of the framework cli tool that is used to generate this project or undefined if none */
 	frameworkCli?: string;
+	/**
+	 * A specific version of the framework cli tool to use instead of the standard one taken from the src/frameworks/package.json
+	 * (which gets managed and bumped by dependabot)
+	 */
+	pinFrameworkCli?: string;
```
vicb, line 82: nit on the name: pin<...> make me think Boolean. Why not framewokCliPinnedVersion ?

petebacondarwin: Any of these work fine for me.

dario-piotrowicz: renamed 🙂👍

PR #7246, opened 2024-11-13
chore: use defineCommand in `wrangler docs`

emily-shen:

related #6655 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/docs/index.ts
+++ b/packages/wrangler/src/docs/index.ts
@@ -1,55 +1,49 @@
-import { printWranglerBanner } from "..";
-import { readConfig } from "../config";
+import { defineCommand } from "../core";
 import { logger } from "../logger";
-import * as metrics from "../metrics";
 import openInBrowser from "../open-in-browser";
 import { runSearch } from "./helpers";
-import type {
-	CommonYargsArgv,
-	StrictYargsOptionsToInterface,
-} from "../yargs-types";
 
-export function docsOptions(yargs: CommonYargsArgv) {
-	return yargs
-		.positional("search", {
+defineCommand({
+	command: "wrangler docs",
+	metadata: {
+		description: "📚 Open Wrangler's command documentation in your browser\n",
+		owner: "Workers: Authoring and Testing",
+		status: "stable",
+	},
+	args: {
+		search: {
 			describe:
 				"Enter search terms (e.g. the wrangler command) you want to know more about",
 			type: "string",
 			array: true,
-		})
-		.option("yes", {
+		},
+		yes: {
 			alias: "y",
 			type: "boolean",
-			description: "Takes you to the docs, even if search fails",
-		});
-}
+			describe: "Takes you to the docs, even if search fails",
+		},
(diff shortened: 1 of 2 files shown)


packages/wrangler/src/docs/index.ts
```diff
@@ -1,55 +1,49 @@
...
-		return;
-	}
-	await printWranglerBanner();
-
-	logger.log(`Opening a link in your default browser: ${urlToOpen}`);
-	await openInBrowser(urlToOpen);
-	const config = readConfig(undefined, args);
-	await metrics.sendMetricsEvent("view docs", {
-		sendMetrics: config.send_metrics,
-	});
```
emily-shen, line 54: removing this in preparation for new telementry

edmundhung: Would it make sense to keep this until we refactor the telemetry code? It might be easier to justify removing it later, and to give us flexibility in case we decide to keep it last minute.

packages/wrangler/src/docs/index.ts
```diff
@@ -1,55 +1,49 @@
...
-	StrictYargsOptionsToInterface,
-} from "../yargs-types";
 
-export function docsOptions(yargs: CommonYargsArgv) {
-	return yargs
-		.positional("search", {
+defineCommand({
+	command: "wrangler docs",
+	metadata: {
+		description: "📚 Open Wrangler's command documentation in your browser\n",
```
edmundhung, line 9: Do you know what is the usage of `\n` here? Is it used to adjust the layout of the help message?

packages/wrangler/src/index.ts
```diff
@@ -49,6 +48,7 @@ import { initHandler, initOptions } from "./init";
 import "./kv";
 import "./workflows";
 import "./user/commands";
+import "./docs";
```
edmundhung, line 51: I think the import order matters in terms of which commands are displayed first on the help message and I though the docs command was the first one..? Can you help confirming?

edmundhung (approved):

packages/wrangler/src/docs/index.ts
emily-shen: yeah makes sense

packages/wrangler/src/index.ts
emily-shen: sort of - right now register.registerNamespace determines the order, because it needs to be interleaved with the non-defineCommand commands (which is why the help tests are passing). i think eventually it will pull from the import order and then we can get rid of registerNamespace. I forgot about that tho, will reorder.

packages/wrangler/src/docs/index.ts
emily-shen: yeah, there are line breaks to group the commands in the help

PR #7250, opened 2024-11-13
feat: support extending the user configuration

petebacondarwin:
Fixes #DEVDASH-330

This feature is described in an internal doc: **MINI-SPEC:Allow 3rd party build tools to generate Worker code and configuration**

The main goal here is to enable tools to generate a partial configuration file that is merged into the user configuration when Wrangler commands are run.

The file must be written to `./.wrangler/config/extra.json`, where the path is relative to the project path, which is the directory containing the wrangler.toml or the current working directory if there is no wrangler.toml.

The format of the file is a JSON object whose properties are the inheritable and non-inheritable options described in the Wrangler configuration documentation. Notably it cannot contain the "top level" configuration properties.

The contents of the file will be merged into the configuration of the currently selected environment before being used in all Wrangler commands.

The user does not need to manually specify that this merging should happen. It is done automatically when the file is found.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18146

### DIFF
--- a/.changeset/slow-dolphins-decide.md
+++ b/.changeset/slow-dolphins-decide.md
@@ -1,0 +1,15 @@
+---
+"wrangler": patch
+---
+
+feat: support extending the user configuration
+
+The main goal here is to enable tools to generate a partial configuration file that is merged into the user configuration when Wrangler commands are run.
+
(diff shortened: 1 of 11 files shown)


petebacondarwin:
Closing as we are not taking this approach.

PR #7251, opened 2024-11-13
Support multiple workers in a single Miniflare instance

penalosa:
This PR allows you to run multiple workers in a single Miniflare instance, allowing for cross-service RPC without the complexities of the dev registry. To try it out, pass multiple `-c` flags to Wrangler: i.e. `wrangler dev -c wrangler.toml -c ../other-worker/wrangler.toml`. The first config will be treated as the _primary_ worker and will be exposed over HTTP as usual (`localhost:8787`) while the rest will be treated as _secondary_ and will only be accessible via a service binding from the primary worker.

Screenshot 2024-11-13 at 17 50 02

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/issues/18449


irvinebroque:
What would we need in order for this to work without the `-c` flag? Something in `wrangler.toml` that told you the directory where the Worker that you're bound to lives? 

(not saying we need to solve for it rn, just curious what the config decision we're facing is)

andyjessop:
The problem with adding a new field to services in wrangler.toml is that you're then encoding this idea of a primary worker, whereas you might want to start up one of the bound workers as a primary worker to work with it directly. It might be useful also to think about a true multi-worker config, where you don't have a single primary worker, and where you can start any of the defined workers as the primary. E.g.:

```js
{
  ...
  workers: [
    {
      name: "a",
      main: "../a/a.ts",
      bindings: { c: { "type": "worker", name: "c" } },
    },
    {
      name: "b",
      main: "../b/b.ts",
      bindings: { c: { "type": "worker", name: "c" } },
    },
    {
      name: "c",
      main: "../c/c.ts",
    },
  ]
}
``` 

```bash
# starts worker "a" at http:localhost:8787 with "c" accessible only through RPC
npx wrangler dev --worker a

# starts worker "b" at http:localhost:8787 with "c" accessible only through RPC
npx wrangler dev --worker b

# starts worker "c" at http:localhost:8787
npx wrangler dev --worker c
```

Multi-worker is a new paradigm for `wrangler`, so I would worry about adding it to the single-worker config as a patch in case we're painting ourselves into a corner. A new style config could be launched as a minor patch and would force us to tidy the unwieldy `config` in the code base.

(I don't say the above to derail the PR - @penalosa's solution is a neat and flexible one that doesn't tie us to any future arrangement).

packages/wrangler/src/dev.ts
```diff
@@ -561,216 +700,158 @@ export async function startDev(args: StartDevOptions) {
...
-				typeof vitest === "undefined"
-			) {
-				process.send(
-					JSON.stringify({
-						event: "DEV_SERVER_READY",
-						ip: url.hostname,
-						port: parseInt(url.port),
-					})
-				);
+		if (Array.isArray(configPath)) {
```
andyjessop, line 725: I feel like we could clean things up if we just treated all scenarios as multi-worker. So instead of having a separate `MultiworkerRuntimeController`, just modify the `LocalRuntimeController` to handle multi-workers, and then cast the configPath to an array in the yargs definition. 

```ts
const configs = Array.isArray(configPath) ? configPath : [configPath];

// Now no need to handle the single worker case. LocalRuntimeController assumes multi.
const runtime = new LocalRuntimeController(configs.length);
```

Or is there a reason to keep the single worker code path?

andyjessop (commented):
This is amazing work - very nice observation that we essentially already have this functionality in Miniflare and it just needed hooking up 👍

packages/wrangler/src/dev.ts
penalosa: I went down this path in [Merge Local & Multi controllers](https://github.com/cloudflare/workers-sdk/pull/7251/commits/d632e70131e553a3814aceccf59d8ea10e55f84f), but I'm starting to get a bit concerned about the potential for disrupting the normal `dev` flow, and I think it might end up being more complexity than it's worth. The single worker dev flow has two runtimes, remote & local, and the remote runtime doesn't (and will never) work for the multiworker use case. Because of that there's some differences in how these flows are set up (as well as the disabling of the dev registry in the multiworker flow, which involves disabling the registry as well as a change to how DOs & entrypoints are registered). It's not that much duplicated code, and I think it's actually clearer to keep them separate, and easier to reason about. What do you think?

zhihengGet:
does this support pages , i tried wrangler pages dev .svelte /cloudfare -c ... but got Pages does not support custom paths for the `wrangler.toml` configuration file

packages/wrangler/src/dev.ts
andyjessop: Yeah, that sounds fair enough. Just for clarity, I assume the remote runtime will never be able to do this because it is always deploying a single worker.

packages/wrangler/src/deployment-bundle/bundle.ts
```diff
@@ -305,6 +307,25 @@ export async function bundleWorker(
...
+			fs.writeFileSync(
+				patchedConsoleFileToInject,
+				fs.readFileSync(
+					path.resolve(getBasePath(), "templates/patch-console.js")
+				)
+			);
+		}
+
+		inject.push(patchedConsoleFileToInject);
+	}
```
andyjessop, line 328: Can this be done inside middleware so that we don't have to hard-code the code path in the main bundle handler?

(sorry this wasn't in the first review)

packages/wrangler/templates/patch-console.js
```diff
@@ -0,0 +1,9 @@
+globalThis.console = new Proxy(globalThis.console, {
+	get(target, p, receiver) {
+		if (p === "log" || p === "debug" || p === "info") {
+			return (...args) =>
+				Reflect.get(target, p, receiver)(WRANGLER_WORKER_NAME, ...args);
+		}
+		return Reflect.get(target, p, receiver);
+	},
+});
```
andyjessop, line 9: Perhaps we could wrap the logger instead of patching the console method directly.

(sorry this wasn't in the first review)

penalosa: I don't think I quite know what you mean—which logger?

packages/wrangler/src/deployment-bundle/bundle.ts
penalosa: https://github.com/cloudflare/workers-sdk/pull/7251/commits/4f2b07044fd5a24d5e626a80c8f02e8387203898

emily-shen:
> does this support pages , i tried `wrangler pages dev .svelte/cloudfare -c ../packages/wranger.toml` ... but got Pages does not support custom paths for the `wrangler.toml` configuration file

@zhihengGet this does not support pages, but there is a [followup](#7290) to get it working with [Workers + static assets](https://developers.cloudflare.com/workers/static-assets/)

packages/wrangler/templates/patch-console.js
penalosa: That's the Wrangler logger, whereas this code is patching the console global _inside_ a Worker, which is slightly different

packages/wrangler/src/api/dev.ts
```diff
@@ -228,6 +228,8 @@ export async function unstable_dev(
 	const devServer = await run(
 		{
 			FILE_BASED_REGISTRY: fileBasedRegistry,
+			// TODO: can we make this work?
```
CarmenPopoviciu, line 231: if TODO is relevant, should we keep track of it in a separate issue?

packages/wrangler/e2e/multiworker-dev.test.ts
```diff
@@ -0,0 +1,321 @@
...
+				async () =>
+					await expect(fetchText(`${url}/service`)).resolves.toBe(
+						"Hello from service worker"
+					),
+				{ interval: 1000, timeout: 10_000 }
+			);
+		});
+	});
+
+	describe("durable objects", () => {
```
emily-shen, line 264: would be nice to get some tests for RPC too (to DO and regular worker), since that's explicitly called out in the changeset. And named entrypoints too maybe. don't want to block on this, but would be good to get confirmation its been tested manually + followup with it?

packages/wrangler/src/api/dev.ts
penalosa: https://github.com/cloudflare/workers-sdk/issues/7369

packages/wrangler/src/dev.ts
```diff
@@ -560,216 +700,154 @@ export async function startDev(args: StartDevOptions) {
...
+
+			if (isInteractive() && args.showInteractiveDevSession !== false) {
+				unregisterHotKeys = registerDevHotKeys(primaryDevEnv, args);
 			}
-		});
 
-		if (!args.disableDevRegistry) {
-			teardownRegistryPromise = devRegistry((registry) =>
-				updateDevEnvRegistry(devEnv, registry)
+			// Set up the primary DevEnv (the one that the ProxyController will connect to)
```
emily-shen, line 734: is there any reason we can't have a proxy controller + proxy worker per devEnv, and thus expose each worker over http?

packages/wrangler/src/dev.ts
```diff
@@ -560,216 +700,154 @@ export async function startDev(args: StartDevOptions) {
...
-				usesTypeScript: /\.tsx?$/.test(
-					devEnv.config.latestConfig?.entrypoint as string
-				),
+				// usesTypeScript: /\.tsx?$/.test(
+				// 	devEnv.config.latestConfig?.entrypoint as string
+				// ),
 			},
 			{
-				sendMetrics: devEnv.config.latestConfig?.sendMetrics,
+				// sendMetrics: devEnv.config.latestConfig?.sendMetrics,
```
emily-shen, line 828: bit concerned about ignoring send_metrics. should we remove this metrics call?

penalosa: Removed in e14865fef

packages/wrangler/src/dev.ts
penalosa: In theory, but I think it can be a follow up—this service binding approach will work for a lot of use cases. The ProxyWorker runs in it's own workerd instance, and it feels a bit weird to spin up loads of workerd instances for the proxies when the user workers all fit in one.

emily-shen: ah I missed that they were in separate workerd instances. Makes sense to do in a followup, was just wondering :)

emily-shen (approved):
🎉 🎉 🎉

irvinebroque:
👏

grocco:
@penalosa This is awesome! 

I still could not figure out how to support multiple workers through [getPlatformProxy](https://developers.cloudflare.com/workers/wrangler/api/#getplatformproxy) though. The documentation states that for Durable Objects bindings to work one has to run a separate "wrangler dev" in another terminal; however RPC calls are not supported when doing so.

Is there an existing solution?

I thought maybe the `configPath` option in `getPlatformProxy` would now support an array but it does not seem to be the case.

I am trying to use Durable Objects that use RPC calls in a Remix application where locally I use `cloudflareDevProxyVitePlugin`.

sassanh:
@grocco I'm not working for cloudflare, but I had a similar issue and I ended up endpoints on the durable object for local environment and calling RPC in remote environments.
I added a handler for each expected-to-be-called-by-RPC method in `fetch` method for this purpose.
It is just a hacky workaround until they provide RPC in local environments.

grocco:
@sassanh Thank you for the workaround suggestion, that's indeed what I'm doing currently, otherwise I'd be stuck. Looking forward to local env RPC support!

cloudkite:
> Let's start the conversation on whether or not users will still want/need workers to be accessed from distinct endpoints, or if the dev registry is something we could remove entirely (perhaps in v4?).

👋 @andyjessop removing the dev registry would break our setup. We have different workers running on different subpaths of localhost. Which unless I am mistaken is not supported by this change since only the primary worker endpoint is exposed

PR #7256, opened 2024-11-14
Export config types/utils

jamesopstad:
In the Vite plugin we are looking at generating `wrangler.json` files as part of the build process. Exporting these utils and types enables us to do this.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: type exports only
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: type exports only
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not currently for general usage

### DIFF
--- a/.changeset/breezy-cats-matter.md
+++ b/.changeset/breezy-cats-matter.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Export RawConfig and RawEnvironment types from Wrangler
--- a/.changeset/breezy-cats-matter.md
+++ b/.changeset/breezy-cats-matter.md
@@ -2,4 +2,5 @@
 "wrangler": patch
 ---
 
-Export RawConfig and RawEnvironment types from Wrangler
+Export unstable_readConfig function and UnstableConfig, UnstableRawConfig, UnstableRawEnvironment and UnstableMiniflareWorkerOptions types from Wrangler.
+Overload unstable_getMiniflareWorkerOptions function to accept a config that has already been loaded.
--- a/packages/wrangler/src/api/integrations/platform/index.ts
+++ b/packages/wrangler/src/api/integrations/platform/index.ts
@@ -16,11 +16,12 @@ import { getLegacyAssetPaths, getSiteAssetPaths } from "../../../sites";
 import { CacheStorage } from "./caches";
 import { ExecutionContext } from "./executionContext";
 import { getServiceBindings } from "./services";
-import type { Config } from "../../../config";
+import type { Config, RawConfig, RawEnvironment } from "../../../config";
 import type { IncomingRequestCfProperties } from "@cloudflare/workers-types/experimental";
 import type { MiniflareOptions, ModuleRule, WorkerOptions } from "miniflare";
 
-export type { RawConfig, RawEnvironment } from "../../../config";
+export { readConfig };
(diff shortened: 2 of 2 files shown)


.changeset/breezy-cats-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Export RawConfig and RawEnvironment types from Wrangler
```
dario-piotrowicz, line 5:
```suggestion
feat: Export RawConfig and RawEnvironment types from Wrangler
```

(if I remember correctly the wrangler team had the convention to follow conventional commit message formats here)

penalosa:
Can we export these with an experimental/unstable name? We're about to work on a first-class API for doing this (DEVX-1423 & DEVX-1424) which is probably what we'd want frameworks to be using longer term

.changeset/breezy-cats-matter.md
penalosa: We did, but we dropped it a few months ago

dario-piotrowicz: 😮 ah ok, good to know 🙂

penalosa (approved):

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -16,10 +16,17 @@ import { getLegacyAssetPaths, getSiteAssetPaths } from "../../../sites";
...
 import { ExecutionContext } from "./executionContext";
 import { getServiceBindings } from "./services";
-import type { Config } from "../../../config";
+import type { Config, RawConfig, RawEnvironment } from "../../../config";
 import type { IncomingRequestCfProperties } from "@cloudflare/workers-types/experimental";
 import type { MiniflareOptions, ModuleRule, WorkerOptions } from "miniflare";
 
+export { readConfig as unstable_readConfig };
+export type {
+	Config as UnstableConfig,
```
petebacondarwin, line 25: Quick question, should we use `Unstable_Config`, etc. similar to `unstable_readConfig`?

petebacondarwin: In the future there might actually be a type that starts with Unstable...

petebacondarwin (approved):

packages/wrangler/src/api/integrations/platform/index.ts
jamesopstad: I was going to do that but noticed there are already some types named like this without the underscore e.g. `UnstableDevOptions`, `UnstableDevWorker`. Thought it was better to keep it consistent.

petebacondarwin: Oh... 😢 
I think that was probably a bad idea. But I guess it's best to stick with the convention.

jamesopstad: You can change them all if you like (they're called unstable so people can't exactly say they haven't been warned). That should probably be in a separate PR though.

petebacondarwin:
We're going to land this and then follow up today with a PR to rename all unstable types to be prefixed with `Unstable_...`

PR #7272, opened 2024-11-15
Make debug log for `.env` not found less scary

### LINKED ISSUE #7042: 🐛 BUG: .env file missing error when none exists and WRANGLER_LOG="debug"

Wrangler

3.81.0

22

Mac Sonoma 14.6

### Describe the Bug

If I set `export WRANGLER_LOG=debug` and then deploy a Worker, and there is no `.env` file in the cwd, I get this error in my logs:

```
>  export WRANGLER_LOG=debug 
>  npx wrangler@latest deploy
🪵  Writing logs to "/Users/brendan/.wrangler/logs/wrangler-2024-10-21_15-25-39_001.log"
Failed to load .env file ".env": Error: ENOENT: no such file or directory, open '.env'
    at Object.openSync (node:fs:573:18)
    at Object.readFileSync (node:fs:452:35)
    at tryLoadDotEnv (/Users/brendan/.npm/_npx/d77349f55c2be1c0/node_modules/wrangler/wrangler-dist/cli.js:163894:72)
    at loadDotEnv (/Users/brendan/.npm/_npx/d77349f55c2be1c0/node_modules/wrangler/wrangler-dist/cli.js:163903:12)
    at /Users/brendan/.npm/_npx/d77349f55c2be1c0/node_modules/wrangler/wrangler-dist/cli.js:214085:20
    at /Users/brendan/.npm/_npx/d77349f55c2be1c0/node_modules/wrangler/wrangler-dist/cli.js:179349:16
    at maybeAsyncResult (/Users/brendan/.npm/_npx/d77349f55c2be1c0/node_modules/wrangler/wrangler-dist/cli.js:177570:44)
    at /Users/brendan/.npm/_npx/d77349f55c2be1c0/node_modules/wrangler/wrangler-dist/cli.js:179348:14
    at /Users/brendan/.npm/_npx/d77349f55c2be1c0/node_modules/wrangler/wrangler-dist/cli.js:177557:22
    at Array.reduce (<anonymous>) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '.env'
}
```

This is misleading since it's not actually a problem, we are just trying to load `.env` and swallowing the error (except when you turn on `WRANGLER_LOG=debug`.

Even when debug logs are on, this seems like an expected scenario not something that should error with a stacktrace?

_No response_

_No response_

penalosa:
Fixes #7042

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: debug log change


packages/wrangler/src/config/index.ts
```diff
@@ -670,7 +670,11 @@ function tryLoadDotEnv(path: string): DotEnv | undefined {
 		const parsed = dotenv.parse(fs.readFileSync(path));
 		return { path, parsed };
 	} catch (e) {
-		logger.debug(`Failed to load .env file "${path}":`, e);
+		if ((e as { code: string }).code === "ENOENT") {
+			logger.debug(`.env file not found at "${path}"`);
```
emily-shen, line 674: definitely less scary but it still could be mistaken for an error - can we add a short message to say this is fine to ignore if you're not trying to use .dev.vars, .env etc.?

packages/wrangler/src/config/index.ts
```diff
@@ -670,7 +670,11 @@ function tryLoadDotEnv(path: string): DotEnv | undefined {
 		const parsed = dotenv.parse(fs.readFileSync(path));
 		return { path, parsed };
 	} catch (e) {
-		logger.debug(`Failed to load .env file "${path}":`, e);
+		if ((e as { code: string }).code === "ENOENT") {
```
vicb, line 673: Wouldn't testing if the file exists before reading it and then catching this exc be more sensible?

petebacondarwin: There is a small chance of a race condition where the file is removed between the check and the load. And since we need to catch any other load error (such as invalid format) we might as well do it all in one catch block.

petebacondarwin (approved):

packages/wrangler/src/config/index.ts
vicb: > There is a small chance of a race condition

The chance is so small that I think not showing a "less scary" message in that case should be ok?

This would avoid scary code `((e as { code: string }).code === "ENOENT")` :)

petebacondarwin: This is a common paradigm in this code base and in Node.js in general. I wouldn't stress about it.

vicb: I think a more common paradigm is:

1. validate the inputs
2. `try { something that is most likely to work }`
3. `catch { "exception"al errors }`

Also I think `e as { code: string }` is an anti-pattern. What it conveys is that at this point we know that `e` is a `{code: string}` which is a false assumption. `e` could be `any`thing here.

I think the correct way to write this is:

```ts
catch (e: any) {
  if (e.code === '...') { ... }
}
```

I have seen this pattern multiple time - the root cause is probably linter rules disallowing `: any`. In this case I think that the correct way to fix is to disable the linter on `catch (e: any) {` rather than introducing a cast which is even more misleading.

That being said it's only a comment about my thoughts on best practice that can be discarded if deemed not valuable.

penalosa: Totally fair, but we use this pattern a lot in the Wrangler codebase, and so for consistency I'm going to keep this as-is

PR #7278, opened 2024-11-18
fix: ensure Vitest Pool Workers doesn't error with nodejs_compat_v2 flag

### LINKED ISSUE #7156: 🐛 BUG: Compatibility Issue: `nodejs_compat_v2` Flag Causes Errors with Vitest Integration `@cloudflare/vitest-pool-workers`

Wrangler, Workers Vitest Integration

3.84.1 [Wrangler], 2.1.4 [Vitest], 0.5.24 [@cloudflare/vitest-pool-workers], 4.5.1 [yarn], 5.6.3 [typescript]

20.11.1

macOS Sequoia 15.1 (24B83)

### Describe the Bug

According to Cloudflare's documentation on Node.js compatibility ([link](https://developers.cloudflare.com/workers/runtime-apis/nodejs/)), setting the compatibility flag to `nodejs_compat_v2` should enable compatibility behavior equivalent to `nodejs_compat` as long as the compatibility date is September 23, 2024, or later. However, using `nodejs_compat_v2` results in compatibility issues with Vitest integration in my project, causing the following error during testing:

#### Error Log

```
Error: In project vitest.config.mts's configuration file wrangler.toml, `compatibility_flags` must contain "nodejs_compat".
This flag is required to use `@cloudflare/vitest-pool-workers`.
 ❯ assertCompatibilityFlagEnabled ...
```

### Observed Behavior

When using `nodejs_compat_v2`, Vitest testing fails with a `compatibility_flags` error indicating that `nodejs_compat` is required. Switching back to `nodejs_compat` resolves this error, but it contradicts the expected functionality stated in the documentation.

### Expected Behavior

With `nodejs_compat_v2` set as the compatibility flag, tests should pass successfully without requiring `nodejs_compat`.

### Steps to Reproduce

1. Set compatibility flag to `nodejs_compat_v2` in `wrangler.toml` with a compatibility date after September 23, 2024.
2. Run tests using `@cloudflare/vitest-pool-workers`.
3. Observe the compatibility error.

This issue seems to be tied to the interaction between `nodejs_compat_v2` and Vitest's Cloudflare integration package.

https://github.com/ZL-Asica/KumoAuth

```bash
Using vars defined in .dev.vars

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

Vitest caught 1 unhandled error during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Error ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: In project vitest.config.mts's configuration file wrangler.toml, `compatibility_flags` must contain "nodejs_compat".
This flag is required to use `@cloudflare/vitest-pool-workers`.
 ❯ assertCompatibilityFlagEnabled node_modules/.store/@cloudflare-vitest-pool-workers-virtual-36b606629f/package/dist/pool/index.mjs:1155:11
 ❯ buildProjectWorkerOptions node_modules/.store/@cloudflare-vitest-pool-workers-virtual-36b606629f/package/dist/pool/index.mjs:1178:3
 ❯ buildProjectMiniflareOptions node_modules/.store/@cloudflare-vitest-pool-workers-virtual-36b606629f/package/dist/pool/index.mjs:1294:47
 ❯ getProjectMiniflare node_modules/.store/@cloudflare-vitest-pool-workers-virtual-36b606629f/package/dist/pool/index.mjs:1321:21
 ❯ Object.runTests node_modules/.store/@cloudflare-vitest-pool-workers-virtual-36b606629f/package/dist/pool/index.mjs:1595:26
 ❯ executeTests node_modules/.store/vitest-virtual-7eb2413551/package/dist/chunks/resolveConfig.DPmbhVlP.js:7678:5
 ❯ node_modules/.store/vitest-virtual-7eb2413551/package/dist/chunks/cli-api.CKrRYkw8.js:10800:9
 ❯ Vitest.runFiles node_modules/.store/vitest-virtual-7eb2413551/package/dist/chunks/cli-api.CKrRYkw8.js:10820:12
 ❯ Vitest.start node_modules/.store/vitest-virtual-7eb2413551/package/dist/chunks/cli-api.CKrRYkw8.js:10663:7

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 Test Files  no tests
      Tests  no tests
     Errors  1 error
   Start at  13:19:48
   Duration  435ms (transform 0ms, setup 0ms, collect 0ms, tests 0ms, environment 0ms, prepare 0ms)


 FAIL  Tests failed. Watching for file changes...
       press h to show help, press q to quit
[vpw:dbg] Shutting down runtimes...
```

andyjessop:
Fixes #7156.

In Vitest Pool Workers, we assert the presence of the `nodejs_compat` flag, but it is also possible to use the `nodejs_compat_v2` flag, so we have to account for this possibility when asserting the validity of the config file.

This PR creates a new class to handle the validation, simplifying the logic as well as the testing. Some small improvements alongside adding the `nodejs_compat_v2` functionality:

- you can now add console logs to the code and see the logs in the unit tests
- the tests now run faster
- the code does not assume where the options path is located (i.e. `miniflare`)
- it is now clear how to add extra functionality should it be required - i.e. just add an extra assertion
- each test setup has a separate `test` block to allow for simpler debugging.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Bug fix


packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -399,25 +338,41 @@ function buildProjectWorkerOptions(
...
+			]),
+	];
+
+	for (const assertion of assertions) {
+		const result = assertion();
+		if (!result.isValid) {
+			throw new Error(result.errorMessage);
+		}
+	}
+
```
petebacondarwin, line 375: Miniflare already exposes a function for working out what the current node_compat mode we are in (it is not trivial due to the negation flags).

See https://github.com/cloudflare/workers-sdk/blob/c1022d66c6536c9e8199ebffe73d2358aa4064ee/packages/miniflare/src/plugins/core/node-compat.ts#L24

I think that `isFlagPresent()` is actually too simple to cover the cases.

I would argue that we should just use the `getNodeCompat()` function from Miniflare, except that I see that we also need this other `export_commonjs_namespace` compat flag, which potentially also has complexity due to compat date default and negation.

I propose that if we are going to have this logic here, we just create a function called `isEnabled(enableFlag: string, disableFlag: string, defaultOnDate?: string)`, then you could just do:

```ts
if (!compatFlagAssertions.isEnabled("export_commonjs_default", "export_commonjs_namespace", "2022-10-31")) {
  throw new Error("export_commandjs_default must be enabled - either explicitly or by compat date being after 2022-10-31");
}
if (
  !compatFlagAssertions.isEnabled("nodejs_compat", "no_nodejs_compat") &&
  !compatFlagAssertions.isEnabled("nodejs_compat_v2", "no_nodejs_compat_v2")
  ) {
  throw new Error("Node.js compatibility must be enabled - by adding either `nodejs_compat` or `nodejs_compat_v2` compat flag");
}
```

petebacondarwin (changes requested):

packages/vitest-pool-workers/src/pool/index.ts
andyjessop: At the moment, we also have messaging depending on the options path, the wrangler path, and the relative project path. Are you suggesting we just drop that for a simpler set of messages? Most of the code in this class is dealing with messaging, so being liberated from that would simplify things significantly.

petebacondarwin: OK looking at the code again, I think we could just reuse most of the logic in `assertCompatibiltyFlagEnabled()` but update it so that `enableFlag: string` becomes `enableFlags: string[]` and similarly with `disableFlag: string` becoming `disableFlags: string[]`.

The logic just needs a bit of tweaking to ensure that none of the `disableFlags` are in there and either at least one of the enableFlags is there or the compatibility date is new enough.

So then still throwing inside that function.

petebacondarwin: And if you still want to move it out into a class that is cool too!

andyjessop: I've updated this to consolidate the `assertDisableFlagNotPresent` and `assertEnableFlagOrCompatibilityDate` methods, but will wait for discussion on the messaging before proceeding further with the refactor.

andyjessop: Sorry Pete, I didn't see your comment before pushing my latest update.

packages/vitest-pool-workers/src/pool/compatibility-flag-assertions.ts
```diff
@@ -0,0 +1,164 @@
...
+function isDateSufficient(
+	compatibilityDate?: string,
+	defaultOnDate?: string
+): boolean {
+	if (!compatibilityDate || !defaultOnDate) {
+		return false;
+	}
+	const compDate = parseDate(compatibilityDate);
+	const reqDate = parseDate(defaultOnDate);
+	return compDate >= reqDate;
```
petebacondarwin, line 163: Compat date comparison can simply be string comparison thanks to the format of the string!

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -399,25 +338,38 @@ function buildProjectWorkerOptions(
 	// compatibility flags enabled. Vitest makes heavy use of Node APIs, and many
 	// of the libraries it depends on expect `require()` to return
 	// `module.exports` directly, rather than `{ default: module.exports }`.
+	// Initialize compatibilityFlags if undefined
```
petebacondarwin, line 341: NIT: I don't think this comment is needed.

petebacondarwin (approved):

Issue #5481, opened 2024-04-02
🐛 BUG: Wrangler executes from the wrong directory when package.json higher than wrangler.toml


GregBrimble:
### Which Cloudflare product(s) does this pertain to?

Wrangler core

### What version(s) of the tool(s) are you using?

3.43.0

### What version of Node are you using?

20.12.0

### What operating system and version are you using?

macOS

### Describe the Bug

```
- package.json
- packageA
  - package.json
  - serviceA
    - wrangler.toml
    - src
      - index.ts
```

Running `wrangler dev` in `packageA/serviceA` will result in Wrangler not picking up `packageA/serviceA/wrangler.toml` (so it'll error with ``✘ [ERROR] Missing entry-point: The entry-point should be specified via the command line (e.g. `wrangler dev path/to/script`) or the `main` config field.``

Instead, it thinks its running in `packageA`. This can be proven by running `wrangler dev -c serviceA/wrangler.toml` (and it correctly picking up the config). 

Wrangler shouldn't hoist itself to the nearest package.json (esp. now that we support Python Workers). It should _maybe_ hoist itself to the nearest `wrangler.toml`. I don't really have strong opinions on that: I personally always run Wrangler in the same dir as `wrangler.toml`.

GregBrimble:
Possibly an `npx` issue?

saturnonearth:
I am experiencing the same issue when trying to use the new monorepo features with the `wrangler.toml` support. It only seems to look for a `wrangler.toml` file inside the root directory of the git repo (even if the root directory is changed manually in cloudflare pages settings).

This does not work because if I have multiple projects in a monorepo, I would want my `wrangler.toml` inside the project/app folder and to only apply to that specific app, not the entire git repo. hopefully that makes sense?

geelen:
FWIW, I just hit this when running `pnpm wrangler ...` in a subdirectory. Thankfully, `npx wrangler ...` in the same place works fine, so I've switched to using that.

PR #7285, opened 2024-11-18
Fix relative project root

penalosa:
https://github.com/cloudflare/workers-sdk/pull/6971 introduced a small regression in the `getEntry()` function that meant that `Entry.directory` was always `process.cwd()` rather than relative to the found `wrangler.toml` file. This PR fixes the regression & adds a regression test that fails on `main` but passes prior to #6971.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/__tests__/get-entry.test.ts
+++ b/packages/wrangler/src/__tests__/get-entry.test.ts
@@ -1,19 +1,20 @@
-import { writeFile } from "fs/promises";
 import path from "path";
 import dedent from "ts-dedent";
-import { Config } from "../config";
 import { defaultWranglerConfig } from "../config/config";
-import { Entry, getEntry } from "../deployment-bundle/entry";
-import guessWorkerFormat from "../deployment-bundle/guess-worker-format";
+import { getEntry } from "../deployment-bundle/entry";
 import { mockConsoleMethods } from "./helpers/mock-console";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { seed } from "./helpers/seed";
+import type { Entry } from "../deployment-bundle/entry";
 
 function normalize(entry: Entry): Entry {
 	return JSON.parse(
-		JSON.stringify(entry).replaceAll(process.cwd(), "/tmp/dir")
+		JSON.stringify(entry)
+			.replaceAll(process.cwd(), "/tmp/dir")
+			.replaceAll(path.sep, "/")
 	);
 }
+
 describe("getEntry()", () => {
 	runInTempDir();
 	mockConsoleMethods();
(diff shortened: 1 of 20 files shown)


CarmenPopoviciu (approved):

packages/wrangler/src/__tests__/get-entry.test.ts
```diff
@@ -0,0 +1,128 @@
...
+import { mockConsoleMethods } from "./helpers/mock-console";
+import { runInTempDir } from "./helpers/run-in-tmp";
+import { seed } from "./helpers/seed";
+
+function normalize(entry: Entry): Entry {
+	return JSON.parse(
+		JSON.stringify(entry).replaceAll(process.cwd(), "/tmp/dir")
+	);
+}
+describe("getEntry()", () => {
```
CarmenPopoviciu, line 17: 🙏 🙇 ♥️

packages/wrangler/src/deployment-bundle/entry.ts
```diff
@@ -53,7 +53,9 @@ export async function getEntry(
 	const directory = process.cwd();
 	const entryPoint = config.site?.["entry-point"];
 
-	let paths: { absolutePath: string; relativePath: string } | undefined;
+	let paths:
+		| { absolutePath: string; relativePath: string; directory?: string }
```
petebacondarwin, line 57:
```suggestion
		| { absolutePath: string; relativePath: string; projectRoot?: string }
```

packages/wrangler/src/__tests__/get-entry.test.ts
```diff
@@ -0,0 +1,128 @@
...
+	runInTempDir();
+	mockConsoleMethods();
+	it.each([
```
petebacondarwin, line 20: I love that you added these tests, but if I am honest, using `it.each()` here actually obscures what the tests are doing and doesn't save much boilerplate compared to writing out a bunch of separate it clauses, given how simple the test code is.

penalosa:
@petebacondarwin I've renamed `directory` to `projectRoot` in the `Entry` type & SDW types, which should clarify things across the codebase. This is a breaking change to the SDW interface, which is fine because that's currently exposed as an experimental API

petebacondarwin (approved):
Cripes that rename had more blast radius than I thought!
Thanks.

A few unused imports to remove?

GregBrimble:
Haven't tested, but this maybe fixes #5481?

penalosa:
> #5481

It shouldn't I don't think—that looks older than this regression? Taking a look

penalosa:
@GregBrimble it looks like it might be a workspaces thing? https://github.com/cloudflare/workers-sdk/issues/5481#issuecomment-2469587208

petebacondarwin (approved):

PR #7290, opened 2024-11-19
fix: allow assets in multiworker setup

emily-shen:
#7251 will allow running multiple workers in one Miniflare instance. Currently we only set up the router/asset worker for the first worker passed to miniflare, assuming this will be the only user worker. This PR will make sure assets works regardless of how many 'user workers' are passed in and at what point.

---

- Tests
  - [ ] TODO (before merge): once #7251 is in. 
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: I don't think there are any existing docs covering this.

### DIFF
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -1282,7 +1282,6 @@ export class Miniflare {
 					address,
 					service: {
 						name: workerOpts.assets.assets
-							? `${ROUTER_SERVICE_NAME}-${workerName}`
 							: getUserServiceName(workerName),
 						entrypoint: entrypoint === "default" ? undefined : entrypoint,
 					},
@@ -1305,7 +1304,7 @@ export class Miniflare {
 				!this.#workerOpts[0].core.name?.startsWith(
 					"vitest-pool-workers-runner-"
 				)
-					? `${ROUTER_SERVICE_NAME}-${this.#workerOpts[0].core.name}`
+					? `${ROUTER_SERVICE_NAME}:${this.#workerOpts[0].core.name}`
 					: getUserServiceName(this.#workerOpts[0].core.name),
 			loopbackPort,
 			log: this.#log,
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -1130,14 +1130,6 @@ export class Miniflare {
 			innerBindings: Worker_Binding[];
 		}[] = [];
 
-		if (this.#workerOpts[0].assets.assets) {
-			// This will be the UserWorker, or the vitest pool worker wrapping the UserWorker
-			// The asset plugin needs this so that it can set the binding between the RouterWorker and the UserWorker
-			// TODO: apply this to ever this.#workerOpts, not just the first (i.e this.#workerOpts[0])
-			this.#workerOpts[0].assets.assets.workerName =
-				this.#workerOpts[0].core.name;
-		}
-
 		for (let i = 0; i < allWorkerOpts.length; i++) {
 			const previousWorkerOpts = allPreviousWorkerOpts?.[i];
 			const workerOpts = allWorkerOpts[i];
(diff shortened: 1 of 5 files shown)


packages/miniflare/src/index.ts
```diff
@@ -1341,7 +1341,39 @@ export class Miniflare {
...
+						if (
+							targetService.worker?.bindings?.some((binding) => {
+								if ("service" in binding) {
+									return binding.service?.name?.startsWith(
+										"assets:assets-service"
+									);
+								}
+							})
+						) {
+							// if target worker does have assets, set binding to point to router worker for that worker ("assets:router-targetWorkerName"), not the user worker itself ("core:user:targetWorkerName")
```
penalosa, line 1368: Is that how this works in production? I'd have expected service bindings to bypass the router worker

emily-shen: yeah production behaviour seriously confused me at first but fetch on a service binding does definitely hit assets, even if it is a binding to a named entrypoint. 





  | prod (ie expected behaviour)
-- | --
Simple fetch handler | assets / UW
Fetch on default worker entrypoint | Assets / UW
RPC on default entrypoint | UW only
Fetch on named entrypoint | Assets / UW's named entrypoint
RPC on named entrypoint | UW only

penalosa: Could we add some tests for this? i.e. in https://github.com/cloudflare/workers-sdk/blob/80a83bb4708ea2e3d4b514d3ebc97aa40c14439c/packages/wrangler/e2e/multiworker-dev.test.ts. In particular it would be good to have a test that validates that RPC service bindings bypass Assets

emily-shen: 47a09dab086217fc873658567cde510f914c9a2a

packages/wrangler/e2e/multiworker-dev.test.ts
```diff
@@ -446,4 +445,357 @@ describe("multiworker", () => {
...
+
+				const { url } = await worker.waitForReady(5_000);
+				await expect(fetchText(`${url}/rpc`)).resolves.toContain(
+					`TypeError: The RPC receiver does not implement the method`
+				);
+			});
+			// wrangler dev just crashses because the named entrypoint does not exist on the router worker
+			// expected behaviour: you should be able to hit the user worker only on named entrypoints
+			// and never get assets (currently this does in production but is a bug)
+			it("named entrypoints will not work", async () => {
```
penalosa, line 731: If possible, these should be written as tests that check the correct behaviour, with `it.fails()`

penalosa (approved):
Some small comments, but overall 🎉

packages/miniflare/src/index.ts
```diff
@@ -1340,7 +1340,39 @@ export class Miniflare {
 					"Ensure wrapped bindings don't have bindings to themselves."
 			);
 		}
-
+		// correct the bindings for bindings to workers with assets
+		// doing this here because all services + bindings will have been set up and because we need the config for all workers
+		for (const service of servicesArray) {
```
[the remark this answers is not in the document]
emily-shen: whoops got stuck trying to that before, had another look now and it was totally fine 😅 
90f8e7b663b26b0233f844dab159e98b8aa3ba1f

packages/wrangler/e2e/multiworker-dev.test.ts
emily-shen: eb4cb620475cb693ad2f3d33c6e4f86d43e8299e

packages/miniflare/src/index.ts
```diff
@@ -1340,7 +1340,39 @@ export class Miniflare {
...
+						// check if target worker has assets
+						const targetService = servicesArray.find(
+							(s) => s.name === binding.service?.name
+						);
+						if (!targetService || !("worker" in targetService)) continue;
+						if (
+							targetService.worker?.bindings?.some((binding) => {
+								if ("service" in binding) {
+									return binding.service?.name?.startsWith(
+										"assets:assets-service"
```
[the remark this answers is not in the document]
emily-shen: 90f8e7b663b26b0233f844dab159e98b8aa3ba1f

PR #7299, opened 2024-11-20
feat(cloudflare_pages): add `CI` detection on deploy

oritwoen:
This pull request adds support for autofilling parameters when deploying cloudflare pages using CI. Currently on GitLab I have to do something like this:

`npx wrangler pages deploy dist --project-name cf-pages-project --commit-hash "$CI_COMMIT_SHORT_SHA" --commit-message "$CI_COMMIT_MESSAGE" --branch $CI_COMMIT_BRANCH`
---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Deploy already contains tests and if it doesn't work correctly the current tests will detect it.
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Deploy already contains tests and if it doesn't work correctly the current tests will detect it.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: It should have worked this way from the beginning.

### DIFF
--- a/packages/wrangler/package.json
+++ b/packages/wrangler/package.json
@@ -87,7 +87,8 @@
 		"source-map": "^0.6.1",
 		"unenv": "npm:unenv-nightly@2.0.0-20241111-080453-894aa31",
 		"workerd": "1.20241106.1",
-		"xxhash-wasm": "^1.0.1"
+		"xxhash-wasm": "^1.0.1",
+		"std-env": "^3.8.0"
 	},
 	"devDependencies": {
 		"@cloudflare/cli": "workspace:*",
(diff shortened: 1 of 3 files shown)


petebacondarwin:
@oritwoen - thank you for this contribution.

- I'm not sure if there is a similar tool for gitlab but we have a Github Action that I believe could do this sort of thing - and if not that would be the place I would add such a feature.
- Also Pages is being superceded by Workers Assets so we are very unlikely to make this change to Pages.

Given the above, I am going to close this PR.

PR #7302, opened 2024-11-20
feat: add banner to indicate when telemetry is on

emily-shen:
Fixes [DEVX-1482](https://jira.cfdata.org/browse/DEVX-1482) in 56c4cfaeff287e56121b0994b0a1720b21704718 and 2d654bc2261272fbec4887a3acd285d57ddb1fb6
Screenshot 2024-11-20 at 13 24 58
(link TBD)

Fixes [DEVX-1506](https://jira.cfdata.org/browse/DEVX-1506) in de7ebf222c787e60e8badfa6f5626e37c3be7589

Also adds some basic tests for sendNewEvent in 6b2233853d691315684dc0ee7c811d2898fc5279

Also make sure events are settled before exiting wrangler (with a timeout) 3df6ab18491b6efbf287d017d8c0548ef35e173a

---
- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: merging into feature branch

### DIFF
--- a/packages/wrangler/src/metrics/metrics-dispatcher.ts
+++ b/packages/wrangler/src/metrics/metrics-dispatcher.ts
@@ -68,9 +68,7 @@ export function getMetricsDispatcher(options: MetricsConfigOptions) {
 			) {
 				return;
 			}
-			if (name === "wrangler command started") {
-				printMetricsBanner();
-			}
+			printMetricsBanner();
 			const commonEventProperties: CommonEventProperties = {
 				amplitude_session_id,
 				amplitude_event_id: amplitude_event_id++,
--- a/packages/wrangler/src/metrics/metrics-dispatcher.ts
+++ b/packages/wrangler/src/metrics/metrics-dispatcher.ts
@@ -21,6 +21,7 @@ export function getMetricsDispatcher(options: MetricsConfigOptions) {
 	// The SPARROW_SOURCE_KEY will be provided at build time through esbuild's `define` option
 	// No events will be sent if the env `SPARROW_SOURCE_KEY` is not provided and the value will be set to an empty string instead.
 	const SPARROW_SOURCE_KEY = process.env.SPARROW_SOURCE_KEY ?? "";
+	const requests: Array<Promise<void>> = [];
 	const wranglerVersion = getWranglerVersion();
 	const platform = getPlatform();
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/metrics/metrics-dispatcher.ts
```diff
@@ -43,13 +48,28 @@ export function getMetricsDispatcher(options: MetricsConfigOptions) {
...
 		): Promise<void> {
+			if (
+				properties.command === "wrangler telemetry disable" ||
+				properties.command === "wrangler metrics disable"
+			) {
+				return;
+			}
+			if (name === "wrangler command started") {
+				printMetricsBanner();
+			}
```
edmundhung, line 72: I think we can remove the condition as the current logic should be safe to make sure it print once only. This also avoids the assumption that `wrangler command started` event to always be the first event.

emily-shen: ce38f17b04457a39cf03a48ec53fbcfd2d8b3e40 sounds good 👍

packages/wrangler/src/metrics/metrics-dispatcher.ts
```diff
@@ -112,13 +135,49 @@ export function getMetricsDispatcher(options: MetricsConfigOptions) {
...
+		const storedSegment = parseInt(storedVersion[i]);
+		const currentSegment = parseInt(currentVersion[i]);
+		if (currentSegment > storedSegment) {
+			return true;
+		} else if (currentSegment < storedSegment) {
+			return false;
+		}
+	}
+	return false;
+};
```
edmundhung, line 183: This might not work with pre-release version (e.g. `0.0.0-76d43d8b1`). What if we don't care about whether it is a newer version but showing the banner whenever the version is changed?

emily-shen: [57eca6c](https://github.com/cloudflare/workers-sdk/pull/7302/commits/57eca6ccfd98cfe174537d2ec59ff1195f2af952)

edmundhung (approved):

Issue #7108, opened 2024-10-27
🐛 BUG: Unnecessary polyfills?


Cherry:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What version(s) of the tool(s) are you using?

3.83.0

### What version of Node are you using?

20.18.0

### What operating system and version are you using?

Windows 11

### Describe the Bug

With a worker like this:
```ts
export default {
	async fetch(request, env, ctx): Promise<Response> {
		return new Response('Hello World!');
	},
} satisfies ExportedHandler<Env>;
```

when using `nodejs_compat`, the output size is:
`Total Upload: 25.40 KiB / gzip: 6.13 KiB`

vs when I use `nodejs_compat` and `no_nodejs_compat_v2`:
`Total Upload: 0.19 KiB / gzip: 0.16 KiB`

The output when using v2 mode seems to include polyfills for `process`, `console`, and much more, drastically inflating my bundle size, despite not using any of these in my worker:
```js
function __cf_cjs(esm) {
  const cjs = 'default' in esm ? esm.default : {};
	for (const [k, v] of Object.entries(esm)) {
		if (k !== 'default') {
			Object.defineProperty(cjs, k, {
				enumerable: true,
				value: v,
			});
		}
	}
	return cjs;
}
var __defProp = Object.defineProperty;
var __defNormalProp = (obj, key, value) => key in obj ? __defProp(obj, key, { enumerable: true, configurable: true, writable: true, value }) : obj[key] = value;
var __name = (target, value) => __defProp(target, "name", { value, configurable: true });
var __publicField = (obj, key, value) => {
  __defNormalProp(obj, typeof key !== "symbol" ? key + "" : key, value);
  return value;
};

// node_modules/wrangler/_virtual_unenv_global_polyfill-clear$immediate.js
globalThis.clearImmediate = clearImmediateFallback;

// node_modules/unenv/runtime/_internal/utils.mjs
function createNotImplementedError(name) {
  return new Error(`[unenv] ${name} is not implemented yet!`);
}
__name(createNotImplementedError, "createNotImplementedError");
function notImplemented(name) {
  const fn2 = /* @__PURE__ */ __name(() => {
    throw createNotImplementedError(name);
  }, "fn");
  return Object.assign(fn2, { __unenv__: true });
}
__name(notImplemented, "notImplemented");

// node_modules/unenv/runtime/mock/noop.mjs
var noop_default = Object.assign(() => {
}, { __unenv__: true });

[...]

PR #7303, opened 2024-11-20
Modify router-worker logic to allow for running user-worker ahead of assets

WillTaylorDev:
This PR provides the option for user worker to run ahead of assets.
- Adds logic to run user-worker ahead of assets
- Moves vitest from worker-shared root into asset-worker
- Creates new vitest unit test suite for router-worker

---

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: not user facing through this mechanism

### DIFF
--- a/packages/workers-shared/package.json
+++ b/packages/workers-shared/package.json
@@ -30,14 +30,16 @@
 		"bundle:router-worker:prod": "pnpm run bundle:router-worker --minify",
 		"check:lint": "eslint . --max-warnings=0",
 		"check:type": "pnpm run check:type:tests && tsc",
-		"check:type:tests": "tsc -p ./asset-worker/tests/tsconfig.json",
+		"check:type:tests": "tsc -p ./asset-worker/tests/tsconfig.json && tsc -p ./router-worker/tests/tsconfig.json",
 		"clean": "rimraf dist",
 		"deploy": "pnpm run deploy:router-worker && pnpm run deploy:asset-worker",
 		"deploy:asset-worker": "CLOUDFLARE_API_TOKEN=$WORKERS_DEPLOY_AND_CONFIG_CLOUDFLARE_API_TOKEN pnpx wrangler versions upload --experimental-versions -c asset-worker/wrangler.toml",
 		"deploy:router-worker": "CLOUDFLARE_API_TOKEN=$WORKERS_DEPLOY_AND_CONFIG_CLOUDFLARE_API_TOKEN pnpx wrangler versions upload --experimental-versions -c router-worker/wrangler.toml",
 		"dev": "pnpm run clean && concurrently -n bundle:asset-worker,bundle:router-worker -c blue,magenta \"pnpm run bundle:asset-worker --watch\" \"pnpm run bundle:router-worker --watch\"",
-		"test": "vitest",
-		"test:ci": "vitest run",
+		"test": "concurrently --group -n router-worker,asset-worker \"pnpm run test:router-worker\" \"pnpm run test:asset-worker\"",
+		"test:asset-worker": "vitest -c asset-worker/vitest.config.mts --dir asset-worker",
+		"test:ci": "pnpm run test",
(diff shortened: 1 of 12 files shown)


WillTaylorDev:
tsc seems angry-- is this broken already, or did I heck something up with the vitest changes?

image

packages/workers-shared/router-worker/__tests__/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+	"compilerOptions": {
+		"target": "ES2020",
+		"module": "ES2020",
+		"lib": ["ES2020"],
+		"types": [
+			"@cloudflare/workers-types/experimental",
+			"@cloudflare/vitest-pool-workers"
+		],
```
petebacondarwin, line 9: I think you need to add `"@types/node"` in here to fix TS checks.

WillTaylorDev: This did the trick, but actually we needed to change it in workers-shared. Still perplexed because I didn't touch any of the code in the root of workers-shared (the utils/helper.ts and script/* file already existed)

CarmenPopoviciu: seems like the ts checks failures are caused by the fact that `vitest.config.mts` was moved out of the root folder...whyyy would that break TS checks o.O????

packages/workers-shared/package.json
```diff
@@ -46,6 +48,7 @@
 	},
 	"devDependencies": {
 		"@cloudflare/eslint-config-worker": "workspace:*",
+		"@cloudflare/vitest-pool-workers": "latest",
```
GregBrimble, line 51: How do y'all in ANT feel about us pulling from npm here rather using the workspace version? Circular dependency otherwise.

packages/workers-shared/router-worker/wrangler.toml
```diff
@@ -12,6 +12,7 @@ account_id = "0f1b8aa119a907021f659042f95ea9ba"
 workers_dev = false
 main = "src/index.ts"
 compatibility_date = "2024-07-31"
+compatibility_flags = ["nodejs_compat"]
```
petebacondarwin, line 15: Can you add a comment explaining that this is only here because of Vitest?
It makes me said that we are turning this on for a Worker that is deeply embedded in the standard pipeline for many of our customer's projects going forward.

packages/workers-shared/router-worker/tests/index.test.ts
```diff
@@ -0,0 +1,66 @@
...
+		const env = {
+			CONFIG: {
+				invoke_user_worker_ahead_of_assets: true,
+				has_user_worker: true,
+			},
+			USER_WORKER: {
+				async fetch(_: Request): Promise<Response> {
+					return new Response("hello from user worker");
+				},
+			},
```
petebacondarwin, line 37: Could you add an ASSET_WORKER in here too, just to prove that having an ASSET_WORKER does not force the request to go there rather than the USER_WORKER.

petebacondarwin (approved):
Non-blocking comments

packages/workers-shared/router-worker/wrangler.toml
Cherry: It might make sense to also add `no_nodejs_compat_v2` here to avoid filling this worker with unnecessary global polyfills from `process`, `console`, etc.: https://github.com/cloudflare/workers-sdk/issues/7108

WillTaylorDev: Thanks both! Updated PR based on comments :)

packages/workers-shared/package.json
WillTaylorDev: Think this is good for now, we can always revisit if it becomes an issue

PR #7315, opened 2024-11-21
Update local AI fetcher to forward method and pathname to upstream

G4brym:
This pr starts forwarding method and pathname to upstream.
This is not a breaking change, because the AI binding currently just does POST requests. [workerd here](https://github.com/cloudflare/workerd/blob/main/src/cloudflare/internal/ai-api.ts#L93)
---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because:
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/ai/fetcher.ts
```diff
@@ -16,12 +16,14 @@ export default function (env) {
 export async function AIFetcher(request: Request): Promise<Response> {
 	const accountId = await getAccountId();
 
-	request.headers.delete("Host");
-	request.headers.delete("Content-Length");
+	const reqHeaders = new Headers(request.headers);
+	reqHeaders.delete("Host");
+	reqHeaders.delete("Content-Length");
+	reqHeaders.set("x-url", request.url);
```
petebacondarwin, line 22: I think there is a standard one can use for proxying requests like this...

petebacondarwin: Oh I was thinking of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

petebacondarwin (commented):
Is there no test you can update or create for this behaviour?

G4brym:
@petebacondarwin sure, let me add a test and get back to you
Regarding the header name, i don't think "X-Forwarded-Host" fits well, because we are forwarding the complete url, not just the hostname, i guess nothing stops us from doing it anyway, it might just be confusing for someone new
The part we are really interested is in the utl pathname + querystring, so i thought i might as well send the complete url

packages/wrangler/src/ai/fetcher.ts
```diff
@@ -16,12 +16,14 @@ export default function (env) {
 export async function AIFetcher(request: Request): Promise<Response> {
 	const accountId = await getAccountId();
 
-	request.headers.delete("Host");
-	request.headers.delete("Content-Length");
+	const reqHeaders = new Headers(request.headers);
+	reqHeaders.delete("Host");
+	reqHeaders.delete("Content-Length");
+	reqHeaders.set("X-Forwarded-Host", request.url);
```
petebacondarwin, line 22: Just to be clear I am not sure that `X-Forwarded-Host` is the correct thing here. It is what I was thinking of but I think is only for hosts not full URLs.

Do you have a requirement for this header on the server?
What is the header it is looking for?

G4brym: i'm writing the server part as well, so we are free to decide what we like best 😄 
the only requirement i have is for this local fetcher to send the full path and query string to the server, so any header name does the job

petebacondarwin:
Still LGTM!

PR #7316, opened 2024-11-21
Add serve_directly option to assets configuration

WillTaylorDev:
Provides serve_directly option in asset metadata

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: not yet implemented on backend

### DIFF
--- a/.changeset/six-coats-rule.md
+++ b/.changeset/six-coats-rule.md
@@ -2,4 +2,6 @@
 "wrangler": minor
 ---
 
-Added serve_directly option to assets
+feat: add `serve_directly` option to Workers with Assets
+
+Users can now specify whether their assets are served directly against HTTP requests or whether these requests always go to the Worker, which can then respond with asset retrieved by its assets binding.
--- a/.changeset/six-coats-rule.md
+++ b/.changeset/six-coats-rule.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Added serve_directly option to assets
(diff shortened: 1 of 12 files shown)


GregBrimble (approved):

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4950,6 +4950,82 @@ addEventListener('fetch', event => {});`
 			await runWrangler("deploy");
 		});
 
+		it("serve_directly correctly overrides default if set to false", async () => {
```
GregBrimble, line 4953: I know it's tested in all the other blocks implicitly, but might be nice to have an explicit check for undefined config being *omitted* in the request body. If it ever regressed and defaulted to false, then that would be a bug.

WillTaylorDev: I actually did have a test for it and then deleted after thinking "oh it's tested everywhere else" 😆  no problem, will add an explicit one!

.changeset/six-coats-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Added serve_directly option to assets
```
petebacondarwin, line 5: The format is a type followed by a present tense description of what is changing from the point of view of someone reading the changelog.

```suggestion
feat: add `serve_directly` option to Workers with Assets

Users can now specify whether their assets are served directly against HTTP requests or whether these requests always go to the Worker, which can then respond with asset retrieved by its assets binding.
```

Something like that.

petebacondarwin (approved):

PR #7317, opened 2024-11-21
test: fix system locale override

ns476:
Apparently setting `LC_ALL` in `vitest.setup.ts` doesn't work, as this is after the locale settings are read, but it does work in a `globalSetup` file.

This can be observed locally with the Workflows tests:
```
➜  wrangler git:(fix-test-locale) npx vitest run src/__tests__/workflows.test.ts >&/dev/null; echo $?
0
➜  wrangler git:(fix-test-locale) git checkout origin/main vitest.config.mts && LC_ALL=en_UK npx vitest run src/__tests__/workflows.test.ts >&/dev/null; echo $?
Updated 1 path from e1554242a
1
```

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: small improvements to tests, though I guess running with `en_UK` in CI would help verify this?
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Small test improvement
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Not public facing

### DIFF
--- a/packages/wrangler/src/__tests__/vitest.global.ts
+++ b/packages/wrangler/src/__tests__/vitest.global.ts
@@ -1,0 +1,7 @@
+export function setup(): void {
+	// Set `LC_ALL` to fix the language as English for the messages thrown by Yargs,
+	// and to make any uses of datetimes in snapshots consistent.
+	// This needs to be in a globalSetup script - it won't work in a setupFile script.
+	// https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1439286286
+	process.env.LC_ALL = "C";
+}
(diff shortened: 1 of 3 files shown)


edmundhung:
I tried running `npx vitest run src/__tests__/workflows.test.ts` locally and it fails regardless of this change 🤔 

@petebacondarwin does it work for you?

petebacondarwin:
Oops only just saw your comment @edmundhung ...

petebacondarwin:
Trying it locally. Hopefully this is not a CI cache hit false positive...

petebacondarwin:
It passes locally for me...

PR #7318, opened 2024-11-21
Normalize workers assets path

WillTaylorDev:
Prevents same-schema redirect attacks by collapsing multiple slashes in the path.

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/fixtures/asset-config/redirects.test.ts
+++ b/fixtures/asset-config/redirects.test.ts
@@ -1,0 +1,89 @@
+import { SELF } from "cloudflare:test";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { applyConfigurationDefaults } from "../../packages/workers-shared/asset-worker/src/configuration";
+import Worker from "../../packages/workers-shared/asset-worker/src/index";
+import { getAssetWithMetadataFromKV } from "../../packages/workers-shared/asset-worker/src/utils/kv";
+import type { AssetMetadata } from "../../packages/workers-shared/asset-worker/src/utils/kv";
+
+const IncomingRequest = Request<unknown, IncomingRequestCfProperties>;
+
+vi.mock("../../packages/workers-shared/asset-worker/src/utils/kv.ts");
+vi.mock("../../packages/workers-shared/asset-worker/src/configuration");
+const existsMock = (fileList: Set<string>) => {
+	vi.spyOn(Worker.prototype, "unstable_exists").mockImplementation(
+		async (pathname: string) => {
+			if (fileList.has(pathname)) {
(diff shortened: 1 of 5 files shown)


packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -96,4 +96,46 @@ describe("[Asset Worker] `handleRequest`", () => {
...
+		expect(response.headers.get("location")).toBe("/abc/def/123/456");
+
+		response = await handleRequest(
+			new Request("https://example.com/%2fexample.com/"),
+			configuration,
+			exists,
+			getByETag
+		);
+
+		expect(response.status).toBe(307);
```
WillTaylorDev, line 128: These should be 200s. Will probably lose the location header when that updates.

fixtures/asset-config/redirects.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+		vi.mocked(applyConfigurationDefaults).mockImplementation(() => {
+			return {
+				html_handling: "none",
+				not_found_handling: "none",
+			};
+		});
+	});
+
+	it("returns 404 for non matched encoded url", async () => {
```
WalshyDev, line 48: i'd probably have the normalisation stuff in it's own test to verify how it matches fully, bit weird it's in an encoded focused test

e.g. `/%2fexample.com/` -> 404
e.g. `/%2fblog/index.html` -> 200 `/blog/index.html`
etc

WillTaylorDev: Added a url-normalization test file to test those specific cases. I think we still want the redirect tests, but you're right they can absolutely be split out.

CarmenPopoviciu (approved):

edmundhung (approved):

fixtures/asset-config/url-normalization.test.ts
```diff
@@ -0,0 +1,77 @@
...
+		async (pathname: string) => {
+			if (fileList.has(pathname)) {
+				return pathname;
+			}
+		}
+	);
+};
+const BASE_URL = "http://example.com";
+
+describe("[Asset Worker] `test redirects`", () => {
```
petebacondarwin, line 23: redirects? or rewrites?

WillTaylorDev: Good catch, I'll update in my next PR shortly!

PR #7326, opened 2024-11-22
Print copyable wrangler.toml text when creating a hyperdrive config

OilyLime:
Fixes #[SQC-348].

This change gives wrangler users a quality of life improvement when creating a Hyperdrive config, giving them a copyable snippet to paste in their `wrangler.toml` file to bind their new Hyperdrive Config to their Worker.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Does not change logic around creating or updating Hyperdrive configs, only printing the results.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Does not change the API surface for the Hyperdrive command.

### DIFF
--- a/packages/wrangler/src/hyperdrive/create.ts
+++ b/packages/wrangler/src/hyperdrive/create.ts
@@ -1,5 +1,4 @@
-import TOML from "@iarna/toml";
-import { readConfig } from "../config";
+import { configFileName, formatConfigSnippet, readConfig } from "../config";
 import { logger } from "../logger";
 import { createConfig } from "./client";
 import { getCacheOptionsFromArgs, getOriginFromArgs, upsertOptions } from ".";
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/hyperdrive/create.ts
```diff
@@ -35,6 +36,10 @@ export async function handler(
 	});
 	logger.log(
 		`✅ Created new Hyperdrive config\n`,
-		JSON.stringify(database, null, 2)
+		TOML.stringify({
```
penalosa, line 39: This should be using the new `formatConfigSnippet()` helper introduced in https://github.com/cloudflare/workers-sdk/pull/7230

penalosa (changes requested):
This should support both TOML & JSON

sejoker:
> This should support both TOML & JSON

this is now addressed.

penalosa (approved):

PR #7334, opened 2024-11-25
Wrangler v4

penalosa:
Wrangler v4 is coming! We’re releasing this release candidate now as a way for you to check the upcoming version works with your projects. TLDR of the breaking changes to expect are:

- Node v18 is now the minimum Node version
- Wrangler now uses `esbuild` v0.24.2
- `wrangler r2` and `wrangler kv` are `--local` by default
- All deprecated commands & config have been removed

A detailed migration guide is available at https://developers.cloudflare.com/workers/wrangler/migration/update-v3-to-v4

To try it out, install and use `wrangler@next`.

Let us know what you think! We'd love to hear any feedback, either as comments on this PR or in the #wrangler channel on Discord.


emily-shen (commented):
nit: can we add a pre-requisites section to the wrangler readme that mentions this

fixtures/node-app-pages/tests/index.test.ts
emily-shen, line 1: the package.json for this fixture still uses the `--node-compat` flag

emily-shen (commented):
the node app pages package.json still uses the `--node-compat` flag

.changeset/funny-pets-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": major
+---
+
+Use `--local` by default for KV & R2
```
emily-shen, line 5: > Use `--local` by default for`kv key` and `r2 object` commands

since kv namespace etc. are _not_ local - seen a fair number of people getting confused so we should probably be careful about phrasing

packages/wrangler/README.md
```diff
@@ -8,10 +8,6 @@
 
 `wrangler` is a command line tool for building [Cloudflare Workers](https://workers.cloudflare.com/).
 
-> [!WARNING]
```
emily-shen, line 11: do we have a similar (but less scary) message anywhere for v3?

penalosa: No, and I kinda don't think we should. I feel like the NPM readme is targeted at new users, who would be a bit put off at seeing a warning as their first interaction with Wrangler?

.changeset/funny-pets-punch.md
penalosa, line 5:
```suggestion
Use `--local` by default for `wrangler kv key` and `wrangler r2 object` commands
```

penalosa:
Remaining test failures should be resolved by https://github.com/cloudflare/next-on-pages/pull/940

packages/wrangler/src/r2/object.ts
```diff
@@ -272,6 +288,7 @@ export const r2ObjectPutCommand = createCommand({
 		);
 
 		if (local) {
+			logger.warn(remoteFlagWarning);
```
emily-shen, line 291: nit: logger.info?
don't want to annoy people with a big warning every time, and i feel that using local by accident is unlikely to be a big deal (unlike the reverse). don't feel very strongly about this though.

packages/wrangler/src/init.ts
```diff
@@ -50,18 +41,6 @@ export function initOptions(yargs: CommonYargsArgv) {
...
-			hidden: true,
-			deprecated: true,
-		})
-		.option("site", {
-			hidden: true,
-			type: "boolean",
-			deprecated: true,
-		})
 		.option("yes", {
 			describe: 'Answer "yes" to any prompts for new projects',
```
emily-shen, line 45: should the delegate-c3 arg not be removed?

packages/wrangler/src/index.ts
```diff
@@ -906,72 +882,11 @@ export function createCLIParser(argv: string[]) {
 	]);
 	registry.registerNamespace("check");
 
-	/******************************************************/
-	/*               DEPRECATED COMMANDS                  */
-	/******************************************************/
-	// [DEPRECATED] build
 	wrangler.command("build", false, buildOptions, buildHandler);
```
emily-shen, line 885: are we keeping this one? undeprecating it even?

emily-shen (commented):
a couple of places to update:

- error message in `getBindings` in wrangler/src/dev.ts mentions kv:namespace create
- workers-chat-demo fixture has [build.upload] in its config
- legacy assets is mentioned in an error message in versions/secrets/index.ts in `parseModules()`

packages/wrangler/src/index.ts
penalosa: Yes, we're undeprecating build (arguably this doesn't need to go in a major but 🤷 )

packages/wrangler/src/init.ts
penalosa: No, because C3 uses it when calling back to wrangler to get existing worker files (`wrangler init --from-dash` -> C3 -> `wrangler init --from-dash --no-delegate-c3`)

packages/wrangler/src/r2/object.ts
penalosa: This was an ask from the R2 team—we can definitely refine over time, but I think it's worth keeping as a warning for now. Changing this to `.info()` in e.g. 4.2 would maybe make sense?

.changeset/major-remove-version.md
```diff
@@ -0,0 +1,5 @@
+---
```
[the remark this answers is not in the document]
penalosa: We could, but if we did that I'm not sure where we'd document it?

(this also isn't a breaking change, so we could do this outside of the v4 scope)

.changeset/major-remove-d1-backups.md
```diff
@@ -0,0 +1,11 @@
+---
+"wrangler": major
+---
+
+Remove `wrangler d1 backups`
+
+BREAKING CHANGE: This change removes `wrangler d1 backups`, a set of alpha-only commands that would allow folks to interact with backups of their D1 alpha DBs.
```
emily-shen, line 7: nit: in context, having BREAKING CHANGE on just a couple of the breaking changes could be a bit confusing, maybe we could remove this from these changesets? 

.changeset/major-remove-d1-backups.md
.changeset/major-remove-d1-deprecated-batch-size.md
.changeset/major-remove-d1-migrations-alpha.md

.changeset/major-remove-legacy-assets.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": major
+---
+
+Remove legacy assets
```
emily-shen, line 5: > Remove the --legacy-assets flag. 

(and config key if there was one?)

packages/kv-asset-handler/package.json
```diff
@@ -50,7 +50,7 @@
 		"service-worker-mock": "^2.0.5"
 	},
 	"engines": {
-		"node": ">=16.13"
+		"node": ">=18.0.0"
```
petebacondarwin, line 53: We should bump the version of this package to 0.4.0 as this is arguably a breaking change.

.changeset/node-compat-v4.md
```diff
@@ -0,0 +1,8 @@
+---
+"wrangler": major
+"miniflare": major
```
petebacondarwin, line 3: add "@cloudflare/vite-plugin": patch since we remove the legacy node_compat check there.

packages/quick-edit-extension/package.json
```diff
@@ -44,7 +44,7 @@
 	"devDependencies": {
 		"@cloudflare/workers-tsconfig": "workspace:^",
 		"@cloudflare/workers-types": "^4.20250224.0",
```
petebacondarwin, line 46: Revert this line?

.changeset/node-compat-v4.md
petebacondarwin: May not be necessary since I think that vite-plugin has a dependency on wrangler?

packages/workers-shared/asset-worker/tests/assets-manifest.test.ts
```diff
@@ -12,6 +13,13 @@ import { binarySearch, compare, hashPath } from "../src/assets-manifest";
 
 const encoder = new TextEncoder();
 
+beforeAll(() => {
+	if (!("crypto" in globalThis)) {
+		// @ts-expect-error will go away once Node.js is bumped
```
petebacondarwin, line 18: is bumped to what? 20?

.changeset/slimy-pumpkins-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+wrangler: major
```
petebacondarwin, line 2: Minor version bump to kv-asset-handler, workers-shared, workflows-shared
Major bump to miniflare

.github/workflows/test-and-check.yml
```diff
@@ -120,7 +120,7 @@ jobs:
           - os: ubuntu-24.04-arm
             os_name: v22, Linux
             node_version: 22
-            filter: '--filter="./packages/*"'
+            filter: '--filter="./packages/*" --filter="!./packages/kv-asset-handler"'
```
petebacondarwin, line 123: Do we have a plan for this when we get to Node 22?

packages/wrangler/src/dev/remote.ts
```diff
@@ -114,7 +114,7 @@ export async function createRemoteWorkerInit(props: {
 		props.modules
 	);
 
-	const legacyAssets = await syncLegacyAssets(
+	const legacyAssets = await syncWorkersSite(
```
petebacondarwin, line 117: NIT: rename `legacyAssets` to `workersSitesAssets`?

packages/kv-asset-handler/package.json
penalosa: b9b42f725

packages/workers-shared/asset-worker/tests/assets-manifest.test.ts
penalosa: c9a6e500c

packages/kv-asset-handler/package.json
penalosa: b9b42f725

.changeset/slimy-pumpkins-tickle.md
penalosa: 8d13f427a

.github/workflows/test-and-check.yml
penalosa: Yes, we'll need to migrate the `kv-asset-handler` tests to use the Vitest integration (right now they run in Node with some global scope mocking)

packages/wrangler/src/__tests__/route.test.ts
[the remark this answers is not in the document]
penalosa: 122adf88c

packages/wrangler/src/dev/remote.ts
penalosa: 2bc289e2e

packages/wrangler/src/r2/object.ts
```diff
@@ -210,12 +226,12 @@ export const r2ObjectPutCommand = createCommand({
 			objectPath,
 			file,
 			pipe,
-			local,
 			persistTo,
 			jurisdiction,
 			storageClass,
 			...options
 		} = objectPutYargs;
+		const local = !objectPutYargs.remote;
```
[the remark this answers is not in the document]
penalosa: fa7e4da92

emily-shen (commented):
This ship has probably already sailed, but the more i think about it, the more hesitant i am about making R2 and KV commands local by default. With R2 in particular, most commands don't have --local, so the default state becomes more inconsistent within a resource - the three object commands are now local, but the vast majority (20+) still operate on remote resources.

The same applies to KV, although less drastically. 

Crucially, the commands you start with - resource creation - are still only remote. I would guess that moving from creating a resource -> operations on that resource is a more commonly encountered scenario than moving between different resources.

I think it would actually make more sense to make _D1_ remote by default, if we really cared about consistency between resources commands, although the easiest solution is probably to just drop this change. Ultimately I think consistency _within_ a set of resource commands is way more important though, and making r2 object and kv key commands local by default are introducing much more inconsistency. 

As a side note, when we do mixed mode again, we should also add local/remote indicators for all resource commands like we did for dev.

Sorry to throw a spanner in the works at this late stage!

emily-shen:
For posterity, I've been convinced this is alright :)
-  if we land provisioning people won't have to run create commands
- we can/should print an `executing on local / remote` message for all resource commands anyway
- accidentally running something locally first probably isn't going to break anything, unlike remote first defaults.

PR #7338, opened 2024-11-25
Require Node v18 in Wrangler v4

penalosa:
Bump minimum supported Node version to Node v18 in Wrangler v4

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: will be part of v4 docs


.github/workflows/test-and-check.yml
```diff
@@ -102,10 +102,21 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-13]
         filter: ["./packages/*", "./fixtures/*"]
+        node_version: [18]
```
emily-shen, line 105: I think CI is pulling something like v18.20.2 which is too old - might need to make these more specific.

andyjessop (approved):

emily-shen:
also there's a line in the c3 nuxt templates that might need to be updated?
`writeFile("./.node-version", "17");`

.github/workflows/test-and-check.yml
penalosa: I've loosened this up a bit to only require v18

penalosa:
> also there's a line in the c3 nuxt templates that might need to be updated? `writeFile("./.node-version", "17");`

I think we can fix this without a major, so I'll open a separate PR

Issue #7341, opened 2024-11-25
🐛 BUG: npm create cloudflare@latest --framework angular has bug in initial template


neomax7:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What version(s) of the tool(s) are you using?

wrangler@3.90.0

### What version of Node are you using?

v20.18.1

### What operating system and version are you using?

Mac Sequoia 15.1.1

### Observed behavior
Please describe.
npm create cloudflare@latest --framework angular 
initialize a template for angular development using angular v19
the initialed template has typo in @/src/server.ts at line 8

`
instead of 
  const res = await angularApp.render(req);`
expect
  const res = await angularApp.handle(req);`
`
reference
[official doc reference](https://angular.dev/api/ssr/AngularAppEngine)

### Expected behavior
Please describe.
`npm run start`
work as intended, right now it fails

### Steps to reproduce
Please provide the following:
- A minimal working subset of your worker code
- A minimal working subset of your `wrangler.toml`
- Commands used to start your local dev server, including custom env and cli args
- Steps to be performed in the browser, curl commands, or a test we can run that reliably fails (at least a percent of the time)
`
npm create cloudflare@latest --framework angular 
npm run start fails
`
A git repo we can clone and run a test suite on, or which has a README with step-by-step instructions, is even better. In this case, please use the field below to provide a link to the minimal repro.

emily-shen:
fixed by https://github.com/cloudflare/workers-sdk/pull/7342

PR #7342, opened 2024-11-25
fix: error in angular c3 template

neomax7:
AngularAppEngine class does not have .render method, instead it should be .handle method

https://github.com/cloudflare/workers-sdk/issues/7341
_Describe your change..._
AngularAppEngine class does not have .render method, instead it should be .handle method

changed method name from ".render" to ".handle" as the official doc below
https://angular.dev/api/ssr/AngularAppEngine

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: only changing template for create-cloudflare with angular framework
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because:  only changing template for create-cloudflare with angular framework
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because:  only changing template for create-cloudflare with angular framework

### DIFF
--- a/.changeset/witty-dolphins-appear.md
+++ b/.changeset/witty-dolphins-appear.md
@@ -2,6 +2,6 @@
 "create-cloudflare": patch
 ---
 
-fixed: related to issues-7341
-AngularAppEngine class does not have '.render' method,
-instead it should be '.handle' method.
+fix: Angular template should call `handle()` method on `AngularAppEngine` not 'render()'
+
+Fixes #7341
--- a/.changeset/witty-dolphins-appear.md
+++ b/.changeset/witty-dolphins-appear.md
@@ -1,0 +1,7 @@
+---
+"create-cloudflare": patch
+---
+
+fixed: related to issues-7341
+AngularAppEngine class does not have '.render' method,
+instead it should be '.handle' method.
--- a/packages/create-cloudflare/templates/angular/c3.ts
+++ b/packages/create-cloudflare/templates/angular/c3.ts
@@ -28,14 +28,11 @@ const configure = async (ctx: C3Context) => {
 };
 
 async function installCFWorker() {
-	await installPackages(
-		["@cloudflare/workers-types", "@miniflare/tre@next", "wrangler@beta"],
-		{
-			dev: true,
-			startText: "Installing adapter dependencies",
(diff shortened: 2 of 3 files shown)


emily-shen (approved):
Thanks for fixing this 🎉

emily-shen:
Ah can you add a changeset actually? should be a patch,  you can run `npx changeset` to generate one. Thanks!

emily-shen:
> I am concerned that this didn't fail in CI before this fix. E.g. https://github.com/cloudflare/workers-sdk/actions/runs/12071631960/job/33663814631#step:5:196

@petebacondarwin was just looking into it! Seems like the generated `package.json` doesn't have a `preview` or `deploy` script, so c3 wasn't able to run tests (it still should've failed when there was no `deploy` though 🤔). The whole generated package.json for angular looks a bit wrong actually - was planning to open a separate PR to fix. Re this PR, I tested this one manually and it's fixed the referenced issue.

neomax7:
> Ah can you add a changeset actually? should be a patch, you can run `npx changeset` to generate one. Thanks!

added changeset patch

petebacondarwin:
Yes you are right. The problem is that there is no preview script. (We don't run deploy scripts in CI tests any more since they were quite flaky). I'll land this fix (which I also verified locally) and then add a preview script to ensure it is tested.

petebacondarwin:
I updated the tests to run locally and they fail because Angular seems to hang. Does a new C3 template (after this PR) workf for you @neomax7? I haven't managed to debug what is causing the hang.

neomax7:
> I updated the tests to run locally and they fail because Angular seems to hang. Does a new C3 template (after this PR) workf for you @neomax7? I haven't managed to debug what is causing the hang.

Hi @petebacondarwin 
image
Here is the screenshot I tried it on my laptop.
it seems working without any issue.
I tried it on mac-sequoia 15.1.1 with node v20.18.1 (lts/iron)

neomax7:
it must be node version.
the highlighted line must be the issue. 
this is same for the current released build of C3 tool. 
no prebuilt binaries for node v22 (lts/jod)
image

PR #7345, opened 2024-11-26
fix(wrangler): keypress event name is optional

edmundhung:
Fixes n/a.

We were incorrectly treating the `name` property of the Keypress event as required. This PR fixes the type issue and updates the implementation accordingly.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: bugfix
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/cli-hotkeys.ts
+++ b/packages/wrangler/src/cli-hotkeys.ts
@@ -62,17 +62,22 @@ export default function (
 	}
 
 	const unregisterKeyPress = onKeyPress(async (key) => {
-		let char = key.name.toLowerCase();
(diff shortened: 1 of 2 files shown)


penalosa (approved):

petebacondarwin (changes requested):
I believe we have tests for `cli-hotkeys`. Can we add a test to cover this case that was missed?

petebacondarwin (approved):
Thank you @penalosa !

PR #7357, opened 2024-11-27
chore: refactor defineCommand to use named imports

andyjessop:
As discussed internally, there have been some issues with refactoring existing commands over to the `defineCommand` utility, relating to side-effect imports. The issues stem from the interaction with the bundler, which reorders side-effect imports under some circumstances, and also creates a shared state between tests, which results in duplicate definitions.

This refactor ensures that commands must be named and imported as named imports, and ensures that there is no potential for shared state between test runs.

The main changes are:

1. `defineCommand` becomes `createCommand` and returns a command object that should be defined with the registry. (https://github.com/cloudflare/workers-sdk/pull/7357/commits/f70664c76dc23a6e075d35300c75288f305b8f4d)

2. Same with `defineNamespace` and `defineAlias` - they become `createNamespace` and `createAlias`.

3. Commands are defined on the registry by providing their full command name and the named definition, (https://github.com/cloudflare/workers-sdk/pull/7357/commits/8284eff65a802dae947c7a6b67ddda5e02eb7fe4) e.g.

**user/login.ts**
```

export const loginCommand = createCommand({
  args: ...,
  handler: ...
});
```

**src/index.ts**
```ts
const registry = new CommandRegistry();

...

registry.define([
  { command: "wrangler login", definition: loginCommand }
]);
registry.registerNamespace("login");
```

4. The `DefinitionTreeRoot` that was previously exposed as a singleton, is now encapsulated inside the `CommandRegistry` class, ensuring that there is a fresh root for each command execution. (https://github.com/cloudflare/workers-sdk/pull/7357/commits/11b0b460f5540ff7e94f634bf53633432bcdc77e)

5. `createCommandRegistry` has been converted into a class to better manage the state. One key issue was that the root was being accessed globally in various functions during the definition and registry phase (`resolveDefinitionNode`, `resolveDefinitionChain`, `findParentFor`, `findNodeFor`), so moving the root into the top level closure would have meant passing the root through various other functions along the way, unless of course you move all those functions into the closure, but at that point you're effectively mimicking a class anyway. It turns out much cleaner if they are all methods on a class and have direct access to the root via `this.#DefinitionTreeRoot`.

6. The actual building of the command with Yargs has been abstracted, so that the class is now less coupled to the implementation, which lives now in the [`createRegisterYargsCommand` function](https://github.com/cloudflare/workers-sdk/pull/7357/files#diff-c5b22ef21f285ccf53ba4a990d6fba5e40050aee0dbe5a0b6737852164771d06R18). There is probably more that could be done here, but at least the `CommandRegistry` is now only concerned with managing the command tree.

7. The `runWrangler` tests have been replaced with unit tests for the `CommandRegistry`. It's fair to say that the `runWrangler` tests are already covered by module tests (r2, kv, etc.) and the `CommandRegistry` unit tests are more readable and debuggable. (https://github.com/cloudflare/workers-sdk/pull/7357/commits/7c2f4c0c3d2e3b9d9a50644c69ce4a7a763cbb04)

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/__tests__/vitest.global.ts
```diff
@@ -4,4 +4,5 @@ export function setup(): void {
 	// This needs to be in a globalSetup script - it won't work in a setupFile script.
 	// https://github.com/vitest-dev/vitest/issues/1575#issuecomment-1439286286
 	process.env.LC_ALL = "C";
+	process.env.TZ = "UTC";
```
andyjessop, line 7: Sometimes I see failures locally when dates are involved. This particular time it was the workflows tests. I think it's easier if we just set the time zone here.

packages/wrangler/src/index.ts
```diff
@@ -71,12 +135,23 @@ import registerTriggersSubcommands from "./triggers";
...
+import { workflowsInstanceNamespace, workflowsNamespace } from "./workflows";
+import { workflowsDeleteCommand } from "./workflows/commands/delete";
+import { workflowsDescribeCommand } from "./workflows/commands/describe";
+import { workflowsInstancesDescribeCommand } from "./workflows/commands/instances/describe";
+import { workflowsInstancesListCommand } from "./workflows/commands/instances/list";
+import { workflowsInstancesPauseCommand } from "./workflows/commands/instances/pause";
+import { workflowsInstancesResumeCommand } from "./workflows/commands/instances/resume";
+import { workflowsInstancesTerminateCommand } from "./workflows/commands/instances/terminate";
+import { workflowsListCommand } from "./workflows/commands/list";
+import { workflowsTriggerCommand } from "./workflows/commands/trigger";
```
andyjessop, line 154: There is an argument to be had to hide namespace commands and sub-commands in their own folder, and just define the whole array at once, e.g. so you go from this:

```
registry.define([
		{ command: "wrangler kv:key", definition: kvKeyAlias },
		{ command: "wrangler kv:namespace", definition: kvNamespaceAlias },
	...
	]);
```

To this:
	
```ts
registry.define(kv);
```

But it's worth focussing on the main benefit to having all the registrations in this file as individuals, which is that anyone coming to the codebase has a single place to look to see the entire command API.

It also means you don't have to know what command you're searching for, they are more discoverable if they're laid out here.

andyjessop:
@emily-shen 

> haven't had time to review properly but this doc might need updating too https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/docs/how-to/add-a-binding.md

Maybe not that one (I can't see a reference to `defineCommand` there), but good call - the `contributing.md` needs updating. Updated here: https://github.com/cloudflare/workers-sdk/pull/7357/commits/d55117307dc5133211aa077c7d5b6b5902b0de44

packages/wrangler/src/index.ts
```diff
@@ -351,8 +433,13 @@ export function createCLIParser(argv: string[]) {
...
 
-	// dev
-	register.registerNamespace("dev");
+	registry.define([
+		{
+			command: "wrangler dev",
+			definition: dev,
+		},
+	]);
+	registry.registerNamespace("dev");
```
edmundhung, line 442: Can we merge `.define` and `.registerNamespace` into one api call?

packages/wrangler/src/index.ts
edmundhung: I like the idea of using a flat structure. 👍🏼

packages/wrangler/src/index.ts
andyjessop: I don't believe we can. The limitation comes from the fact that Yargs requires all the registration of sub-commands up front. So you can't register `wrangler kv` and then later register `wrangler kv put`, because put needs to be registered in the same `.command` call to Yargs.

What this means is that first you need to `define` all the commands that you want, and _then_ you need to register them.

I suppose we could try to register all the commands that have been defined in the array immediately, by parsing the array for any new namespaces, but I think that could cause an issue if you wanted to define a command conditionally in a subsequent block and found that the namespace was already registered (and therefore you couldn't add to it).

Worth noting, however, that once everything is converted to the new API, we can drop the individual `registerNamespace` calls, and just keep one `registerAll` call at the end. The only reason we have to have `registerNamespace` at the moment is to ensure that everything is added to the Yargs object in the right order.

andyjessop: Yeah, I agree with all of that. I'm wary about messing too much with the internals here though - ideally this PR just solves one problem, and we make further improvements in future PRs. What do you think?

penalosa (approved):

PR #7360, opened 2024-11-27
fix: allow `wrangler types` when expected entrypoint doesn't exist

### LINKED ISSUE #7346: 🐛 BUG: typegen command errors when the entry point does not exist

Wrangler

3.90.0 [Wrangler], latest version of C3

_No response_

MacOS Version 15.1.1 (24B91)

### Observed behavior

With `@opennextjs/cloudflare`, the `npm run cf-typegen` command throws the following error:

```
✘ [ERROR] The entry-point file at ".worker-next/index.mjs" was not found.
```

This is because the entry point is not generated until you run the build command

### Expected behavior

The correct types should get added to the `env.d.ts` file.

It looks like the entry point is parsed to determine service worker vs module worker format.

We should default ot module when the ep is not found

### Steps to reproduce
Please provide the following:
- Create a new Next.js project using C3 with the `--experimental` flag: `npx create cloudflare@latest next-app -- --framework=next --experimental`
- Add a binding in `wrangler.toml` or add a variable in `.dev.vars`.
- Run the command `npx run cf-typegen`

This will throw an error.

_No response_

_No response_

emily-shen:
Fixes #7346 
`wrangler types` used to fail if an entrypoint was specified in config, but did not exist (e.g. not built yet), because of an assertion inside `getEntry()`

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/__tests__/type-generation.test.ts
+++ b/packages/wrangler/src/__tests__/type-generation.test.ts
@@ -540,6 +540,31 @@ describe("generateTypes()", () => {
 	`);
 	});
 
+	it("should not error if expected entrypoint is not found and assume module worker", async () => {
+		fs.writeFileSync(
+			"./wrangler.toml",
+			TOML.stringify({
+				main: "index.ts",
+				vars: bindingsConfigMock.vars,
+			} as unknown as TOML.JsonMap),
+			"utf-8"
+		);
+		// note index.ts does not exist
+
+		await runWrangler("types");
+		expect(std.out).toMatchInlineSnapshot(`
+		"Generating project types...
+
+		interface Env {
+			SOMETHING: \\"asdasdfasdf\\";
+			ANOTHER: \\"thing\\";
+			\\"some-other-var\\": \\"some-other-value\\";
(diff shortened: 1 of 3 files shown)


penalosa (approved):

packages/wrangler/src/__tests__/type-generation.test.ts
```diff
@@ -540,6 +540,31 @@ describe("generateTypes()", () => {
...
+	it("should not error if expected entrypoint is not found and assume module worker", async () => {
+		fs.writeFileSync(
+			"./wrangler.toml",
+			TOML.stringify({
+				main: "index.ts",
+				vars: bindingsConfigMock.vars,
+			} as unknown as TOML.JsonMap),
+			"utf-8"
+		);
+		// note index.ts does not exist
```
vicb, line 552:
```suggestion
		expect(fs.existsSync("index.ts")).toEqual(false);`
```

vicb (approved):
Thanks!
One minor comment

vicb (approved):

Issue #7366, opened 2024-11-27
🚀 Feature Request: Import the unenv preset for cloudflare 


vicb:
We want to import the [unenv](https://github.com/unjs/unenv) preset from cloudflare in this monorepo as `@cloudflare/unenv-preset`

@pi0 has extracted the preset from unenv to https://github.com/pi0/unenv-preset-cloudflare 🙏

TODO:
- [x] import https://github.com/pi0/unenv-preset-cloudflare
- [x] Sync with latest unenv
- [x] Add peerDependencies on `workerd` and `unenv`
- [x] Enable testing the dev version
- [x] Setup the npm release for `@cloudflare/unenv-preset`
- [ ] Wrangler should update `workerd` when it is pushed to prod
- [x] Wrangler should depend on the npm release of `@cloudflare/unenv-preset`
  - [x]  update [deps.ts](https://github.com/cloudflare/workers-sdk/blob/98d27250d2b37451e808b4aee623c3b4c4de5a20/packages/wrangler/scripts/deps.ts#L21) to include `@cloudflare/unenv-preset`. 
- [ ] workerd update PR should update `@cloudflare/unenv-preset` (automatically via the peerDep)
- [x] rename `$cloudflare.ts` to `index.ts`, see https://github.com/cloudflare/workers-sdk/pull/7370#discussion_r1880277295
- [x] createEnv, see https://github.com/cloudflare/workers-sdk/pull/7370#discussion_r1880273670

/cc @petebacondarwin

pi0:
As tests are non-blockers and dependent to unenv repo structure, I think we can keep it in unenv repo in meanwhile and potentially use `@cloudflare/unenv-preset` as an external preset until move is done.

PR #7370, opened 2024-11-27
import unenv-preset 

vicb:
@pi0 @petebacondarwin This PR imports & sync the cloudflare preset.

https://github.com/cloudflare/workers-sdk/issues/7366 lists (most of) the tasks that will need to be done.

--

- imported from https://github.com/pi0/unenv-preset-cloudflare
  - sync'd with unjs/unenv at d9d4d03
  - added back the .cjs to fix the build

Smoke tested with https://github.com/cloudflare/workers-sdk/tree/preset-test. In that branche, wrangler uses `@cloudflare/unenv-preset` and the hybrid test pass.

Next steps:
- Add a way to test the dev version of the package (the plugin should resolve modules to the workspace version of the file)
- Add tests
- Setup the NPM release
- ... see tracking issue

--

@pi0 I changed the license to MIT + Apache, is that ok with you?

@petebacondarwin do you think it would cause an issue to merge this before all of #7366 is done?

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: will be added in a follow up PR
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: newly created pkg not used yet
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: there is no user facing change


packages/unenv-preset/build.config.ts
pi0, line 1: while personally recommand unbuild/mkdist, also a simple tsc would work if it is widely used in this repo.

packages/unenv-preset/eslint.config.mjs
pi0, line 1: Do you have any other convention perhaps here?

pi0:
@vicb I think you need to update [deps.ts](https://github.com/cloudflare/workers-sdk/blob/98d27250d2b37451e808b4aee623c3b4c4de5a20/packages/wrangler/scripts/deps.ts#L21) to include this new package.

vicb:
> @vicb I think you need to update [deps.ts](https://github.com/cloudflare/workers-sdk/blob/98d27250d2b37451e808b4aee623c3b4c4de5a20/packages/wrangler/scripts/deps.ts#L21) to include this new package.

Good point, thanks @pi0. I'll add that to the tracking issue for when we switch wrangler to using this package.

.gitignore
```diff
@@ -198,6 +198,8 @@ packages/chrome-devtools-patches/depot
 packages/chrome-devtools-patches/devtools-frontend
 packages/miniflare/dist-types/
 
+packages/unenv-preset/runtime
```
anonrig, line 201: Can we add a codeowners document as well?

packages/unenv-preset/package.json
```diff
@@ -0,0 +1,62 @@
...
+	"scripts": {
+		"build": "unbuild",
+		"check:lint": "eslint",
+		"check:type": "tsc --noEmit"
+	},
+	"dependencies": {
+		"unenv": "npm:unenv-nightly@2.0.0-20241127-105629-7a9f230"
+	},
+	"devDependencies": {
+		"@types/node": "^22.9.3",
```
anonrig, line 48: Should this come from workspace as well?

.gitignore
vicb: @anonrig Add a task in the tracking issue to setup CODEOWNERS

packages/unenv-preset/package.json
vicb: Nope.
This is the version of the types matching the unenv impl.

packages/unenv-preset/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { cloudflare } from "./preset";
+export { env, nodeless } from "unenv";
```
pi0, line 2: Is is likely that this will change before the final unenv v2 release to simpler `createEnv({ presets: [], ...options })` and we drop/deprecate older exports.

Totally up to you (as long as we have unenv as relaxed `^` peer dependency it is fine) but I would very much root for not doing reexport and compose where preset is "consumed" (it can be `wrangler`, esm.sh, nitro or other places using unenv and will be always forward compatible)

packages/unenv-preset/src/runtime/node/async_hooks/$cloudflare.ts
pi0, line 1: You can convert `$cloudflare` files to `index` btw now as it is an actual external preset. IMO it can make things more readable.

packages/unenv-preset/src/index.ts
vicb: I added your 2 comments to the tracking issue #7366

The idea is to get this in to unblock the next steps and then we can tweak.

packages/unenv-preset/src/runtime/node/async_hooks/$cloudflare.ts
vicb: I added your 2 comments to the tracking issue https://github.com/cloudflare/workers-sdk/issues/7366

packages/unenv-preset/src/index.ts
pi0: The issue is that, if we start by having peer dependency of unenv, even if unused, it keeps nitro blocked as it forces a pinned version of unenv. 

While it is actually possible that we keep unenv as a `wrangler` dependency so that all consumers are free to choose their versioning policy for `unenv` (the composer) + their preset compatibility.

pi0: Having peer dep as a range in the future is workable. But we hack a dead-lock situation here.

- unenv v2 won't be released unless we can test it fully against nitro
- unenv-preset won't relax peer dep of unenv until unenv v2 is released
- nitro (nightly) cannot test unenv-preset without setting the pinned dep

I might be missing some of the context in your integration plan and steps but it looks to me, `0.0.1` can be safely released without reexports and peer dependency and tested/integrated against wrangler by composing in wrangler layer and this way starting `0.0.1` we can also start testing it against nitro to finalize both sides.

.gitignore
anonrig: Which tracking issue?

packages/unenv-preset/build.config.ts
```diff
@@ -0,0 +1,19 @@
+import { defineBuildConfig } from "unbuild";
+
+export default defineBuildConfig({
+	declaration: true,
+	rollup: {
+		emitCJS: true,
+	},
+	entries: [
+		"src/index",
+		{ input: "src/runtime/", outDir: "runtime", format: "esm" },
```
petebacondarwin, line 10: Can we output to `dist/runtime` and then fix up the package.json exports, which would also allow us to remove the `.gitignore` change.

packages/unenv-preset/src/index.ts
petebacondarwin, line 2: Why do we re-export this?
Do presets normally also export `env()` etc?

petebacondarwin (commented):
Will need to update `findPackageNames()` in `validate-changesets.test.ts`.

packages/unenv-preset/build.config.ts
vicb: Yes, I can change to output to 2 different dist subfolders

packages/unenv-preset/turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+	"$schema": "http://turbo.build/schema.json",
+	"extends": ["//"],
+	"tasks": {
+		"build": {
+			"outputs": ["dist/**", "runtime/**"]
```
petebacondarwin, line 6: Now that runtime is output to dist is it enough to just do this?

```suggestion
			"outputs": ["dist/**"]
```

petebacondarwin (approved):

petebacondarwin:
Still need to fix this test:

> Will need to update findPackageNames() in validate-changesets.test.ts.

packages/unenv-preset/turbo.json
vicb: good catch!

petebacondarwin:
And appears to be a problem with the lockfile?

packages/unenv-preset/package.json
```diff
@@ -0,0 +1,61 @@
...
+		"wrangler": "3.95.0"
+	},
+	"peerDependencies": {
+		"workerd": "^1.20241205.0"
+	},
+	"peerDependenciesMeta": {
+		"workerd": {
+			"optional": true
+		}
+	}
```
petebacondarwin, line 60: If you add this here:

```suggestion
	},
	"workers-sdk": {
		"prerelease": true
	}
```

Then we should auto-generate pre-release builds per PR.

petebacondarwin:
I think to do the first deployment we need to mark it as public access:

```sh
npm publish --access public
```

packages/unenv-preset/turbo.json
petebacondarwin: ...

petebacondarwin (approved):
Just this one small nit.

packages/unenv-preset/turbo.json
vicb: really fixed this time :)
Thanks!

PR #7378, opened 2024-11-27
Add build and push subcommands to cloudchamber

IRCody:
_Describe your change..._
Add build and push subcommands to cloudchamber that build a docker image and push to the cloudchamber managed registry using a local docker daemon.
---

- Tests
  - [] TODO (before merge)
  - [x] Tests included
  - [] I don't know
  - [ ] Required
  - [x] Not required because: No e2e test for Cloudchamber yet
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: product is private only for certain users.

### DIFF
--- a/packages/wrangler/src/cloudchamber/build.ts
+++ b/packages/wrangler/src/cloudchamber/build.ts
@@ -1,0 +1,209 @@
+import { spawn } from "child_process";
+import { logRaw } from "@cloudflare/cli";
+import { ImageRegistriesService } from "./client";
+import type { Config } from "../config";
+import type {
+	CommonYargsArgvJSON,
+	StrictYargsOptionsToInterfaceJSON,
+} from "../yargs-types";
+import type { ImageRegistryPermissions } from "./client";
+
+// default cloudflare managed registry
+const domain = "registry.cloudchamber.cfdata.org";
+
+export async function dockerLoginManagedRegistry(options: {
+	pathToDocker?: string;
+}) {
+	const dockerPath = options.pathToDocker ?? "docker";
+	const expirationMinutes = 15;
+
+	await ImageRegistriesService.generateImageRegistryCredentials(domain, {
+		expiration_minutes: expirationMinutes,
+		permissions: ["push"] as ImageRegistryPermissions[],
+	}).then(async (credentials) => {
+		const child = spawn(
+			dockerPath,
+			["login", "--password-stdin", "--username", "v1", domain],
+			{ stdio: ["pipe", "inherit", "inherit"] }
+		).on("error", (err) => {
+			throw err;
+		});
+		child.stdin.write(credentials.password);
+		child.stdin.end();
+		await new Promise((resolve) => {
+			child.on("close", resolve);
+		});
+	});
+}
+
+export async function constructBuildCommand(options: {
+	imageTag?: string;
+	pathToDocker?: string;
+	pathToDockerfile?: string;
+	platform?: string;
+}) {
+	// require a tag if we provide dockerfile
+	if (
+		typeof options.pathToDockerfile !== "undefined" &&
+		options.pathToDockerfile !== "" &&
+		(typeof options.imageTag === "undefined" || options.imageTag === "")
+	) {
+		throw new Error("must provide an image tag if providing a docker file");
+	}
+	const dockerFilePath = options.pathToDockerfile;
+	const dockerPath = options.pathToDocker ?? "docker";
+	const imageTag = domain + "/" + options.imageTag;
+	const platform = options.platform ? options.platform : "linux/amd64";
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -0,0 +1,225 @@
...
+	await new Promise((resolve) => {
+		child.on("close", resolve);
+	});
+}
+
+export function buildYargs(yargs: CommonYargsArgvJSON) {
+	return yargs
+		.positional("PATH", {
+			type: "string",
+			describe: "path for the directory containing the dockerfile",
```
gpanders, line 129:
```suggestion
			describe: "path to the directory containing the Dockerfile to build",
```

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -0,0 +1,225 @@
...
+			demandOption: true,
+		})
+		.option("tag", {
+			alias: "t",
+			type: "string",
+			demandOption: true,
+			describe: "Tag to use for the built image",
```
gpanders, line 136: We should mention that this should also include the name of the image, not just the tag (i.e. it should be `-t foo-bar:1.0`, not just `-t 1.0`).

`docker build` describes this as "Name and optionally a tag (format: "name:tag")". I think we should just use the exact same phrasing since it just gets passed to `docker build -t` anyway.

```suggestion
			describe: "Name and optionally a tag (format: \"name:tag\")",
```

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -0,0 +1,225 @@
...
+		})
+		.option("path-to-docker", {
+			type: "string",
+			default: "docker",
+			describe: "path to docker binary",
```
gpanders, line 141: The option is called "path-to-docker" but accepts a command (i.e. `"docker"`) so maybe `docker-cmd` is a better name? Path implies a full path.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -0,0 +1,225 @@
...
+			demandOption: false,
+		})
+		.option("build-command", {
+			type: "string",
+			describe: "custom build command to use",
```
gpanders, line 146: I'm not really sure what this option does from the description or why/how it's different from the `path-to-docker` option. Do we need both? Can we eliminate one or the other?

I'm also curious to read more about the use case for this option. It strikes me as a potential footgun since specifying a custom build command seems to ignore all of the other options provided to this command, which seems like it could be confusing.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -0,0 +1,225 @@
...
+		})
+		.option("push", {
+			alias: "p",
+			type: "boolean",
+			describe: "push the built image to a registry",
+			default: false,
+		})
+		.option("platform", {
+			type: "string",
+			default: "linux/amd64",
```
gpanders, line 157: Can we make the default whatever the user is using (e.g. `linux/arm64` on an M1 Mac and `linux/amd64` on x86)?

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -0,0 +1,225 @@
...
+		})
+		.option("build-command", {
+			type: "string",
+			describe: "custom build command to use",
+			demandOption: false,
+		})
+		.option("push", {
+			alias: "p",
+			type: "boolean",
+			describe: "push the built image to a registry",
```
gpanders, line 152: Nit: since this only supports pushing to _one_ registry, we should make that clear

```suggestion
			describe: "push the built image to Cloudflare's managed registry",
```

I see there's a TODO comment to potentially change this though. If that's the case then this suggestion doesn't apply.

packages/wrangler/src/cloudchamber/build.ts
mikenomitch: I think it probably makes sense to rip this out of "wrangler build", but keep a custom build-command in "wrangler push" and eventually "wrangler cloudchamber apply/deploy".

I think it's fair to say that if you're overriding this here, then you can just run the command yourself.

packages/wrangler/src/cloudchamber/build.ts
mikenomitch: Yeah +1 to using amd64 as the default.

Shoudl we say in the description of build that it defaults to "architecture supported by Workers (linux/amd64)"?

ichernetsky-cf (approved):

penalosa (approved):
This is cloudchamber-only, so stamping with a quick skim

PR #7381, opened 2024-11-28
Turn on `--x-registry` for Pages by default

### LINKED ISSUE #7322: 🐛 BUG: Pages never connects to external workers in local development

C3 (npm create cloudflare), Pages, Wrangler

3.88.0 [Wrangler]

22.9.0

Windows 11, Linux Ubuntu WSL

### Observed Behavior

After configuring and starting a connection from an ES module worker to a page, the page, which was run with `npx wrangler pages dev`, never connects to the worker, which is also running in another terminal with `npx wrangler dev`.

### Expected Behavior

It is expected that after running the dev command for the page project and then running the dev command for the worker project, the page project should be able to connect to the worker running locally, as described in the [documentation](https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/#local-development).

### Steps to Reproduce

This scenario can be easily tested by creating each of the projects with the command `npm create cloudflare@latest`.

#### Setup Worker Project

- Run `npm create cloudflare@latest -- worker-a`
- Select the first option _Hello World example_
- Choose the default options for the remaining choices
- Run `npm run dev`

#### Setup Page Project

- Run `npm create cloudflare@latest -- pages-nuxt`
- Select the second option _Framework Starter_
- Choose the NUXT framework (I also tested using another framework like NEXT, but encountered the same issue)
- Choose the default options for the remaining choices
- Run `npx wrangler pages dev --service WORKER_A=worker-a`

After running both projects in development mode, regardless of the order in which they were started, the output in the pages project will be:

```bash
- Services:
  - WORKER_A: worker-a [not connected]
```

Indicating that the page project did not connect to the worker-a.

I reviewed the documentation, repeated this process on Linux, tested switching from NUXT to NEXT, and had no success. Therefore, I am reporting this issue as a bug.

#### Note

This issue might be related to another [request](https://github.com/cloudflare/workers-sdk/issues/5918)

_No response_

_No response_

penalosa:
Fixes #7322

When flipping the `--x-registry` default, we forgot to turn it on for Pages, resulting in Pages and Workers not being able to connect over service bindings unless `--x-registry=false` was specified on the Worker invocation. Our tests didn't catch this because they always specified the `--x-registry` flag (for both the true and false case), and so the defaults weren't being tested. This PR:
- Switches the default to `true` for Pages
- Changes our E2E tests to correctly test the default (i.e. only the false branch has a flag provided)
- Adds additional Pages E2E tests

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/seven-eels-reflect.md
+++ b/.changeset/seven-eels-reflect.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Turn on `--x-registry` for Pages by default
--- a/packages/wrangler/e2e/dev-registry.test.ts
+++ b/packages/wrangler/e2e/dev-registry.test.ts
@@ -1,4 +1,5 @@
 import { execSync } from "child_process";
+import getPort from "get-port";
 import dedent from "ts-dedent";
 import { fetch } from "undici";
 import { beforeEach, describe, expect, it, vi } from "vitest";
@@ -140,7 +141,7 @@ describe("unstable_dev()", () => {
 });
 
 describe.each([
-	{ cmd: "wrangler dev --x-registry" },
+	{ cmd: "wrangler dev" },
 	{ cmd: "wrangler dev --no-x-registry" },
 ])("dev registry $cmd", ({ cmd }) => {
 	let workerName: string;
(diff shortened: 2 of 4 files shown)


emily-shen (approved):
yay thanks for fixing this 🎉

PR #7385, opened 2024-11-28
feat(wrangler): add inherit bindings support

edmundhung:
Fixes [DEVX-1419](https://jira.cfdata.org/browse/DEVX-1419).

This PR adds inherit bindings support using `wrangler deploy --x-provision`. The current implementation does not take into account the settings of the deployed worker yet. It assumes  resources with `undefined` id to be inherit bindings for the ease of testing.  

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: unreleased feature


packages/wrangler/src/deployment-bundle/provision.ts
```diff
@@ -0,0 +1,64 @@
...
+	config: Config | undefined,
+	options?: {
+		pages?: boolean;
+	}
+): CfWorkerInit["bindings"] {
+	return {
+		kv_namespaces: config?.kv_namespaces?.map((kv) => ({
+			...kv,
+			id: kv.id ?? INHERIT_SYMBOL,
+		})),
```
penalosa, line 19:
```suggestion
		kv_namespaces: config?.kv_namespaces?.map((kv) => ({
			...kv,
			id: kv.id ?? INHERIT_SYMBOL,
		})) ?? [],
```

packages/wrangler/src/deployment-bundle/provision.ts
```diff
@@ -0,0 +1,64 @@
+import type { Config } from "../config";
+import type { CfWorkerInit } from "./worker";
+
+/**
+ * A symbol to inherit a binding from the deployed worker.
+ */
+export const INHERIT_SYMBOL = Symbol.for("inherit_binding");
+
+export function getBindings(
+	config: Config | undefined,
```
penalosa, line 10: Why is config optional here?

penalosa (approved):

packages/wrangler/src/deployment-bundle/provision.ts
edmundhung: It could be undefined for pages project.

packages/wrangler/src/deployment-bundle/provision.ts
edmundhung: The type of `kv_namespaces` is `CfKvNamespace[] | undefined;` so defaulting it to empty array won't help. Should I try making the type `CfKvNamespace[]` instead?

PR #7388, opened 2024-11-29
Remove node_compat flag from C3 templates that use Vitest Pool Workers

andyjessop:
Fixes #000.

Removes the need to add `nodejs_compat` flag in C3 templates just to use Vitest Pool Workers. Vitest Pool Workers will now inject the flag if it doesn't exist.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Covered by existing
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18495

### DIFF
--- a/.changeset/calm-lizards-tell.md
+++ b/.changeset/calm-lizards-tell.md
@@ -1,6 +1,0 @@
----
-"@cloudflare/vitest-pool-workers": minor
-"create-cloudflare": minor
----
-
-chore: add nodejs_compat by default in Vitest Pool Workers, and remove nodejs_compat flag from basic C3 templates
--- a/.changeset/calm-lizards-tell.md
+++ b/.changeset/calm-lizards-tell.md
@@ -1,0 +1,6 @@
+---
+"@cloudflare/vitest-pool-workers": minor
+"create-cloudflare": minor
+---
+
+chore: add nodejs_compat by default in Vitest Pool Workers, and remove nodejs_compat flag from basic C3 templates
--- a/packages/vitest-pool-workers/src/pool/index.ts
+++ b/packages/vitest-pool-workers/src/pool/index.ts
@@ -9,6 +9,7 @@ import { createBirpc } from "birpc";
 import * as devalue from "devalue";
 import {
 	compileModuleRules,
+	getNodeCompat,
 	kCurrentWorker,
 	kUnsafeEphemeralUniqueKey,
 	Log,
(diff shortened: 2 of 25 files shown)


Cherry (commented):
This is incredibly interesting!

Won't this potentially cause issues where a project could work during tests (if it does `nodejs_compat` stuff) but then fail during deploy or in prod, creating a different and now hard to debug environment?

Perhaps simply doing `nodejs_compat` + `no_nodejs_compat_v2` would make more sense, to reduce the amount of cruft added to workers by default (#7108), but avoid the environment differences?

DaniFoldi:
I think unfortunately enabling nodejs_compat_v2 by default would break opentelemetry libs' platform detection - and possibly many others. Checking for globalThis.Buffer is still pretty common in libraries :/

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -369,6 +365,16 @@ function buildProjectWorkerOptions(
...
 	}
 
+	const { mode } = getNodeCompat(
+		runnerWorker.compatibilityDate,
+		runnerWorker.compatibilityFlags
+	);
+
+	if (mode !== "v1" && mode !== "v2") {
+		runnerWorker.compatibilityFlags.push("nodejs_compat");
+		runnerWorker.compatibilityFlags.push("no_nodejs_compat_v2");
```
petebacondarwin, line 375: Pedantic NIT 😄 
```suggestion
		runnerWorker.compatibilityFlags.push("nodejs_compat", "no_nodejs_compat_v2");
```

petebacondarwin (approved):

packages/vitest-pool-workers/src/pool/index.ts
andyjessop: Thanks, I'm not really with it this morning...

.changeset/calm-lizards-tell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cloudflare/vitest-pool-workers": minor
+"create-cloudflare": minor
+---
+
+chore: add nodejs_compat by default in Vitest Pool Workers, and remove nodejs_compat flag from basic C3 templates
```
edmundhung, line 6: Nit: Would it better splitting this into 2 changesets?

andyjessop: Added here: https://github.com/cloudflare/workers-sdk/pull/7388/commits/27f98409e3c45d5802aabfdfc8710b0efc05698d

PR #7389, opened 2024-11-29
test: fix timezone for tests

andyjessop:
Fixes #000.

When running tests locally, the machine's timezone can cause issues with tests that rely on `Date`. This PR pins the timezone to `UTC`, so that it's the same for everyone, whether running locally or in CI.

Before 😞 

Screenshot 2024-11-29 at 11 30 10

After 😄 

Screenshot 2024-11-29 at 11 29 19

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Covered already
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Internal only

### DIFF
--- a/fixtures/vitest-pool-workers-examples/vitest.global.ts
+++ b/fixtures/vitest-pool-workers-examples/vitest.global.ts
@@ -1,0 +1,3 @@
+export function setup(): void {
+	process.env.TZ = "UTC";
+}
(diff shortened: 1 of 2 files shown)


fixtures/vitest-pool-workers-examples/vitest.global.ts
```diff
@@ -0,0 +1,3 @@
+export function setup(): void {
```
CarmenPopoviciu, line 1: for future readers 😋 , see https://github.com/vitest-dev/vitest/issues/1575

CarmenPopoviciu (approved):

petebacondarwin:
Late to the game but I think is also good: https://github.com/cloudflare/workers-sdk/pull/7317

PR #7394, opened 2024-11-30
Update Astro experimental template

alexanderniebuhr:
In theory this should be the only change needed after https://github.com/withastro/adapters/pull/406 lands, not sure if c3 needs any other changes though 🤔 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


alexanderniebuhr:
Since there is an [option to ignore files](https://developers.cloudflare.com/workers/static-assets/binding/#ignoring-assets) now, I think we will stick with that for now. Until Worker Assets is out of Beta, and we know if Pages will stay.

PR #7400, opened 2024-12-02
Automatically backport Wrangler patches to v3

penalosa:
See https://github.com/cloudflare/workers-sdk/pull/7402 for an example.

This PR adds a workflow that will, for every PR with a Wrangler patch change, open a corresponding PR to the `v3-maintenance` branch (not created yet, but will be after v4 is released). The auto-generated PRs will likely not be perfect, but they should be a good starting point to backporting the fix. Some fixes will not be relevant, and so could just be closed (i.e. a fix for a feature introduced in v4 wouldn't make sense to backport)

### DIFF
--- a/.github/workflows/open-v3-maintenance-prs.yml
+++ b/.github/workflows/open-v3-maintenance-prs.yml
@@ -1,0 +1,33 @@
+name: v3 Maintenance
+
+on: pull_request
+
+jobs:
+  open-pr:
+    if: ${{ github.repository_owner == 'cloudflare' }}
+    name: Open backport PR for patches
+    runs-on: ubuntu-latest
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
+    timeout-minutes: 30
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install Dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/open-v3-maintenance-prs.yml
```diff
@@ -0,0 +1,33 @@
+name: v3 Maintenance
+
+on: pull_request
+
+jobs:
+  open-pr:
+    if: ${{ github.repository_owner == 'cloudflare' }}
+    name: Open backport PR for patches
+    runs-on: ubuntu-latest
```
emily-shen, line 9: i don't particularly care, but want to move this to macOs as that might have a bit more CI capacity?

tools/deployments/open-v3-pr.ts
```diff
@@ -0,0 +1,43 @@
+import { execSync } from "node:child_process";
+import { readFileSync } from "node:fs";
+import parseChangeset from "@changesets/parse";
+
+/* eslint-disable turbo/no-undeclared-env-vars */
+if (require.main === module) {
+	const parsedLabels = JSON.parse(process.env.LABELS as string) as string[];
+	if (
+		isWranglerPatch(process.env.FILES as string) &&
+		!parsedLabels.includes("skip-backport")
```
emily-shen, line 10: `skip-v3-PR` might be a clearer  label name maybe?

emily-shen (approved):

PR #7409, opened 2024-12-02
fix: update Angular experimental Workers + Assets template

petebacondarwin:
Fixes #0000

fix: update Angular experimental Workers + Assets template

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: C3 e2e tests run automatically
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: patch fix


packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -74,6 +189,80 @@ function getFrameworkTests(opts: {
 				},
 				unsupportedOSs: ["win32"],
 			},
+			nuxt: {
+				quarantine: true,
```
emily-shen, line 193: just double checking, this is still supposed to quarantined?

packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -74,6 +189,80 @@ function getFrameworkTests(opts: {
...
+						input: [keys.enter],
+					},
+					{
+						matcher:
+							/Which package manager do you want to install dependencies with/,
+						input: [keys.enter],
+					},
+				],
+				testCommitMessage: true,
+				unsupportedOSs: ["win32"],
```
emily-shen, line 255: just noticed no windows tests actually end up running 🤣 only gatsby supports windows, but we run pnpm for the windows test and gatsby doesn't support that. not sure if worth changing that the windows PM to npm or something.

emily-shen (approved):

packages/create-cloudflare/e2e-tests/frameworks.test.ts
petebacondarwin: It fails in this PR - but then passes in after the Nuxt PR 😄

packages/create-cloudflare/e2e-tests/frameworks.test.ts
petebacondarwin: We should tidy that up as a follow up

petebacondarwin: https://github.com/cloudflare/workers-sdk/issues/7412

PR #7410, opened 2024-12-02
Remove await from user Worker fetch in router-worker

WalshyDev:
Fixes N/A

Stops us waiting on asset Worker to return and escalating the errors upwards. We don't need to do this and it currently skews metrics as well.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no material change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no public facing change

### DIFF
--- a/.changeset/gorgeous-paws-compare.md
+++ b/.changeset/gorgeous-paws-compare.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-shared": patch
+---
+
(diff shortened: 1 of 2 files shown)


WillTaylorDev (approved):

WillTaylorDev:
(Pending changeset)

PR #7415, opened 2024-12-02
refactor: move projectRoot computation to config validation

petebacondarwin:
A refactor to move the `projectRoot` field to config (from entry) so that it can be used more flexibly.

### DIFF
--- a/packages/wrangler/src/deployment-bundle/entry.ts
+++ b/packages/wrangler/src/deployment-bundle/entry.ts
@@ -22,8 +22,6 @@ import type { CfScriptFormat } from "./worker";
 export type Entry = {
 	/** A worker's entrypoint */
 	file: string;
-	/** A worker's directory. Usually where the Wrangler configuration file is located */
-	projectRoot: string;
 	/** Is this a module worker or a service worker? */
 	format: CfScriptFormat;
 	/** The directory that contains all of a `--no-bundle` worker's modules. Usually `${directory}/src`. Defaults to path.dirname(file) */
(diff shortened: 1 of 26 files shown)


packages/wrangler/src/deployment-bundle/entry.ts
```diff
@@ -60,9 +58,9 @@ export async function getEntry(
 	if (args.script) {
 		paths = resolveEntryWithScript(args.script);
 	} else if (config.main !== undefined) {
-		paths = resolveEntryWithMain(config.main, config.configPath);
+		paths = resolveEntryWithMain(config.main, config.projectRoot);
 	} else if (entryPoint) {
-		paths = resolveEntryWithEntryPoint(entryPoint, config.configPath);
+		paths = resolveEntryWithEntryPoint(entryPoint, config.projectRoot);
```
penalosa, line 63: I'm not sure why this isn't a type error, but I think the return value here is used further down, and should probably be changed to `config.projectRoot` as well:

```ts
89: const projectRoot = paths.projectRoot ?? process.cwd();
```

packages/wrangler/src/versions/upload.ts
```diff
@@ -69,7 +69,6 @@ type Props = {
 	dryRun: boolean | undefined;
 	noBundle: boolean | undefined;
 	keepVars: boolean | undefined;
```
penalosa, line 71: Unrelated to this PR, but it looks like these props could do with some cleanup... They seem to take both config.rules and config as separate values?

penalosa (approved):

packages/wrangler/src/deployment-bundle/entry.ts
petebacondarwin: Oops I missed that one.
It is not a type error it is just redundant further down.

packages/wrangler/src/versions/upload.ts
petebacondarwin: So in the call site for `versionsUpload()` there is `rules: getRules(config)`, which does som additional validation of deprecated rules options. That function is used in various places.

I think that whole function can be removed in Wrangler V4, since the deprecated fields might be removed?

PR #7424, opened 2024-12-03
Add support for Images binding

ns476:
Fixes IMAGES-1400, IMAGES-1401, IMAGES-1402

Adds support for the Images binding, with two kinds of preview:
- Remote: hits a remote endpoint that exposes the same HTTP interface used in production
- Local: uses a local fake that only supports rotating, resizing and transcoding, intended mainly for use in tests.
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: For the time being this is in private beta - docs are being worked on separately.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,1 @@
 git-checks=false
-public-hoist-pattern[]=@img*
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 git-checks=false
+public-hoist-pattern[]=@img*
--- a/packages/wrangler/e2e/dev-with-resources.test.ts
+++ b/packages/wrangler/e2e/dev-with-resources.test.ts
@@ -691,11 +691,11 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 		await expect(res.text()).resolves.toBe("env.WORKFLOW is available");
 	});
 
-	it("exposes Images bindings", async () => {
-		describe.sequential.each([
-			{ imagesMode: "remote", extraFlags: "" },
-			{ imagesMode: "local", extraFlags: "--experimental-images-local-mode" },
-		] as const)("Images Binding Mode: $imagesMode", async ({ extraFlags }) => {
+	describe.sequential.each([
+		{ imagesMode: "remote", extraFlags: "" },
+		{ imagesMode: "local", extraFlags: "--experimental-images-local-mode" },
+	] as const)("Images Binding Mode: $imagesMode", async ({ extraFlags }) => {
+		it("exposes Images bindings", async () => {
 			await helper.seed({
 				"wrangler.toml": dedent`
 					name = "my-images-demo"
--- a/packages/wrangler/e2e/dev-with-resources.test.ts
+++ b/packages/wrangler/e2e/dev-with-resources.test.ts
@@ -691,6 +691,40 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 		await expect(res.text()).resolves.toBe("env.WORKFLOW is available");
 	});
 
+	it("exposes Images bindings", async () => {
+		describe.sequential.each([
+			{ imagesMode: "remote", extraFlags: "" },
+			{ imagesMode: "local", extraFlags: "--experimental-images-local-mode" },
+		] as const)("Images Binding Mode: $imagesMode", async ({ extraFlags }) => {
+			await helper.seed({
+				"wrangler.toml": dedent`
+					name = "my-images-demo"
+					main = "src/index.ts"
+					compatibility_date = "2024-12-27"
+
+					[images]
+					binding = "IMAGES"
+				`,
+				"src/index.ts": dedent`
+					export default {
+						async fetch(request, env, ctx) {
+							if (env.IMAGES === undefined) {
+								return new Response("env.IMAGES is undefined");
+							}
+
+							return new Response("env.IMAGES is available");
+						}
+					}
+				`,
+			});
+			const worker = helper.runLongLived(`wrangler dev ${flags} ${extraFlags}`);
+			const { url } = await worker.waitForReady();
+			const res = await fetch(url);
+
+			await expect(res.text()).resolves.toBe("env.IMAGES is available");
+		});
+	});
+
 	// TODO(soon): implement E2E tests for other bindings
 	it.todo("exposes hyperdrive bindings");
 	it.skipIf(isLocal).todo("exposes send email bindings");
--- a/packages/wrangler/src/__tests__/config/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/config/configuration.test.ts
@@ -2302,6 +2302,69 @@ describe("normalizeAndValidateConfig()", () => {
 			});
 		});
 
+		// Images
+		describe("[images]", () => {
+			it("should error if images is an array", () => {
+				const { diagnostics } = normalizeAndValidateConfig(
+					{ images: [] } as unknown as RawConfig,
+					undefined,
+					undefined,
+					{ env: undefined }
+				);
+
+				expect(diagnostics.hasWarnings()).toBe(false);
+				expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Processing wrangler configuration:
+			  - The field \\"images\\" should be an object but got []."
+		`);
+			});
+
+			it("should error if images is a string", () => {
+				const { diagnostics } = normalizeAndValidateConfig(
+					{ images: "BAD" } as unknown as RawConfig,
+					undefined,
+					undefined,
+					{ env: undefined }
+				);
+
+				expect(diagnostics.hasWarnings()).toBe(false);
+				expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Processing wrangler configuration:
+			  - The field \\"images\\" should be an object but got \\"BAD\\"."
+		`);
+			});
+
+			it("should error if images is a number", () => {
+				const { diagnostics } = normalizeAndValidateConfig(
+					{ images: 999 } as unknown as RawConfig,
+					undefined,
+					undefined,
+					{ env: undefined }
+				);
+
+				expect(diagnostics.hasWarnings()).toBe(false);
+				expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Processing wrangler configuration:
+			  - The field \\"images\\" should be an object but got 999."
+		`);
+			});
+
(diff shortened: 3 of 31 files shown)


penalosa:
@ns476 It doesn't look like this is ready for review yet, so I've converted this to draft for now

packages/wrangler/src/config/environment.ts
```diff
@@ -743,6 +743,21 @@ export interface EnvironmentNonInheritable {
...
+	/**
+	 * Binding to Cloudflare Images
+	 *
+	 * NOTE: This field is not automatically inherited from the top level environment,
+	 * and so must be specified in every named environment.
+	 *
+	 * @default `{}`
+	 * @nonInheritable
+	 */
+	images:
```
CarmenPopoviciu, line 755: I know we are not doing docs yet, but when we do we should not forget to update https://developers.cloudflare.com/workers/wrangler/configuration/#non-inheritable-keys as well. A lot of folks tend to forget

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+});
+
+const Transforms = z.array(Transform);
+
+export async function imagesLocalFetcher(request: Request): Promise<Response> {
+	const data = await request.formData();
+
+	const body = data.get("image");
+	if (!body || !(body instanceof File)) {
+		return errorResponse(400, 9523, "ERROR: Expected image in request");
```
CarmenPopoviciu, line 21: can we be a bit more helpful here for our users in our err message. Maybe smth like

```suggestion
		return errorResponse(400, 9523, "ERROR: Expected image in request body but got ....");
```

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+
+	const url = new URL(request.url);
+
+	if (url.pathname == "/info") {
+		return runInfo(transformer);
+	} else {
+		const badTransformsResponse = errorResponse(
+			400,
+			9523,
+			"ERROR: Expected JSON array of valid transforms in transforms field"
```
CarmenPopoviciu, line 34: how can we make this err message more useful for users so they know how to action? Should we maybe phrase this as `Expected ... but got .... Please make sure ...`. Also what is a valid transform? Is that documented somewhere? If so, we could add the docs link or a tl;dr so users don't have to switch context

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+				return badTransformsResponse;
+			}
+
+			const outputFormat = data.get("output_format");
+
+			if (outputFormat != null && typeof outputFormat !== "string") {
+				return errorResponse(
+					400,
+					9523,
+					"ERROR: Expected output format to be a string if provided"
```
CarmenPopoviciu, line 55:
```suggestion
					"ERROR: Expected output format to be a string if provided, but got ..."
```

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+			mime = "image/webp";
+			break;
+		case "gif":
+			mime = "image/gif";
+			break;
+		case "avif":
+			mime = "image/avif";
+			break;
+		default:
+			return errorResponse(415, 9520, "ERROR: Unsupported image type");
```
CarmenPopoviciu, line 90: might be worth mentioning which image types are supported, and what type was provided? Or add a link to docs where supported image types are documented....that way we wouldn't have to update here every time a new image type is added to that support list. Up to you

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+	if (mime == "image/svg+xml") {
+		resp = {
+			format: mime,
+		};
+	} else {
+		if (!metadata.size || !metadata.width || !metadata.height) {
+			return errorResponse(
+				500,
+				9523,
+				"ERROR: Expected size, width and height for bitmap input"
```
CarmenPopoviciu, line 103: this is a bit vague...for me personally it is not clear what the actual problem is, just by looking at this error

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+async function runTransform(
+	transformer: Sharp,
+	transforms: z.infer<typeof Transforms>,
+	outputFormat: string | null
+): Promise<Response> {
+	for (const transform of transforms) {
+		if (transform.imageIndex !== undefined && transform.imageIndex !== 0) {
+			// We don't support draws, and this transform doesn't apply to the root
```
CarmenPopoviciu, line 125: I have no context here, but I was wondering if this is smth we should surface to users?

.npmrc
```diff
@@ -1 +1,2 @@
 git-checks=false
+public-hoist-pattern[]=@img*
```
penalosa, line 2: Why is this needed?

packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -691,6 +691,40 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 		await expect(res.text()).resolves.toBe("env.WORKFLOW is available");
 	});
 
+	it("exposes Images bindings", async () => {
+		describe.sequential.each([
```
penalosa, line 695: These should be swapped (the describe on the outside, and it on the inside)

packages/wrangler/src/dev.ts
```diff
@@ -324,6 +324,12 @@ export const dev = createCommand({
 				"Bind to production Vectorize indexes in local development mode",
 			default: false,
 		},
+		"experimental-images-local-mode": {
+			type: "boolean",
+			describe:
+				"Use a local lower-fidelity implementation of the Images binding",
+			default: false,
+		},
```
penalosa, line 332: I know this is a product call, but I would really suggest flipping the default here. Local-mode development in Wrangler is the norm, and I think having remote be an opt-in would fit in with that better

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -952,6 +983,15 @@ export async function buildMiniflareOptions(
...
 	}
 
+	if (config.bindings.images && config.imagesLocalMode) {
+		if (!didWarnImagesLocalModeUsage) {
+			didWarnImagesLocalModeUsage = true;
+			logger.info(
+				"You are using Images local mode. This only supports resizing, rotating and transcoding."
+			);
+		}
+	}
```
penalosa, line 993: This will also need an equivalent warning for remote dev warning about usage charges (assuming there are usage charges)

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+import type { Sharp } from "sharp";
+
+const Transform = z.object({
+	imageIndex: z.number().optional(),
+	rotate: z.number().optional(),
+	width: z.number().optional(),
+	height: z.number().optional(),
+});
+
+const Transforms = z.array(Transform);
```
penalosa, line 14: Adding Zod is a fairly significant bloating of Wrangler's bundle size for what seems like just these two validations. Would it be infeasible to do these manually?

packages/wrangler/src/images/local.ts
```diff
@@ -0,0 +1,187 @@
...
+			}
+
+			return runTransform(transformer, transforms.data, outputFormat);
+		} catch (e) {
+			return badTransformsResponse;
+		}
+	}
+}
+
+async function runInfo(transformer: Sharp): Promise<Response> {
```
penalosa, line 66: Is there going to be a docs page on what local mode supports/vs remote mode? Maybe a compat table?

packages/wrangler/src/pages/dev.ts
```diff
@@ -244,6 +244,12 @@ export function Options(yargs: CommonYargsArgv) {
 					"Bind to production Vectorize indexes in local development mode",
 				default: false,
 			},
+			"experimental-images-local-mode": {
+				type: "boolean",
+				describe:
+					"Use a local lower-fidelity implementation of the Images binding",
+				default: false,
```
penalosa, line 251: As above re default value

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -344,6 +346,13 @@ export function printBindings(
 		});
 	}
 
+	if (images !== undefined) {
+		output.push({
+			name: friendlyBindingNames.images,
+			entries: [{ key: "Name", value: images.binding }],
+		});
+	}
```
penalosa, line 354: This should have some explanation of whether or not it's local, using e.g. `addLocalSuffix(...)`. Although cc @andyjessop who's changing this around a bit

packages/wrangler/src/dev/miniflare.ts
ns476: We don't have any usage charges for the preview endpoint - these requests are pretty cheap compared to e.g. Workers AI

.npmrc
ns476: The fixtures CI checks fail without it, something to do with how sharp imports its native dependencies?

ns476: Actually, I needed to mark `sharp` as unbundled - works if I do that

packages/wrangler/src/dev.ts
ns476: The local mode is really missing a lot of features compared to remote, so I don't think we would really want it to be the default. I suppose nothing really prevents us from changing that down the line though, vs switching the default from local to remote.

packages/wrangler/src/__tests__/config/configuration.test.ts
```diff
@@ -2289,6 +2289,69 @@ describe("normalizeAndValidateConfig()", () => {
...
+				);
+
+				expect(diagnostics.hasWarnings()).toBe(false);
+				expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
+			"Processing wrangler configuration:
+			  - The field \\"images\\" should be an object but got 999."
+		`);
+			});
+
+			it("should error if ai is null", () => {
```
CarmenPopoviciu, line 2339:
```suggestion
			it("should error if images is null", () => {
```

packages/wrangler/src/images/local.ts
ns476: Yup, this will be explained in the docs

andyjessop (approved):
Approving here to unblock for merge, as all the conversations have been resolved.

PR #7426, opened 2024-12-03
fix: allow the asset directory to be omitted in Wrangler config for commands that don't need it

petebacondarwin:
Fixes #0000

We want to allow build tools (e.g. Nuxt/Vite/Next.js etc) to own the `main` and `assets.directory` fields in deployment configuration since they are the ones that generate these files.
In that case it doesn't makes sense for the user to provide them in their user configuration.
The `main` field is already optional from the point of view of the syntax check of the user's configuration file (e.g. wrangler.toml); this PR just makes the `assets.directory` field the same.
It is still an error to attempt to deploy or dev a Worker with an `assets` field but no `assets.directory`.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: the docs don't specify that the wrangler.toml must have a directory, just that a directory is required.

### DIFF
--- a/.changeset/forty-gifts-grab.md
+++ b/.changeset/forty-gifts-grab.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: allow the asset directory to be omitted in Wrangler config for commands that don't need it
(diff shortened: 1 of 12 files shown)


penalosa (approved):
LGTM, but could you add some context to the PR description for why this is needed?

PR #7427, opened 2024-12-03
feat(wrangler): identify draft and inherit bindings and provision draft ones

edmundhung:
Fixes [DEVX-1421](https://jira.cfdata.org/browse/DEVX-1421) and [DEVX-1420](https://jira.cfdata.org/browse/DEVX-1420)

This PR adds the ability to identify draft and inherit bindings by looking up the current binding settings.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: unreleased feature

### DIFF
--- a/packages/wrangler/e2e/provision.test.ts
+++ b/packages/wrangler/e2e/provision.test.ts
@@ -1,0 +1,173 @@
+import assert from "node:assert";
+import dedent from "ts-dedent";
+import { fetch } from "undici";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { CLOUDFLARE_ACCOUNT_ID } from "./helpers/account-id";
+import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
+import { fetchText } from "./helpers/fetch-text";
+import { generateResourceName } from "./helpers/generate-resource-name";
+import { normalizeOutput } from "./helpers/normalize";
+import { retry } from "./helpers/retry";
+
+const TIMEOUT = 500_000;
+const normalize = (str: string) => {
+	return normalizeOutput(str, {
+		[CLOUDFLARE_ACCOUNT_ID]: "CLOUDFLARE_ACCOUNT_ID",
+	}).replaceAll(
+		/- KV: ([0-9a-f]{32})/gm,
+		"- KV: 00000000000000000000000000000000"
+	);
+};
+const workerName = generateResourceName();
+
+describe("provisioning", { timeout: TIMEOUT }, () => {
+	let deployedUrl: string;
+	let kvId: string;
+	let d1Id: string;
+	const helper = new WranglerE2ETestHelper();
+
+	it.skip("can run dev without resource ids", async () => {
+		const worker = helper.runLongLived("wrangler dev --x-provision", {
+			debug: true,
+		});
+
+		const { url } = await worker.waitForReady();
+		await fetch(url);
+
+		const text = await fetchText(url);
+
+		expect(text).toMatchInlineSnapshot(`"Hello World!"`);
+	});
+
+	beforeAll(async () => {
+		await helper.seed({
+			"wrangler.toml": dedent`
+						name = "${workerName}"
+						main = "src/index.ts"
+						compatibility_date = "2023-01-01"
+
+						[[kv_namespaces]]
+						binding = "KV"
+
+						[[r2_buckets]]
+						binding = "R2"
+
+						[[d1_databases]]
+						binding = "D1"
+						`,
+			"src/index.ts": dedent`
+						export default {
+							fetch(request) {
+								return new Response("Hello World!")
+							}
+						}`,
+			"package.json": dedent`
+						{
+							"name": "${workerName}",
+							"version": "0.0.0",
+							"private": true
+						}
+						`,
+		});
+	});
+
+	it("can provision resources and deploy worker", async () => {
+		const worker = helper.runLongLived(
+			`wrangler deploy --x-provision --x-auto-create`
+		);
+		await worker.exitCode;
+		const output = await worker.output;
+		expect(normalize(output)).toMatchInlineSnapshot(`
+			"Total Upload: xx KiB / gzip: xx KiB
+			The following bindings need to be provisioned:
+			- KV Namespaces:
+			  - KV
+			- D1 Databases:
+			  - D1
+			- R2 Buckets:
+			  - R2
+			Provisioning KV (KV Namespace)...
+			🌀 Creating new KV Namespace "tmp-e2e-worker-00000000-0000-0000-0000-000000000000-kv"...
+			✨ KV provisioned with tmp-e2e-worker-00000000-0000-0000-0000-000000000000-kv
+			--------------------------------------
+			Provisioning D1 (D1 Database)...
+			🌀 Creating new D1 Database "tmp-e2e-worker-00000000-0000-0000-0000-000000000000-d1"...
+			✨ D1 provisioned with tmp-e2e-worker-00000000-0000-0000-0000-000000000000-d1
+			--------------------------------------
+			Provisioning R2 (R2 Bucket)...
+			🌀 Creating new R2 Bucket "tmp-e2e-worker-00000000-0000-0000-0000-000000000000-r2"...
+			✨ R2 provisioned with tmp-e2e-worker-00000000-0000-0000-0000-000000000000-r2
+			--------------------------------------
+			🎉 All resources provisioned, continuing with deployment...
+			Your worker has access to the following bindings:
+			- KV Namespaces:
+			  - KV: 00000000000000000000000000000000
+			- D1 Databases:
+			  - D1: 00000000-0000-0000-0000-000000000000
+			- R2 Buckets:
+			  - R2: tmp-e2e-worker-00000000-0000-0000-0000-000000000000-r2
+			Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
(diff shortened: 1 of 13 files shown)


andyjessop (commented):
This is awesome work! The only suggestion I have is to have a clearer separation of concerns in the provisioner, so that it doesn't get too unwieldy later on, and so we can move more confidently towards modularisation of wrangler modules.

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +70,258 @@ export function getBindings(
...
+type PendingResourceOperations = {
+	create: (name: string) => Promise<string>;
+	updateId: (id: string) => void;
+};
+type PendingResources = {
+	kv_namespaces: (CfKvNamespace & PendingResourceOperations)[];
+	r2_buckets: (CfR2Bucket & PendingResourceOperations)[];
+	d1_databases: (CfD1Database & PendingResourceOperations)[];
+};
+export async function provisionBindings(
```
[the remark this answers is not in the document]
emily-shen: I think we would always want this to be done as part of a deploy/upload - the key benefit of this provisioning work is that you can create/link resources at the same as time as main worker deployment.

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +70,258 @@ export function getBindings(
...
+			} else {
+				pendingResources.d1_databases?.push({
+					binding: d1.binding,
+					async create(name) {
+						const db = await createD1Database(accountId, name);
+						d1.database_id = db.uuid;
+						return db.uuid;
+					},
+					updateId(id) {
+						// tODO check d1 isn't doing something funny here
```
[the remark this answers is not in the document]
emily-shen: (it wasn't doing any funny :'))

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +70,258 @@ export function getBindings(
...
+	};
+	let settings: Settings | undefined;
+
+	try {
+		settings = await getSettings(accountId, scriptName);
+	} catch (error) {
+		logger.debug("No settings found");
+	}
+
+	for (const kv of bindings.kv_namespaces ?? []) {
```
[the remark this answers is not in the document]
emily-shen: happy to add to the momentum for modularising wrangler! thanks for the very thorough comment too :)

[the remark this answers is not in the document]
emily-shen: ...although in the interest of time, I'll probably revisit this later (around new year's when i'm on call)?

packages/wrangler/src/config/index.ts
```diff
@@ -618,13 +618,24 @@ export function printBindings(
...
 	const message = [
-		`${context.name && getFlag("MULTIWORKER") ? chalk.blue(context.name) : "Your worker"} has access to the following bindings:`,
+		title,
 		...output
 			.map((bindingGroup) => {
 				return [
 					`- ${bindingGroup.name}:`,
-					bindingGroup.entries.map(({ key, value }) => `  - ${key}: ${value}`),
+					bindingGroup.entries.map(
+						({ key, value }) => `  - ${key}${value ? ":" : ""} ${value}`
```
emily-shen, line 637: this just removes the colon if there is no value (ie for inherited bindings). I do think we maybe should print something for inherited bindings (id? name?) and that's totally possible as we've got the worker settings, but just wanted to discuss first and not block this PR on that.

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +69,258 @@ export function getBindings(
...
+	logger.log();
+	logger.log(chalk.dim("--------------------------------------"));
+	logger.log();
+}
+
+type NormalisedResourceInfo = {
+	name: string;
+	id: string;
+};
+type ResourceType = "d1_databases" | "r2_buckets" | "kv_namespaces";
```
penalosa, line 235: Is this just `keyof PendingResources`?

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +69,258 @@ export function getBindings(
...
+	resourceKeyDescriptor: string,
+	scriptName: string,
+	autoCreate: boolean
+) {
+	const MAX_OPTIONS = 4;
+	if (pending.length) {
+		const options = preExisting
+			.map((resource) => ({
+				title: resource.name,
+				value: resource.id,
```
penalosa, line 249: Given we're mapping to `value`, why not just have `NormalisedResourceInfo` be `{ title: string, value: string }`?

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +69,258 @@ export function getBindings(
...
+			}))
+			.slice(0, MAX_OPTIONS - 1);
+		if (options.length < preExisting.length) {
+			options.push({
+				title: "Other (too many to list)",
+				value: "manual",
```
penalosa, line 255: This should be a symbol—what if a user has an R2 bucket called `manual`?

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +69,258 @@ export function getBindings(
...
+			});
+		}
+
+		for (const item of pending) {
+			logger.log("Provisioning", item.binding, `(${friendlyBindingName})...`);
+			let name: string = "";
+			let selected: string;
+			if (options.length === 0 || autoCreate) {
+				selected = "new";
```
penalosa, line 264: As above with `manual`

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +69,258 @@ export function getBindings(
...
+							return true;
+						} else {
+							return false;
+						}
+					});
+					if (!searchedResource) {
+						logger.log(
+							`No ${friendlyBindingName} with that ${resourceKeyDescriptor} "${input}" found. Please try again.`
+						);
+					}
```
penalosa, line 303: Love this!

packages/wrangler/src/dev.ts
```diff
@@ -377,7 +377,7 @@ export const dev = createCommand({
 			{
 				FILE_BASED_REGISTRY: args.experimentalRegistry,
 				MULTIWORKER: Array.isArray(args.config),
-				RESOURCES_PROVISION: false,
+				RESOURCES_PROVISION: args.experimentalProvision ?? false,
```
penalosa, line 380: `dev` should ever be provisioning, I don't think?

emily-shen: The flag is also used to gate the config id check so it’s needed for now at least otherwise it’ll throw a missing id error

packages/wrangler/src/d1/list.ts
```diff
@@ -55,6 +56,9 @@ export const listDatabases = async (
 		);
 		page++;
 		results.push(...json);
+		if (limitCalls) {
+			break;
+		}
```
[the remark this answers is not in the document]
emily-shen: added this already to KV (should be in the same commit), and R2 seems to be happy to return everything in one go

packages/wrangler/src/deployment-bundle/bindings.ts
emily-shen: I don't think the select component will let me make the value a symbol, but I can add a prefix like WRANGLER_INTERNAL_?

packages/wrangler/src/deployment-bundle/bindings.ts
emily-shen: https://github.com/cloudflare/workers-sdk/pull/7427#discussion_r1880246167

packages/wrangler/src/deployment-bundle/bindings.ts
penalosa: That sounds good!

packages/wrangler/e2e/provision.test.ts
```diff
@@ -0,0 +1,173 @@
...
+
+						[[d1_databases]]
+						binding = "D1"
+						database_name = "${workerName}-d1"
+						database_id = "${d1Id}"
+						`,
+		});
+		let output = await helper.run(`wrangler r2 bucket delete ${workerName}-r2`);
+		expect(output.stdout).toContain(`Deleted bucket`);
+		output = await helper.run(`wrangler d1 delete ${workerName}-d1 -y`, {
```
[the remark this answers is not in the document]
penalosa: Ah cool, okay. cc @rozenmd out of scope for this PR, but is that something D1 wants to support?

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -62,3 +69,263 @@ export function getBindings(
...
+					},
+					updateId(id) {
+						d1.database_id = id;
+					},
+				});
+			}
+		}
+	}
+
+	if (Object.values(pendingResources).some((v) => v && v.length > 0)) {
```
penalosa, line 165: Can we gate this around a service env check and throw otherwise? i.e. `if(!isLegacyEnv(config)){ throw new UserError("You can't provision with a service environment")}`

emily-shen: [ee0a46d](https://github.com/cloudflare/workers-sdk/pull/7427/commits/ee0a46d06e4e767bea362478165237ff531d0f69)

PR #7428, opened 2024-12-03
Changed port in playwright config file

Oreoezi:
One issue that I had creating a Sveltekit project using create-cloudflare was the fact that the playwright configuration file still had its port set to the default of 4173, which would cause the test to hang. I could either change the `webServer.command` property to override the default port that miniflare uses or to change the port to 8788 (which is what I chose).

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: The changes are relatively simple.
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: The changes are relatively simple.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: It does not change in any way the user experience.

### DIFF
--- a/.changeset/lovely-kiwis-grab.md
+++ b/.changeset/lovely-kiwis-grab.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+Use correct port in playwright config file for generated SvelteKit apps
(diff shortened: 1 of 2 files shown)


penalosa (approved):
Probably more trouble than it's worth to add a test here, but I've validated it manually

penalosa:
@Oreoezi could you fix the formatting issues here? Running `pnpm prettify` should be enough

Oreoezi:
@penalosa I ran prettier. lmk if that fixed all the formatting issues.

PR #7437, opened 2024-12-04
chore: convert versions and versions secret to createCommand

andyjessop:
Fixes #000.

Converts `wrangler versions` and `wrangler versions secret` to use the `CommandRegistry`.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: already covered
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/__tests__/versions/versions.deploy.test.ts
```diff
@@ -925,39 +924,28 @@ describe("versions deploy", () => {
...
-			versionsDeployOptions,
-			() => {}
-		);
-
 		test("no args", () => {
-			const input = "versions deploy";
-
-			const args = options.parse(input) as VersionsDeployArgs;
-			const result = parseVersionSpecs(args);
+			const result = parseVersionSpecs({});
```
andyjessop, line 928: These are now pure unit tests, rather than testing any yargs functionality. So the ordering tests have been removed (not the concern of this function), and there is no need to build and parse the yargs object.

packages/wrangler/src/versions/view.ts
```diff
@@ -1,141 +1,141 @@
...
->;
-
-export function versionsViewOptions(yargs: CommonYargsArgv) {
-	return yargs
-		.positional("version-id", {
+export const versionsViewCommand = createCommand({
+	metadata: {
+		description: "View the details of a specific version of your Worker",
+		owner: "Workers: Authoring and Testing",
+		status: "open-beta",
```
penalosa, line 18: I think this is now GA: https://blog.cloudflare.com/builder-day-2024-announcements/#gradual-deployments

packages/wrangler/src/versions/upload.ts
```diff
@@ -112,6 +125,282 @@ function errIsStartupErr(err: unknown): err is ParseError & { code: 10021 } {
...
+			requiresArg: true,
+		},
+		message: {
+			describe:
+				"A descriptive message for this Worker Gradual Rollouts Version",
+			type: "string",
+			requiresArg: true,
+		},
+	},
+	handler: async function versionsUploadHandler(args, { config }) {
```
penalosa, line 282: I don't think we can use this config as-is—this command used `const configPath =
		args.config ||
		(args.script && findWranglerConfig(path.dirname(args.script)));` to locate the config, logic which isn't in defineCommand

packages/wrangler/src/versions/list.ts
```diff
@@ -1,93 +1,79 @@
...
+		json: {
 			describe: "Display output as clean JSON",
 			type: "boolean",
 			default: false,
-		});
-}
+		},
+	},
+	handler: async function versionsSecretListHandler(args) {
+		const config = readConfig(args.config, args, false, true);
```
penalosa, line 32: Do we need this?

packages/wrangler/src/versions/upload.ts
andyjessop: Updated here: https://github.com/cloudflare/workers-sdk/pull/7437/commits/c447e879c97b5a0b435e2926dbff2413e998833e

packages/wrangler/src/versions/list.ts
andyjessop: Updated here: https://github.com/cloudflare/workers-sdk/pull/7437/commits/25c9e043febc6597c04036501909c771b2632722

penalosa (approved):
Deferring beta re-classification for now

PR #7442, opened 2024-12-04
feat: add support for redirecting Wrangler to a generated config when running deploy commands

### LINKED ISSUE #5579: 🚀 Feature Request: support deployments with dynamically generated wrangler.toml (or runtime config in another format)

petebacondarwin:
Fixes #5579

This feature is designed for build tools and frameworks to provide a deploy-specific configuration, which Wrangler can use instead of user configuration when running deploy commands.
It is not expected that developers of Workers will need to use this feature directly.

### Affected commands

The commands that use this feature are:

- `wrangler deploy`
- `wrangler dev`
- `wrangler versions upload`
- `wrangler versions deploy`
- `wrangler pages deploy`
- `wrangler pages build`
- `wrangler pages build-env`

### Config redirect file

When running these commands, Wrangler will look up the directory tree from the current working directory for a file at the path `.wrangler/deploy/config.json`. This file must contain only a single JSON object of the form:

```json
{ "configPath": "../../path/to/wrangler.json" }
```

When this file exists Wrangler will use the `configPath` (relative to the `config.json` file) to find an alternative Wrangler configuration file to load and use as part of this command.

When this happens Wrangler will display a warning to the user to indicate that the configuration has been redirected to a different file than the user's configuration file.

### Custom build tool example

A common approach that a build tool might choose to implement.

- The user writes code that uses Cloudflare Workers resources, configured via a user `wrangler.toml` file.

  ```toml
  name = "my-worker"
  main = "src/index.ts"
  [[kv_namespaces]]
  binding = "<BINDING_NAME1>"
  id = "<NAMESPACE_ID1>"
  ```

  Note that this configuration points `main` at user code entry-point.

- The user runs a custom build, which might read the `wrangler.toml` to find the entry-point:

  ```bash
  > my-tool build
  ```

- This tool generates a `dist` directory that contains both compiled code and a new deployment configuration file, but also a `.wrangler/deploy/config.json` file that redirects Wrangler to this new deployment configuration file:

  ```plain
  - dist
    - index.js
  	- wrangler.json
  - .wrangler
    - deploy
  	  - config.json
  ```

  The `dist/wrangler.json` will contain:

  ```json
  {
  	"name": "my-worker",
  	"main": "./index.js",
  	"kv_namespaces": [{ "binding": "<BINDING_NAME1>", "id": "<NAMESPACE_ID1>" }]
  }
  ```

  And the `.wrangler/deploy/config.json` will contain:

  ```json
  {
  	"configPath": "../../dist/wrangler.json"
  }
  ```

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18757


tools/e2e/runIndividualE2EFiles.ts
```diff
@@ -28,5 +28,6 @@ for (const file of e2eTests) {
 for (const task of tasks.values()) {
 	execSync(task, {
 		stdio: "inherit",
+		env: process.env,
```
petebacondarwin, line 31: This is needed so that you can run `pnpm run test:e2e:wrangler` locally while passing environment variables in at the command line. E.g.

```bash
CLOUDFLARE_ACCOUNT_ID=xxxx CLOUDFLARE_API_TOKEN=yyyy  WRANGLER="node $PWD/packages/wrangler/bin/wrangler.js" WRANGLER_IMPORT=$PWD/packages/wrangler/wrangler-dist/cli.js pnpm run test:e2e:wrangler
```

petebacondarwin: Actually this doesn't seem to help 😢 
I'll revert this change.

fixtures/import-npm/package.json
```diff
@@ -7,9 +7,10 @@
...
 	"scripts": {
-		"check:type": "rm -rf node_modules && npm install && npm run check:type --workspaces",
-		"test:ci": "npm install && npm run test:ci --workspaces",
-		"test:watch": "npm install && npm run test:watch --workspaces",
-		"type:tests": "rm -rf node_modules && npm install && npm run type:tests --workspaces"
+		"_clean_install": "rm -rf node_modules && npm install",
+		"check:type": "npm run _clean_install && npm run check:type --workspaces",
+		"test:ci": "npm run _clean_install && npm run test:ci --workspaces",
+		"test:watch": "npm run _clean_install && npm run test:watch --workspaces",
+		"type:tests": "npm run _clean_install && npm run type:tests --workspaces"
```
petebacondarwin, line 14: I found that this fixture's tests kept failing; and I think it was because we need to ensure that the node_modules directory was in fact created by the `npm i` command here, and not a `pnpm i` command that is naturally run when setting up the workers-sdk project.

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -9,115 +9,117 @@ import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
 import { fetchText } from "./helpers/fetch-text";
 import { normalizeOutput } from "./helpers/normalize";
 
-describe.each([{ cmd: "wrangler pages dev" }])("Pages $cmd", ({ cmd }) => {
-	it("should warn if no [--compatibility_date] command line arg was specified", async () => {
-		const helper = new WranglerE2ETestHelper();
-		await helper.seed({
-			"_worker.js": dedent`
+const port = await getPort();
+const inspectorPort = await getPort();
```
petebacondarwin, line 13: The only changes in this file are to explicitly set both the port and inspector port to avoid concurrent tests from failing due to port 9229 being blocked.

packages/wrangler/src/config/validation.ts
```diff
@@ -327,7 +330,7 @@ function applyPythonConfig(
 	config: Config,
 	args: NormalizeAndValidateConfigArgs
 ) {
-	const mainModule = "script" in args ? args.script : config.main;
+	const mainModule = args.script ?? config.main;
```
petebacondarwin, line 333: It is possible that the `args.script` property exists but is set to `undefined`, which means that we should still defer to `config.main` in that case.

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
```diff
@@ -0,0 +1,254 @@
...
+		});
+
+		it("should error if deploy config does not contain a `configPath` property", async () => {
+			await seed({
+				[".wrangler/deploy/config.json"]: `{}`,
+			});
+
+			let error;
+			try {
+				findWranglerConfig(".", { useRedirect: true });
```
penalosa, line 188: As above re `toThrowErrorMatchingInlineSnapshot()`

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
-	);
+		findUpSync(`wrangler.toml`, { cwd: referencePath });
+
+	return useRedirect
+		? findRedirectedWranglerConfig(referencePath, userConfigPath)
+		: userConfigPath;
+}
+
+/**
+ * Check whether there is a config file that indicates
```
penalosa, line 65: Is this comment complete?

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+ * @param cwd
+ * @param userConfigPath
+ * @returns
+ */
+function findRedirectedWranglerConfig(
+	cwd: string,
+	userConfigPath: string | undefined
+) {
+	const PATH_TO_DEPLOY_CONFIG = ".wrangler/deploy/config.json";
+	const deployConfigPath = findUpSync(PATH_TO_DEPLOY_CONFIG, { cwd });
```
penalosa, line 75: ~Should this be a `findUpSync()`? I almost feel like this should just look directly for that path starting from the dirname of the user config file~

Never mind, I see this is an error further down

packages/wrangler/src/type-generation/index.ts
```diff
@@ -63,7 +63,7 @@ export async function typesHandler(
 
 	await printWranglerBanner();
 
-	const configPath = resolveWranglerConfigPath(args);
+	const configPath = resolveWranglerConfigPath(args, {});
```
penalosa: Should type generation not use this? What if the generated config file has a different set of compat flags to the user one?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -96,7 +101,8 @@ export const versionsDeployCommand = createCommand({
 		},
 	},
 	positionalArgs: ["version-specs"],
-	handler: async function versionsDeployHandler(args, { config }) {
+	handler: async function versionsDeployHandler(args) {
+		const config = readConfig(args, { useRedirect: true });
```
penalosa, line 105: Could this be added as an option to the command definition helpers?

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -23,9 +30,9 @@ export function resolveWranglerConfigPath(
 		script?: string;
 	},
 	options: { useRedirect?: boolean }
-): string | undefined {
+): ConfigPaths {
 	if (config !== undefined) {
-		return config;
+		return { userConfigPath: config, configPath: config };
 	}
```
penalosa, line 36: Does this mean that config redirection won't happen if the user specifies `--config` directly? 

That seems reasonable, just wanted to double check

packages/wrangler/src/deploy/index.ts
```diff
@@ -271,17 +270,17 @@ async function deployWorker(args: DeployArgs) {
...
-	});
-	const projectRoot = userConfigPath && path.dirname(userConfigPath);
 	const config = readConfig(args, { useRedirect: true });
 	if (config.pages_build_output_dir) {
 		throw new UserError(
 			"It looks like you've run a Workers-specific command in a Pages project.\n" +
 				"For Pages, please run `wrangler pages deploy` instead."
 		);
 	}
+	// We use the `userConfigPath` to compute the root of a Pages project,
```
penalosa, line 280: This looks like a normal deploy, not Pages?

packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -8,6 +8,9 @@ import WebSocket from "ws";
 import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
 import { generateResourceName } from "./helpers/generate-resource-name";
 
+const port = await getPort();
+const inspectorPort = await getPort();
```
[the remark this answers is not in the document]
penalosa: These tests don't run concurrently in CI—was this a problem you were seeing locally?

penalosa (approved):
Some small comments, but overall LGTM ✅

packages/wrangler/src/type-generation/index.ts
petebacondarwin: Yes I thought about that. In that case the code that expects to have the compat flag is part of the generated code, not the user code. If the user wants to use such APIs then they should be defining the compat flag in their user config.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: Yes that is precisely the behaviour.

packages/wrangler/src/deploy/index.ts
petebacondarwin: Yes you are right. Thanks

packages/wrangler/src/deploy/index.ts
```diff
@@ -346,11 +346,7 @@ async function deployWorker(args: DeployArgs) {
 
 	if (!args.dryRun) {
 		assert(accountId, "Missing account ID");
-		await verifyWorkerMatchesCITag(
-			accountId,
-			name,
-			path.relative(entry.projectRoot, config.configPath ?? "wrangler.toml")
-		);
+		await verifyWorkerMatchesCITag(accountId, name, config.configPath);
```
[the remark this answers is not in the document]
petebacondarwin: Yes both `configPath` and` userConfigPath` are resolved to absolute paths as part of `resolveWranglerConfigPath()`.

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
```diff
@@ -0,0 +1,254 @@
...
+		});
+
+		it("should error if deploy config is not valid JSON", async () => {
+			await seed({
+				[".wrangler/deploy/config.json"]: `INVALID JSON`,
+			});
+
+			let error;
+			try {
+				findWranglerConfig(".", { useRedirect: true });
```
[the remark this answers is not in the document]
petebacondarwin: Sadly not because of the need to run `normalizeString()` on the error. Otherwise we get Windows test failures.

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
```diff
@@ -0,0 +1,254 @@
...
+			} catch (e) {
+				error = e;
+			}
+
+			expect(normalizeString(`${error}`)).toMatchInlineSnapshot(
+				`"Error: Failed to load the deploy config at .wrangler/deploy/config.json"`
```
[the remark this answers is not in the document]
petebacondarwin: The error itself does contain a `cause` which is the ParseError.
And this would get logged out (sort of) in an actual Wrangler run.
But I can make that better. Thanks

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
petebacondarwin: Same as above. We need to normalize the error messge.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: Nope! Thanks

packages/wrangler/src/versions/deploy.ts
petebacondarwin: Yes but I worried that this would expand the scope of this PR.
I can do it in a follow up?

petebacondarwin (commented):
Thanks for the review @penalosa !

packages/wrangler/src/deploy/index.ts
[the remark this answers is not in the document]
CarmenPopoviciu: can we specify this in `resolveWranglerConfigPath`'s jsdoc? Might be useful for consumers of this fn to know that an abs path is returned, without them having to step into the fn code

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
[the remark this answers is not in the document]
petebacondarwin: I improved this.

packages/wrangler/src/deploy/index.ts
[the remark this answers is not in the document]
petebacondarwin: Good call. Will do.

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+		throw new UserError(
+			dedent`
+				Failed to load the deploy config at ${path.relative(".", deployConfigPath)}
+			`,
+			{ cause: e }
+		);
+	}
+	if (!redirectedConfigPath) {
+		throw new UserError(dedent`
+			A redirect config was found at "${path.relative(".", deployConfigPath)}".
```
CarmenPopoviciu, line 102: `A redirect configuration file was fount at ...`

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+			\`\`\`
+			${deployConfigFile}
+			\`\`\`
+		`);
+	}
+
+	if (redirectedConfigPath) {
+		if (!fs.existsSync(redirectedConfigPath)) {
+			throw new UserError(dedent`
+				There is a redirect configuration at "${path.relative(".", deployConfigPath)}".
```
CarmenPopoviciu, line 114: `There is a redirect configuration file ....`

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+				But the config path it points to, "${path.relative(".", redirectedConfigPath)}", does not exist.
+			`);
+		}
+		if (userConfigPath) {
+			if (
+				path.join(path.dirname(userConfigPath), PATH_TO_DEPLOY_CONFIG) !==
+				deployConfigPath
+			) {
+				throw new UserError(dedent`
+					Found both a user config file at "${path.relative(".", userConfigPath)}"
```
CarmenPopoviciu, line 124: can we please use `configuration` instead of `config` everywhere in our copy, since this is user facing 🙏

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+		: userConfigPath;
+}
+
+/**
+ * Check whether there is a config file that indicates
+ * @param cwd
+ * @param userConfigPath
+ * @returns
+ */
+function findRedirectedWranglerConfig(
```
CarmenPopoviciu, line 72: nit: `findRedirectWranglerConfig()` instead of `..Redirected..`

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+	cwd: string,
+	userConfigPath: string | undefined
+) {
+	const PATH_TO_DEPLOY_CONFIG = ".wrangler/deploy/config.json";
```
CarmenPopoviciu, line 76: should we extract this const somewhere higher level? I see it's being used all over our fixtures.

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+	const deployConfigPath = findUpSync(PATH_TO_DEPLOY_CONFIG, { cwd });
+	if (deployConfigPath === undefined) {
```
CarmenPopoviciu, line 78: what does this mean tho? Basically the code in the calling fn becomes:
```
configPath: useRedirect
			? userConfigPath
			: userConfigPath,
```
Does this mean we set `useRedirect` but forgot to provide a redirect config path? Won't returning userConfigPath without any warning result in an unexpected behaviour for users? Just thinking out loud

packages/wrangler/src/config/config-helpers.ts
CarmenPopoviciu, line 76: why only deploy? what does this mean? that we're using this redirect config file for deploy only? I'm pretty sure the answer is no, but the naming got me confused there

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+			deployConfigFile,
+			deployConfigPath
+		);
+		redirectedConfigPath =
+			deployConfig.configPath &&
+			path.resolve(path.dirname(deployConfigPath), deployConfig.configPath);
+	} catch (e) {
+		throw new UserError(
+			dedent`
+				Failed to load the deploy config at ${path.relative(".", deployConfigPath)}
```
CarmenPopoviciu, line 95: `Failed to load the deploy configuration file ....`

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,127 @@
...
+					Found both a user config file at "${path.relative(".", userConfigPath)}"
+					and a redirect config file at "${path.relative(".", deployConfigPath)}".
+					But these do not share the same base path so it is not clear which should be used.
+				`);
+			}
+		}
+
+		logger.warn(dedent`
+			Using redirected Wrangler configuration.
+			Redirected config path: "${path.relative(".", redirectedConfigPath)}"
```
CarmenPopoviciu, line 133: oh boy...I find this confusing 😵. What's the diff between the redirected/deploy/original config path? The original I'm assuming is `wrangler.toml` proper in the user's project?

packages/wrangler/src/versions/deploy.ts
```diff
@@ -47,6 +48,10 @@ export const versionsDeployCommand = createCommand({
 		owner: "Workers: Authoring and Testing",
 		status: "stable",
 	},
+	behaviour: {
+		provideConfig: false,
```
CarmenPopoviciu, line 52: why don't we need `useConfigRedirect: true,` here too?

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: I did mean to use "redirected" here. It is getting the configuration path "after" redirection.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: This was the name we came up with when we were in London with Igor.
We are only using this file for deployment type commands (with the slightly complicated addition of `dev` commands).
Any commands that don't involve deploying/running actual Worker code do not use this configuration.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: Thanks for this feedback. I'll rewrite this to something that might be better.

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -23,9 +30,9 @@ export function resolveWranglerConfigPath(
 		script?: string;
 	},
 	options: { useRedirect?: boolean }
-): string | undefined {
+): ConfigPaths {
 	if (config !== undefined) {
-		return config;
+		return { userConfigPath: config, configPath: config };
```
CarmenPopoviciu, line 37: so they're both the same value?

packages/wrangler/src/config/config-helpers.ts
CarmenPopoviciu: after reading more of the code, I **think** this makes sense, because `useRedirect` serves as a a flag that indicates we should check whether there's a redirect config? For example our deploy commands, which can't know beforehand whether a redirect config is present, so we set `useRedirect` to `true` by default, which would ensure we always check for a redirect file. If no redirect file is found, then we proceed with the user config file. Did I get this right? And do I win a star if I did? 🌟 😋

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: Inside Wrangler I think this is only used here.
The fixtures are like user applications, so unless we exported this constant from the `wrangler` npm package, I am not sure where else it should go.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: When the user provide a config path at the command line (e.g. `wrangler deploy -c my/wrangler.toml`) then we do not attemtpt to redirect and, indeed, these two are the same.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: Yes, I guess we could rename `useRedirect` to something like `tryRedirect`, which might make it clearer that we are trying to use a redirect if one is available in a `.wrangler/deploy/config.json`?

packages/wrangler/src/config/config-helpers.ts
CarmenPopoviciu: lol...you're ofc right! ignore me this time

packages/wrangler/src/deployment-bundle/resolve-entry.ts
```diff
@@ -12,31 +13,33 @@ export function resolveEntryWithScript(script: string): {
...
 ): {
 	absolutePath: string;
 	relativePath: string;
 	projectRoot: string;
 } {
-	const projectRoot = path.resolve(path.dirname(configPath ?? "."));
-	const file = path.resolve(projectRoot, main);
-	const relativePath = path.relative(projectRoot, file) || ".";
-	return { absolutePath: file, relativePath, projectRoot };
+	const projectRoot = path.resolve(path.dirname(config.userConfigPath ?? "."));
```
CarmenPopoviciu, line 22: could we add a comment here explaining why the `projectRoot` and `entryRoot` can potentially diverge into 2 diff paths?

CarmenPopoviciu: also I wonder if we won;t need to make this kind of distinction in other code paths as well, in which case we'd prolly want to extract these two into a fn that returns them? Still going through the code, so might be over engineering...I'll defer to your judgement. Just wanted to call it out

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -11,15 +11,17 @@ export type ResolveConfigPathOptions = {
 };
 
 export type ConfigPaths = {
-	/** The path to the actual configuration being used (possibly redirected from the user's config). */
+	/** Absolute path to the actual configuration being used (possibly redirected from the user's config). */
```
CarmenPopoviciu, line 14: 🙏 ♥️

packages/wrangler/src/versions/deploy.ts
petebacondarwin: Actually it is added already! `useConfigRedirectIfAvailable`...

packages/wrangler/src/versions/deploy.ts
petebacondarwin: Oh yes! We can.

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
```diff
@@ -0,0 +1,280 @@
...
+		});
+
+		it("should return user config path even if a deploy config is found", async () => {
+			await seed({
+				[`wrangler.toml`]: "DUMMY",
+				[".wrangler/deploy/config.json"]: `{"configPath": "../../dist/wrangler.json" }`,
+				[`dist/wrangler.json`]: "DUMMY",
+			});
+			expect(
+				findWranglerConfig(".", { useRedirectIfAvailable: false })
```
vicb, line 80: `{ useRedirectIfAvailable: false }` must be the default value because it is not specified for other calls to `findWranglerConfig` in this describe.

Should we be consistent in (not) using it. Another option would be to test both either explicitly`expect(
				findWranglerConfig({ useRedirectIfAvailable: false })).toEqual(findWranglerConfig())` or with a parametrized test (`[{useRedirectIfAvailable: false}, undefined]`)

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
```diff
@@ -0,0 +1,280 @@
...
+				  "info": "",
+				  "out": "",
+				  "warn": "[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mUsing redirected Wrangler configuration.[0m
+
+				  Configuration being used: \\"dist/wrangler.json\\"
+				  Original user's configuration: \\"<no user config found>\\"
+				  Deploy configuration file: \\".wrangler/deploy/config.json\\"
+
+
+				[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mUsing redirected Wrangler configuration.[0m
```
vicb, line 119: is it expected to get a duplicated warning?

vicb: Oh, probably we test a child dir

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,139 @@
...
-}): string | undefined {
+export function resolveWranglerConfigPath(
+	{
+		config,
+		script,
+	}: {
+		config?: string;
+		script?: string;
+	},
+	options: { useRedirectIfAvailable?: boolean }
```
vicb, line 34: Should this (and the default value/behavior) be documented?

packages/wrangler/src/config/index.ts
```diff
@@ -182,29 +201,32 @@ export interface DotEnv {
 	parsed: dotenv.DotenvParseOutput;
 }
 
-function tryLoadDotEnv(path: string): DotEnv | undefined {
+function tryLoadDotEnv(basePath: string): DotEnv | undefined {
```
vicb, line 204: not sure to get the renaming here?

fixtures/import-npm/package.json
petebacondarwin: Furthemore I couldn't just run _clean_install in front of each of these scripts because turbo tries to run then in parallel. So the solution was to add a turbo dependency instead, which means that this only runs once.

packages/wrangler/src/versions/deploy.ts
petebacondarwin: Updated this command to use it.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: Renamed to `useRedirectIfAvailable`.

packages/wrangler/src/versions/deploy.ts
petebacondarwin: Done

packages/wrangler/src/deployment-bundle/resolve-entry.ts
petebacondarwin: This is the main place I believe where it matters. The `getEntry()` is effectively the helper that covers this stuff.

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
petebacondarwin: Removed.

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
petebacondarwin: Yes we are calling `findWranglerConfig()` twice so we expect two warnings in the logs.

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: Yes documentation is in https://github.com/cloudflare/cloudflare-docs/pull/18757

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -1,35 +1,139 @@
...
- * Find the wrangler config file by searching up the file-system
+ * Find the wrangler configuration file by searching up the file-system
  * from the current working directory.
  */
 export function findWranglerConfig(
-	referencePath: string = process.cwd()
-): string | undefined {
-	return (
+	referencePath: string = process.cwd(),
+	{ useRedirectIfAvailable = false } = {}
```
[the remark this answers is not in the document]
petebacondarwin: Will be in https://github.com/cloudflare/cloudflare-docs/pull/18757

packages/wrangler/src/config/index.ts
petebacondarwin: Because we are using `path` as the default import from `node:path` at the top of the file now. So this renaming avoids the conflict/shadowing.

PR #7450, opened 2024-12-05
fix: ensure that version secrets commands do not write wrangler config warnings

petebacondarwin:
Fixes #0000

Ensures that version secrets commands do not write wrangler config warnings.
This was a regression introduced by https://github.com/cloudflare/workers-sdk/pull/7437.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/__tests__/versions/secrets/bulk.test.ts
+++ b/packages/wrangler/src/__tests__/versions/secrets/bulk.test.ts
@@ -9,7 +9,7 @@ import { runWrangler } from "../../helpers/run-wrangler";
 import { mockPostVersion, mockSetupApiCalls } from "./utils";
 import type { Interface } from "node:readline";
 
-describe("versions secret put", () => {
+describe("versions secret bulk", () => {
 	const std = mockConsoleMethods();
 	runInTempDir();
 	mockAccountId();
@@ -72,6 +72,18 @@ describe("versions secret put", () => {
 			➡️  To deploy this version to production traffic use the command \\"wrangler versions deploy\\"."
 		`
 		);
+		expect(std.err).toMatchInlineSnapshot(`""`);
+	});
+
+	test("no wrangler configuration warnings shown", async () => {
+		await writeFile("secrets.json", JSON.stringify({ SECRET_1: "secret-1" }));
+		await writeFile("wrangler.json", JSON.stringify({ invalid_field: true }));
+		mockSetupApiCalls();
+		mockPostVersion();
+		await runWrangler(
+			`versions secret bulk secrets.json --name script-name --x-versions`
+		);
+		expect(std.warn).toMatchInlineSnapshot(`""`);
 		expect(std.err).toMatchInlineSnapshot(`""`);
 	});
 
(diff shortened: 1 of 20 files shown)


packages/wrangler/src/__tests__/versions/secrets/bulk.test.ts
```diff
@@ -75,6 +75,18 @@ describe("versions secret put", () => {
 		expect(std.err).toMatchInlineSnapshot(`""`);
 	});
 
+	test("no wrangler configuration warnings shown", async () => {
+		await writeFile("secrets.json", JSON.stringify({ SECRET_1: "secret-1" }));
+		await writeFile("wrangler.json", JSON.stringify({ invalid_field: true }));
```
penalosa, line 80: Nit, but could this use the writeWranglerConfig helper?

packages/wrangler/src/__tests__/versions/secrets/put.test.ts
```diff
@@ -54,41 +55,60 @@ describe("versions secret put", () => {
 		expect(std.err).toMatchInlineSnapshot(`""`);
 	});
 
-	// For some reason, this always hangs. Not sure why
```
petebacondarwin, line 57: This was broken because you cannot call `useMockStdin()` from within an `it()` clause since itself calls things like `beforeEach()`.

packages/wrangler/src/__tests__/versions/secrets/utils.ts
```diff
@@ -82,8 +82,12 @@ export function mockGetVersion(versionInfo?: VersionDetails) {
 function mockGetVersionContent() {
 	msw.use(
 		http.get(
-			`*/accounts/:accountId/workers/scripts/:scriptName/content/v2?version=ce15c78b-cc43-4f60-b5a9-15ce4f298c2a`,
```
petebacondarwin, line 85: This was causing MockServiceWorkers (MSW) to write warnings to the console about a redundant query in the URL matcher.

packages/wrangler/src/versions/utils/config.ts
```diff
@@ -1,13 +1,11 @@
...
-	},
->(args: Pick<T, "config" | "name">) {
+type Args = Parameters<typeof readConfig>[1] & { config?: string };
+type Options = Parameters<typeof readConfig>[2];
+
+export function getConfig(args: Args, options?: Options, entryPath?: string) {
 	const configPath =
-		args.config || (args.name && findWranglerConfig(path.dirname(args.name)));
-	return readConfig(configPath, args);
+		args.config || (entryPath && findWranglerConfig(path.dirname(entryPath)));
```
petebacondarwin, line 9: In the future I think we could somehow roll this into `readConfig()` itself and remove the need for this helper function at all.

packages/wrangler/src/__tests__/versions/secrets/bulk.test.ts
petebacondarwin: Not really because that requires that the input object matches the `RawConfig` type, which then requires an awkward cast.

PR #7455, opened 2024-12-05
feat: provide experimental flags to the defineCommand handler

emily-shen:
currently, in order to access experimental flags you have to wrap each command handler in an AsyncLocalStorage context. this adds wraps it around the defineCommand handler so eventually the flags will be available to all commands.
---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: I think not breaking existing tests that use experimental flags should be sufficient?
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal helper


andyjessop (approved):
I like this change. Maybe we can go further in the future and remove ALS entirely, as it adds an extra layer of complexity and it's usually unnecessary, making the execution flow harder to understand.

PR #7476, opened 2024-12-06
WC-2972: Allow Workers Assets to be mounted to a path

WalshyDev:
Fixes [WC-2972](https://jira.cfdata.org/browse/WC-2972).

Allow Workers Assets to be mounted to a wildcard path. We are still limiting unwildcarded paths like `example.com/test` for the minute though.

We will log a warning saying that we will attempt to serve assets along with the asset dir being matched. Also with a note saying unmatched will go to the Worker
image

Wording TBD here.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not covered there
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18657

### DIFF
--- a/.changeset/early-baboons-fly.md
+++ b/.changeset/early-baboons-fly.md
@@ -1,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+feat: allow routing to Workers with Assets on any HTTP route, not just the root. For example, `example.com/blog/*` can now be used to serve assets.
+These assets will be served as though the assets directly were mounted to the root.
+For example, if you have `assets = { directory = "./public/" }`, a route like `"example.com/blog/*"` and a file `./public/blog/logo.png`, this will be available at `example.com/blog/logo.png`. Assets outside of directories which match the configured HTTP routes can still be accessed with the [Assets binding](https://developers.cloudflare.com/workers/static-assets/binding/#binding) or with a [Service binding](https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/) to this Worker.
--- a/packages/workers-shared/asset-worker/tests/handler.test.ts
+++ b/packages/workers-shared/asset-worker/tests/handler.test.ts
@@ -1,4 +1,5 @@
 import { vi } from "vitest";
+import { applyConfigurationDefaults } from "../src/configuration";
 import { handleRequest } from "../src/handler";
 import type { AssetConfig } from "../../utils/types";
 
@@ -99,5 +100,54 @@ describe("[Asset Worker] `handleRequest`", () => {
 		);
 
 		expect(response.status).toBe(200);
+	});
+
+	it("cannot fetch assets outside of configured path", async () => {
+		const assets: Record<string, string> = {
+			"/blog/test.html": "aaaaaaaaaa",
+			"/blog/index.html": "bbbbbbbbbb",
+			"/index.html": "cccccccccc",
+			"/test.html": "dddddddddd",
+		};
+
+		// Attempt to path traverse down to the root /test within asset-server
+		let response = await handleRequest(
+			new Request("https://example.com/blog/../test"),
+			applyConfigurationDefaults({}),
+			async (pathname: string) => {
+				if (pathname.startsWith("/blog/")) {
+					// our route
+					return assets[pathname] ?? null;
+				} else {
+					return null;
+				}
+			},
+			async (_: string) => ({
+				readableStream: new ReadableStream(),
+				contentType: "text/html",
+			})
+		);
+
+		expect(response.status).toBe(404);
+
+		// Attempt to path traverse down to the root /test within asset-server
+		response = await handleRequest(
+			new Request("https://example.com/blog/%2E%2E/test"),
+			applyConfigurationDefaults({}),
+			async (pathname: string) => {
+				if (pathname.startsWith("/blog/")) {
+					// our route
+					return assets[pathname] ?? null;
+				} else {
+					return null;
+				}
+			},
+			async (_: string) => ({
+				readableStream: new ReadableStream(),
+				contentType: "text/html",
+			})
+		);
+
+		expect(response.status).toBe(404);
 	});
 });
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -1602,95 +1602,336 @@ Update them to point to this script instead?`,
 			});
 		});
 
-		it("should error on routes with paths if assets are present", async () => {
-			writeWranglerConfig({
-				routes: [
-					"simple.co.uk/path",
-					"simple.co.uk/path/*",
-					"simple.co.uk/",
-					"simple.co.uk/*",
-					"simple.co.uk",
-					{ pattern: "route.co.uk/path", zone_id: "asdfadsf" },
-					{ pattern: "route.co.uk/path/*", zone_id: "asdfadsf" },
-					{ pattern: "route.co.uk/*", zone_id: "asdfadsf" },
-					{ pattern: "route.co.uk/", zone_id: "asdfadsf" },
-					{ pattern: "route.co.uk", zone_id: "asdfadsf" },
-					{ pattern: "custom.co.uk/path", custom_domain: true },
-					{ pattern: "custom.co.uk/*", custom_domain: true },
-					{ pattern: "custom.co.uk", custom_domain: true },
-				],
+		describe("deploy asset routes", () => {
+			it("shouldn't error on routes with paths if there are no assets", async () => {
+				writeWranglerConfig({
+					routes: [
+						"simple.co.uk/path",
+						"simple.co.uk/path/*",
+						"simple.co.uk/",
+						"simple.co.uk/*",
+						"simple.co.uk",
+						{ pattern: "route.co.uk/path", zone_id: "asdfadsf" },
+						{ pattern: "route.co.uk/path/*", zone_id: "asdfadsf" },
+						{ pattern: "route.co.uk/*", zone_id: "asdfadsf" },
+						{ pattern: "route.co.uk/", zone_id: "asdfadsf" },
+						{ pattern: "route.co.uk", zone_id: "asdfadsf" },
+						{ pattern: "custom.co.uk/path", custom_domain: true },
+						{ pattern: "custom.co.uk/*", custom_domain: true },
+						{ pattern: "custom.co.uk", custom_domain: true },
+					],
+				});
+				writeWorkerSource();
+
+				await expect(runWrangler(`deploy ./index`)).rejects
+					.toThrowErrorMatchingInlineSnapshot(`
+					[Error: Invalid Routes:
+					custom.co.uk/path:
+					Paths are not allowed in Custom Domains
+
+					custom.co.uk/*:
+					Wildcard operators (*) are not allowed in Custom Domains
+					Paths are not allowed in Custom Domains]
+				`);
 			});
-			writeWorkerSource();
-			writeAssets([{ filePath: "asset.txt", content: "Content of file-1" }]);
 
-			await expect(runWrangler(`deploy --assets="assets"`)).rejects
-				.toThrowErrorMatchingInlineSnapshot(`
-				[Error: Invalid Routes:
-				simple.co.uk/path:
-				Workers which have static assets cannot be routed on a URL which has a path component. Update the route to replace /path with /*
+			it("should warn on mounted paths", async () => {
+				writeWranglerConfig({
+					routes: [
+						"simple.co.uk/path/*",
+						"simple.co.uk/*",
+						"*/*",
+						"*/blog/*",
+						{ pattern: "example.com/blog/*", zone_id: "asdfadsf" },
+						{ pattern: "example.com/*", zone_id: "asdfadsf" },
+						{ pattern: "example.com/abc/def/*", zone_id: "asdfadsf" },
+					],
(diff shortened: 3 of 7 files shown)


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -189,26 +191,20 @@ export const validateRoutes = (routes: Route[], hasAssets: boolean) => {
...
 			const components = pattern.split("/");
 
-			if (
-				// = ["route.com"]  bare domains are invalid as it would only match exactly that
-				components.length === 1 ||
-				// = ["route.com",""] as above
-				(components.length === 2 && components[1] === "")
-			) {
+			// If the route does not end with a wildcard, then error
+			// We may want to allow this in the future, but for now let's error
```
WalshyDev, line 199: Wondering what you think @GregBrimble @korinne - imo this is a totally valid use case but not sure we want to unleash it _just yet_.

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -220,6 +216,21 @@ export const validateRoutes = (routes: Route[], hasAssets: boolean) => {
 					.join(`\n\n`)
 		);
 	}
+
+	if (mountedAssetRoutes.length > 0 && assetsDir !== undefined) {
+		const relativeAssetsDir = path.relative(process.cwd(), assetsDir);
+
+		logger.once.warn(
+			`Warning: The following routes will attempt to serve Assets on a mounted path:\n${mountedAssetRoutes
```
WalshyDev, line 224: probably worth checking serve_directly here
also @korinne on wording

.changeset/early-baboons-fly.md
```diff
@@ -0,0 +1,6 @@
+---
+"wrangler": minor
+---
+
+feat: allow mounting Workers Assets on wildcard paths not just the root - e.g. `example.com/blog/*` can now serve assets.
+These assets will be served matching the assets dir, so `example.com/blog/*` will serve assets from the `blog` dir within your assets output folder.
```
petebacondarwin, line 6: We have gone with the "mask" approach, not the "mount" approach, following the naming conventions used in previous discussions.

To be clear, if you have two routes:

- `example.com/blog/*` and
- `example.com/admin/*`

(and assuming the assets directory is `public`) then the first route would only ever serve assets from the `public/blog` directory and the second route would only ever serve assets from the `public/admin` directory, "for the same Worker deployment".

WalshyDev: yup, i forgot what the naming was and mount was what made sense in my head. naming is fun - i'll leave to korinne on whatever we want to describe it as publicly

Cherry:
This is awesome to see, but "masking" is an odd way to describe this feature. "Masking" to me implies a sense of temporary-ness, or fake-ness, vs the industry standard of mounting that you see across lots of different software when you mount files/folders at a specific location, like inside containers.

Hopefully the naming is finalised, because "mount" makes way more senses than "mask" in my opinion.

packages/wrangler/src/deploy/deploy.ts
GregBrimble: I'd personally just remove this restriction entirely. Otherwise, we'll have to remember to follow-up if we ever do infix splats in routes or whatever.

packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -458,6 +454,34 @@ describe.sequential("wrangler dev", () => {
...
+			await runWranglerUntilConfig("dev --assets assets");
+
+			expect(std.warn).toMatchInlineSnapshot(`
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mWarning: The following routes will attempt to serve Assets on a mounted path:[0m
+
+				    • simple.co.uk/path/* (Will match assets: assets/path/*)
+				    • example.com/blog/* (Will match assets: assets/blog/*)
+				    • example.com/abc/def/* (Will match assets: assets/abc/def/*)
+
+				  Requests not matching an asset will be forwarded to the Worker.
```
GregBrimble, line 480:
```suggestion
				  Requests not matching an asset will be forwarded to the Worker script.
```

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -1689,6 +1683,81 @@ Update them to point to this script instead?`,
...
+			await runWrangler(`deploy --assets assets`);
+
+			expect(std.warn).toMatchInlineSnapshot(`
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mWarning: The following routes will attempt to serve Assets on a mounted path:[0m
+
+				    • simple.co.uk/path/* (Will match assets: assets/path/*)
+				    • example.com/blog/* (Will match assets: assets/blog/*)
+				    • example.com/abc/def/* (Will match assets: assets/abc/def/*)
+
+				  Requests not matching an asset will be forwarded to the Worker.
```
GregBrimble, line 1743:
```suggestion
				  Requests not matching an asset will be forwarded to the Worker script.
```

.changeset/early-baboons-fly.md
GregBrimble, line 6:
```suggestion
feat: allow routing to Workers with Assets on any HTTP route, not just the root. For example, `example.com/blog/*` can now be used to serve assets.

These assets will be served as though the assets directly were mounted to the root. If the `asset.directory = "./public/", and `./public/blog/logo.png` will be available at `example.com/blog/logo.png`. Assets outside of directories which match the configured HTTP routes can still be accessed with the [Assets binding](https://developers.cloudflare.com/workers/static-assets/binding/#binding) or with a [Service binding](https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/) to this Worker.
```

packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -458,6 +454,34 @@ describe.sequential("wrangler dev", () => {
 				Paths are not allowed in Custom Domains]
 			`);
 		});
+
+		it("should warn on mounted paths in dev", async () => {
+			writeWranglerConfig({
+				routes: [
```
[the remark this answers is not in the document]
WalshyDev: yes good callout!

packages/wrangler/src/__tests__/dev.test.ts
WalshyDev: Do we want to say script? We moved away from the script terminology a while ago... though assets are part of the worker so also understandable

GregBrimble: We're not perfectly consistent, but yes, I'm trying to not make it seem like assets are "outside" of the Worker. We've used "script" in the docs for this a few times: https://developers.cloudflare.com/workers/static-assets/binding/.

We originally turned away from "script" because "service" replaced it. That's not really the same issue here, but yeah, totally get your concern. If you've got a better way of phrasing it, I'm all ears!

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -220,6 +216,24 @@ export const validateRoutes = (routes: Route[], hasAssets: boolean) => {
...
+				.map((route) => {
+					const routeNoScheme = route.replace(/https?:\/\//g, "");
+					const assetPath = path.join(
+						relativeAssetsDir,
+						routeNoScheme.substring(routeNoScheme.indexOf("/"))
+					);
+					return `  • ${route} (Will match assets: ${assetPath})`;
+				})
+				.join("\n")}` +
+				"\n\nRequests not matching an asset will be forwarded to the Worker."
```
penalosa, line 234: Are we ignoring the not found behaviour config setting here?

packages/wrangler/src/__tests__/deploy.test.ts
penalosa: I'd push back on script—we've been trying to move away from this language and towards Worker. To me, this is naturally a Worker + Assets situation, which makes the Worker language make sense for your code. Script is limiting, and makes it sound like workers are small scripts rather than complex fully featured applications with multiple modules & bindings

GregBrimble: https://github.com/cloudflare/workers-sdk/pull/7476/files#r1873572047

Yeah, I'm not super married to it. But lay out my thinking ^

packages/wrangler/src/deploy/deploy.ts
WalshyDev: Not matching assets where there is a Worker always go to the Worker (which can then handle in env.ASSETS)
I'll need to add something for no-Worker, need to check how we even detect that today

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -220,6 +213,26 @@ export const validateRoutes = (routes: Route[], hasAssets: boolean) => {
...
+					const routeNoScheme = route.replace(/https?:\/\//g, "");
+					const assetPath = path.join(
+						relativeAssetsDir,
+						routeNoScheme.substring(routeNoScheme.indexOf("/"))
+					);
+					return `  • ${route} (Will match assets: ${assetPath})`;
+				})
+				.join("\n")}` +
+				(assets?.routingConfig.has_user_worker
+					? "\n\nRequests not matching an asset will be forwarded to the Worker script."
```
GregBrimble, line 232: We've been given ["your Worker's code" from Brendan](https://github.com/cloudflare/cloudflare-docs/pull/18508#discussion_r1878520026). Let's go with that. I'll update the rest of the docs later this week.

```suggestion
					? "\n\nRequests not matching an asset will be forwarded to your Worker's code."
```

.changeset/early-baboons-fly.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+feat: allow routing to Workers with Assets on any HTTP route, not just the root. For example, `example.com/blog/*` can now be used to serve assets.
+These assets will be served as though the assets directly were mounted to the root.
+For example, if you have `asset.directory = "./public/"` then `./public/blog/logo.png` will be available at `example.com/blog/logo.png`. Assets outside of directories which match the configured HTTP routes can still be accessed with the [Assets binding](https://developers.cloudflare.com/workers/static-assets/binding/#binding) or with a [Service binding](https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/) to this Worker.
```
GregBrimble, line 7:
```suggestion
For example, if you have `assets = { directory = "./public/" }`, a route like `"example.com/blog/*"` and a file `./public/blog/logo.png`, this will be available at `example.com/blog/logo.png`. Assets outside of directories which match the configured HTTP routes can still be accessed with the [Assets binding](https://developers.cloudflare.com/workers/static-assets/binding/#binding) or with a [Service binding](https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/) to this Worker.
```

packages/workers-shared/asset-worker/tests/handler.test.ts
GregBrimble, line 1: thanks :)

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -1600,95 +1600,270 @@ Update them to point to this script instead?`,
...
+					  example.com/blog/* (zone id: asdfadsf)
+					  example.com/* (zone id: asdfadsf)
+					  example.com/abc/def/* (zone id: asdfadsf)
+					Current Version ID: Galaxy-Class"
+				`);
+			});
 
-				route.co.uk/path/*:
-				Workers which have static assets cannot be routed on a URL which has a path component. Update the route to replace /path/* with /*
+			it("assets only will not mention invoking Worker on 404", async () => {
```
GregBrimble, line 1722: What do you mean by this test title? I'm not picking up its significance.

GregBrimble: Ah, is this [the `has_user_worker` piece](https://github.com/cloudflare/workers-sdk/pull/7476/files#diff-31fd97cfffe59015a2d5a5cb87902e8e8c016fddd9151455b80129c82567b6a0R231)? I don't see anything checking when that message actually does get added?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -220,6 +213,26 @@ export const validateRoutes = (routes: Route[], hasAssets: boolean) => {
 					.join(`\n\n`)
 		);
 	}
+
+	if (mountedAssetRoutes.length > 0 && assets?.directory !== undefined) {
```
GregBrimble, line 217: tiny nit: you've already checked that `assets.directory` is defined.
```suggestion
	if (mountedAssetRoutes.length > 0) {
```

GregBrimble (approved):

packages/wrangler/src/deploy/deploy.ts
WalshyDev: I do but ts wasn't happy with it 😅

GregBrimble: Oh fair enough!

packages/wrangler/src/__tests__/deploy.test.ts
WalshyDev: Yep, apparently I missed that, ty. Added now

CarmenPopoviciu (approved):

PR #7477, opened 2024-12-06
Remove the need for `getConfig` by centralising logic surrounding finding the config file for given args

andyjessop:
Fixes #000.

There was a lot of logic spread around relating to how to locate the config path. Whether to use `args.config`, `args.script`, some specified `configPath`, etc. If we centralise this logic inside `readConfig`, we can remove a lot of the complexity.

In fact, it removes entirely the need for the first argument to `readConfig`, so in most cases, the call now just reads `readConfig(args)`.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: already covered
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -97,19 +97,19 @@ export function handleFailure<
 		? K
 		: never,
 >(
-	cb: (t: CommandArgumentsObject, config: Config) => Promise<void>
+	cb: (args: CommandArgumentsObject, config: Config) => Promise<void>
```
andyjessop, line 100: Some housekeeping 😁

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -0,0 +1,36 @@
+import path from "path";
+import { findUpSync } from "find-up";
+
+/**
+ * Resolve the path to the configuration file, given the `config` and `script` optional command line arguments.
+ * `config` takes precedence, then `script`, then we just use the cwd.
+ */
+export function resolveWranglerConfigPath({
```
andyjessop, line 8: This function is the meat of the PR, centralising the logic surrounding `config`, `script`, and the default option.

packages/wrangler/src/index.ts
```diff
@@ -1109,7 +1109,7 @@ export async function main(argv: string[]): Promise<void> {
 		// key to fetch) or flags
 
 		try {
-			const configPath = args.config ?? findWranglerConfig(process.cwd());
+			const configPath = resolveWranglerConfigPath(args);
 			const rawConfig = readRawConfig(args.config);
```
penalosa, line 1113: This logic doesn't look right (not a change in your PR). I _think_ we should be passing `configPath` to `readRawConfig`? `args.config` won't always be defined

packages/wrangler/src/triggers/index.ts
```diff
@@ -57,7 +57,7 @@ async function triggersDeployHandler(
 ) {
 	await printWranglerBanner();
 
-	const config = readConfig(undefined, args);
+	const config = readConfig(args);
```
penalosa, line 60: This is a change in behaviour, but it looks like the previous behaviour must've been a bug

andyjessop: Oh yeah, you're right. Didn't spot that.

PR #7478, opened 2024-12-06
fix: ensure that non-inherited fields are not removed when using an inferred named environment

petebacondarwin:
Fixes #0000

It is an error for the the user to provide an environment name that doesn't match any of the named environments in the Wrangler configuration. But if there are no named environments defined at all in the Wrangler configuration, we special case the top-level environment as though it was a named environment. Previously, when this happens, we would remove all the nonInheritable fields from the configuration (essentially all the bindings) leaving an incorrect configuration. Now we correctly generate a flattened named environment that has the nonInheritable fields, plus correctly applies any transformFn on inheritable fields.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/__tests__/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/configuration.test.ts
@@ -3854,22 +3854,35 @@ describe("normalizeAndValidateConfig()", () => {
 
 	describe("named environments", () => {
 		it("should warn if we specify an environment but there are no named environments", () => {
-			const rawConfig: RawConfig = {};
-			const { diagnostics } = normalizeAndValidateConfig(rawConfig, undefined, {
-				env: "DEV",
-			});
+			const rawConfig: RawConfig = {
+				name: "my-worker",
+				kv_namespaces: [{ binding: "KV", id: "xxxx-xxxx-xxxx-xxxx" }],
+			};
+			const { diagnostics, config } = normalizeAndValidateConfig(
+				rawConfig,
+				undefined,
+				{
+					env: "dev",
+				}
+			);
+			expect(config).toEqual(
+				expect.objectContaining({
+					name: "my-worker-dev",
+					kv_namespaces: [{ binding: "KV", id: "xxxx-xxxx-xxxx-xxxx" }],
+				})
+			);
 			expect(diagnostics.renderErrors()).toMatchInlineSnapshot(`
 			        "Processing wrangler configuration:
 			        "
 		      `);
 			expect(diagnostics.renderWarnings()).toMatchInlineSnapshot(`
 				"Processing wrangler configuration:
-				  - No environment found in configuration with name \\"DEV\\".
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/__tests__/configuration.test.ts
```diff
@@ -4010,70 +4023,70 @@ describe("normalizeAndValidateConfig()", () => {
...
 			});
-		});
 
-		it("should error if named environment contains a `account_id` field, even if there is no top-level name", () => {
-			const rawConfig: RawConfig = {
-				legacy_env: false,
-				env: {
-					DEV: {
-						account_id: "some_account_id",
+			it("should error if named environment contains a `account_id` field, even if there is no top-level name", () => {
```
petebacondarwin, line 4027: This whole section has just been moved inside the "non-legacy" describe block as that is what they are related to. There is no code change here.

packages/wrangler/src/config/validation-helpers.ts
```diff
@@ -69,7 +69,11 @@ export function inheritable<K extends keyof Environment>(
 ): Environment[K] {
 	validate(diagnostics, field, rawEnv[field], topLevelEnv);
 	return (
-		(rawEnv[field] as Environment[K]) ??
+		// `rawEnv === topLevelEnv` is a special case where the user has provided an environment name
+		// but that named environment is not actually defined in the configuration.
+		// In that case we have reused the topLevelEnv as the rawEnv,
+		// and so we need to process the `transformFn()` anyway rather than just using the field in the `rawEnv`.
+		(rawEnv !== topLevelEnv ? (rawEnv[field] as Environment[K]) : undefined) ??
```
jamesopstad, line 76: Trying to understand the need for the extra logic here. What would happen if it was left as it was?

packages/wrangler/src/config/validation.ts
```diff
@@ -218,7 +218,7 @@ export function normalizeAndValidateConfig(
 			activeEnv = normalizeAndValidateEnvironment(
```
penalosa, line 218: Given they're now the same, we should move the `activeEnv` calculation outside of the Pages/Workers gate

packages/wrangler/src/config/validation-helpers.ts
petebacondarwin: Without this the inherited field transform does not run. For example the `name` field gets transformed when it is inherited from 'my-worker' to 'my-worker-env'.

jamesopstad: Got it. Thanks!

jamesopstad (approved):

penalosa (approved):

PR #7490, opened 2024-12-09
Update bug-template.yaml

penalosa:


CarmenPopoviciu:
@penalosa just for posterity, can you pls provide some context as to why we're making these changes 🙏 ?

.github/ISSUE_TEMPLATE/bug-template.yaml
```diff
@@ -34,50 +34,29 @@ body:
...
     attributes:
       label: Please provide a link to a minimal reproduction
-      description: Although not required, we often request a minimal reproduction to help troubleshoot the issue, so providing this up-front streamlines the process towards resolution.
       placeholder: https://github.com/foobar-user/minimal-repro
     validations:
       required: false
+  - type: textarea
+    attributes:
+      label: Describe the Bug
+      description: Please describe the bug in as much detail as possible, including steps to reproduce, the behaviour you're seeing, and the behaviour you expect.
```
CarmenPopoviciu, line 56: IMHO `A git repo we can clone and run a test suite on, or which has a README with step-by-step instructions, is even better. In this case, please use the field below to provide a link to the minimal repro.` (or a shortened version of that) would be worth leaving in, because gh repo repros are super helpful, and would be good to keep encouraging folks to provide them when possible

penalosa: I've left that field in, it's just been moved above the description field

CarmenPopoviciu (approved):

PR #7494, opened 2024-12-09
chore: remove unnecessary svelteKit C3 variants

dario-piotrowicz:
The C3 sveltekit template has two variants, one for `js` and one for `ts`, but both variants (unless I am missing something) contain the exact same files, so I don't think that the two variants are necessary

This PR doesn't change any behavior at all simply removes extra unnecessary files

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included - there are already C3 svelteKit e2e tests
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


CarmenPopoviciu (approved):

CarmenPopoviciu:
@dario-piotrowicz feel free to merge when you're ready 🚀

PR #7506, opened 2024-12-10
feat: notify users when they are running C3 in experimental mode

dario-piotrowicz:
This is just to fix that usually bothers me, when I run C3 in experimental mode there is no indication wether it is running or not in such mode, having this confirmation seems quite nice to me since, besides the final output, there is no way for me to double check/confirm that I didn't I mess up (like a simple typo etc..) inputting the `--experimental` flag

Result:
Screenshot 2024-12-10 at 11 59 06

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required (just to make sure that this is not breaking some e2es that check the welcome banner or something like that)
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this simply adds some extra info in the C3 banner

### DIFF
--- a/packages/create-cloudflare/src/__tests__/dialog.test.ts
+++ b/packages/create-cloudflare/src/__tests__/dialog.test.ts
@@ -15,33 +15,69 @@ describe("dialog helpers", () => {
 		process.stdout.columns = originalColumns;
 	});
 
-	test("printWelcomeMessage with telemetry disabled", () => {
-		printWelcomeMessage("0.0.0", false);
+	describe("printWelcomeMessage", () => {
+		test("with telemetry disabled", () => {
+			printWelcomeMessage("0.0.0", false, {});
 
-		expect(normalizeOutput(std.out)).toMatchInlineSnapshot(`
-			"────────────────────────────────────────────────────────────
-			👋 Welcome to create-cloudflare v0.0.0!
-			🧡 Let's get started.
-			────────────────────────────────────────────────────────────
+			expect(normalizeOutput(std.out)).toMatchInlineSnapshot(`
+				"────────────────────────────────────────────────────────────
+				👋 Welcome to create-cloudflare v0.0.0!
+				🧡 Let's get started.
+				────────────────────────────────────────────────────────────
 
-			"
-		`);
-	});
+				"
+			`);
+		});
 
-	test("printWelcomeMessage with telemetry enabled", () => {
-		printWelcomeMessage("0.0.0", true);
+		test("with telemetry enabled", () => {
+			printWelcomeMessage("0.0.0", true, {});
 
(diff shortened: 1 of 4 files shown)


packages/create-cloudflare/src/dialog.ts
```diff
@@ -27,18 +27,27 @@ export function createDialog(lines: string[]) {
...
 		`👋 Welcome to create-cloudflare v${version}!`,
 		`🧡 Let's get started.`,
 	];
 
+	if (args.experimental) {
+		lines.push("", blue`🧪 Running in experimental mode`);
+	}
+
 	if (telemetryEnabled) {
+		if (args.experimental) {
```
CarmenPopoviciu, line 42: this seems unnecessary 🤔  ... why is it needed?

dario-piotrowicz: it's just so that if you are in experimental mode and telemetry is enabled, there is a clean clean newline between the two, as you can see in the test snapshot:
Screenshot 2024-12-10 at 15 30 58

without this the telemetry first line would be right below the external one, which, given the fact that the telemetry into already has an empty line itself, would make things pretty unclear in my opinion

CarmenPopoviciu: oh, gotcha! thank you for pointing that out 🙏

CarmenPopoviciu: one last question....in that case, wouldn't it make more sense to 👇 ? 
```
if (args.experimental) {
		lines.push("", blue`🧪 Running in experimental mode\n`);
	}
```

CarmenPopoviciu: right right...no let's just leave it as is. it'll be alright

CarmenPopoviciu (approved):

packages/create-cloudflare/src/__tests__/dialog.test.ts
```diff
@@ -15,33 +15,69 @@ describe("dialog helpers", () => {
...
+		});
+
+		test("with telemetry enabled", () => {
+			printWelcomeMessage("0.0.0", true, {});
+
+			expect(normalizeOutput(std.out)).toMatchInlineSnapshot(`
+				"────────────────────────────────────────────────────────────
+				👋 Welcome to create-cloudflare v0.0.0!
+				🧡 Let's get started.
+				📊 Cloudflare collects telemetry about your usage of Create-Cloudflare.
```
petebacondarwin, line 39: Not a big deal but just chiming in here after the fact 😄 

In the case that it is not experimental mode there is no empty line between "let's get started" and the telemetry line. So adding extra lines when in experimental mode is a change of look.

PR #7509, opened 2024-12-10
Add column name to Insert statement (#7507)

### LINKED ISSUE #7507: 🐛 BUG: Exported SQL Fails to Import Due to Missing Column Names in INSERT Statements

D1

3.94.0[wrangler]

20.16.0

macOS 15.1.1

### Observed behavior
I encountered issues when trying to export and re-import data during iterative development. The INSERT statements generated during the export process did not include column names. This caused problems because the column creation order in the tables during development differed from the order in the finalized table definitions.

As a result, the exported SQL file failed to correctly import the data into the updated database schema.

### Expected behavior
The exported SQL file should include column names in the INSERT statements to ensure compatibility when importing data, even if the column order differs between the development and finalized table definitions.

### Steps to reproduce

Use drizzle-orm with the following migration files:

```
ls -al migrations/
meta
0000_init.sql
0001_~
0002_~
・・・
```

Export data from the database:

```
for table in $(echo "artists" "artist_timelines" "albums" "album_timelines" "tracks" )
do
  npx wrangler d1 export spotify-db --table=$table --output ./data/$table.sql --no-schema --local
done
```

Delete the database and recreate it:

```
rm -rf migrations/*
npx drizzle-kit generate --name=init
npx wrangler d1 migrations apply spotify-db --local
```

Attempt to re-import the exported data:

```
for table in $(echo "artists" "artist_timelines" "albums" "album_timelines" "tracks" )
do
  npx wrangler d1 execute spotify-db --file=./data/$table.sql --local
done
```

During development, I frequently modify the table structure by adding and removing columns using ALTER statements. Once the table structure is finalized, I consolidate these changes into a single CREATE TABLE statement.

To preserve the data generated during application testing, I exported the data and attempted to re-import it into the finalized schema. However, because the exported SQL file lacked column names in the INSERT statements and the column order differed, the import failed.

https://github.com/cloudflare/workers-sdk/blob/75be0d54abdd3593b98ac4064f1ca50bde61f7a6/packages/miniflare/src/workers/d1/dumpSql.ts#L82

_No response_

nmemoto:
Fixes #7507.

Add column name to Insert statement for D1 export

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/miniflare/src/workers/d1/dumpSql.ts
```diff
@@ -79,7 +79,7 @@ export function* dumpSql(
 				}
 			});
 
-			yield `INSERT INTO ${sqliteQuote(table)} VALUES(${formattedCells.join(",")});`;
+			yield `INSERT INTO ${sqliteQuote(table)}(${column_names}) VALUES(${formattedCells.join(",")});`;
```
lambrospetrou, line 82: There are a few reasons why this change cannot be accepted as is.

1. This file only applies for local development, and not for what runs in production, so if we would do such change we would need to do it both for miniflare, and also for our internal version that actually runs the worker code in production otherwise it will be inconsistent.
2. Most importantly, this is a breaking change. D1 has a limit on the statement length for imports at 100KB. Adding column names to these `INSERT` commands means that some databases that used to work (export and then import), might not work anymore if they have many columns and this will cause their `INSERT` statements to go above the limit.

Thank you for the PR but we will need to think more about the issue you originally raised for the inconsistent column names and fix it in a different non-breaking way, e.g. sorting deterministically the column names in our exported statements.

nmemoto: Thank you for reviewing the pull request. I understand the reasons why it cannot be merged. I hope this will lead to better discussions and considerations.

PR #7520, opened 2024-12-11
Workflows: fix getting a workflow when such workflow was not yet created

bruxodasilva:
Fixes WOR-440.

Fixed a bug in local development where fetching a Workflow instance by ID would return a Workflow status, even if that instance did not exist. This only impacted the `get()` method on the Worker bindings. 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s):

### DIFF
--- a/.changeset/lovely-fireants-roll.md
+++ b/.changeset/lovely-fireants-roll.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workflows-shared": patch
+---
+
+Fix getting a workflow when such workflow was not yet created
--- a/.changeset/lovely-fireants-roll.md
+++ b/.changeset/lovely-fireants-roll.md
@@ -2,4 +2,4 @@
 "@cloudflare/workflows-shared": patch
 ---
 
-Fixed a bug in local development where fetching a Workflow instance by ID would return a Workflow status, even if that instance did not exist. This only impacted the `get()` method on the Worker bindings. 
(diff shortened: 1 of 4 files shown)


.changeset/lovely-fireants-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/workflows-shared": patch
+---
+
+Fix getting a workflow when such workflow was not yet created
```
elithrar, line 5:
```suggestion
Fixed a bug in local development where fetching a Workflow instance by ID would return a Workflow status, even if that instance did not exist. This only impacted the `get()` method on the Worker bindings. 
```

Skye-31 (approved):

packages/workflows-shared/src/engine.ts
```diff
@@ -142,6 +142,11 @@ export class Engine extends DurableObject<Env> {
 		_accountId: number,
 		_instanceId: string
 	): Promise<InstanceStatus> {
+		// NOTE(dferreira): we need to verify if the engine was initialized or not
```
CarmenPopoviciu, line 145: are you tracking this anywhere? if yes, we can remove the comment 🙏

bruxodasilva: It was more of an explanation of why it was necessary. Removing it now.

CarmenPopoviciu (approved):

PR #7521, opened 2024-12-11
feat: add experimental_patchConfig

emily-shen:
Fixes DEVX 1424

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: should be covered by unit tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not user facing change

### DIFF
--- a/packages/wrangler/src/__tests__/patch-config.test.ts
+++ b/packages/wrangler/src/__tests__/patch-config.test.ts
@@ -8,7 +8,8 @@ import type { RawConfig } from "../config";
 type TestCase = {
 	name: string;
 	original: RawConfig;
-	patch: RawConfig;
+	additivePatch: RawConfig;
+	replacingPatch: RawConfig;
 	expectedToml: string;
 	expectedJson: string;
 };
@@ -16,7 +17,14 @@ const testCases: TestCase[] = [
 	{
 		name: "add a binding",
 		original: {},
-		patch: {
+		additivePatch: {
+			kv_namespaces: [
+				{
+					binding: "KV",
+				},
+			],
+		},
+		replacingPatch: {
 			kv_namespaces: [
 				{
 					binding: "KV",
@@ -52,8 +60,18 @@ const testCases: TestCase[] = [
 				},
 			],
 		},
-		patch: {
+		additivePatch: {
 			kv_namespaces: [
+				{
+					binding: "KV2",
+				},
+			],
+		},
+		replacingPatch: {
+			kv_namespaces: [
+				{
+					binding: "KV",
+				},
 				{
 					binding: "KV2",
 				},
@@ -94,7 +112,14 @@ const testCases: TestCase[] = [
 				},
 			],
 		},
-		patch: {
+		additivePatch: {
+			d1_databases: [
+				{
+					binding: "DB",
+				},
+			],
+		},
+		replacingPatch: {
 			d1_databases: [
 				{
 					binding: "DB",
@@ -138,7 +163,10 @@ const testCases: TestCase[] = [
 				},
 			],
 		},
-		patch: {
+		additivePatch: {
+			compatibility_flags: ["nodejs_compat"],
+		},
+		replacingPatch: {
 			compatibility_flags: ["nodejs_compat"],
 		},
 		expectedToml: dedent`
@@ -179,7 +207,7 @@ const testCases: TestCase[] = [
 				},
 			],
 		},
-		patch: {
+		additivePatch: {
 			kv_namespaces: [
 				{
 					binding: "KV2",
@@ -189,6 +217,27 @@ const testCases: TestCase[] = [
 				},
 			],
 			d1_databases: [
+				{
+					binding: "DB2",
+				},
+			],
+		},
+		replacingPatch: {
+			kv_namespaces: [
+				{
+					binding: "KV",
+				},
+				{
+					binding: "KV2",
+				},
+				{
+					binding: "KV3",
+				},
+			],
+			d1_databases: [
+				{
+					binding: "DB",
+				},
 				{
 					binding: "DB2",
 				},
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/config/patch-config.ts
```diff
@@ -0,0 +1,63 @@
...
+	if (!configPath) {
+		return;
+	}
+
+	let configString = readFileSync(configPath);
+
+	if (configPath.endsWith("toml")) {
+		// the TOML parser we use does not preserve comments
+		if (configString.includes("#")) {
+			return;
```
edmundhung, line 20: We need to know if the patch does not happens so we can show a different message to the user?

packages/wrangler/src/config/patch-config.ts
```diff
@@ -0,0 +1,63 @@
+import TOML from "@iarna/toml";
+import { applyEdits, format, parse as JSONCParse, modify } from "jsonc-parser";
+import { parseTOML, readFileSync } from "../parse";
+import type { RawConfig } from "./config";
+import type { JSONPath } from "jsonc-parser";
+
+export const experimental_patchConfig = (
+	configPath: string | undefined,
```
edmundhung, line 8: Why `configPath` could be `undefined`?

packages/wrangler/src/config/patch-config.ts
edmundhung, line 20: We might need to know if the patch does not happens so we can show a different message to the user?

packages/wrangler/src/config/patch-config.ts
```diff
@@ -0,0 +1,63 @@
...
+		});
+		configString = applyEdits(configString, edit);
+		const formatEdit = format(configString, undefined, {});
+		configString = applyEdits(configString, formatEdit);
+	}
+
+	if (configPath.endsWith(".toml")) {
+		return TOML.stringify(JSONCParse(configString));
+	}
+	return configString;
```
edmundhung, line 43: Nit: If we are taking a file path on this function, why don't we write the result back to the file?

packages/wrangler/src/config/patch-config.ts
penalosa: This path should throw, so that the caller can tell the patch failed

packages/wrangler/src/config/patch-config.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const getPath = (
```
penalosa, line 46: Is the `jsonc-parser` helper `getNodePath()` usable here at all?

penalosa, line 46: Could you add a comment to this function explaining what it does?

packages/wrangler/src/__tests__/patch-config.test.ts
```diff
@@ -0,0 +1,248 @@
...
+import type { RawConfig } from "../config";
+
+type TestCase = {
+	name: string;
+	original: RawConfig;
+	patch: RawConfig;
+	expectedToml: string;
+	expectedJson: string;
+};
+const testCases: TestCase[] = [
```
penalosa, line 15: Could you add some tests for things like adding a D1 binding when only KV bindings exist, and adding a new field like compat date?

packages/wrangler/src/config/patch-config.ts
```diff
@@ -0,0 +1,63 @@
...
+		const edit = modify(configString, patchPath, value, {
+			isArrayInsertion: true,
+		});
+		configString = applyEdits(configString, edit);
+		const formatEdit = format(configString, undefined, {});
+		configString = applyEdits(configString, formatEdit);
+	}
+
+	if (configPath.endsWith(".toml")) {
+		return TOML.stringify(JSONCParse(configString));
```
penalosa, line 41: This should use the `parseJsonc` helper

packages/wrangler/src/config/patch-config.ts
```diff
@@ -0,0 +1,63 @@
...
+
+	const patchPaths: JSONPath[] = [];
+	getPath(patch, patchPaths);
+	for (const patchPath of patchPaths) {
+		const value = patchPath.pop();
+		const edit = modify(configString, patchPath, value, {
+			isArrayInsertion: true,
+		});
+		configString = applyEdits(configString, edit);
+		const formatEdit = format(configString, undefined, {});
```
penalosa, line 36: Do we need to format in the loop? Would one formatting pass at the end work?

packages/wrangler/src/config/patch-config.ts
emily-shen: readRawConfig, resolveWranglerConfigPath all return a configPath as `string | undefined` for when there isn't one, so I preferred to handle it inside experimental_patchConfig but I don't feel strongly about it.

@penalosa the spec has this function taking `configPath?: string, config: RawConfig` - is there any particular reason for configPath being optional?

packages/wrangler/src/__tests__/patch-config.test.ts
emily-shen: a9e4d48ec0a781a6f8defb8245f2a263b9a32c33

packages/wrangler/src/__tests__/patch-config.test.ts
```diff
@@ -0,0 +1,340 @@
...
+					]
+				}
+			`,
+	},
+];
+
+describe("experimental_patchConfig()", () => {
+	runInTempDir();
+	describe.each(["json", "toml"])("%s", (configType) => {
+		it.each(testCases)(
```
edmundhung, line 248: You might want to add one more test to check if the config file is not found

packages/wrangler/src/config/patch-config.ts
penalosa: The spec isn't super clear admittedly, but this should write the result back to the file

packages/wrangler/src/config/patch-config.ts
emily-shen: not particularly, because it takes a `Node`, which we would need to use their parseTree function to get, which is basically this with some extra steps like stringifying the patch as well. I'm not sure we would gain much.

packages/wrangler/src/__tests__/patch-config.test.ts
emily-shen: made the configPath required now instead

packages/wrangler/src/config/patch-config.ts
```diff
@@ -26,24 +25,26 @@ export const experimental_patchConfig = (
...
 	if (configPath.endsWith(".toml")) {
-		return TOML.stringify(JSONCParse(configString));
+		configString = TOML.stringify(parseJSONC(configString));
 	}
+	writeFileSync(configPath, configString);
 	return configString;
 };
 
-const getPath = (
+// gets all the json paths for the patch which are needed to create the edit
```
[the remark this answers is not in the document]
emily-shen: more comments [dd02c5d](https://github.com/cloudflare/workers-sdk/pull/7521/commits/dd02c5da29ed8ead304e792b19240483dc6d768e)

packages/wrangler/src/__tests__/patch-config.test.ts
```diff
@@ -0,0 +1,880 @@
...
+	// 				binding: "KV2",
+	// 			},
+	// 		],
+	// 	},
+	// 	replacingPatch: {
+	// 		kv_namespaces: [
+	// 			{
+	// 				binding: "KV",
+	// 			},
+	// 			undefined,
```
penalosa, line 408: Does this work if you take out the `undefined`?

packages/wrangler/src/__tests__/patch-config.test.ts
```diff
@@ -0,0 +1,880 @@
...
+					"name": "test-name",
+					"compatibility_flags": [
+						"nodejs_compat",
+						"flag"
+					]
+				}
+			`,
+	},
+	{
+		name: "delete a compat flag",
```
penalosa, line 477: There don't seem to be any compat flags in the original here?

packages/wrangler/src/__tests__/patch-config.test.ts
emily-shen: no :/ I probably need to rethink the 'delete' api, but since no one is asking for it now and all of this is quite separate from the patchConfig flow that is actually used, I was just going to leave it alone for now. 

Happy to take the edit with replacement stuff out of this PR as well, and just have the additive patch if you'd rather.

packages/wrangler/src/__tests__/patch-config.test.ts
emily-shen: argh excellent point, got my tests mixed up [6c2ce25](https://github.com/cloudflare/workers-sdk/pull/7521/commits/6c2ce25fd009b90bb124783d827bedb069d12fe9)

packages/wrangler/src/__tests__/patch-config.test.ts
emily-shen: for future reference: 

the edit-with-replacement is going through each terminal? leaf? node and updating that, in order to preserve comments. that's why it needs 'undefined' in order to actually delete something. I suppose the better way to do it would be work out the diff between the patch and original, and only edit those. Will iterate on this later 👍

PR #7522, opened 2024-12-11
feat: add support for custom unenv resolve path

vicb:
We are moving the unenv preset to this repo.

We will need to be able to test the version that is package with wrangler as well as the dev version. We can do that by passing the path to the dev version in the added env var.

/cc @pi0

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [X] Tests not necessary because: tested locally
  - [ ] I don't know
  - [X] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: internal use


packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -90,14 +98,16 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
 	alias: Record<string, string>,
-	external: string[]
+	external: string[],
+	unenvPresetPath: string | undefined
 ) {
 	// esbuild expects alias paths to be absolute
 	const aliasAbsolute: Record<string, string> = {};
 	for (const [module, unresolvedAlias] of Object.entries(alias)) {
 		try {
+			const paths = unenvPresetPath ? [unenvPresetPath] : undefined;
```
petebacondarwin, line 108: Can we just call 

```
const unenvPresetPath = getUnenvPresetPathFromEnv();
```

here and avoid having to thread it through the code?

petebacondarwin (changes requested):

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
pi0: (quick call with Vic, we might eventually move the plugin to unenv core as part of https://github.com/unjs/unenv/issues/263 -- this unifies impl between nitro and wrangler as a bonus)

petebacondarwin: Not sure I follow here.
If we move this to the preset (or even to unenv core) then this `unenvPresetPath` would be moot, right? So we'd need to redo this code then anyway.
In that case just loading the env var where it is needed is simplest and touches the least code.

pi0: So there are two things:

- To move forward add a feature to unenv core to allow full path resolves
- In the future, to move the plugin stuff as an unplugin

petebacondarwin: Does that affect my original comment to move the reading of the environment variable into this code?

vicb: > If we move this to the preset (or even to unenv core) then this unenvPresetPath would be moot, right?

It would only be a different path but same code.

> In that case just loading the env var where it is needed is simplest and touches the least code.
> Does that affect my original comment to move the reading of the environment variable into this code?

In general I find injecting values cleaner.
i.e. in this case if we have an hardcoded `WRANGLER_...` name, we would have to change that if we move the plugin. Injecting the value is more flexible (i.e. the first point above).

pi0:
LGTM! Small point: It is technically UNENV_RESOLVE_PATHS not specific to preset paths only 😄

.changeset/fluffy-eggs-film.md
```diff
@@ -0,0 +1,10 @@
+---
+"wrangler": minor
+---
+
+feat(wrangler): allow overriding the unenv preset paths
+
+By default wrangler uses the unenv preset installed via package.json.
+
+Setting `WRANGLER_UNENV_RESOLVE_PATHS` allow to use a local version of the preset.
```
petebacondarwin, line 9: I find this naming a bit confusing. I think that we are only using this path to resolve from imports like `unenv/foo/bar` to some absolute path, right? It doesn't affect other imports? Are we really just providing custom paths to the unenv package? Is there some other use for thes paths? If not, then I would expect this to be more like `WRANGLER_UNENV_PACKAGE_PATH`.

.changeset/fluffy-eggs-film.md
```diff
@@ -0,0 +1,10 @@
+---
+"wrangler": minor
+---
+
+feat(wrangler): allow overriding the unenv preset paths
+
+By default wrangler uses the unenv preset installed via package.json.
```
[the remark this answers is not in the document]
vicb: I update the description, PTAL.

petebacondarwin (approved):

PR #7523, opened 2024-12-11
cloudchamber/create: always print deployment and placement ID

gpanders:
We currently only print the full deployment ID when the deployment has an IPv4 address. This commit makes the deployment ID always print, as well as the placement ID. We also move the printing of the IPv4 address (if one exists) to the same place as the IPv6 address so that they are printed together.

There are a few other nit-picky changes as well. I've tried to make capitalization of labels more consistent (there doesn't appear to be a rule on when someone should be capitalized and when it shouldn't be, but I've at least made it more consistent within the subcommands themselves). I also changed the green "SUCCESS" message to only print once at the very end of the `cloudchamber create` command. This is consistent with the flow for creating a Worker.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: minor UI changes only
  - [ ] Required
  - [x] Not required because: minor UI changes only
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: minor UI changes only

### DIFF
--- a/packages/wrangler/src/cloudchamber/cli/deployments.ts
+++ b/packages/wrangler/src/cloudchamber/cli/deployments.ts
@@ -1,5 +1,5 @@
 import { exit } from "process";
-import { cancel, crash, endSection, log } from "@cloudflare/cli";
+import { cancel, crash, endSection, log, newline } from "@cloudflare/cli";
 import { processArgument } from "@cloudflare/cli/args";
 import { brandColor, dim, yellow } from "@cloudflare/cli/colors";
 import { spinner } from "@cloudflare/cli/interactive";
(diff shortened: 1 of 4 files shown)


gabivlj (approved):

skepticfx (approved):

CarmenPopoviciu (approved):

CarmenPopoviciu:
@gpanders while these are indeed just some UI adjustments, I would have expected some existing tests to capture the changes. Are there no tests at all that capture these codepaths?

gpanders:
> @gpanders while these are indeed just some UI adjustments, I would have expected some existing tests to capture the changes. Are there no tests at all that capture these codepaths?

We don't (currently) have any tests for the interactive flow for creating a deployment, which is what this PR changes. Our existing tests cover non-interactive creation and error cases, neither of which are affected by this PR.

PR #7540, opened 2024-12-13
fix: Fix changeset

CarmenPopoviciu:
Fixes #000

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: changeset fix
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: changeset fix
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: changeset fix

### DIFF
--- a/.changeset/clean-peas-exist.md
+++ b/.changeset/clean-peas-exist.md
@@ -1,5 +1,5 @@
 ---
-"wrangler": patch
+"wrangler": minor
 ---
 
 Add file prefix option to wrangler pipelines commands


petebacondarwin (approved):
Ideally I think we should have a bit more context in changesets than this but not blocking. For instance:
- is this every pipeline command? Perhaps list them if not all.
- Is the option `--prefix`?  Perhaps show an example
- what exactly does a prefix do to the command?

PR #7541, opened 2024-12-13
chore(wrangler): update unenv dependency version

vicb:
Update unenv to the latest release.

unenv now resolves aliases to absolute path so the plugin does not it to do it any more.

/cc @petebacondarwin @pi0

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: tested by existing tests
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #7566 
- <kbd>&nbsp;1&nbsp;</kbd> #7541 👈

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -10069,7 +10069,7 @@ export default{
 			fs.writeFileSync(
 				"index.js",
 				`
-      import path from 'path';
+      import path from 'node:path';
       console.log(path);
       export default {}
       `
@@ -10090,7 +10090,7 @@ export default{
 			}
 		`);
 			expect(fs.readFileSync("dist/index.js", { encoding: "utf-8" })).toContain(
-				`import path from "path";`
+				`import path from "node:path";`
 			);
 		});
 
--- a/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
+++ b/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
@@ -1,7 +1,7 @@
 import { builtinModules } from "node:module";
 import nodePath from "node:path";
 import dedent from "ts-dedent";
-import { cloudflare, env, nodeless } from "unenv";
+import { cloudflare, defineEnv } from "unenv";
 import { getBasePath } from "../../paths";
 import type { Plugin, PluginBuild } from "esbuild";
 
@@ -9,13 +9,25 @@ const REQUIRED_NODE_BUILT_IN_NAMESPACE = "node-built-in-modules";
 const REQUIRED_UNENV_ALIAS_NAMESPACE = "required-unenv-alias";
 
 export const nodejsHybridPlugin: () => Plugin = () => {
-	const { alias, inject, external } = env(nodeless, cloudflare);
+	// Get the resolved environment.
+	const { env } = defineEnv({
+		nodeCompat: true,
+		presets: [cloudflare],
+		resolve: true,
+	});
+	const { alias, inject, external } = env;
+	// Get the unresolved alias.
+	const unresolvedAlias = defineEnv({
+		nodeCompat: true,
+		presets: [cloudflare],
+		resolve: false,
+	}).env.alias;
 	return {
 		name: "hybrid-nodejs_compat",
 		setup(build) {
 			errorOnServiceWorkerFormat(build);
 			handleRequireCallsToNodeJSBuiltins(build);
-			handleUnenvAliasedPackages(build, alias, external);
+			handleUnenvAliasedPackages(build, unresolvedAlias, alias, external);
 			handleNodeJSGlobals(build, inject);
 		},
 	};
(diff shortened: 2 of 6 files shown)


packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -10090,7 +10090,7 @@ export default{
 			}
 		`);
 			expect(fs.readFileSync("dist/index.js", { encoding: "utf-8" })).toContain(
-				`import path from "path";`
+				`import path from "node:path";`
```
vicb, line 10093: This change is required as because unenv module resolution will always append the `node:` prefix

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -10069,7 +10069,7 @@ export default{
 			fs.writeFileSync(
 				"index.js",
 				`
-      import path from 'path';
+      import path from 'node:path';
```
vicb, line 10072: This change is not required - the output is the same with or without the prefix.
Added only for consistency with l10093 below

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -1,21 +1,33 @@
...
+		presets: [cloudflare],
+		resolve: true,
+	});
+	const { alias, inject, external } = env;
+	// Get the unresolved alias.
+	const unresolvedAlias = defineEnv({
+		nodeCompat: true,
+		presets: [cloudflare],
+		resolve: false,
+	}).env.alias;
```
pi0, line 24: Interesting would you need it for longer them you think? We could also expose `resolveEnv()` from unenv to avoid defining twice. (but also it is so to call it twice haha!)

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -87,45 +99,41 @@ function handleRequireCallsToNodeJSBuiltins(build: PluginBuild) {
...
 	build.onResolve({ filter: UNENV_ALIAS_RE }, (args) => {
-		const unresolvedAlias = alias[args.path];
+		const unresolved = unresolvedAlias[args.path];
 		// Convert `require()` calls for NPM packages to a virtual ES Module that can be imported avoiding the require calls.
 		// Note: Does not apply to Node.js packages that are handled in `handleRequireCallsToNodeJSBuiltins`
 		if (
 			args.kind === "require-call" &&
-			(unresolvedAlias.startsWith("unenv/runtime/npm/") ||
-				unresolvedAlias.startsWith("unenv/runtime/mock/"))
+			(unresolved.startsWith("unenv/runtime/npm/") ||
```
pi0, line 124: I recommend using `resolveAlias` from `pathe/utils` ([pathe](https://npmjs.com/pathe) is safe against windows paths when comparing and also handles multiple recursive aliases a => b => c)

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
vicb: Having a `resolveEnv()` or `env.unresolved` would make things slightly nicer here but I'm fine with calling it twice.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
vicb: Not sure to understand how it would help here?
`"unenv/runtime/npm/"` comes from the preset alias.

petebacondarwin:
> unenv now resolves aliases to absolute path so the plugin does not it to do it any more.

🎉

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -1,21 +1,33 @@
...
 
 const REQUIRED_NODE_BUILT_IN_NAMESPACE = "node-built-in-modules";
 const REQUIRED_UNENV_ALIAS_NAMESPACE = "required-unenv-alias";
 
 export const nodejsHybridPlugin: () => Plugin = () => {
-	const { alias, inject, external } = env(nodeless, cloudflare);
+	// Get the resolved environment.
+	const { env } = defineEnv({
+		nodeCompat: true,
+		presets: [cloudflare],
```
petebacondarwin, line 15: No need to include the `nodeless` anymore?

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
petebacondarwin: Happy to go with the current approach. But it seems to me, what would be nice is if `unenv` exposed something like `env.resolve(path)` function that can be used to get the absolute path of any import specifier.
Then there would be no need to hold two distinct alias objects.

petebacondarwin (approved):

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
vicb: Another good question.

We are iterating with Pooya to find the best way to do this. We might revisit this code as we progress.

PR #7560, opened 2024-12-16
Load Wrangler in extension

penalosa:
Fixes DEVX-1470

Load Wrangler from the workspace. A followup PR will fallback to a bundled version of Wrangler if needed.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: extension currently has no tests—coming soon
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: E2e doesn't cover the extension
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: extension not yet documented


packages/cloudflare-workers-bindings-extension/src/extension.ts
```diff
@@ -1,9 +1,24 @@
...
+		vscode.workspace.workspaceFolders &&
+		vscode.workspace.workspaceFolders.length > 0
+			? vscode.workspace.workspaceFolders[0].uri.fsPath
+			: undefined;
+
+	if (!rootPath) {
+		return;
+	}
+
+	const wrangler = importWrangler(rootPath);
```
edmundhung, line 21: Nit: Should we make `importWrangler` accepts `undefined` so it could fallback to the bundled wrangler as well?

edmundhung (approved):

packages/cloudflare-workers-bindings-extension/src/wrangler.ts
```diff
@@ -0,0 +1,21 @@
+import * as path from "path";
+
+export function importWrangler(
+	workspaceRoot: string
+): typeof import("wrangler") {
+	const wrangler = path.join(
+		path.join(
```
emily-shen, line 7: why do we need two path.joins?

packages/cloudflare-workers-bindings-extension/src/extension.ts
penalosa: I think the root path should always be defined (if it's not, there's no workspace loaded, and so there's nothing this extension can usefully do anyway)

packages/cloudflare-workers-bindings-extension/package.json
```diff
@@ -69,7 +69,8 @@
 		"eslint": "^8.57.0",
 		"npm-run-all": "^4.1.5",
 		"typescript": "^5.4.5",
-		"vsce": "^2.15.0"
+		"vsce": "^2.15.0",
+		"wrangler": "^3.95.0"
```
edmundhung, line 73: Can we use the workspace version instead? Just so I can use the new APIs from wrangler before it is released 🤔 

```suggestion
		"wrangler": "workspace:*"
```

PR #7544, opened 2024-12-13
Remove now-unused asset preservation cache v1 from Pages

GregBrimble:
Asset preservation cache v2 fully superceded v1 since 2024-05-17.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e coverage
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal change

PR #7564, opened 2024-12-16
Key the Early Hints off of the asset key rather than request path

GregBrimble:
Keys the Early Hints cache entries off of the asset key rather than the request path.

I could do [something like we did for asset preservation v2](https://github.com/cloudflare/workers-sdk/pull/7544/files) where we simultaneously checked the old and new formats to prevent a flood of new cache misses as we roll this out.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/pages-shared/asset-server/handler.ts
+++ b/packages/pages-shared/asset-server/handler.ts
@@ -366,6 +366,7 @@ export async function generateHandler<
 
 	async function attachHeaders(response: Response) {
 		const existingHeaders = new Headers(response.headers);
+		const eTag = existingHeaders.get("eTag")?.match(/^"(.*)"$/)?.[1];
 
 		const extraHeaders = new Headers({
 			"access-control-allow-origin": "*",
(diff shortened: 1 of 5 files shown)


packages/pages-shared/asset-server/handler.ts
```diff
@@ -364,13 +365,14 @@ export async function generateHandler<
...
-			isHTMLContentType(response.headers.get("Content-Type"))
+			isHTMLContentType(response.headers.get("Content-Type")) &&
+			eTag
 		) {
 			const preEarlyHintsHeaders = new Headers(headers);
 
 			// "Early Hints cache entries are keyed by request URI and ignore query strings."
 			// https://developers.cloudflare.com/cache/about/early-hints/
-			const earlyHintsCacheKey = `${protocol}//${host}${pathname}`;
+			const earlyHintsCacheKey = `${protocol}//${host}/${eTag}`;
```
GregBrimble, line 375: I could do [something like we did for asset preservation v2](https://github.com/cloudflare/workers-sdk/pull/7544/files) where we simultaneously checked the old and new formats to prevent a flood of new cache misses as we roll this out.

WillTaylorDev (approved):

packages/pages-shared/asset-server/handler.ts
GregBrimble: Nah, we added this in the first place without an issue, I'm sure it'll be fine without bothering to do that.

CarmenPopoviciu (approved):

PR #7575, opened 2024-12-17
Make workflow `instance.status()` return output equal to production workflows

LuisDuarte1:
Previously, in local dev, the `output` field would return the list of successful steps outputs in the workflow. This is not expected behavior compared to production workflows (where the output is the actual return of the `run` function), probably just an implementation detail not set left-over from before beta.

This commit makes it so that `output` is equal to production behavior. 

For observability’s sake, I kept the old step output list in a different field `__LOCAL_DEV_STEP_OUTPUTS` - I think this is a good enough compromise right now since we want local dev to be correct against prod first. We can remove `__LOCAL_DEV_STEP_OUTPUTS` later, once we figure out on how to add custom stuff to the devtools page.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: just local dev ig
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this was expected behavior


packages/workflows-shared/src/binding.ts
```diff
@@ -92,16 +92,30 @@ export class WorkflowHandle extends RpcTarget implements WorkflowInstance {
...
 	}
 
-	public async status(): Promise<InstanceStatus> {
+	public async status(): Promise<
+		InstanceStatus & { __LOCAL_DEV_STEP_OUTPUTS: unknown[] }
+	> {
 		const status = await this.stub.getStatus(0, this.id);
-		const { logs } = await this.stub.readLogs();
-		// @ts-expect-error TODO: Fix this
+		const { logs } = (await this.stub.readLogs()) as { logs: unknown[] };
```
andyjessop, line 99: Why do we type a log as `unknown` here? We could instead create a type assertion function.

```ts
type Log = {
  event: InstanceEvent,
  metadata: {
    result: unknown
  }
};

function assertHasLogs(obj: unknown): obj is { logs: Log[] } {
  // check for props
}
```

Could probably have just one TODO to remove once correct typings are added to `readLogs`.

andyjessop:
I think we'll need E2E tests here because we test workflows in our dev E2E suite.

packages/workflows-shared/src/binding.ts
LuisDuarte1: I've fixed the `readLogs` return type - there's just one @ts-expect-error on the binding return because, agh, workers-types are a bit wrong. I'll follow-up with another PR once binding types are fixed.

andyjessop (approved):
Great, thanks!

PR #7576, opened 2024-12-17
Remove defaults for `batch-max-*` pipeline parameters and define value ranges

cmackenzie1:
Closes https://jira.cfdata.org/browse/PIPE-152

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: it is CLI help info
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: it is documentation

### DIFF
--- a/packages/wrangler/src/pipelines/index.ts
+++ b/packages/wrangler/src/pipelines/index.ts
@@ -196,7 +196,9 @@ export function pipelines(pipelineYargs: CommonYargsArgv) {
 					args.compression === undefined ? "gzip" : args.compression;
 
 				const batch = {
-					max_mb: args["batch-max-mb"],
+					max_bytes: args["batch-max-mb"]
+						? args["batch-max-mb"] * 1000 * 1000 // convert to bytes for the API
+						: undefined,
 					max_duration_s: args["batch-max-seconds"],
 					max_rows: args["batch-max-rows"],
 				};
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/pipelines/index.ts
```diff
@@ -101,20 +101,23 @@ function addCreateAndUpdateOptions(yargs: Argv<CommonYargsOptions>) {
 		})
 		.option("batch-max-mb", {
 			describe:
-				"The approximate maximum size of a batch before flush in megabytes \nDefault: 10",
+				"The approximate maximum size (in megabytes) for each batch before flushing (range: 1 - 100)",
 			type: "number",
+			default: 10,
```
edmundhung, line 106: I see that we mentioned the default value in the description already. Is this the same default value we have on the API side? That's why this is a non-breaking change?

cmackenzie1: Yeah, 10 MB is the current default on the API side. Regarding breaking changes, this is currently in private beta and changes like this are expected.

cmackenzie1: The defaults are still being determined, so I removed here and will instead only set them on the API side.

packages/wrangler/src/pipelines/index.ts
```diff
@@ -386,7 +388,8 @@ export function pipelines(pipelineYargs: CommonYargsArgv) {
 					pipelineConfig.destination.compression.type = args.compression;
 				}
 				if (args["batch-max-mb"]) {
-					pipelineConfig.destination.batch.max_mb = args["batch-max-mb"];
+					pipelineConfig.destination.batch.max_mb =
```
penalosa, line 391: Why is this bytes if the API property is called `max_mb`?

cmackenzie1: Good catch, let me update it!

PR #7582, opened 2024-12-18
feat(cloudflare-workers-bindings-extensions): list bindings on the sidebar

edmundhung:
Fixes [DEVX-1472](https://jira.cfdata.org/browse/DEVX-1472)

This PR introduce the bindings view on the vscode extension that lists all the supported bindings in the sidebar. This implementation assumes the wrangler config is present on the root of the workspace folder and does not support monorepo at the moment.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: No e2e tests for VSCode extension
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: unreleased


packages/cloudflare-workers-bindings-extension/package.json
```diff
@@ -19,6 +19,7 @@
 		"package": "pnpm run check:type && pnpm run check:lint && node esbuild.js --production",
 		"compile-tests": "tsc -p . --outDir out",
 		"watch-tests": "tsc -p . -w --outDir out",
+		"test": "node ./out/test/runTest.js",
```
penalosa, line 22: We use `tsx` in various other places for stuff like this. Would that work? I'd alos love to see if it's possible to run this in CI—does it fail if you add a `test:ci` script?

penalosa (approved):
Minor comment about tests, but otherwise looking good!

packages/cloudflare-workers-bindings-extension/package.json
edmundhung: Just tried that but it fails with this message:

```
Error: Cannot find module '/Users/edmund/Workspace/workers-sdk/packages/cloudflare-workers-bindings-extension/src/test/suite/index'
```

I have also pushed a commit with the `test:ci` script. The tests seems to be running fine on windows and macos but fails on ubuntu (https://github.com/cloudflare/workers-sdk/actions/runs/12401086099/job/34619688337). I can look at this next as a follow-up.

edmundhung:
> Minor comment about tests, but otherwise looking good! We should also make sure to render all bindings, not just the ones we can edit

I can do a follow-up and extend it to all bindings tomorrow 👍🏼 (or on this branch if the main is still locked)

packages/cloudflare-workers-bindings-extension/package.json
edmundhung: Looks like we need to have Xvfb enabled specifically for ubuntu:
https://github.com/cloudflare/workers-sdk/actions/runs/12401086099/job/34619688337

Issue #7577, opened 2024-12-18
🐛 BUG: Wrangler pages fails to deploy


michaelpayne02:
### Which Cloudflare product(s) does this pertain to?

Pages, Wrangler

### What version(s) of the tool(s) are you using?

3.97.0 [Wrangler]

### What version of Node are you using?

Deno 2.1.4

### What operating system and version are you using?

Mac Sequoia 15.2

### Observed behavior
When deploying to pages via `npx wrangler pages deploy dist --project-name=project`, wrangler fails at some point when configuring unenv. Not sure why it's trying to load the package.json from the dist subdirectory when it's actually in the parent folder.

### Expected behavior
Wrangler should upload pages projects.

### Steps to reproduce
Please provide the following:
- Try to upload any pages project with wrangler deploy

### Please provide any relevant error logs

✘ [ERROR] ENOTDIR: not a directory, open '.../node_modules/unenv-nightly/dist/index.cjs/package.json'

penalosa:
Thanks for reporting this! We're investigating—in the meantime downgrading to Wrangler 3.96.0 should be a workaround

PR #7583, opened 2024-12-18
Revert "feat: add support for custom unenv resolve path"

penalosa:
Reverts cloudflare/workers-sdk#7522

This should resolve https://github.com/cloudflare/workers-sdk/issues/7577, and fix the C3 E2E failures we're seeing, which seem to be due to a failure to resolve unenv properly


CarmenPopoviciu (approved):

CarmenPopoviciu:
@penalosa and I have both tested with this pre-release version and can confirm the revert fixes the issue. Will go ahead and merge

vicb:
Many thanks @penalosa for handling the revert of something I broke :(

FYI @pi0 has fixed [the root cause in unenv](https://github.com/unjs/unenv/pull/378) - it is linked to package resolution and the fact that `pnpm` set the name of `"unenv": "npm:unenv-nightly@..."` to `unenv-nightly` in the lock file.

I'll resubmit the PRs and make sure I test the pre-release version locally.

🙏

petebacondarwin:
I think we need to go further and add tests (probably to nodejs-hybrid fixture) that will prevent this regression.

PR #7604, opened 2024-12-20
feat(wrangler): Capture Workers with static assets in the telemetry data

CarmenPopoviciu:
Fixes DEVX-1525

During many Workers+Assets conversations, folks have mentioned that we expect usage for assets in remote mode to be fairly small. We should measure accurately what this number is, as it will be a very helpful data point down the road, when more decisions around remote mode will have to be taken.

This PR adds this kind of insight to our telemetry data, by capturing whether the command running is in the context of a Workers + Assets project. 

N.B. With these changes in place we will be capturing the Workers + Assets context for all commands, not just `wrangler dev --remote`.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/metrics/metrics-dispatcher.ts
+++ b/packages/wrangler/src/metrics/metrics-dispatcher.ts
@@ -111,6 +111,7 @@ export function getMetricsDispatcher(options: MetricsConfigOptions) {
 					isPagesCI: isPagesCI(),
 					isWorkersCI: isWorkersCI(),
 					isInteractive: isInteractive(),
+					hasAssets: options.hasAssets ?? false,
 					argsUsed,
 					argsCombination,
 				};
--- a/.changeset/long-houses-mate.md
+++ b/.changeset/long-houses-mate.md
@@ -1,0 +1,11 @@
+---
+"wrangler": minor
+---
+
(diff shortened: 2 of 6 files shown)


packages/wrangler/src/metrics/metrics-dispatcher.ts
```diff
@@ -111,6 +111,7 @@ export function getMetricsDispatcher(options: MetricsConfigOptions) {
 					isPagesCI: isPagesCI(),
 					isWorkersCI: isWorkersCI(),
 					isInteractive: isInteractive(),
+					hasAssets: options.hasAssets ?? false,
```
CarmenPopoviciu, line 114: this might be a tad misleading, since if `options.hasAssets` is `undefined`, we'll default to `false`, which might not be entirely true. However, the only use case I can see this happening is for code paths that call the old `metrics.sendMetricsEvent()`. The new metricsDispatcher will handle this correctly

andyjessop (approved):
So glad to be seeing these metrics in the wild!!

edmundhung (approved):

PR #7605, opened 2024-12-20
feat: add options to readConfig() to specify control how we deal with missing environment definitions

petebacondarwin:
Fixes #0000

This will be used by the Vite plugin, since `vite dev` command will always set the environment
to "development", and we don't want the user to have to define this explicitly if they don't want to.
In such cases there is no need for the user to define this environment explicitly,
even if there are other environments defined.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal API only

### DIFF
--- a/packages/wrangler/src/__tests__/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/configuration.test.ts
@@ -3,6 +3,7 @@ import path from "node:path";
 import { experimental_readRawConfig, readConfig } from "../config";
 import { normalizeAndValidateConfig } from "../config/validation";
 import { run } from "../experimental-flags";
+import { mockConsoleMethods } from "./helpers/mock-console";
 import { normalizeString } from "./helpers/normalize";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { writeWranglerConfig } from "./helpers/write-wrangler-config";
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/__tests__/configuration.test.ts
```diff
@@ -45,6 +46,73 @@ describe("readConfig()", () => {
...
+					bar: {},
+				},
+			});
+			const config = readConfig(
+				{ config: "wrangler.toml", env: "foo" },
+				{ optionalEnvironments: ["foo"] }
+			);
+			expect(config.name).toEqual("test-name");
+		});
+	});
```
jamesopstad, line 115: It might be clearer to split this into two separate tests, one to check that no warnings are logged and one to check the worker name is not updated.

jamesopstad (commented):
This looks good. Did you mean to also include the change where environment names are not appended if the environment is specified but there is only a top-level environment?

packages/wrangler/src/__tests__/configuration.test.ts
petebacondarwin: Done

petebacondarwin:
> Did you mean to also include the change where environment names are not appended if the environment is specified but there is only a top-level environment?

Thanks. I missed that. I fixed it now but it resulted in some renaming of the options to make them make sense. PTAL @jamesopstad

jamesopstad (commented):
Thanks Pete. I feel like the following test cases need to be added. They may already pass but I'm not totally confident from looking at the code.

`it("should update the worker name if an environment is requested that is defined and allowEnvironmentsWhenNoneAreDefined is true")`

`it("should update the worker name if an environment name in optionalEnvironments is requested that is defined")`

Is the second one the behaviour we agreed on?

petebacondarwin:
> it("should update the worker name if an environment name in optionalEnvironments is requested that is defined")
> Is the second one the behaviour we agreed on?

Yes, I believe this is what we agreed. If the user has defined the environment that is selected, then it will do its normal processing.

I am still wondering about this though. If the user doesn't have a production environment defined and they deploy the worker name will not be suffixed. But as soon as the define an environment it will be suffixed, and they would have to provide a specific Worker name in that environment to get the same name as before. Is this what we want?

jamesopstad:
> I am still wondering about this though. If the user doesn't have a production environment defined and they deploy the worker name will not be suffixed. But as soon as the define an environment it will be suffixed, and they would have to provide a specific Worker name in that environment to get the same name as before. Is this what we want?

That's one of the reasons I was in favour of always adding the suffix (including in the default `production` case). My feeling is that environments should always be used (either implicitly or explicitly) rather than something that only take effect when the user defines them explicitly. I appreciate that we've already ruled this out but I think there are tradeoffs with each solution.

petebacondarwin:
OK we changed our minds about how best to do this. So I will close this and make a new PR.

PR #7612, opened 2024-12-22
fix: better handler malformed paths

### LINKED ISSUE #7611: URI malformed error with Workers Assets and malformed paths
If you send a request to the Assets server with a malformed URL, it throws an error like `Error: URIError: URI malformed`

This currently makes it easy to cause errors on Assets sites, like `https://developers.cloudflare.com/%A0`, or `https://placeholders.dev/%A0`.

Cherry:
Fixes #7611

This more gracefully handles malformed URLs and simply throws a 404 on them instead of the entire Worker throwing an exception.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e tests for this
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no user-facing changes

### DIFF
--- a/packages/workers-shared/asset-worker/src/handler.ts
+++ b/packages/workers-shared/asset-worker/src/handler.ts
@@ -18,12 +18,7 @@ export const handleRequest = async (
 ) => {
 	const { pathname, search } = new URL(request.url);
 
-	let decodedPathname;
-	try {
-		decodedPathname = decodePath(pathname);
-	} catch (err) {
-		return new NotFoundResponse();
-	}
+	let decodedPathname = decodePath(pathname);
 	// normalize the path; remove multiple slashes which could lead to same-schema redirects
 	decodedPathname = decodedPathname.replace(/\/+/g, "/");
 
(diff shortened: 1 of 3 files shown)


petebacondarwin (changes requested):
I am not sure how this can happen in practice - i.e. whether it is possible to access a real asset via a malformed URL segment.

But I feel like we should just be passing through these invalid segments rather than 404ing.
So instead of this fix we just change `decodePath()` so that it is resilient there: catching the error for each segment and just returning the original string instead of the decoded one.

If there is no such asset, then it would still 404 but at least we are not potentially hiding a real asset.

Cherry:
Good call, that makes sense, thanks Pete. I've updated the changes and tests to account for this more gracefully.

packages/workers-shared/asset-worker/src/handler.ts
```diff
@@ -692,6 +700,14 @@ export const decodePath = (pathname: string) => {
 const encodePath = (pathname: string) => {
 	return pathname
 		.split("/")
-		.map((x) => encodeURIComponent(x))
+		.map((x) => {
+			let encoded;
+			try {
+				encoded = encodeURIComponent(x);
```
petebacondarwin, line 706: Can encodeURIComponent actually throw?

Cherry: If given something like a lone surrogate, yes:

`encodeURIComponent("\uD800")`

How likely/possible this is to happen in a URL? Probably not, but I felt like it couldn't hurt to wrap both.

petebacondarwin (approved):

PR #7648, opened 2024-12-30
bump node.js to v.20 for all C3 workflows

dario-piotrowicz:
The C3 e2es (both experimental and not) tests for Nuxt using yarn under ubuntu seem to be consistently failing on every PR that runs them.

By looking at the logs it looks like Nuxt is incompatible with the version of node that we're using in our workflows:
Screenshot 2024-12-30 at 11 53 44

So in this PR I am simply bumping the node version used by all the C3 workflows as that should be pretty safe to do I think.

I did not change the [default node version for all workflows](https://github.com/cloudflare/workers-sdk/blob/acbea32c6df82269b435e34d8c51fe8a26c973f1/.github/actions/install-dependencies/action.yml#L6) as I assume that we have a specific reason as to why we want to use `18.20.2` there (if not I am happy to change that value instead)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: this is just a CI change
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this is just a CI change

### DIFF
--- a/.github/workflows/c3-dependabot-versioning-prs.yml
+++ b/.github/workflows/c3-dependabot-versioning-prs.yml
@@ -18,6 +18,9 @@ jobs:
           token: ${{ secrets.GH_ACCESS_TOKEN }}
       - name: Install Dependencies
         uses: ./.github/actions/install-dependencies
+        with:
+          node-version: 20
+
       - name: Configure Git
         run: |
           git config --global user.email wrangler@cloudflare.com
--- a/.github/workflows/c3-e2e-experimental.yml
+++ b/.github/workflows/c3-e2e-experimental.yml
@@ -50,6 +50,7 @@ jobs:
           turbo-team: ${{ secrets.TURBO_TEAM }}
           turbo-token: ${{ secrets.TURBO_TOKEN }}
           turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
+          node-version: 20
 
       - name: E2E Tests
         if: steps.changes.outputs.everything_but_markdown == 'true'
--- a/.github/workflows/c3-e2e-quarantine.yml
+++ b/.github/workflows/c3-e2e-quarantine.yml
@@ -39,6 +39,7 @@ jobs:
           turbo-team: ${{ secrets.TURBO_TEAM }}
           turbo-token: ${{ secrets.TURBO_TOKEN }}
           turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
+          node-version: 20
 
       - name: E2E Tests
         uses: ./.github/actions/run-c3-e2e
--- a/.github/workflows/c3-e2e.yml
+++ b/.github/workflows/c3-e2e.yml
@@ -50,6 +50,7 @@ jobs:
           turbo-team: ${{ secrets.TURBO_TEAM }}
           turbo-token: ${{ secrets.TURBO_TOKEN }}
           turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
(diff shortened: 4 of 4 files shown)


edmundhung:
Wrangler v3 currently supports Node v16+ as stated in the docs. But we are changing it to 18+ in v4 following the Nodejs support window (https://nodejs.org/en/about/previous-releases). So the wrangler workflow should probably stay in node16 for now.

But if we are just changing the c3 tests for frameworks to use node 18, i think it is fine?

dario-piotrowicz:
> Wrangler v3 currently supports Node v16+ as stated in the docs. But we are changing it to 18+ in v4 following the Nodejs support window (https://nodejs.org/en/about/previous-releases). So the wrangler workflow should probably stay in node16 for now.

The value I was referring to is the default one for the install-dependecies action:
https://github.com/cloudflare/workers-sdk/blob/ab7204a6eb8f5bab00172a6a2623251ba2d757ba/.github/actions/install-dependencies/action.yml#L6

So, I think that it should be the one most commonly used across all the various GitHub workflows.

There are a few gh workflows that use the action: [GitHub search](https://github.com/search?q=repo%3Acloudflare%2Fworkers-sdk+%2Fuses%3A+.*%3Finstall-dependencies%2F&type=code), I am not too involved/familiar with the monorepo so I cannot really say what default would make most sense without digging deeper into things a bit.

Interestingly the wrangler e2es are already purposely run under node 18 actually (instead of 16):
https://github.com/cloudflare/workers-sdk/blob/ab7204a6eb8f5bab00172a6a2623251ba2d757ba/.github/workflows/e2e.yml#L23

> But if we are just changing the c3 tests for frameworks to use node 18, i think it is fine?

yeah I do think so, alternatively we skip the nuxt+yarn+ubuntu tests for the time being, but I would personally prefer to just bump the node version as I think we can reasonably expect people to use a relatively newer version of node when scaffolding a new application.

Actually 20 is not an LTS, so I wonder if it would be better to just straight jump to 22 🤔 
[Screenshot 2024-12-30 at 14 25 16](https://nodejs.org/en/about/previous-releases)

One thing to note is that c3 has support for node 18 and up:
https://github.com/cloudflare/workers-sdk/blob/ab7204a6eb8f5bab00172a6a2623251ba2d757ba/packages/create-cloudflare/package.json#L91

Anyways I am fine with whatever the team prefers 🙂

edmundhung:
> One thing to note is that c3 has support for node 18 and up:

In that case I think c3 workflow should use node 18 in general, but node 20+ just for framework tests. We have no control over the node version requirements of the frameworks but the rest of the template should remain supported with node 18.

dario-piotrowicz:
> In that case I think c3 workflow should use node 18 in general, but node 20+ just for framework tests. We have no control over the node version requirements of the frameworks but the rest of the template should remain supported with node 18.

This makes sense to me 👍 , unfortunately C3 e2es run all together, we don't currently have a way to run separately the ones that use frameworks from the ones that don't 🥲

Of course we could implement that, but it would, in my opinion, require quite some refactoring and make the e2es process generally more complicated, so I don't think that it would be worth it 😕

Personally I would go with just using node 20+ (or 22+ as I mentioned) and:
 - as I mentioned assume that people do use modern node versions for new apps scaffolding
 - assume that if things work fine with node 20+/22+ then most likely they run well with node 18+ (since we don't use any newer fancy node feature as far as I know)
 - on the next major version of C3 we bump the node requirement accordingly realigning everything

dario-piotrowicz:
Closing in favour of #7649 (turns out a non-controversial patch bump was enough 🤦)

PR #7660, opened 2025-01-03
docs: clarify difference between external contributors and Cloudflare employees

andyjessop:
Fixes #000.

We sometimes get PRs from Cloudflare employees via a forked repo, which makes it difficult to contribute to that PR (we can't push directly to their fork like we can with PRs created from a branch on this repo). 

This PR clarifies the difference in the CONTRIBUTING.md guide to try to avoid the above situation.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: docs change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: docs change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,8 @@ Wrangler is built and run on the Node.js JavaScript runtime.
 
 ### Fork and clone this repository
 
+#### For External Contributors
+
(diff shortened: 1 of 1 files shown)


andyjessop:
> > we can't push directly to their fork like we can with PRs created from a branch on this repo
> 
> We should be able to if they check the checkbox allowing that, I think? You can do `gh pr checkout PR_NUMBER` and then push to it

@penalosa yes, I think they/we can do that - what I'm trying to do here is just alleviate the issue a little.

penalosa (approved):

PR #7674, opened 2025-01-06
add dotenv support to wrangler secret bulk

### LINKED ISSUE #6388: 🚀 Feature Request: wrangler secret:bulk should support uploading .dev.vars file format
### Describe the solution

Currently the secret:bulk command only supports secrets formatted as JSON. 

This creates an extra step for people uploading secrets that they also use in local dev using the .dev.vars format.

It would be super convenient if the secret:bulk upload command could support the .env file format :)

Ankcorn:
Fixes #6388

Adds support for env files in all the secret bulk commands

Questions for the team: 

currently the parsing is toggled between JSON and env by adding a flag `--experimental-env-file`. 
Should this be made "no config" by trying to parse both and seeing which succeeds? 

Will update the docs once the flag stuff is confirmed

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: changes don't require workers and can be effectively unit tested
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/19056


packages/wrangler/src/pages/secret/index.ts
```diff
@@ -167,18 +165,22 @@ export const secret = (secretYargs: CommonYargsArgv, subHelp: SubHelp) => {
...
 		)
 		.command(
-			"bulk [json]",
+			"bulk [data]",
 			"Bulk upload secrets for a Pages project",
 			(yargs) => {
 				return yargs
-					.positional("json", {
-						describe: `The JSON file of key-value pairs to upload, in form {"key": value, ...}`,
+					.positional("data", {
```
penalosa, line 172: Unfortunately this is a breaking change, since yargs supports passing positionals as named args (`--json`). As it happens, we were going to make this change in v4 anyway, so would you be able to rename the arg back to `json` in this PR and open a second PR to the v4 branch that just does this argument rename?

Ankcorn: Ah thanks for the heads up 🧡

packages/wrangler/src/pages/secret/index.ts
```diff
@@ -167,18 +165,22 @@ export const secret = (secretYargs: CommonYargsArgv, subHelp: SubHelp) => {
...
 					})
 					.option("project-name", {
 						type: "string",
 						alias: ["project"],
 						description: "The name of your Pages project",
+					})
+					.option("experimental-env-file", {
+						describe: "Parse an env file",
+						type: "boolean",
 					});
```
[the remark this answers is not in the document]
Ankcorn: Sounds good

irvinebroque (commented):
So this is basically

*Before:* you can pass `wrangler secret bulk` a path to a file, but has to be JSON

*After:* you can pass `wrangler secret bulk` a path to a file, and that file can be either JSON, or KEY=VALUE ala env vars

...making it easy to go "I have a `.env` file, I want to add all of it to secrets.
...but not coupling it specifically to the filename `.env`, or to `dotenv` behaviors, etc.

Correct?

Ankcorn:
> So this is basically
> 
> _Before:_ you can pass `wrangler secret bulk` a path to a file, but has to be JSON
> 
> _After:_ you can pass `wrangler secret bulk` a path to a file, and that file can be either JSON, or KEY=VALUE ala env vars
> 
> ...making it easy to go "I have a `.env` file, I want to add all of it to secrets. ...but not coupling it specifically to the filename `.env`, or to `dotenv` behaviors, etc.
> 
> Correct?

Exactly

penalosa:
> So this is basically
> 
> _Before:_ you can pass `wrangler secret bulk` a path to a file, but has to be JSON
> 
> _After:_ you can pass `wrangler secret bulk` a path to a file, and that file can be either JSON, or KEY=VALUE ala env vars
> 
> ...making it easy to go "I have a `.env` file, I want to add all of it to secrets. ...but not coupling it specifically to the filename `.env`, or to `dotenv` behaviors, etc.
> 
> Correct?

In Wrangler's case, I think it'd likely be a case of having a `.dev.vars` file which you want to upload

.changeset/silver-plums-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Add support for env files to wrangler secret bulk
```
penalosa, line 5: Could you expand this with an example, and call out `.dev.vars`?

packages/wrangler/src/__tests__/secret.test.ts
```diff
@@ -1065,7 +1087,7 @@ describe("wrangler secret", () => {
 				"wrangler secret:bulk [json]
 
 				POSITIONALS
-				  json  The JSON file of key-value pairs to upload, in form {\\"key\\": value, ...}  [string]
+				  json  The file of key-value pairs to upload, as JSON in form {\\"key\\": value, ...} or .env file in the form KEY=VALUE  [string]
```
penalosa, line 1090:
```suggestion
				  json  The file of key-value pairs to upload, as JSON in form {\\"key\\": value, ...} or .dev.vars file in the form KEY=VALUE  [string]
```

penalosa (approved):

irvinebroque:
> currently the parsing is toggled between JSON and env by adding a flag --experimental-env-file. Should this be made "no config" by trying to parse both and seeing which succeeds?

I'm having a hard time seeing a reason _not_ to do that — you give us file, we detect format

packages/wrangler/src/secret/index.ts
```diff
@@ -563,3 +538,47 @@ export function validateJSONFileSecrets(
...
+	let content: Record<string, string>;
+	if (input) {
+		const jsonFilePath = path.resolve(input);
+		try {
+			const fileContent = readFileSync(jsonFilePath);
+			try {
+				content = parseJSON<Record<string, string>>(fileContent);
+			} catch (e) {
+				content = dotenvParse(fileContent);
+				// dotenvParse does not error unless fileContent is undefined, no keys === error
```
penalosa, line 552: What if it's not a valid env file format? Something like:

```
2:4
3==5
```

Ankcorn:
```javascript
import { parse } from "dotenv";

console.log(
	parse(`2:4
4==5`)
);
//  { '4': '5' }

```

dotenv parse is rather lax, it only throws an exception if you pass undefined which is guarded against. in the case it can't parse the file it just returns an empty object

PR #7681, opened 2025-01-06
chore: add d1 and kv to e2e cleanup job

emily-shen:
Fixes https://jira.cfdata.org/browse/DEVX-1537

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: doesn't affect e2e's (cleanup is a separate job)
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal

### DIFF
--- a/tools/e2e/e2eCleanup.ts
+++ b/tools/e2e/e2eCleanup.ts
@@ -71,7 +71,7 @@ async function run() {
 	const d1DatabasesToDelete = await listTmpDatabases();
 	for (const db of d1DatabasesToDelete) {
 		console.log("Deleting D1 database: " + db.name);
-		await deleteDatabase(db.name);
+		await deleteDatabase(db.uuid);
 	}
 	if (d1DatabasesToDelete.length === 0) {
 		console.log(`No D1 databases to delete.`);
--- a/tools/e2e/e2eCleanup.ts
+++ b/tools/e2e/e2eCleanup.ts
@@ -1,8 +1,12 @@
 import {
+	deleteDatabase,
+	deleteKVNamespace,
 	deleteProject,
 	deleteWorker,
+	listTmpDatabases,
 	listTmpE2EProjects,
 	listTmpE2EWorkers,
+	listTmpKVNamespaces,
 } from "./common";
 
 if (!process.env.CLOUDFLARE_API_TOKEN) {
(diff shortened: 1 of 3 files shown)


tools/e2e/e2eCleanup.ts
```diff
@@ -22,6 +26,22 @@ run().catch((e) => {
 });
 
 async function run() {
+	// KV namespaces don't have a creation timestamp, but deletion will fail if there is a worker bound to it
+	// so delete these first to avoid interrupting running e2e jobs (unless you are very very unlucky)
+	const kvNamespacesToDelete = await listTmpKVNamespaces();
+	for (const kvNamespace of kvNamespacesToDelete) {
+		console.log("Deleting KV namespace: " + kvNamespace.title);
+		await deleteKVNamespace(kvNamespace.id);
+	}
```
andyjessop, line 35: We could potentially save some seconds by doing these deletions in parallel:

```ts
const promises = [];
for (const kvNamespace of kvNamespacesToDelete) {
	console.log("Deleting KV namespace: " + kvNamespace.title);
	promises.push(deleteKVNamespace(kvNamespace.id));
}
await Promise.all(promises);
```

tools/e2e/e2eCleanup.ts
```diff
@@ -47,4 +67,17 @@ async function run() {
 	} else {
 		console.log(`Successfully deleted ${workersToDelete.length} workers`);
 	}
+
+	const d1DatabasesToDelete = await listTmpDatabases();
+	for (const db of d1DatabasesToDelete) {
+		console.log("Deleting D1 database: " + db.name);
+		await deleteDatabase(db.uuid);
+	}
```
andyjessop, line 75: As above. Unless they rely on each other, I presume we can do these in parallel.

andyjessop (commented):
Looking good! Just an optimisation comment that might save a fair amount of time if we're doing this a lot.

tools/e2e/e2eCleanup.ts
emily-shen: Would this end up spamming the API too much? I'm not sure what sort of rate limiting there is on there. Also this only runs once a day out of EU/US hours, so hopefully CI time shouldn't be a problem. I'm inclined to leave it for now, and if we end up having problems we can revisit it?

andyjessop (approved):
Good stuff!!

PR #7684, opened 2025-01-07
refactor(CI): explicitly set encoding when reading package.json

vicb:
readFileSync returns a `Buffer` if the encoding is not set.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: CI
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.github/version-script.js
+++ b/.github/version-script.js
@@ -16,7 +16,7 @@ const { execSync } = require("child_process");
 try {
 	const packageName = getArgs()[0] ?? "wrangler";
 	const packageJsonPath = `./packages/${packageName}/package.json`;
-	const pkg = JSON.parse(readFileSync(packageJsonPath));
+	const pkg = JSON.parse(readFileSync(packageJsonPath, "utf-8"));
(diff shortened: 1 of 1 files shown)


petebacondarwin (commented):
This PR is unnecessary. The first thing that `JSON.parse()` does is stringify its input, which for a Buffer will convert it to a string. So this all works fine.

vicb:
> This PR is unnecessary

Yes but

1) This fixes a TS error 

image

2) Are you 100% that is is not necessary? At least I have no idea what encoding is used to create the buffer nor the encoding used to convert back to string from top of my head.

But we can close this if you prefer.

petebacondarwin (approved):
OK - let's land it, since it resolves the typings issue.
I am 100% sure it works as-is though. The default encoding for a buffer when converting to a string is utf8.

PR #7694, opened 2025-01-07
Default D1 export to --local

joshthoward:
Copy behavior of `wrangler d1 execute` for `wrangler d1 export`.

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
  - [X] Not required because: Change is covered by unit tests
- Public documentation
  - [ ] TODO (before merge)
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/19088


isaac-mcfadyen:
IMO this is a very breaking change (all users using D1 export will suddenly be exporting their local DB instead of their remote DB).

Because of that, is this planned to be a major version bump? Landing this in a minor/patch has the potential to disrupt many user workflows.

joshthoward:
This is not a breaking change. This is a feature enhancement where if `--local` or `--remote` are not specified `wrangler d1 export` will default to local rather than failing with a user error. All existing uses will continue working. This will get a minor version bump.

rozenmd:
To be clear, `wrangler d1 export` did not default to either local or remote. If you didn't specify, it would error:

```
% npx wrangler d1 export northwind --output ./test.sql

 ⛅️ wrangler 3.100.0
--------------------


✘ [ERROR] You must specify either --local or --remote
```

IMO this is a `feat`/minor bump

penalosa (approved):

penalosa:
For posterity, merging with a failing C3 check because this PR doesn't touch C3 and the same test is failing (flaking, probably) on main

PR #7697, opened 2025-01-08
chore(unenv-preset): drop unused .cjs files

vicb:
Use tsc to compile `@cloudflare/unenv-preset`

/cc @pi0 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/unenv-preset/build.config.ts
+++ b/packages/unenv-preset/build.config.ts
@@ -1,19 +1,7 @@
 import { defineBuildConfig } from "unbuild";
 
 export default defineBuildConfig({
-	declaration: true,
-	rollup: {
-		emitCJS: true,
-	},
 	entries: [
-		"src/index",
-		{ input: "src/runtime/", outDir: "dist/runtime", format: "esm" },
-		{
-			input: "src/runtime/",
-			outDir: "dist/runtime",
-			format: "cjs",
-			ext: "cjs",
-			declaration: false,
-		},
+		{ input: "src/", outDir: "dist/src", format: "esm", declaration: true },
 	],
 });
--- a/packages/unenv-preset/package.json
+++ b/packages/unenv-preset/package.json
@@ -21,22 +21,21 @@
 	},
 	"license": "MIT OR Apache-2.0",
 	"sideEffects": false,
+	"type": "module",
 	"exports": {
 		".": {
-			"types": "./dist/index.d.ts",
-			"require": "./dist/index.cjs",
-			"import": "./dist/index.mjs"
+			"types": "./dist/src/index.d.ts",
+			"import": "./dist/src/index.mjs"
 		},
 		"./runtime/*": {
-			"types": "./dist/runtime/*.d.ts",
-			"require": "./dist/runtime/*.cjs",
-			"import": "./dist/runtime/*.mjs"
+			"types": "./dist/src/runtime/*.d.ts",
+			"import": "./dist/src/runtime/*.mjs"
 		}
 	},
-	"main": "./dist/index.cjs",
-	"types": "./dist/index.d.ts",
+	"main": "./dist/src/index.mjs",
+	"types": "./dist/src/index.d.ts",
 	"files": [
-		"dist"
+		"dist/src"
 	],
 	"scripts": {
 		"build": "unbuild",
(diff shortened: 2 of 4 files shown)


petebacondarwin (approved):

packages/unenv-preset/build.config.ts
```diff
@@ -1,19 +0,0 @@
...
-	},
-	entries: [
-		"src/index",
-		{ input: "src/runtime/", outDir: "dist/runtime", format: "esm" },
-		{
-			input: "src/runtime/",
-			outDir: "dist/runtime",
-			format: "cjs",
-			ext: "cjs",
-			declaration: false,
```
pi0, line 16: This line was actually instructing unbuild to emit CJS (it is not a default behavior).

Using tsc makes sense for consistency with cloudflare repo. unbuild uses mkdist and transforms such as ensuring output relative imports are ESM-compatible (with explicit `.mjs` extension). The behavior might vary when using tsc alone..

packages/unenv-preset/package.json
```diff
@@ -21,32 +21,31 @@
...
-			"import": "./dist/index.mjs"
+			"types": "./dist/src/index.d.ts",
+			"import": "./dist/src/index.js"
 		},
 		"./runtime/*": {
-			"types": "./dist/runtime/*.d.ts",
-			"require": "./dist/runtime/*.cjs",
-			"import": "./dist/runtime/*.mjs"
+			"types": "./dist/src/runtime/*.d.ts",
+			"import": "./dist/src/runtime/*.js"
```
pi0, line 32: I highly recommend to use `.mjs` extension for tooling compatibility. `.js` at **least** has performance overhead that tools need to travese and read closest `package.json` only to find about right format.

packages/unenv-preset/build.config.ts
vicb: > The behavior might vary when using tsc alone and less consistent with unenv artifacts (which use unbuild/mkdist for ESM compliance)

Do you foresee any issues?

I switched to `tsc` for 2 reasons:

- consistency with other packages in `workers-sdk`
- I had troubles compiling tests to a different `outDir` with `unbuild`

packages/unenv-preset/package.json
vicb: Is there a way to output `.mjs` with `tsc`? (I checked but didn't find a way)

pi0: AFAIK there was a recent pr they introduced to rewrite paths. still `/index` is not auto fixed (tsc does not adheres ESM all the times -- that's the major issue and why mkdist process is safer)

packages/unenv-preset/build.config.ts
pi0: > Consistency with other packages in workers-sdk

consistency with unenv output and (end-user) that consume the dist I believe is more important for this case. strict ESM compliance i think is also an important point for us.

> I had troubles compiling tests to a different outDir with unbuild

Do you have a branch that was trying to do this? (also why tests need compilation?)

vicb: > also why tests need compilation?

We need to launch `wrangler deploy --dry-run` with a custom resolution path (the current package) and a custom `workerd` (from the current package), both are set by env variables. So what we need to compile if the file that will launch this build.

> Do you have a branch that was trying to do this?

Basically what I had was:

```
	entries: [
		"src/index",
		{ input: "src/runtime/", outDir: "dist/runtime", format: "esm" },
		{ input: "tests/build.ts", outDir: "dist-test", format: "esm" },
	]
```    

From the log, I could see that `dist-test/` was cleared but still build.ts would be compiled into `dist/tests/build.esm`

pi0: (haven't locally tested) you might want to set `builder: 'mkdist'` for second entry. When there is traling slash like `runtime/` it is infered (default is rollup build with multi entries).

Alternatively you could tsc only to emit test artifacts..

vicb:
Thanks for the reviews and feedback @petebacondarwin and @pi0 

I have updated the PR to keep unbuild but drop .cjs files.
The output dir has been changed to `dist/src` as `dist/tests` will be created in a follow up PR.

PR #7700, opened 2025-01-08
test(unenv-preset): move tests from nodejs-hybrid-app

vicb:
Move the tests from nodejs-hybrid-app

The tests are executed with the dev version of `@cloudflare/unenv-preset` and the pkg local version of `workerd` (instead of the pinned versions that will be used by wrangler).

Note that currently wrangler still uses the cloudflare preset from `unjs/unenv` so that is what is tested for now. Once wrangler switches to using `@cloudflare/unenv-preset` the resolution override will be in effect (we are trying to resolve `$clouflare.ts` files that are not in `@cloudflare/unenv-preset`) 

Tests are disabled on Windows as `MINIFLARE_WORKERD_PATH` is [not supported](https://github.com/cloudflare/workers-sdk/blob/773bda8b38d43102c2a66126df92d3bbc7e80861/packages/miniflare/test/index.spec.ts#L2036).

/cc @pi0

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/fixtures/nodejs-hybrid-app/src/index.ts
+++ b/fixtures/nodejs-hybrid-app/src/index.ts
@@ -6,7 +6,6 @@ import { Stream } from "node:stream";
 import { Context } from "vm";
 import { Client } from "pg";
 import { s } from "./dep.cjs";
-import { testUnenvPreset } from "./unenv-preset";
 
 testBasicNodejsProperties();
 
@@ -29,8 +28,6 @@ export default {
 				return testX509Certificate();
 			case "/test-require-alias":
 				return testRequireUenvAliasedPackages();
-			case "/test-unenv-preset":
-				return await testUnenvPreset();
 		}
 
 		return new Response(
@@ -39,7 +36,6 @@ export default {
 <a href="test-random">Test getRandomValues()</a>
 <a href="test-x509-certificate">Test X509Certificate</a>
 <a href="test-require-alias">Test require unenv aliased packages</a>
-<a href="test-unenv-preset">Test unenv preset</a>
 `,
 			{ headers: { "Content-Type": "text/html; charset=utf-8" } }
 		);
--- a/packages/unenv-preset/tests/index.test.ts
+++ b/packages/unenv-preset/tests/index.test.ts
@@ -1,0 +1,64 @@
+import fs from "node:fs";
+import path from "node:path";
+import { platform } from "node:process";
+import { fileURLToPath } from "node:url";
+import { fetch } from "undici";
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
+import { runWranglerDev } from "../../../fixtures/shared/src/run-wrangler-long-lived";
+import { TESTS } from "./worker/index";
+
+// Root of the current package
+const pkgDir = path.resolve(fileURLToPath(import.meta.url), "../..");
+// workerd binary
+const localWorkerdPath = path.join(pkgDir, "node_modules/.bin/workerd");
+// Base path for resolving `@cloudflare/unenv-preset` files
+const localPresetResolveBaseDir = path.join(pkgDir, "package.json");
+// Base path for resolving `unjs/unenv` files
+const localUnenvResolveBaseDir = path.join(
+	pkgDir,
+	"node_modules/unenv/package.json"
+);
+
+// `runWranglerDev` + `MINIFLARE_WORKERD_PATH` is not supported on Windows
+// See https://github.com/nodejs/node/issues/52554
+describe.skipIf(platform === "win32")(
+	`@cloudflare/unenv-preset ${platform} ${localWorkerdPath} ${fs.existsSync(localWorkerdPath)}`,
+	() => {
+		let wrangler: Awaited<ReturnType<typeof runWranglerDev>> | undefined;
+
+		beforeAll(async () => {
+			// Use workerd binary install in `@cloudflare/unenv-preset`
+			// rather than the one bundled with wrangler.
+			const MINIFLARE_WORKERD_PATH = localWorkerdPath;
+
+			// Use the preset from the local `@cloudflare/unenv-preset` and `unjs/unenv`
+			// rather than the one bundled with wrangler.
+			const WRANGLER_UNENV_RESOLVE_PATHS = [
+				localPresetResolveBaseDir,
+				localUnenvResolveBaseDir,
+			].join(",");
+
+			wrangler = await runWranglerDev(
+				path.join(__dirname, "worker"),
+				["--port=0", "--inspector-port=0"],
+				{
+					MINIFLARE_WORKERD_PATH,
+					WRANGLER_UNENV_RESOLVE_PATHS,
+				}
+			);
+		});
+
+		afterAll(async () => {
+			await wrangler?.stop();
+			wrangler = undefined;
+		});
+
+		test.for(Object.keys(TESTS))("%s", async (testName) => {
+			expect(wrangler).toBeDefined();
+			const { ip, port } = wrangler!;
+			const response = await fetch(`http://${ip}:${port}/${testName}`);
+			const body = await response.text();
+			expect(body).toMatch("OK!");
+		});
+	}
+);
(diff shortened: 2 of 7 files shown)


packages/unenv-preset/tests/index.test.ts
```diff
@@ -0,0 +1,49 @@
...
+// Skip tests on windows where MINIFLARE_WORKERD_PATH is not supported
+const isWindows = platform === "win32";
+
+describe("@cloudflare/unenv-preset", () => {
+	let wrangler: Awaited<ReturnType<typeof runWranglerDev>>;
+
+	beforeAll(async () => {
+		if (isWindows) {
+			return;
+		}
```
petebacondarwin, line 20: Can you not just use `describe.skipIf(isWindows)` rather than adding this in every hook and test?

petebacondarwin: Actually, is it not possible to make these tests work on Windows?
You say that MINIFLARE_WORKERD_PATH is not supported on Windows. Is that strictly accurate or is it that you just need to specify a different path for Windows (e.g. `"node_modules/.bin/workerd.cmd"` or `"node_modules/.bin/workerd.ps1"`)?

packages/unenv-preset/tests/build.ts
```diff
@@ -0,0 +1,28 @@
...
+// wrangler binary
+const wranglerPath = path.join(pkgDir, "../wrangler/bin/wrangler.js");
+// Base path for resolving `@cloudflare/unenv-preset` files
+const localPresetResolveBaseDir = path.join(pkgDir, "dist/src");
+// Base path for resolving `unjs/unenv` files
+const localUnenvResolveBaseDir = path.join(pkgDir, "node_modules/unenv");
+
+console.log(pkgDir);
+
+execSync(`node ${wranglerPath} deploy --dry-run --outdir ./dist/worker`, {
```
petebacondarwin, line 16: Sorry to be dumb but why are we building a Worker here?

petebacondarwin (commented):
I'm a bit confused about the testing strategy here.

It seems that you are compiling a Worker in order to then run it via `wrangler dev`. Is that right? Why do that rather than just point Wrangler at the Worker directly, telling it to use the different unenv preset?

packages/unenv-preset/tests/build.ts
vicb: That the same thing we do for `nodejs-hybrid-app` in the `"build": "wrangler deploy --dry-run --outdir=./dist",`

That is we need to serve the worker for the tests.

We have to `execSync` here to override the resolution base folders (via `WRANGLER_UNENV_RESOLVE_PATHS`).

That's a good question, I'll add a comment to the file.

vicb:
> I'm a bit confused about the testing strategy here.
> 
> It seems that you are compiling a Worker in order to then run it via `wrangler dev`. Is that right? Why do that rather than just point Wrangler at the Worker directly, telling it to use the different unenv preset?

Yep that should work too.

I only did what `nodejs-hybrid-app` does without thinking too much.

packages/unenv-preset/tests/index.test.ts
petebacondarwin: I think that comment in Miniflare is not the same thing. What they are saying there (I believe) is that their test uses a shebang binary to trigger workerd, which would not work on Windows. But when you install workerd it should provide a Windows compatible binary, which usually ends in .cmd or .ps1, and that ought to work in this scenario.

I am not convinced that testing on Max+Ubuntu is adequate to ensure Windows compatibility 😆

vicb: > I am not convinced that testing on Max+Ubuntu is adequate to ensure Windows compatibility 😆

What I mean is that I don't really care about the dev version working on Windows because I don't think any use of dev on Windows. Wrangler is tested on windows which should be what matters.

But I'll see if there is something I can do without spending too much time.

petebacondarwin:
OK. So I forgot that we have this problem with when we bump the unenv version in Wrangler.
I see what you are trying to do now, which is to allow us to test bleeding edge unenv+workerd before we bump the version in Wrangler.

I think we should drop that extra build from the preset package build command, and allow the developer to run it independently if they want. Or if we want to actually have that as a test do the build as part of a test.

I still think we should try to test on Windows. While it might not be part of a CF developer's daily workflow, it would be annoying to find that an unenv preset change breaks on Windows only after we try to bump the preset version in Wrangler.

packages/unenv-preset/wrangler.json
petebacondarwin, line 1: Move this inside the `tests/worker` directory?

packages/unenv-preset/tests/index.test.ts
```diff
@@ -0,0 +1,62 @@
...
+			["--port=0", "--inspector-port=0"],
+			{
+				MINIFLARE_WORKERD_PATH,
+				WRANGLER_UNENV_RESOLVE_PATHS,
+			}
+		);
+	}, 30000);
+
+	afterAll(async () => {
+		await wrangler.stop();
```
petebacondarwin, line 53: It's possible that this could be undefined if `beforeAll()` fails? So perhaps

```suggestion
		await wrangler?.stop();
```

which should solve the types error?

fixtures/nodejs-hybrid-app/src/index.ts
```diff
@@ -29,8 +28,6 @@ export default {
 				return testX509Certificate();
 			case "/test-require-alias":
 				return testRequireUenvAliasedPackages();
-			case "/test-unenv-preset":
-				return await testUnenvPreset();
```
petebacondarwin, line 33: Given that the testing strategy is now to test local changes to the unenv preset in the package, it might be worth keeping this test here so that we also get test coverage on the published unenv preset?

petebacondarwin (approved):
A couple of suggestions.

packages/unenv-preset/tests/index.test.ts
```diff
@@ -0,0 +1,62 @@
...
+import { platform } from "node:process";
+import { fileURLToPath } from "node:url";
+import { fetch } from "undici";
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
+import { runWranglerDev } from "../../../fixtures/shared/src/run-wrangler-long-lived";
+import { TESTS } from "./worker/index";
+
+const pkgDir = path.resolve(fileURLToPath(import.meta.url), "../..");
+// workerd binary
+const workerdBin = platform === "win32" ? "workerd.ps1" : "workerd";
```
petebacondarwin, line 12: Perhaps .cmd instead?

```suggestion
const workerdBin = platform === "win32" ? "workerd.cmd" : "workerd";
```

Not sure if that is why the test is timing out or not...

packages/unenv-preset/wrangler.json
vicb: Good idea

packages/unenv-preset/tests/index.test.ts
vicb: I had `let wrangler: ... | undefined;` first but then `const { ip, port } = wrangler;` complains because it might be undefined.
Because tests will fail anyway after a timeout I left as it but let me know if you think I should change.

fixtures/nodejs-hybrid-app/src/index.ts
vicb: I think it makes sens to have preset related tests in the `@cloudflare/unenv-preset` and in a single location (I don't think we will think about updating 2 places especially because they should not be updated at the same time)

An improvement I want to do (as a follow up) is to test both the stable and dev versions from the preset package. This is why I asked James and Yagiz about figuring out the workerd version at runtime (to only run applicable tests). It sounds like it is not possible but we can still pass the version via the request.

packages/unenv-preset/tests/index.test.ts
vicb: We have 0 test in this repo with either .cmd or .ps1.

I'll try .cmd once and if it doesn't work, I'll skipIf(win32). Does that sound good?

packages/unenv-preset/tests/index.test.ts
vicb: I made the update

vicb:
@petebacondarwin thanks a lot for your patience and insights with this review. I ended up disabling windows tests (adding a comment with the link to the relevant Node issue). Let me know if there is more to do here.

🙏

anonrig (approved):

IgorMinar:
yay!

Issue #7588, opened 2024-12-18
🚀 Feature Request: Make CF-Connecting-IP header available in dev mode locally


godfat:
## Problem statement

The application is behind a reverse proxy, and it cannot access the client IP. All the application sees is the IP from `wrangler`, because it doesn't add `CF-Connecting-IP`, nor does it append the client IP to `X-Forwarded-For`.

## Feature request

* Have header `CF-Connecting-IP` available. Or:
* Append the client IP to `X-Forwarded-For`. Or:
* Have a way to access the client IP, for example, it can be done via `request.socket.remoteAddress`

## Background

We (GitLab) need a way to access the client IP when running `wrangler` for end-to-end tests. We do not want to deploy it to Cloudflare in order to run end-to-end tests.

This is needed for the end-to-end tests because the application which `wrangler` is talking to is behind another reverse proxy, and given this the application sees only the `wrangler` IP, not the actual client IP. This looks like this:

    client <-> wrangler <-> (optional nginx) <-> workhorse <-> application

We want wrangler to append the client IP in `X-Forwarded-For` header and configure `wrangler`'s IP as a trusted proxy, so that the application can find the real IP from `X-Forwarded-For`.

If `CF-Connecting-IP` is always available and can be trusted, then we can use that as well.

The main point is that we need to have a way to access the client IP, so that we can test features with regard to restricting IP. At the moment all IPs we can see are from `wrangler` itself, which means we cannot test this.

## Workaround

We can potentially run another Nginx in front of `wrangler` to append the IP to `X-Forwarded-For`, like this:

    client <-> nginx <-> wrangler <-> (optional nginx) <-> workhorse <-> application

But due to all the complexity involved with this, it would be nice to avoid needing to do this, just for accessing the client IP.

## References

* This issue described what we want to do: https://gitlab.com/gitlab-org/cells/http-router/-/issues/140
* This issue described the original problem we want to solve: https://gitlab.com/gitlab-org/gitlab/-/issues/499657

penalosa:
Once https://github.com/cloudflare/workers-sdk/pull/7702 is merged, `CF-Connecting-IP` will be available on Linux and macOS. Once https://github.com/cloudflare/workerd/issues/3310 is fixed in workerd, it should be available on Windows too—I'll close this issue once #7702 is merged, but feel free to comment on the workerd issue if Windows support is crucial.

godfat:
Terrific! Thank you so much!

PR #7702, opened 2025-01-08
Inject `CF-Connecting-IP` from workerd `clientIp`

penalosa:
Fixes CUSTESC-46864, fixes https://github.com/cloudflare/workers-sdk/issues/7588

When deployed, Workers can access the IP of the client that initiated a request using the `CF-Connecting-IP` header. However, this wasn't available locally, and so there was previously no way to know the IP address of a connecting client from a `wrangler dev` session. 

As it turns out, `workerd` injects a `clientIp` (that appears to be of the form `{ip}:{port}`— @danlapid or @kentonv could you confirm the format here?) using the `request.cf` property. Previously, the `request.cf` injected by `workerd` wouldn't make it to a user worker, because Miniflare injects a more fleshed out `request.cf` value that reflects what would be seen in production. As such, this PR extracts the `clientIp` from `workerd`'s `request.cf` and puts it in the `CF-Connecting-IP` header _before_ Miniflare injects it's own `request.cf` value.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: covered by Miniflare tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: matching (documented) production behaviour

### DIFF
--- a/packages/miniflare/test/index.spec.ts
+++ b/packages/miniflare/test/index.spec.ts
@@ -2631,7 +2631,7 @@ test("Miniflare: dispatchFetch() can override cf", async (t) => {
 test("Miniflare: CF-Connecting-IP is injected", async (t) => {
 	const mf = new Miniflare({
 		script:
-			"export default { fetch(request) { return new Response(request.headers.get('MF-Raw-workerd-CF')) } }",
+			"export default { fetch(request) { return new Response(request.headers.get('CF-Connecting-IP')) } }",
 		modules: true,
 		cf: {
 			myFakeField: "test",
@@ -2640,13 +2640,18 @@ test("Miniflare: CF-Connecting-IP is injected", async (t) => {
 	t.teardown(() => mf.dispose());
 
 	const ip = await mf.dispatchFetch("http://example.com/");
-	t.deepEqual(await ip.text(), "127.0.0.1");
+	// Tracked in https://github.com/cloudflare/workerd/issues/3310
+	if (!isWindows) {
+		t.deepEqual(await ip.text(), "127.0.0.1");
+	} else {
+		t.deepEqual(await ip.text(), "");
+	}
 });
 
 test("Miniflare: CF-Connecting-IP is injected (ipv6)", async (t) => {
 	const mf = new Miniflare({
 		script:
-			"export default { fetch(request) { return new Response(request.headers.get('MF-Raw-workerd-CF')) } }",
+			"export default { fetch(request) { return new Response(request.headers.get('CF-Connecting-IP')) } }",
 		modules: true,
 		cf: {
 			myFakeField: "test",
@@ -2656,7 +2661,13 @@ test("Miniflare: CF-Connecting-IP is injected (ipv6)", async (t) => {
 	t.teardown(() => mf.dispose());
 
 	const ip = await mf.dispatchFetch("http://example.com/");
-	t.deepEqual(await ip.text(), "::1");
+
+	// Tracked in https://github.com/cloudflare/workerd/issues/3310
+	if (!isWindows) {
+		t.deepEqual(await ip.text(), "::1");
+	} else {
+		t.deepEqual(await ip.text(), "");
+	}
 });
 
 test("Miniflare: CF-Connecting-IP is preserved when present", async (t) => {
--- a/packages/miniflare/test/index.spec.ts
+++ b/packages/miniflare/test/index.spec.ts
@@ -2610,6 +2610,39 @@ test("Miniflare: getCf() returns a user provided cf object", async (t) => {
 	t.deepEqual(cf, { myFakeField: "test" });
 });
 
+test("Miniflare: dispatchFetch() can override cf", async (t) => {
+	const mf = new Miniflare({
+		script:
+			"export default { fetch(request) { return Response.json(request.cf) } }",
+		modules: true,
+		cf: {
+			myFakeField: "test",
+		},
+	});
+	t.teardown(() => mf.dispose());
+
+	const cf = await mf.dispatchFetch("http://example.com/", {
+		cf: { myFakeField: "test2" },
+	});
+	const cfJson = (await cf.json()) as { myFakeField: string };
+	t.deepEqual(cfJson.myFakeField, "test2");
+});
+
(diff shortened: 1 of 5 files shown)


danlapid:
clientIp can be: "<ipv4>:<port>" or "[<ipv6>]:port" or "(inet_ntop error)" or "unix-abstract:<unix_socket_path>" or "unix:<unix_socket_path>" or "(unknown address family <address family int>)".
There are many cases here but I'm not sure all of them really need to be handled.
IMHO the logic should be:
1. Try to find ":" in the clientIp string (otherwise don't set CF-Connecting-IP at all)
2. Try to find "[" and "]" in the clientIp string split on ":"'s first part and if found set CF-Connecting-IP to be the ipv6 address between those
3. Try to find three dots in the clientIp string split on ":"'s second part and if found set CF-Connecting-IP to be the first part of the split.
4. Don't set CF-Connecting-IP.

(Notice header format in https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip-in-worker-subrequests for IPV6 does not contain square brackets).

danlapid (approved):
LGTM but I don't feel like I can comment on mini flare code 😅 
Up to you to decide if you want more eyes on this

petebacondarwin (approved):
LGTM - nice solution

packages/miniflare/test/index.spec.ts
```diff
@@ -2610,6 +2610,85 @@ test("Miniflare: getCf() returns a user provided cf object", async (t) => {
...
+
+	// Tracked in https://github.com/cloudflare/workerd/issues/3310
+	if (!isWindows) {
+		t.deepEqual(await ip.text(), "::1");
+	} else {
+		t.deepEqual(await ip.text(), "");
+	}
+});
+
+test("Miniflare: CF-Connecting-IP is preserved when present", async (t) => {
```
godfat, line 2673: Does this header come from the client?

If so, wouldn't this cause a security concern that this header can be spoofed (because we have to trust it)? I know that running `wrangler dev` is for the development anyway, but I don't feel too comfortable if this can cause a security concern even if it's just for local development (and running on CI without deploying to Cloudflare)

penalosa: It does come from the client, yeah—since this is intended to be used for local development I'm not sure there's a huge security concern, and I think the benefit of being able to mock this value might outweigh it.

In general, we don't recommend running a `wrangler dev` server exposed to the public internet, as it's not really designed for that purpose.

godfat: You're right and that makes sense. However, I still have concerns with regard to how GitLab might be using this. For now it's purely on CI, and it's probably not accessible from the public internet. However, this does increase the attack vector that if the client is compromised somehow, containing malicious code, it could expose some internal access.

Also, GitLab utilizes so called "Review Apps", which runs a GitLab instance for each merge requests (pull requests) requesting it. For now it does not run `wrangler` in any way, but if we need, we have to use `wrangler` because for that deployment process it cannot deploy to Cloudflare. This Review Apps are public accessible, which needs to be protected.

Is it possible to provide an option to toggle this behaviour? If so, we GitLab can toggle it so that we never trust the client's headers. Actually, I think it should be an option to allow the client to mock it, not the other way around, if an option is provided. The default should be secure, and I am open to have options for anything reasonable like mocking the IP under trusted environments.

What do you think?

Skye-31: Even if the machine that wrangler is running on is open to the internet, wrangler will listen on `127.0.0.1` by default, not `0.0.0.0`, so it won't be reachable unless you explicitly set it to be. 

> we have to use wrangler because for that deployment process it cannot deploy to Cloudflare

This change is also specific to wrangler dev, not deploying, so I'm not really sure what you mean by this

godfat: > Even if the machine that wrangler is running on is open to the internet, wrangler will listen on `127.0.0.1` by default, not `0.0.0.0`, so it won't be reachable unless you explicitly set it to be.

This is not about being reachable or not, it's about the client passing `CF-Connecting-IP: 127.0.0.1` (as an example) header, to pretend that the client IP is `127.0.0.1` but in fact it's something else. The server trusting `CF-Connecting-IP` would think that is true, and allow the client to access internal resources which should normally be blocked with the actual client IP.

To mitigate this, the header from the client should be completely ignored and dropped.

> This change is also specific to wrangler dev, not deploying, so I'm not really sure what you mean by this

I mean we have to run `wrangler` in some environments which should be secure, and not just for local development. We have to run `wrangler` because in those cases we cannot run them via Cloudflare. This means we kinda need to treat `wrangler` as a semi-production environment, which is also public facing.

penalosa: > I mean we have to run wrangler in some environments which should be secure, and not just for local development. We have to run wrangler because in those cases we cannot run them via Cloudflare. This means we kinda need to treat wrangler as a semi-production environment, which is also public facing.

The Wrangler/miniflare security model is not designed for being exposed on the public internet—they're tools intended for local development and testing. There will be other security concerns, but off the top of my head Wrangler exposes a devtools connection that allows for arbitrary code execution.

We can potentially add an option for dropping `CF-Connecting-IP`, but I'd strongly recommend against running `wrangler dev` in production and relying on this value for security checks to production resources.

> We have to run wrangler because in those cases we cannot run them via Cloudflare

Could you elaborate on this? Deploying the worker to Cloudflare sounds like the right solution for your use case.

godfat: I have a follow up question and it would be nice if you can answer that. (It's the same from Cloudflare support channel in the GitLab Slack workspace, posting here for transparency)

While I think it's not ideal that we cannot use `wrangler` in a secure way, I see it's not designed to be used to run securely. My follow up question will be, is there any plans to make it secure? While I understand it's only designed to be used as a local development tool, as we run more complex stuffs on it, I think it's important to have a way to run it securely. It doesn't need to be feature full like it is running on Cloudflare, but for end-to-end tests I think it'll be valuable.

PR #7081, opened 2024-10-24
Default `--x-registry` to on

penalosa:
Fixes #6955

Use the file based registry by default. It can still be disabled by setting `--x-registry=false` when running `wrangler dev`

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: existing tests should continue to pass
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: experimental flag

Issue #4633, opened 2023-12-19
🚀 Feature Request: Option to output esbuild metafile


janpio:
### Describe the solution

Esbuild's metafile is a great way to get additional information about a build, and for example use it in [Esbuild Size Analyzer](https://esbuild.github.io/analyze/) to understand the bundle size. If the CLI could "export" the Esbuild metafile it creates under the hood, that would be optimal.

As this is currently not implements, I quickly hacked together this almost-close-enough esbuild script that uses similar plugins. It brings most of my workers pretty close in size and code - so I am ok using this for now:

```ts
import { NodeGlobalsPolyfillPlugin } from '@esbuild-plugins/node-globals-polyfill'
import { NodeModulesPolyfillPlugin } from '@esbuild-plugins/node-modules-polyfill'
import { build } from 'esbuild'
import fs from 'fs'
import { relative, resolve } from "node:path";
import path from "path"


// ChatGPT
function isNodeCompat() {
  const filePath = 'wrangler.toml'; // Hardcoded file path as 'wrangler.toml'

  try {
    // Read the content of the file
    const data = fs.readFileSync(filePath, 'utf8');

    // Parse the content to extract variables
    const lines = data.split('\n');
    const variables = {};
    lines.forEach((line) => {
      const match = line.match(/^\s*([\w_]+)\s*=\s*(.+)/);
      if (match) {
        const key = match[1];
        const value = match[2].trim();
        variables[key] = value;
      }
    });

    // Check for the presence and value of node_compat
    return variables.node_compat && variables.node_compat.toLowerCase() === 'true';
  } catch (err) {
    console.error('Error reading the file:', err);
    return false;
  }
}

let plugins = [
  {
    name: "Copy wasm next to bundle",
    setup(build) {
      build.onResolve({ filter: /.*\.(wasm)$/ }, async (args) => {
        const to = resolve(path.join("dist-esbuild", args.path))
        fs.mkdirSync(path.dirname(to), { recursive: true })
        fs.copyFileSync(resolve(args.resolveDir, args.path), to)
        return {
          path: args.path,
          external: true
        };
      });
    },
  },
  // via https://github.com/cloudflare/workers-sdk/blob/3b5407a968189e60974233c5db8615162db37fd2/packages/wrangler/src/deployment-bundle/esbuild-plugins/cloudflare-internal.ts
  {
    name: "Cloudflare internal imports plugin",
    setup(pluginBuild) {
[...]

PR #7706, opened 2025-01-08
Remove express-based dev registry

penalosa:
After turning the file based registry on by default in #7081, usage of the legacy registry has been near 0 ([internal metrics](https://cflare.co/legacy-registry-metrics)). We've also had no recent bug reports related to the file based registry. As such, this PR removes the server based registry implementation.

Additionally, I did some deduping of dependencies (we previously had undici 5.28.4 & 5.28.2 included in the Wrangler bundle).

The result of both of the above is that Wrangler's bundle size is now just 4.4MB, down from the current 7.07MB 🎉 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: functionality is the same


GregBrimble (approved):
image

packages/wrangler/metafile.json
Skye-31, line 1: Is this file intentional?

DaniFoldi: if only this was https://github.com/cloudflare/workers-sdk/issues/4633 😅 for a second I thought that was finally implemented

edmundhung (approved):

PR #7720, opened 2025-01-10
chore(wrangler): use the unenv preset from `@cloudflare/unenv-preset`

vicb:
The cloudflare preset was migrated to this mono-repo (`@cloudflare/unenv-preset`)

This PR is switching to using this version

To do before merging:
- [x] diff the output of a wrangler build before/after
- [x] test the pre-release version of wrangler
- [ ] make sure the Vite plugin works with `@cloudflare/unenv-preset`
- [x] publish `@cloudflare/unenv-preset@1.0.0` and use that version

/cc @petebacondarwin @pi0 @dario-piotrowicz 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: preset is tested, will do a diff test
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no user facing change

PR #7721, opened 2025-01-10
chore(unenv-preset): Add .cjs output for the preset

vicb:
We do not need .cjs for the runtime files but `wrangler` requires the preset so we need to have a `dist/index.cjs`

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: tested via #7720
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: No user facing change

### DIFF
--- a/.changeset/shy-flies-float.md
+++ b/.changeset/shy-flies-float.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/unenv-preset": patch
+---
+
+chore(unenv-preset): Add .cjs output for the preset
--- a/packages/unenv-preset/build.config.ts
+++ b/packages/unenv-preset/build.config.ts
@@ -1,7 +1,13 @@
 import { defineBuildConfig } from "unbuild";
 
 export default defineBuildConfig({
+	declaration: true,
+	rollup: {
+		emitCJS: true,
+		cjsBridge: true,
+	},
 	entries: [
-		{ input: "src/", outDir: "dist/src", format: "esm", declaration: true },
+		"src/index",
+		{ input: "src/runtime/", outDir: "dist/runtime", format: "esm" },
 	],
 });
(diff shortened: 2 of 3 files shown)


petebacondarwin:
What do you mean "wrangler requires the preset"? 
I don't see that happening in the code on `main`. E.g. https://github.com/cloudflare/workers-sdk/blob/9d2740aa582c76040baf8aded1ac73d8bb2edeeb/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts#L4

petebacondarwin:
Or do you mean that if a library "requires" a Node.js module then our aliasing would result in the polyfilled module being required?

vicb:
> What do you mean "wrangler requires the preset"? I don't see that happening in the code on `main`. E.g.
> 
> https://github.com/cloudflare/workers-sdk/blob/9d2740aa582c76040baf8aded1ac73d8bb2edeeb/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts#L4
> 
> and also in the PR https://github.com/cloudflare/workers-sdk/pull/7720/files#diff-d5f05e17ffb161174d2bfea37bf23d39679290784e9aff28ac77bc06e3492459R3

Wrangler is compiled to cjs

https://github.com/cloudflare/workers-sdk/blob/54a84758ebb93765aea2744b4554efb048075dcb/packages/wrangler/scripts/bundle.ts#L38

petebacondarwin:
OMG that seems unnecesssary. Perhaps we could just change that?

vicb:
> OMG that seems unnecesssary. Perhaps we could just change that?

The PR could still be useful if we change that - _and I don't want to go down this rabbit hole today_

petebacondarwin (approved):

vicb:
Thanks @petebacondarwin

PR #7733, opened 2025-01-10
feat: pull resource names for provisioning from config if provided 

emily-shen:
Fixes DEVX-1556

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: still experimental

### DIFF
--- a/.changeset/smart-lions-suffer.md
+++ b/.changeset/smart-lions-suffer.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+feat: pull resource names for provisioning from config if provided
+
+Uses `database_name` and `bucket_name` for provisioning if specified. For R2, this only happens if there is not a bucket with that name already. Also respects R2 `jurisdiction` if provided.
--- a/packages/wrangler/src/d1/list.ts
+++ b/packages/wrangler/src/d1/list.ts
@@ -56,7 +56,7 @@ export const listDatabases = async (
 		);
 		page++;
 		results.push(...json);
-		if (limitCalls) {
+		if (limitCalls && page > 3) {
 			break;
 		}
 		if (json.length < pageSize) {
--- a/packages/wrangler/src/deployment-bundle/bindings.ts
+++ b/packages/wrangler/src/deployment-bundle/bindings.ts
@@ -13,7 +13,6 @@ import { createR2Bucket, getR2Bucket, listR2Buckets } from "../r2/helpers";
 import { isLegacyEnv } from "../utils/isLegacyEnv";
 import { printBindings } from "../utils/print-bindings";
 import type { Config } from "../config";
-import type { R2BucketInfo } from "../r2/helpers";
 import type { WorkerMetadataBinding } from "./create-worker-upload-form";
 import type {
 	CfD1Database,
@@ -129,7 +128,6 @@ export async function provisionBindings(
 		}
 	}
 
-	let preExistingR2: R2BucketInfo[] | undefined;
 	for (const r2 of bindings.r2_buckets ?? []) {
 		assert(typeof r2.bucket_name !== "symbol");
 		if (
@@ -240,7 +238,7 @@ export async function provisionBindings(
 			);
 		}
 		if (pendingResources.r2_buckets?.length) {
-			const preExisting = preExistingR2 ?? (await listR2Buckets(accountId));
+			const preExisting = await listR2Buckets(accountId);
 			await runProvisioningFlow(
 				pendingResources.r2_buckets,
 				preExisting.map((bucket) => ({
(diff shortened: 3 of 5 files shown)


packages/wrangler/src/d1/list.ts
```diff
@@ -56,7 +56,7 @@ export const listDatabases = async (
 		);
 		page++;
 		results.push(...json);
-		if (limitCalls) {
+		if (limitCalls && page > 3) {
```
emily-shen, line 59: tangential, but noticed that `d1 list` pagination seems to default to 10 per page, which seems quite low to me - is there a particular reason this is the case? KV is 100 per page for reference

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -151,9 +169,9 @@ export async function provisionBindings(
 			} else {
 				pendingResources.d1_databases?.push({
 					binding: d1.binding,
+					name: d1.database_name,
```
emily-shen, line 172: slight problem - if you have a db, specify the name in config and provision, and then you change the name in your config, it will still inherit the first binding. the settings api call doesn't include the db name, only the db ID and binding, so we can't really tell when this has changed.

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -124,23 +127,38 @@ export async function provisionBindings(
...
-						r2.bucket_name = bucketName;
-					},
-				});
+		assert(typeof r2.bucket_name !== "symbol");
+		if (inBindingSettings(settings, "r2_bucket", r2.binding, r2.bucket_name)) {
+			// does not inherit if the bucket name has changed
+			r2.bucket_name = INHERIT_SYMBOL;
+		} else {
+			if (r2.bucket_name) {
+				preExistingR2 ??= await listR2Buckets(accountId);
```
penalosa, line 138: I'm a tad concerned about this—what if the user has 100s of R2 buckets? Is there a get bucket endpoint or something we could call to more explicitly check if this bucket exists?

packages/wrangler/src/deployment-bundle/bindings.ts
penalosa: Could we call the D1 API to check specifically if this D1 database exists?

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -219,11 +237,19 @@ export async function provisionBindings(
 function inBindingSettings<Type extends WorkerMetadataBinding["type"]>(
 	settings: Settings | undefined,
 	type: Type,
-	bindingName: string
+	bindingName: string,
+	bucket_name?: string | undefined
```
penalosa, line 241: Could this be a more general ID matcher function so it worked for more than just R2?

emily-shen: [2ed581f](https://github.com/cloudflare/workers-sdk/pull/7733/commits/2ed581fb60faea21bed25f85f91eae7dc5efa3cf) - types are a bit gross though :/

.changeset/smart-lions-suffer.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": patch
```
penalosa, line 2:
```suggestion
"wrangler": minor
```

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -198,7 +240,7 @@ export async function provisionBindings(
 			);
 		}
 		if (pendingResources.r2_buckets?.length) {
-			const preExisting = await listR2Buckets(accountId);
+			const preExisting = preExistingR2 ?? (await listR2Buckets(accountId));
```
penalosa, line 243: I think `preExistingR2` is always undefined now?

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -268,67 +321,90 @@ async function runProvisioningFlow(
...
-				selected = await select(
-					`Would you like to connect an existing ${friendlyBindingName} or create a new one?`,
-					{
-						choices: options.concat([{ title: "Create new", value: "new" }]),
-						defaultOption: options.length,
+			if (name) {
+				logger.log("Resource name found in config:", name);
+				const foundResourceId = preExisting.find(
+					(r) => r.title === name
+				)?.value;
```
penalosa, line 331: What case is this catering for? Surely if the item has a name (D1 or R2) which maps to an existing resource we should have already bound to it further up? Why do we need to ask the user?

penalosa (approved):
Couple comments, but in general lgtm

.changeset/smart-lions-suffer.md
emily-shen: think it should be a patch because this is experimental?

PR #7735, opened 2025-01-10
Unwrap SDW errors before sending to Sentry

penalosa:
This should make Sentry errors more actionable, by unwrapping error `cause` (which was causing all the `captureGlobalException` issues)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: some existing tests cover sentry, and this isn't really testable
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/honest-spies-prove.md
+++ b/.changeset/honest-spies-prove.md
@@ -2,4 +2,4 @@
 "wrangler": patch
 ---
 
-Send error cause to Sentry when available
+Unwrap the error cause when available to send to Sentry
--- a/.changeset/honest-spies-prove.md
+++ b/.changeset/honest-spies-prove.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Send error cause to Sentry when available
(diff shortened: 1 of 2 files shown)


packages/wrangler/src/index.ts
```diff
@@ -1179,17 +1180,18 @@ export async function main(argv: string[]): Promise<void> {
 			}
 		}
 
+		const reportableError = loggableException ?? e;
```
petebacondarwin, line 1183: I'm not sure this line is really needed. Above you assign `e` to `loggableError already and only reassign (to `e.cause`) if that is itself an `Error` instance.  It think it is enough to use `loggableError` below rather than `reportableError`.

.changeset/honest-spies-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Send error cause to Sentry when available
```
petebacondarwin, line 5:
```suggestion
Unwrap the error cause when available to send to Sentry
```

petebacondarwin (approved):

CarmenPopoviciu (approved):

PR #7742, opened 2025-01-12
chore: update wrangler versions in templates

threepointone:
We should probably automate this, but doing it manually for now since it's so behind. 

Fixes #0000

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no functional change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no functional change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no functional change

### DIFF
--- a/packages/create-cloudflare/templates-experimental/hello-world-durable-object-with-assets/js/package.json
+++ b/packages/create-cloudflare/templates-experimental/hello-world-durable-object-with-assets/js/package.json
@@ -8,6 +8,6 @@
 		"start": "wrangler dev"
 	},
 	"devDependencies": {
-		"wrangler": "^3.60.3"
+		"wrangler": "^3.101.0"
 	}
 }
--- a/packages/create-cloudflare/templates/hello-world-durable-object/ts/src/index.ts
+++ b/packages/create-cloudflare/templates/hello-world-durable-object/ts/src/index.ts
@@ -14,7 +14,7 @@ import { DurableObject } from "cloudflare:workers";
  */
 
 /** A Durable Object's behavior is defined in an exported Javascript class */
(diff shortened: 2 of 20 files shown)


packages/create-cloudflare/templates/hello-world-durable-object/ts/src/index.ts
```diff
@@ -14,7 +14,7 @@ import { DurableObject } from "cloudflare:workers";
  */
 
 /** A Durable Object's behavior is defined in an exported Javascript class */
-export class MyDurableObject extends DurableObject {
+export class MyDurableObject extends DurableObject<Env> {
```
threepointone, line 17: also added this fix here

petebacondarwin: Doh! I see now. This is imported from the `cloudflare:workers` module, which is what brings in this class.

petebacondarwin (approved):

packages/create-cloudflare/templates-experimental/hello-world-durable-object-with-assets/js/package.json
```diff
@@ -8,6 +8,6 @@
 		"start": "wrangler dev"
 	},
 	"devDependencies": {
-		"wrangler": "^3.60.3"
+		"wrangler": "^3.101.0"
```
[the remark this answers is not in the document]
threepointone: Considering it more of a cleanliness thing really. Looks bad if our own official templates don't point to the latest version. Ideally this would be automated.

PR #7756, opened 2025-01-14
generate types for the next experimental template

vicb:
Instead of copying an empty `cloudflare-env.d.ts`, the file is generated when the template is configured.

Then it contains the existing bindgins:

```ts
// Generated by Wrangler by running `wrangler types --env-interface CloudflareEnv cloudflare-env.d.ts`

interface CloudflareEnv {
	NEXTJS_ENV: string;
	ASSETS: Fetcher;
}
```

Notes:

- I've moved the call to `transformPackageJson` before `configure` so that the latest can execute the added scripts
- If this get merged, we probably should update other templates (we should check `usesTypescript(...)` for template supporting JS).

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: already tested & checked locally
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not user facing change


packages/create-cloudflare/templates-experimental/next/c3.ts
```diff
@@ -27,6 +32,14 @@ const configure = async () => {
 		startText: "Adding the Cloudflare adapter",
 		doneText: `${brandColor(`installed`)} ${dim(packages.join(", "))}`,
 	});
+
+	const { npm } = detectPackageManager();
+	await runCommand([npm, "run", NPM_TYPE_GEN_SCRIPT], {
```
petebacondarwin, line 37: I feel like a simpler change that doesn't require changes to how C3 hooks are run, is just to put the command that generates the types into a variable. Then run it here and use it to update the package.json below.

E.g.

```suggestion
  const GENERATE_TYPES = ["wrangler", "types", "--env-interface", "CloudflareEnv", "cloudflare-env.d.ts"];
	await runCommand([npx, ...GENERATE_TYPES], {
```

packages/create-cloudflare/templates-experimental/next/c3.ts
```diff
@@ -49,7 +62,7 @@ export default {
 		scripts: {
 			deploy: `opennextjs-cloudflare && wrangler deploy`,
 			preview: `opennextjs-cloudflare && wrangler dev`,
-			"cf-typegen": `wrangler types --env-interface CloudflareEnv cloudflare-env.d.ts`,
+			[NPM_TYPE_GEN_SCRIPT]: `wrangler types --env-interface CloudflareEnv cloudflare-env.d.ts`,
```
petebacondarwin, line 65:
```suggestion
			"cf-typegen: GENERATE_TYPES.join(" "),
```

petebacondarwin (changes requested):
Can we do this more simply?

packages/create-cloudflare/templates-experimental/next/c3.ts
vicb: That would work too but I prefer using the actual script.

Do you think that calling `updatePackageScripts()` before `configure` might ever cause issues?

petebacondarwin: I don't see the benefit in making this change to how C3 hooks run when a simple change to the c3.ts file would suffice. The update to the package.json is simply to make user facing scripts ready to go, not to support internal scripts.

vicb: 1) What I mentionned

https://github.com/cloudflare/workers-sdk/blob/df0e5bef817c09754471ac4842531a1e14e5576a/packages/create-cloudflare/src/frameworks/index.ts#L31-L35

> The update to the package.json is simply to make user facing scripts ready to go, not to support internal scripts.

the `.d.ts` is supposed to be generated/updated by running the "user facing script", that is what this PR implements.

petebacondarwin: That comment is not relevant here, I believe. We are not running the CLI generator script in this command, which has to be run outside of a node package.
Here we are just running `wrangler types` command inside the package, which works fine with yarn.

PR #7761, opened 2025-01-14
Add jaeger bindings to asset-worker

WillTaylorDev:
This commit adds Jaeger tracing to asset-worker, in order for us to be able to have better insight into failures from KV.

---

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/workers-shared/asset-worker/src/index.ts
+++ b/packages/workers-shared/asset-worker/src/index.ts
@@ -7,10 +7,15 @@ import { applyConfigurationDefaults } from "./configuration";
 import { decodePath, getIntent, handleRequest } from "./handler";
 import { InternalServerErrorResponse } from "./responses";
 import { getAssetWithMetadataFromKV } from "./utils/kv";
-import type { AssetConfig, UnsafePerformanceTimer } from "../../utils/types";
+import { mockJaegerBinding } from "./utils/mocks";
+import type {
+	AssetConfig,
+	JaegerTracing,
+	UnsafePerformanceTimer,
+} from "../../utils/types";
 import type { ColoMetadata, Environment, ReadyAnalytics } from "./types";
 
-type Env = {
+export type Env = {
 	/*
 	 * ASSETS_MANIFEST is a pipeline binding to an ArrayBuffer containing the
 	 * binary-encoded site manifest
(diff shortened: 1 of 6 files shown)


packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -56,6 +62,12 @@ export default class extends WorkerEntrypoint<Env> {
 		const startTimeMs = performance.now();
 
 		try {
+			if (!this.env.JAEGER) {
+				console.log("USING MOCK");
```
WalshyDev, line 66: test log

packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -89,12 +101,21 @@ export default class extends WorkerEntrypoint<Env> {
 				});
 			}
 
-			return handleRequest(
-				request,
-				config,
-				this.unstable_exists.bind(this),
-				this.unstable_getByETag.bind(this)
-			);
+			return await this.env.JAEGER.enterSpan("handleRequest", (span) => {
```
WalshyDev, line 104: any reason to await and not async like it was before?

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -17,8 +18,14 @@ describe("[Asset Worker] `handleRequest`", () => {
...
 		});
 
+		const mockEnv = {
+			JAEGER: mockJaegerBinding(),
+		};
+
 		const response = await handleRequest(
 			new Request("https://example.com/"),
+			// @ts-expect-error Only jaeger
+			mockEnv,
```
WalshyDev, line 28: we could just pass empty object now

packages/workers-shared/asset-worker/src/index.ts
WillTaylorDev: None, just oversight!

packages/workers-shared/asset-worker/tests/handler.test.ts
WillTaylorDev: We do the override in index.ts unfortunately; this is calling handleRequest which doesn't have the env override for jaeger.

WalshyDev (approved):

PR #7769, opened 2025-01-14
Refactor pipelines commands into separate files, rename some options

cmackenzie1:
**Pipelines is currently in closed beta, renaming these without providing aliases _should_ have limited to zero impact**

Moving each of the subcommand handlers and options for each subcommand into its own file. This also renames many of the parameters to be more specific.

The following parameters have been renamed:

| Previous Name | New Name |
| ---- | ---- |
| access-key-id | r2-access-key-id |
| secret-access-key | r2-secret-access-key |
| transform | transform-worker |
| r2 | r2-bucket |
| prefix | r2-prefix |
| binding | enable-worker-binding |
| http | enable-http |
| authentication | require-http-auth |
| filename | file-template |
| filepath | partition-template |

Adds the following new option for `create` and `update` commands:

```
--cors-origins           CORS origin allowlist for HTTP endpoint (use * for any origin)  [array]
```

Closes https://jira.cfdata.org/browse/PIPE-160.

---

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
  - [X] Not required because: Unit tests cover the change
- Public documentation
  - [ ] TODO (before merge)
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18595

### DIFF
--- a/.changeset/plenty-birds-melt.md
+++ b/.changeset/plenty-birds-melt.md
@@ -2,7 +2,7 @@
 "wrangler": patch
 ---
 
-Adds the following new option for `create` and `update` commands:
+Adds the following new option for `wrangler pipelines create` and `wrangler pipelines update` commands:
 
 ```
 --cors-origins           CORS origin allowlist for HTTP endpoint (use * for any origin)  [array]
--- a/.changeset/plenty-birds-melt.md
+++ b/.changeset/plenty-birds-melt.md
@@ -1,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+Adds the following new option for `create` and `update` commands:
+
+```
+--cors-origins           CORS origin allowlist for HTTP endpoint (use * for any origin)  [array]
+```
--- a/.changeset/short-kids-hunt.md
+++ b/.changeset/short-kids-hunt.md
@@ -1,5 +1,5 @@
 ---
-"wrangler": minor
+"wrangler": patch
 ---
 
 Rename wrangler pipelines <create|update> flags
--- a/.changeset/short-kids-hunt.md
+++ b/.changeset/short-kids-hunt.md
@@ -1,0 +1,26 @@
+---
+"wrangler": minor
+---
+
+Rename wrangler pipelines <create|update> flags, add `--cors-origins` flag
+
+The following parameters have been renamed:
+
+| Previous Name     | New Name              |
+| ----------------- | --------------------- |
+| access-key-id     | r2-access-key-id      |
+| secret-access-key | r2-secret-access-key  |
+| transform         | transform-worker      |
+| r2                | r2-bucket             |
+| prefix            | r2-prefix             |
+| binding           | enable-worker-binding |
+| http              | enable-http           |
+| authentication    | require-http-auth     |
+| filename          | file-template         |
+| filepath          | partition-template    |
+
+Adds the following new option for `create` and `update` commands:
+
+```
+--cors-origins           CORS origin allowlist for HTTP endpoint (use * for any origin)  [array]
+```
--- a/.changeset/short-kids-hunt.md
+++ b/.changeset/short-kids-hunt.md
@@ -2,7 +2,7 @@
 "wrangler": minor
 ---
 
-Rename wrangler pipelines <create|update> flags, add `--cors-origins` flag
+Rename wrangler pipelines <create|update> flags
 
 The following parameters have been renamed:
 
@@ -18,9 +18,3 @@ The following parameters have been renamed:
 | authentication    | require-http-auth     |
 | filename          | file-template         |
 | filepath          | partition-template    |
-
-Adds the following new option for `create` and `update` commands:
-
-```
---cors-origins           CORS origin allowlist for HTTP endpoint (use * for any origin)  [array]
-```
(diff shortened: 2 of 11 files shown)


packages/wrangler/src/pipelines/cli/list.ts
```diff
@@ -0,0 +1,28 @@
+import { readConfig } from "../../config";
+import { logger } from "../../logger";
+import * as metrics from "../../metrics";
+import { requireAuth } from "../../user";
+import { listPipelines } from "../client";
+import type { CommonYargsOptions } from "../../yargs-types";
+import type { ArgumentsCamelCase } from "yargs";
+
+export async function listPipelinesHandler(
```
CarmenPopoviciu, line 9: shouldn't this fn call `await printWranglerBanner();` as well?

cmackenzie1: I am not sure, I copied it directly from https://github.com/cloudflare/workers-sdk/pull/7769/files#diff-bd3457733e5c7c3cdc0ea531bc35fc5858ec21ea84e1f65a11bee0f222efd727L330-L347.

Should all commands call `printWranglerBanner`?

cmackenzie1: Added it.

maheshwarip (approved):

.changeset/short-kids-hunt.md
```diff
@@ -0,0 +1,26 @@
+---
+"wrangler": minor
+---
+
+Rename wrangler pipelines <create|update> flags, add `--cors-origins` flag
```
andyjessop, line 5: I think ideally this is two separate changesets.

packages/wrangler/src/__tests__/pipelines.test.ts
```diff
@@ -283,47 +284,70 @@ describe("pipelines", () => {
...
+				      --require-http-auth      Require Cloudflare API Token for HTTPS endpoint authentication  [boolean] [default: false]
+				      --cors-origins           CORS origin allowlist for HTTP endpoint (use * for any origin)  [array]
+
+				Batch hints:
+				      --batch-max-mb       Maximum batch size in megabytes before flushing  [number]
+				      --batch-max-rows     Maximum number of rows per batch before flushing  [number]
+				      --batch-max-seconds  Maximum age of batch in seconds before flushing  [number]
+
+				Transformations:
+				      --transform-worker  PipelineTransform worker and entrypoint (<worker>.<entrypoint>)  [string]
```
andyjessop, line 299:
```suggestion
				      --transform-worker  PipelineTransform Worker and entrypoint (<worker>.<entrypoint>)  [string]
```

packages/wrangler/src/pipelines/cli/create.ts
```diff
@@ -0,0 +1,316 @@
...
+				demandOption: false,
+				coerce: validateInRange("batch-max-seconds", 1, 300),
+			})
+
+			// Transform options
+			.group(["transform-worker"], "Transformations:")
+			.option("transform-worker", {
+				type: "string",
+				describe:
+					"PipelineTransform worker and entrypoint (<worker>.<entrypoint>)",
```
andyjessop, line 97:
```suggestion
					"PipelineTransform Worker and entrypoint (<worker>.<entrypoint>)",
```

andyjessop (commented):
I have no comments on the code, as it looks like the functions have just been moved to separate files. The only thing it would be good to address here is the consistency in naming "Pipelines" vs "pipelines", "Workers" vs "workers". I believe we should always capitalise product names.

.changeset/plenty-birds-melt.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+Adds the following new option for `create` and `update` commands:
```
edmundhung, line 5: How about this? Just to make sure people know this is about the pipeline commands 

```suggestion
Adds the following new option for `wrangler pipelines create` and `wrangler pipelines update` commands:
```

packages/wrangler/src/pipelines/validate.ts
```diff
@@ -0,0 +1,36 @@
+export function validateName(label: string, name: string) {
+	if (!name.match(/^[a-zA-Z0-9-]+$/)) {
+		throw new Error(`Must provide a valid ${label}`);
```
edmundhung, line 3: It would be better to fail these check with an UserError. We used it to differentiate whether it is a possible bug on our side or if it is simply a user error.

edmundhung (commented):
Looks good to me overall! Just a few minor suggestions.

Also, we’re migrating existing commands to a new setup as documented [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/CONTRIBUTING.md#defining-your-new-command-in-wrangler). This isn’t blocking for this change, but just an FYI. 👍🏼

.changeset/short-kids-hunt.md
```diff
@@ -0,0 +1,20 @@
+---
+"wrangler": minor
```
edmundhung, line 2: We will mark every changes as patch as Pipeline is still in closed beta.
```suggestion
"wrangler": patch
```

edmundhung (approved):
Looks good to me.

.changeset/short-kids-hunt.md
cmackenzie1: Sounds good. Thanks @edmundhung !

PR #7773, opened 2025-01-15
fix: make sure asset routing config is passed to vitest 

emily-shen:
Fixes  WC-3076

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e's for this
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/vitest-pool-workers/src/pool/config.ts
+++ b/packages/vitest-pool-workers/src/pool/config.ts
@@ -221,6 +221,7 @@ async function parseCustomPoolOptions(
 		options.miniflare.assets = {
 			...options.miniflare.assets,
 			routingConfig: {
+				...options.miniflare.assets.routingConfig,
 				has_user_worker: Boolean(options.main),
 			},
 		};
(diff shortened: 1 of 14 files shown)


packages/vitest-pool-workers/src/pool/config.ts
```diff
@@ -221,6 +221,7 @@ async function parseCustomPoolOptions(
 		options.miniflare.assets = {
 			...options.miniflare.assets,
 			routingConfig: {
+				...options.miniflare.assets.routingConfig,
 				has_user_worker: Boolean(options.main),
 			},
 		};
```
edmundhung, line 227: Nit: Would it be simpler to modify the config directly?, e.g.

```suggestion
		options.miniflare.assets.routingConfig ??= {};
		options.miniflare.assets.routingConfig.has_user_worker = Boolean(options.main);
```

edmundhung (commented):
Should we add a test to the vitest-pool-workers fixture?

packages/vitest-pool-workers/src/pool/config.ts
emily-shen: 6a110217c546b6033b5ad939cb48a69c2bcd104e

edmundhung (approved):

PR #7676, opened 2025-01-06
Output `.json` by default in C3

penalosa:
Fixes DEVX-1491

Generate all new projects from C3 with a `wrangler.json` file (cc @dario-piotrowicz), as well as simplifying the included comments (cc @vicb I know you had thoughts around this).

I did the TOML -> JSON conversion by running (in `packages/create-cloudflare`):

```sh
for file in **/wrangler.toml; do       
    (toml2json "$file" | jq '{ name: .name, main: .main, compatibility_date: .compatibility_date } + .') > "${file%.toml}.json"
done
```

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: C3 E2E will run without the label
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/19125
  - [ ] Documentation not necessary because:

PR #7774, opened 2025-01-15
fix(create-cloudflare): Fix regression in C3's next template

### LINKED ISSUE #7770: 🐛 BUG: create-cloudflare@2.37.0, Output the wrangler.json config format when running create-cloudflare. But when generating the next template, the wrangler.toml file will be read.

C3 (npm create cloudflare)

2.37.0[create-cloudflare]

Mac Sonoma 14.1.1

_No response_

### Describe the Bug

Version: create-cloudflare@2.37.0.
Output the wrangler.json config format when running create-cloudflare, but when generating the next template, the wrangler.toml file will be read.
[https://github.com/cloudflare/workers-sdk/blob/main/packages/create-cloudflare/templates/next/c3.ts](https://github.com/cloudflare/workers-sdk/blob/main/packages/create-cloudflare/templates/next/c3.ts)
`
const generate = async (ctx: C3Context) => {
	const projectName = ctx.project.name;

	await runFrameworkGenerator(ctx, [projectName]);

	const wranglerToml = readFile(join(getTemplatePath(ctx), "wrangler.toml"));

	// Note: here we add `# KV Example:` to the toml file for the KV example, we don't actually
	//       include the comment in the template wrangler.toml file just so to keep it identical
	//       and consistent with that of all the other frameworks
	//       (instead of making it a special case which needs extra care)
	const newTomlContent = wranglerToml.replace(
		/#\s+\[\[kv_namespaces\]\]\n#\s+binding\s+=\s+"MY_KV_NAMESPACE"\n#\s+id\s+=\s+"[a-zA-Z0-9]+?"/,
		($1) => `# KV Example:\n${$1}`,
	);

	if (!/# KV Example/.test(newTomlContent)) {
		// This should never happen to users, it is a check mostly so that
		// if the toml file is changed in a way that breaks the "KV Example" addition
		// the C3 Next.js e2e runs will fail with this
		throw new Error("Failed to properly generate the wrangler.toml file");
	}

	writeFile(join(ctx.project.path, "wrangler.toml"), newTomlContent);

	updateStatus("Created wrangler.toml file");
};

`

╰  ERROR  Error: Error: ENOENT: no such file or directory, open '/Users/user/.npm/_npx/971abe9c5e1ca794/node_modules/create-cloudflare/templates/next/wrangler.toml'

CarmenPopoviciu:
[#7676](https://github.com/cloudflare/workers-sdk/pull/7676) switched C3 templates to default to `wrangler.json` instead of `wrangler.toml`. Unfortunately, this unintendedly broke the `next` template, which was still attempting to read `wrangler.toml` specifically.

This commit fixes the regression.

Fixes #7770

These changes were also tested using a test project and the C3 pre-release url . See snapshot of generated readme file 👇 

Screenshot 2025-01-15 at 23 42 32

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: regression fix


packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -475,7 +475,6 @@ function getFrameworkTests(opts: {
 					},
 				],
 				testCommitMessage: true,
-				quarantine: true,
```
CarmenPopoviciu, line 478: we should start adding comments when we quarantine tests

packages/create-cloudflare/templates/next/README.md
```diff
@@ -47,10 +47,13 @@ In order to enable the example:
   // KV Example:
   ```
   and uncomment the commented lines below it (also uncomment the relevant imports).
-- Do the same in the `wrangler.toml` file, where
-  the comment is:
+- In the `wrangler.json` file search for the block below, and uncomment uncomment the line that contains the `"kv_namespaces"` declaration:
```
CarmenPopoviciu, line 50: to keep things consistent with the rest of the comments in the config file, I added the KV comment in a similar format. IMHO, it made no sense to add the extra `// KV example` line, so I didn't, and modified the readme instead. Lemme know if anyone strongly disagrees

packages/create-cloudflare/e2e-tests/frameworks.test.ts
dario-piotrowicz: we should also evaluate why tests still need to be quarantined (maybe an investigation + adding respective comments) would be a good start 🙂

emily-shen (approved):
Not blocking this PR, but there's also a bunch of other places that use `wrangler.toml`, I think next is just more commonly use so it's come up here.
I think wrangler.toml is still used in 
- hello world + hello world with assets vitest config
- download pre-existing worker from dash
- comment on the worker-configuration.d.ts files

sorry should've caught these while reviewing the original PR!

packages/create-cloudflare/e2e-tests/frameworks.test.ts
dario-piotrowicz: PS: as far as I can tell quarantined tests get forgotten and their templates can completely break without anyone noticing, I wonder if we could increase the visibility of these too

dario-piotrowicz: (that's actually what happened here isn't it? 🤔 next broke but no one noticed right away since it's quarantined)

packages/create-cloudflare/templates/next/README.md
dario-piotrowicz, line 50:
```suggestion
- In the `wrangler.json` file search for the block below, and uncomment the line that contains the `"kv_namespaces"` declaration:
```

packages/create-cloudflare/e2e-tests/frameworks.test.ts
CarmenPopoviciu: that is correct yes. and I agree that we should evaluate what tests are quarantined and why. I took a quick look and both `astro` and `analog` tests are quarantined. I have context over the `analog` ones but not the `astro` ones

packages/create-cloudflare/templates/next/README.md
dario-piotrowicz: I don't know if adding the `kv_namespaces` only for next makes sense... (why KV is the only binding included? what about DOs, R2, etc... why the preferential treatment?) as we discussed maybe we should just get rid of the KV example in the next template instead 🤔 

I'm happy to progress with this solution but (maybe as a followup) I think removing the example would be better

packages/create-cloudflare/src/wrangler/__tests__/config.test.ts
```diff
@@ -83,6 +83,12 @@ describe("update wrangler config", () => {
 			# [[services]]
 			# binding = "MY_SERVICE"
 			# service = "my-service"
+
+			# KV Namespaces
```
dario-piotrowicz, line 87: shouldn't this be under the bindings section above? (line 61) 🤔

CarmenPopoviciu: oh...but I see now what u mean...I'll fix

packages/create-cloudflare/templates/next/README.md
CarmenPopoviciu: > why KV is the only binding included? what about DOs, R2, etc... why the preferential treatment?

I totally get what you mean. But then, wasn't this question equally relevant in the prev version. Why did nextjs project ship only with a KV example? I have no context whatsoever as to why/how this decision was taken. 

Don't get me wrong, I don't mind removing it, but then my question is, does C3 come with anything similar out of the box...as in an example or template that exemplify how such bindings are set. Does it need to? Arguably, these are all larger prod questions

dario-piotrowicz: > I totally get what you mean. But then, wasn't this question equally relevant in the prev version?

yeah I agree with you, this was basically the same before, the Next.js template being the only framework one having an extra KV example 👍 

I'm just bringing it up because before it was less expensive to have this divergence (basically a single extra comment in the shared toml file) while now it seems like it requires us to add the KV declaration comment to all wrangler configs, given this extra downside maybe it's no longer worth it

> Why did the nextjs starter ship only with a KV example?

We discussed it on the frameworks team and concluded that it'd be nice to show such an example to make sure that people knew how to access bindings in their Next.js app, that's all

>  does C3 come with anything similar out of the box?

I'd have to check, I am quite sure the Next.js one is the only fullstack template with this, as for workers I would imagine that they also have extra configs as well

> Don't get me wrong, I don't mind removing it

I don't particularly mind if you keep it in, I just personally think that the extra (even if very small) complexity + having the KV for all template configs makes the example not that KV Next.js example worth keeping, but it's also not a big deal 🙂

dario-piotrowicz:
It's pretty minor and could be done in a followup PR, but if you follow the readme instructions:
https://github.com/cloudflare/workers-sdk/blob/99f27df059c211428d4cf3cc07bb99c164d22369/packages/create-cloudflare/templates/next/README.md?plain=1#L40-L64

things don't really work well as a comma is missing in the `wrangler.json` 😓 

Screenshot 2025-01-15 at 17 03 27

People should not get too confused by this but to fix it we should add a comma here (I think?):
https://github.com/cloudflare/workers-sdk/blob/99f27df059c211428d4cf3cc07bb99c164d22369/packages/create-cloudflare/src/wrangler/config.ts#L57
(or just mention in the readme to add the comma?)

dario-piotrowicz:
I have no idea what's going on, but also `next dev` doesn't seem to work! 😨 
(which I think is unrelated to this PR/regression)

Screenshot 2025-01-15 at 17 11 17

dario-piotrowicz (approved):
Looks good to me, thanks @CarmenPopoviciu 😄

packages/create-cloudflare/templates/next/README.md
CarmenPopoviciu: > while now it seems like it requires us to add the KV declaration comment to all wrangler configs

I am 100% with you on this! And tbh, I do have the same concern as you. On the other hand, I 100% believe that the KV example is valuable for our users, which is why I'm holding on to it with my dear life 😆 . IMHO if it's not meant to live in the starter project it should def live somewhere else...tho the starter is *such* a great place, because it requires zero context switch on the dev's side

> We discussed it on the frameworks team and concluded that it'd be nice to show such an example to make sure that people knew how to access bindings in their Next.js app, that's all

yes 💯  this!!!!

having said all this I feel like there is not enough consensus (wrt what the right fix is) to land this PR just yet. How about if I just revert the `next` template back to `wrangler.toml`. The remix template is still using a `.toml` config file, so it wouldn't be the only exception. Just so we can bump the C3 version with a fix. And in the meanwhile, we can iterate this PR without the pressure of a fix release on our back

CarmenPopoviciu: oh wait...we actually can't revert back to `wrangler.toml` because of how we now standardise comments. NVM me

dario-piotrowicz: Nono, I am totally happy to go with the current solution! we can always re-evaluate later! no big deal 🙂

CarmenPopoviciu:
just for posterity...

1. https://github.com/cloudflare/workers-sdk/pull/7774#discussion_r1916824409
- there was some back forth in this conversation, rightfully so, about whether this PR fixes things in a way the we want things to be moving forward. Landing the PR as it was at the time of this conversation, would also have left a couple of loose ends that would have had to be addressed in follow-up PRs. Therefore we pulled in prod, and asked them to take a decision, since this felt very much like a prod decision to take. The solution we landed, which is reflected in the current state of this PR, was to keep the `next` template as an exception, in terms of providing that KV example, but instead of the template relying on particular commented code being present in the config file, we instead changed the readme to instruct users to add the necessary config configuration to the distributed config file. This means that we ultimately didn't have to pollute the config file for *all* C3 templates, with comments about KV bindings and such.

2. https://github.com/cloudflare/workers-sdk/pull/7774#issuecomment-2593489357
- because of the adopted solution, this is no longer a point of concern

3. https://github.com/cloudflare/workers-sdk/pull/7774#pullrequestreview-2553023270
- we went ahead and solved this as well. In theory, there should be no lingering references of `wrangler.toml` in the C3 templates, with the exception of Remix, which for reasons unknown to me, was not yet ported to `wrangler.json`

andyjessop (approved):
Really nice investigation and PR. Thanks so much for this 🚀

PR #7639, opened 2024-12-27
fix(workers-playground): sends raw request method through the `X-CF-HTTP-Method` header

edmundhung:
Fixes [DEVX-1449](https://jira.cfdata.org/browse/DEVX-1449).

This PR utilizes the `X-CF-HTTP-Method` header introduced on #7630 and is **blocked** until the relevant change is released.

You can verify this change on the [preview environment](https://d42f344a.workers-playground.pages.dev/playground) following these steps:
[...]

PR #7793, opened 2025-01-16
fix(playground-preview-worker,edge-preview-authenticated-proxy): ensure no body is passed when constructing a GET or HEAD request to the preview worker

edmundhung:
Fixes #7791, #7630

This PR fixes the following error we see after releasing the change on #7639.

> Error: Request with a GET or HEAD method cannot have a body.

The implementation has been updated in both `playground-preview-worker` and `edge-preview-authenticated-proxy`, which are used by the worker playground and the dashboard.

**Note**: This change itself will not affect how playground and dashboard works as the behavior is opt-in. We will have to un-revert #7639 and we should be able to verify this on the preview playground.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/edge-preview-authenticated-proxy/tests/index.test.ts
+++ b/packages/edge-preview-authenticated-proxy/tests/index.test.ts
@@ -48,9 +48,7 @@ describe("Preview Worker", () => {
 							return Response.redirect("https://example.com", 302)
 						}
 						if(url.pathname === "/method") {
-							return new Response(request.method, {
-								headers: { "X-Test-Http-Method": request.method },
-							});
+							return new Response(request.method)
 						}
 						if(url.pathname === "/status") {
 							return new Response(407)
@@ -337,7 +335,9 @@ describe("Raw HTTP preview", () => {
 							return Response.redirect("https://example.com", 302)
 						}
 						if(url.pathname === "/method") {
-							return new Response(request.method)
+							return new Response(request.method, {
+								headers: { "Test-Http-Method": request.method },
+							})
 						}
 						if(url.pathname === "/status") {
 							return new Response(407)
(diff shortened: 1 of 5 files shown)


packages/edge-preview-authenticated-proxy/tests/index.test.ts
```diff
@@ -487,6 +489,30 @@ compatibility_date = "2023-01-01"
 		expect(await resp.text()).toEqual("PUT");
 	});
 
+	it.each(["GET", "POST", "PUT", "PATCH", "DELETE", "HEAD", "OPTIONS"])(
```
CarmenPopoviciu, line 492: ❤️ 🙏 🚀

packages/playground-preview-worker/src/index.ts
```diff
@@ -79,17 +79,14 @@ async function handleRawHttp(request: Request, url: URL, env: Env) {
...
-			redirect: "manual",
-		})
-	);
+	headers.append("cf-run-user-worker", "true");
+
+	const workerResponse = await userObject.fetch(url, {
+		method,
+		headers,
+		body: method === "GET" || method === "HEAD" ? null : request.body,
+		redirect: "manual",
```
edmundhung, line 88: I have checked which properties we can pass here and it looks like only these 4 are relevant. But if you think differently, please let me know 👍🏼

packages/edge-preview-authenticated-proxy/tests/index.test.ts
```diff
@@ -487,6 +489,30 @@ compatibility_date = "2023-01-01"
...
+		"should support %s method specified on the X-CF-Http-Method header",
+		async (method) => {
+			const token = randomBytes(4096).toString("hex");
+			const resp = await worker.fetch(
+				`https://0000.rawhttp.devprod.cloudflare.dev/method`,
```
andyjessop, line 497: What's the reason for testing a remote worker here? Can we just use Miniflare for this?

edmundhung: It is local 👍🏼 . We are running two separate instances through `unstable_dev` to mimick proxy <-> remote communication: https://github.com/cloudflare/workers-sdk/blob/795620c924c744a97d35182c2708ae4eef73d9e6/packages/edge-preview-authenticated-proxy/tests/index.test.ts#L309-L379

andyjessop (approved):
Looks great, thanks!

penalosa (approved):

PR #7797, opened 2025-01-16
Revert "chore(wrangler): use the unenv preset from `@cloudflare/unenv-preset`"

CarmenPopoviciu:
Reverts cloudflare/workers-sdk#7720 as it introduced breakage in some of the C3 templates (eg. Nuxt)

**How was this tested**

- create a Nuxt project via `npm create cloudflare@latest` (`2.36.0` at the time of this writing)
- change the wrangler version in the generated project's `package.json` to the wrangler pre-release version of this PR
- `npm i` again
- `npm run preview` - which previously failed
- 🎉  👇 

Screenshot 2025-01-16 at 17 14 07

**Why was this not caught by our C3 e2e tests?**

When our C3 e2e tests run, they (in a nutshell) build C3 out of the `main` branch, and create a starter project (via that previously built executable) for each framework under test. Once the starter project is created, certain assertions are made on it.

A starter project will come with a list of predefined dependencies, necessary for the project to run and be deployed, defined in `package.json`. All (I think) C3 starter projects will have a dependency on `wrangler`, specifically, on the **latest released version** of wrangler. This is crucial to note here, because while the e2e tests are running in the "context" of a wrangler built off of `main` (let's call this version `N`), they are in fact making assertions against starter projects that were built with wrangler version `N-1` (`wrangler@latest`).

In our case this resulted in a false positive CI run, where C3 e2e tests were successful, when they should have instead failed

### DIFF
--- a/packages/wrangler/package.json
+++ b/packages/wrangler/package.json
@@ -70,7 +70,6 @@
 	},
 	"dependencies": {
 		"@cloudflare/kv-asset-handler": "workspace:*",
-		"@cloudflare/unenv-preset": "1.x",
 		"@esbuild-plugins/node-globals-polyfill": "0.2.3",
 		"@esbuild-plugins/node-modules-polyfill": "0.2.2",
 		"blake3-wasm": "2.1.5",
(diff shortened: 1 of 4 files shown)


penalosa (approved):

lrapoport-cf (approved):

lrapoport-cf:
@CarmenPopoviciu  this is approved to merge despite c3 e2e tests failing. c3 tests will only be fixed by landing this change and releasing wrangler. see PR description for additional detail/context

PR #7799, opened 2025-01-16
Fix wrangler config detection with json and toml formats

DaniFoldi:

_Describe your change..._
Assume this (admittedly suboptimal) folder layout:
```
- my-fancy-worker/
  - worker.ts
  - wrangler.json
  - some-feature
    - some-worker.ts
    - wrangler.toml
```

Wrangler, without this change, would pick up the outer config file due to the order `findUpSync` was run in. While it is an edge case, we did run into it with a worker where some functionality has been extracted to a separate one for use with jsrpc, and I was migrating them to use `.json` instead of `.toml`.

Fortunately, findUpSync supports an array of strings as input, and even the priority of json > jsonc > toml is kept, if all three are present at the same level.
---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: it's such an edge case
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/red-mice-divide.md
+++ b/.changeset/red-mice-divide.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: Fix wrangler config file detection edge case. Previously, a folder containing wrangler.json, and a subfolder containing wrangler.toml would actually detect the top-level wrangler.json due to the order they were checked.
--- a/packages/wrangler/src/config/config-helpers.ts
+++ b/packages/wrangler/src/config/config-helpers.ts
@@ -50,8 +50,10 @@ export function findWranglerConfig(
 	referencePath: string = process.cwd(),
 	{ useRedirectIfAvailable = false } = {}
 ): ConfigPaths {
-	const userConfigPath =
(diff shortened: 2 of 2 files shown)


DaniFoldi:
Hey team,

From the failing test it looks like this was intentional. Was there a specific use case in mind? If so, I can close this PR, as now that I have migrated our root worker, I shouldn't run into this again.

petebacondarwin (commented):
I think the intention was to look for any wrangler.json before looking for any wrangler.toml. But I guess this could be reconsidered...

penalosa:
@DaniFoldi @petebacondarwin This feels like a bit of an edge case that most people aren't likely to run into. However, I think changing it at this stage would probably be a breaking change, so we might be stuck wwith it. What do you think?

DaniFoldi:
While I think my changed version is the one that makes logical sense, I've reorganized the repo that had workers within another worker, so as we're not going to run into it again, I don't really mind

petebacondarwin:
I'm going to close this as it looks like it would be a breaking change. Thanks for the contribution @DaniFoldi

PR #7802, opened 2025-01-16
Add fixtures using startWorker() and Miniflare directly

penalosa:
Just adding some additional fixtures to demostrate direct use of the Miniflare API and `startWorker()` in a non-Vitest testing framework.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: only changes fixtures
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: only changes fixtures


fixtures/miniflare-node-test/src/index-build.test.js
```diff
@@ -0,0 +1,41 @@
...
+import { Miniflare } from "miniflare";
+
+before(() => {
+	spawnSync("npx wrangler build -c wrangler-build.json", {
+		shell: true,
+		stdio: "pipe",
+	});
+});
+
+describe("worker build", () => {
```
vicb, line 13: Very nice PR!

nit: "worker build" and "worker" names in the next file are not very descriptive.

(question: how to use a different worked when interacting with Miniflare? I don't see an option so I guess `MINIFLARE_WORKERD_VERSION` is respected?)

penalosa: `MINIFLARE_WORKERD_VERSION` should be respected, yes

vicb: Do you think it would be worth adding some docs [on the class](https://github.com/cloudflare/workers-sdk/blob/59c7c8ee177d9345948a416377c6625269d58925/packages/miniflare/src/index.ts#L706)?

penalosa: It's documented in the README: https://github.com/cloudflare/workers-sdk/tree/main/packages/miniflare/README.md#L873

fixtures/miniflare-node-test/src/index.ts
```diff
@@ -0,0 +1,10 @@
+import isEven from "is-even";
+
+export default {
+	async fetch(request): Promise<Response> {
+		const url = new URL(request.url);
+		return new Response(
+			isEven(Number(url.searchParams.get("number") ?? "0")) ? "even" : "odd"
```
dario-piotrowicz, line 7: why are you using `is-even` instead of just doing `% 2 === 0`? is it to test module resolution? 🤔

penalosa: Yeah—tested in https://github.com/cloudflare/workers-sdk/pull/7802/files/fe10741b093ffbece2ebac749aa746b31ee717b5#diff-3023250eb0476b0beae6fb7213ed15a640ed39f45a701803ace0c77d847fa10b

dario-piotrowicz: ah ok, I would probably put a comment somewhere on why the function is used, but I'm also ok not to have such a comment if you think that this is obvious 🙂

dario-piotrowicz (approved):
looks good to me 😄

PR #7830, opened 2025-01-20
vite-plugin: update to use the new `@cloudflare/unenv-preset`

### LINKED ISSUE #8054: 🐛 BUG: @cloudflare/vite-plugin Vite 6.1.0 `vite dev` errors when `nodejs_compat` is enabled

Other

@cloudflare/vite-plugin 0.0.6, Vite 6.1.0, Wrangler 3.107.3, Node.js v22.13.1

Mac Sequoia 15.3 and Ubuntu 24.10

https://github.com/patsaas/cloudflare-vite-plugin-repro

### Describe the Bug

With the [new Vite plugin](https://github.com/cloudflare/workers-sdk/tree/main/packages/vite-plugin-cloudflare) and Vite 6.1.0, `vite` errors on request when the `nodejs_compat` or `nodejs_compat_v2` compatibility flag is enabled.

Error on request to the Worker at `/api/`:

```
Incorrect type for Promise: the Promise did not resolve to 'Response'.
at async ProxyServer.fetch (file:///path/to/cloudflare-vite-tutorial/node_modules/miniflare/src/workers/core/proxy.worker.ts:173:11
```

(I'm unsure if it matters, but I'll point out the `miniflare/src` directory doesn't/shouldn't exist in my `node_modules`.

Errors from the `vite` dev server (on macOS at least):
```
error running /Users/pat/cloudflare-vite-tutorial/node_modules/.vite/deps_cloudflare_vite_tutorial/chunk-RWKZ2VVH.js?v=4e1f1797
Error: Failed to load url unenv/runtime/node/timers/promises/index (resolved id: unenv/runtime/node/timers/promises/index) in /Users/pat/cloudflare-vite-tutorial/node_modules/.vite/deps_cloudflare_vite_tutorial/chunk-RWKZ2VVH.js. Does the file exist?
    at loadAndTransform (file:///Users/pat/cloudflare-vite-tutorial/node_modules/vite/dist/node/chunks/dep-CfG9u7Cn.js:41295:17)

error running /Users/pat/cloudflare-vite-tutorial/node_modules/.vite/deps_cloudflare_vite_tutorial/unenv_runtime_node_timers_$cloudflare.js?v=4e1f1797
Error: Failed to load url unenv/runtime/node/timers/promises/index (resolved id: unenv/runtime/node/timers/promises/index) in /Users/pat/cloudflare-vite-tutorial/node_modules/.vite/deps_cloudflare_vite_tutorial/chunk-RWKZ2VVH.js. Does the file exist?
    at loadAndTransform (file:///Users/pat/cloudflare-vite-tutorial/node_modules/vite/dist/node/chunks/dep-CfG9u7Cn.js:41295:17)
error running /Users/pat/cloudflare-vite-tutorial/api/index.ts
Error: Failed to load url unenv/runtime/node/timers/promises/index (resolved id: unenv/runtime/node/timers/promises/index) in /Users/pat/cloudflare-vite-tutorial/node_modules/.vite/deps_cloudflare_vite_tutorial/chunk-RWKZ2VVH.js. Does the file exist?
    at loadAndTransform (file:///Users/pat/cloudflare-vite-tutorial/node_modules/vite/dist/node/chunks/dep-CfG9u7Cn.js:41295:17)
```

(These files _do_ exist.)

[My repro](https://github.com/patsaas/cloudflare-vite-plugin-repro) follows [@cloudflare/vite-plugin's tutorial](https://github.com/cloudflare/workers-sdk/tree/main/packages/vite-plugin-cloudflare#tutorial) and simply adds the `nodejs_compat` compatibility flag to `wrangler.toml`.

`npm install && npm run dev` then visit `/api/` or click the "Name from API is" button on the homepage to reproduce the error(s).

I confirmed this isn't an issue with Vite 6.0.11; only Vite 6.1.0 as [pointed out](https://github.com/cloudflare/workers-sdk/issues/8054#issuecomment-2642373588).

petebacondarwin:
Fixes #8054

* Updates the Vite plugin to use the new Cloudflare unenv preset
* Updates the Vite plugin to support Vite 6.1 (fixing the issues with the node-compat)
* Refactors the node-compat stuff to isolate it all in a single Vite plugin

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not related to Wrangler e2e tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: refactoring/bug fix


packages/vite-plugin-cloudflare/package.json
```diff
@@ -40,9 +40,10 @@
 		"test:ci": "vitest run"
 	},
 	"dependencies": {
+		"@cloudflare/unenv-preset": "~1.1.1",
```
vicb, line 43: - Not sure if this should be "1.1.1" (no `~`)? We can relax the range once 2.0 is released
- Is it expected that the plugin has no (peer) dep on `workerd`?

vicb: wrangler is in both devDep and peerDep

petebacondarwin: Wrangler is there twice because the `devDep` is only for testing (and we use the workspace version); while the peerDep is a range ("wrangler": "^3.101.0") which constrains the minimum version.

vicb (commented):
I only took a quick look.

How urgent is this? If this can be put on hold until we release the v2 of the preset, that would be ideal

petebacondarwin:
> I only took a quick look.
> 
> How urgent is this? If this can be put on hold until we release the v2 of the preset, that would be ideal

The Vite plugin is currently broken with Vite 6.1 and this solves that. If we want to hold off on landing this then I would have to put up a fix for the old unenv usage.

Is there a reason to hold this back? This Vite plugin is, itself, in beta, so it feels like a good place to try out new stuff like the Cloudflare unenv update.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -1,18 +1,124 @@
...
+} from "./plugin-config";
+import type { Plugin } from "vite";
+
+/**
+ * A Vite plugin that can provide Node.js compatibility support for Vite Environments that are hosted in Cloudflare Workers.
+ */
+export function nodejsCompatPlugin(pluginConfig: PluginConfig): Plugin {
+	let resolvedPluginConfig: ResolvedPluginConfig;
+	return {
+		name: "vite-plugin-nodejs-compat",
```
jamesopstad, line 20: Let's use this naming convention for child plugins.

```suggestion
		name: "vite-plugin-cloudflare:nodejs-compat",
```

packages/vite-plugin-cloudflare/package.json
petebacondarwin: the cf unenv preset has its own peer dependency on workerd.

vicb:
> Is there a reason to hold this back? This Vite plugin is, itself, in beta, so it feels like a good place to try out new stuff like the Cloudflare unenv update.

I'm asking mostly because I don't have time to do a deep review.
But if everything is still in beta, there should be no problem.

There might be changes to do after https://github.com/cloudflare/workers-sdk/pull/7915 is merged / 2.0 is released

packages/vite-plugin-cloudflare/package.json
vicb: > Happy to pin this dep to 1.1.1 if that helps

It's more that we do not expect any 1.1.2...

> the cf unenv preset has its own peer dependency on workerd.

it's a peer dep so it should be installed somewhere...

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -1,18 +1,124 @@
...
+}
+
+const { env } = defineEnv({
+	nodeCompat: true,
+	presets: [cloudflare],
+});
 
-const require = createRequire(import.meta.url);
-const preset = unenv.env(unenv.nodeless, unenv.cloudflare);
 const CLOUDFLARE_VIRTUAL_PREFIX = "\0cloudflare-";
```
jamesopstad, line 116: Maybe let's adopt this convention for private module prefixes to reduce the risk of collisions:

```suggestion
const CLOUDFLARE_VIRTUAL_PREFIX = "\0__CLOUDFLARE_NODEJS_COMPAT__";
```

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -1,18 +1,124 @@
...
+				return;
+			}
+
+			const resolvedAlias = await this.resolve(
+				unresolvedAlias,
+				import.meta.url
+			);
+			if (!resolvedAlias) {
+				return;
+			}
```
jamesopstad, line 77: Should we be asserting rather than returning `undefined` here? Otherwise, won't it resolve with the prefixed ID?

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -40,24 +146,16 @@ export function isNodeCompat({
...
+	const injectedCode = Object.entries(env.inject)
 		.map(([globalName, globalInject]) => {
 			if (typeof globalInject === "string") {
 				const moduleSpecifier = globalInject;
 				// the mapping is a simple string, indicating a default export, so the string is just the module specifier.
 				return `import var_${globalName} from "${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName};\n`;
 			}
 
-			// the mapping is a 2 item tuple, indicating a named export, made up of a module specifier and an export name.
+			// the mapping is a 2 item tuple, indicating a named export, made up of a module specifier and an name.
```
jamesopstad, line 158:
```suggestion
			// the mapping is a 2 item tuple, indicating a named export, made up of a module specifier and a name.
```

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -76,65 +174,50 @@ export function injectGlobalCode(
...
+	if (!isNodeCompat(workerConfig)) {
+		// We are not in node compat mode so just return the original module specifier
+		return from;
+	}
 
-	if (alias && preset.external.includes(alias)) {
-		throw new Error(`Alias to external: ${source} -> ${alias}`);
+	const alias = env.alias[from];
+	if (!alias) {
+		return;
```
jamesopstad, line 215: What's actually happened if we get here and `alias` is `undefined`? If this an error and should it be an assertion?

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -1,18 +1,124 @@
...
+/**
+ * A Vite plugin that can provide Node.js compatibility support for Vite Environments that are hosted in Cloudflare Workers.
+ */
+export function nodejsCompatPlugin(pluginConfig: PluginConfig): Plugin {
+	let resolvedPluginConfig: ResolvedPluginConfig;
+	return {
+		name: "vite-plugin-nodejs-compat",
+		config(userConfig, env) {
+			// Capture the configuration of the Cloudflare Workers for use in other hooks, below.
+			resolvedPluginConfig = resolvePluginConfig(pluginConfig, userConfig, env);
```
jamesopstad, line 23: I don't think that we should be resolving this multiple times. Please can we move this plugin into `index.ts` and use the `resolvedPluginConfig` variable that is declared there.

jamesopstad (commented):
Looks good, as far as I understand the new preset. I would prefer the plugins to be defined in the same file though so that they have access to the `cloudflare` function scope.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: d03e2df996b448cb58abbf3ed7c2cfba3a7d7bf1

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: OK :sad:

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: I'm not sure the best practice here. If we return `undefined` then other plugins might have a crack at resolving it, right?
And then if nothing else can resolve this you'd get an error later on.

But if we are sure that failure to resolve here should be unrecoverable, we can throw instead.

petebacondarwin: be3a0132eabda03a989eb68936ce748e6138f45f

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: cb4388afa97d8a7132c7dc5b7203558d47acc668

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: 92913887f12d48ed1f54992fa162c2ccb2af99d1

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: Right I think this is an error state.

petebacondarwin: 3b726f727e59f14f69c2dc098dc60737340e298f

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: One reason I was thinking about this is that the plugin in this form is very nearly ready to be reused for the Vitest stuff.

packages/vite-plugin-cloudflare/package.json
petebacondarwin: It'll be installed as part of Wrangler

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -48,17 +51,13 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 		{
 			name: "vite-plugin-cloudflare",
 			config(userConfig, env) {
-				if (env.isPreview) {
-					return { appType: "custom" };
-				}
-
```
jamesopstad, line 54: One issue with removing this is that any logs/warnings emitted by the Wrangler `readConfig` function will be emitted for both the input and output Worker config files.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -419,7 +384,84 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+				const resolvedId = await this.resolve(workerConfig.main);
+				if (id === resolvedId?.id) {
+					return injectGlobalCode(id, code);
+				}
+			},
+		},
 	];
+
+	function getWorkerConfig(environmentName: string) {
+		return resolvedPluginConfig.type !== "assets-only"
```
jamesopstad, line 461: If we did keep the reduced config for preview mode then this change should work:

```suggestion
		return resolvedPluginConfig && resolvedPluginConfig.type !== "assets-only"
```

jamesopstad (approved):
Looks good! I think there's still a bit of an open question about what's correct for preview mode but happy for you to make the decision on this.

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: Ah interesting. Now that I have refactored things I think you are right. We don't actually need there to be a worker config anywhere in preview mode, right?

The previous concern was the dealiasing of the virtual modules but now I strip the virtual prefix whether or not there is node.js compat turned on.

jamesopstad: Yep, I think if it's OK with you let's keep the reduced preview config. We can always reevaluate later in a separate PR.

penalosa (approved):

PR #7832, opened 2025-01-20
Relax the messaging when Wrangler uses redirected configuration

petebacondarwin:
Previously the messaging was rendered as a warning, which implied that the user had done something wrong. Now it is just a regular info message.

Fixes #0000

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: minor change that doesn't affect docs

### DIFF
--- a/packages/wrangler/src/config/config-helpers.ts
+++ b/packages/wrangler/src/config/config-helpers.ts
@@ -128,11 +128,11 @@ function findRedirectedWranglerConfig(
 			}
 		}
 
-		logger.warn(dedent`
+		logger.info(dedent`
 			Using redirected Wrangler configuration.
-			Configuration being used: "${path.relative(".", redirectedConfigPath)}"
-			Original user's configuration: "${userConfigPath ? path.relative(".", userConfigPath) : "<no user config found>"}"
-			Deploy configuration file: "${path.relative(".", deployConfigPath)}"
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/config/config-helpers.ts
```diff
@@ -128,7 +128,7 @@ function findRedirectedWranglerConfig(
 			}
 		}
 
-		logger.warn(dedent`
+		logger.info(dedent`
```
penalosa, line 131: Purely aesthetic, but it feels like it might be nice to indent a bit to make the grouping clearer? Something like:

```
Using redirected Wrangler configuration:
  Configuration being used: ...
  Original user's configuration: ...
  Deploy configuration file: ...
```

penalosa (approved):

packages/wrangler/src/config/config-helpers.ts
petebacondarwin: OK

petebacondarwin: done

dario-piotrowicz (approved):
LGTM 😄

jamesopstad (approved):

PR #7833, opened 2025-01-20
Relax the Wrangler peer dependency for the Vite plugin

petebacondarwin:
We only need the Wrangler peer dependency to be at least 3.101.0,
which is when we added support for the redirected config.

By using the peer dependency range version in the playgrounds, we ensure that
we are testing against what will be used in practice. It will also ensure that this
dependency gets bumped when needed, since the playgrounds might start to fail otherwise.

Fixes #000.

(No changeset required because this is covered by the current changeset for the first release of the plugin)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Vite plugin not covered by e2e tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no documentation facing change

### DIFF
--- a/packages/vite-plugin-cloudflare/package.json
+++ b/packages/vite-plugin-cloudflare/package.json
@@ -59,7 +59,7 @@
 	},
 	"peerDependencies": {
 		"vite": "catalog:vite-plugin",
-		"wrangler": "workspace:^"
+		"wrangler": "catalog:vite-plugin"
 	},
 	"publishConfig": {
 		"access": "public"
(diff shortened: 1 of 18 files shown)


dario-piotrowicz (approved):
LGTM 🙂

penalosa (approved):

jamesopstad (approved):

PR #7846, opened 2025-01-21
[vite-plugin-cloudflare] make sure that runner initialization is properly validated

dario-piotrowicz:
Fixes https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/59

The plugin validation around the module runner initialization is not properly propagated, the changes here make sure
that it is and that a (hopefully) helpful error message is shown to the user

An easy way to see the different behavior here is to make the `createModuleRunner` function always throw:
Screenshot 2025-01-21 at 15 30 14
and run a playground application, without the changes here it's not clear/surfaced that the runner failed initialization

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: this is an extra check that, it is just in case, we don't expect the error to be triggered and testing this would require some potentially hacking tweaks in the runner initialization code (so that it can throw with no proper reason), so it's most likely not worth it, I am happy to have a look if we do strongly believe that this needs testing though
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no wrangler changes
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: just better error handling

### DIFF
--- a/packages/vite-plugin-cloudflare/src/runner-worker/index.ts
+++ b/packages/vite-plugin-cloudflare/src/runner-worker/index.ts
@@ -177,9 +177,7 @@ export function createWorkerEntrypointWrapper(
 						await createModuleRunner(this.env, server);
 					} catch (e) {
 						return new Response(
-							JSON.stringify(
-								e instanceof Error ? e.message : JSON.stringify(e)
-							),
+							e instanceof Error ? e.message : JSON.stringify(e),
 							{ status: 500 }
 						);
 					}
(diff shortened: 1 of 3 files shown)


packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -173,7 +173,16 @@ export function createWorkerEntrypointWrapper(
...
 				if (url.pathname === INIT_PATH) {
 					const { 0: client, 1: server } = new WebSocketPair();
-					createModuleRunner(this.env, server);
+					try {
+						await createModuleRunner(this.env, server);
+					} catch (e) {
+						return new Response(
+							JSON.stringify(
+								e instanceof Error ? e.message : JSON.stringify(e)
+							),
```
jamesopstad, line 182: Is it intentional to stringify twice if the error is not an instance of `Error` here?

dario-piotrowicz: no, sorry my bad, thanks for spotting it 🙏

PR #7854, opened 2025-01-21
resolves #925

beratbayram:
Fixes #925.

WHAT: It changes the output of next.js with TS template.
WHY: Basic templating is broken.
HOW: Existing customers are not affected. New customers will have a better experience.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: already written
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Changes the output not the usage

---

I've run tests and two fails as follows but I don't believe they are related to my changes.

image

image

### DIFF
--- a/.changeset/moody-walls-shake.md
+++ b/.changeset/moody-walls-shake.md
@@ -1,0 +1,9 @@
+---
+"create-cloudflare": major
+---
+
+# Typescript transpilation error with clean Next.js template #925
+
+WHAT: It changes the output of next.js with TS template.
+WHY: Basic template is broken (see: #925).
+HOW: Existing customers are not affected. New customers will have a better experience.
--- a/packages/create-cloudflare/templates/next/c3.ts
+++ b/packages/create-cloudflare/templates/next/c3.ts
@@ -8,6 +8,7 @@ import {
 	probePaths,
 	readFile,
 	readJSON,
+	removeFile,
 	usesEslint,
 	usesTypescript,
 	writeFile,
@@ -31,29 +32,55 @@ const generate = async (ctx: C3Context) => {
 	updateStatus("Created wrangler.json file");
 };
 
-const updateNextConfig = (usesTs: boolean) => {
+const setupDevPlatformCode =
+	`import { setupDevPlatform } from '@cloudflare/next-on-pages/next-dev';
+
+	// Here we use the @cloudflare/next-on-pages next-dev module to allow us to
+	// use bindings during local development (when running the application with \`next dev\`).
+	// For more information see:
+	// https://github.com/cloudflare/next-on-pages/blob/main/internal-packages/next-dev/README.md
+	if (process.env.NODE_ENV === 'development') {
+		await setupDevPlatform();
+	}
+
+	`.replace(/\n\t*/g, "\n");
+
+const updateNextJsConfig = () => {
 	const s = spinner();
 
-	const configFile = `next.config.${usesTs ? "ts" : "mjs"}`;
+	const configFile = `next.config.mjs`;
 	s.start(`Updating \`${configFile}\``);
 
 	const configContent = readFile(configFile);
 
-	const updatedConfigFile =
-		`import { setupDevPlatform } from '@cloudflare/next-on-pages/next-dev';
-
-		// Here we use the @cloudflare/next-on-pages next-dev module to allow us to use bindings during local development
-		// (when running the application with \`next dev\`), for more information see:
-		// https://github.com/cloudflare/next-on-pages/blob/main/internal-packages/next-dev/README.md
-		if (process.env.NODE_ENV === 'development') {
-		  await setupDevPlatform();
-		}
-
-		`.replace(/\n\t*/g, "\n") + configContent;
+	const updatedConfigFile = setupDevPlatformCode + configContent;
 
 	writeFile(configFile, updatedConfigFile);
 
-	s.stop(`${brandColor(`updated`)} ${dim(`\`${configFile}\``)}`);
+	s.stop(`${brandColor(`Updated`)} ${dim(`\`${configFile}\``)}`);
+};
+
+const updateNextTsConfig = () => {
+	const s = spinner();
+
+	const oldConfigFile = `next.config.ts`;
+	const newConfigFile = `next.config.mjs`;
+
+	s.start(`Removing \`${oldConfigFile}\``);
+
+	const updatedConfigFile = setupDevPlatformCode +
+	`/** @type {import('next').NextConfig} */
+	const nextConfig = {
+	  /* config options here */
+	};
+
+	export default nextConfig;
+	`.replace(/\n\t*/g, "\n");
+
+	removeFile(oldConfigFile);
+	writeFile(newConfigFile, updatedConfigFile);
+
+	s.stop(`${brandColor(`Created`)} ${dim(`\`${newConfigFile}\``)}`);
 };
 
 const configure = async (ctx: C3Context) => {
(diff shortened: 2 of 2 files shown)


beratbayram:
[dario-piotrowicz](https://github.com/dario-piotrowicz)

packages/create-cloudflare/templates/next/c3.ts
```diff
@@ -31,29 +32,55 @@ const generate = async (ctx: C3Context) => {
...
 
-const updateNextConfig = (usesTs: boolean) => {
+const setupDevPlatformCode =
+	`import { setupDevPlatform } from '@cloudflare/next-on-pages/next-dev';
+
+	// Here we use the @cloudflare/next-on-pages next-dev module to allow us to
+	// use bindings during local development (when running the application with \`next dev\`).
+	// For more information see:
+	// https://github.com/cloudflare/next-on-pages/blob/main/internal-packages/next-dev/README.md
+	if (process.env.NODE_ENV === 'development') {
```
vicb, line 42: @dario-piotrowicz do we need `(process.env.NODE_ENV === 'development')` here? (or could it be inside `setupDevPlatform` ?

(The code here is only moved from l42 so not directly related to the PR)

.changeset/moody-walls-shake.md
```diff
@@ -0,0 +1,9 @@
+---
+"create-cloudflare": major
```
vicb, line 2: probably only a patch? (bug fix)

vicb (commented):
I took a quick look here but I don't really think the changes are optimal, there should be no reason to need both a TS and JS version of `updateNextXConfig`.

One thing that look weird is that we do not pass `--js` or `--ts` to `create-next-app
`https://github.com/cloudflare/workers-sdk/blob/6c2e94318dd079f9cba3f763e67ba6698ef597f6/packages/create-cloudflare/templates/next/c3.ts#L27

beratbayram:
> I took a quick look here but I don't really think the changes are optimal, there should be no reason to need both a TS and JS version of `updateNextXConfig`.

JS version only prepends some code to existing config while TS version removes the existing one completely and creates a new hardcoded one from scratch. I can make JS version also behave like TS version but in case of a possible config change by next.js team, both versions will be affected instead of only the TS version.

> One thing that look weird is that we do not pass `--js` or `--ts` to `create-next-app `
> 
> https://github.com/cloudflare/workers-sdk/blob/6c2e94318dd079f9cba3f763e67ba6698ef597f6/packages/create-cloudflare/templates/next/c3.ts#L27

We don't need that, user decides that during next-cli prompt

.changeset/moody-walls-shake.md
vicb: > I've run pnpm changeset, and it decided that it is a major change

I have to admit that the "UI" is weird but `changeset` does not decide anything for you, it only depends on what you type :)

vicb:
> We don't need that, user decides that during next-cli prompt

Thanks for the clarification here.

> JS version only prepends some code to existing config while TS version removes the existing one completely and creates a new hardcoded one from scratch. I can make JS version also behave like TS version but in case of a possible config change by next.js team, both versions will be affected instead of only the TS version.

Could you expand on why the TS config would need to be fully replaced? (maybe add the generated TS config + explanation to the PR description). Thanks!

beratbayram:
> Thanks for the clarification here.

No problem

> Could you expand on why the TS config would need to be fully replaced? (maybe add the generated TS config + explanation to the PR description). Thanks!

I've just edited the description. Please also check the conversation at the linked issue.

.changeset/moody-walls-shake.md
beratbayram: > changeset does not decide anything for you

I had thought there was some logic behind it. My bad. I've just changed it to patch as you suggested

dario-piotrowicz:
> > because C3 generates an await statement inside a .ts file which is illegal
> 
> I think we should drop the `await`, see:
> 
> [cloudflare/next-on-pages@`2cd4c3c`/internal-packages/next-dev/src/index.ts#L6-L13](https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/internal-packages/next-dev/src/index.ts#L6-L13)
> 
> [cloudflare/next-on-pages@`2cd4c3c`/internal-packages/next-dev/README.md?plain=1#L56-L59](https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/internal-packages/next-dev/README.md?plain=1#L56-L59)
> 
> @dario-piotrowicz what do you think?

yeah I forgot about the `await`ing not being actually necessary, yes probably removing it would be the simplest thing, I feel like awaiting the function provides a nice safety net (since we've seen that not awaiting the function is totally fine, but there is no hard guarantee about that, in theory it could no longer be true if Next.js changed how they run their dev server etc...), but since now it complicates the C3 logic here I think we might just get rid of it

packages/create-cloudflare/templates/next/c3.ts
dario-piotrowicz: This check is totally unnecessary, moving it inside `setupDevPlatform` wouldn't actually accomplish anything

it's here to try to make it extra clear to the developer that this function is only needed in development, in reality this only has any effect during development anyways

I'm happy to get rid of it if you want 👍

vicb: > This check is totally unnecessary

Then I would drop it.

And in [next-on-pages](https://github.com/cloudflare/next-on-pages/blob/2cd4c3c704a00e6b693229f1f14102abc6318d11/pages-e2e/fixtures/issue797app/next.config.mjs#L6) as well

beratbayram:
I've reverted all of my previous work.

> yeah I forgot about the awaiting not being actually necessary, yes probably removing it would be the simplest thing,

Removed `await` and added `.catch(console.error)`

> it's here to try to make it extra clear to the developer that this function is only needed in development

Removed the if and improved the comment to emphasize that.

.changeset/famous-coats-admire.md
```diff
@@ -0,0 +1,7 @@
+---
+"create-cloudflare": patch
+---
+
+# \[create-cloudflare\] next.js with ts fix #7854
```
vicb, line 5:
```suggestion
Update the Next.js template
```

`#` will generate a weird formatting in the changelog

.changeset/famous-coats-admire.md
```diff
@@ -0,0 +1,7 @@
...
+
+Removed top-level await in next.config.ts
```
vicb:
```suggestion
Removed top-level `await` in `next.config.ts` as it is not allowed there and not required.
```

vicb (approved):
LGTM, thanks for the PR and the updates!

I triggered the e2e

.changeset/famous-coats-admire.md
```diff
@@ -0,0 +1,8 @@
+---
+"create-cloudflare": patch
+---
+
+Update the Next.js template
+
+* Removed top-level `await` in `next.config.ts` as it is not allowed there and not required.
+* Improved setupDevPlatform() comment
```
vicb, line 8: well then:

```suggestion
* Improved `setupDevPlatform()` comment
```

beratbayram:
Thanks for your directions and your patience. It was fun to work on this.

vicb:
@beratbayram it looks like the formatting is wrong, could you please run `pnpm fix` on the change?

beratbayram:
> @beratbayram it looks like the formatting is wrong, could you please run `pnpm fix` on the change?

done

CarmenPopoviciu (approved):

PR #7856, opened 2025-01-21
feat: add sanitised error messages to wrangler telemetry collection

emily-shen:
Fixes DEVX-1529

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: shouldn't affect e2e paths
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: documentation included in PR (see telemetry.md changes)

### DIFF
--- a/packages/wrangler/src/errors.ts
+++ b/packages/wrangler/src/errors.ts
@@ -5,11 +5,26 @@
  * messaging.
  */
 export class UserError extends Error {
-	constructor(...args: ConstructorParameters<typeof Error>) {
-		super(...args);
+	cleanMessage: string | undefined;
+	constructor(
+		message?: string | undefined,
+		options?:
+			| (ErrorOptions & {
+					/**
+					 * This is used to provide telemetry with a sanitised error
+					 *  message that could not have any user-identifying information.
+					 * Use "*" as shorthand to duplicate `message`.
+					 *  */
+					cleanMessage?: string;
+			  })
+			| undefined
+	) {
+		super(message, options);
 		// Restore prototype chain:
 		// https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#support-for-newtarget
 		Object.setPrototypeOf(this, new.target.prototype);
+		this.cleanMessage =
+			options?.cleanMessage === "*" ? message : options?.cleanMessage;
 	}
 }
 
@@ -22,9 +37,10 @@ export class DeprecationError extends UserError {
 export class FatalError extends UserError {
 	constructor(
 		message?: string,
-		readonly code?: number
+		readonly code?: number | undefined,
+		options?: { cleanMessage?: string }
 	) {
-		super(message);
+		super(message, options);
 	}
 }
 
--- a/packages/wrangler/src/errors.ts
+++ b/packages/wrangler/src/errors.ts
@@ -1,7 +1,7 @@
 /**
  * This is used to provide telemetry with a sanitised error
- *  message that could not have any user-identifying information.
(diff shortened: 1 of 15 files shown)


packages/wrangler/src/errors.ts
```diff
@@ -5,11 +5,26 @@
...
+	constructor(
+		message?: string | undefined,
+		options?:
+			| (ErrorOptions & {
+					/**
+					 * This is used to provide telemetry with a sanitised error
+					 *  message that could not have any user-identifying information.
+					 * Use "*" as shorthand to duplicate `message`.
+					 *  */
+					cleanMessage?: string;
```
edmundhung, line 18: Nit: How about `telemetryMessage`? I would prefer making it a bit more explicit for its purpose.

Also, instead of having a special shorthand, we can consider making this `boolean | string`. So that we can specify `true` if we are fine with the current message.

packages/wrangler/src/errors.ts
```diff
@@ -22,9 +37,10 @@ export class DeprecationError extends UserError {
 export class FatalError extends UserError {
 	constructor(
 		message?: string,
-		readonly code?: number
+		readonly code?: number | undefined,
+		options?: { cleanMessage?: string }
```
edmundhung, line 41: **Nit**: Just an idea

```ts
export class FatalError extends UserError {
	constructor(
		message?: string | {
       logMessage: string;
       telemetryMessage?: boolean | string;
    },
		readonly code?: number | undefined,
  )
}
```

packages/wrangler/src/errors.ts
```diff
@@ -1,15 +1,29 @@
+/**
+ * This is used to provide telemetry with a sanitised error
+ *  message that could not have any user-identifying information.
+ * Use "*" as shorthand to duplicate `message`.
```
penalosa, line 4: Is this comment accurate re *?

packages/wrangler/src/errors.ts
```diff
@@ -22,13 +36,18 @@ export class DeprecationError extends UserError {
...
+		options?: TelemetryMessage
 	) {
-		super(message);
+		super(message, options);
 	}
 }
 
-export class CommandLineArgsError extends UserError {}
+export class CommandLineArgsError extends UserError {
+	constructor(message: string, options?: { telemetryMessage?: string | true }) {
```
penalosa, line 47:
```suggestion
	constructor(message: string, options?: TelemetryMessage) {
```

packages/wrangler/src/errors.ts
```diff
@@ -41,15 +60,18 @@ export class CommandLineArgsError extends UserError {}
...
-		super(message);
+		super(message, code, options);
 	}
 }
 
 export class MissingConfigError extends Error {
+	telemetryMessage: string | undefined;
 	constructor(key: string) {
 		super(`Missing config value for ${key}`);
+		this.telemetryMessage = `Missing config value for ${key}`;
```
penalosa, line 74: Could this just be `super(..., {telemetryMessage: true})`?

penalosa (approved):
This looks great! I love the `telemetryMessage` option

packages/wrangler/src/errors.ts
emily-shen: i'm not sure I understand, MissingConfigError extends Error, so I don't think that works?

penalosa: Ah—thought it extended UserError—never mind!

PR #7860, opened 2025-01-22
refactor(wrangler): make JSON parsing independent of Node

vicb:
This first started as a PR to relax a constraint on an exception message...

It turned out to be because Node throws different exceptions when `JSON.parse`ing across different version.

So let's use `jsonc-parser` instead that will be consistent across Node versions/runtimes.

One more change is that `parseJSON(c)` are no more parametrized. They returns `unknow` because that's the truth.
Callers can change that with an explicit cast.

(Any JSONC specifics are disabled when parsing JSON).

____

Initial description:

While the test passes on the CI, it fails locally for me:

```
wrangler:test:ci:  FAIL  src/__tests__/versions/secrets/bulk.test.ts > versions secret bulk > should error on invalid json file
wrangler:test:ci: AssertionError: expected [Function] to throw error including 'The contents of "secrets.json" is not…' but got 'The contents of "secrets.json" is not…'
wrangler:test:ci: 
wrangler:test:ci: Expected: "The contents of "secrets.json" is not valid JSON: "ParseError: Unexpected token o""
wrangler:test:ci: Received: "The contents of "secrets.json" is not valid JSON: "SyntaxError: Unexpected token 'o', "not valid json :(" is not valid JSON""
wrangler:test:ci: 
```

It might be linked to the Node version?

Anyway details are not meaningful, dropping them.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: not affected
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: bug fix / refactor

### DIFF
--- a/.changeset/thick-bikes-teach.md
+++ b/.changeset/thick-bikes-teach.md
@@ -1,0 +1,10 @@
+---
+"wrangler": patch
+---
+
+refactor(wrangler): make JSON parsing independent of Node
+
+Switch `jsonc-parser` to parse json:
+
+- `JSON.parse()` exception messages are not stable across Node versions
+- While `jsonc-parser` is used, JSONC specific syntax is disabled
--- a/packages/wrangler/src/__tests__/parse.test.ts
+++ b/packages/wrangler/src/__tests__/parse.test.ts
@@ -225,17 +225,15 @@ describe("parseJSON", () => {
 				kind: "error",
 				location: {
 					line: 3,
-					column: 9,
+					column: 10,
+					length: 2,
 					lineText: '"version" "1',
 					file: undefined,
 					fileText: `\n{\n"version" "1\n}\n`,
 				},
 				notes: [],
 			});
-			expect(text).oneOf([
-				/* Node.js v16/v18 */ "Unexpected string",
-				/* Node.js v20+ */ "Expected ':' after property name",
-			]);
+			expect(text).toEqual("UnexpectedEndOfString");
 		}
 	});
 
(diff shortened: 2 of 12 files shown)


petebacondarwin:
This is actually highlighting a bug due to the change in Node version.
In our `parseJSON()` helper we are looking for a specific string in the error and then converting to a `ParseError`.


In Node 18:

```
> JSON.parse("not valid json :(")
Uncaught SyntaxError: Unexpected token o in JSON at position 1
```

In Node 22:

```
> JSON.parse("not valid json :(")
Uncaught SyntaxError: Unexpected token 'o', "not valid json :(" is not valid JSON
```

The helper is looking for this text: `" in JSON at position "`

petebacondarwin:
So just ignoring this failure by changing the expectation is not the correct solution here.
Instead we shoud improve the helper function to handle Node 22

vicb:
> So just ignoring this failure by changing the expectation is not the correct solution here. Instead we shoud improve the helper function to handle Node 22

We should also probably drop the generic while we are at it:

```ts
export function parseJSON<T>(input: string, file?: string): T
```

It should just return `unknown` (which is the truth) and let the caller cast.

petebacondarwin:
> We should also probably drop the generic while we are at it:
> 
> ```ts
> export function parseJSON<T>(input: string, file?: string): T
> ```
> 
> It should just return `unknown` (which is the truth) and let the caller cast.

I don't feel strongly about that suggestion. It will mean that most calls will end up going from:

```ts
const foo = parseJSON<MyType>(...);
```

to

```ts
const foo = parseJSON(...) as MyType;
```

It is unlikely that we will get better validation by changing this.

vicb:
> I don't feel strongly about that suggestion. It will mean that most calls will end up going from:

I do - probably because that pattern (generic only for return type) was strictly forbidden by my previous employer ;)

PR #7864, opened 2025-01-22
improve and test vite-plugin-cloudflare `.dev.vars` files support

### LINKED ISSUE #7850: Support `.dev.vars` in Vite plugin
Add support for loading secrets from a `.dev.vars` file during local development.

dario-piotrowicz:
fixes #7850

the changes here add tests for making sure that the vite plugin correctly reads secrets from `.dev.vars` files with and without a specified environment

as part of this wragler's `unstable_getMiniflareWorkerOptions` utility needed to also be updated to accept the environment name as its second parameter

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included (tests added for the vite-plugin, as for wrangler, it looks like the `unstable_getMiniflareWorkerOptions` utility is not currently tested, so I don't think it's necessary to include any here either)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: 
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: documentation added as part of our vite-plugin README (which is currently the only documentation the package has)


packages/vite-plugin-cloudflare/README.md
```diff
@@ -382,6 +382,17 @@ It accepts an optional `PluginConfig` parameter.
...
+## Secrets
+
+The vite plugin supports local development via secrets using the [`.dev.vars` file convention](https://developers.cloudflare.com/workers/configuration/secrets/#local-development-with-secrets).
+
+Meaning that when you run `vite dev` or `vite preview` the plugin picks up secrets set in your `.dev.vars` files and makes them available to your application.
+
+To instead make such secrets available to deployed applications follow the [relative Cloudflare documentation](https://developers.cloudflare.com/workers/configuration/secrets/#secrets-on-deployed-workers).
+
+> [!Warning]
+> Currently `vite preview` does not actually pick up `.dev.vars` files, the team is working on it and that should be available soon
```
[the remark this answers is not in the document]
dario-piotrowicz: done 👍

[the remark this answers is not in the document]
jamesopstad: My preference is that we don't mention `.dev.vars` in the docs until it is fully supported. I think having warnings can come across as a bit intense and put people off. There shouldn't be too much work to do once we've decided on what the best approach is for supporting them in preview. Let's have that discussion and then hopefully it can also be done in this PR.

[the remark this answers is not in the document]
dario-piotrowicz: removed as the `vite preview` support has been added (d46955c41dff3d44ae4b26c1bc324bced2c43511)

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -297,3 +307,41 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
...
 	};
 }
+
+/**
+ * Copies a potential `.dev.vars` target over to the worker's output directory
+ * (so that it can get picked up by `vite preview`)
+ *
+ * Note: This resolves the .dev.vars file path following the same logic
+ *       as `loadDotEnv` in `/packages/wrangler/src/config/index.ts`
+ *       the two need to be kept in sync
```
dario-piotrowicz, line 317: this is not ideal.... ideally we should share the functionality 😓

we could do that by exposing some API from wrangler but that would feel a bit gnarly to me (since this seems like internal logic more than anything else to me)

in my opinion, one of the biggest strengths in having a monorepo setup would be to share code... so I would really consider having some internal package for shared logic such as this, in that way code could be shared very easily without needing to expose it publicly in any our packages

anyways that's quite overkill for something as simple thing, but anyways I wanted to mention it just because I think setting something like that can in general be very useful (and this need to share code does come up from time to time)

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -198,15 +198,25 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
 				}
 
 				config = workerConfig;
+
+				if (workerConfig.configPath) {
+					copyDotDevDotVarsFileToOutputDir(
+						workerConfig.configPath,
+						resolvedPluginConfig.cloudflareEnv,
+						this.emitFile
+					);
```
jamesopstad, line 207: This is really picky so feel free to ignore but my preference would be to call `this.emitFile` here and for the function to just return the content rather than also writing the file. Makes it easier to see what's happening.

```suggestion
					this.emitFile({
					  type: "asset",
					  fileName: ".dev.vars",
					  source: getDotDevDotVarsContent(
					    workerConfig.configPath,
					    resolvedPluginConfig.cloudflareEnv
					  )
					});
```

packages/vite-plugin-cloudflare/playground/dev-vars/tsconfig.node.json
```diff
@@ -0,0 +1,4 @@
+{
+	"extends": ["@cloudflare/workers-tsconfig/base.json"],
+	"include": ["vite.config.ts", "__tests__"]
```
jamesopstad, line 3:
```suggestion
	"include": ["vite.config.ts", "vite.config.with-specified-env.ts", "__tests__"]
```

packages/vite-plugin-cloudflare/playground/dev-vars/src/index.ts
```diff
@@ -0,0 +1,10 @@
+export default {
+	async fetch(_req, env) {
+		const { ENV_NAME, MY_DEV_VAR_A, MY_DEV_VAR_B, MY_DEV_VAR_C } = env;
+		const dotDevDotVarsVariables = { MY_DEV_VAR_A, MY_DEV_VAR_B, MY_DEV_VAR_C };
```
jamesopstad, line 4:
```suggestion
		const { ENV_NAME, ...dotDevDotVarsVariables } = env;
```

packages/vite-plugin-cloudflare/playground/dev-vars/package.json
```diff
@@ -0,0 +1,20 @@
+{
+	"name": "@playground/dev-vars",
+	"private": true,
+	"type": "module",
+	"scripts": {
+		"build": "vite build --app",
+		"check:types": "tsc --build",
+		"dev": "vite dev",
+		"prepreview": "node ./copy-dev-vars-to-dist.mjs",
```
jamesopstad, line 9: Is this still needed?

packages/vite-plugin-cloudflare/playground/dev-vars/.gitignore
```diff
@@ -0,0 +1 @@
+!.dev.vars
```
jamesopstad, line 1: Shall we do this in case others are added to the root `.gitignore` in the future.

```suggestion
!.dev.vars*
```

I'm wondering if we need to do this for `.env*` files that are used in playgrounds too now that we're in this repo?

packages/vite-plugin-cloudflare/README.md
```diff
@@ -382,6 +382,17 @@ It accepts an optional `PluginConfig` parameter.
...
+
+The vite plugin supports local development via secrets using the [`.dev.vars` file convention](https://developers.cloudflare.com/workers/configuration/secrets/#local-development-with-secrets).
+
+Meaning that when you run `vite dev` or `vite preview` the plugin picks up secrets set in your `.dev.vars` files and makes them available to your application.
+
+> [!NOTE]
+> The `vite build` command copies the potential `.dev.vars` target file into the same output directory in which it creates the build output `wrangler.json`, this copied file is the one that `vite preview` will then use
+
+To instead make such secrets available to deployed applications follow the [relative Cloudflare documentation](https://developers.cloudflare.com/workers/configuration/secrets/#secrets-on-deployed-workers).
+
```
jamesopstad, line 395: I think this section can be replaced as suggested below and that it should go after the 'Cloudflare environments' section.

```suggestion
## Secrets

Secrets can be provided to your Worker in local development using the [`.dev.vars`](https://developers.cloudflare.com/workers/configuration/secrets/#local-development-with-secrets) file convention. If you are using [Cloudflare Environments](#cloudflare-environments) then the relevant `.dev.vars` file will be selected. For example, `CLOUDFLARE_ENV=staging vite dev` will load `.dev.vars.staging` if it exists and otherwise fall back to `.dev.vars`.

> [!NOTE]
> The `vite build` command copies the relevant `.dev.vars` file to the output directory. This is only used when running `vite preview` and is not deployed with your Worker.
```

packages/vite-plugin-cloudflare/README.md
```diff
@@ -414,6 +425,11 @@ This is because the environment name is automatically appended to the top-level
 > The default Vite environment name for a Worker is always the top-level Worker name.
 > This enables you to reference the Worker consistently in your Vite config when using multiple Cloudflare environments.
 
+> [!NOTE]
+> Setting `CLOUDFLARE_ENV` also effects the `.dev.vars` being read, as files ending with a dot and the environment
+> name get prioritized over the standard `.dev.vars` ones (e.g. if I run `CLOUDFLARE_ENV=staging vite dev`, the vite plugin will
+> look for a `.dev.vars.staging` file to gather the secrets from, falling back to `.dev.vars` in case the file doesn't exist)
+
```
jamesopstad, line 432: This can be removed once the 'Secrets' section has been updated and moved.

```suggestion
```

.changeset/orange-camels-hunt_wrangler.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": minor
+---
+
+allow `unstable_getMiniflareWorkerOptions` to always accept an `env` argument
+
+the `unstable_getMiniflareWorkerOptions` utility, when accepting a config object as its first argument,
+doesn't accept an `env` one, the changes here make sure it does since even if a config object is passed
+the `env` one is still relevant for picking up variables from potential `.dev.vars` files
```
jamesopstad, line 9:
```suggestion
Update the `unstable_getMiniflareWorkerOptions` types to always include an `env` parameter.

The `unstable_getMiniflareWorkerOptions` types, when accepting a config object as the first argument,
didn't accept a second `env` argument. The changes here make sure they do, since the `env` is still relevant for picking up variables from potential `.dev.vars` files.
```

.changeset/orange-camels-hunt_vite-plugin.md
```diff
@@ -0,0 +1,9 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+add full support for `.dev.vars` files
+
+this change makes sure that `.dev.vars` files work for a specified environment, it also
+copies the target `.dev.vars` file (which might be environment specific, eg: `.dev.vars.prod`)
+to the worker's dist directory, always named `.dev.vars` (so that `vite preview` can pick it up)
```
jamesopstad, line 9:
```suggestion
Add full support for `.dev.vars` files.

This change makes sure that `.dev.vars` files work when the environment is specified. It also
copies the target `.dev.vars` file (which might be environment specific, e.g. `.dev.vars.prod`)
to the worker's dist directory so that `vite preview` can pick it up. The copied file will always be named `.dev.vars`.
```

jamesopstad (commented):
Looks good! Just a few code comments and some changes to the readme.

packages/vite-plugin-cloudflare/playground/dev-vars/.gitignore
dario-piotrowicz: > I'm wondering if we need to do this for .env* files that are used in playgrounds too now that we're in this repo?

We can, or we can just wait until the need arises, it is a single line addition anyways, we can do it at any point 👍

packages/vite-plugin-cloudflare/playground/dev-vars/package.json
dario-piotrowicz: damn, no sorry this was for some temporary hacking, thanks a lot for noticing it 🙏

packages/vite-plugin-cloudflare/playground/dev-vars/src/index.ts
dario-piotrowicz: I purposely didn't use the rest operation to be more explicit here

my version is I think also more future proof since later we can expand this app (adding other bindings for example) without breaking this test

I'm happy to go with the proposed code anyways (neither make a huge difference in my opinion) 👍

packages/vite-plugin-cloudflare/playground/dev-vars/tsconfig.node.json
dario-piotrowicz: good catch! 🙂👍

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -297,3 +307,41 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
...
+ */
+function copyDotDevDotVarsFileToOutputDir(
+	configPath: string,
+	cloudflareEnv: string | undefined,
+	emitFile: vite.Rollup.PluginContext["emitFile"]
+) {
+	const configDir = path.dirname(configPath);
+
+	const defaultDotDevDotVarsPath = `${configDir}/.dev.vars`;
+	const inputDotDevDotVarsPath = `${defaultDotDevDotVarsPath}${!cloudflareEnv ? "" : `.${cloudflareEnv}`}`;
```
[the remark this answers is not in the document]
dario-piotrowicz: yeah 👍

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: With your suggestion is sounds like you'd want to always save the `.dev.vars` file even if there is no content to save? That'd be ok, but I think it'd be nicer and clearer if we saved the file only when actually needed (keep in mind that `this.emitFile` also logs the information to the user)

So I've applied a similar change: https://github.com/cloudflare/workers-sdk/pull/7864/commits/3bb0dd1695d4b5cad543bfe4fb208c972d29fdcf

Let me know what you think 🙂

(I've also added a comment to clarify why we're saving such a file, although I understand that it is a bit unnecessary)

.changeset/orange-camels-hunt_wrangler.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": minor
+---
+
+Update the `unstable_getMiniflareWorkerOptions` types to always include an `env` parameter.
+
+The `unstable_getMiniflareWorkerOptions` types, when accepting a config object as the first argument,
+didn't accept a second `env` argument. The changes here make sure they do, since the `env` is still
+relevant for picking up variables from potential `.dev.vars` files.
```
vicb, line 9:
```suggestion
relevant for picking up variables from `.dev.vars` files.
```

packages/vite-plugin-cloudflare/README.md
```diff
@@ -466,6 +466,13 @@ The value of `MY_VAR` will therefore be `'Production var'`.
 If you run `vite build --mode staging` then the 'staging' Vite mode will be used and the 'staging' Cloudflare environment will be selected.
 The value of `MY_VAR` will therefore be `'Staging var'`.
 
+## Secrets
+
+Secrets can be provided to your Worker in local development using the [`.dev.vars`](https://developers.cloudflare.com/workers/configuration/secrets/#local-development-with-secrets) file convention. If you are using [Cloudflare Environments](#cloudflare-environments) then the relevant `.dev.vars` file will be selected. For example, `CLOUDFLARE_ENV=staging vite dev` will load `.dev.vars.staging` if it exists and otherwise fall back to `.dev.vars`.
```
vicb, line 471:
```suggestion
Secrets can be provided to your Worker in local development using a [`.dev.vars`](https://developers.cloudflare.com/workers/configuration/secrets/#local-development-with-secrets) file. If you are using [Cloudflare Environments](#cloudflare-environments) then the relevant `.dev.vars` file will be selected. For example, `CLOUDFLARE_ENV=staging vite dev` will load `.dev.vars.staging` if it exists and otherwise fall back to `.dev.vars`.
```

packages/vite-plugin-cloudflare/README.md
```diff
@@ -466,6 +466,13 @@ The value of `MY_VAR` will therefore be `'Production var'`.
...
+
+> [!NOTE]
+> The `vite build` command copies the relevant `.dev.vars` file to the output directory. This is only used when running `vite preview` and is not deployed with your Worker.
```
vicb, line 474: is this relevant?
```suggestion
> The `vite build` command copies the relevant `.dev.vars[.env-name]` file to the output directory. This is only used when running `vite preview` and is not deployed with your Worker.
```

packages/vite-plugin-cloudflare/playground/dev-vars/.dev.vars.staging
```diff
@@ -0,0 +1,3 @@
+ENV_NAME = "staging"
+MY_DEV_VAR_A = "my .dev.vars staging variable A"
+MY_DEV_VAR_B = "my .dev.vars staging variable B"
```
vicb, line 3:
```suggestion
MY_DEV_VAR_A = "my .dev.vars.staging variable A"
MY_DEV_VAR_B = "my .dev.vars.staging variable B"
```

packages/vite-plugin-cloudflare/playground/dev-vars/package.json
```diff
@@ -0,0 +1,19 @@
...
+		"build": "vite build --app",
+		"check:types": "tsc --build",
+		"dev": "vite dev",
+		"preview": "vite preview"
+	},
+	"devDependencies": {
+		"@cloudflare/vite-plugin": "workspace:*",
+		"@cloudflare/workers-tsconfig": "workspace:*",
+		"@cloudflare/workers-types": "^4.20241230.0",
+		"typescript": "catalog:vite-plugin",
```
vicb, line 15: Note, mostly unrelated to this PR:

The `vite-plugin` catalog pins `@types/node` but that will not work as expected because of

```json
		"overrides": {
			// ...      
			"@types/node": "$@types/node"
		},
```

So there would need to be a more specific override there for the plugin package

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -198,15 +198,33 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
...
 
 				config = workerConfig;
+
+				if (workerConfig.configPath) {
+					const dotDevDotVarsContent = getDotDevDotVarsContent(
+						workerConfig.configPath,
+						resolvedPluginConfig.cloudflareEnv
+					);
+					// if present we save a .dev.vars file onto the worker's build output
+					// directory so that it will be then detected by `vite preview`
```
vicb, line 208: I find the following clearer

```suggestion
					// Save a .dev.vars file to the worker's build output directory 
					// when it exists so that it will be then detected by `vite preview`
```

So when I read those kind of comments, I only wonder and what if not???

vicb: Oh well, looks like I misunderstood the comment... so maybe it wasn't clear :)

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -198,15 +198,33 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
...
 
 				assetsOnlyConfig.assets.directory = ".";
 
+				const filesToAssetsIgnore = ["wrangler.json", ".dev.vars"];
+
 				this.emitFile({
 					type: "asset",
 					fileName: ".assetsignore",
-					source: "wrangler.json",
+					source: `${filesToAssetsIgnore.join("\n")}\n`,
```
vicb, line 227: is there any benefit in having a variable here (i.e. do we expect to add more files)
Otherwise it looks looks inlining would be less convoluted

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -82,7 +83,11 @@ export function resolvePluginConfig(
 	const configPaths = new Set<string>();
 	const persistState = pluginConfig.persistState ?? true;
 	const root = userConfig.root ? path.resolve(userConfig.root) : process.cwd();
-	const { CLOUDFLARE_ENV } = vite.loadEnv(viteEnv.mode, root, "");
+	const { CLOUDFLARE_ENV: cloudflareEnv } = vite.loadEnv(
+		viteEnv.mode,
+		root,
+		""
+	);
```
vicb, line 90: rant:

Seeing this annoys me, same thing as with `function(path, config, true, true, false, "")`

```
	const { CLOUDFLARE_ENV: cloudflareEnv } = vite.loadEnv(
		viteEnv.mode,
		root,
		/* what is this??? usually the name of the parameter */""
	);
```

vicb (approved):
Looks good - with my limited understanding of the plugin.

But I always find things to comment ;)

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: the benefit of the variable are:
 - it is nicer to have if we can indeed add more files in the future
 - if is a self documenting variable, the variable name helps (at least to me) understanding that it does
 
 those are not huge benefits I don't mind inlining the array if that's what you strongly prefer

packages/vite-plugin-cloudflare/src/plugin-config.ts
dario-piotrowicz: I agree, this could be cleared if we owned the API and used a config object instead

with Vite's API we can still make this clearer by having a temporary documenting variable, but that might not be what you personally want since [before you said that you find such a variable "convoluted"](https://github.com/cloudflare/workers-sdk/pull/7864/files/5cc21f3adf1dfddd3f45d2b4da1f7c87fb8af648#r1929161497)

packages/vite-plugin-cloudflare/playground/dev-vars/.dev.vars.staging
dario-piotrowicz: I prefer not to have the `.` so that it is clearer that this this is a manually inputted string (otherwise without seeing the implementation someone could thing that this was generated by something like `my ${devVarsFileName} variable A`)

if you find it confusing I can change the strings to something like `MY_DEV_VAR_A = "my .dev.vars variable A for the staging environment` (or something like that)

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Yes, you're right. Good point. I like the way you've done it now.

packages/vite-plugin-cloudflare/src/plugin-config.ts
petebacondarwin: I seem to remember that some Google code guideline encouraged putting an inline comment next to this sort of magic string...

```suggestion
	const { CLOUDFLARE_ENV: cloudflareEnv } = vite.loadEnv(
		viteEnv.mode,
		root,
		/* prefixes */ ""
	);
```

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: We could certainly move this code into the wrangler package and reuse it (in the future).

Meanwhile this is pretty simple and intuitive to make it OK to have a copy here.

Can you put a comment in the Wrangler code to ensure that if that logic changes they should also update here?

petebacondarwin (approved):

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: > Can you put a comment in the Wrangler code to ensure that if that logic changes they should also update here?

that slipped my mind! good suggestion! thanks 🙂

packages/vite-plugin-cloudflare/src/plugin-config.ts
dario-piotrowicz: ah ok 🤔👍

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: done 🙂👍  https://github.com/cloudflare/workers-sdk/pull/7864/commits/744f80bfcfbba3315f3770e31a38776d64a863a7

PR #7867, opened 2025-01-22
Create revert-proxy-assets changeset

penalosa:


.changeset/revert-proxy-assets.md
```diff
@@ -0,0 +1,5 @@
+---
+wrangler: patch
+---
+
+Revert https://github.com/cloudflare/workers-sdk/pull/7816
```
petebacondarwin, line 5: For the future, I always find it helpful to explain why we reverted something.

petebacondarwin (approved):

.changeset/revert-proxy-assets.md
emily-shen: Haha here from the future, trying to hunt down the original bug report…

PR #7891, opened 2025-01-24
fix(C3): Use today's date for compat date when the latest workerd release date is in the future

jahands:
Fixes an issue where deployments would fail when there is a workerd release on the same day due to workerd releases having a date in the future because Workers does not allow using a compat date in the future.

I found this really frustrating that C3 templates fail to deploy if there is a workerd release on the same day:
image

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: small change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/create-cloudflare/src/helpers/__tests__/compatDate.test.ts
+++ b/packages/create-cloudflare/src/helpers/__tests__/compatDate.test.ts
@@ -4,7 +4,7 @@ import {
 	getWorkerdCompatibilityDate,
 } from "helpers/compatDate";
 import { getGlobalDispatcher, MockAgent, setGlobalDispatcher } from "undici";
-import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, test, vi } from "vitest";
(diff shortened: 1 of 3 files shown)


packages/create-cloudflare/src/helpers/__tests__/compatDate.test.ts
```diff
@@ -78,13 +79,29 @@ describe("Compatibility Date Helpers", () => {
 
 			const date = await getWorkerdCompatibilityDate();
 
-			const fallbackDate = "2023-05-18";
+			const fallbackDate = "2024-11-11";
```
jahands, line 82: Bumped the fallback date while I was here. THis was the latest date I found in https://developers.cloudflare.com/workers/configuration/compatibility-flags/ (not sure if we go off of that or workerd releases for this).

penalosa (approved):
Thanks for fixing this!

jahands:
@penalosa Mind kicking those failed e2e jobs? They look like they flaked

PR #7902, opened 2025-01-24
Implement verdaccio local npm registry for vitest-pool-workers tests

### LINKED ISSUE #7800: Ensure C3 E2E tests run against a locally built version of Wrangler
see https://github.com/cloudflare/workers-sdk/pull/7797 for reference

When our C3 e2e tests run, they (in a nutshell) build C3 out of the main branch, and create a starter project (via that previously built executable) for each framework under test. Once the starter project is created, certain assertions are made on it.

A starter project will come with a list of predefined dependencies, necessary for the project to run and be deployed, defined in `package.json`. All (I think) C3 starter projects will have a dependency on `wrangler`, specifically, on the latest released version of `wrangler`, `wrangler@latest`. This is crucial to note here, because while the C3 e2e tests are running in the "context" of a wrangler built off of `main` (let's call this version `N`), they are in fact making assertions against starter projects that were built with wrangler version `N-1` (`wrangler@latest`).

In our case this resulted in a false positive CI run, where C3 e2e tests were successful, when they should have instead failed

petebacondarwin:
Fixes #7800

This changes how e2e tests (C3 and vitest-pool-workers) make use of local packages.

**Reviewers: it is easier to follow along if you review commit by commit.**

This is achieved by spinning up a local mock npm registry in front of the public npm registry.
Then when you run `pnpm publish` or `pnpm i` it tries to use this local mock registry first.
This makes it easier to test the local versions of these packages, without having to manually set things up, which is error prone and fragile.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: actually updating tests
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: doesn't affect the Wrangler e2e tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: test change

### DIFF
--- a/packages/mock-npm-registry/README.md
+++ b/packages/mock-npm-registry/README.md
@@ -1,0 +1,35 @@
+# mock-npm-registry
+
+This internal package can be used in tests by othe packages to set up a mock local npm registry, publish locally built copies of the packages and their dependencies, then install those packages into test fixtures.
+
+## Usage
+
+A good example is the vitest-pool-workers e2e tests. See packages/vitest-pool-workers/test/global-setup.ts
+
+```ts
+export default async function ({ provide }: GlobalSetupContext) {
+	const stop = await startMockNpmRegistry("@cloudflare/vitest-pool-workers");
+
+	// Create temporary directory
+	const projectPath = await createTestProject();
+	execSync("pnpm install", { cwd: projectPath, stdio: "ignore" });
+
+	provide("tmpPoolInstallationPath", projectPath);
+```
+
+Here you can see that in Vitest global setup file, we are calling `startMockNpmRegistry()`.
+We pass in the `@cloudflare/vitest-pool-workers` package name, which will ensure that this package
+(and all its local dependencies, such as Wrangler, Miniflare, etc) is built and published to the mock registry.
+
+We then create a temporary test project that has a dependency on `@cloudflare/vitest-pool-workers`
+and then run `pnpm install`, which will install the locally published packages.
+
+The path to this temporary test project is provided to tests.
+
+## Debugging
+
+If you are having problems with the mock npm registry, you can get additional debug logging by setting the `NODE_DEBUG` environment variable.
+
+```env
+NODE_DEBUG=mock-npm-registry
+```
--- a/packages/mock-npm-registry/src/index.ts
+++ b/packages/mock-npm-registry/src/index.ts
@@ -1,0 +1,269 @@
+/* eslint-disable turbo/no-undeclared-env-vars */
+import childProcess, { execSync } from "node:child_process";
+import fs, { writeFile } from "node:fs/promises";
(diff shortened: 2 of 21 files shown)


lrapoport-cf:
https://github.com/cloudflare/workers-sdk/issues/7800

packages/mock-npm-registry/README.md
```diff
@@ -0,0 +1,35 @@
+# mock-npm-registry
+
+This internal package can be used in tests by othe packages to set up a mock local npm registry, publish locally built copies of the packages and their dependencies, then install those packages into test fixtures.
```
penalosa, line 3: "by other packages"

packages/mock-npm-registry/src/index.ts
```diff
@@ -0,0 +1,269 @@
...
+	);
+
+	let stopServer = await startVerdaccioServer(configPath);
+
+	const tempConfig = path.join(registryPath, ".npmrc");
+	await writeFile(
+		tempConfig,
+		dedent`
+			registry=http://localhost:${registryPort}
+			//localhost:${registryPort}/:_authToken=xxxx-xxxx-xxxx-xxxx
```
edmundhung, line 54: Is this `authToken` actually used? I don't see us setting it on Verdaccio. Or is it meant to ensure we don't accidentally publish to NPM?

packages/mock-npm-registry/src/index.ts
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	const revert_NPM_CONFIG_USERCONFIG = overrideProcessEnv(
+		"NPM_CONFIG_USERCONFIG",
+		tempConfig
+	);
+	const revert_npm_config_userconfig = overrideProcessEnv(
+		"npm_config_userconfig",
+		tempConfig
+	);
```
edmundhung, line 70: Why do we need both?

edmundhung (approved):
This looks really good! Thanks for putting this together!

penalosa (approved):

packages/mock-npm-registry/src/index.ts
petebacondarwin: Well.....
It turns out that either of these can be set and node.js/npm will read either, using the lowercase one in preference to the upper case one. But in my testing I found that if there were both these and they were not the same it could cause problems in some cases - I can't remember exactly which. So it was safer to set them both.

packages/mock-npm-registry/README.md
petebacondarwin:
```suggestion
This internal package can be used in tests by other packages to set up a mock local npm registry, publish locally built copies of the packages and their dependencies, then install those packages into test fixtures.
```

PR #7907, opened 2025-01-25
refactor(wrangler): runWranglerDev always log error

vicb:
@petebacondarwin this is reverting [one change](https://github.com/cloudflare/workers-sdk/commit/e4716cc87893a0633bd2d00543b351e83e228970) you made a couple weeks ago.

I had CI (and local) that were impossible to debug without this.

I see that you mention that the logs would be printed on timeout anyway... but that's unless your test timeout is less than the timeout in `runLongLivedWrangler` - it is the case with the failures I had (30 vs 50s in updated file).

Adding `stderr` only should hopefully not be too spammy?

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/fixtures/shared/src/run-wrangler-long-lived.ts
+++ b/fixtures/shared/src/run-wrangler-long-lived.ts
@@ -87,9 +87,7 @@ async function runLongLivedWrangler(
 		chunks.push(chunk);
 	});
 	wranglerProcess.stderr?.on("data", (chunk) => {
-		if (process.env.WRANGLER_LOG === "debug") {
-			console.log(`[${command}]`, chunk.toString());
-		}
+		console.log(`[${command}]`, chunk.toString());
 		chunks.push(chunk);
 	});
 	const getOutput = () => Buffer.concat(chunks).toString();


petebacondarwin:
I am not convinced this is the right thing to do. When the tests are passing we do not want to see all these logs.
Locally it is enough to just add the WRANGLER_LOG env var and you get the logging back for debugging.

I'd prefer to look into how we could capture the logs in CI for tests that fail without constantly spamming the output.

petebacondarwin:
Rather worryingly, the fixtures that depend upon this e.g. additional-modules got a cache hit, which means that Turbo didn't realise that they had a dependency on `fixtures/shared`. So it is not easy to see in the current CI runs what the impact of this change is.

vicb:
This would only display errors though. 
Maybe we could have an extra option to silent errors when they are expected? 

Yes the problem is mostly CI as it used different platforms which make it impossible to test locally - in my case the error was reproducible locally but seeing the error on the CI would have speed up the process

petebacondarwin (approved):
OK so I had a play around and I think that the stderr is pretty minimal when there are no failures. Can you confirm that this would have actually shown you the errors you needed @vicb?

vicb:
> Can you confirm that this would have actually shown you the errors you needed @vicb?

Yes it does - thanks for merging!

PR #7909, opened 2025-01-26
Allow unsafe param to pass through vite build

byule:
@jamesopstad this allows unsafe params to pass through the wrangler.json file while using vite as long as unsafe isn't set to an empty object. At least according to the comment, this was done for a reason to prevent empty objects from passing through, but seemed to do it unconditionally (even if not empty).

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: don't apply to the vite plugin
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: small bugfix

### DIFF
--- a/.changeset/old-bees-design.md
+++ b/.changeset/old-bees-design.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+Support unsafe params
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -218,8 +218,10 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin {
 
 			config.no_bundle = true;
 			config.rules = [{ type: "ESModule", globs: ["**/*.js"] }];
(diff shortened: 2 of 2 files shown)


jamesopstad (approved):
Thanks Ben! I'll make a note about fixing the root of the issue too so we can remove this.

jamesopstad:
It turns out that @petebacondarwin merged a PR to allow the empty object for `unsafe` in Wrangler in December (https://github.com/cloudflare/workers-sdk/pull/7461). Sorry I forgot about that. I still think we need to only pass it through if it's explicitly set by the user, however, or they will get the `"unsafe" fields are experimental and may change or break at any time.` error when they deploy. Happy to merge this as is for now and then we can have another look at the behaviour in Wrangler (perhaps to set the default `unsafe` value to `undefined` rather than an empty object).

jamesopstad:
> @jamesopstad this could probably do with a test, but I'm not super familiar with the testing setup for the vite plugin—do you know where the right place to add one would be?

Most tests are written as integration tests in the `playground` directory. Probably not necessary for this though as the fix is just to avoid a warning.

PR #7915, opened 2025-01-27
Sync unenv-preset with unenv@2.0.0-rc.8

### LINKED ISSUE #8111: 🚀 Feature Request: [unenv-preset] implement `globalThis.performance` in nodejs_compat mode
### Describe the solution

This issue is ported from https://github.com/unjs/unenv/issues/320

`globalThis.performance` comes from `workerd` and as such is missing quite a few methods.

We should use the `unenv` polyfill - there is a circular dependency that was blocking us.

Needs some investigation

vicb:
Sync `@cloudflare/unenv-preset` with `unenv@2.0.0-rc.8`

Changelog:
- The preset is now esm only
- Polyfill perf_hooks (fixes #8111)

Notes: 

The ESBuild plugin needs to be updated because
- The preset now only publish ESM
- To support the `polyfill` property of the preset `Environment`

You can check the [WIP/test-preset](https://github.com/cloudflare/workers-sdk/tree/WIP/test-preset) branch that includes the updates and test the preset.

I have deliberately not include included a changelog because this package should not be released before 2.0 is ready

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: the preset is not used yet in wrangler
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/unenv-preset/package.json
```diff
@@ -47,15 +47,15 @@
 	},
 	"devDependencies": {
 		"@types/node-unenv": "npm:@types/node@^22.10.5",
-		"typescript": "catalog:default",
+		"typescript": "^5.7.2",
 		"unbuild": "^3.2.0",
 		"undici": "catalog:default",
+		"unenv": "file:/Users/vberchet/code/work/unenv",
```
pi0, line 53: Really not sure about effect of this on resolution but it won't be same as normal installations in `node_modules`.

We should adopt it with next rc (or nightly if you have it still working here)

vicb: Yep, that was just pushing the changes that I have for (p)review.
This PR is in draft mode, I just have been more clear in the description.

anonrig:
@vicb Can you rebase?

vicb:
> @vicb Can you rebase?

I will, probably tomorrow. For now this is only a preview syncing with the next version of unenv

vicb:
Drop timers + rebased

pnpm-lock.yaml
```diff
@@ -1686,9 +1686,6 @@ importers:
 
   packages/unenv-preset:
     dependencies:
-      unenv:
-        specifier: 2.0.0-rc.1
-        version: 2.0.0-rc.1
```
petebacondarwin, line 1691: What's happening here?

vicb: `pnpm link` to test the dev version as the new structure has not been released to npm.
Will need to be updated after the release and then remove draft from the PR

vicb:
@petebacondarwin I have started to sync with unenv - there are still a few changes to sync. I also have 2 PRs on unenv that are pendings.

This PR also needs to be tested with a version of wrangler actually using the preset - currently `wrangler` (then also `pnpm -F @cloudflare/unenv-preset test:ci`) is using `unjs/unenv`.

I hope to be done with that later today so that we can release `@cloudflare/unenv-preset@2` this week. This would be a prerequisite before switching `wrangler` back to the preset.

packages/unenv-preset/src/preset.ts
```diff
@@ -64,29 +65,25 @@ export const cloudflare: Preset = {
 		),
```
petebacondarwin, line 65: Out of curiosity, why do we create all these identity aliases? It seems a bit pointless?
Shouldn't we only add aliases for things that need to be mapped to polyfills?
(I could understand creating `${p}` -> `node:${p}` perhaps?)

packages/unenv-preset/src/preset.ts
```diff
@@ -64,29 +65,25 @@ export const cloudflare: Preset = {
...
-				[m, `@cloudflare/unenv-preset/runtime/node/${m}/index`],
-				[`node:${m}`, `@cloudflare/unenv-preset/runtime/node/${m}/index`],
+				[m, `@cloudflare/unenv-preset/node/${m}`],
+				[`node:${m}`, `@cloudflare/unenv-preset/node/${m}`],
 			])
 		),
-
-		// TODO: this is a hotfix and breaks unenv/fetch
-		// https://github.com/unjs/unenv/issues/364
-		"unenv/runtime/node/stream/index": "node:stream",
```
petebacondarwin, line 80: How was this resolved? I see that the polyfill was removed in https://github.com/unjs/unenv/pull/411 but it is difficult to see how that fixes this.

packages/unenv-preset/src/preset.ts
```diff
@@ -64,29 +65,25 @@ export const cloudflare: Preset = {
...
 	},
 	inject: {
-		// workerd already defines `global` and `Buffer`
-		// override the previous presets so that we use the native implementation
```
petebacondarwin, line 84: I quite liked this comment.

But perhaps you removed it because we are no longer piggy backing on top of the nodeless preset?
In which case can we get rid of

```
		Buffer: false,
		global: false,
```

too?

packages/unenv-preset/src/runtime/node/module.ts
```diff
@@ -53,17 +53,20 @@ export {
...
 	file: string | URL
 ) => {
 	return Object.assign(workerdModule.createRequire(file), {
-		resolve: Object.assign(notImplemented("module.require.resolve"), {
-			paths: notImplemented("module.require.resolve.paths"),
-		}),
+		resolve: Object.assign(
+			/*@__PURE__*/ notImplemented("module.require.resolve"),
+			{
+				paths: /*@__PURE__*/ notImplemented("module.require.resolve.paths"),
```
petebacondarwin, line 67: The idea, here, being that by marking these as PURE the bundlers can remove these if not used?

petebacondarwin (approved):
Nice work! Got there finally!
As well as updating Wrangler, we'll need to update Vite plugin.

petebacondarwin:
Is it worth adding a changeset now?

packages/unenv-preset/src/preset.ts
vicb: I have added this comment:

```
		// `nodeCompatModules` are implemented in workerd.
		// Create alias to override polyfills defined in based environments.
```

Does it help?

packages/unenv-preset/src/preset.ts
vicb: unenv used to import from `unenv/runtime/node/stream/index` to get a stream implementation needed for direct fetch. It was updated to import from `node:stream`.
This is why we do need this alias to use the workerd implementation of the stream module.

packages/unenv-preset/src/preset.ts
vicb: I added back `// Setting symbols implemented by workerd to `false` so that `inject`s defined in base presets are not used.`

packages/unenv-preset/src/runtime/node/module.ts
vicb: That's exact.
This is reflecting the many `/*@__PURE__*/` annotation added by @pi0 in unenv@2

vicb:
> Is it worth adding a changeset now?

I'll first merge the PR, test the PR releases and if that looks good we can add a changeset

packages/unenv-preset/src/runtime/node/async_hooks.ts
pi0, line 1: I noticed latest prod release of workerd has all `async_hooks` (nitro nightly already has `async_hooks` as full-native.

packages/unenv-preset/src/runtime/node/module.ts
pi0: I think it won't affect anything for bundlers when used _inside_ functions (might be accident if I also added it in unenv). PURE is for module top level context.

packages/unenv-preset/src/runtime/node/async_hooks.ts
vicb: I'll create an issue to tackle new additions in a follow up PR (async_hooks + crypto)

vicb: https://github.com/cloudflare/workers-sdk/issues/8297

pi0: will workerd crypto add all missing exports :o 

from [platform-node-compat](https://platform-node-compat.pi0.workers.dev/) quite lot a missing:

image

packages/unenv-preset/src/runtime/node/module.ts
vicb: At least is will not harm, let's get the PR in and we can investigate later

Issue #7847, opened 2025-01-21
🚀 Feature Request: Vitest v3 support


SegaraRai:
### Describe the solution

Vitest has recently released v3 and it would be great if vitest-pool-workers would support Vitest v3.

When I tried Vitest v3 I got the following error message.
It seems that the test runner now depends on `node:timers` and `node:timers/promises`.

```
 DEV  v3.0.2 D:/path/to/project/packages/backend

[vpw:wrn] You're running `vitest@3.0.2`, but this version of `@cloudflare/vitest-pool-workers` only officially supports `vitest 2.0.x - 2.1.x`.
`@cloudflare/vitest-pool-workers` currently depends on internal Vitest APIs that are not protected by semantic-versioning guarantees.
Your tests may work without issue, but we can not guarantee compatibility outside of the above version range.
Using vars defined in .dev.vars
[vpw:inf] Starting single runtime for vitest.config.ts...
[mf:wrn] The latest compatibility date supported by the installed Cloudflare Workers Runtime is "2024-12-30",
but you've requested "2025-01-01". Falling back to "2024-12-30"...
workerd/server/server.c++:3319: error: Fallback service failed to fetch module; payload = ; spec = /?specifier=%2Fnode%3Atimers%2Fpromises&referrer=%2FD%3A%2Fpath%2Fto%2Fproject%2Fnode_modules%2F.pnpm%2Fvitest%403.0.2_%40types%2Bnode%4022.10.7%2Fnode_modules%2Fvitest%2Fdist%2Fchunks%2FrunBaseTests.Ba8jtu6O.js&rawSpecifier=node%3Atimers%2Fpromises
workerd/server/server.c++:3319: error: Fallback service failed to fetch module; payload = ; spec = /?specifier=%2Fnode%3Atimers%2Fpromises&referrer=%2FD%3A%2Fpath%2Fto%2Fproject%2Fnode_modules%2F.pnpm%2Fvitest%403.0.2_%40types%2Bnode%4022.10.7%2Fnode_modules%2Fvitest%2Fdist%2Fchunks%2FrunBaseTests.Ba8jtu6O.js&rawSpecifier=node%3Atimers%2Fpromises
Error running worker: Error: No such module "node:timers/promises".
  imported from "D:/path/to/project/node_modules/.pnpm/vitest@3.0.2_@types+node@22.10.7/node_modules/vitest/dist/chunks/runBaseTests.Ba8jtu6O.js"
[...]

PR #7923, opened 2025-01-27
Support Vitest 3

penalosa:
Fixes https://github.com/cloudflare/workers-sdk/issues/7847

Update `@cloudflare/vitest-pool-workers` to support Vitest v3. Additionally, support the `vitest list` API.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: existing tests provide coverage
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/issues/19990


packages/vitest-pool-workers/src/worker/lib/cloudflare/test-runner.ts
```diff
@@ -209,15 +209,6 @@ export default class WorkersTestRunner extends VitestTestRunner {
 		initialState.current = this.state.current;
 	}
 
-	async onBeforeRunFiles() {
```
penalosa, line 212: This has been removed in Vitest v3

edmundhung: It is a bit strange that `onBeforeRunFiles` is removed while we are still using `onAfterRunFiles` 🤔 

Any chance it is a different issue? I can also find `onBeforeRunFiles` on vitest docs:
https://github.com/vitest-dev/vitest/blob/9acc13f367988ff5711f51d8b360a7f6d6602d2c/docs/advanced/runner.md?plain=1#L69-L76

packages/vitest-pool-workers/src/worker/lib/cloudflare/test-runner.ts
```diff
@@ -209,15 +209,6 @@ export default class WorkersTestRunner extends VitestTestRunner {
...
-		if (DEBUG) {
-			__console.log("onBeforeRunFiles");
-			await scheduler.wait(100);
-		}
-
-		resetMockAgent(fetchMock);
```
edmundhung, line 218: We will need to figure out where to reset the MockAgent if `onBeforeRunFiles` is removed

packages/vitest-pool-workers/src/worker/lib/cloudflare/test-runner.ts
penalosa: It's still in the docs, but it's been removed from the actual code

penalosa: https://github.com/vitest-dev/vitest/blob/9acc13f367988ff5711f51d8b360a7f6d6602d2c/packages/vitest/src/runtime/runners/test.ts

SegaraRai:
At the risk of meddling, I suggest updating Vitest to >= v3.0.5 before merging. pnpm-lock contains v3.0.4, which contains [an RCE vulnerability](https://github.com/vitest-dev/vitest/security/advisories/GHSA-9crc-q9x8-hgqq) recently discovered in Vitest.

fixtures/shared/src/mock-console.ts
```diff
@@ -0,0 +1,60 @@
...
+
+function normalizeOutput(spy: MockInstance, join = "\n"): string {
+	return captureCalls(spy, join);
+}
+
+function captureCalls(spy: MockInstance, join = "\n"): string {
+	return spy.mock.calls
+		.map((args: unknown[]) => util.format("%s", ...args))
+		.join(join);
+}
```
edmundhung, line 42: Nit: What do we need an extra `captureCalls` helper here? I don't see it being used outside of `normalizeOutput`.

packages/wrangler/src/__tests__/d1/convert-timestamp-to-iso.test.ts
```diff
@@ -3,7 +3,7 @@ import { convertTimestampToISO } from "../../d1/timeTravel/utils";
 
 describe("convertTimestampToISO", () => {
 	beforeAll(() => {
-		vi.useFakeTimers();
+		vi.useFakeTimers({ toFake: ["setTimeout", "clearTimeout", "Date"] });
```
edmundhung, line 6: Why do we need to specify these? It feels like we are coupling the test with the implementation if we have to specify the methods to fake.

packages/wrangler/src/index.ts
```diff
@@ -1172,10 +1172,12 @@ export async function main(argv: string[]): Promise<void> {
...
 			await closeSentry();
+			const controller = new AbortController();
+
 			await Promise.race([
-				await Promise.allSettled(dispatcher?.requests ?? []),
-				setTimeout(1000), // Ensure we don't hang indefinitely
-			]);
+				Promise.allSettled(dispatcher?.requests ?? []),
+				setTimeout(1000, undefined, controller), // Ensure we don't hang indefinitely
+			]).then(() => controller.abort()); // Ensure the Wrangler process doesn't hang waiting for setTimeout(1000) to complete
```
edmundhung, line 1180: Super Nit: `Promise.allSettled` won't throw, so this shouldn't have any actual difference 🤔 

```suggestion
			]).finally(() => controller.abort()); // Ensure the Wrangler process doesn't hang waiting for setTimeout(1000) to complete
```

edmundhung (approved):
LGTM once you have the changeset ready 👍🏼

packages/wrangler/src/__tests__/d1/convert-timestamp-to-iso.test.ts
penalosa: This restores the Vitest 2 defaults: https://vitest.dev/guide/migration#fake-timers-defaults. Without this, our tests were hanging (presumably because `performance.now()` was not previously being mocked?

fixtures/shared/src/mock-console.ts
penalosa: Are you suggesting it should be inline in `normalizeOutput()`?

edmundhung: Yes. But I guess you are just following the patterns. nevermind!

packages/wrangler/src/index.ts
edmundhung: Not really, just so we can guarantee the controller is aborted if it throws for whatever reasons. But it shouldn't have any actual difference with the current promise setup 🤷

edmundhung:
I might have missed something. But it looks like the only changes needed within `vitest-pool-workers` is to stop calling `super.onBeforeRunFiles()`. Is it worth keeping Vitest 2 support?

We could technically check whether `super.onBeforeRunFiles` is defined before calling it instead of removing it.

PR #7925, opened 2025-01-27
Add workflow for uploading asset-worker and router-worker

WillTaylorDev:
Creates internal workflow to allow us to upload new versions of router-worker and asset-worker without an explicit wrangler release.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: internal workflow

### DIFF
--- a/.github/workflows/workers-shared-deploy-production.yml
+++ b/.github/workflows/workers-shared-deploy-production.yml
@@ -1,0 +1,21 @@
+name: Deploy Workers Shared Production
+
+on:
+  workflow_dispatch:
+
+jobs:
+  deploy-workers:
+    name: Deploy Workers Shared Production
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Install Dependencies
+        uses: ./.github/actions/install-dependencies
+
+      - name: Deploy
+        run: pnpm deploy
+        working-directory: packages/workers-shared
+        env:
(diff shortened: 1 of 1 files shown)


WalshyDev (approved):

.github/workflows/workers-shared-deploy-production.yml
```diff
@@ -0,0 +1,21 @@
+name: Deploy Workers Shared Production
+
+on:
```
CarmenPopoviciu, line 3: I think it should be enough to just say 👇  based on https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch

```suggestion
on: workflow_dispatch
```

CarmenPopoviciu (approved):

CarmenPopoviciu (changes requested):
pending on an internal discussion

.github/workflows/workers-shared-deploy-production.yml
WillTaylorDev: image
Unfortunately has to be newline'd :(

edmundhung (approved):
for posterity, this is discussed within the team and we confirmed both workers are not deployed directly to production as the deploy script is only doing a version upload.

@WillTaylorDev FYI, The changeset workflow will still upload a new version if there is any changes on `@cloudflare/workers-shared`. You might need to modify the setup if you don't want it.

PR #7927, opened 2025-01-28
refactor(wrangler): generate bundle meta file

### LINKED ISSUE #4633: 🚀 Feature Request: Option to output esbuild metafile

vicb:
Fixes #4633 

Hey @penalosa or @petebacondarwin, could you please point me to the best place to add tests for that?
I tested locally and it works fine but we need to add test.
Thanks!

_I'll also add a changeset before merging._

---

- Tests
  - [x] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: I'll update the docs once merged

### DIFF
--- a/packages/wrangler/src/__tests__/pages/functions-build.test.ts
+++ b/packages/wrangler/src/__tests__/pages/functions-build.test.ts
@@ -161,11 +161,21 @@ describe("pages functions build", () => {
 
 		expect(readdirSync("dist").sort()).toMatchInlineSnapshot(`
 		Array [
+		  "bundle-meta.json",
 		  "e8f0f80fe25d71a0fc2b9a08c877020211192308-name.wasm",
 		  "f7ff9e8b7bb2e09b70935a5d785e0cc5d9d0abf0-greeting.wasm",
 		  "index.js",
 		]
 	`);
+
+		// `bundle-meta.json` contains ESBuild build metadata
+		// see https://esbuild.github.io/api/#build-metadata
+		const meta = JSON.parse(
+			readFileSync("dist/bundle-meta.json", { encoding: "utf8" })
+		);
+
+		expect(meta.inputs).toBeDefined();
+		expect(meta.outputs).toBeDefined();
 	});
 
 	it("should build _worker.js", async () => {
--- a/packages/wrangler/src/deployment-bundle/bundle.ts
+++ b/packages/wrangler/src/deployment-bundle/bundle.ts
@@ -398,16 +398,18 @@ export async function bundleWorker(
 		entryPoints: [entry.file],
 		bundle,
 		absWorkingDir: entry.projectRoot,
-		outdir: destination,
 		keepNames: true,
-		entryNames: entryName || path.parse(entryFile).name,
 		...(isOutfile
 			? {
 					outdir: undefined,
 					outfile: destination,
 					entryNames: undefined,
 				}
-			: {}),
+			: {
+					outdir: destination,
+					outfile: undefined,
+					entryNames: entryName || path.parse(entryFile).name,
+				}),
 		inject,
 		external: bundle
 			? ["__STATIC_CONTENT_MANIFEST", ...(external ? external : [])]
@@ -483,6 +485,19 @@ export async function bundleWorker(
 			};
 		} else {
 			result = await esbuild.build(buildOptions);
+			// Write the bundle metafile to disk.
+			if (result.metafile) {
+				let metaFilePath: string | undefined;
+				if (buildOptions.outdir) {
+					metaFilePath = path.join(buildOptions.outdir, "bundle-meta.json");
+				} else if (buildOptions.outfile) {
+					metaFilePath = buildOptions.outfile + ".bundle-meta.json";
+				}
+				if (metaFilePath) {
+					const metaJson = JSON.stringify(result.metafile, null, 2);
+					fs.writeFileSync(metaFilePath, metaJson);
+				}
+			}
 			// Even when we're not watching, we still want some way of cleaning up the
 			// temporary directory when we don't need it anymore
 			stop = async function () {


vicb:
I added a test.

Could anybody please confirm that generating the file in the `outdir` will not get it uploaded?

petebacondarwin:
> I added a test.
> 
> Could anybody please confirm that generating the file in the `outdir` will not get it uploaded?

That will depend upon whether it matches a "additional module rule" and whether `find_additional_modules` and/or `no_bundle` are true.

petebacondarwin (changes requested):
I think this should be explicitly opt-in rather than inferred from the outdir.
E.g. `--metafile` or similar

The opt-in would default to false. A value equivalent to `true` could default to putting the metafile in the outdir if no other value is provided.

E.g. `--metafile=path/to/metafile.json`.

vicb:
Thanks for the review @petebacondarwin, I have a few follow up questions:

> >Could anybody please confirm that generating the file in the outdir will not get it uploaded?
>
> That will depend upon whether it matches a "additional module rule" and whether find_additional_modules and/or no_bundle are true.

Could you please point me to the code?
_(That was my fear but the additional modules should be looked for in the source folder, not the outdir)_

> I think this should be explicitly opt-in rather than inferred from the outdir.
> E.g. --metafile=path/to/metafile.json

 `--metafile=path/to/metafile.json` was my first thought.
 
Then I thought that we already have an `outdir` and we can re-use that, so it should probably just be `--metafile` (i.e. only to enable but do not offer to override the path).

I also see that we always generate the metadata so the only override here is to dump to disk. So I'm wondering if we really need that `--metafile` at all - which is what I implemented in this PR. I think simpler is better but happy to change that if needed.

petebacondarwin:
> Could you please point me to the code?
> (That was my fear but the additional modules should be looking for files in the source folder, not the outdir?)

It is not uncommon to run `wrangler build` to generate a compiled Worker and then `wrangler deploy` with `no_bundle` on to deploy from that output directory.

petebacondarwin:
> I also see that we always generate the metadata so the only override here is to dump to disk. So I'm wondering if we really need that --metafile at all - which is what I implemented in this PR. I think simpler is better but happy to change that if needed.

Yes I was proposing that we have a `--metafile` option that by itself would dump the metadata to disk in the `outdir` directory. I think this should still be opt-in, i.e. not to dump the file unless this flag is on.

vicb:
> It is not uncommon to run `wrangler build` to generate a compiled Worker and then `wrangler deploy` with `no_bundle` on to deploy from that output directory.

Oh I see, good point - and in this case the `.map` is not an issue because it doesn't match any of the implicit rules.

I'll update the PR to add `--metafile` later this week.

DaniFoldi:
Hey @vicb,

Is there anything I can help with to land this? The feature request issue has been open for a long time, and I'd love to be able to see which dependency is increasing the bundle size (which is an important metric for startup time) without having to patch wrangler.

vicb:
> Hey @vicb,
> 
> Is there anything I can help with to land this? The feature request issue has been open for a long time, and I'd love to be able to see which dependency is increasing the bundle size (which is an important metric for startup time) without having to patch wrangler.

We need to add a `--metafile` flag to enable the generation of the file - do you want to get a stab at this?

ItsWendell:
> > Hey @vicb,
> > Is there anything I can help with to land this? The feature request issue has been open for a long time, and I'd love to be able to see which dependency is increasing the bundle size (which is an important metric for startup time) without having to patch wrangler.
> 
> We need to add a `--metafile` flag to enable the generation of the file - do you want to get a stab at this?

I did a attempt here, I added a simple `--metafile` flag and adjusted / added some more tests. I was doubting whether to allow the --metafile flag to be string to change the location but didn't see any value in that in this case since the build process is highly managed by the worker itself anyway.

Curious on how I could best document this further for docs, and whether I missed a build path if you have any insights into this.

vicb:
Superseded by https://github.com/cloudflare/workers-sdk/pull/8716

Issue #7886, opened 2025-01-23
🐛 BUG: `process.env.NODE_ENV` is `null` but should be `'production'` instead.


brillout:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What versions are you using?

3.102.0 [Wrangler]

### What operating system and version are you using?

Ubuntu 24.04.1

### Describe the Bug

If I set the following:

```toml
# wrangler.toml

[vars]
NODE_ENV = "production"
```

Then I expect wrangler to replace `process.env.NODE_ENV` with `'production'`. It's a ubiquitous JavaScript convention.

I also tried the following, but it doesn't work either.

```toml
# wrangler.toml

define = { "process.env.NODE_ENV" = "production" }
```

I can't dig why it doesn't work as it doesn't seem possible to inspect the worker code as per https://github.com/cloudflare/workers-sdk/issues/7885.

vicb:
```toml
# wrangler.toml

[vars]
NODE_ENV = "production"
```

Makes `NODE_ENV` available on `env` (as in `fetch(request, env, ctx)`) - it is not the same as `process.env`

See https://developers.cloudflare.com/workers/runtime-apis/nodejs/process/#processenv for details

```
# wrangler.toml

define = { "process.env.NODE_ENV" = "production" }
```

I am not sure what `define` is used for, maybe @petebacondarwin knows?

brillout:
Is it even possible to set the `process.env.NODE_ENV` value then?

In my specific case, Vike shows a warning if it isn't `'production'`:

> [vike][Warning] The environment is set to be a development environment by process.env.NODE_ENV===null, which seems contradictory because the environment seems to be a production environment (Vite isn't loaded), see https://vike.dev/NODE_ENV

This is where Vike reads `process.env.NODE_ENV`: https://github.com/vikejs/vike/blob/bb8341b7144f1e9a1e9fa47a916042cb94dbd76a/vike/utils/assertSetup.ts#L141.

vicb:
> Is it even possible to set the `process.env.NODE_ENV` value then?

You can have `process.env.NODE_ENV = env.NODE_ENV;` in your code.
Note that as `env` in only available in the request context, top-level code might see `process.env.NODE_ENV === undefined`

Ideally we should support defining `process.env.NODE_ENV` in wrangler.json, I'm not sure if it is the intent of the `define` entry in the configuration file but if it is it looks like it doesn't work

brillout:
[...]

PR #7932, opened 2025-01-28
fix(wrangler): handling of `process.env.NODE_ENV` in bundling mode

vicb:
Relates to #7886.

Before this PR:

- The first check `process.env.NODE_ENV` should have checked the runtime value (as done a few lines below)
- This check should only (de)activate `define["process.env.NODE_ENV"]` and not `define` as a whole.

This fix a few points from the linked issue, not all.

Notably there is still an issue when process is imported (`import process from "node:process";`) as "process" will be renamed by the hybrid plugin and not imported!

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/lucky-jeans-relax.md
+++ b/.changeset/lucky-jeans-relax.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix(wrangler): handling of `process.env.NODE_ENV` in bundling mode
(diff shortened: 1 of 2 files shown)


penalosa (approved):

vicb:
Thanks for the review @penalosa !

PR #7942, opened 2025-01-28
remove functions that are removed from node:util

anonrig:
Removes methods that are removed from node.js

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: they are removed from node.js
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: removing deprecated/removed functions
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: they're removed from node.js


Cherry:
Is there a documented backwards compatibility strategy for `nodejs_compat`? If I have a worker in production with this:
```ts
import {isNumber} from 'node:util';

export default {
	async fetch(request, env, ctx): Promise<Response> {
		const bar = Math.random() > 0.5 ? 1 : 'foo';
		if(isNumber(bar)) {
			return new Response('Number!');
		}
		return new Response('Hello World!');
	},
} satisfies ExportedHandler<Env>;

```
which works completely fine with `nodejs_compat`, will it break after this lands in the latest version of wrangler? In the latest version of Node.js 22, `util.isNumber('cd')` works without issue, though is indeed marked as deprecated.

vicb (changes requested):
Let's wait before merging this...

2 reasons:

- First use the preset in this repo matching unenv and start diverging later
- We have at least one user telling us he is using the API

Idea: maybe we can add a flag (`--disable-node-deprecated-apis-v18`) or something.

(Requesting changes to make sure this doesn't get in too early)

Cherry:
> You will not have any issues when you redeploy. If you use a specific version of wrangler it will always come with the polyfill. On another hand, if you update wrangler, then you'll just see that isNumber doesn't exist anymore.
> 
> These functions are removed on April 29, 2024: [nodejs/node#52744](https://github.com/nodejs/node/pull/52744)

If I'm understanding correctly, these will be removed in Node.js 23, and therefore any future versions as well. This makes sense as it's a new major version bump, and breaking changes are expected.

wrangler / `nodejs_compat` wouldn't have that here, so expecting these functions to disappear in a minor version would not be a great DX. This is why I was asking primarily if there was a documented strategy around `nodejs_compat` versioning and backwards compatibility.

jasnell:
hmmm... I think we need to carefully think about this. If these were implemented in the runtime we would need to keep them around forever with or without a compatibility flag. I think breaking changes in polyfill implementations that we choose to bundle are fundamentally different from ones that the user chooses to bundle and the burden of support is significantly higher.

Instead of removing these like this, it would likely be a much better idea to noisily deprecate them for a few versions of wrangler, detecting when they are used and complaining loudly for a while before dropping them.

petebacondarwin:
@Cherry  makes a good point. This is actually a breaking change to Wrangler so should be rolled out in v4?

petebacondarwin (changes requested):
Let's hold on this removal and follow the path suggested by @vicb.

packages/unenv-preset/src/runtime/node/util/index.ts
```diff
@@ -1,51 +1,15 @@
...
 	styleText,
 } from "unenv/runtime/node/util/index";
 import type nodeUtil from "node:util";
 
 export {
-	_errnoException,
-	_exceptionWithHostPort,
 	getSystemErrorMap,
 	getSystemErrorName,
-	isArray,
```
vicb, line 33: Note: `isArray` is not removed from Node and the timeline to remove the other functions is Node 23.

anonrig: isArray is available and implemented on workerd

vicb: I should have commented on l94 instead.
What I meant is that `isArray` is not part of the functions removed in Node23.

But if we have some of the functions here implemented in workerd (and not going away) we should definitely use the workerd version instead - I guess it will be a different PR.

Thanks

PR #7947, opened 2025-01-28
[wrangler] fix: avoid `getPlatformProxy` logging twice that it is using vars defined in `.dev.vars` files

### LINKED ISSUE #7944: 🐛 BUG: `getPlatformProxy` logs `Using vars defined in .dev.vars` twice

Wrangler

3.106.0

MacOS Sequoia 15.2

https://github.com/dario-piotrowicz/getplatformproxy-devvars-double-log

### Describe the Bug

If I call wrangler's `getPlatformProxy` with `.dev.vars` file in my directory the message `Using vars defined in .dev.vars` gets logged twice

it should only be logged once

_No response_

dario-piotrowicz:
Fixes #7944

when `getPlatformProxy` is called and it retrieves values from `.dev.vars` files, it logs twice a message like: `Using vars defined in .dev.vars`, the changes here make sure that in such cases this log only appears once

(Prerequisite for fixing https://github.com/opennextjs/opennextjs-cloudflare/issues/288)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: this functionality is already tested, the only noticeable change here is the removal of the extra terminal log, I am not sure if that can be reliably tested, I can look into that if we believe it beneficial but it feels not completely worth it to me
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/gentle-foxes-float.md
+++ b/.changeset/gentle-foxes-float.md
@@ -1,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+fix: avoid `getPlatformProxy` logging twice that it is using vars defined in `.dev.vars` files
+
+when `getPlatformProxy` is called and it retrieves values from `.dev.vars` files, it logs twice
+a message like: `Using vars defined in .dev.vars`, the changes here make sure that in such cases
+this log only appears once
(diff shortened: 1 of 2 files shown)


petebacondarwin (approved):
The point being that `getMiniflareOptionsFromConfig()` now also reads the `.dev.vars` so there is no need to do it again in `getPlatformProxy()`?

dario-piotrowicz:
> The point being that `getMiniflareOptionsFromConfig()` now also reads the `.dev.vars` so there is no need to do it again in `getPlatformProxy()`?

yes, exactly 🙂👍

PR #7950, opened 2025-01-29
Add Worker Pipelines Binding to Miniflare

cmackenzie1:
Adding support to Miniflare to reference bindings during `wrangler dev` or during tests via vitest.

This binding uses newer JSRPC method instead of propagating changes through `workerd.capnp`.

Fixes https://jira.cfdata.org/browse/PIPE-166

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
  - [X] Not required because: no e2e test needed
- Public documentation
  - [ ] TODO (before merge)
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/18595

### DIFF
--- a/.changeset/slimy-otters-rush.md
+++ b/.changeset/slimy-otters-rush.md
@@ -1,5 +1,6 @@
 ---
-"miniflare": minor
+"miniflare": patch
+"wrangler": patch
 ---
 
 Add local binding support for Worker Pipelines
--- a/.changeset/slimy-otters-rush.md
+++ b/.changeset/slimy-otters-rush.md
@@ -1,0 +1,5 @@
+---
+"miniflare": minor
+---
+
+Add local binding support for Worker Pipelines
--- a/fixtures/vitest-pool-workers-examples/pipelines/test/pipeline-send-integration-self.test.ts
+++ b/fixtures/vitest-pool-workers-examples/pipelines/test/pipeline-send-integration-self.test.ts
@@ -1,0 +1,20 @@
+import { SELF } from "cloudflare:test";
+import { expect, it, vi } from "vitest";
+
+it("produces and consumers queue message", async () => {
+	// Send data to the Pipeline
+	let response = await SELF.fetch("https://example.com/ingest", {
+		method: "POST",
+		body: "value",
+	});
+	expect(response.status).toBe(202);
+	expect(await response.text()).toBe("Accepted");
+
+	// Wait until data is sent to the pipeline
+	const result = await vi.waitUntil(async () => {
+		const response = await SELF.fetch("https://example.com/ingest");
+		const text = await response.text();
+		if (response.ok) return text;
+	});
+	expect(result).toBe("Accepted");
+});
--- a/fixtures/vitest-pool-workers-examples/pipelines/test/pipeline-send-integration-self.test.ts
+++ b/fixtures/vitest-pool-workers-examples/pipelines/test/pipeline-send-integration-self.test.ts
@@ -1,7 +1,7 @@
 import { SELF } from "cloudflare:test";
-import { expect, it, vi } from "vitest";
+import { expect, it } from "vitest";
 
-it("produces and consumers queue message", async () => {
+it("sends message to pipeline", async () => {
 	// Send data to the Pipeline
 	let response = await SELF.fetch("https://example.com/ingest", {
 		method: "POST",
@@ -9,12 +9,4 @@ it("produces and consumers queue message", async () => {
 	});
 	expect(response.status).toBe(202);
 	expect(await response.text()).toBe("Accepted");
-
-	// Wait until data is sent to the pipeline
-	const result = await vi.waitUntil(async () => {
-		const response = await SELF.fetch("https://example.com/ingest");
-		const text = await response.text();
-		if (response.ok) return text;
-	});
-	expect(result).toBe("Accepted");
 });
--- a/fixtures/vitest-pool-workers-examples/pipelines/vitest.config.ts
+++ b/fixtures/vitest-pool-workers-examples/pipelines/vitest.config.ts
@@ -1,0 +1,17 @@
+import { defineWorkersProject } from "@cloudflare/vitest-pool-workers/config";
+
+export default defineWorkersProject({
+	test: {
+		poolOptions: {
+			workers: {
+				singleWorker: true,
+				miniflare: {
+					pipelines: ["PIPELINE"],
+				},
+				wrangler: {
+					configPath: "./wrangler.toml",
+				},
+			},
+		},
+	},
+});
--- a/fixtures/vitest-pool-workers-examples/pipelines/vitest.config.ts
+++ b/fixtures/vitest-pool-workers-examples/pipelines/vitest.config.ts
@@ -5,9 +5,6 @@ export default defineWorkersProject({
 		poolOptions: {
 			workers: {
 				singleWorker: true,
-				miniflare: {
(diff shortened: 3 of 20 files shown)


packages/miniflare/src/plugins/pipelines/index.ts
```diff
@@ -0,0 +1,104 @@
...
+	className: PIPELINE_OBJECT_CLASS_NAME,
+};
+
+export const PIPELINE_PLUGIN: Plugin<typeof PipelineOptionsSchema> = {
+	options: PipelineOptionsSchema,
+	getBindings(options) {
+		const pipelines = bindingEntries(options.pipelines);
+		return pipelines.map<Service>(([name]) => ({
+			name,
+			service: { name: SERVICE_PIPELINE_PREFIX },
```
emily-shen, line 36: think this needs the id?
```suggestion
			service: { name: `SERVICE_PIPELINE_PREFIX:${id}` },
```

emily-shen (commented):
https://github.com/cloudflare/workers-sdk/blob/cee82faca692faec0d9cc271e5390c945f1b1c40/packages/wrangler/src/dev/miniflare.ts#L670 here is the bit you need to wire up wrangler and miniflare

packages/miniflare/src/plugins/pipelines/index.ts
cmackenzie1: I don't think that is needed here since these are not backed by DOs, but JSPRC bindings. @penalosa do you recall if that is correct?

https://github.com/cloudflare/workers-sdk/blob/9560ae5fed94d1858e380804c903c5020318501d/packages/miniflare/src/workers/pipelines/pipeline.worker.ts#L3

emily-shen: so it looks like your set up is 
user worker ---binding--> pipeline entry worker --> pipeline service worker

where the pipeline entryworker is the `services` declared in line 49 with `objectEntryWorker()`
so those are per pipeline binding and have an id in their name, so when you use `getBinding` miniflare is hooking up user worker -> pipeline entry worker

along the lines of d1 and kv in this diagram:

image

packages/miniflare/src/plugins/pipelines/index.ts
```diff
@@ -0,0 +1,104 @@
...
+							esModule: SCRIPT_PIPELINE_OBJECT(),
+						},
+					],
+					durableObjectNamespaces: [
+						{
+							className: PIPELINE_OBJECT_CLASS_NAME,
+							uniqueKey,
+						},
+					],
+					durableObjectStorage: { inMemory: kVoid },
```
emily-shen, line 71: i don't think these bits are needed either since it looks like you're not currently using DOs

packages/miniflare/src/plugins/pipelines/index.ts
cmackenzie1: I am seeing this error in the tests now (was passing before without ID param). Did I miss a part in [5110fca](https://github.com/cloudflare/workers-sdk/pull/7950/commits/5110fcaa4c3752d8267745dcc27a80d15e3f0623)?

```
workerd/server/server.c++:2009: warning: A DurableObjectNamespace in the config referenced the class "Pipeline", but no such Durable Object class is exported from the worker. Please make sure the class name matches, it is exported, and the class extends 'DurableObject'. Attempts to call to this Durable Object class will fail at runtime, but historically this was not a startup-time error. Future versions of workerd may make this a startup-time error.
Worker "core:user:"'s binding "PIPELINE" refers to a service "SERVICE_PIPELINE_PREFIX:PIPELINE", but no such service is defined.
```

packages/miniflare/src/plugins/pipelines/index.ts
```diff
@@ -0,0 +1,104 @@
...
+	className: PIPELINE_OBJECT_CLASS_NAME,
+};
+
+export const PIPELINE_PLUGIN: Plugin<typeof PipelineOptionsSchema> = {
+	options: PipelineOptionsSchema,
+	getBindings(options) {
+		const pipelines = bindingEntries(options.pipelines);
+		return pipelines.map<Service>(([name, id]) => ({
+			name,
+			service: { name: `SERVICE_PIPELINE_PREFIX:${id}` },
```
emily-shen, line 36:
```suggestion
			service: { name: `${SERVICE_PIPELINE_PREFIX}:${id}` },
```

just a typo hopefully!

cmackenzie1: 🤦‍♂️

.changeset/slimy-otters-rush.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": minor
```
emily-shen, line 2: should be a patch as its experimental, and should also apply to wrangler

fixtures/vitest-pool-workers-examples/pipelines/test/pipeline-send-integration-self.test.ts
emily-shen, line 1: can you check for the console output? (for the unit test version too)

fixtures/vitest-pool-workers-examples/pipelines/vitest.config.ts
```diff
@@ -0,0 +1,17 @@
+import { defineWorkersProject } from "@cloudflare/vitest-pool-workers/config";
+
+export default defineWorkersProject({
+	test: {
+		poolOptions: {
+			workers: {
+				singleWorker: true,
+				miniflare: {
```
emily-shen, line 8: hmm you shouldn't need to specify the pipelines binding separately, it should get pulled from wrangler.toml - does it not work if you don't do that?

packages/miniflare/src/plugins/pipelines/index.ts
```diff
@@ -0,0 +1,68 @@
...
+	async getServices({ options }) {
+		const pipelines = bindingEntries(options.pipelines);
+
+		const services = [];
+		for (const pipeline of pipelines) {
+			services.push({
+				name: `${SERVICE_PIPELINE_PREFIX}:${pipeline[1]}`,
+				worker: {
+					compatibilityDate: "2024-12-30",
+					compatibilityFlags: ["nodejs_compat"],
```
emily-shen, line 37: this shouldn't need nodejs_compat?

packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -609,7 +609,7 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 		await fetch(`${url}/connect`);
 	});
 
-	it.skipIf(!isLocal).fails("exposes Pipelines bindings", async () => {
+	it.skipIf(!isLocal)("exposes Pipelines bindings", async () => {
```
emily-shen, line 612: are you able to check for the console output as well - otherwise its not actually testing the miniflare plugin

emily-shen (changes requested):
Also if you'd be able to document the new miniflare options here:
https://github.com/cloudflare/workers-sdk/blob/main/packages/miniflare/README.md

emily-shen (approved):
🚀

emily-shen:
oh looks like those e2e failures are legitimate - should be fixed if you just change to a `toContain` etc. check or use `worker.readUntil`

PR #7964, opened 2025-01-29
fix: make `wrangler deploy` not call Workflows API if it binds to a different script

LuisDuarte1:
Basically, if a worker bounds to a Workflow defined in another script, it would do a PUT call to the Workflow API, overriding the config and because it would change the `script_name`, it would break new instances (because this one will probably not have the class exported).

This commit makes it so that Wrangler does not call the API if the `script_name` is different.

Fixes WOR-327.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: just a fix


edmundhung (commented):
Looks good to me. We will get this merged once CI is green 👍🏼

edmundhung (approved):

PR #7967, opened 2025-01-29
Revert "fix(wrangler): handling of `process.env.NODE_ENV` in bundling mode"

edmundhung:
Reverts cloudflare/workers-sdk#7932.

### DIFF
--- a/.changeset/lucky-jeans-relax.md
+++ b/.changeset/lucky-jeans-relax.md
@@ -1,5 +1,0 @@
----
-"wrangler": patch
----
-
-fix(wrangler): handling of `process.env.NODE_ENV` in bundling mode
--- a/packages/wrangler/src/deployment-bundle/bundle.ts
+++ b/packages/wrangler/src/deployment-bundle/bundle.ts
@@ -393,11 +393,6 @@ export async function bundleWorker(
 
 	const unenvResolvePaths = getUnenvResolvePathsFromEnv()?.split(",");
 
-	// Use `process.env["NODE_ENV" + ""]` to get the runtime value when wranler is executed.
-	// Esbuild would replace `process.env.NODE_ENV` with the value it has when wrangler gets bundled.
-	// See: https://github.com/cloudflare/workers-sdk/issues/1477)
-	const runtimeProcessEnv = process.env["NODE_ENV" + ""];
-
 	const buildOptions = {
 		// Don't use entryFile here as the file may have been changed when applying the middleware
 		entryPoints: [entry.file],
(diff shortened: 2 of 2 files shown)


lrapoport-cf:
> Can you provide some explanation why this is being reverted?

my understanding is that although https://github.com/cloudflare/workers-sdk/pull/7932 indicates in the PR description that it has tests, they're actually missing from that PR. so this revert is to go back until tests can be added. @edmundhung @vicb is this correct?

lrapoport-cf (approved):

vicb:
I think that is the case, my answer to @edmundhung yesterday

> You're absolutely right, tests are not included, I definitely should have mentioned that. 
There is one more point to be fixed in the original issue (when process is imported rather than using the global) that definitely needs testing.
This PR only includes the fix that are obvious but I can understand if it cannot be merged without tests.

vicb:
To be clear:

- I think that the current code is broken
- I think that the reverted PR fixes some of the issues
- The current is not tested (properly or at all, that I don't know)
- I think the reverted PR is an improvement
- BUT I fully understand why no test is a reason to revert the PR... i.e. we wouldn't have broken in the first place if we caught that it was not tested.

I think it would be good to add test soon and fixed the remaining point in the original issue

edmundhung:
I have prepared this PR just in case we are feeling uncomfortable with the changes for today's release. But happy to close this if we have no concern :) Personally I think the change is minimal and it is quite clear what it fixed.

vicb:
I'll defer to others for the decision.

Whatever the decision, I think #7886 should be prioritized.

penalosa:
@vicb @edmundhung I think the change is minimal and shouldn't cause issues (if it would, we'd see other tests failing). However, perhaps @vicb you could add a simple test that checks that Wrangler correctly statically defines process.env.NODE_ENV? The navigator define is already tested.

In terms of fixing the imported process, personally I think we should punt on that until the runtime solution lands, since the situation right now is stable.

vicb:
> ever, perhaps @vicb you could add a simple test that checks that Wrangler correctly statically defines process.env.NODE_ENV? The navigator define is already tested.

vicb:
> @vicb @edmundhung I think the change is minimal and shouldn't cause issues (if it would, we'd see other tests failing). However, perhaps @vicb you could add a simple test that checks that Wrangler correctly statically defines process.env.NODE_ENV? The navigator define is already tested.

Could you point me to a starting point (i.e. existing test)?

> In terms of fixing the imported process, personally I think we should punt on that until the runtime solution lands, since the situation right now is stable.

Are you thinking of https://github.com/cloudflare/workerd/pull/3311?
If that's the case I think that it should be considered orthogonal.
Quite a few libs out there have debug code guarded by `if (process.env.NODE === "development") { ... }` (or similar). While having the value at runtime would make sure the code is not executed, it would still be in the bundle (including potential dependencies).
Having `process.env.NODE` replaced at build time would ensure the code is tree shaken and would allow reducing the bundle size.
_(Also 3311 is paused and will be guarded by a compatibility flag when released)_

penalosa:
@vicb the navigator tests are in https://github.com/cloudflare/workers-sdk/blob/ce8bfbdada90d1eac99f3a35dfdf1707e88eb73f/packages/wrangler/src/__tests__/navigator-user-agent.test.ts, which should be a good starting point.

edmundhung:
We will revert this as agreed and will land this change again once we have the test ready 👍🏼

Issue #7905, opened 2025-01-25
🐛 BUG: New Hono project from C3 creates wrangler.toml and wrangler.json


irvinebroque:
### Which Cloudflare product(s) does this pertain to?

C3 (npm create cloudflare), Wrangler

### What versions are you using?

latest

### What operating system and version are you using?

latest

### Please provide a link to a minimal reproduction

https://x.com/thinanchor/status/1881410770208272729

### Describe the Bug

1. `npm create cloudflare@latest`
2. choose hono
3. observe that the project has both a wrangler.json and a wrangler.toml

```
➜  src cd hono
➜  hono ls
README.md                 package.json              worker-configuration.d.ts
node_modules              src                       wrangler.json
package-lock.json         tsconfig.json             wrangler.toml
```

Beyond the obvious fix — I think when we see that there is both a wrangler.json and a wrangler.toml present — we should consdier:

1. warn people loudly
2. say when you run wrangler dev which configuration file is being used

irvinebroque:
refs https://github.com/cloudflare/workers-sdk/pull/7676 or https://github.com/cloudflare/workers-sdk/pull/7774 ?

irvinebroque:
think maybe because hono is a framework but also a worker, while default for other frameworks are to target pages?

yusukebe:
Hey @irvinebroque 

The starter template in Hono now uses `wrangler.toml`, but anyway, using `wrangler.json` is good if the C3 starts using `wrangler.json.` We'll work on making the template using `wrangler.json` instead of `wrangler.toml`.

PR #7972, opened 2025-01-30
Bumps create-hono now that upstream has updated to use wrangler.json

andyjessop:
Fixes https://github.com/cloudflare/workers-sdk/issues/7905

`create-hono` has now been updated to generate `wrangler.json` instead of `wrangler.toml`. Because `create-hono` points to `main` on the starter instead of a specific version, we need to bump it to ensure the two are in-sync.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/fair-waves-punch.md
+++ b/.changeset/fair-waves-punch.md
@@ -2,4 +2,4 @@
 "create-cloudflare": patch
 ---
 
-chore: bumps create-hono version to ensure correct Wrangler config file
+chore: bumps create-hono version to ensure correct Wrangler config file. Fixes the issue where C3 was generating both a `wrangler.toml` and a `wrangler.json` file. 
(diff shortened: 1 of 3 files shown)


.changeset/fair-waves-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+bump create-hono version
```
petebacondarwin: Might be worth adding some context here - i.e. this fixes a problem related to C3 now using wrangler.json rather than wrangler.toml

And mention that this breaks the old version since the hono starter template has been updated.

petebacondarwin (approved):

edmundhung (approved):

PR #7974, opened 2025-01-30
fix(wrangler): handling of process.env.NODE_ENV in bundling mode

vicb:
Re-introduce #7932 with a test (reverted in #7967 because it was missing a test)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included - really this time :)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: the initial PR was E2E tested, let's save resources :)
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


penalosa (commented):
Could you add a test for NODE_ENV?

vicb:
> Could you add a test for NODE_ENV?

Just did that by moving the test to a specific ESBuild/define file.
We can re-use that for the last bug in the original issue.

Note that a default value is set somewhere - it probably explains why the bug went unnoticed...

vicb:
latest forced push is a rebase

packages/wrangler/src/__tests__/esbuild-define.test.ts
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+async function seedFs(files: Record<string, string>): Promise<void> {
+	for (const [location, contents] of Object.entries(files)) {
+		await mkdir(path.dirname(location), { recursive: true });
+		await writeFile(location, contents);
+	}
+}
+
+// Does bundleWorker respect the value of `defineNavigatorUserAgent`?
```
penalosa, line 24: I don't think this comment is correct?

packages/wrangler/src/__tests__/esbuild-define.test.ts
```diff
@@ -0,0 +1,118 @@
...
+		const code = await getBundledWorker("production");
+
+		expect(code).toContain(`userAgent = "Cloudflare-Workers";`);
+		expect(code).toContain(`g = global;`);
+		expect(code).toContain(`nodeEnv = "production";`);
+	});
+});
+
+async function getBundledWorker(nodeEnv: string | undefined) {
+	const startNodeEnv = process.env.NODE_ENV;
```
penalosa, line 63: Is the `vi.stubEnv()` API usable here?

packages/wrangler/src/__tests__/esbuild-define.test.ts
```diff
@@ -0,0 +1,118 @@
...
+		expect(code).toContain(`g = global;`);
+		expect(code).toContain(`nodeEnv = "development";`);
+	});
+
+	it("Replaces defined values when process.env.NODE_ENV is undefined", async () => {
+		const code = await getBundledWorker(undefined);
+
+		expect(code).toContain(`userAgent = "Cloudflare-Workers";`);
+		expect(code).toContain(`g = global;`);
+		expect(code).toContain(`nodeEnv = "development";`);
```
penalosa, line 42: Where does this `development` value come from?

vicb: Not sure... but this is not changed by this PR.
Is it ok if I create an issue to investigate and link it here to tackle in a follow up PR?

vicb: Oh actually this comes from the `...define`:

```ts
		define: {
			...(defineNavigatorUserAgent
				? { "navigator.userAgent": `"Cloudflare-Workers"` }
				: {}),
			...(runtimeProcessEnv
				? { "process.env.NODE_ENV": `"${runtimeProcessEnv}"` }
				: {}),
			...(nodejsCompatMode === "legacy" ? { global: "globalThis" } : {}),
			...define,
		},
```

I need to investigate this more

penalosa:
Closing for now as this is stale. We may re-address in future.

PR #7977, opened 2025-01-31
Adding wrangler commands for R2 bucket lock rule configuration

jkoe-cf:
What this PR solves / how to test
Allows user to configure lock configuration on an R2 bucket

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: changes covered by unit tests
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/19846

### DIFF
--- a/packages/wrangler/src/__tests__/r2.test.ts
+++ b/packages/wrangler/src/__tests__/r2.test.ts
@@ -6,17 +6,59 @@ import { actionsForEventCategories } from "../r2/helpers";
 import { endEventLoop } from "./helpers/end-event-loop";
 import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
 import { mockConsoleMethods } from "./helpers/mock-console";
-import { mockConfirm } from "./helpers/mock-dialogs";
+import { mockConfirm, mockPrompt } from "./helpers/mock-dialogs";
 import { useMockIsTTY } from "./helpers/mock-istty";
 import { createFetchResult, msw, mswR2handlers } from "./helpers/msw";
 import { runInTempDir } from "./helpers/run-in-tmp";
 import { runWrangler } from "./helpers/run-wrangler";
 import { writeWranglerConfig } from "./helpers/write-wrangler-config";
 import type {
+	BucketLockRule,
 	PutNotificationRequestBody,
 	R2EventableOperation,
 	R2EventType,
 } from "../r2/helpers";
+
+function mockBucketLockPutNew(bucketName: string, rules: BucketLockRule[]) {
+	mockBucketLockPutWithExistingRules(bucketName, [], rules);
+}
+
+function mockBucketLockPutWithExistingRules(
+	bucketName: string,
+	existingRules: BucketLockRule[],
+	newRules: BucketLockRule[]
+) {
+	msw.use(
+		http.get(
+			"*/accounts/:accountId/r2/buckets/:bucketName/lock",
+			async ({ params }) => {
+				const { accountId, bucketName: bucketParam } = params;
+				expect(accountId).toEqual("some-account-id");
+				expect(bucketParam).toEqual(bucketName);
+				return HttpResponse.json(
+					createFetchResult({
+						rules: existingRules,
+					})
+				);
+			},
+			{ once: true }
+		),
+		http.put(
+			"*/accounts/:accountId/r2/buckets/:bucketName/lock",
+			async ({ request, params }) => {
+				const { accountId, bucketName: bucketParam } = params;
+				expect(accountId).toEqual("some-account-id");
+				expect(bucketName).toEqual(bucketParam);
+				const requestBody = await request.json();
+				expect(requestBody).toEqual({
+					rules: newRules,
+				});
+				return HttpResponse.json(createFetchResult({}));
+			},
+			{ once: true }
+		)
+	);
+}
 
 describe("r2", () => {
 	const std = mockConsoleMethods();
@@ -98,6 +140,7 @@ describe("r2", () => {
 				  wrangler r2 bucket dev-url          Manage public access via the r2.dev URL for an R2 bucket
 				  wrangler r2 bucket lifecycle        Manage lifecycle rules for an R2 bucket
 				  wrangler r2 bucket cors             Manage CORS configuration for an R2 bucket
+				  wrangler r2 bucket lock             Manage lock rules for an R2 bucket
 
 				GLOBAL FLAGS
 				  -c, --config   Path to Wrangler configuration file  [string]
@@ -137,6 +180,7 @@ describe("r2", () => {
 				  wrangler r2 bucket dev-url          Manage public access via the r2.dev URL for an R2 bucket
 				  wrangler r2 bucket lifecycle        Manage lifecycle rules for an R2 bucket
 				  wrangler r2 bucket cors             Manage CORS configuration for an R2 bucket
+				  wrangler r2 bucket lock             Manage lock rules for an R2 bucket
 
 				GLOBAL FLAGS
 				  -c, --config   Path to Wrangler configuration file  [string]
@@ -1958,7 +2002,7 @@ describe("r2", () => {
 				});
 			});
 			describe("add", () => {
-				it("it should add a lifecycle rule using command-line arguments", async () => {
+				it("it should add an age lifecycle rule using command-line arguments", async () => {
 					const bucketName = "my-bucket";
 					const ruleId = "my-rule";
 					const prefix = "images/";
@@ -2009,6 +2053,64 @@ describe("r2", () => {
 					);
 					await runWrangler(
 						`r2 bucket lifecycle add ${bucketName} --id ${ruleId} --prefix ${prefix} --expire-days ${conditionValue}`
+					);
+					expect(std.out).toMatchInlineSnapshot(`
+						"Adding lifecycle rule 'my-rule' to bucket 'my-bucket'...
+						✨ Added lifecycle rule 'my-rule' to bucket 'my-bucket'."
+					  `);
+				});
+
+				it("it should add a date lifecycle rule using command-line arguments", async () => {
+					const bucketName = "my-bucket";
+					const ruleId = "my-rule";
+					const prefix = "images/";
+					const conditionType = "Date";
+					const conditionValue = "2025-01-30";
+
+					msw.use(
+						http.get(
+							"*/accounts/:accountId/r2/buckets/:bucketName/lifecycle",
+							async ({ params }) => {
+								const { accountId, bucketName: bucketParam } = params;
+								expect(accountId).toEqual("some-account-id");
+								expect(bucketParam).toEqual(bucketName);
+								return HttpResponse.json(
+									createFetchResult({
+										rules: [],
+									})
+								);
+							},
+							{ once: true }
+						),
+						http.put(
+							"*/accounts/:accountId/r2/buckets/:bucketName/lifecycle",
+							async ({ request, params }) => {
+								const { accountId, bucketName: bucketParam } = params;
+								expect(accountId).toEqual("some-account-id");
+								expect(bucketName).toEqual(bucketParam);
+								const requestBody = await request.json();
+								expect(requestBody).toEqual({
+									rules: [
+										{
+											id: ruleId,
+											enabled: true,
+											conditions: { prefix: prefix },
+											deleteObjectsTransition: {
+												condition: {
+													type: conditionType,
+													date: "2025-01-30T00:00:00.000Z",
+												},
+											},
+										},
+									],
+								});
+								return HttpResponse.json(createFetchResult({}));
+							},
+							{ once: true }
+						)
+					);
+					await runWrangler(
+						`r2 bucket lifecycle add ${bucketName} --id ${ruleId} --prefix ${prefix} --expire-date ${conditionValue}`
 					);
 					expect(std.out).toMatchInlineSnapshot(`
 						"Adding lifecycle rule 'my-rule' to bucket 'my-bucket'...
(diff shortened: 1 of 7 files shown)


packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,330 @@
...
+			conditionValue = date.toISOString();
+			newRule.condition = {
+				type: conditionType,
+				date: conditionValue,
+			};
+		} else {
+			conditionType = "Indefinite";
+		}
+
+		if (!prefix && !force) {
```
emily-shen, line 184: this should combined with the prefix prompt earlier

so
```
 if(prefix === undefined && !force){
   confirm("No prefix has been specified, meaning this will apply to all prefixes. The lock rule will apply to all objects in your bucket. Are you sure you want to do that?
   ")
}
```
rough wording, but you get the point 😅

packages/wrangler/src/index.ts
```diff
@@ -696,6 +703,26 @@ export function createCLIParser(argv: string[]) {
...
+		{
+			command: "wrangler r2 bucket lock add",
+			definition: r2BucketLockAddCommand,
+		},
+		{
+			command: "wrangler r2 bucket lock remove",
+			definition: r2BucketLockRemoveCommand,
+		},
+		{
+			command: "wrangler r2 bucket lock set",
```
emily-shen, line 723: `wrangler r2 bucket lock add bulk` would be more in fitting with wrangler conventions

emily-shen (changes requested):

emily-shen:
would also be good to get some tests for invalid input 🧡

packages/wrangler/src/index.ts
jkoe-cf: We already have other older commands such as `wrangler r2 bucket lifecycle set` to upload from json, I'm not sure if this was an older convention but should those all be updated to support as `add bulk` or is it fine to leave existing commands as-is and make this one different?

emily-shen: ah fair enough if its the same pattern as established r2 commands its better to continue with `set` 👍

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,334 @@
...
+			});
+		}
+
+		const newRule: BucketLockRule = {
+			id: id,
+			enabled: true,
+			condition: { type: "Indefinite" },
+		};
+		if (prefix === undefined) {
+			if (force) {
```
emily-shen, line 135: I think if someone has specified force, they would expect there to be no interaction necessary? in this case i would expect it to skip the prompt and the confirm

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,334 @@
...
+				prefix = await prompt(
+					"Enter a prefix for the bucket lock rule (leave empty for all prefixes)",
```
emily-shen, line 137:
```suggestion
					"Enter a prefix for the bucket lock rule (set to "" for all prefixes)",
```

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,334 @@
...
+					`Are you sure you want to add lock rule '${id}' to bucket '${bucket}' without a prefix? ` +
+						`The lock rule will apply to all objects in your bucket.`
+				);
+				if (!confirmedAdd) {
+					logger.log("Add cancelled.");
+					return;
+				}
+			}
+		}
+		if (lockDays === undefined && lockDate === undefined) {
```
emily-shen, line 151: what about force here? A suggestion: what if you made the default assumptions, and then presented the user with the lock config, and then asking for confirmation once?

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,334 @@
...
+				logger.log("Add cancelled.");
+				return;
+			} else {
+				newRule.condition = {
+					type: "Indefinite",
+				};
+			}
+		} else {
+			if (lockDays !== undefined) {
+				if (isNonNegativeNumber(String(lockDays))) {
```
emily-shen, line 167: since you know lockDays is a number you don't need to cast it to a string and use the helper

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,334 @@
...
+					const conditionDaysValue = Number(lockDays) * 86400; // Convert days to seconds
```
emily-shen, line 168: don't need to cast it to a string as yargs should error if its not a number. the only check you'll need to make is if its positive

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,334 @@
...
+				} else {
+					throw new UserError(
+						`Must be a valid date in the YYYY-MM-DD format: ${String(lockDate)}`,
+						{ telemetryMessage: true }
+					);
+				}
+			}
+		}
+
+		if (prefix) {
```
emily-shen, line 196: nit: this could be moved up so its nearer the prefix validation

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,334 @@
...
+		} catch (e) {
+			if (e instanceof Error) {
+				throw new UserError(
+					`Failed to read or parse the lock configuration config file: '${e.message}'`,
+					{ telemetryMessage: true }
+				);
+			} else {
+				throw e;
+			}
+		}
```
emily-shen, line 309:
```suggestion
		} catch (e) {
				throw new ParseError(
					`Failed to read or parse the lock configuration config file: '${e.message}'`,
					{ telemetryMessage: "Failed to read or parse the lock configuration config file" }
				);
			} 
		}
```

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2277,6 +2337,575 @@ describe("r2", () => {
...
+						Lock rule 'my-rule' removed from bucket 'my-bucket'."
+					  `);
+				});
+				it("should handle removing non-existant rule ID as expected", async () => {
+					const bucketName = "my-bucket";
+					const ruleId = "my-rule";
+					const lockRules = {
+						rules: [],
+					};
+					msw.use(
```
emily-shen, line 2828: nit: you could extract these api mocks into helpers to reuse

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,333 @@
...
+					{ telemetryMessage: true }
+				);
+			}
+		} else {
+			newRule.condition = {
+				type: "Indefinite",
+			};
+		}
+
+		if (lockDays === undefined && lockDate === undefined && !force) {
```
emily-shen, line 187: similarly to how you added prefix and id, you need to do this before the section setting lockdays/lockdate above. since right now even if people confirm they do want it to be indefinite, that's not getting set?

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,333 @@
...
+				}
+			}
+		}
+
+		if (prefix) {
+			newRule.prefix = prefix;
+		}
+
+		if (lockDays !== undefined) {
+			if (isNonNegativeNumber(lockDays)) {
```
emily-shen, line 156:
```suggestion
			if (lockDays >0) {
```

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,333 @@
...
+				const conditionDaysValue = Number(lockDays) * 86400; // Convert days to seconds
```
emily-shen, line 157: I added a suggestion above to make the yargs type a number. Then you won't have to cast it here

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,333 @@
...
+				newRule.condition = {
+					type: "Age",
+					maxAgeSeconds: conditionDaysValue,
+				};
+			} else {
+				throw new UserError(`Must be a positive number: ${String(lockDays)}`, {
```
emily-shen, line 163: you don't need to cast it to a string, a template literals do that automatically :)

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,333 @@
...
+					telemetryMessage: true,
```
emily-shen, line 164:
```suggestion
					telemetryMessage:{lock days not a positive number},
```
(easier for us to aggregate if we don't include the actual number)

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2277,6 +2379,362 @@ describe("r2", () => {
...
+					id:         rule-indefinite
+					enabled:    Yes
+					prefix:     images/indefinite
+					condition:  indefinitely"
+				  `);
+				});
+			});
+			describe("add", () => {
+				it("it should add a lock rule without prefix using command-line arguments", async () => {
+					setIsTTY(false);
```
emily-shen, line 2455:
```suggestion
					setIsTTY(true);
```
otherwise the prompts are just using the defaults

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2277,6 +2379,362 @@ describe("r2", () => {
...
+
+					id:         rule-indefinite
+					enabled:    Yes
+					prefix:     images/indefinite
+					condition:  indefinitely"
+				  `);
+				});
+			});
+			describe("add", () => {
+				it("it should add a lock rule without prefix using command-line arguments", async () => {
```
emily-shen, line 2454: can you also add a test for `wrangler lock add bucket-name` ie no cli args

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2277,6 +2379,362 @@ describe("r2", () => {
...
+						✨ Added lock rule 'rule-no-prefix' to bucket 'my-bucket'."
+					  `);
+				});
+				it("it should add an age lock rule using command-line arguments", async () => {
+					setIsTTY(true);
+					const bucketName = "my-bucket";
+					const ruleIdAge = "rule-age";
+					const prefixAge = "prefix-age";
+					const conditionTypeAge = "Age";
+					const conditionValueAge = 1;
```
emily-shen, line 2499: nit: Unless you were going to loop through various test cases where these values would be different, I think it would be more readable to just use the actual values in place.

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2277,6 +2379,362 @@ describe("r2", () => {
...
+					// date
+					await runWrangler(
+						`r2 bucket lock add ${bucketName} --id ${ruleIdDate} --prefix ${prefixDate} --lock-date ${conditionValueDate}`
+					);
+					expect(std.out).toMatchInlineSnapshot(`
+						"Adding lock rule 'rule-date' to bucket 'my-bucket'...
+						✨ Added lock rule 'rule-date' to bucket 'my-bucket'."
+					  `);
+				});
+				it("it should add a date lock rule using command-line arguments", async () => {
```
emily-shen, line 2577:
```suggestion
				it("should fail with an invalid date"", async () => {
```

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2277,6 +2379,362 @@ describe("r2", () => {
...
+					setIsTTY(true);
+					const bucketName = "my-bucket";
+					const ruleIdDate = "rule-date";
+					const prefixDate = "prefix-date";
+					const conditionTypeDate = "Date";
+					const conditionValueDate = "January 30, 2025";
+
+					mockBucketLockPutNew(bucketName, [
```
emily-shen, line 2585: mock is not necessary as it should fail before you get to api call

packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2277,6 +2379,362 @@ describe("r2", () => {
...
+						`r2 bucket lock add ${bucketName} --id ${ruleIdIndefinite} --prefix ${prefixIndefinite}`
+					);
+					expect(std.out).toMatchInlineSnapshot(`
+						"? Are you sure you want to add lock rule 'rule-indefinite' to bucket 'my-bucket' without an expiration? The lock rule will apply to all matching objects indefinitely.
+						🤖 Using fallback value in non-interactive context: yes
+						Adding lock rule 'rule-indefinite' to bucket 'my-bucket'...
+						✨ Added lock rule 'rule-indefinite' to bucket 'my-bucket'."
+					  `);
+				});
+			});
```
emily-shen, line 2640: Are you also able to add a test for when lockdays and lockdate are not defined via cli args

emily-shen (changes requested):
Also, its best practice to refrain from squashing your commits/force pushing after a review, so the reviewer can see what has changed since 🧡 
If you want to tidy up your commit history you can do that after approval, but also we squash merge to main in this repo anyway so it doesn't really matter a whole lot here 🧡

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,333 @@
...
+		description: "Manage lock rules for an R2 bucket",
+		status: "stable",
+		owner: "Product: R2",
+	},
+});
+
+export const r2BucketLockListCommand = createCommand({
+	metadata: {
+		description: "List lock rules for an R2 bucket",
+		status: "stable",
```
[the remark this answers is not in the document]
jkoe-cf: We're fine with marking these as stable

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,344 @@
...
+				{ defaultValue: true }
+			);
+			if (confirmIndefinite !== true) {
+				logger.log("Add cancelled.");
+				return;
+			}
+		}
+
+		if (lockDays !== undefined) {
+			if (!isNaN(lockDays)) {
```
emily-shen, line 167: nit: this is not necessary, because if the input isn't a number yargs will have failed already when its passing the cli args

packages/wrangler/src/r2/lock.ts
```diff
@@ -0,0 +1,344 @@
...
+				if (lockDays > 0) {
+					const conditionDaysValue = Number(lockDays) * 86400; // Convert days to seconds
```
emily-shen, line 169: nit - yargs will have set the type of lockDays as a number, so there's no need to cast this to a number again. typescript would have let you know if the type was wrong :)

emily-shen (approved):

packages/wrangler/src/r2/lock.ts
jkoe-cf: Not sure if it's something with the test setup vs. how wrangler works in practice but it looks like it still reaches there if the type is wrong so I might leave the NaN logic as-is if that's ok https://github.com/cloudflare/workers-sdk/blob/99db3e4a76b3e2350185271eb56c7566a6317014/packages/wrangler/src/__tests__/r2.test.ts#L2565

jonesphillip (approved):

PR #7980, opened 2025-01-31
feat(fixtures): Add new fixtures Workers + Assets with service bindings

CarmenPopoviciu:

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+	 * as a Request object.
+	 *
+	 * see https://developers.cloudflare.com/workers/runtime-apis/handlers/fetch/
+	 */
+	async fetch(request, env, ctx) {
+		return new Response("Hello from worker-🐝 fetch()");
+	},
+
+	/*
+	 * Named functions ??
```
CarmenPopoviciu, line 18: AFAICT named fns on default Fetchers are not documented anywhere, but this does seem to work. I am not sure if `named functions` is the correct naming here. Happy to be directed to the correct docs if they exist <3

fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
```diff
@@ -0,0 +1,59 @@
+/*
+ * default entrypoint
+ */
+export default {
```
CarmenPopoviciu, line 4: I have used https://developers.cloudflare.com/workers/runtime-apis/handlers/ as a point of reference in terms of what handlers we want to test. 
- I am unsure whether `tail` is something that applies in this context
- I believe `alarm` does not, since that is DO specific? and these fixtures don't cover DOs yet

fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
```diff
@@ -0,0 +1,59 @@
...
+	/*
+	 * Queues
+	 *
+	 * Allows a consumer Worker to consume messages from a Queue. To define
+	 * a consumer Worker, add a queue() function to the default export of the
+	 * Worker.
+	 *
+	 * see https://developers.cloudflare.com/queues/configuration/javascript-apis/#consumer
+	 */
+	// TODO
```
CarmenPopoviciu, line 46: - [ ]  @CarmenPopoviciu add queue handler

fixtures/workers-with-assets-and-rpc-named-entrypoints/tests/index.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+			it("should respond to cron trigger", async ({ expect }) => {
+				let response = await fetch(
+					`http://${ipWorkerB}:${portWorkerB}/__scheduled`
+				);
+				let text = await response.text();
+				expect(response.status).toBe(200);
+				expect(text).toContain("Ran scheduled event");
+
+				// for some reason scheduled() never logs to stdout, but no idea why
```
CarmenPopoviciu, line 82: anyone has any ideas why this wouldn't work? I have tested manually with `wrangler dev --test-scheduled` and can confirm that no logging from the `scheduled()` fn happens 😕

fixtures/workers-with-assets-and-http-based-service-bindings/tests/index.test.ts
```diff
@@ -0,0 +1,163 @@
...
+				expect(getOutputWorkerB()).toContain(
+					"Hello from worker-🐝 scheduled()"
+				);
+			});
+		});
+
+		/*
+		 * Assert that WorkerA works as intended
+		 */
+		describe("WorkerA", () => {
```
CarmenPopoviciu, line 93: to clarify, `workerA` makes all the requests we are testing for and returns **one** response that includes all those responses. But for the sake of keeping things readable and having these tests serve as docs as well, I have split the tests into separate, individual bits per expected response.

CarmenPopoviciu: if folks feel like this is confusing, pls call that out

fixtures/workers-with-assets-and-http-based-service-bindings/README.md
```diff
@@ -0,0 +1,23 @@
+# workers-with-assets-and-http-based-service-bindings
```
CarmenPopoviciu, line 1: y'all let me know if this verbiage is confusing. I sort of followed the naming from https://blog.cloudflare.com/javascript-native-rpc/#example-authentication-service

fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
emily-shen: 'named functions' are just RPC methods - I think they work because the workerEntrypoint class and default object export (idk what to call this, but its exporting an object 😅 ) are treated the same when they get to workerd. However the types get unhappy if you add your own RPC methods to the default object export like you do here (see my comment above https://github.com/cloudflare/workers-sdk/pull/7980/files#r1977268625). I think we should drop these as we should not be recommending people add RPC methods to anything except for WorkerEntrypoints (or DO entrypoints etc.).

fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
emily-shen: We should probably test a queue() consumer handler - you can make workerB the queue consumer _and_ producer and test that the queue handler successfully receives the queued item. 

Tail workers do make sense imo, but I'm not sure if the tail worker and producer worker can/should be the same worker? I think someone from D&C will know, @GregBrimble has definitely tried tail worker + worker with assets. 

I don't think either of these are urgent though, you could definitely do them in a followup :)

fixtures/workers-with-assets-and-rpc-default-entrypoints/README.md
```diff
@@ -0,0 +1,23 @@
+# workers-with-assets-and-rpc-default-entrypoints
```
emily-shen, line 1: Up to you, but i think the named and default entrypoints could be one fixture. It would be good to test a worker B that has a default and named entrypoints, and make sure we hit the right entrypoints.

fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
CarmenPopoviciu: sounds abs reasonable to me. I'll remove

fixtures/workers-with-assets-and-http-based-service-bindings/README.md
CarmenPopoviciu: this fixture needs to test `fetch()`es in particular, because that's currently broken in dev. I'll remove the named fns, but `fetch()`es and `scheduled` and whatever else is in there, needs to stay, because we are intentionally testing for those things

fixtures/workers-with-assets-and-rpc-default-entrypoints/tests/index.test.ts
```diff
@@ -0,0 +1,158 @@
...
+					"Hello from worker-🐝 scheduled()"
+				);
+			});
+		});
+
+		/*
+		 * Assert that WorkerA works as intended
+		 */
+		describe("WorkerA", () => {
+			// [[ CURRENTLY BROKEN ]]
```
emily-shen, line 94: should this be `it.fails` then?

fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
CarmenPopoviciu: I have queues on my list to add to this PR

fixtures/workers-with-assets-and-rpc-default-entrypoints/tests/index.test.ts
```diff
@@ -0,0 +1,158 @@
...
+				// `http://fakehost/cdn-cgi/mf/scheduled` instead
+				let response = await fetch(`http://${ipWorkerA}:${portWorkerA}`);
+				let text = await response.text();
+				expect(response.status).toBe(200);
+				expect(text).toContain("env.WORKER_B.scheduled() response: undefined");
+			});
+		});
+	});
+
+	describe.todo("Workers running in the same wrangler dev session", () => {});
```
emily-shen, line 157: these tests already exist in `e2e/multiworker-dev.test.ts`, maybe you could just add a comment pointing there 😄

fixtures/workers-with-assets-and-http-based-service-bindings/workerB/src/index.ts
CarmenPopoviciu: ✅

fixtures/workers-with-assets-and-http-based-service-bindings/README.md
CarmenPopoviciu: ✅

fixtures/workers-with-assets-and-rpc-default-entrypoints/README.md
CarmenPopoviciu: actually...I'm gonna hold off on this one. merging the two makes everything hardly readable anymore

fixtures/workers-with-assets-and-rpc-named-entrypoints/tests/index.test.ts
penalosa: Scheduled handlers are only for default entrypoints—as with routes, there's no way to attach them to a names entrypoint (in prod or otherwise). I think this is expected behaviour rather than a bug

fixtures/workers-with-assets-and-rpc-default-entrypoints/workerA/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+		const beeResult = await env.WORKER_B.bee();
+
+		// test named functions with params
+		const busyBeeResult = await env.WORKER_B.busyBee("🐝");
+
+		// tests Cron Triggers
+		const scheduledRequest = new Request(
+			"http://fakehost/cdn-cgi/mf/scheduled"
+		);
+		const scheduledResponse = await env.WORKER_B.scheduled(scheduledRequest);
```
penalosa, line 19: Same comment as for the other fixture re scheduled handlers

penalosa (changes requested):
I've left some comments about scheduled handler in particular, but I'm also a bit confused about the the three fixtures here:
- What's the difference between the first and the second? They're named as RPC and HTTP, but Wrangler will always connect service bindings via JSRPC (assuming up-to-date Wrangler versions etc..., which applies here)
- Do we need a separate fixture for default and named? It feels like that could be part of the same fixture

fixtures/workers-with-assets-and-service-bindings/workerA/src/workerB.util.ts
```diff
@@ -0,0 +1,28 @@
...
+	const beeResult = await response.text();
+
+	// test named functions with multiple parameters
+	// named functions with variable number of args are only supported for
+	// Workers using the class-based syntax
+
+	// tests Cron Triggers
+	const scheduledRequest = new Request("http://fakehost/cdn-cgi/mf/scheduled");
+	const scheduledResponse =
+		await env.DEFAULT_EXPORT.scheduled(scheduledRequest);
```
GregBrimble, line 19: I've never anything a caller invoke a callee's `scheduled` handler before. Does that work in production?

Second, the parameter shouldn't be a `Request`. It should be a [`ScheduledEvent`](https://unpkg.com/browse/@cloudflare/workers-types@4.20250214.0/experimental/index.d.ts#:~:text=declare%20abstract%20class%20ScheduledEvent,%7D).

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
```diff
@@ -0,0 +1,201 @@
...
+						"env.DEFAULT_EXPORT.bee() response: Hello from worker-b bee()"
+					);
+					expect(text).toContain(
+						"env.DEFAULT_ENTRYPOINT.bee() response: Hello from worker-c bee()"
+					);
+					expect(text).toContain(
+						"env.NAMED_ENTRYPOINT.bee() response: Hello from worker-d bee()"
+					);
+					expect(text).toContain(
+						'env.DEFAULT_EXPORT.busyBee("🐝") response: Not supported. When calling a top-level handler function that is not declared as part of a class, you must always send exactly one argument. In order to support variable numbers of arguments, the server must use class-based syntax (extending WorkerEntrypoint) instead.'
```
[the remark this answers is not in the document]
CarmenPopoviciu: I could have sworn I tested this and I got that error if the param wasn't of type `Request`. I tested again and you're right, seems to be working just fine :/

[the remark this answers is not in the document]
CarmenPopoviciu: ✅

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+			afterAll(async () => {
+				await stopWorkerA?.();
+				await stopWorkerB?.();
+				await stopWorkerC?.();
+				await stopWorkerD?.();
+			});
+
+			describe("WorkerA", () => {
+				// this currently incorrectly returns worker-d's User Worker response
```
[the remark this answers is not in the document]
CarmenPopoviciu: ✅

fixtures/workers-with-assets-and-service-bindings/workerA/src/workerB.util.ts
CarmenPopoviciu: sorry y'all ... this is me being very confused about how to trigger these `scheduled` events 🙇 . 

Right, so basically we could either trigger the event via the API
```
env.DEFAULT_EXPORT.scheduled({cron: "* * * * *"});
```

or by calling
```
env.DEFAULT_EXPORT.fetch("https://example.com/__scheduled");
```
which would require local dev to run with the `--test-scheduled` flag on, which these fixtures are already doing.

I tested `env.DEFAULT_EXPORT.scheduled({cron: "* * * * *"});` in prod by deploying `worker-a` and `worker-b`, and tailing `worker-b` and can confirm it works, so perhaps worth settling on that

👉  https://worker-a.carmen-popoviciu643.workers.dev/
👇 
Screenshot 2025-03-05 at 12 44 26

fixtures/workers-with-assets-and-service-bindings/workerA/src/workerC.util.ts
```diff
@@ -0,0 +1,23 @@
...
+	// test named functions without parameters
+	const beeResult = await env.DEFAULT_ENTRYPOINT.bee();
+
+	// test named functions with parameters
+	const busyBeeResult = await env.DEFAULT_ENTRYPOINT.busyBee("🐝");
+
+	// tests Cron Triggers
+	const scheduledRequest = new Request("http://fakehost/cdn-cgi/mf/scheduled");
+	const scheduledResponse =
+		await env.DEFAULT_ENTRYPOINT.scheduled(scheduledRequest);
```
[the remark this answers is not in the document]
CarmenPopoviciu: this should be resolved now, but not closing so you can dbl check

GregBrimble (approved):
Clever setup!

I'd like to extend this with some more of the risky things (e.g. RPC stubs and WebSockets). And the Cron/Queues concerns that I had were more about making sure that the normal front door for those features was still working (i.e. I should be able to hit `/__scheduled` as a client rather than anything cross-Worker). But this is absolutely worth landing as it is and we can add in these bits when we get to them :)

fixtures/workers-with-assets-and-service-bindings/workerA/src/workerB.util.ts
GregBrimble: TIL! Not really surprising, I guess, but still, cross-Worker invocation of `scheduled()` like this is weird!

fixtures/workers-with-assets-and-service-bindings/workerA/src/workerB.util.ts
```diff
@@ -0,0 +1,25 @@
+export async function getWorkerBResponses(request: Request, env) {
+	let response: Response;
+
+	// test fetch requests (includes both assets and User Worker routes)
+	response = await env.DEFAULT_EXPORT.fetch(request);
+	const fetchResponse = await response.text();
+
+	// test named functions without parameters
+	const beeResult = await env.DEFAULT_EXPORT.bee("🐝");
```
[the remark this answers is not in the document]
CarmenPopoviciu: oeps!

CarmenPopoviciu:
> Clever setup!
> 
> I'd like to extend this with some more of the risky things (e.g. RPC stubs and WebSockets). And the Cron/Queues concerns that I had were more about making sure that the normal front door for those features was still working (i.e. I should be able to hit `/__scheduled` as a client rather than anything cross-Worker). But this is absolutely worth landing as it is and we can add in these bits when we get to them :)

Oh, I actually had those tests for `/__scheduled` and I removed them because I thought the cross Worker stuff was the most interesting part. I'll re-add them, no biggie

GregBrimble:
A bit different to how you had them before, I think. Worker A shouldn't be able to call `http://workerb/__scheduled` over the binding. Instead, I'm interested in ensuring that a request to `http://localhost:workerBPort/__scheduled` should work, even when Worker B has the RPC proxy in front of it. Just as I'm interested in making sure it can still consume queue messages when it has the RPC proxy in front of it.

CarmenPopoviciu:
> A bit different to how you had them before, I think. Worker A shouldn't be able to call `http://workerb/__scheduled` over the binding. Instead, I'm interested in ensuring that a request to `http://localhost:workerBPort/__scheduled` should work, even when Worker B has the RPC proxy in front of it. Just as I'm interested in making sure it can still consume queue messages when it has the RPC proxy in front of it.

right on. Me and my rebasing 😒 ... but what I had before was something along the lines of 
`await fetch(`http://${ipWorkerB}:${portWorkerB}/__scheduled`);`

GregBrimble:
Ah sorry, I maybe just missed that or misunderstood in my earlier review. Great, that sounds good to me, thanks!

penalosa:
@GregBrimble @CarmenPopoviciu I'm not sure there's much value in testing that `/__scheduled` still works. That's an in-worker feature powered by middleware, so if a general fetch works, so will `/__scheduled`

GregBrimble:
If you don't think so, then I can trust your judgement. To me, we're testing the _platform feature_ of scheduled, rather than the _implementation detail_ that it happens to go through `fetch`. Same with queues (and it just happens to be that this one isn't implemented with `fetch`).

Yup, fair point about that working cross-Worker, and I agree: let's avoid promoting that. Feels like is just a leak (and honestly even the `http://localhost:workerBPort/__scheduled` feels a bit hacky to me too).

penalosa (approved):
LGTM! Thanks for all the hard work on this!

PR #7981, opened 2025-01-31
Fix assets inheritance

WalshyDev:
Fixes N/A

Fixes a regression introduced in Wrangler 3.107.0 in which `[assets]` was not being inherited from the top-level environment.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no change

### DIFF
--- a/packages/wrangler/src/__tests__/config/configuration.test.ts
+++ b/packages/wrangler/src/__tests__/config/configuration.test.ts
@@ -6396,6 +6396,154 @@ describe("normalizeAndValidateConfig()", () => {
 				expect(result2.diagnostics.hasWarnings()).toBe(false);
 			});
 		});
+
+		describe("[assets]", () => {
+			it("should inherit from top-level assets", () => {
+				const rawConfig: RawConfig = {
+					assets: {
+						directory: "dist",
+						binding: "ASSETS",
+					},
+					env: {
+						ENV1: {},
+					},
+				};
+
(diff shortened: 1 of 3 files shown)


WillTaylorDev (approved):

packages/wrangler/src/__tests__/config/configuration.test.ts
```diff
@@ -6396,6 +6396,117 @@ describe("normalizeAndValidateConfig()", () => {
 				expect(result2.diagnostics.hasWarnings()).toBe(false);
 			});
 		});
+
+		describe("[assets]", () => {
+			it("should inherit from top-level assets", () => {
```
CarmenPopoviciu, line 6401: I feel like we should be already having a test for this, but then again here we are...so just to be abs sure, can we test that if we config assets at both top level and env level, the config resolves correctly?

CarmenPopoviciu (approved):
pending on that test comment, this LGTM. Thank you so much for fixing Walshy ♥️

packages/wrangler/src/__tests__/config/configuration.test.ts
WalshyDev: Good call, added

PR #7984, opened 2025-01-31
add telemetry to auto provisioning

emily-shen:
Fixes [DEVX-1618]
Sends a new adhoc telemetry event since the dispatcher cannot (currently) be used in command handlers.
draft amplitude dash: https://app.amplitude.com/analytics/cloudflare/dashboard/4h1lunzd

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: unreleased

### DIFF
--- a/.changeset/bright-islands-sleep.md
+++ b/.changeset/bright-islands-sleep.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+feat: add telemetry to experimental auto-provisioning
--- a/packages/wrangler/src/__tests__/provision.test.ts
+++ b/packages/wrangler/src/__tests__/provision.test.ts
@@ -1,4 +1,5 @@
 import { http, HttpResponse } from "msw";
+import { logger } from "../logger";
 import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
 import { mockConsoleMethods } from "./helpers/mock-console";
 import { clearDialogs, mockPrompt, mockSelect } from "./helpers/mock-dialogs";
@@ -30,6 +31,7 @@ describe("--x-provision", () => {
 	const { setIsTTY } = useMockIsTTY();
 
 	beforeEach(() => {
+		logger.loggerLevel = "debug";
 		setIsTTY(true);
 		msw.use(
 			...mswSuccessDeploymentScriptMetadata,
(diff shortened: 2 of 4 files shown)


.changeset/bright-islands-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+feat: add telemetry to experimental auto-provisioning
```
edmundhung, line 5: Just a sanity check: This is a patch because it is an experimental feature, right?

packages/wrangler/src/deployment-bundle/bindings.ts
```diff
@@ -85,7 +86,9 @@ abstract class ProvisionResourceHandler<
 		public type: T,
 		public binding: B,
 		public idField: keyof B,
-		public accountId: string
+		public accountId: string,
+		/** For telemetry */
+		public summary: Record<string, number>
```
penalosa, line 91: An instance of this class represents a _single_ binding—it feels odd to be passing in a record of _multiple_ bindings here? Why does this need to be done insdie the class? Can't we just count the number of items in pendingResources?

packages/wrangler/src/__tests__/provision.test.ts
```diff
@@ -30,6 +31,7 @@ describe("--x-provision", () => {
 	const { setIsTTY } = useMockIsTTY();
 
 	beforeEach(() => {
+		logger.loggerLevel = "debug";
```
edmundhung, line 34: You might need to restore the loggerLevel after the test or it will create a side effect for the rest of the tests.

packages/wrangler/src/deployment-bundle/bindings.ts
emily-shen: yeah it is odd :/ i wanted access to the inherit/create/connect methods to record what happens to each resource, not just that they get provisioned in some way (connected or created), especially as some resources get connected before `runProvisioningFlow` and some after. 

Although now that i'm revisiting it, i think that's probably a level of granularity that isn't necessary/important and we can just count pendingResources... yeah i'll simplify this

emily-shen:
#8077

PR #7994, opened 2025-01-31
feat(wrangler): support `-C, --cwd` cli argument

### LINKED ISSUE #7993: 🚀 Feature Request: Support argument for wrangler to run in different working directory
### Describe the solution

With the introduction of the generated wrangler config (https://github.com/cloudflare/workers-sdk/pull/7442), some commands like `wrangler pages dev|deploy` require to be running in the correct working directory to be working as intended and reading generated wrangler config (or user config).

A global option like `-C, --cwd` in wrangler CLI can guarantee we can reliably run all commands in the project working directory without depending on the package manager monorepo features (which is not always possible).

----

**Context:** Nitro (and Nuxt) generates command hints after building for cloudflare pages. These commands are meant to be directly running in the user CLI and relative to where the build was issued for example:

```
$ <pm> nitro build path/to/project
✔ You can preview this build using npx wrangler pages dev path/to/project                              
✔ You can deploy this build using npx wrangler pages deploy path/to/project
```

Working on https://github.com/nitrojs/nitro/pull/2949, noticed the above hint is not reliable because it won't use any of the wrangler configs from `path/to/project`

It is possible to manually point to the config path with `-c, --config` for modules in the example above but for pages, the only way is to somehow change CWD we need to show something like `cd path/to/project && npx wrangler pages dev`.

Quickly discussed with @petebacondarwin if we could make it a better DX from Wrangler CLI itself.

pi0:
Fixes #7993 (/cc @petebacondarwin)

This PR adds `-C, --cwd` global argument to the `wrangler` CLI to allow changing the current working directory before running any command.

This feature is used in most of CLI tools that depend on right working directory but different names. I have chosen `-C, --cwd` following with [pnpm](https://pnpm.io/pnpm-cli#-c-path---dir-path) (`-C`) + [yarn](https://classic.yarnpkg.com/lang/en/docs/cli/#toc-cwd) (`--cwd`), but open to ideas changing it to anything else.

This allows better DX of using `wrangler` commands from any working directory  to another (where project is).

```
$ wrangler --help

GLOBAL FLAGS
  -c, --config   Path to Wrangler configuration file  [string]
  -C, --cwd      Run as if wrangler was started in <path> instead of the current working directory  [string]
  -e, --env      Environment to use for operations, and for selecting .env and .dev.vars files  [string]
  -h, --help     Show help  [boolean]
  -v, --version  Show version number  [boolean]
```

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because: It is generic CLI behavior (is it needed?)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: effect of global arg has unit test
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/19665

### DIFF
--- a/packages/wrangler/src/api/dev.ts
+++ b/packages/wrangler/src/api/dev.ts
@@ -183,6 +183,7 @@ export async function unstable_dev(
 		ip: "127.0.0.1",
 		inspectorPort: options?.inspectorPort ?? 0,
 		v: undefined,
+		cwd: undefined,
 		localProtocol: options?.localProtocol,
 		httpsKeyPath: options?.httpsKeyPath,
 		httpsCertPath: options?.httpsCertPath,
--- a/.changeset/khaki-years-act.md
+++ b/.changeset/khaki-years-act.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Add `-C, --cwd` global argument to the `wrangler` CLI to allow changing the current working directory before running any command.
--- a/packages/wrangler/src/__tests__/ai.test.ts
+++ b/packages/wrangler/src/__tests__/ai.test.ts
@@ -27,6 +27,7 @@ describe("ai help", () => {
 
 			GLOBAL FLAGS
 			  -c, --config   Path to Wrangler configuration file  [string]
+			  -C, --cwd      Run as if Wrangler was started in the specified directory instead of the current working directory  [string]
 			  -e, --env      Environment to use for operations, and for selecting .env and .dev.vars files  [string]
 			  -h, --help     Show help  [boolean]
 			  -v, --version  Show version number  [boolean]"
@@ -55,6 +56,7 @@ describe("ai help", () => {
 
 			GLOBAL FLAGS
 			  -c, --config   Path to Wrangler configuration file  [string]
+			  -C, --cwd      Run as if Wrangler was started in the specified directory instead of the current working directory  [string]
 			  -e, --env      Environment to use for operations, and for selecting .env and .dev.vars files  [string]
 			  -h, --help     Show help  [boolean]
 			  -v, --version  Show version number  [boolean]"
--- a/packages/wrangler/src/__tests__/cert.test.ts
+++ b/packages/wrangler/src/__tests__/cert.test.ts
@@ -440,6 +440,7 @@ describe("wrangler", () => {
 
 						GLOBAL FLAGS
 						  -c, --config   Path to Wrangler configuration file  [string]
+						  -C, --cwd      Run as if Wrangler was started in the specified directory instead of the current working directory  [string]
 						  -e, --env      Environment to use for operations, and for selecting .env and .dev.vars files  [string]
 						  -h, --help     Show help  [boolean]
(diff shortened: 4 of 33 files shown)


petebacondarwin (approved):
LGTM and happy with the naming.

petebacondarwin:
A minor linting issue to fix https://github.com/cloudflare/workers-sdk/actions/runs/13079905513/job/36575863616?pr=7994#step:5:916

pi0:
updated.

penalosa (commented):
I'm a bit uneasy with the `-C` shortening here—I know it's a relatively common convention, but we also have a global flag `-c` for config, and those seem pretty likely to be mixed up to me. Not a blocker though.

However, this really should have tests.

pi0:
Added basic test in cfd81f1909c007e0a39ba9c5c1120ef12e7f1a3f

pi0:
@penalosa can you please review this? 🙏🏼

packages/wrangler/src/api/dev.ts
```diff
@@ -183,6 +183,7 @@ export async function unstable_dev(
 		ip: "127.0.0.1",
 		inspectorPort: options?.inspectorPort ?? 0,
 		v: undefined,
+		cwd: undefined,
```
petebacondarwin, line 186: Not related directly to this PR... I wonder if we should actually change `StartDevOptions` to omit properties like this? Rather than having to pass in undefined?

petebacondarwin (approved):
Still LGTM

CarmenPopoviciu:
I share @penalosa 's concern about `-c` and `-C` being confusing for users. I spoke with @petebacondarwin offline and he seemed fine with dropping `-C` altogether in favour of just `--cwd`. I'd like to propose we do that, unless there is other shortening that would make sense in this case. @pi0 any strong objections to that? I'm happy to help get this over the finish line if you don't have the bandwidth ♥️

pi0:
thanks for followup, dear @CarmenPopoviciu. `-C` is really standard in most of UNIX tools but of course totally fine to only keep `--cwd` (it was main requirement)

Do you mind to push change to this PR if have time? 🙏🏼

CarmenPopoviciu:
> -C is really standard in most of UNIX tools 
yeah, I noticed that being mentioned, so I was a bit torn on what the right thing to do is. Since changing the `-c` shortening would have much bigger impact, I figured dropping `-C` would be the most straightforward thing to do. Let's see where this change lands with users, and if there is an ask for `-C` we can revisit this decision and add it later.

> Do you mind to push change to this PR if have time
totally happy to do that ♥️ . Don't think I can push directly to this PR since it's from a forked repo, so I might open a PR to your branch instead if that's alright?

CarmenPopoviciu:
@pi0 👉 https://github.com/pi0/workers-sdk/pull/1 ^.^

CarmenPopoviciu:
I've opened a new [PR](https://github.com/cloudflare/workers-sdk/pull/8253) based off of this one, incl. the changes I suggested, so we can get this across the finish line

PR #8016, opened 2025-02-03
fix(vite-plugin-cloudflare): Usage with Workers AI

justinvdm:
## Context
Hi there o/

At RedwoodJS, we're trying out `@cloudflare/vite-plugin` alongside Workers AI, but we've run into an issue where the two don't seem to work together.

I've made a PR mainly to illustrate where the issue is - feel free to use or disregard the solution as needed. For this same reason I'm marking this as a draft, and will bring it up with Cloudflare folks in Redwood's slack.

 ## Problem

When using `@cloudflare/vite-plugin` for a worker with an `[ai]` binding in the `worker.toml`, we end up getting this error:

```
workerd/server/workerd-api.c++:753: error: wrapped binding module can't be resolved (internal modules only); moduleName = miniflare-internal:wrapped:__WRANGLER_EXTERNAL_AI_WORKER
```

This seems to be because the miniflare options include the wrapped bindings mapping to an internal AI worker script, yet the corresponding worker options for this worker are not included in the miniflare options.

 ## Solution

* In `wrangler`: Include `externalWorkers` in the results of [`unstable_getMiniflareWorkerOptions`](https://github.com/cloudflare/workers-sdk/blob/a7163b3a21f56c9bd839e34c9d5f31c3099a585a/packages/wrangler/src/api/integrations/platform/index.ts#L273) - it contains the worker options for the AI worker
* In `@cloudflare/vite-plugin-cloudflare`: For the bindings in `wrappedBindings`, if their corresponding scripts are in `externalWorkers`, add these worker options to the miniflare options

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included: Only e2e test included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: There was a backwards compatible change to `unstable_getMiniflareWorkerOptions` - the returned object now has `externalWorkers` added to it.


packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -288,69 +288,99 @@ export function getDevMiniflareOptions(
...
-							},
-							serviceBindings: {
-								...workerOptions.serviceBindings,
-								...(environmentName ===
-									resolvedPluginConfig.entryWorkerEnvironmentName &&
-								workerConfig.assets?.binding
-									? {
-											[workerConfig.assets.binding]: ASSET_WORKER_NAME,
+							externalWorkers,
+							worker: {
```
justinvdm, line 298: The diff here is misleading - this was the only change for this part of code

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -288,69 +288,99 @@ export function getDevMiniflareOptions(
...
+			const scriptName =
+				typeof binding === "string" ? binding : binding.scriptName;
+
+			const externalWorker = resolvedExternalWorkersMap.get(scriptName);
+
+			if (externalWorker) {
+				externalWorkersMap.set(scriptName, externalWorker);
+			}
+		}
+	}
```
justinvdm, line 382: The idea is:
* only include the workers that map to script names referenced in `wrappedBindings`
* only include them once - in case, for e.g. there were multiple workers each with their own AI binding, I think we'd end up with more than once reference to the internal AI worker script

packages/vite-plugin-cloudflare/playground/ai/__tests__/basic-functionality.spec.ts
```diff
@@ -0,0 +1,6 @@
+import { expect, test } from "vitest";
+import { getTextResponse } from "../../__test-utils__";
+
+test("basic hello-world functionality", async () => {
+	expect(JSON.parse(await getTextResponse())).toContain("PONG");
+});
```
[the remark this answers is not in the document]
petebacondarwin: Normally we provide a CLOUDFLARE_ACCOUNT_ID env variable to avoid such login popups, but so far I don't think we've needed that in the Vite plugin work, since it has all been local only. I think this might be one of the first wrappedBindings that has a test?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
petebacondarwin: Is this actually necessary?

In Wrangler's `MultiWorkerRuntimeController` it appears that any external workers are all just concatenated together. It doesn't do this kind of filtering and deduping.

I am pretty sure that Miniflare only adds external workers that are relevant to the bindings in the config. So if there is no AI binding there would be no equivalent external worker. See https://github.com/cloudflare/workers-sdk/blob/444a6302f194150b0678da5b564cfd2de8a3dad6/packages/wrangler/src/dev/miniflare.ts#L589-L608

petebacondarwin (commented):
I think this is generally the right idea. Thanks for the investigation and suggested solution. Let's iterate and get it in.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
petebacondarwin: Actually, while I think the filtering is not needed, the deduping probably is. The fact that the `MultiWorkerRuntimeController` doesn't do this might be a bug! cc @penalosa

justinvdm: @petebacondarwin if we remove this `for` block so that we're no longer filtering to only the workers that have corresponding `wrappedBindings`, then I get this when I have workers referencing Durable Objects:

```
service core:user:__WRANGLER_EXTERNAL_DURABLE_OBJECTS_WORKER: Worker "core:user:__WRANGLER_EXTERNAL_DURABLE_OBJECTS_WORKER"'s binding "SessionDO" refers to a service "core:user:worker", but no such service is defined.
```

packages/vite-plugin-cloudflare/playground/ai/__tests__/basic-functionality.spec.ts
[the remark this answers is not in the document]
justinvdm: Looking at the built types for options the plugin takes in, I think there might not be a way to set an account id dynamically.

```
type PersistState = boolean | {
    path: string;
};
interface BaseWorkerConfig {
    viteEnvironment?: {
        name?: string;
    };
}
interface EntryWorkerConfig extends BaseWorkerConfig {
    configPath?: string;
}
interface AuxiliaryWorkerConfig extends BaseWorkerConfig {
    configPath: string;
}
interface PluginConfig extends EntryWorkerConfig {
    auxiliaryWorkers?: AuxiliaryWorkerConfig[];
    persistState?: PersistState;
}

declare function cloudflare(pluginConfig?: PluginConfig): vite.Plugin;
```

Considering that, to avoid scope creep in the PR, I skipped the test for now with a comment: https://github.com/cloudflare/workers-sdk/pull/8016/commits/b4d16b5ab247638ff15e62d1060ad22e1cb04132

Happy to go a different route though, or even just remove the test or example :)

justinvdm:
@petebacondarwin thank you for taking a look! I think this is ready for another look now.

wadefletch:
I'm also encountering this issue with `@cloudflare/vitest-pool-workers`–is it correct to read that this same fix could be applied there as well?

justinvdm:
> @petebacondarwin thank you for taking a look! I think this is ready for another look now.

Hi @petebacondarwin, just following up - can you take a look when you have a moment?

I also totally don't mind us closing this one if it would be better for someone at CF to make this change instead. The PR was just a way to show the parts that I saw needed changing to get Workers AI working, in case it helped.

petebacondarwin:
Sorry for the delay. Thank you for driving this. We were discussing what to do about auth and where it should be placed.

thomasgauvin:
Looking forward to vite-plugin supporting ai bindings, thanks for driving folks

packages/vite-plugin-cloudflare/src/miniflare-options.ts
penalosa: > Actually, while I think the filtering is not needed, the deduping probably is. The fact that the `MultiWorkerRuntimeController` doesn't do this might be a bug! cc @penalosa

Ooh, that does sound like a bug! I suspect multiple workers with an AI binding would cause Miniflare to complain

packages/vite-plugin-cloudflare/playground/ai/__tests__/basic-functionality.spec.ts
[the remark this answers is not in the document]
penalosa: I think this should probably be in an E2E test? We have an existing test for Workers AI with Wrangler

(not a blocker for this PR though—it'll be a bit involved to set up Vite with E2E testing)

packages/vite-plugin-cloudflare/src/miniflare-options.ts
penalosa: I'm not sure we need this filtering though—external workers should only be added when needed in miniflare.ts. I think instead we might need to add worker-name based prefixes to the external workers? i.e. `scriptName: workerName-EXTERNAL_AI_WORKER_NAME` or something like that

packages/vite-plugin-cloudflare/playground/ai/__tests__/basic-functionality.spec.ts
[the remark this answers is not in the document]
petebacondarwin: I recently set up Vite e2e testing with the mock-npm-registry, so might not be so difficult.

justinvdm:
@penalosa thank you so much for continuing this in #8273! 🙇 I think I'll close this PR then, but let me know if I can help :)

PR #8021, opened 2025-02-04
Fix pages functions using unenv aliased packages

### LINKED ISSUE #7572: 🐛 BUG: wrangler pages dev failing with error on Identifier '__cf_cjs' has already been declared

Pages

Wrangler 3.96.0

N/A

Windows 11

### Observed behavior
```
X [ERROR] service core:user:bitter-bread-d12c: Uncaught SyntaxError: Identifier '__cf_cjs' has already been declared

    at null.<anonymous> (fn7s4pcgxfq.js:1026)
```

### Expected behavior
App starts

### Steps to reproduce

Occurs when the code below is added as a function

Note: You will need to modify line 251 in json-bigint and comment out //case 'null': not valid because your linter also causes a fatal failure on this line
```
import { PromptRole } from "@llumiverse/core";
import { TogetherAIDriver } from "@llumiverse/drivers";
export async function onRequestPost(ctx) {
    const driver = new TogetherAIDriver({
        apiKey: "<some key>"
    });

    const prompt = [
        {
            role: PromptRole.user,
            content: 'Please, write a short story about winter in Paris, in no more than 512 characters.'
        }
    ]

    const body = await driver.execute(prompt, {
        model:"meta-llama/Meta-Llama-3.1-8B-Instruct-Turbo",
        temperature: 0.6,
        max_tokens: 1024
    });
    return new Response(200,{body:JSON.stringify(body)})
    //return new Response(200,{body:JSON.stringify(prompt)});
}
```

_No response_

[wrangler-2024-12-17_15-58-15_864.log](https://github.com/user-attachments/files/18168308/wrangler-2024-12-17_15-58-15_864.log)

0xD34DC0DE:
Fixes #7572

- Added a test fixture to test the edge case where pages functions using `nodejs_compat` (v2) together with the import of an npm package getting aliased by unenv would trigger the addition of `__cf_cjs` multiple times
- Replace the `__cf_cjs` esm to cjs converter function by an inline expression to avoid the use of esbuild banners that might get included in the bundle multiple time.

A separate fixture has been created to avoid making all tests in `fixtures\pages-nodejs-v2-compat` fail for an unrelated reason since the runtime fails to start when the issue is present.
The test fixture was inspired by the `nodejs-hybrid-app` and `pages-nodejs-v2-compat` fixtures.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: should be adequately covered by fixtures
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Bugfix


packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -143,29 +143,13 @@ function handleUnenvAliasedPackages(
...
-		`;
-
 	build.onLoad(
 		{ filter: /.*/, namespace: REQUIRED_UNENV_ALIAS_NAMESPACE },
 		({ path }) => {
 			return {
 				contents: dedent`
 					import * as esm from '${path}';
-					module.exports = __cf_cjs(esm);
+					module.exports = Object.assign(esm.default || {}, ...Object.entries(esm).filter(([k,]) => k !== 'default'));
```
petebacondarwin, line 152: Did we use `defineProperty` for a reason? Is it possible that the `default` property object is locked down from being written to?

petebacondarwin:
The good news is that our test coverage seems to check these changes!
The bad news is that they appear to be unhappy with this change.

0xD34DC0DE:
> The good news is that our test coverage seems to check these changes!
> The bad news is that they appear to be unhappy with this change.

Thanks for the review, I'll try to see why the tests aren't successful when they're executed on the CI, my first guess would be the use of  transitive includes (cross-fetch and cross-env)?

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
vicb, line 152: `esm.default` can be false(y)

```suggestion
					module.exports = Object.assign('default' in esm ? esm.default : {}, ...Object.entries(esm).filter(([k,]) => k !== 'default'));
```

vicb: Good catch, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#description

```ts
module.exports = Object
  .defineProperties(
    esm.default || {}, 
    Object.fromEntries(
      Object.entries(esm)
         .filter(([k,]) => k !== 'default'))
         .map(([k, value]) => [k, {value, enumerable: true}])
    )    
  );

Comment (0xD34DC0DE) [reply packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts]: Got it! Falsy values are now handled.

Comment (0xD34DC0DE) [reply packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts]: I fixed the exports assigment to use Object.defineProperty to ensure that the properties are enumerable.

File: fixtures/pages-functions-unenv-alias/functions-deps/dep.cjs [hunk 2]
```diff
@@ -0,0 +1,9 @@
+// 2 imports required to trigger https://github.com/cloudflare/workers-sdk/issues/7572
```
vicb, line 1: Are you sure it takes 2 aliased modules? 
I don't understand why as the banner seems to correctly be added only once?

0xD34DC0DE: You're right, only 1 import triggers the issue. I've removed the unnecessary require (cross-env).

I was under the impression that the error got triggered when using at least 2 imports, but this was during the time I was trying to get the tests to run correctly at first before the fix. I was getting weird results with the tests after I started the fork.

fixtures/pages-functions-unenv-alias/tests/index.test.ts
```diff
@@ -0,0 +1,18 @@
...
+    it("should run dev server when importing 2 unenv aliased packages", async ({ expect, onTestFinished }) => {
+        const { ip, port, stop } = await runWranglerPagesDev(
+            resolve(__dirname, ".."),
+            "./functions",
+            ["--port=0", "--inspector-port=0"]
+        );
+        onTestFinished(stop);
+        const response = await fetch(`http://${ip}:${port}/`);
+        const body = await response.text();
+        expect(body).toMatchInlineSnapshot(`"OK!"`);
```
vicb, line 16:
```suggestion
        expect(body).toEqual("OK!");
```

vicb (commented):
Thanks a lot for your work on this - and coping with my reviews ;)

One last comment and this should be good to go

fixtures/pages-functions-unenv-alias/functions-deps/dep.cjs
```diff
@@ -0,0 +1,6 @@
+const fetch = require("cross-fetch");
+
+module.exports = () => {
+    return "OK!";
+};
+module.exports.fetch = fetch;
```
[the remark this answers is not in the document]
0xD34DC0DE: The cjs is here to ensure an esm style import from a cjs module.
"cross-fetch" might be exporting both esm and cjs.

I did try to run the unpatched version together with a direct import `import * as fetch from "cross-fetch"` but `__cf_cjs` wasn't even getting called in the resulting bundle.
I am not sure if the duplication happens when importing a module that imports something (together with the double bundling) or just when importing cjs.

However, I'm willing to try with a cjs only npm package, but I don't know which one would fit the bill.

fixtures/pages-functions-unenv-alias/tests/index.test.ts
0xD34DC0DE: 👍

fixtures/pages-functions-unenv-alias/functions-deps/dep.cjs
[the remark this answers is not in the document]
vicb: My understanding of the issue is that it happen with `require(<package aliased in unenv>)`,  so would `const fetch = require("cross-fetch");` in `[[path]].ts` build?

`require("cross-fetch")` should actually import https://github.com/unjs/unenv/blob/main/src/runtime/npm/cross-fetch.ts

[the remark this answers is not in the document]
vicb: it has to be a require over an import:

https://github.com/cloudflare/workers-sdk/blob/ee4873c963b89b0091a2ff0952c274ef9dbf05ad/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts#L129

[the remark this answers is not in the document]
0xD34DC0DE: Yep, I tested the unpatched version using a direct `require` and it does cause the issue.
I've removed the `dep.cjs` file in favor of just a single `require`.

Thank you for taking the time to explain and pinpointing the code in question.

vicb:
@0xD34DC0DE Sorry I was bad at following up here. Thanks for your work here and let's get this in today

vicb (approved):

vicb:
Merci Émerick and sorry again for letting this unattended for so long :(

PR #8026, opened 2025-02-04
Support profiling the startup phase of a Worker

penalosa:
This PR adds two features (separate changesets):
* Support for the `--outfile` argument for `wrangler deploy`, which outputs the exact Worker bundle that Wrangler will upload to the Cloudflare API. This has more information than just `--outdir` because it bakes in bindings & module types.
* `wrangler check startup`, which profiles the startup time of a Worker and generates a `.cpuprofile` file that can be opened in Chrome DevTools/directly in VSCode for further investigation. Additionally, this is automatically generated when Wrangler encounters a startup time error from the API.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/deploy/index.ts
```diff
@@ -301,6 +306,10 @@ async function deployWorker(args: DeployArgs) {
 		);
 	}
 
+	if (args.outfile && args.outdir) {
+		throw new UserError("Cannot use `--outfile` and `--outdir` together");
```
emily-shen, line 310: why not? don't they output different things...?

packages/wrangler/src/utils/friendly-validator-errors.ts
```diff
@@ -0,0 +1,75 @@
...
+			false
+		);
+		const profile = path.relative(
+			projectRoot ?? process.cwd(),
+			path.join(tmpDir.path, `worker.cpuprofile`)
+		);
+		await writeFile(profile, JSON.stringify(cpuProfile));
+		throw new UserError(dedent`
+			Your Worker failed validation because it exceeded startup limits.
+			To ensure fast responses, we place constraints on Worker startup—like how much CPU it can use, or how long it can take. Your Worker failed validation, which means it hit one of these startup limits. Try reducing the amount of work done during startup (outside the event handler), either by removing code or relocating it inside the event handler.
```
emily-shen, line 69:
```suggestion
			To ensure fast responses, there are constraints on Worker startup, such as how much CPU it can use, or how long it can take. Your Worker has hit one of these startup limits. Try reducing the amount of work done during startup (outside the event handler), either by removing code or relocating it inside the event handler.
```

packages/wrangler/src/utils/friendly-validator-errors.ts
```diff
@@ -0,0 +1,75 @@
...
+
+			A CPU Profile of your Worker's startup phase has been written to ${profile} Load it into the Chrome DevTools profiler (or directly in VSCode) to view a flamegraph.
```
emily-shen, line 71:
```suggestion
			A CPU Profile of your Worker's startup phase has been written to ${profile} - load it into the Chrome DevTools profiler (or directly in VSCode) to view a flamegraph.
```

packages/wrangler/src/deploy/index.ts
```diff
@@ -63,6 +63,11 @@ export function deployOptions(yargs: CommonYargsArgv) {
 				type: "string",
 				requiresArg: true,
 			})
+			.option("outfile", {
```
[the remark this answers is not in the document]
penalosa: It's pretty generically useful, since it gives you the exact Worker bundle that Wrangler will upload to the API. It could be good for running a build and then deploying that later on or keeping a record of the build for later analysis.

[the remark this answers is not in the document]
penalosa: (it's also pretty helpful for debugging how Wrangler builds)

packages/wrangler/src/deploy/index.ts
penalosa: Fair point!

emily-shen (approved):

packages/wrangler/src/check/commands.ts
```diff
@@ -0,0 +1,242 @@
...
+import { UserError } from "../errors";
+import { logger } from "../logger";
+import { getWranglerTmpDir } from "../paths";
+import type { Config } from "../config";
+import type { ModuleDefinition } from "miniflare";
+import type { FormData, FormDataEntryValue } from "undici";
+
+const mimeTypeModuleType = flipObject(moduleTypeMimeType);
+
+export const checkNamespace = createNamespace({
```
emily-shen, line 27: not blocking, especially since its hidden, but it feels odd this has its own command. Do we anticipate other 'check' commands?


...maybe we should make `wrangler deploy --dry-run` a check command?? (come to think of it, maybe we should resurrect `wrangler build` for `wrangler deploy --dry-run` 🤔)
(these are only semi-serious suggestions)

emily-shen (commented):
Pages build already has an outfile, is this supposed to reuse that...? Can pages deploy fail due to startup time, and should this generate a cpuprofile if that's the case?

packages/wrangler/src/check/commands.ts
penalosa: I agree–we should definitely resurrect `wrangler build`!

In terms of the command structure here, I went with `check startup` since it seemed feasible that we might want to have other `check *` commands in future (`check types`? `check config`?), but since this is an alpha command we can always change that later.

penalosa:
> Pages build already has an outfile, is this supposed to reuse that...? 

It's meant to be doing that, yeah—does that seem buggy to you?

emily-shen:
> > Pages build already has an outfile, is this supposed to reuse that...?
> 
> It's meant to be doing that, yeah—does that seem buggy to you?

Ah no that’s fine, the lack of changes in the pages deploy path just threw me for a moment but it makes sense now. 

Will the cpuprofile be generated automatically if a pages deploy fails?

emily-shen (approved):

irvinebroque (commented):
👏🚀

PR #8031, opened 2025-02-05
Support WebAssembly modules

### LINKED ISSUE #7851: Add WebAssembly playground to Vite plugin
We should test support for WebAssembly with the Vite plugin and provide an example.

See https://vite.dev/guide/features.html#webassembly

### LINKED ISSUE #8022: 🚀 Feature Request: Support for ".wasm" and other data blobs imports for `vite-plugin-cloudflare`
### Describe the solution

It would be great if `vite-plugin-cloudflare` resolves esm `.wasm`, `.bin` imports just like wrangler does.
For example:
```typescript
// src/worker.ts
import photonWasmModule from "./photon.wasm";
import fontData from "./font.ttf.bin";
```
It should also resolve these imports if they are used in a package for example [`@cf-wasm/photon`](https://github.com/fineshopdesign/cf-wasm/blob/main/packages/photon/src/workers/index.ts).

jamesopstad:
Fixes #7851. Partially addresses #8022.

This adds support for Wasm in the Vite plugin. Imported Wasm modules are loaded via the module fallback service in dev and included as part of the bundle in build.

I've prioritised Wasm first as users are asking for it. Once this PR is merged, however, I will create a follow up PR that generalises the approach to include `CompiledWasm`, `Text` and `Data` modules defined by the user's `rules`.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Vite plugin has its own tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: `rules` documentation to be added after follow up PR


packages/vite-plugin-cloudflare/playground/prisma/package.json
```diff
@@ -0,0 +1,26 @@
+{
+	"name": "@playground/prisma",
+	"private": true,
+	"type": "module",
+	"scripts": {
+		"build": "vite build --app",
+		"check:types": "tsc --build",
+		"dev": "vite dev",
+		"generate": "pnpm prisma generate",
+		"postinstall": "pnpm generate",
```
jamesopstad, line 10: I had to add this so that the build wouldn't fail when all packages are built by TurboRepo. It might be better for us to omit the playground packages from being built automatically as the builds aren't used elsewhere.

packages/vite-plugin-cloudflare/src/modules-plugin.ts
```diff
@@ -0,0 +1,102 @@
...
+
+			return {
+				external: true,
+				id: createModuleReference("CompiledWasm", resolved.id),
+			};
+		},
+		renderChunk(code, chunk) {
+			const moduleRE = new RegExp(MODULE_PATTERN, "g");
+			let match: RegExpExecArray | null;
+			let s: MagicString | undefined;
```
dario-piotrowicz, line 53: nit: I'm usually against single letter variables, especially if they are used in more than a handful of lines of code, so here it would be nice if we could have a better name for the variable, although yeah I do think that this is a bit awkward and I can come up with a great name either 😕 🤷 

`codeUpdater`? `codeMagicString`? `content`?

it's up to you, I'm also fine with `s` 👍

dario-piotrowicz:
PS: @jamesopstad there's a merge conflict with your lock file

packages/vite-plugin-cloudflare/src/modules-plugin.ts
jamesopstad: Yeah, totally agree about single letter variables but couldn't think of a good name. I'll just call it `magicString`.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -26,29 +27,37 @@ import { getWarningForWorkersConfigs } from "./workers-configs";
...
+export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 	let resolvedPluginConfig: ResolvedPluginConfig;
 	let resolvedViteConfig: vite.ResolvedConfig;
 	let miniflare: Miniflare | undefined;
 
 	// this flag is used to show the workers configs warning only once
 	let workersConfigsWarningShown = false;
 
-	return {
+	// This is temporary pending a better solution when we refactor the plugin into multiple plugins
```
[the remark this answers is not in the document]
jamesopstad: [56e7e98](https://github.com/cloudflare/workers-sdk/pull/8031/commits/56e7e9839fdaa1743c0e73e6a7ebfc24cd3cbdf5)

I've refactored to move the modules plugin into the `index.ts` file. Makes it easier to follow and removes the need for this variable. I think as the plugin grows we should refactor parts of it into sub-plugins (`vite-plugin-cloudflare:nodejs-compat`, for example).

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -438,9 +450,40 @@ export function getDevMiniflareOptions(
 							),
 						},
 					],
+					unsafeUseModuleFallbackService: true,
 				} satisfies WorkerOptions;
 			}),
 		],
+		unsafeModuleFallbackService(request) {
+			const url = new URL(request.url);
+			const specifier = url.searchParams.get("specifier");
```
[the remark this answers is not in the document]
jamesopstad: [433e70f](https://github.com/cloudflare/workers-sdk/pull/8031/commits/433e70fe9c4a59755b35b7597b44357124df65d0)

I've switched out `specifier` for `rawSpecifier`.

packages/vite-plugin-cloudflare/playground/wasm/minimal.c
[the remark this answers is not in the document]
jamesopstad: I added a README and removed the file like you suggested.

packages/vite-plugin-cloudflare/playground/prisma/package.json
jamesopstad: Do you think it's a problem that this will run `prisma generate` whenever someone runs `pnpm i` in the repo?

jamesopstad:
> Very nice! 🤩 🚀
> 
> I've just left a few minor comments

Thanks @dario-piotrowicz! I've addressed all the feedback.

See also my comment here - https://github.com/cloudflare/workers-sdk/pull/8031#discussion_r1946292278.

packages/vite-plugin-cloudflare/playground/prisma/package.json
dario-piotrowicz: mh... yeah it's not ideal 😕 

what about using `predev` and `prebuild` scripts instead?

Something like:
```
"predev": "node ./prisma-setup.mjs",
"prebuild": "node ./prisma-setup.mjs",
```
with `prista-setup.mjs` checking if the prisma is already set up (hopefully there's a simple way to do that?) and running (via child_process) `prisma generate` only when it isn't? (so that on subsequent dev/build runs this script does nothing)

this should avoid the `pnpm i` issue with close to zero overhead when running dev/build

what do you think?

packages/vite-plugin-cloudflare/src/index.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah yeah I wasn't against the variable per-se, I was just saying that more context (exactly what you commented here) would be valuable in the code comment

I also agree with what you said 👍

packages/vite-plugin-cloudflare/playground/prisma/package.json
jamesopstad: [2c2bf54](https://github.com/cloudflare/workers-sdk/pull/8031/commits/2c2bf5449f0a645eea4e15553e022631ffb01823)

I went with just putting `prisma generate` in the `predev` and `prebuild` hooks. It's fast and ensures Prisma is always up to date with the schema.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -438,9 +450,40 @@ export function getDevMiniflareOptions(
...
+
+			try {
+				source = fs.readFileSync(modulePath);
+			} catch (error) {
+				throw new Error(`Import ${modulePath} not found. Does the file exist?`);
+			}
+
+			return MiniflareResponse.json({
+				// Cap'n Proto expects byte arrays for `:Data` typed fields from JSON
+				wasm: source instanceof Uint8Array ? Array.from(source) : source,
```
petebacondarwin, line 484: Isn't `source` always a `Buffer`? In which case it is also always a `Uint8Array`?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -438,9 +450,40 @@ export function getDevMiniflareOptions(
...
+			const url = new URL(request.url);
+			const rawSpecifier = url.searchParams.get("rawSpecifier");
+			assert(
+				rawSpecifier,
+				`Unexpected error: no specifier in request to module fallback service.`
+			);
+
+			const moduleRE = new RegExp(MODULE_PATTERN);
+			const match = moduleRE.exec(rawSpecifier);
+			assert(match, `Unexpected error: no match for module ${rawSpecifier}.`);
```
petebacondarwin, line 467: I guess that in some future where we start using the module fallback service for other things we would convert this from an asset to some kind of multi-branch if statement.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -33,289 +36,390 @@ import type { Unstable_RawConfig } from "wrangler";
...
+					id: createModuleReference("CompiledWasm", resolved.id),
+				};
+			},
+			renderChunk(code, chunk) {
+				const moduleRE = new RegExp(MODULE_PATTERN, "g");
+				let match: RegExpExecArray | null;
+				let magicString: MagicString | undefined;
+
+				while ((match = moduleRE.exec(code))) {
+					magicString ||= new MagicString(code);
```
petebacondarwin, line 372: NIT:
```suggestion
					magicString ??= new MagicString(code);
```

petebacondarwin (approved):

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: Yes. I wasn't sure if this would be the case when adding the other module types so put it here to remind me. Copied from https://github.com/cloudflare/workers-sdk/blob/0c0374cce3908a47f7459ba4810855c1ce124349/packages/vitest-pool-workers/src/pool/module-fallback.ts#L432.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: I think the `asserts` will still be valid and then the `moduleType` in the line below will be used to branch.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: [57ed257](https://github.com/cloudflare/workers-sdk/pull/8031/commits/57ed2575baaf617408c22005ca21cbca862bfc42)

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: [d4c4077](https://github.com/cloudflare/workers-sdk/pull/8031/commits/d4c4077d4107d02456af625fa674202142d1764f)

wesbos:
@jamesopstad is there any way to make this plugin agnostic to other Vite setups? Trying to import a .wasm file in a Waku project, but I can't seem to find any way to make it work. Ref: https://github.com/dai-shi/waku/issues/1245

jamesopstad:
> @jamesopstad is there any way to make this plugin agnostic to other Vite setups? Trying to import a .wasm file in a Waku project with Cloudflare, but I can't seem to find any way to make it work. Ref: [dai-shi/waku#1245](https://github.com/dai-shi/waku/issues/1245)

We would love to properly support using the Cloudflare Vite plugin with Waku in the future but it's not something that will work at the moment, I'm afraid. The Wasm support is also not something that could easily be extracted into a standalone plugin at present.

From the linked issue it looks like there might be a potential solution so hope that works out for the time being.

wesbos:
Thank you! Looks like we're getting it sorted

PR #8039, opened 2025-02-05
Remove Bun support

penalosa:
Fixes https://jira.cfdata.org/browse/DEVX-1621

Adds a fast fail error to Wrangler when run using Bun, as well as removing Bun support from C3. Bun is not supported by Wrangler or Miniflare, and so allowing users to use Bun to run C3 was not setting people up for success—we see issues come in relatively frequently about people trying to use Bun with Wrangler/Miniflare, thinking it'd work because of C3's support.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/19813

### DIFF
--- a/packages/create-cloudflare/src/cli.ts
+++ b/packages/create-cloudflare/src/cli.ts
@@ -9,6 +9,7 @@ import {
 	logRaw,
 	startSection,
 } from "@cloudflare/cli";
+import { bgYellow, yellow } from "@cloudflare/cli/colors";
 import { CancelError } from "@cloudflare/cli/error";
 import { isInteractive } from "@cloudflare/cli/interactive";
 import { cliDefinition, parseArgs } from "helpers/args";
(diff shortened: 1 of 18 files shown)


packages/create-cloudflare/src/cli.ts
```diff
@@ -41,6 +42,13 @@ import type { C3Args, C3Context } from "types";
 const { npm } = detectPackageManager();
 
 export const main = async (argv: string[]) => {
+	if (process.versions.bun || npm === "bun") {
+		console.warn(
```
Jarred-Sumner, line 46: There was a regression in Bun v1.2.0 that impacted this, but that regression was fixed in Bun v1.2.1. I manually verified it works without error in Bun v1.2.1 and Bun v1.2.2 (and that it regressed in Bun v1.2.0, which matches what you saw in the CI runs). We are adding more tests from Node.js' test suite to catch bugs like this in the future.

Is this warning still needed? We will fix anything else that comes up

Cherry: I think so. Bun’s continued disregard for ecosystem compatibility puts burden on third party maintainers, as well as the community where often “use Node.js” is the solution to most issues once Bun is discovered to be used.

In my opinion, until Bun commits to join something like WinterTC and work closely with other runtimes before implementing ecosystem breaking or diverging features, I don’t imagine this burden will change.

Cherry:
Would be good to see this merged soon. To this day we're still seeing people hit issues running wrangler with bun, where "use node" is almost always the resolution.

PR #8044, opened 2025-02-06
Add code for interpolation search for asset manifest lookup

GregBrimble:
Fixes WC-3154

Adds analytics, a bunch of tests, and introduces a zero percent gated interpolation search method as an alternative to the binary search that we're doing today. It should be a bit faster.

Zero percent for now. Let's just get the analytics in place, and then we can play with the percentage a bit and see what effect it has in production.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal package


.changeset/stupid-turtles-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
```
WillTaylorDev, line 1: Cool changeset name 😆

packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -212,7 +212,45 @@ export default class extends WorkerEntrypoint<Env> {
 	}
 
 	async unstable_exists(pathname: string): Promise<string | null> {
-		const assetsManifest = new AssetsManifest(this.env.ASSETS_MANIFEST);
-		return await assetsManifest.get(pathname);
+		const analytics = new Analytics(this.env.ANALYTICS);
+		const performance = new PerformanceTimer(this.env.UNSAFE_PERFORMANCE);
+
+		const INTERPOLATION_EXPERIMENT_SAMPLE_RATE = 0;
```
WillTaylorDev, line 218: We could also very precisely target internal accounts, might be more desirable than sampling of all accounts.

packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -212,7 +212,45 @@ export default class extends WorkerEntrypoint<Env> {
 	}
 
 	async unstable_exists(pathname: string): Promise<string | null> {
-		const assetsManifest = new AssetsManifest(this.env.ASSETS_MANIFEST);
-		return await assetsManifest.get(pathname);
+		const analytics = new Analytics(this.env.ANALYTICS);
```
WillTaylorDev, line 215: This might propose a problem, because I think we would be double counting requests in our analytics. Can't reall think of an alternative though..

packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -212,7 +212,45 @@ export default class extends WorkerEntrypoint<Env> {
...
+				coloTier: this.env.COLO_METADATA.coloTier,
+
+				coloRegion: this.env.COLO_METADATA.coloRegion,
+				version: this.env.VERSION_METADATA.id,
+			});
+		}
+
+		const startTimeMs = performance.now();
+		try {
+			const assetsManifest = new AssetsManifest(this.env.ASSETS_MANIFEST);
```
WillTaylorDev, line 245: We're setting the read method, but not the read time (not sure if that was intentional)

.changeset/stupid-turtles-hammer.md
GregBrimble: 🤪🐢🔨

packages/workers-shared/asset-worker/src/index.ts
GregBrimble: I'm kinda just wanting a random (but very low) sample. Something like 0.005%. Just enough to get some data on real life performance with actually jeopardizing any one specific site. I could also add in a header or something which always selects interpolation if you wanted to force opt in for testing? Up to you.

packages/workers-shared/asset-worker/src/index.ts
GregBrimble: We can just change our current queries to only consider records which have a hostname set (or don't have an manifest read time or something like that).

And then, yes, multiple new records will appear, one for each asset existence check. But everything else request-y will be missing from those.

packages/workers-shared/asset-worker/src/index.ts
WillTaylorDev: Nah, non blocking either way :)

WillTaylorDev (approved):

packages/workers-shared/asset-worker/src/index.ts
WalshyDev, line 215: instead of doing this and having the double counting issue - I suggest we just wrap `exists` in index/handler with the timing and set the metric. it'd clean up this code a lot and avoid analytics issues.

GregBrimble: I did consider that. Two points there though:

1. `unstable_exists` is available through the assets binding without first going through `fetch`.
2. `unstable_exists` can be called multiple times in the course of `fetch`. Last write would win. Definitely not the end of the world (and in fact, even entirely surmountable if we sent an array of results instead — we spoke about this a while ago).

Ultimately decided this was the lesser of two evils.

WalshyDev: hmm, we should at least give this it's own indexId (currently that's typed to scriptId) so it doesn't break current analytics - i do not want to go update 40 different panels haha

i think scriptId is a string but we could change that to `indexId: number | string` and do a `indexId: 'interpolation-experiment'` for this.

GregBrimble: We've gone with a second analytics namespace :)

WalshyDev (approved):

PR #8050, opened 2025-02-06
chore: pin Miniflare deps

andyjessop:
Fixes #000

This pins dependencies of Miniflare to ensure that transitive dependencies don't break Miniflare releases.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: No logic changed
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal change only

### DIFF
--- a/packages/miniflare/package.json
+++ b/packages/miniflare/package.json
@@ -43,16 +43,16 @@
 	},
 	"dependencies": {
 		"@cspotcode/source-map-support": "0.8.1",
-		"acorn": "^8.8.0",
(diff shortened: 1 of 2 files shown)


emily-shen (approved):

petebacondarwin (approved):
I wonder if we should start updating some of these deps using Dependabot?

PR #8058, opened 2025-02-07
Change type to "inherit" for all fetched bindings in versions secret put

WillTaylorDev:
This commit provides a means of inheriting all known bindings, rather than attempting to rebuild them in the versions secret put call. Some bindings (like DOs and internal bindings) don't have all fields returned. Instead, we can provide a type "inherit" for each known binding-by-name, and the API will carry over the binding from the last version.

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/versions/secrets/index.ts
+++ b/packages/wrangler/src/versions/secrets/index.ts
@@ -122,18 +122,15 @@ export async function copyWorkerVersionWithNewSecrets({
 	);
 
 	// Filter out secrets because we're gonna inherit them
-	const bindings: WorkerMetadataBinding[] =
-		versionInfo.resources.bindings.filter(
-			(binding) => binding.type !== "secret_text"
-		);
-
-	// We cannot upload a DO with a namespace_id so remove it
-	for (const binding of bindings) {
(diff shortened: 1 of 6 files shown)


petebacondarwin (approved):

packages/wrangler/src/versions/secrets/index.ts
```diff
@@ -122,18 +122,17 @@ export async function copyWorkerVersionWithNewSecrets({
...
+	// Inherit all of the existing bindings
+	bindings = bindings.map((binding) => {
+		return {
+			name: binding.name,
+			type: "inherit",
+		};
+	});
 
 	// Add the new secrets
 	for (const secret of secrets) {
```
WalshyDev, line 138: Could simplify inheritance here, we filter out, then map for inherit and then push the filtered out ones to inherit in this for loop

WalshyDev (approved):

PR #8061, opened 2025-02-07
fix: respect WRANGLER_LOG in dev

### LINKED ISSUE #7525: 🐛 BUG: wrangler dev doesn't respect WRANGLER_LOG=debug

Wrangler

n/a

_No response_

mac

### Describe the Bug

running `WRANGLER_LOG=debug npx wrangler dev` does not show the same logs as `npx wrangler dev --log-level=debug`

this is because the log-level arg has a default, which overrides the env variable.

This can be fixed by removing the default from yargs options but then that breaks tests that use runWranglerUntilConfig for some reason. 

_No response_

_No response_

emily-shen:
Fixes #7525 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/__tests__/dev.test.ts
+++ b/packages/wrangler/src/__tests__/dev.test.ts
@@ -142,7 +142,7 @@ describe.sequential("wrangler dev", () => {
 		msw.resetHandlers();
 		spy.mockClear();
 		setSpy.mockClear();
-		logger.loggerLevel = "log";
+		logger.resetLoggerLevel();
 	});
 
 	async function runWranglerUntilConfig(
--- a/packages/wrangler/src/__tests__/dev.test.ts
+++ b/packages/wrangler/src/__tests__/dev.test.ts
@@ -1774,6 +1774,17 @@ describe.sequential("wrangler dev", () => {
 				"
 			`);
 		});
+
+		it("should also set log level using WRANGLER_LOG'", async () => {
+			fs.writeFileSync("index.js", `export default {};`);
+			vi.stubEnv("WRANGLER_LOG", "none");
+			await runWranglerUntilConfig("dev index.js --inspect");
+			expect(std.warn).toMatchInlineSnapshot(`""`);
+
+			vi.stubEnv("WRANGLER_LOG", "debug");
+			await runWranglerUntilConfig("dev index.js");
+			expect(std.debug).toContain(".env file not found at");
+		});
 	});
 
 	describe("--show-interactive-dev-session", () => {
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -142,6 +142,7 @@ describe.sequential("wrangler dev", () => {
 		msw.resetHandlers();
 		spy.mockClear();
 		setSpy.mockClear();
+		logger.loggerLevel = "log";
```
edmundhung, line 145: The logger implementation seems to treat the setter as an override. So this won't be able to restore the default behavior which used `env` and fallback to `log`:

https://github.com/cloudflare/workers-sdk/blob/35c1dc32137f6440c3e514fdc54b5f471fdff7aa/packages/wrangler/src/logger.ts#L64-L66

edmundhung: Looks like you could use `logger.resetLoggerLevel()`

packages/wrangler/src/logger.ts
```diff
@@ -65,6 +65,10 @@ export class Logger {
 		this.overrideLoggerLevel = val;
 	}
 
+	resetLoggerLevel() {
+		this.overrideLoggerLevel = undefined;
+	}
+
```
edmundhung, line 71: I have no idea where I see `resetLoggerLevel` before you make this changes now... 😅

packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -1831,6 +1774,17 @@ describe.sequential("wrangler dev", () => {
...
+		it("should also set log level using WRANGLER_LOG'", async () => {
+			fs.writeFileSync("index.js", `export default {};`);
+			vi.stubEnv("WRANGLER_LOG", "none");
+			await runWranglerUntilConfig("dev index.js --inspect");
+			expect(std.warn).toMatchInlineSnapshot(`""`);
+
+			vi.stubEnv("WRANGLER_LOG", "debug");
+			await runWranglerUntilConfig("dev index.js");
+			expect(std.debug).toContain(".env file not found at");
+		});
```
edmundhung, line 1787: Nit: it would be great to test what would happen if users defined both `WRANGLER_LOG` and `--log-level`.

edmundhung (approved):

PR #8078, opened 2025-02-10
clarify e2e refers to wrangler e2e tests in PR template

### LINKED ISSUE #7744: Infra fix/improvement - the `e2e` term is too generic
### Describe the solution

In PR descriptions we have the following checkbox:
https://github.com/cloudflare/workers-sdk/blob/d102b60238c1dddfdd829ffee62c451cb526717a/.github/pull_request_template.md?plain=1#L16

we also have the `e2e` label for PRs which description is: `Run e2e tests on a PR`

I think that here the term e2e is way too generic and ambiguous since for example the C3 e2es aren't actually included in the above set of tests (they always run when there are C3 related changes).

I think the above should be made clearer, should these be called "wrangler e2e tests"?

emily-shen:
Fixes #7744 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: pr template change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: pr template change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: pr template change

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,7 +13,7 @@ The following selections do not need to be completed if this PR only contains ch
   - [ ] TODO (before merge)
   - [ ] Tests included
   - [ ] Tests not necessary because:
-- E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
+- Wrangler E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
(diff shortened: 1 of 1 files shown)


petebacondarwin (approved):
Love this!!

PR #8080, opened 2025-02-10
No longer call `next` in server middleware.

jamesopstad:
Fixes #000.

This prevents `next` from being called in the dev and preview server middleware. This is for framework compatibility so that we can override middleware provided by the framework.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included:
  - [x] Tests not necessary because: compatible with existing tests
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: only affects Vite plugin
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal feature

### DIFF
--- a/.changeset/stale-zoos-drop.md
+++ b/.changeset/stale-zoos-drop.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": minor
+---
+
+No longer call `next` in server middleware.
--- a/.changeset/stale-zoos-drop.md
+++ b/.changeset/stale-zoos-drop.md
@@ -3,3 +3,5 @@
 ---
 
 No longer call `next` in server middleware.
+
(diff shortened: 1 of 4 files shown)


.changeset/stale-zoos-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": minor
+---
+
+No longer call `next` in server middleware.
```
dario-piotrowicz, line 5: Nit: it would be great if (after an empty line) you could add a bit of details as to why we're no longer calling next

dario-piotrowicz (approved):
Looks good to me, but have you considered making this configurable?

I can imagine that there might be use cases for that?

petebacondarwin (approved):

PR #8089, opened 2025-02-11
[DO NOT MERGE] RPC POC (nonsense)

GregBrimble:

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/miniflare/src/plugins/core/index.ts
+++ b/packages/miniflare/src/plugins/core/index.ts
@@ -33,7 +33,10 @@ import {
 	CoreHeaders,
 	viewToBuffer,
 } from "../../workers";
-import { ROUTER_SERVICE_NAME } from "../assets/constants";
+import {
+	ROUTER_SERVICE_NAME,
+	RPC_PROXY_SERVICE_NAME,
+} from "../assets/constants";
 import { getCacheServiceName } from "../cache";
 import { DURABLE_OBJECTS_STORAGE_SERVICE_NAME } from "../do";
 import {
@@ -262,7 +265,7 @@ function getCustomServiceDesignator(
 	} else if (service === kCurrentWorker) {
 		// Sets SELF binding to point to router worker instead if assets are present.
 		serviceName = hasAssetsAndIsVitest
-			? `${ROUTER_SERVICE_NAME}:${refererName}`
+			? `${RPC_PROXY_SERVICE_NAME}:${refererName}`
 			: getUserServiceName(refererName);
 	} else {
 		// Regular user worker
--- a/packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
+++ b/packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
@@ -1,0 +1,33 @@
+import { WorkerEntrypoint } from "cloudflare:workers";
+
+export default class SpecialDefault extends WorkerEntrypoint {
+	fetch(request) {
+		return this.env.ROUTER_WORKER.fetch(request);
+	}
+
+	constructor(...args) {
+		return new Proxy(super(...args), {
+			get(target, prop, receiver) {
+				// totally shit
+				if (prop === "fetch") {
+					const originalFetch = target[prop];
+					return function (...args) {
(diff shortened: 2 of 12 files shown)


packages/miniflare/src/plugins/core/index.ts
```diff
@@ -262,7 +265,7 @@ function getCustomServiceDesignator(
 	} else if (service === kCurrentWorker) {
 		// Sets SELF binding to point to router worker instead if assets are present.
 		serviceName = hasAssetsAndIsVitest
-			? `${ROUTER_SERVICE_NAME}:${refererName}`
+			? `${RPC_PROXY_SERVICE_NAME}:${refererName}`
```
CarmenPopoviciu, line 268: I am SO confused why this solves anything 🙃

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export default class SpecialDefault extends WorkerEntrypoint {
+	fetch(request) {
+		return this.env.ROUTER_WORKER.fetch(request);
+	}
+
+	constructor(...args) {
+		return new Proxy(super(...args), {
+			get(target, prop, receiver) {
+				// totally shit
```
kentonv, line 11: Is this expressing some sort of anger at the whole idea of using a `Proxy` here or just expressing that the current implementation needs work?

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -0,0 +1,33 @@
...
+
+				if (prop === "env") {
+					return Reflect.get(target, prop, receiver);
+				}
+
+				if (target[prop] !== undefined) {
+					return target[prop];
+				}
+
+				return function (...args) {
```
kentonv, line 27: Could this just `return target.env.USER_WORKER[prop]`? Why is it necessary to wrap in a pass-through function?

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -0,0 +1,33 @@
...
+					return receiver.env.USER_WORKER[prop](...args);
```
kentonv, line 28: Shouldn't this be `target.env` instead of `receiver.env`? `receiver` is the proxy itself.

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -0,0 +1,33 @@
...
+			get(target, prop, receiver) {
+				// totally shit
+				if (prop === "fetch") {
+					const originalFetch = target[prop];
+					return function (...args) {
+						return originalFetch.apply(target, args);
+					};
+				}
+
+				if (prop === "env") {
```
kentonv, line 19: This seems like it exposes `env` over RPC, which doesn't seem like what you want?

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -0,0 +1,33 @@
+import { WorkerEntrypoint } from "cloudflare:workers";
+
+export default class SpecialDefault extends WorkerEntrypoint {
+	fetch(request) {
+		return this.env.ROUTER_WORKER.fetch(request);
+	}
+
+	constructor(...args) {
+		return new Proxy(super(...args), {
```
kentonv, line 9: Wow did not know you could return from a constructor. That's a neat trick that certainly makes proxying cleaner.

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
GregBrimble: The latter :) I just threw this together and knew there were going to be issues with this implementation.

PR #8091, opened 2025-02-11
Omit files from public directory in Worker builds

jamesopstad:
Fixes #[000].

Omit files from public directory in Worker builds

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: N/A
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
+++ b/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
@@ -148,6 +148,7 @@ export function createCloudflareEnvironmentOptions(
 			// We need to enable `emitAssets` in order to support additional modules defined by `rules`
 			emitAssets: true,
 			outDir: getOutputDirectory(userConfig, environmentName),
+			copyPublicDir: false,
 			ssr: true,
 			rollupOptions: {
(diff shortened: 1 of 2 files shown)


packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -148,6 +148,7 @@ export function createCloudflareEnvironmentOptions(
 			// We need to enable `emitAssets` in order to support additional modules defined by `rules`
 			emitAssets: true,
 			outDir: getOutputDirectory(userConfig, environmentName),
+			copyPublicDir: false,
```
petebacondarwin, line 151: This is not wanted when we are assets-only, right?
But then `createCloudflareEnvironmentOptions()` is not called in that case, right?

petebacondarwin (approved):

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
jamesopstad: Yes, there is no Worker environment when it's assets only, just the `client` environment.

PR #8094, opened 2025-02-11
Upload sourcemaps to sentry for workers-shared workers

WillTaylorDev:
This PR generates sourcemaps, creates sentry release, and uploads sourcemaps, for all production releases of router-worker and asset-worker.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


WalshyDev (approved):

packages/workers-shared/scripts/get-version-tag.ts
```diff
@@ -0,0 +1,10 @@
+// Generates a version tag which can be used with version upload and associated with sentry releases
+
+import { execSync } from "child_process";
+
+try {
+	const hash = execSync("git rev-parse --short=10 HEAD").toString().trim();
+	console.log(hash);
+} catch (error) {
+	console.log("UNKNOWN");
```
CarmenPopoviciu, line 9: are we sure we don't want to log the error here?

WillTaylorDev: We're capturing the output as the version tag; we definitely don't want to log the error. Providing UNKNOWN is probably the best we can do. This wouldn't ever be expected though, since we're always going to be running this in a git repo :)

PR #8104, opened 2025-02-11
[vite-plugin] use the new `resolve.builtins` vite config to specify the cloudflare builtins

dario-piotrowicz:
Closes https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/52

Note: in https://github.com/flarelabs-net/vite-plugin-cloudflare/pull/52 I've done some extra work to also conditionally include/exclude the nodejs externals based on the nodejs_compat flag, I am happy to do that here as well but I figured that maybe it'd be cleaner to do one thing at a time, and first simply use the new `resolve.builtins` option before optimizing further

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: the module resolution is already tested
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: unrelated to wrangler
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal change, no user facing changes


petebacondarwin:
Rebased to see if this still works with the changes I made to nodejs compat

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -25,6 +26,10 @@ import type {
 import type { MiniflareOptions, SharedOptions, WorkerOptions } from "miniflare";
 import type { FetchFunctionOptions } from "vite/module-runner";
 
+const nodeBuiltInModules = new Set(
+	builtinModules.concat(builtinModules.map((m) => `node:${m}`))
+);
+
```
jamesopstad, line 32: Can we export this from `cloudflare-environment.ts` as it's basically the same (and doesn't necessarily need to be a `Set`)?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -328,22 +333,16 @@ export function getDevMiniflareOptions(
...
-									if (moduleRE.test(moduleId)) {
-										const result = {
-											externalize: moduleId,
-											type: "module",
-										} satisfies vite.FetchResult;
+									const shouldExternalize =
+										// Worker modules (CompiledWasm, Text, Data)
+										moduleRE.test(moduleId) ||
+										// Node.js builtin node modules (they will be resolved to unenv aliases)
+										nodeBuiltInModules.has(moduleId);
```
jamesopstad, line 340: Know we've discussed it before but it still doesn't really make sense to me that we have to externalize things again here. Agree this is correct for now though.

jamesopstad (approved):

packages/vite-plugin-cloudflare/src/miniflare-options.ts
dario-piotrowicz: exporting it from `cloudflare-environment.ts` feels incorrect to me, I can move it to a third file imported by both if that's ok with you?

PS: it's very minimal but I do like it being a set here since it will speed up (very trivially) the lookup below

packages/vite-plugin-cloudflare/src/miniflare-options.ts
dario-piotrowicz: I 100% totally agree 👍

I'll go and investigate this 👍  (but yeah for this PR I would keep it as is since this is also effects the other modules)

packages/vite-plugin-cloudflare/src/miniflare-options.ts
dario-piotrowicz: I've moved this to the `utils.ts` file: https://github.com/cloudflare/workers-sdk/pull/8104/commits/3ce2e36db36cd86d339919ef3602f9b17f72ea3b

(and kept is a `Set` 🙈)

I hope this works for you? 🙂

Issue #8101, opened 2025-02-11
🐛 BUG: `wrangler secret list --format json` emits non-JSON characters


benvinegar:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What versions are you using?

3.108.0

### What operating system and version are you using?

Mac Sequoia 15.1.1

### Describe the Bug

As of 3.108, wrangler is emitting non-JSON characters when specifying `--format json`.

Before (3.107):

```
npx wrangler@3.107 secret list --format json --name WORKER_NAME
[
  {
    "name": "secret_A",
    "type": "secret_text"
  },
  {
    "name": "secret_B",
    "type": "secret_text"
  }
]
```

Current version (3.108):

```
npx wrangler@3.108 secret list --format json --name WORKER_NAME

 ⛅️ wrangler 3.108.0
--------------------

[
  {
    "name": "secret_A",
    "type": "secret_text"
  },
  {
    "name": "secret_B",
    "type": "secret_text"
  }
]
```

This is causing my code to fail because I capture this output and assume it is JSON.

lrapoport-cf:
@benvinegar  thanks for reporting 👍 we have deprecated 3.108.0 and are working on a fix now

Issue #7487, opened 2024-12-06
🐛 BUG: wrangler secret list not valid json 3.93.0 - resolved by downgrading to 3.92.0


jdanyow:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What version(s) of the tool(s) are you using?

3.93.0

### What version of Node are you using?

v23.3.0

### What operating system and version are you using?

gh actions ubuntu-latest

### Describe the Bug

I updated to wranger 3.93.0 today. Afterwards, executing wrangler secret lists produces the following output/error:

```
[ { "name": "XYZ", "type": "secret_text" } ]Cloudflare collects anonymous telemetry about your usage of Wrangler. Learn more at htt
```

If I downgrade to 3.92.0, wrangler secrets list works as expected

This is an issue for our deployments because the [secrets JSON is parsed by jq as part of an assertion that all expected secrets are configured](https://github.com/cloudflare/workers-sdk/issues/2101#issuecomment-1813043234). I'd love to be able to assert secrets in my wrangler.toml. Any chance #2101 can be re-opened?

penalosa:
Thanks for opening an issue! Would it be possible to put together a repro of this issue? It would also be really helpful to see Wrangler's debug logs (which can be enabled by setting WRANGLER_LOG=debug) to help us investigate further.

SayefReyadh:
Is there any update on this issue? Some of our pipelines are currently broken because of it. We use the list command to fetch the recent version and retrieve the ID based on our custom tag. However, the new line ⛅️ wrangler 3.93.0 at the top of the output is causing our jq commands to fail with the error: jq: parse error: Invalid numeric literal at line 2, column 8.

SayefReyadh:
> @SayefReyadh is correct - this was an issue with `wrangler secret list`, not deploy. I've updated the title and description to reflect this. Need to remove the "Cloudflare collects anonymous telemetry about your usage of Wrangler" notice.

For a work around we downgraded to 3.92.0 for now.

jdanyow:
I just verified you can add the following to your wranger.toml to remove the telemetry notice:

```
send_metrics = false
```

https://github.com/cloudflare/workers-sdk/blob/669d7ad1e44c07cf74202c4d0fc244a9c50dec81/packages/wrangler/src/metrics/metrics-dispatcher.ts#L190

lrapoport-cf:
we are working on a fix now 👍

lrapoport-cf:
[...]

PR #8109, opened 2025-02-12
chore: include banner in output of tests for commands with json/pretty mode

emily-shen:
The wrangler banner is normally [mocked](https://github.com/cloudflare/workers-sdk/blob/82a89371250a1ec3e56fa5a518c685d362bde0ad/packages/wrangler/src/__tests__/vitest.setup.ts#L86) in tests so it isn't included in the console output that we check. However, we do want to test for the _absence_ of the banner when a particular command has a 'json' mode where the output should be json only.

This PR adds/updates tests for d1 and tail commands that have a 'json' option.
There may be some commands that have an _implicit_ guarantee that they output json without a separate flag, but I don't have a good way of identifying those - `secret list` now has tests for it but there may be others?

This should ensure issues like #7487 and #8101 don't happen again 😅 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: change to tests
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: only change unit tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: test change

### DIFF
--- a/packages/wrangler/src/__tests__/d1/execute.test.ts
+++ b/packages/wrangler/src/__tests__/d1/execute.test.ts
@@ -6,6 +6,10 @@ import { runInTempDir } from "../helpers/run-in-tmp";
 import { runWrangler } from "../helpers/run-wrangler";
 import { writeWranglerConfig } from "../helpers/write-wrangler-config";
 
+// we want to include the banner to make sure it doesn't show up in the output
+// when --json=true
+vi.unmock("../../wrangler-banner");
+
 describe("execute", () => {
 	const std = mockConsoleMethods();
 	runInTempDir();
@@ -136,6 +140,29 @@ describe("execute", () => {
 				2
 			)
 		);
+	});
+
+	it("should show banner by default", async () => {
+		setIsTTY(false);
+		writeWranglerConfig({
+			d1_databases: [
+				{ binding: "DATABASE", database_name: "db", database_id: "xxxx" },
+			],
+		});
+
+		await runWrangler("d1 execute db --command 'select 1;'");
+		expect(std.out).toContain("⛅️ wrangler x.x.x");
+	});
+	it("should not show banner if --json=true", async () => {
+		setIsTTY(false);
+		writeWranglerConfig({
+			d1_databases: [
+				{ binding: "DATABASE", database_name: "db", database_id: "xxxx" },
+			],
+		});
+
+		await runWrangler("d1 execute db --command 'select 1;' --json");
+		expect(std.out).not.toContain("⛅️ wrangler x.x.x");
 	});
 });
 
--- a/packages/wrangler/src/__tests__/d1/info.test.ts
+++ b/packages/wrangler/src/__tests__/d1/info.test.ts
@@ -8,6 +8,9 @@ import { runInTempDir } from "../helpers/run-in-tmp";
 import { runWrangler } from "../helpers/run-wrangler";
 import { writeWranglerConfig } from "../helpers/write-wrangler-config";
 
+// we want to include the banner to make sure it doesn't show up in the output when
+// when --json=true
+vi.unmock("../../wrangler-banner");
 describe("info", () => {
 	mockAccountId({ accountId: null });
 	mockApiToken();
@@ -16,7 +19,7 @@ describe("info", () => {
 	const std = mockConsoleMethods();
 	const { setIsTTY } = useMockIsTTY();
 
-	it("should display version when alpha", async () => {
+	beforeEach(() => {
 		setIsTTY(false);
 		mockGetMemberships([
 			{ id: "IG-88", account: { id: "1701", name: "enterprise" } },
@@ -30,6 +33,8 @@ describe("info", () => {
 				},
 			],
 		});
+	});
+	it("should display version when alpha", async () => {
 		msw.use(
 			http.get("*/accounts/:accountId/d1/database/*", async () => {
 				return HttpResponse.json(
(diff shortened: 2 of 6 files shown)


packages/wrangler/src/__tests__/pages/pages-deployment-tail.test.ts
```diff
@@ -21,6 +21,9 @@ import type {
 import type { RequestInit } from "undici";
 import type WebSocket from "ws";
 
+// we want to include the banner to make sure it doesn't show up in
```
vicb, line 24: Do you mean:

```suggestion
// we want to include the banner to make sure it shows up in
```

vicb: Also I have no clue about what "json mode" is?

emily-shen: i've update the comments and PR description, but tldr we want to test for the absence of the banner, so we need to enable banner printing in the first place

Skye-31:
Out of curiosity: instead of each command having to think about `--json` and conditionally add the banner, could `printBanner()` itself do that check?

emily-shen:
> Out of curiosity: instead of each command having to think about `--json` and conditionally add the banner, could `printBanner()` itself do that check?

yeah? i don't know if it would be huge benefit at the moment as the switches are so inconsistent: sometimes we use `--json` and sometimes we use `--format`, and sometimes we just print json and hope we remember to not mess it up in the future 🙃  
I think there's some standardising in v4, but there's still two default behaviours that vary command by command.

penalosa (approved):

packages/wrangler/src/__tests__/d1/execute.test.ts
```diff
@@ -6,6 +6,10 @@ import { runInTempDir } from "../helpers/run-in-tmp";
 import { runWrangler } from "../helpers/run-wrangler";
 import { writeWranglerConfig } from "../helpers/write-wrangler-config";
 
+// we want to include the banner to make sure it doesn't show up in the output
+// when --json=true
+vi.unmock("../../wrangler-banner");
```
CarmenPopoviciu, line 11: for posterity, this is mocked [here](https://github.com/cloudflare/workers-sdk/blob/82a89371250a1ec3e56fa5a518c685d362bde0ad/packages/wrangler/src/__tests__/vitest.setup.ts#L86)

CarmenPopoviciu: updated PR description to reflect this detail

packages/wrangler/src/__tests__/d1/info.test.ts
```diff
@@ -129,4 +121,70 @@ describe("info", () => {
 		}"
 	`);
 	});
+
+	it("should print banner in pretty mode", async () => {
```
CarmenPopoviciu, line 125: nit:

```suggestion
	it("should pretty print by default, incl. the wrangler banner", async () => {
```

packages/wrangler/src/__tests__/d1/list.test.ts
```diff
@@ -0,0 +1,85 @@
...
+			  {
+			    \\"uuid\\": \\"2\\",
+			    \\"name\\": \\"b\\",
+			    \\"binding\\": \\"B\\"
+			  }
+			]"
+		`);
+	});
+
+	it("should pretty print by default", async () => {
```
CarmenPopoviciu, line 69: not sure if it is relevant to signal that we're also explicitly testing for the presence of the banner? Or is that specifically relevant only for json print mode?

```suggestion
	it("should pretty print by default, including the wrangler banner", async () => {
```

packages/wrangler/src/__tests__/d1/timeTravel.test.ts
```diff
@@ -110,4 +114,129 @@ describe("time-travel", () => {
...
+					`d1 time-travel restore db --timestamp=2011-09-05T14:48:00.000Z --json`
+				);
+				expect(std.out).toMatchInlineSnapshot(`
+					"{
+					  \\"bookmark\\": \\"a\\"
+					}"
+				`);
+			});
+
+			it("should pretty print by default", async () => {
```
CarmenPopoviciu, line 184: not sure if it is relevant to signal that we're also explicitly testing for the presence of the banner? Or is that specifically relevant only for json print mode?

```suggestion
			it("should pretty print by default, including the wrangler banner", async () => {
```

packages/wrangler/src/__tests__/d1/timeTravel.test.ts
```diff
@@ -110,4 +114,129 @@ describe("time-travel", () => {
...
+				)
+			);
+			vi.useFakeTimers();
+			vi.setSystemTime(new Date("2011-10-05T14:48:00.000Z"));
+		});
+		afterEach(() => {
+			vi.useRealTimers();
+		});
+		describe("restore", () => {
+			it("should print as json if specified, without wrangler banner", async () => {
```
CarmenPopoviciu, line 173: since we're already in a `--json` describe block...

```suggestion
			it("should print as json, without wrangler banner", async () => {
```

packages/wrangler/src/__tests__/d1/timeTravel.test.ts
```diff
@@ -110,4 +114,129 @@ describe("time-travel", () => {
...
+					🤖 Using fallback value in non-interactive context: yes
+					⚡️ Time travel in progress...
+					✅ Database db restored back to bookmark a
+
+					↩️ To undo this operation, you can restore to the previous bookmark: undefined"
+				`);
+			});
+		});
+		describe("info", () => {
+			it("should print as json if specified, without wrangler banner", async () => {
```
CarmenPopoviciu, line 208: since we're already in a `--json` describe block...

```suggestion
			it("should print as json, without wrangler banner", async () => {
```

CarmenPopoviciu (approved):
Left some small nit but LGTM overall. Thank you for adding those extra tests <3 🚀

PR #8116, opened 2025-02-12
Add RewriteFrames to correct issue with source mappings

WillTaylorDev:
Source maps were not correctly mapping.
See: https://docs.docbase.ai/platforms/node/guides/express/configuration/integrations/pluggable-integrations/#rewriteframes

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/eighty-panthers-rule.md
+++ b/.changeset/eighty-panthers-rule.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-shared": patch
+---
+
+Adds a RewriteFrames integration for workers-shared Sentry source-mappings.
--- a/packages/workers-shared/utils/sentry.ts
+++ b/packages/workers-shared/utils/sentry.ts
@@ -1,4 +1,4 @@
-import { Toucan } from "toucan-js";
+import { RewriteFrames, Toucan } from "toucan-js";
 import type { ColoMetadata } from "./types";
 
 export function setupSentry(
@@ -22,6 +22,14 @@ export function setupSentry(
 		context,
 		sampleRate: 1.0,
 		release: versionMetadata?.tag,
+		integrations: [
+			new RewriteFrames({
+				iteratee(frame) {
+					frame.filename = "/index.js";
+					return frame;
+				},
+			}),
+		],
 		requestDataOptions: {
 			allowedHeaders: [
 				"user-agent",


WalshyDev (approved):

.changeset/eighty-panthers-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/workers-shared": minor
+---
+
+Adds a RewriteFrames for workers-shared sentry sourcemappings.
```
petebacondarwin, line 5:
```suggestion
Adds a RewriteFrames integration for workers-shared Sentry source-mappings.
```

.changeset/eighty-panthers-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/workers-shared": minor
```
petebacondarwin, line 2: Is this really a bug fix?
If a minor, what is the new feature that the user is getting?

packages/workers-shared/utils/sentry.ts
```diff
@@ -22,6 +23,14 @@ export function setupSentry(
 		context,
 		sampleRate: 1.0,
 		release: versionMetadata?.tag,
+		integrations: [
+			new RewriteFrames({
+				iteratee: (frame: StackFrame): StackFrame => {
+					frame.filename = "/index.js";
+					return frame;
```
petebacondarwin, line 30: It is OK (or expected) to mutate the frame rather than making a copy with changes?

.changeset/eighty-panthers-rule.md
penalosa, line 2: This is a bugfix, right? It should be a patch

penalosa (approved):

.changeset/eighty-panthers-rule.md
petebacondarwin: Really the `RewriteFrames` bit is an implementation detail.
What you are really doing here is fixing the filename on sourcemap frames sent to Sentry.

petebacondarwin (changes requested):

packages/workers-shared/utils/sentry.ts
WillTaylorDev: Yup, this is fine!

.changeset/eighty-panthers-rule.md
WillTaylorDev: Changed to patch :)

packages/workers-shared/package.json
```diff
@@ -61,6 +61,7 @@
 		"@cloudflare/workers-tsconfig": "workspace:*",
 		"@cloudflare/workers-types": "^4.20250204.0",
 		"@sentry/cli": "^2.37.0",
+		"@sentry/types": "^7.51.0",
```
petebacondarwin, line 64: No need for these types. They are already inferrable from the RewriteFrames type.

packages/workers-shared/utils/sentry.ts
```diff
@@ -1,5 +1,6 @@
-import { Toucan } from "toucan-js";
+import { RewriteFrames, Toucan } from "toucan-js";
 import type { ColoMetadata } from "./types";
+import type { StackFrame } from "@sentry/types";
```
petebacondarwin, line 3: Not needed

petebacondarwin (approved):
Thanks for the quick turnaround and bearing with my pedantry.

PR #8118, opened 2025-02-12
fix node.js compat module resolution

### LINKED ISSUE #8108: 🐛 BUG: @cloudflare/vite-plugin `vite dev` errors when `nodejs_compat` is enabled

Other

@cloudflare/vite-plugin 0.0.8, Vite 6.1.0 and 6.0.11, Wrangler 3.108.1, Node.js v22.13.1

Mac Sequoia 15.3.1 and Ubuntu 24.10

https://github.com/patsaas/cloudflare-vite-plugin-repro

### Describe the Bug

The Cloudflare Vite plugin `0.0.8` should've had the `nodejs_compat` issue (#8054) fixed in #7830, but there's a different error now.

[Repro](https://github.com/patsaas/cloudflare-vite-plugin-repro) > `npm install && npm run dev` > visit `/api/` or click the "Name from API is" button on the homepage to reproduce the error.

Error on request to the Worker at `/api/`:

```
node:internal/fs/promises:638
  return new FileHandle(await PromisePrototypeThen(
                        ^

Error: ENOENT: no such file or directory, open '/Users/pat/cloudflare-vite-tutorial/node_modules/unenv/runtime/node/tty/index.mjs?v=79e7f969'
    at async open (node:internal/fs/promises:638:25)
    at async Object.readFile (node:internal/fs/promises:1242:14)
    at async extractExportsData (file:///Users/pat/cloudflare-vite-tutorial/node_modules/vite/dist/node/chunks/dep-CfG9u7Cn.js:19285:24) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/pat/cloudflare-vite-tutorial/node_modules/unenv/runtime/node/tty/index.mjs?v=79e7f969'
}
```

This occurs with both Vite 6.0.11 and 6.1.0.

_No response_

petebacondarwin:
Fixes #8108

In v0.0.8 we landed support for Vite 6.1 and also switched to using the new Cloudflare owned unenv preset.
Unfortunately, the changes made in that update caused a regression in Node.js support.
This became apparent only when the plugin was being used with certain package managers and outside of the workers-sdk monorepo.

The unenv polyfills that get compiled into the Worker are transitive dependencies of this plugin, not direct dependencies of the user's application were the plugin is being used.
This is on purpose to avoid the user having to install these dependencies themselves.

Unfortunately, the changes in 0.0.8 did not correctly resolve the polyfills from `@cloudflare/unenv-preset` and `unenv` when the dependencies were not also installed directly into the user's application.

The approach was incorrectly relying upon setting the `importer` in calls to Vite's `resolve(id, importer)` method to base the resolution in the context of the vite plugin package rather than the user's application.
This doesn't work because the `importer` is only relevant when the `id` is relative, and not a bare module specifier in the case of the unenv polyfills.

This change fixes how these id are resolved in the plugin by manually resolving the path at the appropriate point, while still leveraging Vite's resolution pipeline to handle aliasing, and dependency optimization.

This change now introduces e2e tests that checks that isolated installations of the plugin works with npm, pnpm and yarn.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Not Wrangler
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -416,35 +416,34 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+							unresolved,
+							resolved
+						).id;
+					// We must pass the id from the depsOptimizer through the rest of the
+					// resolving pipeline to ensure that the optimized version gets used.
+					return this.resolve(optimized);
 				}
 
-				return resolvedAlias;
+				return resolved;
```
petebacondarwin, line 446: Not sure it we should pass this through the `this.resolve()` call too, but since it is going to be an absolute path at this stage, I assume that no other plugin would do anything with it anyway?

petebacondarwin:
Agh! Testing with the prerelease seems broken again 😢

.changeset/wild-days-jog.md
```diff
@@ -0,0 +1,11 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+fix node.js compat module resolution
+
+The polyfills that get compiled into the Worker were not being found
+because they were dependencies of the plugin, not the user's project.
```
jamesopstad, line 8: This has always been the case so I don't think it explains why the issue occurred and under what conditions (package manager etc.).

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -104,18 +105,26 @@ export function maybeStripNodeJsVirtualPrefix(
...
+			!env.external.includes(alias),
+			`Unexpected unenv alias to external module: ${source} -> ${alias}`
+		);
+		source = alias;
+	}
+	// Resolve to an absolute path using the path to this file as the resolution starting point.
+	// This is essential so that we can find the `@cloudflare/unenv-preset` and `unenv` packages,
+	// which are dependencies of this package but may not be direct dependencies of the user's project.
+	const resolved = resolvePathSync(source, {
+		url: import.meta.url,
```
vicb, line 123: Why not using the [unenv resolution](https://github.com/unjs/unenv/blob/008bf1c19a716faa5f2cf86d58d8531447d0e230/lib/index.d.mts#L51)?

jamesopstad:
Please could you add a description to the PR explaining what was broken and how it's been fixed. It will be useful to look back on and will make it easier to review.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -104,18 +105,26 @@ export function maybeStripNodeJsVirtualPrefix(
...
-	assert(
-		alias,
-		`Expected "${source}" to have a Node.js compat alias, but none was found`
-	);
-	assert(
-		!env.external.includes(alias),
-		`Unexpected unenv alias to external module: ${source} -> ${alias}`
-	);
-	return alias;
+	if (alias) {
```
jamesopstad, line 112: So, if we can't find the `alias`, we use the `source` as is. In what scenario would this happen?

petebacondarwin: Primarily, the case where the id was on that was injected into the code (to support globals).

jamesopstad: Don't quite follow. Aren't the injected IDs (in `injectGlobalCode`) always prefixed with `CLOUDFLARE_VIRTUAL_PREFIX`? In that case if they aren't an `alias` then they would be an `external` and the assertion would fail.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -51,12 +52,12 @@ export function injectGlobalCode(id: string, code: string) {
 			if (typeof globalInject === "string") {
 				const moduleSpecifier = globalInject;
 				// the mapping is a simple string, indicating a default export, so the string is just the module specifier.
-				return `import var_${globalName} from "${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName};\n`;
+				return `import var_${globalName} from "${CLOUDFLARE_VIRTUAL_PREFIX}${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName};\n`;
```
jamesopstad, line 55: Is it safe to always prefix with `CLOUDFLARE_VIRTUAL_PREFIX`? What if the `moduleSpecifier` is external?

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: Yeah good question.

I realise the design of the unenv resolution stuff is supposed to aid combining different presets in different packages with the unenv library itself but in practice I have found that it adds extra complexity to the code and is hidden logic that is not always easy to reason about.

While trying to get this stuff working I switched to controlling the resolution directly because it was easier to be sure how it was working.

As far as I can tell doesn't appear to be needed for our purposes. It maybe that there are other use cases where it is valuable but not here.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: Our resolve hook will strip this prefix and reapply normal resolution if it is not an unenv alias, so things should just work as expected.

jamesopstad: But, as I understand it, externals are aliases in unenv.

Just trying to square it with the logic here for defining when to apply the prefix:

https://github.com/cloudflare/workers-sdk/blob/8b49bebc4d60cf9b3d9e6804c1f8da5cb51596a8/packages/vite-plugin-cloudflare/src/node-js-compat.ts#L77-L86

It feels like they should be consistent. Wouldn't this also resolve my confusion about why `source` may not be an `alias` here - https://github.com/cloudflare/workers-sdk/pull/8118#discussion_r1954401903?

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: The injected id are of the form: `@cloudflare/unenv-preset/runtime/node/console/index` and similar.
These are not aliased in the preset's alias, they are just files that need to be injected into the code.
So the transformed code will look like:

```ts
import var_console from "\0__CLOUDFLARE_NODEJS_COMPAT__@cloudflare/unenv-preset/runtime/node/console/index";
globalThis.console = var_console;
```

jamesopstad: Ah, OK. This is making more sense to me now. Thanks for explaining.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
jamesopstad: You can ignore my comment above. Makes sense now.

petebacondarwin: Let me double check this aspect. It was certainly the case before that if the alias key is an actual external (i.e. it was already builtin to workerd) then this breaks how Vite deals with it.

There is this bit of code in the unenv preset:

https://github.com/cloudflare/workers-sdk/blob/8b49bebc4d60cf9b3d9e6804c1f8da5cb51596a8/packages/unenv-preset/src/preset.ts#L58-L64

which basically means that the aliases contain a lot of:

```
{
  "path": "path",
  "node:path": "node:path",
  ...
}
```

where `path` is already built-in to workerd.

It might be that we could remove these from the preset. I wonder if they were there just to ensure that we don't inherit polyfill aliases for these modules from the "nodeless" preset. But now I don't believe we rely upon any other preset. (Perhaps @vicb can shed some light here?)

packages/vite-plugin-cloudflare/src/node-js-compat.ts
jamesopstad:
```suggestion
	// If `alias` is `undefined` then `source` was injected in the `transform` hook and we can resolve it directly.
	// Else we assign `alias` to `source` before resolving
	if (alias) {
```

jamesopstad (approved):

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: I passed it through `this.resolve()` and the tests pass.

.changeset/wild-days-jog.md
petebacondarwin: Updated - hopefully gives enough context.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: Added the comments

petebacondarwin:
These new tests do actually run and are passing: https://github.com/cloudflare/workers-sdk/actions/runs/13316922133/job/37193117087#step:5:2148

PR #8123, opened 2025-02-12
fix: nonretryable error and errored step caching

LuisDuarte1:
Fixes WOR-468

NonretryableError and errored step caching didn't previously match prod behavior.
 
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: already documented behavior


packages/workflows-shared/package.json
```diff
@@ -40,13 +40,14 @@
...
 		"@cloudflare/eslint-config-worker": "workspace:*",
+		"@cloudflare/vitest-pool-workers": "^0.6.16",
 		"@cloudflare/workers-tsconfig": "workspace:*",
 		"@cloudflare/workers-types": "^4.20250204.0",
 		"@types/mime": "^3.0.4",
 		"esbuild": "0.17.19",
 		"rimraf": "catalog:default",
 		"typescript": "catalog:default",
-		"vitest": "catalog:default"
+		"vitest": "^2.1.0"
```
penalosa, line 50: This should ideally be using a workspace dependency and Vitest 3—was there a specific reason that didn't work?

CarmenPopoviciu:
let's pls get a Workflows approval first on this PR, as it mostly touches `workflows-shared`

sidharthachatterjee (commented):
Approved

packages/workflows-shared/package.json
LuisDuarte1: I've updated vitest-pool-workers to 0.7, removing the need for vitest 2.0. But I couldn't use the workspace dependency for it because of cyclic dependencies still :sweat_smile:

bruxodasilva (approved):

.changeset/odd-numbers-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/workflows-shared": minor
```
petebacondarwin, line 2: As a package that has a 0 major version, non-breaking changes (even features) should be patch bumps.

```suggestion
"@cloudflare/workflows-shared": patch
```

petebacondarwin (approved):
From the point of view out the workers-sdk team this LGTM - I am leaning on the Workflows team to decide if the logic of this worker is correct.

.changeset/odd-numbers-cheat.md
LuisDuarte1: Yhep, that that's right thanks :)

PR #8130, opened 2025-02-13
feat: allowlist some types defaults

emily-shen:
Include default values for wrangler types --path and --x-include-runtime in telemetry

User provided strings are still left redacted as always.

Note i've also tested this manually for wrangler types

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: metrics
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: expanding documented behaviour

### DIFF
--- a/packages/wrangler/src/metrics/metrics-dispatcher.ts
+++ b/packages/wrangler/src/metrics/metrics-dispatcher.ts
@@ -32,12 +32,15 @@ export function getMetricsDispatcher(options: MetricsConfigOptions) {
 	let amplitude_event_id = 0;
 
 	/** We redact strings in arg values, unless they are named here */
-	const allowList = {
+	const allowList: Record<string, AllowedValues> & { "*": AllowedValues } = {
 		// applies to all commands
 		// use camelCase version
-		"*": ["format", "logLevel"],
-		// specific commands
-		tail: ["status"],
+		"*": { format: "*", logLevel: "*" },
+		"wrangler tail": { status: "*" },
+		"wrangler types": {
+			xIncludeRuntime: [".wrangler/types/runtime.d.ts"],
+			path: ["worker-configuration.d.ts"],
+		},
 	};
 
 	return {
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/metrics/metrics-dispatcher.ts
```diff
@@ -253,35 +256,47 @@ const sanitiseUserInput = (
...
-	allowedKeys: string[]
+	allowedValues: AllowedValues
 ) => {
 	const result: Record<string, unknown> = {};
 
-	for (const [k, value] of Object.entries(args)) {
-		const key = normalise(k);
+	for (let [key, value] of Object.entries(args)) {
+		key = normalise(key);
+		// the default is not set by yargs :/
```
penalosa, line 275: Why not?

packages/wrangler/src/metrics/metrics-dispatcher.ts
```diff
@@ -253,35 +256,47 @@ const sanitiseUserInput = (
...
+		}
+		const allowedValuesForArg = allowedValues[key] ?? [];
 		if (exclude.has(key)) {
 			continue;
 		}
 		if (
 			typeof value === "number" ||
 			typeof value === "boolean" ||
-			allowedKeys.includes(normalise(key))
+			allowedValuesForArg.includes(normalise(key))
```
penalosa, line 286: Isn't key already normalised here?

packages/wrangler/src/metrics/metrics-dispatcher.ts
```diff
@@ -253,35 +256,47 @@ const sanitiseUserInput = (
...
+			!(allowedValuesForArg === "*" || allowedValuesForArg.includes(value))
+		) {
 			result[key] = "<REDACTED>";
 		} else if (Array.isArray(value)) {
 			result[key] = value.map((v) =>
-				typeof v === "string" ? "<REDACTED>" : v
+				typeof v === "string" &&
+				!(allowedValuesForArg === "*" || allowedValuesForArg.includes(v))
+					? "<REDACTED>"
+					: v
```
penalosa, line 299: Could you add some comments explaining this logic? It's gotten pretty hairy—but be nice if there was a way to simplify

emily-shen: yeah i'd like to tidy all of this up, because the sanitise + redact + normalise functions have gotten very redundant and messy. Would you be okay with leaving it for now so I can start collecting metrics, and doing that in a fast follow?

packages/wrangler/src/metrics/metrics-dispatcher.ts
emily-shen: because we only want to set the default if someone has passed in the flag without a value, and yargs doesn't have a way of doing it afaict. if someone didn't pass in --x-include-runtime, we want that to be undefined, not the default value.

packages/wrangler/src/metrics/metrics-dispatcher.ts
emily-shen: [6b15d37](https://github.com/cloudflare/workers-sdk/pull/8130/commits/6b15d374e72b901818843b0eee050230a16f53a8)

packages/wrangler/src/metrics/metrics-dispatcher.ts
emily-shen: depends how long until the release really 😅

penalosa: Sounds good--let's get this in as-is for the current release

penalosa (approved):

PR #8138, opened 2025-02-13
Delete interpolation seach

GregBrimble:
WC-3184

Ah well, we gave it a go. Although interpolation search definitely narrows the search space faster and needs fewer comparisons, and does seem to outperform binary in Node.js, I think all that array slicing and dicing brought the cost back up to just above binary's when running in workerd. Even if they were equal, I'd still favor binary search given how much easier it is to read, understand and maintain, so we'd have needed to see some >5% difference for me to personally think this was worth it. Now we know!

I'm keeping around all the tests I wrote for binary, as well as the perf measurement for binary in prod. Useful to have even if we aren't comparing it to anything else.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: interpolation search


WillTaylorDev (approved):
🫡

PR #8141, opened 2025-02-14
[vite-plugin-cloudflare] simplify playground playwright install

dario-piotrowicz:
This PR simplified the vite-plugin-cloudflare playground playwright installation by only installing the chromium browser and nothing else.

This should unblock: https://github.com/vitejs/vite-ecosystem-ci/pull/361

<details><summary>Details</summary>
 
[`playwright install-deps`](https://playwright.dev/docs/browsers#install-system-dependencies) doesn't work great everywhere, for example [it doesn't work on arch linux](https://www.reddit.com/r/linux4noobs/comments/rp76h1/how_can_i_do_playwright_installdeps_on_arch_linux/), the playwright docs are a bit vague on what it installs but it sounds like it might not be generally necessary (as you can also see from the successful CI runs in this PR).

Besides that we only use the chromium browser so only that needs to actually be installed.

</details> 

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: infra change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: unrelated to wrangler
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: infra change

### DIFF
--- a/packages/vite-plugin-cloudflare/playground/package.json
+++ b/packages/vite-plugin-cloudflare/playground/package.json
@@ -6,7 +6,7 @@
 	"scripts": {
 		"check:types": "tsc --build",
 		"test:build": "cross-env VITE_TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
-		"pretest:ci": "pnpm exec playwright install-deps && pnpm exec playwright install",
+		"pretest:ci": "pnpm playwright install chromium",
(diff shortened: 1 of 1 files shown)


petebacondarwin (approved):
Should speed things up a bit too, eh?

PR #8163, opened 2025-02-17
Prevent infinite `console.log()`/`console.info()` loops with Wrangler E2E tests

penalosa:
One of the Wrangler E2E tests is checking to see if inspector messages >1MB can be succesfully sent (a regression test for a bug introduced during phase 1 of the start dev worker work). However, since we don't really want to _actually_ log a 1MB message during tests, we had been mocking the output of `console.log()`/`console.info()`. In Vitest v3, the way we'd done this seems to have been causing an infinite loop, so instead I've replaced the mock implementation with an empty function. We don't really care about _any_ console output in the specific test, so that shouldn't cause any issues.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal testing change

### DIFF
--- a/packages/wrangler/e2e/startWorker.test.ts
+++ b/packages/wrangler/e2e/startWorker.test.ts
@@ -10,7 +10,7 @@ import WebSocket from "ws";
 import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
 import type { DevToolsEvent } from "../src/api";
 
-const OPTIONS = [{ remote: false }, { remote: true }] as const;
+const OPTIONS = [{ remote: true }, { remote: false }] as const;
 
 type Wrangler = Awaited<ReturnType<WranglerE2ETestHelper["importWrangler"]>>;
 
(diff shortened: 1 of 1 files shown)


packages/wrangler/e2e/startWorker.test.ts
```diff
@@ -10,7 +10,7 @@ import WebSocket from "ws";
 import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
 import type { DevToolsEvent } from "../src/api";
 
-const OPTIONS = [{ remote: false }, { remote: true }] as const;
+const OPTIONS = [{ remote: true }, { remote: false }] as const;
```
petebacondarwin, line 13: Why this change? I think it was originally the other way round to get a fast fail.

petebacondarwin (approved):
A couple of e2e tests failed but are probably classic flakes.
But there is a linting error to resolve: https://github.com/cloudflare/workers-sdk/actions/runs/13369830840/job/37335707533?pr=8163#step:5:941

petebacondarwin (approved):

PR #8166, opened 2025-02-17
feat: graduate `--x-include-runtime`

emily-shen:
TODO: rebase on /vnext for inclusion in wrangler v4

Might be best to review commit by commit
- DEVX-1615 improve logging output: d33b18aae5277ce046454779d55977252067c2e1, 1b0d071a4d25c7e6fb81edc0a769bcc70e18023d
- DEVX-1611 remove --x-include-runtime and consolidate type output into one file: 2b990757ed478d1550ea501069a46781a2265f9c
- DEVX-1612 prompt users to rerun wrangler types during dev: 4872cb9ec9556ddc0b3bb0ef6301140acc52f95c

---

This PR will turn on runtime type generation by default for `wrangler types`. 

The `--experimental-include-runtime` flag will be superseded by `include-runtime` and `include-env`, both of which are `true` by default.

Also, the runtime types will be generated in the same file as the env types, at the location specified by path (`worker-configuration.d.ts` by default).

Before:
Screenshot 2025-02-19 at 12 09 27

After:
Screenshot 2025-02-18 at 13 42 47

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: covered by unit tests
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/20082


.changeset/nasty-kiwis-talk.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": minor
```
emily-shen, line 2: major?

packages/wrangler/e2e/types.test.ts
```diff
@@ -29,159 +29,110 @@ const seed = {
...
 		const helper = new WranglerE2ETestHelper();
 		await helper.seed(seed);
-		const output = await helper.run(
-			`wrangler types --x-include-runtime="./types.d.ts"`
-		);
+		await helper.run(`wrangler types "./types.d.ts" `);
 
-		const fileExists = existsSync(path.join(helper.tmpPath, "./types.d.ts"));
+		const file = (await readFile(path.join(helper.tmpPath, "./types.d.ts")))
+			.toString()
```
vicb, line 70: Looks like this can be updated at multiple places
```suggestion
		const file = (await readFile(path.join(helper.tmpPath, "./types.d.ts"), "utf8"))
```

packages/wrangler/e2e/types.test.ts
```diff
@@ -29,159 +29,110 @@ const seed = {
...
-				"compilerOptions": {
-					"types": ["node"]
-				}
-			}
-			`,
-		});
-		const output = await helper.run(
-			`wrangler types --x-include-runtime="./types.d.ts"`
-		);
+		const file2 = (await readFile(typesPath)).toString();
```
vicb, line 102: utf8

packages/wrangler/src/type-generation/helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+import { logger } from "../logger";
+import { generateEnvTypes } from ".";
+import type { Config } from "../config";
+import type { Entry } from "../deployment-bundle/entry";
+
+export const checkTypesDiff = async (config: Config, entry: Entry) => {
+	if (!entry.file.endsWith(".ts")) {
+		return;
+	}
+	let maybeExistingTypesFile: string[];
```
vicb, line 12:
```suggestion
	let maybeExistingTypesFileLines: string[];
```

packages/wrangler/src/type-generation/helpers.ts
```diff
@@ -0,0 +1,66 @@
+import { readFileSync } from "fs";
+import { version } from "workerd";
+import { logger } from "../logger";
+import { generateEnvTypes } from ".";
+import type { Config } from "../config";
+import type { Entry } from "../deployment-bundle/entry";
+
+export const checkTypesDiff = async (config: Config, entry: Entry) => {
```
vicb, line 8: Add a comment about what this fn does?

packages/wrangler/src/type-generation/helpers.ts
```diff
@@ -0,0 +1,66 @@
...
+	)?.groups?.result;
+	const previousEnvInterface = existingEnvHeader?.match(
+		/--env-interface(=|\s)(?<result>[a-zA-Z][a-zA-Z0-9_]*)/
+	)?.groups?.result;
+
+	let newEnvHeader: string | undefined;
+	try {
+		const { envHeader } = await generateEnvTypes(
+			config,
+			{ strictVars: previousStrictVars === "false" ? false : true },
```
vicb, line 38:
```suggestion
			{ strictVars: previousStrictVars !== "false" },
```

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -8,53 +9,60 @@ import { logger } from "../../logger";
...
+		logger.log("");
 	}
 	if (isNodeCompat && !isNodeTypesInstalled) {
-		logger.info(
-			'📣 Since you have Node.js compatibility mode enabled, you should consider adding Node.js for TypeScript by running "npm i --save-dev @types/node@20.8.3". Please see the docs for more details: https://developers.cloudflare.com/workers/languages/typescript/#transitive-loading-of-typesnode-overrides-cloudflareworkers-types'
+		logAction("Install types@node");
+		logger.log(
+			chalk.dim(
+				`Since you have the \`nodejs_compat\` flag, you should install Node.js types by running "npm i --save-dev @types/node${isWorkersTypesInstalled ? '@20.8.3".\nFor more details: https://developers.cloudflare.com/workers/languages/typescript/#known-issues' : '".'}`
+			)
```
vicb, line 58: Something like the following might improve readibility?

```suggestion
				`Since you have the \`nodejs_compat\` flag, you should install Node.js types by running "npm i --save-dev ${isWorkersTypesInstalled ? "@types/node@20.8.3.\nFor more details: https://developers.cloudflare.com/workers/languages/typescript/#known-issues' : '@types/node.'}`
			)
```

vicb:
There's one thing I was thinking about: should we also update the C3 templates?

Currently the `worker-configuration.d.ts` are static files that only contain `interface Env` while the `cf-typegen` scripts will generate both.

Ideally the types should not be in a static file but generated, I have [a pending PR](https://github.com/cloudflare/workers-sdk/pull/7756) that does that for Next.

I think whatever `worker-configuration.d.ts` contains should match the output of the command. And the best way to do that would be to generate this file (whatever config we decide to use).

If we decide to use the runtime types, `tsconfig.json` should also be update and we should not install the old types.

/cc @petebacondarwin

emily-shen:
> There's one thing I was thinking about: should we also update the C3 templates?
> 
> Currently the `worker-configuration.d.ts` are static files that only contain `interface Env` while the `cf-typegen` scripts will generate both.
> 
> Ideally the types should not be in a static file but generated, I have [a pending PR](https://github.com/cloudflare/workers-sdk/pull/7756) that does that for Next.
> 
> I think whatever `worker-configuration.d.ts` contains should match the output of the command. And the best way to do that would be to generate this file (whatever config we decide to use).
> 
> If we decide to use the runtime types, `tsconfig.json` should also be update and we should not install the old types.
> 
> /cc @petebacondarwin

I'm actually working on a PR right now to update C3 to run `wrangler types` instead of using @cloudflare/workers-types. The plan is to release this in v4, make sure v4 is stable, then bump wrangler in C3 and then add switch over types in C3.

fixtures/type-generation/tests/type-generation.file-comment.test.ts
```diff
@@ -28,7 +28,7 @@ describe("`wrangler types` - file comment", () => {
 			"packages",
 			"wrangler"
 		);
-		execSync(`npx ${wranglerPath} types ${args}`, {
+		execSync(`npx ${wranglerPath} types ${args} --include-runtime=false`, {
```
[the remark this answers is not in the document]
emily-shen: hmm thinking about it now, this entire fixture seems a bit redundant to me, i've subsumed the relevant bits into the types e2e and deleted it.

packages/wrangler/src/type-generation/helpers.ts
emily-shen: I prefer the current one as its easier to understand (at least for me)

.changeset/nasty-kiwis-talk.md
```diff
@@ -0,0 +1,11 @@
...
+---
+
+Include runtime types in the output of `wrangler types` by default
+
+`wrangler types` will now produce one file that contains both `Env` types and runtime types based on your compatibility date and flags. This is located at `worker-configuration.d.ts` by default.
```
penalosa, line 7: Worth taking this opportunity to rename this file, maybe to `cloudflare.d.ts`? cc @irvinebroque if you have any thoughts here

.changeset/nasty-kiwis-talk.md
```diff
@@ -0,0 +1,11 @@
...
+
+This behaviour was previously gated behind `--experimental-include-runtime`. That flag is no longer necessary and has been removed. It has been replaced by `--include-runtime` and `--include-env`, both of which are set to `true` by default. If you were previously using `--x-include-runtime`, you can drop that flag and remove the separate `runtime.d.ts` file.
+
+If you were previously using `@cloudflare/workers-types` we recommend you run uninstall and run `wrangler types` instead. Note that `@cloudflare/workers-types` will continue to be published without any changes.
```
penalosa, line 11: Could you add an example for `run uninstall`?

penalosa, line 11: I suspect we will be making some changes (if only to remove the non-latest entrypoints)

```suggestion
If you were previously using `@cloudflare/workers-types` we recommend you run uninstall and run `wrangler types` instead. Note that `@cloudflare/workers-types` will continue to be published.
```

fixtures/type-generation/package.json
penalosa, line 1: Why is this fixture being removed?

packages/wrangler/e2e/types.test.ts
```diff
@@ -29,159 +31,156 @@ const seed = {
...
-		expect(output.stdout).toContain(
-			`"types": ["./.wrangler/types/runtime.d.ts"]`
-		);
+		const output = await helper.run(`wrangler types`);
+		expect(output.stdout).toContain("Generating project types...");
+		expect(output.stdout).toContain("interface Env {");
+		expect(output.stdout).toContain("Generating runtime types...");
+		expect(output.stdout).toContain("Runtime types generated.");
 		expect(output.stdout).toContain(
-			`📣 Since you have Node.js compatibility mode enabled, you should consider adding Node.js for TypeScript by running "npm i --save-dev @types/node@20.8.3". Please see the docs for more details: https://developers.cloudflare.com/workers/languages/typescript/#transitive-loading-of-typesnode-overrides-cloudflareworkers-types`
```
penalosa, line 59: Is the node language being changed?

packages/wrangler/e2e/types.test.ts
```diff
@@ -29,159 +31,156 @@ const seed = {
...
+		expect(file).contains("interface Env");
 	});
 
-	it("should generate runtime types at the provided path", async () => {
+	it("should be able to generate an Env type only", async () => {
 		const helper = new WranglerE2ETestHelper();
 		await helper.seed(seed);
-		const output = await helper.run(
-			`wrangler types --x-include-runtime="./types.d.ts"`
+		const output = await helper.run(`wrangler types --include-runtime=false`);
```
penalosa, line 69: Does `--no-include-runtime` work too?

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -8,53 +9,60 @@ import { logger } from "../../logger";
...
-	if (isWorkersTypesInstalled) {
-		logger.info('📣 You can now uninstall "@cloudflare/workers-types".');
+		`)
+		);
+		logger.log("");
 	}
 	if (isNodeCompat && !isNodeTypesInstalled) {
-		logger.info(
-			'📣 Since you have Node.js compatibility mode enabled, you should consider adding Node.js for TypeScript by running "npm i --save-dev @types/node@20.8.3". Please see the docs for more details: https://developers.cloudflare.com/workers/languages/typescript/#transitive-loading-of-typesnode-overrides-cloudflareworkers-types'
+		logAction("Install types@node");
```
penalosa, line 54:
```suggestion
		logAction("Install @types/node");
```

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -8,53 +9,60 @@ import { logger } from "../../logger";
...
+		logger.log(
+			chalk.dim(
+				`Since you have the \`nodejs_compat\` flag, you should install Node.js types by running "npm i --save-dev ${isWorkersTypesInstalled ? '@types/node@20.8.3".\nFor more details: https://developers.cloudflare.com/workers/languages/typescript/#known-issues' : '@types/node".'}`
```
penalosa, line 57: The `20.8.3` restriction is no longer needed, regardless of whether they use the workers-types package

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -8,53 +9,60 @@ import { logger } from "../../logger";
...
+			)
 		);
+		logger.log("");
 	}
-	logger.info(
-		`📣 Remember to run 'wrangler types --x-include-runtime' again if you change 'compatibility_date' or 'compatibility_flags' in your ${configFileName(configPath)} file.\n`
```
penalosa, line 57: Should we not keep this reminder?

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -42,20 +41,28 @@ export async function generateRuntimeTypes({
 }: {
 	config: Pick<Config, "compatibility_date" | "compatibility_flags">;
 	outFile?: string;
-}) {
+}): Promise<{ runtimeHeader: string; runtimeTypes: string }> {
 	if (!compatibility_date) {
 		throw new Error("Config must have a compatibility date.");
 	}
 
-	await ensureDirectoryExists(outFile);
```
penalosa, line 50: I think we still need this—they could provide a custom outfile

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -42,20 +41,28 @@ export async function generateRuntimeTypes({
...
+	const header = `// Runtime types generated with workerd@${version} ${compatibility_date} ${compatibility_flags.sort().join(",")}`;
 
 	try {
-		const existingTypes = await readFile(outFile, "utf8");
-		if (existingTypes.split("\n")[0] === header) {
+		const lines = (await readFile(outFile, "utf8")).split("\n");
+		const existingHeader = lines.find((line) =>
+			line.startsWith("// Runtime types generated with workerd@")
+		);
+		const existingTypesStart = lines.findIndex(
```
penalosa, line 56: Why do we need this?

packages/wrangler/src/type-generation/runtime/index.ts
```diff
@@ -42,20 +41,28 @@ export async function generateRuntimeTypes({
...
-}) {
+}): Promise<{ runtimeHeader: string; runtimeTypes: string }> {
 	if (!compatibility_date) {
 		throw new Error("Config must have a compatibility date.");
 	}
 
-	await ensureDirectoryExists(outFile);
-
-	const header = `// Runtime types generated with workerd@${version} ${compatibility_date} ${compatibility_flags.join(",")}`;
+	const header = `// Runtime types generated with workerd@${version} ${compatibility_date} ${compatibility_flags.sort().join(",")}`;
```
penalosa, line 49: Since this is now a combined types file, the header should probably be `Types generated with...`, which would also bust all existing caches

packages/wrangler/src/type-generation/index.ts
```diff
@@ -73,86 +115,80 @@ export const typesCommand = createCommand({
...
-			mtls_certificates: config.mtls_certificates,
-			browser: config.browser,
-			ai: config.ai,
-			version_metadata: config.version_metadata,
-			secrets,
-			assets: config.assets,
-			workflows: config.workflows,
-		};
+		logger.log(
+			chalk.dim("────────────────────────────────────────────────────────────")
```
penalosa, line 169: This probably won't look great on differently sized terminals

packages/wrangler/src/type-generation/index.ts
```diff
@@ -458,72 +521,60 @@ async function generateTypes(
...
+		envTypeStructure.length || modulesTypeStructure.length;
+	if (entrypointFormat === "modules" || typesHaveBeenFound) {
+		const { fileContent, consoleOutput } = generateTypeStrings(
+			entrypointFormat,
+			envInterface,
+			envTypeStructure.map(([key, value]) => `${key}: ${value};`),
+			modulesTypeStructure
+		);
+		// todo replace envInterface with generic
+		const hash = blake3hash(consoleOutput).toString("hex").slice(0, 32);
```
penalosa, line 536: What about just using `crypto.createHash("sha256")` here? https://nodejs.org/docs/latest-v18.x/api/crypto.html#cryptocreatehashalgorithm-options

penalosa:
Could we have a more unified single header? Something like:

```
// Generated by Wrangler by running `wrangler types` (workerd@1.20250214.0, compatibility_date: 2023-05-04, compatibility_flags: [], env hash: e82ba4d7b995dd9ca6fb0332d81f889b)
```

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -8,53 +9,60 @@ import { logger } from "../../logger";
...
-		logger.info(dedent`
-			📣 Add the generated types to the types array in your tsconfig.json:
-
+		logAction("Update your tsconfig.json to include the generated types");
+		logger.log(
+			chalk.dim(dedent`
 				{
 					"compilerOptions": {
-						...
 						"types": ${updatedTypesString}
```
penalosa, line 46: I don't think this is actually required? It seems to include that file for me by default

fixtures/type-generation/package.json
emily-shen: as far as i can tell, its not doing anything that the e2e tests aren't already doing - happy to put it back though if i'm wrong!

packages/wrangler/src/type-generation/runtime/index.ts
emily-shen: afaict, we want to be able to use cached runtime types while updating env types, so this just lets us extract the part of the file that is runtime types.

packages/wrangler/src/type-generation/runtime/index.ts
emily-shen: this is checked in the validateArgs bit of the command definition now - `checkPath` in line 99 or so, can't figure out how to link that line... :')

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
emily-shen: huh yeah so it does. Is it because its covered by `**/*.ts` in `include`??

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
emily-shen: oh cool - can i remove that from the docs then?

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
emily-shen: yep its still around, just moved it to print at the end of everything and cover changing bindings too - see the screenshot in the pr description (its logged out in a test somewhere too)

packages/wrangler/e2e/types.test.ts
emily-shen: yes but that bothers me grammatically :')

.changeset/nasty-kiwis-talk.md
emily-shen: as in, something like "You can run `npm uninstall @cloudflare/workers-types`"?

emily-shen:
> Could we have a more unified single header? Something like:
> 
> ```
> // Generated by Wrangler by running `wrangler types` (workerd@1.20250214.0, compatibility_date: 2023-05-04, compatibility_flags: [], env hash: e82ba4d7b995dd9ca6fb0332d81f889b)
> ```

hmm we could, but since we can use cached runtime types independently of env changes, its mildly more convenient to just have that bit on its own line, and i don't think it'll particularly matter to users.  happy to change it if you feel strongly about it.

packages/wrangler/src/type-generation/runtime/index.ts
emily-shen: https://github.com/cloudflare/workers-sdk/pull/8166#issuecomment-2688636266 
also why do we want to bust existing caches?

packages/wrangler/e2e/types.test.ts
emily-shen: removed in [7e0c6a1](https://github.com/cloudflare/workers-sdk/pull/8166/commits/7e0c6a19e3443b3903da579d58746214b4f1642d)

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
penalosa: Looks like the default is `**/*`: https://www.typescriptlang.org/tsconfig/#include

fixtures/type-generation/package.json
penalosa: No, that sounds good!

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
penalosa: Yep (I thought it already was removed, but maybe not)

packages/wrangler/src/type-generation/index.ts
```diff
@@ -73,86 +115,78 @@ export const typesCommand = createCommand({
...
-			workflows: config.workflows,
-		};
+		logHorizontalRule();
 
-		await generateTypes(
-			configBindingsWithSecrets,
-			config,
-			envInterface,
-			outputPath
+		if ((header.length && content.length) || entrypointFormat === "modules") {
```
penalosa, line 170: I realise this was probably here before, but I don't really understand this logic. Presumably at this stage there will _always_ be something in header & content? What does the entrypointFormat check do?

packages/wrangler/src/type-generation/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+	);
+	const newRuntimeHeader = `// Runtime types generated with workerd@${version} ${config.compatibility_date} ${config.compatibility_flags.sort().join(",")}`;
+
+	const envOutOfDate = existingEnvHeader && maybeExistingHash !== newHash;
+	const runtimeOutOfDate =
+		existingRuntimeHeader && existingRuntimeHeader !== newRuntimeHeader;
+
+	if (envOutOfDate || runtimeOutOfDate) {
+		logger.log(
+			"❓ It looks like your types might be out of date. Have you updated your config file since last running `wrangler types`?"
```
penalosa, line 67: Could we make this more assertive? What's the recommended user action here?

Right now it's not clear what the user should do as a result of their answer to the question posed

packages/wrangler/src/type-generation/runtime/index.ts
penalosa, line 56: It seems like there's some duplication here between checkTypesDiff and here—is there any consolidating possible, or is that more effort than it's worth?

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
```diff
@@ -8,71 +8,50 @@ import { logger } from "../../logger";
...
 	outFile: string,
 	tsconfigTypes: string[],
-	isNodeCompat = false,
-	configPath: string | undefined
+	isNodeCompat = false
 ) {
 	const isWorkersTypesInstalled = tsconfigTypes.find((type) =>
 		type.startsWith("@cloudflare/workers-types")
 	);
 	const isNodeTypesInstalled = tsconfigTypes.find((type) => type === "node");
```
penalosa, line 16: I'm not sure this logic is fully correct. If the types array is empty, but `@types/node` is in `node_modules` it'll automatically be picked up as well.

(I realise this is existing logic, but we might as well fix it up)

penalosa (approved):
Couple comments, but overall this is looking great!

packages/wrangler/src/type-generation/index.ts
emily-shen: this is for the extremely niche situation where its a service syntax worker and --include-runtime=false - we don't want to write an empty Env interface for that case.
I have no idea  _why_ that would be a problem, but we do have a test making sure we don't... any ideas? 😅

packages/wrangler/src/type-generation/runtime/index.ts
emily-shen: gave it a go, ends up being a bit convoluted - think its not really worth it

packages/wrangler/src/type-generation/runtime/log-runtime-types-message.ts
emily-shen: [600a895](https://github.com/cloudflare/workers-sdk/pull/8166/commits/600a895292d840048b29b5fdf2fc51f7fc7e4e4f) fixed here, but i can't figure out how to test it because the tests always run in a context where @types/nodes exists 😢

PR #8167, opened 2025-02-17
Reduce C3 CI runs

penalosa:
To address CI queuing issues (C3 E2E is the most expensive job across the entire Cloudflare org), reduce C3 E2E CI runs to:
* Run normal C3 E2E on VP and PRs with the `c3-e2e` label
* Run C3 E2E quarantine at 3AM every day, when no other jobs are usually running (additionally, this can be manually run on a specific branch if needed)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: change to CI setup
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: change to CI setup
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: added in the README

### DIFF
--- a/.github/pr-labeler.config.yml
+++ b/.github/pr-labeler.config.yml
@@ -4,6 +4,11 @@ e2e:
       - changed-files:
           - any-glob-to-any-file: "packages/wrangler/e2e/**/*"
 
+c3-e2e:
+  - all:
+      - changed-files:
+          - any-glob-to-any-file: "packages/create-cloudflare/e2e-tests/**/*"
+
 # add 'skip-pr-description-validation' label to Version Packages PRs
 skip-pr-description-validation:
   - all:
--- a/.github/pr-labeler.config.yml
+++ b/.github/pr-labeler.config.yml
@@ -7,7 +7,7 @@ e2e:
 c3-e2e:
   - all:
       - changed-files:
-          - any-glob-to-any-file: "packages/create-cloudflare/e2e-tests/**/*"
+          - any-glob-to-any-file: "packages/create-cloudflare/**/*"
 
 # add 'skip-pr-description-validation' label to Version Packages PRs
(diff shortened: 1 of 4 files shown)


.github/pr-labeler.config.yml
```diff
@@ -4,6 +4,11 @@ e2e:
       - changed-files:
           - any-glob-to-any-file: "packages/wrangler/e2e/**/*"
 
+c3-e2e:
+  - all:
+      - changed-files:
+          - any-glob-to-any-file: "packages/create-cloudflare/e2e-tests/**/*"
```
petebacondarwin, line 10: How about also running whenever any files in `packages/create-cloudflare` change?

.github/workflows/c3-e2e-quarantine.yml
```diff
@@ -1,12 +1,8 @@
...
-
 name: C3 E2E (Quarantine)
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - packages/create-cloudflare/**
+  schedule:
+    - cron: "0 3 * * *"
```
petebacondarwin, line 4: I wonder if for quarantined tests we should fail if any of the tests actually pass and vice-versa. This would be a nice setup, since we would have to go and remove tests from quarantine when they start passing.
Then we can continue to run these in VP etc.

petebacondarwin (approved):
Couple of non-blocking suggestions

.github/pr-labeler.config.yml
penalosa: [09f7cc2](https://github.com/cloudflare/workers-sdk/pull/8167/commits/09f7cc2eee6e4d16de0eaec3307ed9a8489a1f2d)

PR #8175, opened 2025-02-17
fix(wrangler): ignore comments when splitting sql into statements

### LINKED ISSUE #7739: 🐛 BUG: applyD1Migrations fails with output from readD1Migrations if last line of a migration is a comment

Workers Vitest Integration

3.101.0

Mac Sequoia 15.2

_No response_

### Describe the Bug

Following the example vitest configuration from:
https://github.com/cloudflare/workers-sdk/blob/d102b60238c1dddfdd829ffee62c451cb526717a/fixtures/vitest-pool-workers-examples/d1/vitest.config.ts

and

https://github.com/cloudflare/workers-sdk/blob/d102b60238c1dddfdd829ffee62c451cb526717a/fixtures/vitest-pool-workers-examples/d1/test/apply-migrations.ts

Will fail if one of the migrations contains a comment as its last line.  Given that generating migrations with wrangler will insert a line like: 

`-- Migration number: 0001 	 2024-12-30T21:23:40.357Z`

It's easy to get a failure if the comment ends up at the bottom of the file. 

Failure message is:
```
Error: D1_ERROR: SQL code did not contain a statement.
    at D1DatabaseSessionAlwaysPrimary._sendOrThrow (cloudflare-internal:d1-api:129:19)
    at D1DatabaseSessionAlwaysPrimary.batch (cloudflare-internal:d1-api:81:23)
```

_No response_

edmundhung:
Fixes #7739.

This makes `splitSqlQuery` ignores both inline and block comments in the result query when splitting it into statements to avoid comments at the end of the SQL makes it into an extra statement.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


CarmenPopoviciu:
I have spoken with the D1 team and they will have a first pass at this PR as there are some concerns on their side this might not match prod behaviour. Will wait on their thoughts before reviewing the PR myself ⏳

CarmenPopoviciu (approved):

PR #8176, opened 2025-02-17
fix: refactor Node.js compat support to ensure all polyfills are pre-bundled before the first request

petebacondarwin:
Fixes #0000.

This is not fixing a bug per-se, except that prior to this change some of the Node.js polyfills would only be processed by the dependency optimizer during/after the first request was made. This meant there was a danger of the dependencies that had been prebundled already becoming stale and the app being in a bad state.

- makes use of the new custom builtin support that @dario-piotrowicz added to Vite recently
- proactively pre-bundles all the Node.js compat polyfill entry-points
- removes the need to use the Vite aliasing mechanism (which is not configurable per environment) and so the problematic virtual modules
- dramatically simplifies the Node.js compat plugin code base

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: the current tests cover the refactoring
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Not Wrangler related
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -1,5 +1,6 @@
 import assert from "node:assert";
 import * as fs from "node:fs";
+import { builtinModules } from "node:module";
 import * as path from "node:path";
 import { createMiddleware } from "@hattip/adapter-node";
 import MagicString from "magic-string";
@@ -24,11 +25,7 @@ import {
 } from "./node-js-compat";
 import { resolvePluginConfig } from "./plugin-config";
 import { MODULE_PATTERN } from "./shared";
-import {
-	getOutputDirectory,
-	nodeBuiltInModules,
-	toMiniflareRequest,
-} from "./utils";
+import { getOutputDirectory, toMiniflareRequest } from "./utils";
 import { handleWebSocket } from "./websockets";
 import { getWarningForWorkersConfigs } from "./workers-configs";
 import type { ModuleType } from "./constants";
@@ -410,30 +407,41 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 				// Skip this whole plugin if we are in preview mode
 				return !env.isPreview;
 			},
+			configEnvironment(name) {
+				// Only configure this environment if it is a Worker using Node.js compatibility.
+				if (isNodeCompat(getWorkerConfig(name))) {
+					return {
+						resolve: {
+							builtins: getNodeCompatExternals(),
+						},
+						optimizeDeps: {
+							// This is a list of dependency entry-points that should be pre-bundled.
+							// In this case we provide a list of all the possible polyfills so that they are pre-bundled,
+							// ready ahead the first request to the dev server.
+							// Without this the dependency optimizer will try to bundle them on-the-fly in the middle of the first request,
+							// which can potentially cause problems if it leads to previous pre-bundling to become stale and needing to be reloaded.
+							include: [...getNodeCompatEntries()],
+							// This is a list of module specifiers that the dependency optimizer should not follow when doing import analysis.
+							// In this case we provide a list of all the Node.js modules, both those built-in to workerd and those that will be polyfilled.
+							// Obviously we don't want/need the optimizer to try to process modules that are built-in;
+							// But also we want to avoid following the ones that are polyfilled since the dependency-optimizer import analyzer does not
+							// resolve these imports using our `resolveId()` hook causing the optimization step to fail.
+							exclude: [
+								...builtinModules,
+								...builtinModules.map((m) => `node:${m}`),
+							],
+						},
+					};
+				}
+			},
 			applyToEnvironment(environment) {
-				const workerConfig = getWorkerConfig(environment.name);
-				return isNodeCompat(workerConfig);
+				// Only run this plugin's runtime hooks if it is a Worker using Node.js compatibility.
+				return isNodeCompat(getWorkerConfig(environment.name));
 			},
 			// We need the resolver from this plugin to run before built-in ones, otherwise Vite's built-in
 			// resolver will try to externalize the Node.js module imports (e.g. `perf_hooks` and `node:tty`)
 			// rather than allowing the resolve hook here to alias then to polyfills.
 			enforce: "pre",
-			configEnvironment() {
-				return {
-					resolve: {
-						builtins: getNodeCompatExternals(),
-					},
-					optimizeDeps: {
-						// We don't want to follow any `node:*` type imports when optimizing deps,
-						// whether or not they are built-in to workerd or polyfilled here.
-						// The ones that are not built-in will just get resolved to their polyfill on demand.
-						exclude: [...nodeBuiltInModules],
-						// But we do want to ensure that all the possible entry-points for the polyfills get pre-bundled
-						// so that they are ready for the first request to the dev server.
-						include: [...getNodeCompatEntries()],
-					},
-				};
-			},
 			async resolveId(source, importer, options) {
 				// See if we can map the `source` to a Node.js compat alias.
 				const result = resolveNodeJSImport(source);
@@ -442,11 +450,15 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					return this.resolve(source, importer, options);
 				}
 
-				if (this.environment.mode === "dev" && this.environment.depsOptimizer) {
+				if (this.environment.mode === "dev") {
+					assert(
+						this.environment.depsOptimizer,
+						"depsOptimizer is required in dev mode"
+					);
 					// We are in dev mode (rather than build).
 					// Node.js compat polyfill modules have already been pre-bundled,
 					// so we can use the unresolved path to the polyfill
-					// and let the dependency optimizer resolve the path to the cached version.
+					// and let the dependency optimizer resolve the path to the pre-bundled version.
 					return this.resolve(result.unresolved, importer, options);
 				}
 
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -16,16 +16,19 @@ import {
 	getPreviewMiniflareOptions,
 } from "./miniflare-options";
 import {
-	getNodeCompatAliases,
+	getNodeCompatEntries,
 	getNodeCompatExternals,
 	injectGlobalCode,
 	isNodeCompat,
-	maybeStripNodeJsVirtualPrefix,
 	resolveNodeJSImport,
 } from "./node-js-compat";
 import { resolvePluginConfig } from "./plugin-config";
 import { MODULE_PATTERN } from "./shared";
-import { getOutputDirectory, toMiniflareRequest } from "./utils";
+import {
+	getOutputDirectory,
+	nodeBuiltInModules,
+	toMiniflareRequest,
+} from "./utils";
 import { handleWebSocket } from "./websockets";
 import { getWarningForWorkersConfigs } from "./workers-configs";
 import type { ModuleType } from "./constants";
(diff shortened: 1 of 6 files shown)


dario-piotrowicz:
@petebacondarwin with this change I think we can remove the following check:
https://github.com/cloudflare/workers-sdk/blob/0b7b263921e1a85eadd83490c069bf7f831af261/packages/vite-plugin-cloudflare/src/miniflare-options.ts#L336

Screenshot 2025-02-17 at 21 49 45

And with that completely get rid of this variable:
https://github.com/cloudflare/workers-sdk/blob/0b7b263921e1a85eadd83490c069bf7f831af261/packages/vite-plugin-cloudflare/src/utils.ts#L7-L9

Could you have a look into it and if everything seems right to you remove these in the PR? 🙂

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -407,67 +410,53 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
-							unresolved,
-							resolved
-						).id;
-					// We must pass the id from the depsOptimizer through the rest of the
-					// resolving pipeline to ensure that the optimized version gets used.
-					return this.resolve(optimized, importer, options);
+					// We are in dev mode (rather than build).
+					// Node.js compat polyfill modules have already been pre-bundled,
+					// so we can use the unresolved path to the polyfill
+					// and let the dependency optimizer resolve the path to the cached version.
```
dario-piotrowicz, line 449: very nit:
```suggestion
					// and let the dependency optimizer resolve the path to the pre-bundled version.
```
(cached can potentially lead to confusion? 🤔)

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -407,67 +410,53 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+					return this.resolve(result.unresolved, importer, options);
 				}
 
-				return this.resolve(resolved, importer, options);
+				// We are in build mode so return the absolute path to the polyfill.
```
dario-piotrowicz, line 453: Nit:
Here you say that we are in build mode, which I think it's correct, but I couldn't tell by the `if` statement above:
```ts
if (this.environment.mode === "dev" && this.environment.depsOptimizer) {
```

could we remove the `&& this.environment.depsOptimizer` and use an `assert` instead?
Something like:
Screenshot 2025-02-17 at 22 16 00

I think this would make the code easier to follow 🙂
(and having the `depsOptimizer` is a hard requirement in dev for us, isn't it?)

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -43,21 +41,38 @@ export function isNodeCompat(
 }
 
 /**
- * If the current environment needs Node.js compatibility,
- * then inject the necessary global polyfills into the code.
+ * Get a set of module specifiers for all possible Node.js compat polyfill entry-points
```
dario-piotrowicz, line 44:
```suggestion
 * Gets a set of module specifiers for all possible Node.js compat polyfill entry-points
```

😅 🙈

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -43,21 +41,38 @@ export function isNodeCompat(
...
+ */
+export function getNodeCompatEntries() {
+	const entries = new Set<string>(Object.values(env.alias));
+	for (const globInject of Object.values(env.inject)) {
+		if (typeof globInject === "string") {
+			entries.add(globInject);
+		} else {
+			entries.add(globInject[0] as string);
```
dario-piotrowicz, line 52: what about
```suggestion
			entries.add(globInject[0]!);
```
?

packages/vite-plugin-cloudflare/src/node-js-compat.ts
dario-piotrowicz: (this comment also applies for the other jsdocs 😅)

dario-piotrowicz (approved):
Looks great to me 🤩 🚀

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -407,67 +410,53 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+						builtins: getNodeCompatExternals(),
+					},
+					optimizeDeps: {
+						// We don't want to follow any `node:*` type imports when optimizing deps,
+						// whether or not they are built-in to workerd or polyfilled here.
+						// The ones that are not built-in will just get resolved to their polyfill on demand.
+						exclude: [...nodeBuiltInModules],
+						// But we do want to ensure that all the possible entry-points for the polyfills get pre-bundled
+						// so that they are ready for the first request to the dev server.
+						include: [...getNodeCompatEntries()],
```
jamesopstad, line 433: Does `include` override `exclude` here? All Node builtins are excluded, including those that are unprefixed and those that are prefixed with `node:` and then `include` adds the polyfilled builtins back in? If so, I think the comments could make this clearer.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -43,21 +41,38 @@ export function isNodeCompat(
...
+		}
+	}
```
jamesopstad, line 54:
```suggestion
	for (const globalInject of Object.values(env.inject)) {
		if (typeof globalInject === "string") {
			entries.add(globalInject);
		} else {
			assert(globalInject[0]);
			entries.add(globalInject[0]);
		}
	}
```

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -43,21 +41,38 @@ export function isNodeCompat(
...
 				const moduleSpecifier = globalInject;
 				// the mapping is a simple string, indicating a default export, so the string is just the module specifier.
-				return `import var_${globalName} from "${CLOUDFLARE_VIRTUAL_PREFIX}${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName};\n`;
+				return `import var_${globalName} from "${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName};\n`;
 			}
 
 			// the mapping is a 2 item tuple, indicating a named export, made up of a module specifier and an export name.
 			const [moduleSpecifier, exportName] = globalInject;
-			return `import var_${globalName} from "${CLOUDFLARE_VIRTUAL_PREFIX}${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName}.${exportName};\n`;
+			return `import var_${globalName} from "${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName}.${exportName};\n`;
```
jamesopstad, line 75:
```suggestion
			assert(moduleSpecifier && exportName, `Unexpected error: globalInject is not a string or tuple`);
			return `import var_${globalName} from "${moduleSpecifier}";\nglobalThis.${globalName} = var_${globalName}.${exportName};\n`;
```

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -93,40 +92,22 @@ export function getNodeCompatExternals(): string[] {
...
-		: undefined;
-}
-
-/**
- * Resolve the source import to an absolute path, potentially via its alias.
+ * Resolve the `source` to a Node.js compat alias if possible.
+ *
+ * If there is an alias, the return value is an object with:
+ * - `unresolved`: a bare import path to the polyfill (e.g. `unenv/runtime/node/crypto`)
+ * - `resolved` an absolute path to the polyfill (e.g. `/path/to/project/node_modules/unenv/runtime/node/child_process/index.mjs`)
```
jamesopstad, line 99:
```suggestion
 * - `resolved`: an absolute path to the polyfill (e.g. `/path/to/project/node_modules/unenv/runtime/node/child_process/index.mjs`)
```

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad, line 433: Just checking that it's been confirmed across package managers that ESBuild can resolve the unenv imports here? Do the new E2E tests for the plugin run on every PR?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -407,67 +410,53 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+			applyToEnvironment(environment) {
+				const workerConfig = getWorkerConfig(environment.name);
+				return isNodeCompat(workerConfig);
+			},
+			// We need the resolver from this plugin to run before built-in ones, otherwise Vite's built-in
+			// resolver will try to externalize the Node.js module imports (e.g. `perf_hooks` and `node:tty`)
+			// rather than allowing the resolve hook here to alias then to polyfills.
+			enforce: "pre",
+			configEnvironment() {
 				return {
```
jamesopstad, line 422: `configEnvironment` runs before `applyToEnvironment` so this configuration is currently being applied to all environments (including the default `client` and `ssr` environments).

```suggestion
			configEnvironment(name) {
				const workerConfig = getWorkerConfig(name);
				
				if (isNodeCompat(workerConfig)) {
				  return {
```

(Can't close the if statement in the suggestion because GitHub won't let me).

I would also move the `applyToEnvironment` hook after the `configEnvironment` hook to make this clearer.

jamesopstad (approved):
Amazing work @petebacondarwin and @dario-piotrowicz! ❤️ I love that this is so much simpler now.

I'm approving pending the suggested changes.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: Hmm. This is definitely subjective. But since you asked (and the [Google JS style-guide](https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments:~:text=Method%20descriptions%20begin%20with%20a%20verb%20phrase%20that%20describes%20what%20the%20method%20does.%20This%20phrase%20is%20not%20an%20imperative%20sentence%2C%20but%20instead%20is%20written%20in%20the%20third%20person%2C%20as%20if%20there%20is%20an%20implied%20This%20method%20...%20before) agrees with you) I'll update these.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: We actively have linting rules against use of !, so I would have to also add a lint comment to turn off that rule.
I'll fix it.

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: Include is the list of packages that should be pre-bundled.
Exclude is the list of imports that should not be followed when doing the import analysis.
I am not quite sure what happens when there is overlap.
But in our case here, there is no overlap. The entries are all of the form `unenv/runtime/...` or `@cloudflare/unenv-preset/runtime/...`.
I'll try to make the comments clearer.

petebacondarwin: The new e2e tests run on every PR - as part of the `test:ci` run for the `@cloudflare/vite-plugin` package.
I will also double check with a manually created project.

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: This is a great spot @jamesopstad - thanks. I didn't realise that the applyToEnvironments only affects which runtime hooks get called and not whether configureEnvironments is called.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Sorry, I should have deleted this comment after looking at the rest of the code because it made sense then.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: 4de9f48d9c6af8eb2319b4cb8ce5f4f892b99c53

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: 4de9f48d9c6af8eb2319b4cb8ce5f4f892b99c53

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: 4de9f48d9c6af8eb2319b4cb8ce5f4f892b99c53

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: 4de9f48d9c6af8eb2319b4cb8ce5f4f892b99c53

petebacondarwin: Is this necessary? It'll totally break anyway if these are not defined.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: 4de9f48d9c6af8eb2319b4cb8ce5f4f892b99c53

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: 4de9f48d9c6af8eb2319b4cb8ce5f4f892b99c53

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: Tested manually and all seems good.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
jamesopstad: Up to you. I would rather control the error that's reported if it breaks. Though, as I also mentioned it would be better if unenv typed tuples as `[string, string]`.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
jamesopstad: Sorry to be picky but I would rather `globInject` is renamed to `globalInject`. `globInject` just makes me think it's something to do with globs.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: If the types were correct then we wouldn't add the assertion, and we'd get the same error as right now.
But given that you asked, I'll add it 🤗

petebacondarwin: 9471f87860fef681269c0b8d7a31335c1ebd952f

packages/vite-plugin-cloudflare/src/node-js-compat.ts
petebacondarwin: 2c335f21ddba84e69051d9a74a5a1a9c6be997df

packages/vite-plugin-cloudflare/src/node-js-compat.ts
dario-piotrowicz: > Hmm. This is definitely subjective.

yeah I know 😅, but I also comment about it as a consistency thing, since I think that in this monorepo the third person form for jsdocs is the most common one

packages/vite-plugin-cloudflare/src/node-js-compat.ts
dario-piotrowicz: > We actively have linting rules against use of !, so I would have to also add a lint comment to turn off that rule.

yeah I was expecting that, but I did try in vscode to add `!` and vscode didn't seem to complain about it 🤷‍♂️  (I did not try to run the linting script though)

PR #8177, opened 2025-02-17
More dependency updates

penalosa:
Continuing on theme...


packages/workers-shared/package.json
```diff
@@ -57,7 +57,7 @@
 	},
 	"devDependencies": {
 		"@cloudflare/eslint-config-worker": "workspace:*",
-		"@cloudflare/vitest-pool-workers": "^0.5.31",
+		"@cloudflare/vitest-pool-workers": "^0.5.41",
```
Skye-31, line 60: Could that be a workspace dep?

penalosa: Not in this case—there's a circular dep:

`@cloudflare/vitest-pool-workers` -> `miniflare` -> `@cloudflare/workers-shared`

emily-shen (approved):

PR #8186, opened 2025-02-19
Add cloudchamber images list/delete commands

IRCody:
Adds a cloudchamber images {list,delete} commands for operating on the cloudchamber managed registry.
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: No e2e test for Cloudchamber yet
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: The cloudchamber subcommand is not documented in that repo.

### DIFF
--- a/.changeset/silly-dragons-doubt.md
+++ b/.changeset/silly-dragons-doubt.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add cloudchamber images {list,delete} commands to list and delete images stored in cloudchamber managed registry.
--- a/packages/wrangler/src/cloudchamber/images/list.ts
+++ b/packages/wrangler/src/cloudchamber/images/list.ts
@@ -1,0 +1,248 @@
+import { logger } from "../../logger";
+import { ImageRegistriesService } from "../client";
+import { handleFailure, promiseSpinner } from "../common";
+import type { Config } from "../../config";
+import type {
+	CommonYargsArgvJSON,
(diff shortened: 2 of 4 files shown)


packages/wrangler/src/cloudchamber/images/list.ts
```diff
@@ -0,0 +1,248 @@
...
+	});
+
+	if (!deleteResponse.ok) {
+		throw new Error(
+			`Failed to delete tag "${tag}" (digest: ${digest}): ${deleteResponse.status} ${deleteResponse.statusText}`
+		);
+	}
+}
+
+async function getCreds(): Promise<string> {
```
skepticfx, line 241: Should we memoize this for the duration of the expiration and only fetch another if the JWT has expired?

skepticfx: Yeah. Let us handle this another time, since it should be its own PR.

skepticfx (approved):

.changeset/silly-dragons-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
emily-shen, line 2: should be a patch as cloudchamber is experimental

emily-shen (approved):

PR #8191, opened 2025-02-19
fix(hybrid-nodejs-compat): inject modules only once

vicb:
This PR makes sure that modules are injected only once before injected all the related values.

You can see the effect for (from the Cloudflare preset):

```ts
  inject: {
    setImmediate: ["unenv/runtime/node/timers/$cloudflare", "setImmediate"],
    clearImmediate: ["unenv/runtime/node/timers/$cloudflare", "clearImmediate"]
  },
```

The generated code before this PR:

```ts
// ../../packages/wrangler/_virtual_unenv_global_polyfill-set$immediate.js
var init_virtual_unenv_global_polyfill_set_immediate = __esm({
  "../../packages/wrangler/_virtual_unenv_global_polyfill-set$immediate.js"() {
    init_cloudflare();
    globalThis.setImmediate = setImmediateFallback;
  }
});

// ../../packages/wrangler/_virtual_unenv_global_polyfill-clear$immediate.js
var init_virtual_unenv_global_polyfill_clear_immediate = __esm({
  "../../packages/wrangler/_virtual_unenv_global_polyfill-clear$immediate.js"() {
    init_cloudflare();
    globalThis.clearImmediate = clearImmediateFallback;
  }
});

var init_utils = __esm({
  "../../node_modules/.pnpm/unenv@2.0.0-rc.1/node_modules/unenv/runtime/_internal/utils.mjs"() {
    init_virtual_unenv_global_polyfill_process();
    init_virtual_unenv_global_polyfill_performance();
    init_virtual_unenv_global_polyfill_console();
    // BEFORE {
    init_virtual_unenv_global_polyfill_set_immediate();
    init_virtual_unenv_global_polyfill_clear_immediate();
    // } BEFORE
    __name(mergeFns, "mergeFns");
    __name(createNotImplementedError, "createNotImplementedError");
    __name(notImplemented, "notImplemented");
    __name(notImplementedAsync, "notImplementedAsync");
    __name(notImplementedClass, "notImplementedClass");
  }
});
```

and after this PR:

```ts
// ../../packages/wrangler/_virtual_unenv_global_polyfill-unenv-runtime-node-timers-$cloudflare
var init_virtual_unenv_global_polyfill_unenv_runtime_node_timers_cloudflare = __esm({
  "../../packages/wrangler/_virtual_unenv_global_polyfill-unenv-runtime-node-timers-$cloudflare"() {
    init_cloudflare();
    globalThis.setImmediate = setImmediateFallback;
    globalThis.clearImmediate = clearImmediateFallback;
  }
});

var init_utils = __esm({
  "../../node_modules/.pnpm/unenv@2.0.0-rc.1/node_modules/unenv/runtime/_internal/utils.mjs"() {
    init_virtual_unenv_global_polyfill_unenv_runtime_node_process_cloudflare();
    init_virtual_unenv_global_polyfill_unenv_runtime_polyfill_performance();
    init_virtual_unenv_global_polyfill_unenv_runtime_node_console_cloudflare();
    // AFTER {
    init_virtual_unenv_global_polyfill_unenv_runtime_node_timers_cloudflare();
    // } AFTER
    __name(mergeFns, "mergeFns");
    __name(createNotImplementedError, "createNotImplementedError");
    __name(notImplemented, "notImplemented");
    __name(notImplementedAsync, "notImplementedAsync");
    __name(notImplementedClass, "notImplementedClass");
  }
});
```

So the first thing is that the code size is reduced - as a data point `init_virtual_unenv_global_polyfill_unenv_runtime_node_console_cloudflare(); /   init_virtual_unenv_global_polyfill_unenv_runtime_node_timers_cloudflare();` is repeated 106 times in the `nodejs-hybrid-app` test.

Another benefit (which actually is the reason for which I created this PR) is that you can inject multiple export from the same file when they have a dependency.

For examples, we would like to inject `performance` and `Performance` from `perf_hooks`. However `performance` needs to instantiate a `Performance` so there is no way we could initialize with the current code.

Let's say `Performance` is injected first from `perf_hooks`. So `perf_hooks` will be imported. The first thing that module will do (see `init_utils` above) would be to inject all the values. It will then try to inject `performance` which imports `perf_hooks` too which result in ESBuild doing nothing (because of the circular dependency) and then it can instantiate `Performance` because it is undefined.

One last benefit is that we would stop using the polyfills (i.e. `init_virtual_unenv_global_polyfill_unenv_runtime_polyfill_performance();`) as I think the preferred way is to use `inject` (the performance polyfill directly adds props on `globalThis`)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: covered by the existing tests (I also ran the OpenNext tests)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/wrangler/src/__tests__/hybrid-nodejs-compat.test.ts
petebacondarwin, line 1: You are deleting this file because these are tested elsewhere?

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -169,71 +169,61 @@ function handleNodeJSGlobals(
...
+		string,
+		{ injectedName: string; exportName: string; importName: string }[]
+	>();
+
+	const moduleIdSpecifier = new Map<string, string>();
+
+	for (const [injectedName, moduleSpecifier] of Object.entries(inject)) {
+		const [module, exportName, importName] = Array.isArray(moduleSpecifier)
+			? [moduleSpecifier[0], moduleSpecifier[1], moduleSpecifier[1]]
+			: [moduleSpecifier, "default", "defaultExport"];
```
petebacondarwin, line 194: This is a really nice way to collect the data from this awkward format.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -169,71 +169,61 @@ function handleNodeJSGlobals(
...
+	 * - `injectedName`: the name injected on `globalThis`
+	 * - `exportName`: the export name from the module
+	 * - `importName`: the imported name
+	 */
+	const injectsByModule = new Map<
+		string,
+		{ injectedName: string; exportName: string; importName: string }[]
+	>();
+
+	const moduleIdSpecifier = new Map<string, string>();
```
petebacondarwin, line 189: It was not clear to me what this map is for. Perhaps renaming would help?

```suggestion
	const virtualModuleToModuleSpecifierMap = new Map<string, string>();
```

I think we could simplify things a bit if the keys of this map contained the whole path (i.e. `${prefix}${id}`), which would mean we don't need to use the regex below to extract that bit from the path, we can just use it directly.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -169,71 +169,61 @@ function handleNodeJSGlobals(
...
+
+	for (const [injectedName, moduleSpecifier] of Object.entries(inject)) {
+		const [module, exportName, importName] = Array.isArray(moduleSpecifier)
+			? [moduleSpecifier[0], moduleSpecifier[1], moduleSpecifier[1]]
+			: [moduleSpecifier, "default", "defaultExport"];
+
+		if (!injectsByModule.has(module)) {
+			injectsByModule.set(module, []);
+			moduleIdSpecifier.set(module.replaceAll("/", "-"), module);
```
petebacondarwin, line 198: E.g.

```suggestion
			moduleIdSpecifier.set(`${prefix}${module.replaceAll("/", "-")}`, module);
```

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -169,71 +169,61 @@ function handleNodeJSGlobals(
...
+		injectsByModule.get(module)!.push({ injectedName, exportName, importName });
+	}
+
 	build.initialOptions.inject = [
 		...(build.initialOptions.inject ?? []),
 		//convert unenv's inject keys to absolute specifiers of custom virtual modules that will be provided via a custom onLoad
-		...Object.keys(inject).map(
-			(globalName) => `${prefix}${encodeToLowerCase(globalName)}.js`
-		),
+		...[...moduleIdSpecifier.keys()].map((id) => `${prefix}${id}`),
```
petebacondarwin, line 207: and

```suggestion
		...moduleIdSpecifier.keys(),
```

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -169,71 +169,61 @@ function handleNodeJSGlobals(
...
 	];
 
 	build.onResolve({ filter: UNENV_GLOBALS_RE }, ({ path }) => ({ path }));
 
 	build.onLoad({ filter: UNENV_GLOBALS_RE }, ({ path }) => {
 		// eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
-		const globalName = decodeFromLowerCase(path.match(UNENV_GLOBALS_RE)![1]);
-		const { importStatement, exportName } = getGlobalInject(inject[globalName]);
+		const module = moduleIdSpecifier.get(path.match(UNENV_GLOBALS_RE)![1])!;
```
petebacondarwin, line 214: and finally:

```suggestion
		const module = moduleIdSpecifier.get(path);
		assert(module, `Expected ${path} to be mapped to a module specifier`);
```

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -169,71 +169,61 @@ function handleNodeJSGlobals(
...
+		// eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
+		const injects = injectsByModule.get(module)!;
```
petebacondarwin, line 216: In general I prefer to use runtime asserts rather than `!` assertions since it allows us to control the error shown to give better context:

```suggestion
		const injects = injectsByModule.get(module);
		assert(injects, `Expected ${module} to map to an inject`);
```

this also avoids the need to add eslint overrides.

petebacondarwin (approved):
LGTM with a few nits that you can take a call on.

petebacondarwin:
I wonder if we could build these helpers into our preset package somehow?
Then we could reuse it in the Vite plugin, etc.

vicb:
> I wonder if we could build these helpers into our preset package somehow? Then we could reuse it in the Vite plugin, etc.

Let's add polyfill (and other potential updates for unenv) and then we can think about sharing this code but it would be great.

vicb:
Thanks for the review @petebacondarwin.
I have integrated your feedback - it makes the code much simpler!
PTAL

PR #8197, opened 2025-02-19
create-cloudflare: add types/autocomplete for Python projects

dom96:
Adds some files to the Python hello world template to enable users to easily install packages for autocomplete/type info.

Screenshot 2025-02-19 at 17 35 47

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: tested manually
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: simple change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: documentation in the template's README.md file


packages/create-cloudflare/templates/hello-world/py/.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
hoodmane, line 1: What is this for?

hoodmane (approved):
Looks fine to me, thanks!

packages/create-cloudflare/templates/hello-world/py/.python-version
dom96: yep

dom96:
Looks like we need a review from @cloudflare/c3. Btw anyone have any ideas why the CI is failing for these changes?

dom96:
Fixed linter errors via `pnpm prettier . --write --ignore-unknown` (thanks @petebacondarwin for the help!)

packages/create-cloudflare/templates/hello-world/py/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+name = "<TBD>"
```
petebacondarwin, line 2: This `<TBD>` doesn't get updated by `create-cloudflare` automatically, unlike those in `wrangler.jsonc` for instance. We should fix that

petebacondarwin (changes requested):
I created a new python hello world app using the prerelease.

```
npx https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/13587583761/npm-package-create-cloudflare-8197 --no-auto-update
```

Then I fixed up the pyproject.toml and uv.lock files to remove the `<TBD>` (which needs fixing).

Then I ran

```
uv venv
uv sync
```

(the `.venv` directory probably needs adding to the `.gitignore` right?)

But I still errors in VS Code...

image

packages/create-cloudflare/templates/hello-world/py/pyproject.toml
petebacondarwin: They do now!

petebacondarwin (approved):

dom96:
Thank you for your help on this @petebacondarwin!

PR #8200, opened 2025-02-19
Remove Miniflare 2

penalosa:
Remove the last vestiges of Miniflare v2 from `workers-sdk` (by way of `pages-shared`). In particular, this PR:
- Directly depends on `html-rewriter-wasm` rather than the Miniflare wrapper
- Migrates the `@cloudflare/pages-shared` package to run tests via `@cloudflare/vitest-pool-workers` 

cc @mrbbot —the end of an era!

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


DaniFoldi:
@penalosa does this mean that the repo will also be archived?

packages/pages-shared/environment-polyfills/miniflare.ts
```diff
@@ -1,7 +1,7 @@
 import { polyfill } from ".";
 
 export default async () => {
-	const { HTMLRewriter } = await import("@miniflare/html-rewriter");
+	const { HTMLRewriter } = await import("html-rewriter-wasm");
```
mrbbot, line 4: Unless things have changed, you'll need to be careful with this. `html-rewriter-wasm` has a slightly different API and semantics that isn't compatible with the Workers `HTMLRewriter`. You should be able to inline the code from `@miniflare/html-rewriter` though, there's not that much. 👍

penalosa: https://github.com/cloudflare/workers-sdk/pull/8200/commits/d270a805ccfe8cafdfe6f975ef4317be93c7eada

packages/pages-shared/package.json
```diff
@@ -22,15 +22,13 @@
...
 		"@cloudflare/workers-tsconfig": "workspace:*",
 		"@cloudflare/workers-types": "^4.20250214.0",
-		"@miniflare/cache": "^2.14.4",
-		"@miniflare/core": "^2.14.4",
-		"@miniflare/html-rewriter": "^2.14.4",
-		"@miniflare/storage-memory": "^2.14.4",
 		"concurrently": "^8.2.2",
 		"glob": "^10.4.5",
-		"vitest": "~2.1.0"
+		"html-rewriter-wasm": "^0.4.1",
```
petebacondarwin, line 30: This can be a dev dependency because the WASM will be bundled into the Worker, right?

petebacondarwin (approved):

packages/pages-shared/package.json
penalosa: I might be wrong, but I don't think this code is executed as a Worker in workers-sdk—Wrangler runs this code in a node context. Internally I think it's a worker, but since this is roughly the same as the `@miniflare/html-rewriter` setup I think it _should_ be fine.

petebacondarwin: Oh I see. These polyfills are only loaded into Wrangler's Miniflare setup.

petebacondarwin:
Happy to merge once @GregBrimble has taken a look from a production deployment point of view.

packages/pages-shared/vitest.config.mts
```diff
@@ -1,12 +1,15 @@
...
-import configShared from "../../vitest.shared";
+import { defineWorkersProject } from "@cloudflare/vitest-pool-workers/config";
 
-export default mergeConfig(
-	configShared,
-	defineProject({
-		test: {
-			include: ["**/__tests__/**/*.{test,spec}.{ts,js,tsx,jsx}"],
-			setupFiles: "./vite.setup.ts",
+export default defineWorkersProject({
```
GregBrimble, line 3: Does this make the tests run in workerd, even if it's not a real Worker? If yes, is HTMLRewriter not just available as a global there? Do we even still need a polyfill?

GregBrimble (approved):
Approved with a question and the suspicion that this could be made even simpler now. But if you just want to ship what you've got, that's cool with me.

No prod changes — this is just a test refactoring.

packages/pages-shared/vitest.config.mts
penalosa: The polyfill is for the wrangler dev flow, which runs this in node

GregBrimble: Makes sense!

petebacondarwin: Because this is Pages and not Workers+Assets?

PR #8201, opened 2025-02-19
fix: interactively list Cloudchamber deployments using labels

ichernetsky-cf:
The `--label` arguments were ignored when listing deployments interactively.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no test coverage of interactive flow
  - [ ] I don't know
  - [x] Required (triggered action https://github.com/cloudflare/workers-sdk/actions/runs/13572114935)
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not a documented aspect of the feature

### DIFF
--- a/packages/wrangler/src/cloudchamber/cli/deployments.ts
+++ b/packages/wrangler/src/cloudchamber/cli/deployments.ts
@@ -77,6 +77,7 @@ export async function loadDeployments(
 		image?: string;
 		state?: string;
 		ipv4?: string;
+		labels?: string[];
 	}
 ): Promise<DeploymentV2[]> {
 	const { start, stop } = spinner();
@@ -87,7 +88,8 @@ export async function loadDeployments(
 			deploymentsParams?.location,
 			deploymentsParams?.image,
 			deploymentsParams?.state as DeploymentPlacementState | undefined,
-			deploymentsParams?.state
+			deploymentsParams?.state,
+			deploymentsParams?.labels
 		)
 	);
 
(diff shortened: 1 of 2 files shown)


skepticfx (approved):

packages/wrangler/src/cloudchamber/cli/deployments.ts
```diff
@@ -87,7 +88,8 @@ export async function loadDeployments(
 			deploymentsParams?.location,
 			deploymentsParams?.image,
 			deploymentsParams?.state as DeploymentPlacementState | undefined,
-			deploymentsParams?.state
+			deploymentsParams?.state,
+			deploymentsParams?.labels
```
CarmenPopoviciu, line 92: is this codepath not covered by any tests?

ichernetsky-cf: No, I believe the interactive flows are not covered by any of the existing tests.

petebacondarwin (approved):
I see that in the tests there was already a TODO to add interactive tests.
See https://github.com/cloudflare/workers-sdk/blob/26a66d76d99f5a308a0bdbb328d5254aa7fab4f5/packages/wrangler/src/__tests__/cloudchamber/list.test.ts#L70-L71
Given the lack of any tests, and that this is a pretty small addition which matches the typings of the API call, I think we can land this.

PR #8202, opened 2025-02-20
fix: make services.environment config work with wrangler dev

ka2n:
Fix an trivial issue where environment settings in the services array of wrangler.jsonc were not being properly applied when running `wrangler dev` command.

---

- Tests
  - [x] TODO (before merge)
  - [ ] Tests included
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/wrangler/src/dev-registry/index.ts
+++ b/packages/wrangler/src/dev-registry/index.ts
@@ -69,7 +69,7 @@ export async function getBoundRegisteredWorkers(
 	existingWorkerDefinitions?: WorkerRegistry | undefined
 ): Promise<WorkerRegistry | undefined> {
 	const serviceNames = (services || []).map(
-		(serviceBinding) => serviceBinding.service
+		(serviceBinding) => serviceBinding.environment ? `${serviceBinding.service}-${serviceBinding.environment}` : serviceBinding.service
 	);
 	const durableObjectServices = (
 		durableObjects || { bindings: [] }


petebacondarwin:
Thanks for putting up this PR @ka2n. I can see that this would be a problem.

We will need a test to prove the fix, and also a patch changeset describing the change.
Would you be up for providing that? We could add an environment test to this set of e2e tests: https://github.com/cloudflare/workers-sdk/blob/c63f1b0790d7487074152c958ad10a910d4eae34/packages/wrangler/e2e/dev-registry.test.ts

I am worried that even this fix is a little too simplistic, since it is possible that a Worker can provide an explicit name for an environment, which means that even the approach here (of concatenating the environment name to the worker name, which is the default behaviour) might fail.

I agree that this is better than what we have right now. But really I think might need to do something more complex to cover all cases in the long run.

penalosa:
I'm not sure this actually works as you'd expect. The `environment` field in a service binding refers to the legacy Service Environments feature, which is very deprecated. Instead, to bind to e.g. the Wrangler environment "foo" on the service "bar", you should reference the worker by it's _deployed_ name. e.g.

```json
"services": [{"service": "bar-foo"}]
```

ka2n:
> The environment field in a service binding refers to the legacy Service Environments feature, which is deprecated

I wasn't aware that this field was for a legacy feature. That wasn't my intention, so I think I'll close this. Thanks for pointing it out.

PR #8205, opened 2025-02-20
Use today's date as the default starter playground compat date

penalosa:
As title—refactor the default starters for the playground to not bake in a compat date. Instead, inject the compat date when loading a starter, which ensures that:
- New users will always have today's compat date
- Playground links will continue to bake in their compat date, ensuring reproducibility

_How to test_

Go to the playground preview generated for this PR (https://4625f2e8.workers-playground.pages.dev/playground), and verify that newer runtime features work (e.g. `fetch()` with `{ cache: "no-store" }`.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: manually tested
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: doesn't affect Wrangler
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/workers-playground/src/QuickEditor/defaultHashes.ts
```diff
@@ -1,5 +1,14 @@
...
+	"/python": {
```
CarmenPopoviciu, line 2: are these generated somehow?

penalosa: Yes, using `pnpm generate:default-hashes`

CarmenPopoviciu:
I tested as per description wit a Worker script that contained 
```
const resp = new Request(request, { cache: 'no-store' })
console.log(`cache: no-store response: ${await resp.body}`)
```
and can confirm that while in prod I'm seeing 
Screenshot 2025-02-20 at 16 27 23

in the playground preview there are no similar errors
Screenshot 2025-02-20 at 16 28 45

CarmenPopoviciu:
@penalosa might totally not be trivial, but is there no way to test this behaviour via an automated test? Otherwise LGTM ♥️

irvinebroque (commented):
🙌

CarmenPopoviciu (approved):

PR #8206, opened 2025-02-20
Add support for binding to named entrypoints in the same Worker

jamesopstad:
Fixes #000

This adds support for binding to named entrypoints in the same Worker.

Additionally, this adds `cloudflare:workers` imports to the `node-compat/worker-basic` playground so that we have tests to confirm that Cloudflare builtins can be used when `nodejs_compat` is enabled.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/bitter-teeth-think.md
+++ b/.changeset/bitter-teeth-think.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+Add support for binding to named entrypoints in the same worker
--- a/packages/vite-plugin-cloudflare/playground/node-compat/worker-basic/index.ts
+++ b/packages/vite-plugin-cloudflare/playground/node-compat/worker-basic/index.ts
@@ -2,6 +2,11 @@ import assert from "node:assert/strict";
 import { join } from "path";
 // Check that we can actually import unenv polyfilled modules in user source.
 import "perf_hooks";
+// Check that `cloudflare:worker` imports work when `nodejs_compat` is enabled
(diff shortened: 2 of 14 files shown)


.changeset/bitter-teeth-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
```
petebacondarwin, line 2: I think this should be minor. It's a feature, not a bug fix, right?

.changeset/bitter-teeth-think.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add support for binding to named entrypoints in the same worker
```
petebacondarwin, line 5: Do we also need a patch changeset for fixing the bug around `cloudflare:workers` not being treated as a builtin?

.changeset/bitter-teeth-think.md
jamesopstad: This is what was agreed in the meeting yesterday for beta and `0.x.x` packages.

> Minor for breaking changes, patch for bug fixes and new backwards-compatible features

.changeset/bitter-teeth-think.md
jamesopstad: This PR doesn't fix that. I have another one incoming.

petebacondarwin (approved):

PR #8209, opened 2025-02-20
Fix bug with usage of Cloudflare builtins in dependencies

jamesopstad:
Fixes #000.

Fixes bug with usage of Cloudflare builtins in dependencies. These are now externalized during dependency optimization.

I have also added a PartyServer playground so that we have test coverage for this scenario.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/big-pillows-cough.md
+++ b/.changeset/big-pillows-cough.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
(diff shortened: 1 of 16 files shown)


dario-piotrowicz (approved):
looks good to me 🙂

Issue #8187, opened 2025-02-19
[c3] can't ctrl-c out of waiting for new zone propagation


lrapoport-cf:
see https://x.com/theo/status/1890634143085605199

CarmenPopoviciu:
I have been trying to reproduce the behaviour via `npm run create cloudflare@latest`, with using both the `Hello World Worker` and `Next.js` starter templates. I am seeing the following:

- for the `Hello World Worker` starter template, the DNS propagation takes on average a bit over `10s` (`11s` in this particular screenshot, though I was not able to capture it more accurately)

<img width="476" alt="Image" src="https://github.com/user-attachments/assets/6e11c9d7-4244-4cc2-a5cb-dc3af31c67e1" />

- for the `Next.js` starter template the DNS propagation times average at  > `200s` (we time out at `300s`). This gets rather confusing for users I would imagine, since I'm not sure if they know what sort of propagation times to expect. This is something we can very easily action on by changing the copy to inform users that propagation times might take up to 2mins

<img width="742" alt="Image" src="https://github.com/user-attachments/assets/500644ac-d34d-4d4a-928e-eb22147faf55" />

- in terms of exiting this process, I have tried to `ctrl+C` at various points in time while waiting for DNS to propagate and for both starter projects I am experiencing the same: first `ctrl+C` does register the key press (`^C` in stdout) but does **not** exit the process (this is a bug and should be fixed), second `ctrl+C` does register the key press and **does** exit the process. I am seeing this behaviour consistently, over the > 10 projects I have tested with, and did not have to, at any point during my testing, have to manually kill the process

<img width="679" alt="Image" src="https://github.com/user-attachments/assets/2fc0d1b3-24ec-4524-8922-66c2738a9776" />

To tldr there are 2 concrete things I can action on here based on my findings:

1. make the copy adjustment to inform users about DNS propagation times being abt 2 mins
2. figure out why first `ctrl+C` does not exit the process and fix that

PR #8215, opened 2025-02-21
fix(create-cloudflare): make DNS poling copy more user friendly

CarmenPopoviciu:
This change makes the user facing message C3 displays while waiting for DNS propagation, more friendly/informative. The idea is to inform users that DNS propagation might sometimes take even up to 2 minutes. This will hopefully prevent confusion around whether how long the process will take, or whether the process is stuck, etc.

See https://github.com/cloudflare/workers-sdk/issues/8187

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: just copy change. Also the fn that contains this message is not covered by tests
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: just copy change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: just copy change

### DIFF
--- a/packages/create-cloudflare/src/helpers/poll.ts
+++ b/packages/create-cloudflare/src/helpers/poll.ts
@@ -22,7 +22,7 @@ export const poll = async (url: string): Promise<boolean> => {
 	const domain = new URL(url).host;
 	const s = spinner();
 
-	s.start("Waiting for DNS to propagate. This might take up to two minutes.");
+	s.start("Waiting for DNS to propagate. This might take a few minutes.");
 
 	// Start out by sleeping for 10 seconds since it's unlikely DNS changes will
 	// have propogated before then
--- a/packages/create-cloudflare/src/helpers/poll.ts
+++ b/packages/create-cloudflare/src/helpers/poll.ts
@@ -22,7 +22,7 @@ export const poll = async (url: string): Promise<boolean> => {
 	const domain = new URL(url).host;
 	const s = spinner();
 
-	s.start("Waiting for DNS to propagate");
+	s.start("Waiting for DNS to propagate. This might take up to two minutes.");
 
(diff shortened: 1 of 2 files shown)


packages/create-cloudflare/src/helpers/poll.ts
```diff
@@ -22,7 +22,7 @@ export const poll = async (url: string): Promise<boolean> => {
 	const domain = new URL(url).host;
 	const s = spinner();
 
-	s.start("Waiting for DNS to propagate");
+	s.start("Waiting for DNS to propagate. This might take up to two minutes.");
```
CarmenPopoviciu, line 25: this will be mostly true for Pages but I'm not sure if it's worth calling that out. IMHO it'll just make the copy more noisy

CarmenPopoviciu, line 25:
```suggestion
	s.start("Waiting for DNS to propagate. This might take a few minutes.");
```

packages/create-cloudflare/src/helpers/poll.ts
```diff
@@ -47,7 +47,9 @@ const pollDns = async (
 	s: ReturnType<typeof spinner>,
 ) => {
 	while (Date.now() - start < TIMEOUT) {
-		s.update(`Waiting for DNS to propagate (${secondsSince(start)}s)`);
+		s.update(
+			`Waiting for DNS to propagate. This might take up to two minutes. (${secondsSince(start)}s)`,
```
CarmenPopoviciu, line 51:
```suggestion
			`Waiting for DNS to propagate. This might take a few minutes. (${secondsSince(start)}s)`,
```

edmundhung (approved):

PR #8220, opened 2025-02-21
Stop cloudchamber build from trying to push image when build fails

IRCody:
_Describe your change..._
Fixes issue where cloudchamber build --push ... would continue trying to push the image even if the build failed. Adds a check to verify the passed in context dir is actually a directory and also checks the builds error code on exit to stop progressing to pushing if that fails.

The behavior now is:

With a file that doesn't exist:
```
% wrangler cloudchamber build -t test:1 --push ./noexist
 Error when checking ./noexist: Error: ENOENT: no such file or directory, stat './noexist'   
```

When given a file and not a directory:
```
% wrangler cloudchamber build -t test:1 --push ./Dockerfile
PATH must be a directory
```

When docker build  returns a non-zero exit code:
```
% wrangler cloudchamber build -t test:1 --push ./DockerfileDir
.... <Docker build output including dockers failure message>

Error: Build exited with code: 1
```

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Would probably belong in e2e tests which cloudchamber doesn't have yet.
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Cloudchamber has no e2e tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Cloudchamber is not documented externally

### DIFF
--- a/packages/wrangler/src/cloudchamber/build.ts
+++ b/packages/wrangler/src/cloudchamber/build.ts
@@ -69,9 +69,7 @@ export async function constructBuildCommand(options: {
 }
 
 // Function for building
-export async function dockerBuild(options: {
-	buildCmd: string;
-}): Promise<void> {
+export function dockerBuild(options: { buildCmd: string }): Promise<void> {
 	return new Promise((resolve, reject) => {
 		const buildCmd = options.buildCmd.split(" ").slice(1);
 		const buildExec = options.buildCmd.split(" ").shift();
@@ -190,7 +188,7 @@ export async function buildCommand(
 			pathToDockerfile: args.PATH,
 			pathToDocker: args.pathToDocker,
 		})
-			.then(async (bc) => await dockerBuild({ buildCmd: bc }))
+			.then((bc) => dockerBuild({ buildCmd: bc }))
 			.then(async () => {
 				if (args.push) {
 					await dockerLoginManagedRegistry({


packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -160,17 +164,33 @@ export function pushYargs(yargs: CommonYargsArgvJSON) {
...
+		logRaw(`Error when checking ${args.PATH}: ${error}`);
+		return;
+	}
+
 	try {
 		await constructBuildCommand({
 			imageTag: args.tag,
 			pathToDockerfile: args.PATH,
 			pathToDocker: args.pathToDocker,
 		})
```
gabivlj, line 192: maybe `.then(bc => dockerBuild({ buildCmd: bc })`?

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -68,17 +69,20 @@ export async function constructBuildCommand(options: {
...
-	const buildExec = options.buildCmd.split(" ").shift();
-	const child = spawn(String(buildExec), buildCmd, { stdio: "inherit" }).on(
-		"error",
-		(err) => {
-			throw err;
-		}
-	);
-	await new Promise((resolve) => {
-		child.on("close", resolve);
+export async function dockerBuild(options: {
```
gabivlj, line 72: nit: I think we shouldnt mark the function async as we are already returning a promise and we dont call await

skepticfx (approved):

emily-shen (approved):

PR #8231, opened 2025-02-23
fix: use ESM WebSocketServer import to avoid crashing vite dev

### LINKED ISSUE #8222: 🐛 BUG: Vite plugin dev-server fails to start in some ESM environments

Vite Plugin

0.1.1

Linux

https://github.com/zebp/vite-test

### Describe the Bug

In some environments running under ESM, although I'm not sure why it's not always the case, the Vite dev server fails to start because `ws`'s default export doesn't have a `Server` property. This is a [known issue](https://github.com/websockets/ws/issues/1538#issuecomment-944859160) with the `ws` package.

```
>  vite-test at main ✓ npm run dev                                                                                              mirage

> vite-test@1.0.0 dev
> vite

The latest compatibility date supported by the installed Cloudflare Workers Runtime is "2025-02-04",
but you've requested "2025-02-14". Falling back to "2025-02-04"...
error when starting dev server:
TypeError: ws.Server is not a constructor
    at handleWebSocket (file:///tmp/vite-test/node_modules/@cloudflare/vite-plugin/dist/index.js:9111:25)
    at configureServer (file:///tmp/vite-test/node_modules/@cloudflare/vite-plugin/dist/index.js:9354:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async _createServer (file:///tmp/vite-test/node_modules/vite/dist/node/chunks/dep-Cg8OuIew.js:45017:20)
    at async CAC.<anonymous> (file:///tmp/vite-test/node_modules/vite/dist/node/cli.js:750:20)
```

petebacondarwin:
Fixes #8222

It appears that if there are multiple versions of the `ws` package in a user's project
then the Node.js resolution picks up the ESM "import" package export rather than the "require" package export.
This results in the entry-point having different JS exports:
In particular the default export no longer contains a `Server` property; instead on must import the `WebSocketServer` named JS export.
While it is not clear why the Node.js behaviour changes in this way, the cleanest fix is to import the `WebSocketServer` directly.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included (the vite-plugin e2e test caught this)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not Wrangler related
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: just a fix


.changeset/green-hornets-camp.md
```diff
@@ -0,0 +1,11 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+fix: use ESM WebSocketServer import to avoid crashing vite dev
+
+It appears that if there are multiple versions of the `ws` package in a user's project
+then the Node.js resolution picks up the ESM "import" package export rather than the "require" package export.
+This results in the entry-point having different JS exports:
+In particular the default export no longer contains a `Server` property; instead on must import the `WebSocketServer` named JS export.
```
jamesopstad, line 10:
```suggestion
In particular the default export no longer contains a `Server` property; instead one must import the `WebSocketServer` named JS export.
```

packages/vite-plugin-cloudflare/e2e/fixtures/basic/package.json
```diff
@@ -14,7 +14,7 @@
 		"react-dom": "^19.0.0"
 	},
 	"devDependencies": {
-		"@cloudflare/vite-plugin": "https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/13293829928/npm-package-cloudflare-vite-plugin-8118",
+		"@cloudflare/vite-plugin": "0.x",
```
jamesopstad, line 17: Should this be the workspace package version? I'm not sure exactly how this works with the custom registry. Does this correctly test against code changes in a PR?

packages/vite-plugin-cloudflare/src/websockets.ts
```diff
@@ -1,4 +1,4 @@
-import ws from "ws";
+import * as ws from "ws";
```
jamesopstad, line 1: I think importing directly is clearer as we're only using one export.

```suggestion
import { WebSocketServer } from "ws";
```

packages/vite-plugin-cloudflare/src/websockets.ts
```diff
@@ -15,7 +15,7 @@ export function handleWebSocket(
 	fetcher: ReplaceWorkersTypes<Fetcher>["fetch"],
 	logger: vite.Logger
 ) {
-	const nodeWebSocket = new ws.Server({ noServer: true });
+	const nodeWebSocket = new ws.WebSocketServer({ noServer: true });
```
jamesopstad, line 18:
```suggestion
	const nodeWebSocket = new WebSocketServer({ noServer: true });
```

jamesopstad (approved):
Approving but I think it would also be good to set the the `ws` version in pnpm workspaces and then use that version in Miniflare and the plugin. As the plugin will always install Miniflare, it would be good to enforce the same version.

petebacondarwin:
> Approving but I think it would also be good to set the the `ws` version in pnpm workspaces and then use that version in Miniflare and the plugin. As the plugin will always install Miniflare, it would be good to enforce the same version.

Yes very good idea.

packages/vite-plugin-cloudflare/e2e/fixtures/basic/package.json
petebacondarwin: This fixture is copied to a temporary directory outside the workspace so we cannot use `workspace:*` type dependencies.
What will happen is that we will run `npm install` (or equivalent) in the temporary directory but the verdaccio npm registry will intercept the requests and try to handle them itself before falling back to the public registry.
The idea he was to try to get the most recent vite-plugin package, which should be the one we published to the verdaccio registry.

jamesopstad: In that case, shouldn't the version be `*`? Want to make sure it's not something we might forget to update when we move to 1.0.

petebacondarwin: I did try * previously and it failed. But just tried again now and it seems better.

PR #8232, opened 2025-02-24
fix: remove extra config files that are generated by c3 with `--existing-script `

### LINKED ISSUE #8188: 🐛 BUG: c3 and wrangler init from dash broken with duplicate config files
To replicate, create a worker on the dashboard, then run 
`npx create-cloudflare@latest <worker-name> --existing-script <worker-name>` or
`npx wrangler init --from-dash <worker-name>`

### Discussed in https://github.com/cloudflare/workers-sdk/discussions/8183

<div type='discussions-op-text'>

<sup>Originally posted by **qwadrox** February 18, 2025</sup>
After running this which I got from the dashboard 

```console
npx create-cloudflare@latest trigger-cms-build --existing-script trigger-cms-build
```

I see a confusing output like this:

image

For me the 3 wrangler file ambiguous. Which one I should extend with my own configuration? 
Which package.json file should contain the dependencies? 

using 

```console
npm create cloudflare@latest
```

the project structure is more flat and clean

image

How it differs from the more complex structure? </div>

emily-shen:
Fixes #8188

This was generating broken projects due to the fact there were three (!) separate config files.

Screenshot 2025-02-24 at 10 54 25

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: i don't really know how to test this, but i have tried it manually though?
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/.changeset/sixty-spoons-unite.md
+++ b/.changeset/sixty-spoons-unite.md
@@ -2,4 +2,6 @@
 "create-cloudflare": patch
 ---
 
-fix: remove duplicate config files that are created when using --existing-script
+fix: stop c3 adding a duplicate `./wrangler.jsonc` when using --existing-script
+
+This should mean dev and deploy on projects initialised using `create-cloudflare --existing-script` start working again. Note there will still be an extraneous `./src/wrangler.toml`, which will require a separate fix in the dashboard. This file can be manually deleted in the meantime.
--- a/.changeset/sixty-spoons-unite.md
+++ b/.changeset/sixty-spoons-unite.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+fix: remove duplicate config files that are created when using --existing-script
--- a/packages/create-cloudflare/templates/pre-existing/c3.ts
+++ b/packages/create-cloudflare/templates/pre-existing/c3.ts
@@ -1,11 +1,9 @@
-import { existsSync } from "fs";
 import { cp, mkdtemp } from "fs/promises";
 import { tmpdir } from "os";
 import { join } from "path";
 import { brandColor, dim } from "@cloudflare/cli/colors";
 import { processArgument } from "helpers/args";
 import { runCommand } from "helpers/command";
-import { removeFile } from "helpers/files";
 import { detectPackageManager } from "helpers/packageManagers";
 import { chooseAccount, wranglerLogin } from "../../src/wrangler/accounts";
 import type { C3Context } from "types";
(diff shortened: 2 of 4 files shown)


packages/create-cloudflare/templates/pre-existing/c3.ts
```diff
@@ -49,14 +51,22 @@ export async function copyExistingWorkerFiles(ctx: C3Context) {
 		},
 	);
 
+	// dash hello-world template generates unused files:
```
CarmenPopoviciu, line 54: I have 0 context here, but can't we fix the dash template to not generate unused files?

emily-shen: Yes WP-1474 is going to remove them. This PR doesn't strictly need to remove those files, because the wrangler.jsonc was the only file that was stopping it from deploying + running, but it _is_ still very confusing for a user to have 2 wrangler.tomls.
So i thought as a stop gap we could remove it in C3, and once WP-1474 is through then we could just remove this delete.

CarmenPopoviciu: I totally agree the current user experience is ... not ideal ... and that we should fix asap ... but IMHO this is not the way to solve it. AFAICT that WP- issue might actually be handled today. If that is the case, we should wait for that. Or...if this is so urgent it can't wait, either ping the responsible team to highlight that in case they are not aware, or just fix it ourselves in Dash, assuming this lives in Dash, but it prolly does

.changeset/sixty-spoons-unite.md
```diff
@@ -0,0 +1,7 @@
+---
+"create-cloudflare": patch
+---
+
+fix: stop c3 adding a duplicate `./wrangler.jsonc` when using --existing-script
```
CarmenPopoviciu, line 5: eternal love and kudos goes to @emily-shen ♥️

CarmenPopoviciu (approved):
thank you Emily 🙏 🚀 ♥️

penalosa:
Could we add an E2E test for this? Something like:
- deploying a worker
- initialising it with C3 `--existing-script`
- running `wrangler dev` to verify it works?

emily-shen:
> Could we add an E2E test for this? Something like:

[6c0284d](https://github.com/cloudflare/workers-sdk/pull/8232/commits/6c0284d7229b5793347cab215c574c742100247e)

penalosa:
Not to pile more on here, but could we make this generate jsonc instead of toml? Not a blocker for this PR though

PR #8243, opened 2025-02-24
fix(create-cloudflare): Ensure we exit the process on "SIGINT" and "SIGTERM"

CarmenPopoviciu:
Currently C3 does not explicitly exit the process if an error is thrown, or if a "SIGINT" or "SIGTERM" signal is received. This leads to situations when, if `ctrl+C` is pressed while there are still tasks in the stack/microtask queues (think in flight async xhr calls, or polling, or long running
`while` loops), the current process will continue running until all those tasks are run to completion, and the queues are empty.

This commit fixes this by explicitly calling `process.exit()` when an error is thrown (our internal "SIGINT"/"SIGTERM" handlers will throw a `CancelError`), thus ensuring we always exit the process.

Fixes https://github.com/cloudflare/workers-sdk/issues/8187

---

- Tests
  - [x] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not covered by e2e tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/create-cloudflare/src/cli.ts
+++ b/packages/create-cloudflare/src/cli.ts
@@ -201,4 +201,9 @@ main(process.argv)
 	})
 	.finally(async () => {
 		await reporter.waitForAllEventsSettled();
+
+		// ensure we explicitly exit the process, otherwise any ongoing async
+		// calls or leftover tasks in the stack queue will keep running until
+		// completed
+		process.exit();
 	});
--- a/.changeset/ripe-socks-stay.md
+++ b/.changeset/ripe-socks-stay.md
@@ -1,0 +1,10 @@
+---
+"create-cloudflare": patch
+---
+
+fix: Ensure we exit the process on "SIGINT" and "SIGTERM"
+
+Currently C3 does not explicitly exit the process if an error is thrown, or if a "SIGINT" or "SIGTERM" signal is received. This leads to situations when, if `ctrl+C` is pressed while there are still tasks in the stack/microtask queues (think in flight async xhr calls, or polling, or long running
+`while` loops), the current process will continue running until all those tasks are run to completion, and the queues are empty.
+
+This commit fixes this by explicitly calling `process.exit()` when an error is thrown (our internal "SIGINT"/"SIGTERM" handlers will throw a `CancelError`), thus ensuring we always exit the process.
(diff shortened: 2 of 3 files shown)


packages/create-cloudflare/src/cli.ts
```diff
@@ -201,4 +201,6 @@ main(process.argv)
 	})
 	.finally(async () => {
 		await reporter.waitForAllEventsSettled();
+		// await setTimeout(100);
+		process.exit();
```
CarmenPopoviciu, line 205: my question here is ... do we **always** want to exit the process? or should we restrict this to sigint and sigterm signals only?

edmundhung: I think it is totally fine to call `process.exit()` all the time. We were calling it everywhere before we added telemetry... 😅

edmundhung: Wrangler is doing it all the time: https://github.com/cloudflare/workers-sdk/blob/47f56a2a4dcfa216d768fec974c2fa8c5bb3c21c/packages/wrangler/src/cli.ts#L20-L26

edmundhung (approved):

petebacondarwin (approved):
Feels a bit worrying that we have to do this at all... it would be good to understand what child processes are hanging and have a way to kill those directly, rather than hard killing the parent.
But in the interests of moving forward and not blocking this, happy to approve for now.

CarmenPopoviciu:
> Feels a bit worrying that we have to do this at all... it would be good to understand what child processes are hanging and have a way to kill those directly, rather than hard killing the parent. But in the interests of moving forward and not blocking this, happy to approve for now.

It can really depend. In the particular case I was trying to solve, [this](https://github.com/cloudflare/workers-sdk/blob/a9a4c33143b9f58673ac0cdd251957997275fa10/packages/create-cloudflare/src/helpers/poll.ts#L49-L59) `while` loop was running in the parent process AFAICT. Because collectAsyncMetrics [registers](https://github.com/cloudflare/workers-sdk/blob/a9a4c33143b9f58673ac0cdd251957997275fa10/packages/create-cloudflare/src/metrics.ts#L204) "SIGINT"/"SIGTERM" listeners, nodejs' default behaviour to exit the process is overriden, at least that's how I understand the docs on this 👇 

> 'SIGTERM' and 'SIGINT' have default handlers on non-Windows platforms that reset the terminal mode before exiting with > code 128 + signal number. If one of these signals has a listener installed, its default behavior will be removed (Node.js will > no longer exit). 

This means if we want to exit the process we'll need to explicitly call that...which is what this PR did.

PR #8247, opened 2025-02-25
Prepare to omit Content-Type header for files of unknown extension in Workers Assets

GregBrimble:
Fixes WC-3028.

This isn't quite everything. This makes it work in `wrangler dev` and prepares the asset-worker in production to be able to handle missing contentType metadata. We'll need to land this, then API support, and then follow up with the final piece here.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not live yet

### DIFF
--- a/fixtures/workers-with-assets/public/totallyinvalidextension.greg
+++ b/fixtures/workers-with-assets/public/totallyinvalidextension.greg
@@ -1,0 +1,1 @@
+I'm a narcissist.
--- a/packages/wrangler/src/assets.ts
+++ b/packages/wrangler/src/assets.ts
@@ -138,7 +138,7 @@ export const syncAssets = async (
 						[(await readFile(absFilePath)).toString("base64")],
 						manifestEntry[1].hash,
 						{
-							type: getContentType(absFilePath),
+							type: getContentType(absFilePath) || "application/octet-stream",
 						}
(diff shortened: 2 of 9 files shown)


fixtures/workers-with-assets/public/totallyinvalidextension.greg
```diff
@@ -0,0 +1 @@
+I'm a narcissist.
```
CarmenPopoviciu, line 1: 😅

CarmenPopoviciu (approved):

packages/wrangler/src/assets.ts
```diff
@@ -138,7 +138,7 @@ export const syncAssets = async (
 						[(await readFile(absFilePath)).toString("base64")],
 						manifestEntry[1].hash,
 						{
-							type: getContentType(absFilePath),
+							type: getContentType(absFilePath) || "application/octet-stream",
```
petebacondarwin, line 141: For future readers, this is the line that allows us to gradually roll this feature out before production support is in place.

PR #8253, opened 2025-02-25
feat(wrangler): support `--cwd` cli argument

CarmenPopoviciu:
Original PR is here https://github.com/cloudflare/workers-sdk/pull/7994. This PR was created from a fork, but required a couple of adjustments.

I tried creating a PR with the required changes to the fork, but it didn't get reviewed. Opening a new PR instead so we can get this over the finish line

---- from original PR

This PR adds `-C, --cwd` global argument to the `wrangler` CLI to allow changing the current working directory before running any command.

This feature is used in most of CLI tools that depend on right working directory but different names. I have chosen `-C, --cwd` following with [pnpm](https://pnpm.io/pnpm-cli#-c-path---dir-path) (`-C`) + [yarn](https://classic.yarnpkg.com/lang/en/docs/cli/#toc-cwd) (`--cwd`), but open to ideas changing it to anything else.

This allows better DX of using `wrangler` commands from any working directory  to another (where project is).

```
$ wrangler --help

GLOBAL FLAGS
  -c, --config   Path to Wrangler configuration file  [string]
  -C, --cwd      Run as if wrangler was started in <path> instead of the current working directory  [string]
  -e, --env      Environment to use for operations, and for selecting .env and .dev.vars files  [string]
  -h, --help     Show help  [boolean]
  -v, --version  Show version number  [boolean]
```

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/19665

### DIFF
--- a/.changeset/khaki-years-act.md
+++ b/.changeset/khaki-years-act.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Add `--cwd` global argument to the `wrangler` CLI to allow changing the current working directory before running any command.
--- a/packages/wrangler/src/__tests__/ai.test.ts
+++ b/packages/wrangler/src/__tests__/ai.test.ts
@@ -27,6 +27,7 @@ describe("ai help", () => {
 
 			GLOBAL FLAGS
 			  -c, --config   Path to Wrangler configuration file  [string]
+			      --cwd      Run as if Wrangler was started in the specified directory instead of the current working directory  [string]
 			  -e, --env      Environment to use for operations, and for selecting .env and .dev.vars files  [string]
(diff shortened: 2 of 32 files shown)


petebacondarwin:
> on a second thought...I don't think we can support this for Pages, because I **think** Pages assumed the process is always running at the root? I need to double check this...

Why not support this for Pages? It is no different to cd-ing into that directory and running `wrangler pages ...`, right?

PR #8258, opened 2025-02-25
add mysql options to hyperdrive

knickish:
Part of RM-17419

Adds mysql-related options to the wrangler hyperdrive CLI

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Feature not publicly available yet


packages/wrangler/src/hyperdrive/index.ts
```diff
@@ -159,15 +159,21 @@ export function getOriginFromArgs<
...
 			);
-		} else if (url.protocol !== "postgresql:" && url.protocol !== "postgres:") {
+		} else if (
+			url.protocol !== "postgresql:" &&
+			url.protocol !== "postgres:" &&
+			url.protocol !== "mysql:"
+		) {
 			throw new UserError(
-				"Only PostgreSQL or PostgreSQL compatible databases are currently supported."
+				"Only PostgreSQL or MySQL compatible databases are currently supported."
```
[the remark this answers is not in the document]
knickish: done

.changeset/new-cooks-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+add mysql configuration options to hyperdrive
```
mtlemilio, line 5: Could be more descriptive here. These end up in the release notes. Something like "Enable the creation of MySQL Hypedrive configs via the Wrangler CLI."

packages/wrangler/src/hyperdrive/index.ts
```diff
@@ -159,15 +159,21 @@ export function getOriginFromArgs<
...
 		}
 
 		if (url.protocol === "") {
 			throw new UserError(
-				"You must specify the database protocol - e.g. 'postgresql'."
+				"You must specify the database protocol - e.g. 'postgresql'/'mysql'."
 			);
-		} else if (url.protocol !== "postgresql:" && url.protocol !== "postgres:") {
+		} else if (
+			url.protocol !== "postgresql:" &&
```
mtlemilio, line 171: Could be nice to use `url.protocol.startsWith` to handle the colon and also consider normalizing the case here since we're displaying the schemes with the "correct" mixed case.

packages/wrangler/src/__tests__/hyperdrive.test.ts
```diff
@@ -402,7 +437,42 @@ describe("hyperdrive commands", () => {
...
+			    }
+			  ]
+			}"
+		`);
+	});
+
+	it("should create a hyperdrive config given individual params instead of a connection string", async () => {
+		const reqProm = mockHyperdriveCreate();
+		await runWrangler(
+			"hyperdrive create test123 --host=example.com --database=neondb --user=test --password=password --port=1234 --origin-scheme=mysql"
```
mtlemilio, line 457: The `origin-scheme` flag is currently set here: https://github.com/cloudflare/workers-sdk/blob/1320f77725b8ee4d2f8ead16287428026c047ae6/packages/wrangler/src/hyperdrive/create.ts#L19

I think it should be moved to the `upsertOptions` function in `index.ts` and be given an alias to just `scheme` for consistency. We could also consider removing the default.

knickish: It's already present in `upsertOptions` with the requested alias

packages/wrangler/src/hyperdrive/shared.ts
```diff
@@ -0,0 +1,12 @@
...
+	let ret = "";
+
+	if (scheme === "mysql") {
+		ret = "MySQL";
+	}
+	if (scheme === "postgres" || scheme === "postgresql") {
+		ret = "PostgreSQL";
+	}
+
+	return ret;
```
petebacondarwin, line 11: NIT this could simply be:

```suggestion
	switch (scheme) {
	  case "mysql":
		 return "MySQL";
	  case "postgres":
	  case "postgresql":
		  return "PostgreSQL";
		default:
		  return "";
	}
```

Ltadrian (commented):
LGTM - hyperdrive

packages/wrangler/src/hyperdrive/shared.ts
```diff
@@ -0,0 +1,12 @@
+export function capitalizeScheme(scheme: string | undefined) {
+	let ret = "";
```
petebacondarwin, line 2: If the scheme is unknown this will return `""`.  Is that what you want?
I would expect to either throw an error or to just return the `scheme` unaltered.

packages/wrangler/src/hyperdrive/index.ts
```diff
@@ -177,9 +185,9 @@ export function getOriginFromArgs<
 			throw new UserError(
 				"You must provide a port number - e.g. 'user:password@database.example.com:port/databasename"
 			);
-		} else if (url.pathname === "") {
+		} else if (!url.pathname || url.pathname === "") {
```
petebacondarwin, line 188: If the pathname is the empty string then `!url.pathname` is true.

```suggestion
		} else if (!url.pathname) {
```

petebacondarwin (approved):
Does the backend already support mysql? In other words if we land and release this change will things break if a user tries to connect to mysql?

While I have approve this (with some suggestions) we should not merge it until the API is ready.

ReppCodes:
> Does the backend already support mysql? In other words if we land and release this change will things break if a user tries to connect to mysql?
> 
> While I have approve this (with some suggestions) we should not merge it until the API is ready.

The control plane will intentionally return an error, it's currently gated to "@cloudflare.com" users while we're developing. It won't just choke, though.

packages/wrangler/src/hyperdrive/shared.ts
knickish: This is only used in displaying some information that should be non-critical, I can have it throw if that would wrangler conventions better, but I think an empty string is fine for this

mtlemilio (approved):

knickish:
This is the error message returned when attempting to create a mysql hyperdrive. This has been reviewed and approved by the team internally, and it is ready to be merged when possible

```bash
$ npx wrangler hyperdrive create hd-mysql --host hyperdrive-mysql-testing.. --port 3306 --database test_db --user test_user --password test_pw --scheme mysql

🚧 Creating 'hd-mysql'

✘ [ERROR] A request to the Cloudflare API (/accounts/f3f246136c89583f8ad438a7edb1cc60/hyperdrive/configs) failed.
  MySQL support is not yet publicly available [code: 2000] 
```

knickish:
This is the error message returned when attempting to create a mysql hyperdrive. This has been reviewed and approved by the team internally, and it is ready to be merged when possible

```bash
$ npx wrangler hyperdrive create hd-mysql --host hyperdrive-mysql-testing.. --port 3306 --database test_db --user test_user --password ... --scheme mysql

🚧 Creating 'hd-mysql'

✘ [ERROR] A request to the Cloudflare API (/accounts/.../hyperdrive/configs) failed.
  MySQL support is not yet publicly available [code: 2000] 
```

Issue #6948, opened 2024-10-11
🚀 Feature Request: in a mono repo, don't add prettierrc and editorconfig


rozenmd:
### Describe the solution

In a mono-repo (whic it seems like wrangler already can detect, since it's not adding a `.git` folder), the first thing I do after adding a new Worker (`npx wrangler@latest init <name> -y`) is delete the opinionated `.editorconfig` and `.prettierrc` files, since my repo already has them.

Separately: It's cool if you want to enforce certain coding styles on your team, let me enforce my own standards on my own team

PR #8263, opened 2025-02-26
Remove .prettierrc from templates generated by npm create cloudflare

irvinebroque:
refs https://github.com/cloudflare/workers-sdk/issues/6948

Removes `.prettierrc` from templates that `npm create cloudflare` generates.

Prettier is awesome, spaces are awesome, tabs are awesome — choose what you want to use, what is best for your project. Many people have their own preferences, their own `.prettierrc`, etc. — and this gets in the way of that when creating new projects.

- Tests
  - [x] TODO (before merge)
  - [x] Tests included
  - [x] Tests not necessary because: What to test here
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: What to document


Cherry (commented):
I would love to see Cloudflare lead the industry with accessible code style configurations by default, perhaps behind an option for if folks want it or not in c3's prompts, instead of removing this altogether.

For power users, they have their own styles and preferences, but for new users which is where I personally see most c3 usage, setting some great accessible defaults is an awesome thing. :orange_heart:

vicb:
-1 to remove this

PR #8264, opened 2025-02-26
fixes #8065

### LINKED ISSUE #8065: npm create cloudflare — make prompt about initializing a git repo unnecessary
### Describe the solution

When you run `npm create cloudflare`, we ask:

```
├ Do you want to use git for version control?
│ yes git
│
├ Initializing git repo
│ initialized git
```

But:

1. If I'm starting an entirely new project, I always want a git repository to be created
2. If I'm running `npm create cloudflare` from within an existing git repository, I never want a new git repository to be created (I'm adding this Worker to my existing repository)

Point is — we should figure out ^ what logic and checks are needed so that it is never necessary to ask someone this question when they run `npm create cloudflare`

dcartertwo:
Fixes https://github.com/cloudflare/workers-sdk/issues/8065.

If in an existing git repository, don't ask the user if they would like to use git.
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: minor change in default behavior

### DIFF
--- a/.changeset/silver-flowers-notice.md
+++ b/.changeset/silver-flowers-notice.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": minor
+---
+
+Create Cloudflare will no longer ask if the developer would like to use git if already in a git repository.
--- a/packages/create-cloudflare/src/__tests__/git.test.ts
+++ b/packages/create-cloudflare/src/__tests__/git.test.ts
@@ -198,11 +198,20 @@ describe("git helpers", () => {
 			mockGitInstalled(true);
 			mockGitConfig();
 			mockInsideGitRepo(true);
+			
+			// Mock user selecting true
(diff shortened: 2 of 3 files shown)


irvinebroque (commented):
will let someone else review but 👍 

(think ci going to ask you for `npx changeset`)

petebacondarwin (approved):
Looks great!

.changeset/silver-flowers-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-cloudflare": minor
```
petebacondarwin, line 2: I debated with myself whether this is just a bug fix (i.e. patch) but I think we can go with it being a small feature.

PR #8273, opened 2025-02-26
Fix workers ai binding for cf vite plugin

penalosa:
Continuation of https://github.com/cloudflare/workers-sdk/pull/8016 (unfortunately I can't push to that branch)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: The Vite plugin is not documented yet


justinvdm:
@penalosa thank you so much for continuing this in #8273! 🙇 I think I'll close this PR then, but let me know if I can help :)

irvinebroque:
Viable to extend this to Vectorize?

penalosa:
I've added the Images binding and the Vectorize binding to the playground—both should work as well as AI

jamesopstad (commented):
Looks great!

This needs to be added to preview mode too, which should be straightforward. You just need to update the `getPreviewMiniflareOptions` function.

The other thing this calls into question is the way we depend on Wrangler. As you're updating the `unstable_getMiniflareWorkerOptions` function, we'll need to bump the version we depend on in the `vite-plugin` catalog. We did previously discuss making Wrangler a dependency (and using the workspace version) rather than a peer dependency, which I would be in favour of so I'll open that discussion again outside this PR.

jamesopstad (approved):
Looks great! Just need to remove the `console.log` and then I think it's good to go. Thanks Samuel.

PR #8279, opened 2025-02-27
Workers Assets metafiles in asset-worker

GregBrimble:
Fixes WC-3266.

Implements support for Workers Assets metafiles (`_headers` and `_redirects`) in the production asset-worker.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: we'll do it once all the pieces land


packages/workers-shared/asset-worker/src/utils/headers.ts
```diff
@@ -29,3 +32,48 @@ export function getHeaders(
 function isCacheable(request: Request) {
 	return !request.headers.has("Authorization") && !request.headers.has("Range");
 }
+
+export function attachCustomHeaders(
```
petebacondarwin, line 36: Feels like this should be shared between pages and workers too?

packages/workers-shared/asset-worker/tests/assets-manifest.test.ts
petebacondarwin, line 1: Why the changes in this file?

petebacondarwin: Oh because you are running the test in vitest-pool-workers which doesn't have access to the file system?

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -129,7 +129,7 @@ export default {
 			if (userWorkerInvocation) {
 				// Don't send user Worker errors to sentry; we have no way to distinguish between
 				// CF errors and errors from the user's code.
-				return;
+				throw err;
```
petebacondarwin, line 132: If you are now throwing, is the comment above still accurate?

packages/workers-shared/utils/responses.ts
```diff
@@ -1,56 +1,123 @@
...
 			...init,
-			status: 304,
+			status: NotModifiedResponse.status,
 			statusText: "Not Modified",
 		});
 	}
 }
 
+export class MovedPermanentlyResponse extends Response {
+	static readonly status = 301 as const;
```
petebacondarwin, line 67: Why is this one `as const` while the others are not?

packages/workers-shared/asset-worker/src/handler.ts
```diff
@@ -1,40 +1,125 @@
...
+
+	let proxied = false;
+
+	if (redirectMatch) {
+		if (redirectMatch.status === 200) {
+			// A 200 redirect means that we are proxying to a different asset, for example,
+			// a request with url /users/12345 could be pointed to /users/id.html. In order to
+			// do this, we overwrite the pathname, and instead match for assets with that url,
+			// and importantly, do not use the regular redirect handler - as the url visible to
+			// the user does not change
```
petebacondarwin, line 76: In other words this is a "rewrite" rather than a "redirect", right?

packages/workers-shared/asset-worker/src/handler.ts
```diff
@@ -1,40 +1,125 @@
...
-	exists: typeof EntrypointType.prototype.unstable_exists,
-	getByETag: typeof EntrypointType.prototype.unstable_getByETag
+	exists: typeof EntrypointType.prototype.unstable_exists
 ) => {
-	const { pathname, search } = new URL(request.url);
+	const url = new URL(request.url);
+	const { host, search } = url;
+	let { pathname } = url;
+
+	const staticRedirectsMatcher = () => {
```
petebacondarwin, line 38: The additions here seem very similar to the `generateResponse()` function in pages-shared. I assume you considered sharing that logic between the two as well? Why did you not?

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1,9 +1,13 @@
 import { vi } from "vitest";
 import { mockJaegerBinding } from "../../utils/tracing";
 import { applyConfigurationDefaults } from "../src/configuration";
-import { handleRequest } from "../src/handler";
+import { canFetch, handleRequest } from "../src/handler";
 import type { AssetConfig } from "../../utils/types";
 
+const mockEnv = {
+	JAEGER: mockJaegerBinding(),
+};
```
petebacondarwin, line 9: Given that the production code has to do a check for a missing JAEGER binding anyway, is this needed?

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -224,4 +204,950 @@ describe("[Asset Worker] `handleRequest`", () => {
 		);
 		expect(response2.status).toBe(307);
 	});
+
+	describe("_headers", () => {
```
petebacondarwin, line 208: I assume these are basically just cut-n-pasted from the pages-shared package?
Any significant differences?

petebacondarwin (approved):
bunch of questions but fundamentally LGTM

packages/workers-shared/asset-worker/src/index.ts
```diff
@@ -175,20 +175,17 @@ export default class extends WorkerEntrypoint<Env> {
...
-		const intent = await getIntent(
-			decodedPathname,
-			{
-				...applyConfigurationDefaults(this.env.CONFIG),
-				not_found_handling: "none",
-			},
+		if (!this.env.JAEGER) {
+			// For wrangler tests, if we don't have a jaeger binding, default to a mocked binding
+			this.env.JAEGER = mockJaegerBinding();
+		}
```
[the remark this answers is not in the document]
GregBrimble: I agree. This was just a copy-paste from above. Let me try removing it from the Worker itself.

packages/workers-shared/asset-worker/tests/handler.test.ts
GregBrimble: Workers Assets' asset-serving behavior is different with it's `canFetch` and `html_handling` etc. so just wanted to write some explicit tests ensuring we had behavior we were happy with for Workers Assets. These are all new tests written by me, and are very comprehensive, I think.

packages/workers-shared/asset-worker/src/handler.ts
GregBrimble: I moved the ahead-of-time parsing and config, but left the runtime evaluation. Felt right, but I definitely won't die on that hill.

My vague thinking was that if we ever fixed a bug in metafiles, we'd probably only do so in Workers Assets and leave Pages frozen (we have access to compatibility date in Workers Assets!). This split makes it _reasonably_ easy to do that.

packages/workers-shared/asset-worker/src/utils/headers.ts
GregBrimble: https://github.com/cloudflare/workers-sdk/pull/8279#discussion_r1985290531

packages/workers-shared/router-worker/src/index.ts
GregBrimble: Returning `undefined` in a `fetch()` will throw an error anyway, so I haven't changed the behavior here: just "fixed" the implicit return type to `Promise<Response>` rather than `Promise<Response | undefined>` which it was before (needed as part of the [TS update commit](https://github.com/cloudflare/workers-sdk/pull/8279/commits/f57d1f830e9f29472d4af77a6491b605e5eba890).

It still doesn't log to Sentry since that would happen below only if we didn't return or throw here: https://github.com/cloudflare/workers-sdk/pull/8279/files#diff-d666c170036315c34fb63daeaa6abe85b7f96a3facd431c49ae1d7b8777f0869R138-R140

packages/workers-shared/asset-worker/tests/handler.test.ts
petebacondarwin: Damn, I was hoping that I wouldn't need to look through these 😆

packages/workers-shared/asset-worker/src/index.ts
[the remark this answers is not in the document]
GregBrimble: Failed. I can't mock an object-y binding like that in Miniflare very easily. Took your simpler nit and left a TODO.

PR #8288, opened 2025-02-27
feat(miniflare): Add assets Proxy Worker skeleton in miniflare

CarmenPopoviciu:
This PR implements a very basic Proxy Worker skeleton, and wires it in the "pipeline" miniflare creates for assets. This Worker will be incrementally worked on, but for now, the current implementation will forward all incoming requests to the Router Worker, thus leaving the current assets behaviour in local dev, unchanged.

TLDR 👇 

Screenshot 2025-02-27 at 16 44 34

**Why are we doing this?** Because we expect the proxy Worker code to be quite involved, so we don't want to worry at that point whether the wiring works properly or not. We just want to focus on that Worker code

Fixes DEVX-1641, DEVX-1642

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because: covered by existing tests
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/fixtures/workers-with-assets-only/tests/index.test.ts
+++ b/fixtures/workers-with-assets-only/tests/index.test.ts
@@ -3,176 +3,184 @@ import { fetch } from "undici";
 import { afterAll, beforeAll, describe, it } from "vitest";
 import { runWranglerDev } from "../../shared/src/run-wrangler-long-lived";
 
-describe("[Workers + Assets] static-assets only site`", () => {
-	let ip: string, port: number, stop: (() => Promise<unknown>) | undefined;
+const devCmds = [{ args: [] }, { args: ["--x-assets-rpc"] }];
 
-	beforeAll(async () => {
-		({ ip, port, stop } = await runWranglerDev(resolve(__dirname, ".."), [
-			"--port=0",
-			"--inspector-port=0",
-		]));
-	});
+describe.each(devCmds)(
+	"[wrangler dev $args][Workers + Assets] static-assets only site`",
+	({ args }) => {
+		let ip: string, port: number, stop: (() => Promise<unknown>) | undefined;
 
-	afterAll(async () => {
-		await stop?.();
-	});
-
-	it("should respond with static asset content", async ({ expect }) => {
-		let response = await fetch(`http://${ip}:${port}/index.html`);
-		let text = await response.text();
-		expect(response.status).toBe(200);
-		expect(text).toContain(`<h1>Hello Workers + Assets World 🚀!</h1>`);
-
-		response = await fetch(`http://${ip}:${port}/about/index.html`);
-		text = await response.text();
-		expect(response.status).toBe(200);
-		expect(text).toContain(`<p>Learn more about Workers with Assets soon!</p>`);
-	});
-
-	it("should resolve '/' to '/index.html' ", async ({ expect }) => {
-		let response = await fetch(`http://${ip}:${port}/`);
-		expect(response.status).toBe(200);
-		expect(await response.text()).toContain(
-			`<h1>Hello Workers + Assets World 🚀!</h1>`
-		);
-	});
-
-	it("should 404 if asset is not found in the asset manifest", async ({
-		expect,
-	}) => {
-		let response = await fetch(`http://${ip}:${port}/hello.html`);
-		expect(response.status).toBe(404);
-
-		response = await fetch(`http://${ip}:${port}/hello.txt`);
-		expect(response.status).toBe(404);
-	});
-
-	it("should handle content types correctly", async ({ expect }) => {
-		let response = await fetch(`http://${ip}:${port}/index.html`);
-		let text = await response.text();
-		expect(response.status).toBe(200);
-		expect(response.headers.get("Content-Type")).toBe(
-			"text/html; charset=utf-8"
-		);
-
-		response = await fetch(`http://${ip}:${port}/README.md`);
-		text = await response.text();
-		expect(response.status).toBe(200);
-		expect(response.headers.get("Content-Type")).toBe(
-			"text/markdown; charset=utf-8"
-		);
-		expect(text).toContain(`Welcome to Workers + Assets YAY!`);
-
-		response = await fetch(`http://${ip}:${port}/yay.txt`);
-		text = await response.text();
-		expect(response.status).toBe(200);
-		expect(response.headers.get("Content-Type")).toBe(
-			"text/plain; charset=utf-8"
-		);
-		expect(text).toContain(`.----------------.`);
-
-		response = await fetch(`http://${ip}:${port}/lava-lamps.jpg`);
-		expect(response.status).toBe(200);
-		expect(response.headers.get("Content-Type")).toBe("image/jpeg");
-	});
-
-	it("should return 405 for non-GET or HEAD requests if asset route exists", async ({
-		expect,
-	}) => {
-		// as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
-		// excl. TRACE and CONNECT which are not supported
-
-		let response = await fetch(`http://${ip}:${port}/index.html`, {
-			method: "POST",
+		beforeAll(async () => {
+			({ ip, port, stop } = await runWranglerDev(resolve(__dirname, ".."), [
+				"--port=0",
+				"--inspector-port=0",
+				...args,
+			]));
 		});
-		expect(response.status).toBe(405);
-		expect(response.statusText).toBe("Method Not Allowed");
 
-		response = await fetch(`http://${ip}:${port}/index.html`, {
-			method: "PUT",
+		afterAll(async () => {
+			await stop?.();
 		});
-		expect(response.status).toBe(405);
-		expect(response.statusText).toBe("Method Not Allowed");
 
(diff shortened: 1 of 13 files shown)


fixtures/workers-with-assets-run-worker-first/tests/index.test.ts
```diff
@@ -3,35 +3,43 @@ import { fetch } from "undici";
...
-	let ip: string, port: number, stop: (() => Promise<unknown>) | undefined;
+const devCmds = [{ args: [] }, { args: ["--x-assets-rpc"] }];
 
-	beforeAll(async () => {
-		({ ip, port, stop } = await runWranglerDev(resolve(__dirname, ".."), [
-			"--port=0",
-			"--inspector-port=0",
-		]));
-	});
+describe.each(devCmds)(
```
CarmenPopoviciu, line 8: no tests changes were added here!! We are basically running the existing set of tests for both `wrangler dev` and `wrangler dev --x-assets-rpc`

fixtures/workers-with-assets/tests/index.test.ts
```diff
@@ -3,211 +3,221 @@ import { fetch } from "undici";
...
-		// POST etc. request -> RW -> AW -> check manifest --405--> RW --405--> eyeball
-
-		// methods as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
-		// excl. TRACE and CONNECT which are not supported
-
-		let response = await fetch(`http://${ip}:${port}/index.html`, {
-			method: "POST",
+const devCmds = [{ args: [] }, { args: ["--x-assets-rpc"] }];
+
+describe.each(devCmds)(
```
CarmenPopoviciu, line 8: no tests changes were added here!! We are basically running the existing set of tests for both `wrangler dev` and `wrangler dev --x-assets-rpc`

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -267,9 +270,13 @@ function getCustomServiceDesignator(
 			serviceName = getBuiltinServiceName(workerIndex, kind, name);
 		}
 	} else if (service === kCurrentWorker) {
-		// Sets SELF binding to point to router worker instead if assets are present.
+		// Sets SELF binding to point to the Router Worker or the Assets
+		// Proxy Worker (depending on whether `unsafeEnableAssetsRpc` is set)
+		// if assets are present.
 		serviceName = hasAssetsAndIsVitest
-			? `${ROUTER_SERVICE_NAME}:${refererName}`
+			? unsafeEnableAssetsRpc
```
CarmenPopoviciu, line 277: supporting this for vitest-pool-workers made things more complicated than I would have wanted, because it required introducing `unsafeEnableAssetsRpc` on the vitest pool workers side as well. Does it make sense to support this experimental feat in pool workers at the moment?

packages/vitest-pool-workers/src/pool/config.ts
```diff
@@ -222,6 +222,7 @@ async function parseCustomPoolOptions(
 		options.miniflare.assets.routingConfig.has_user_worker = Boolean(
 			options.main
 		);
+		options.miniflare.assets.unsafeEnableAssetsRpc ??= false;
```
CarmenPopoviciu, line 225: to verify my understanding, since vitest-pool-workers is not my jam, does this mean we can now do smth like 👇 ?

```
test: {
    poolOptions: {
      workers: {
        miniflare: {
          assets: { unsafeEnableAssetsRpc: true }
        }
      },
    },
  },
```

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -152,6 +155,7 @@ const CoreOptionsSchemaInput = z.intersection(
 		 (If there are assets but we're not using vitest, the miniflare entry worker can point directly to RW.)
 		 */
 		hasAssetsAndIsVitest: z.boolean().optional(),
+		unsafeEnableAssetsRpc: z.boolean().optional(),
```
penalosa, line 158: Is this the right place for this? I'm pretty sure Assets RPC should be a single on/off switch for _all_ workers, not per-worker

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -791,6 +791,8 @@ export function buildAssetOptions(config: Pick<ConfigBundle, "assets">) {
 				binding: config.assets.binding,
 				routingConfig: config.assets.routingConfig,
 				assetConfig: config.assets.assetConfig,
+				// temporary and to be removed once flag is removed
+				unsafeEnableAssetsRpc: getFlag("ASSETS_RPC"),
```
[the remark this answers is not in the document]
penalosa: I'm slightly uneasy about this, since it means it's per-worker rather than a miniflare-level config, but since it's temporary it's probably not that important

packages/miniflare/src/index.ts
```diff
@@ -1205,13 +1208,23 @@ export class Miniflare {
...
+							 * Worker, if the `unsafeEnableAssetsRpc` flag is enabled
+							 */
 							const maybeAssetTargetService = allWorkerOpts.find(
 								(worker) =>
 									worker.core.name === targetWorkerName && worker.assets.assets
 							);
 							if (maybeAssetTargetService && !binding.service?.entrypoint) {
 								assert(binding.service?.name);
-								binding.service.name = `${ROUTER_SERVICE_NAME}:${targetWorkerName}`;
+								binding.service.name = this.#workerOpts[0].assets.assets
```
penalosa, line 1224: Yeah, I think this type of thing is why it should be a top-level setting in Miniflare—it feels a bit weird to be using the _first_ worker's assets settings to decide if all the other workers should have RPC enabled

fixtures/workers-with-assets-only/tests/index.test.ts
```diff
@@ -3,176 +3,184 @@ import { fetch } from "undici";
...
-		expect,
-	}) => {
-		// as per https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
-		// excl. TRACE and CONNECT which are not supported
-
-		let response = await fetch(`http://${ip}:${port}/index.html`, {
-			method: "POST",
+const devCmds = [{ args: [] }, { args: ["--x-assets-rpc"] }];
+
+describe.each(devCmds)(
```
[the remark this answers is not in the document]
penalosa: Should we do that with the multiworker E2E tests as well?

penalosa (approved):

packages/wrangler/src/dev/miniflare.ts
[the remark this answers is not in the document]
CarmenPopoviciu: based on feedback, reverting this change

packages/miniflare/src/plugins/core/index.ts
CarmenPopoviciu: removed

packages/miniflare/src/index.ts
CarmenPopoviciu: you're right. reverting the change

packages/vitest-pool-workers/src/pool/config.ts
CarmenPopoviciu: > But the user does set `unsafeEnableAssetsRpc`

How? `unsafeEnableAssetsRpc` can only be set if we pass the `--x-assets-rpc` flag to `wrangler dev`. AFAICT `unstable_getMiniflareWorkerOptions` only looks at the config

packages/miniflare/src/index.ts
```diff
@@ -1239,6 +1252,7 @@ export class Miniflare {
 				unsafeEphemeralDurableObjects,
 				queueProducers,
 				queueConsumers,
+				unsafeEnableAssetsRpc,
```
CarmenPopoviciu, line 1255: we need this so that the assets plugin can conditionally add the rpc proxy Worker

packages/vitest-pool-workers/src/pool/config.ts
emily-shen: whoops forgot how the flags were set - thought it was an environment variable. nvm 😅 I think yes to your original question but probably best to try it out

packages/miniflare/src/plugins/core/index.ts
CarmenPopoviciu: that will actually be needed since `unsafeEnableAssetsRpc` is only ever set if `--x-assets-rpc` is passed to `wrangler dev`. Therefore, vitest-pool-workers needs a similar config key, smth like 👇 , to be able to correctly execute this conditional logic. Otherwise where would the `unsafeEnableAssetsRpc` value ever came from, if not in a `wrangler dev` context?

```
test: {
    poolOptions: {
      workers: {
        miniflare: {
          unsafeEnableAssetsRpc: true
        }
      },
    },
  },
  ```

fixtures/workers-with-assets-only/tests/index.test.ts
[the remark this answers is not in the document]
CarmenPopoviciu: ✅

edmundhung (approved):

PR #8290, opened 2025-02-27
C3: ensure that git commits are generated

petebacondarwin:
Fixes #8264

This fixes a small regression introduced by #8264. In that PR we wanted to avoid running `git init` if there was already a repository in place, but it also resulted in the git commit being skipped in every case.

Even if you are not in a git repo you would see:
image

Here we fix the regression but also tidy up the code (and unit tests) because there was a lot of unnecessary repeated commands being run, and it was not always clear when values should have been set by.

It also adds in support for testing that the git commit is actually applied in e2e tests, which was not being tested before.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: C3 only
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

### DIFF
--- a/packages/create-cloudflare/e2e-tests/frameworks.test.ts
+++ b/packages/create-cloudflare/e2e-tests/frameworks.test.ts
@@ -25,6 +25,7 @@ import {
 	spawnWithLogging,
 	test,
 	testDeploymentCommitMessage,
+	testGitCommitMessage,
 	waitForExit,
 } from "./helpers";
 import type { TemplateConfig } from "../src/templates";
@@ -120,7 +121,7 @@ function getFrameworkTests(opts: {
 				},
 			},
 			hono: {
-				testCommitMessage: false,
+				testCommitMessage: true,
 				unsupportedOSs: ["win32"],
 				verifyDeploy: {
 					route: "/",
@@ -180,7 +181,16 @@ function getFrameworkTests(opts: {
 				flags: ["--typescript", "--no-install", "--no-git-init"],
 			},
 			next: {
-				testCommitMessage: false,
+				testCommitMessage: true,
+				flags: [
+					"--ts",
+					"--tailwind",
+					"--eslint",
+					"--app",
+					"--import-alias",
+					"@/*",
+					"--src-dir",
+				],
 				verifyBuildCfTypes: {
 					outputFile: "cloudflare-env.d.ts",
 					envInterfaceName: "CloudflareEnv",
@@ -403,7 +413,7 @@ function getFrameworkTests(opts: {
 				},
 			},
 			hono: {
-				testCommitMessage: false,
+				testCommitMessage: true,
 				unsupportedOSs: ["win32"],
 				verifyDeploy: {
 					route: "/",
@@ -658,6 +668,7 @@ describe.concurrent(
 							{
 								argv: [
 									...(experimental ? ["--experimental"] : []),
+									...(testConfig.testCommitMessage ? ["--git"] : ["--no-git"]),
 									...(testConfig.flags ? ["--", ...testConfig.flags] : []),
 								],
 								promptHandlers: testConfig.promptHandlers,
@@ -674,6 +685,14 @@ describe.concurrent(
 						expect(wranglerPath).toExist();
 
 						await addTestVarsToWranglerToml(project.path);
+
+						if (testConfig.testCommitMessage) {
+							await testGitCommitMessage(
+								project.name,
+								frameworkId,
+								project.path,
+							);
+						}
 
 						// Make a request to the deployed project and verify it was successful
 						await verifyDeployment(
@@ -741,7 +760,6 @@ const runCli = async (
 		framework,
 		NO_DEPLOY ? "--no-deploy" : "--deploy",
 		"--no-open",
-		"--no-git",
 	];
 
 	args.push(...argv);
--- a/packages/create-cloudflare/e2e-tests/helpers.ts
+++ b/packages/create-cloudflare/e2e-tests/helpers.ts
@@ -440,6 +440,9 @@ const testProjectDir = (suite: string, test: string) => {
 	return { getName, getPath, clean };
 };
 
+/**
+ * Test that we pushed the commit message to the deployment correctly.
+ */
 export const testDeploymentCommitMessage = async (
 	projectName: string,
 	framework: string,
@@ -491,6 +494,9 @@ export const testDeploymentCommitMessage = async (
 	expect(projectLatestCommitMessage).toContain(`framework = ${framework}`);
 };
 
+/**
+ * Test that C3 added a git commit with the correct message.
+ */
 export async function testGitCommitMessage(
 	projectName: string,
 	framework: string,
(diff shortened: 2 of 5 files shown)


packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -180,7 +181,16 @@ function getFrameworkTests(opts: {
...
+				testCommitMessage: true,
+				flags: [
+					"--ts",
+					"--tailwind",
+					"--eslint",
+					"--app",
+					"--import-alias",
+					"@/*",
+					"--src-dir",
+				],
```
petebacondarwin, line 193: Without these the create-next-app CLI hangs waiting for a response.

petebacondarwin: At least locally. Perhaps it doesn't ask when in non-interactive mode?

packages/create-cloudflare/src/__tests__/git.test.ts
```diff
@@ -147,9 +147,8 @@ describe("git helpers", () => {
 
 	describe("offerGit", async () => {
 		test("happy path", async () => {
-			const ctx = createTestContext();
+			const ctx = createTestContext("test", { projectName: "test" });
```
petebacondarwin, line 150: The `testCreateContext()` helper sets up the `ctx.args` to be all the `C3_DEFAULT_ARGS` values, which undermines the idea that the user has not provided any command line args.  By providing an args object here we ensure that none of the CLI args are set so that we have control over them in the tests.

packages/create-cloudflare/src/__tests__/git.test.ts
```diff
@@ -163,11 +162,23 @@ describe("git helpers", () => {
 				["git", "init", "--initial-branch", "main"],
 				expect.any(Object),
 			);
+			expect(ctx.gitRepoAlreadyExisted).toBe(false);
 			expect(ctx.args.git).toBe(true);
```
petebacondarwin, line 166: By the end of `offerGit()` we should have decided whether `git` is required but also whether `gitRepoAlreadyExisted` is true. This is then inherited/used in the `gitCommit()` function later.

packages/create-cloudflare/src/git.ts
```diff
@@ -26,37 +27,11 @@ export const offerGit = async (ctx: C3Context) => {
...
-				"You're in an existing git repository. Do you want to use git for version control?",
-			label: "git",
-			defaultValue: C3_DEFAULTS.git,
-		});
-
-		return;
-	}
-
-	ctx.args.git = await processArgument(ctx.args, "git", {
+	ctx.args.git ??= await processArgument(ctx.args, "git", {
```
petebacondarwin, line 30: This simplification means that we only ask the user if `ctx.args.git === undefined`.

packages/create-cloudflare/e2e-tests/frameworks.test.ts
vicb: Add that as a comment?

vicb: Also consider using `--yes` https://nextjs.org/docs/app/api-reference/cli/create-next-app#reference

packages/create-cloudflare/e2e-tests/helpers.ts
```diff
@@ -490,6 +491,24 @@ export const testDeploymentCommitMessage = async (
 	expect(projectLatestCommitMessage).toContain(`framework = ${framework}`);
 };
 
+export async function testGitCommitMessage(
```
vicb, line 494: Maybe add comments here - one liner is probably enough

packages/create-cloudflare/src/__tests__/git.test.ts
vicb: Should your comment be a code comment?

packages/create-cloudflare/src/__tests__/git.test.ts
```diff
@@ -163,11 +162,23 @@ describe("git helpers", () => {
...
 		});
 
 		test("git not installed", async () => {
```
vicb, line 169: Comment:

> "git not installed" (This particular instance is not updated in this PR)

A guideline I find very useful (from Google) is to describe both the context and the outcome for test.

i.e. here it could help understand:

```ts
			expect(updateStatus).not.toHaveBeenCalled();
			expect(processArgument).not.toHaveBeenCalled();
```

packages/create-cloudflare/src/git.ts
```diff
@@ -26,37 +27,11 @@ export const offerGit = async (ctx: C3Context) => {
...
 		type: "confirm",
-		question: "Do you want to use git for version control?",
+		question: ctx.gitRepoAlreadyExisted
+			? "You're in an existing git repository. Do you want to use git for version control?"
```
vicb, line 33: Comment:

I don't really understand "You're in an existing git repository. Do you want to use git for version control?"

i.e. am I not using git for version control if I am in a git repo?

packages/create-cloudflare/src/git.ts
```diff
@@ -83,10 +58,15 @@ export const offerGit = async (ctx: C3Context) => {
...
+	assert.notStrictEqual(
+		ctx.args.git,
+		undefined,
+		"Expected git context to be defined by now",
+	);
 	// Note: createCommitMessage stores the message in ctx so that it can
 	//       be used later even if we're not in a git repository, that's why
 	//       we unconditionally run this command here
-	const commitMessage = await createCommitMessage(ctx);
+	ctx.commitMessage = await createCommitMessage(ctx);
```
vicb, line 69: Comment:

That's indeed confusing - maybe renaming the property to "creationMessage" and using that for the commit message could help alleviate some confusion?

(not asking to chnage in this PR)

vicb (approved):
LGTM.

Nice cleanup

I added some minor comments (some unrelated to the PR)

packages/create-cloudflare/src/__tests__/git.test.ts
petebacondarwin: 9442311fdea8dfd51738da4b6d541919e1d46c2f

packages/create-cloudflare/e2e-tests/helpers.ts
petebacondarwin: a2d8f89fcfbdac55b36978321feadcf75d8590af

packages/create-cloudflare/src/__tests__/git.test.ts
petebacondarwin: Yes I agree.

petebacondarwin: 323e23694d7c483868c4f17085c8ff9092597efb

packages/create-cloudflare/src/git.ts
petebacondarwin: This was added by Dan in the previous PR that caused the regression that this PR is fixing.
The idea is that you might be in some monorepo that is using git, but still not want this particular directory to be tracked via git - and so no git commit should be created.

packages/create-cloudflare/e2e-tests/frameworks.test.ts
petebacondarwin: will try it out in a follow up

PR #8295, opened 2025-02-27
feat: Add Cloudflare system prompt to newly created projects

threepointone:
AI driven editors like Cursor can use a system prompt to guide the AI in writing code. This function adds a system prompt to the project directory if it doesn't already exist. We use the cursor standard path, other editors expect the user to explicitly specify the path to the system prompt (for now)

Fixes #0000

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: non functional change 
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: non functional, should happen in the background

### DIFF
--- a/packages/create-cloudflare/src/__tests__/templates.test.ts
+++ b/packages/create-cloudflare/src/__tests__/templates.test.ts
@@ -10,10 +10,12 @@ import {
 } from "helpers/files";
 import { beforeEach, describe, expect, test, vi } from "vitest";
 import {
+	addCloudflareSystemPrompt,
 	addWranglerToGitIgnore,
 	deriveCorrelatedArgs,
 	downloadRemoteTemplate,
 } from "../templates";
+import { createTestContext } from "./helpers";
 import type { PathLike } from "fs";
 import type { C3Args, C3Context } from "types";
 
@@ -316,5 +318,34 @@ describe("deriveCorrelatedArgs", () => {
 		).toThrow(
 			"The `--ts` argument cannot be specified in conjunction with the `--lang` argument",
 		);
+	});
+});
+
+describe("addCloudflareSystemPrompt", () => {
+	const writeFileResults: {
+		file: string | undefined;
+		content: string | undefined;
+	} = { file: undefined, content: undefined };
+
+	beforeEach(() => {
+		vi.mocked(writeFile).mockImplementation((file: string, content: string) => {
+			writeFileResults.file = file;
+			writeFileResults.content = content;
+		});
+	});
+
+	beforeEach(() => {
+		vi.mocked(existsSync).mockReset();
+		vi.mocked(readFile).mockReset();
+		vi.mocked(directoryExists).mockReset();
+
+		writeFileResults.file = undefined;
+		writeFileResults.content = undefined;
+	});
+	test("it should add the system prompt to the project", async () => {
+		const ctx = createTestContext();
+		await addCloudflareSystemPrompt(ctx);
+		expect(writeFileResults.file).toContain(".cursor");
+		expect(writeFileResults.content).toContain("cloudflare");
 	});
 });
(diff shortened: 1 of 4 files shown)


packages/create-cloudflare/src/__tests__/templates.test.ts
```diff
@@ -318,3 +320,17 @@ describe("deriveCorrelatedArgs", () => {
...
 });
+
+describe("addCloudflareSystemPrompt", () => {
+	test("it should add the system prompt to the project", async () => {
+		const ctx = createTestContext();
+		await addCloudflareSystemPrompt(ctx);
+		// list all files in the project
+
+		const files = readdirSync(ctx.project.path, { recursive: true });
+		console.log(files); // this comes up as undefined?
```
threepointone, line 331: I don't know why this fails (tbh I don't know what kind of test to write here). can someone help me?

packages/create-cloudflare/src/__tests__/templates.test.ts
```diff
@@ -318,3 +320,13 @@ describe("deriveCorrelatedArgs", () => {
 		);
 	});
 });
+
+describe.only("addCloudflareSystemPrompt", () => {
+	test("it should add the system prompt to the project", async () => {
+		const ctx = createTestContext();
+		await addCloudflareSystemPrompt(ctx);
+		expect(existsSync(ctx.project.path + "/.cursor/rules/cloudflare.mdc")).toBe(
```
threepointone, line 328: bah this fails too

penalosa (commented):
Not going to block this PR, but I have a couple of concerns:
- This is completely unversioned, and there's no mechanism for users to get updates/know they're using the latest version of the prompt/update the prompt. With a platform like Cloudflare that has relatively frequent changes that seems particularly bad—why is this a text file rather than a package? Could we publish this as e.g. `@cloudflare/workers-prompt` and symlink the cursor file to `node_modules/@cloudflare/workers-prompt/prompt.txt` (or configure the cursor file location?). As one small example of the potential problems—the prompt contains this line:

  > Strongly prefer the agent-sdk to build AI Agents when asked,
  
  which I assume should be _agents-sdk_? ([`agent-sdk`](https://www.npmjs.com/package/agent-sdk) is something unrelated)
- How has this been tested? The prompt is over 4000 words long, and contains a _lot_ of detail. Have we done qualitative tests on the Workers this generates? How do they compare to Workers generated without this prompt? How did we arrive at this final version—what bits of the prompt are important/what bits could be dropped/what bits should be expanded?

threepointone:
both good questions. we're moving a little quickly here because it's a fast moving space and we want to get it into people's hands and iterate. 

- symlinking is a good idea. but we also want it live on the url. so I'm not sure whether the source should be workers-sdk or cloudflare-docs.
- but also, people edit these files themselves to make their own modifications, so pointing to a versioned package might not be ideal.  
- good catch on the package name, will fix asap. 
- right now we're testing by hand, on [redacted internal project 🧢], we'll have evals for it later. in general it's working really well atm, better than nothing.

threepointone:
oh, and the reason it works despite the bad package name there is because of the tons of examples later in the prompt

Cherry (commented):
I'm personally a little concerned/confused about folks wanting to removing bloat from C3 like apparently the [`.prettierrc` file](https://github.com/cloudflare/workers-sdk/pull/8263), but then happy to add things like an AI system prompt by default that'll end up in so many git repos moving forward.

I understand that AI is hype right now, but some folks are pretty against it and have their own preferences to use (or not use) specific things, or may even work at places where not using generative AI is banned entirely - so is including this by default on every single project and so many different versions of this ending up in git repos across the web over time a good call?

I'd encourage this to be an explicit opt-in.

rickyrobinett:
Can we put this in an opt-in (like we do with git)? I think that will help in two ways:

1) Help drive awareness that this is being included and you can take advantage of it
2) Let anyone who may have reasons of not wanting this (enterprise restrictions, concerns around AI, etc) quickly opt-out

threepointone:
closing this because it might need some rework. I see windsurf also has it's own thing now (.windsurfrules). I also don't want to push this on the team if they're resisting of course haha. I'll figure something better.

petebacondarwin:
Moving this to draft while we work out how best we could land this.

threepointone:
closing this one, feel free to take another stab at it fresh

PR #8300, opened 2025-02-28
Use @cloudflare/unenv-preset in wrangler

vicb:
To make sure the dependency resolution works fine, I have create 3 projects (`create-cloudflare@latest --framework=nuxt`) with `npm`, `pnpm`, and `yarn`. I then override the wrangler version in the `package.json` with the preview version from this PR.

I then logged the `aliasAbsolute` (in `cli.js`) and checked that all the paths were resolved as expected (i.e. they do not point to unenv@1 which is used by Nuxt).

---

- Tests
  - [ ] TODO (before merge)
  - [] Tests included
  - [x] Tests not necessary because: using existing tests
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 

---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #8300 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #7915


petebacondarwin:
We should cherry-pick this commit to this PR: 678773a228ec5b7ec92f81bebb090f0ef9a290d5

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
```diff
@@ -15,15 +14,24 @@ const REQUIRED_UNENV_ALIAS_NAMESPACE = "required-unenv-alias";
  * @param _unenvResolvePaths Root paths used to resolve absolute paths.
  * @returns ESBuild plugin
  */
-export function nodejsHybridPlugin(_unenvResolvePaths?: string[]): Plugin {
-	const { alias, inject, external } = env(nodeless, cloudflare);
+export async function nodejsHybridPlugin(
+	_unenvResolvePaths?: string[]
```
petebacondarwin, line 18: These paths are no longer used, so we can remove it all the way to the top env vars.
But this also means that the current approach to testing in `packages/unenv-preset/tests/index.test.ts` isn't actually going to test against the local build.

Instead, I think it is time to move those checks (which is really an integration test with Wrangler) into the `fixtures/nodejs-hybrid-app` fixture alongside all the other tests that give the same coverage.

Then we can remove the `devDep` from this package to Wrangler, resolving the cyclic dependency, allowing  Wrangler to have a `workspace:*` dependency on this package.

Then we have clean simple dependency setup, the latest changes are automatically get tested and released together.

petebacondarwin (changes requested):
I think it is time to move to a workspace dependency on this package.

packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts
vicb: Let's handle this in a follow up PR.

I think the unenv-preset should use miniflare for `packages/unenv-preset/tests/index.test.ts` because we do not want to wait for wrangler to have the updated workerd before testing - we want to be able to create a preset PR has soon as workerd is out (or even with a dev build).

As discussed before while we should be able to create the PR early, it should not be mergeable (i.e. tests failing) until whatever workerd version it relies on is deployed.

Does that sound good?

vicb: And ideally we should keep the ability use those paths to have the ability to run tests a dev version of the preset.

For now this PR is using `require.resolve` to minimize the risk of having wrong resolved paths. But it would be great to switch to whatever resolution is used by unenv latest RC (it has changed recently) which allow overriding the base path.

vicb:
> I think it is time to move to a workspace dependency on this package.

soon :)

petebacondarwin (approved):
OK let's land this and iterate. I am still unhappy with the dependency setup but we can work on that. Just for clarity here is a diagram of the current situation:

image

packages/wrangler/package.json
```diff
@@ -70,13 +70,14 @@
...
 		"@cloudflare/kv-asset-handler": "workspace:*",
+		"@cloudflare/unenv-preset": "2.0.0",
 		"@esbuild-plugins/node-globals-polyfill": "0.2.3",
 		"@esbuild-plugins/node-modules-polyfill": "0.2.2",
 		"blake3-wasm": "2.1.5",
 		"esbuild": "0.17.19",
 		"miniflare": "workspace:*",
 		"path-to-regexp": "6.3.0",
-		"unenv": "2.0.0-rc.1",
+		"unenv": "2.0.0-rc.8",
```
petebacondarwin, line 80: Perhaps move this to the pnpm workspace catalog?

```suggestion
		"unenv": "catalog:default",
```

petebacondarwin: Since the vite plugn will use the same version

vicb: > Since the vite plugn will use the same version

Can this be done when the cute plugin is ready? 
I have no idea about when it's ready to move to rc8

PR #8304, opened 2025-02-28
chore: add concurrency and caching for Zone IDs and Workers routes lookups

jahands:
Fixes [BANDA-768](https://jira.cfdata.org/browse/BANDA-768)

_Describe your change..._

Workers with many routes can result in duplicate Zone lookups during deployments, making deployments unnecessarily slow. This compounded by the lack of concurrency when making these API requests.

This change deduplicates these requests and adds concurrency to help speed up deployments.

In my Worker containing 21 Routes, this reduces the number of requests from 66 -> 36, and the total deploy time from 24 seconds -> 10

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Optimization only


packages/wrangler/src/__tests__/helpers/run-wrangler.ts
```diff
@@ -11,7 +11,11 @@ export async function runWrangler(
 ) {
 	const originalEnv = process.env;
 	try {
-		process.env = { ...originalEnv, ...env };
+		process.env = {
+			...originalEnv,
+			...env,
+			VITEST_CACHE_BUST: new Date().toISOString(),
```
petebacondarwin, line 17: can you explain the need for this here?

jahands: Wasn't sure if there was a more preferred way of solving this...
In zones.ts, I created a global variable to cache zone/ID mappings.
A couple tests were failing because that global variable does not get reset between tests, leading to incorrect results being returned.

jahands:
TODO: Fix lint issues (it doesn't like my nested functions)

packages/wrangler/src/__tests__/helpers/run-wrangler.ts
jahands: This would not be an issue in actual usage because wrangler isn't deploying multiple times in a row, and (more importantly) an account/zone combo will always return the same ID from the API (at least within the lifetime of a wrangler process)

jahands:
Tested the wrangler preview release:
- added/removed route from Worker and validated it worked by making requests to the affected route

irvinebroque (commented):
Curious — why does wrangler need to do a zone lookup for each route?

Is it because we've designed things where to add/update/delete routes you need to go through the zones API: `/zones/{zone_id}/workers/routes/{route_id}`

https://developers.cloudflare.com/api/resources/workers/subresources/routes/methods/update/

...rather than just providing the routes directly to the Workers API, and letting the Workers control plane handle any updates?

jahands:
> Curious — why does wrangler need to do a zone lookup for each route?
> 
> Is it because we've designed things where to add/update/delete routes you need to go through the zones API: `/zones/{zone_id}/workers/routes/{route_id}`
> 
> https://developers.cloudflare.com/api/resources/workers/subresources/routes/methods/update/
> 
> ...rather than just providing the routes directly to the Workers API, and letting the Workers control plane handle any updates?

We currently update all routes in one API call to `/accounts/{account_id}/workers/scripts/{script_id}/routes` (can't find this endpoint in the docs, unfortunately) ([source](https://github.com/cloudflare/workers-sdk/blob/jhands/deploy-concurrency/packages/wrangler/src/deploy/deploy.ts#L1004-L1044)).

There is a fallback for updating routes using the per-zone API you mentioned ([source](https://github.com/cloudflare/workers-sdk/blob/jhands/deploy-concurrency/packages/wrangler/src/deploy/deploy.ts#L1051-L1167)), but I'm unsure how much this code path gets used in practice.

However, even if we use the first API that only requires one API call, we're resolving all routes to a Zone ID for client-side validation to ensure the same Route isn't used across multiple Workers: https://github.com/cloudflare/workers-sdk/blob/jhands/deploy-concurrency/packages/wrangler/src/triggers/deploy.ts#L222-L238
So, if I have 5 subdomains on the same Zone, we're making 5 fetch calls for each of those subdomains (since we don't know if they are a liberated Zone or not), all for client-side validation.

It's unclear to me why we need to do client-side validation that routes aren't being reused across Workers instead of letting the server determine that. If we could let the server validate, we could remove another N API calls (where N=# of routes).

For reference, here's a before/after logs from my Worker with 7 routes showing all the API calls being made: https://gist.github.com/jahands/e047a56c9ff7fba729464f33b149698f

irvinebroque:
Rad. Really helpful answer.

> It's unclear to me why we need to do client-side validation that routes aren't being reused across Workers instead of letting the server determine that. If we could let the server validate, we could remove another N API calls to the Zones API (where N=# of routes).

Wonder if worth exploring! Also seems like big win for anything that does not using Wrangler.

jahands:
I looked further into the route validation and noticed currently do this validation before [calling multiple APIs] (https://github.com/cloudflare/workers-sdk/blob/jhands/deploy-concurrency/packages/wrangler/src/triggers/deploy.ts#L245-L323) to deploy:
- Routes
- Cron schedules
- Custom domains
- Queue producers
- Queue consumers
- Workflows

All of the above deploys are done concurrently. I validated that the Routes API does validate route conflicts, so we could use that rather than fetching all Zones to determine if there are conflicts beforehand. But if we want to validate routes before deploying cron/queues/etc, we need to wait for the Routes API call to complete first. Unfortunately, this Routes API call takes 2-3x longer than fetching 7 zones concurrently (1 per route, so 7 in my case). This means that updating Routes before anything else would be slower than fetching Zones with the current optimization this PR introduces.

**Next steps:**
1. Figure out if we can publish cron/queues/workflows/etc. concurrently with Routes, rather than validating routes first. If so, we can eliminate the Zones API calls.
2. Investigate whether we can speed up the Routes API (I checked and we sequentially lookup the Zone for each Route, the same reason our client code here is slow).

irvinebroque:
🔥 👏 

> Routes API call takes 2-3x longer than fetching 7 zones concurrently

😱

dcartertwo:
amazing work 👏🏾

packages/wrangler/src/zones.ts
```diff
@@ -131,17 +152,41 @@ async function getZoneIdFromHost(from: {
...
-				name: hostPieces.join("."),
-				"account.id": from.accountId,
-			})
-		);
-		if (zones.length > 0) {
-			return zones[0].id;
+		const cacheKey =
+			hostPieces.join(".") +
+			from.accountId +
+			(process.env.VITEST ? process.env.VITEST_CACHE_BUST ?? "" : "");
```
petebacondarwin, line 158: I'm really not keen on this approach, even though it is just apparent in testing.
I think I'd rather we create the cache (or cache key) in the code that makes all these requests and pass it down so that it is clear where the state is being used.

petebacondarwin (commented):
Other than my concerns over the VITEST_CACHE_BUST approach the rest looks good to me.

packages/wrangler/src/zones.ts
petebacondarwin: For example, `process.env` itself is a global so is there not a chance of conflict between parallel tests already?

jahands: Agreed - I like the approach you took of passing in a map!

petebacondarwin (approved):

PR #8309, opened 2025-02-28
fix(unenv-preset): define __filename

vicb:
This PR adds `__filename` which was not defined before:
(The whole `// -- Unbuild CommonJS Shims --` section is added by this PR)

```ts
// -- Unbuild CommonJS Shims --
import __cjs_url__ from 'url';
import __cjs_path__ from 'path';
import __cjs_mod__ from 'module';
const __filename = __cjs_url__.fileURLToPath(import.meta.url);
const __dirname = __cjs_path__.dirname(__filename);
const require = __cjs_mod__.createRequire(import.meta.url);

// [...] 
const cloudflare = {
  meta: {
    name: "unenv:cloudflare",
    version,
    url: __filename
  },
```

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: it will work ;)
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no change

### DIFF
--- a/packages/unenv-preset/build.config.ts
+++ b/packages/unenv-preset/build.config.ts
@@ -2,6 +2,9 @@ import { defineBuildConfig } from "unbuild";
 
 export default defineBuildConfig({
 	declaration: true,
+	rollup: {
+		cjsBridge: true,
+	},
 	entries: [
 		"src/index",
 		{ input: "src/runtime/", outDir: "dist/runtime", format: "esm" },


petebacondarwin:
It should be straightforward to add a test for this, no?

vicb:
> It should be straightforward to add a test for this, no?

it's all tested with 8300

petebacondarwin (approved):
Ideally a test would be good, since someone might just remove that setting in the future, not realising what it does.

vicb:
> Ideally a test would be good, since someone might just remove that setting in the future, not realising what it does.

We are in this weird state where wrangler does not use this pkg yet. Once that is sorted out, changing this setting will fail the tests. My priority for now is to get the the new preset out and I'll work on improving the test story after that. I hope this makes sense.

PR #8310, opened 2025-02-28
Enable nodejs_compat by default for new projects created by C3

irvinebroque:
refs https://github.com/cloudflare/workers-sdk/pull/7388

We've gone back and forth on this in the past — but in practice, have found that:

1. People are most successful in getting started when `nodejs_compat` is enabled. Hit the ground running, and NPM packages you want to use are more likely to work.
2. When `nodejs_compat` is _not_ enabled by default for new projects, products like Hyperdrive, or tutorials that use any NPM packages, need to explain extra steps and caveats to developers. Steps that are easy to miss, or misunderstand.

Getting the ecosystem using shared standards is important. That's why WinterTC exists, why all the awesome work that @jasnell and others do matters. How we get there is by working with WinterTC and the broader community — not by holding back `nodejs_compat` to developers getting started building Workers. The whole point is that developers and open-source maintainers shouldn't have to navigate a maze of differences — and enabling `nodejs_compat` for new projects furthers that goal.

We're going to keep making `nodejs_compat` better, more native to the Workers Runtime, and more comprehensive. And listen to what the vast majority of people want and expect when starting new projects. Disabling nodejs_compat is just one line of configuration to remove :)

---

#### Implementation approach

Instead of manually adding `nodejs_compat` to each template's wrangler.jsonc file, this change adds the flag automatically in C3's `updateWranglerConfig` function. This approach:

- Centralizes the logic in one place rather than duplicating across templates
- Handles both JSON/JSONC and TOML config formats
- Preserves any existing compatibility flags
- Skips adding `nodejs_compat` if any of the following are already present:
  - `nodejs_compat` (already enabled)
  - `nodejs_compat_v2` (newer version)
  - `no_nodejs_compat` (explicitly disabled)
- **Skips adding `nodejs_compat` for Python projects** since nodejs_compat is not compatible with Python workers

The Python check (`ctx.args.lang !== "python"`) is performed in `updateWranglerConfig` before calling the helper functions, making the control flow clearer.

---

- Tests
  - [x] Tests included/updated
  - [ ] Automated tests not possible - manual testing has been completed as follows:
  - [ ] Additional testing not necessary because:
- Public documentation
  - [ ] Cloudflare docs PR(s):
  - [x] Documentation not necessary because: This is a sensible default change that doesn't require user action or documentation updates

**Human review checklist:**
- [ ] Verify `addNodejsCompatFlag` and `addNodejsCompatFlagToToml` correctly handle edge cases (existing flags, nodejs_compat_v2, no_nodejs_compat)
- [ ] Confirm the flag ordering (nodejs_compat first, then existing flags) is acceptable
- [ ] Verify Python project detection (`ctx.args.lang !== "python"`) correctly identifies all Python templates

Devin PR requested by @petebacondarwin

Link to Devin run: https://app.devin.ai/sessions/7a606cdbe6d148a48d02975f027bf808

### DIFF
--- a/packages/create-cloudflare/src/wrangler/config.ts
+++ b/packages/create-cloudflare/src/wrangler/config.ts
@@ -68,7 +68,10 @@ export const updateWranglerConfig = async (ctx: C3Context) => {
 		wranglerJson = appendJSONProperty(wranglerJson, "observability", {
 			enabled: true,
 		});
-		wranglerJson = addNodejsCompatFlag(wranglerJson, ctx);
+		// Skip adding nodejs_compat for Python projects since it's not compatible with Python workers
+		if (ctx.args.lang !== "python") {
+			wranglerJson = addNodejsCompatFlag(wranglerJson);
+		}
 
 		addHintsAsJsonComments(wranglerJson);
 
@@ -88,7 +91,10 @@ export const updateWranglerConfig = async (ctx: C3Context) => {
 			ctx.project.path,
 		);
 		wranglerToml.observability ??= { enabled: true };
-		addNodejsCompatFlagToToml(wranglerToml, ctx);
+		// Skip adding nodejs_compat for Python projects since it's not compatible with Python workers
+		if (ctx.args.lang !== "python") {
+			addNodejsCompatFlagToToml(wranglerToml);
+		}
 
 		writeWranglerToml(
 			ctx,
@@ -328,21 +334,11 @@ function generateHintsAsTomlComments(wranglerConfig: TomlTable): string {
  * Adds the `nodejs_compat` flag to the `compatibility_flags` array in a JSON wrangler config.
  * If the array doesn't exist, it will be created. If `nodejs_compat`, `nodejs_compat_v2`,
  * or `no_nodejs_compat` is already present, no changes are made.
- * For Python projects, no changes are made since nodejs_compat is not compatible with Python workers.
  *
  * @param wranglerConfig The wrangler JSON configuration object.
- * @param ctx The C3 context.
  * @returns The updated configuration object.
  */
-function addNodejsCompatFlag(
-	wranglerConfig: CommentObject,
-	ctx: C3Context,
-): CommentObject {
-	// Skip adding nodejs_compat for Python projects
-	if (ctx.args.lang === "python") {
-		return wranglerConfig;
-	}
-
+function addNodejsCompatFlag(wranglerConfig: CommentObject): CommentObject {
 	const existingFlags = Array.isArray(wranglerConfig.compatibility_flags)
 		? (wranglerConfig.compatibility_flags as string[])
 		: [];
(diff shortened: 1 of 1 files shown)


anonrig (approved):

thomasgauvin:
I would lean towards all docs that require nodejs_compat to avoid errors should include it in the steps, even if nodejs_compat becomes a default, in case for some reason, someone has a cloudflare worker without nodejs compat enabled. So I would lean towards leaving https://developers.cloudflare.com/hyperdrive/get-started/#enable-nodejs-compatibility (maybe switch 'Enable' to 'Ensure') and I think we should still merge https://github.com/cloudflare/cloudflare-docs/pull/19860

Cherry:
Generally I love to see this. `nodejs_compat` in the runtime is a great addition and something I use on basically all my projects. I have two concerns that would be great to get some insight on though:

- https://github.com/cloudflare/workers-sdk/issues/7108 - today, the new `nodejs_compat` inflates basic worker sizes by almost 1000%. It's not a major problem, but it would be nice if this wasn't so inflated and looking at the output work was easier.

- And then the versioning/backwards compatibility story for `nodejs_compat`. As so much of this is baked into wrangler now via unenv, and as we recently saw in https://github.com/cloudflare/workers-sdk/pull/7942, I think `nodejs_compat` not being versioned is going to really cause issues/confusion/frustration in the future.

thomasgauvin (approved):

mikenomitch (approved):

anonrig (approved):

packages/create-cloudflare/templates/svelte/pages/templates/wrangler.jsonc
```diff
@@ -1,6 +1,6 @@
 {
   "name": "<WORKER_NAME>",
   "compatibility_date": "<COMPATIBILITY_DATE>",
-  "compatibility_flags": ["nodejs_als"],
+  "compatibility_flags": ["nodejs_compat", "nodejs_als"],
```
vicb: als (async local strorage) is a subset of node compat
```suggestion
  "compatibility_flags": ["nodejs_compat"],
```

packages/create-cloudflare/templates/svelte/workers/templates/wrangler.jsonc
```diff
@@ -2,7 +2,7 @@
   "name": "<WORKER_NAME>",
   "main": ".svelte-kit/cloudflare/_worker.js",
   "compatibility_date": "<COMPATIBILITY_DATE>",
-  "compatibility_flags": ["nodejs_als"],
+  "compatibility_flags": ["nodejs_compat", "nodejs_als"],
```
vicb: als (async local strorage) is a subset of node compat

```suggestion
  "compatibility_flags": ["nodejs_compat"],
```

packages/create-cloudflare/templates/waku/wrangler.jsonc
```diff
@@ -5,7 +5,7 @@
 	"compatibility_date": "<COMPATIBILITY_DATE>",
 	// nodejs_als is required for Waku server-side request context
 	// It can be removed if only building static pages
-	"compatibility_flags": ["nodejs_als"],
+	"compatibility_flags": ["nodejs_compat", "nodejs_als"],
```
vicb: Savm comment as before for als.
And please update the comment.
(Sorry GH UI wouldn't let me add a suggestion here)

vicb:
> Tests not necessary because: Don't think anything tests what compat flags enabled

I would rephrase to already tested - because there are trivial ways to tests what the flag enables

vicb:
> Generally I love to see this. `nodejs_compat` in the runtime is a great addition and something I use on basically all my projects. I have two concerns that would be great to get some insight on though:
> 
> * [🐛 BUG: Unnecessary polyfills? #7108](https://github.com/cloudflare/workers-sdk/issues/7108) - today, the new `nodejs_compat` inflates basic worker sizes (gzip) by almost 1000%. It's not a major problem, but it would be nice if this wasn't so inflated and looking at the output worker was easier.
> * And then the versioning/backwards compatibility story for `nodejs_compat`. As so much of this is baked into wrangler now via unenv, and as we recently saw in [remove functions that are removed from node:util #7942](https://github.com/cloudflare/workers-sdk/pull/7942), I think `nodejs_compat` not being versioned is going to really cause issues/confusion/frustration in the future.

Unnecessary polyfills -> we have removed a lot of them already and will keep removing more. "almost 1000%" is exaggerating, it mostly adds a constant size (a few kB).

I don't think versioning should be a concern. It is handled via the compat date. But please let us know if you have a specific (recent) concern. Thanks.

Cherry:
> I don't think versioning should be a concern. It is handled via the compat date. But please let us know if you have a specific (recent) concern. Thanks.

The versioning concerns were solved with `remove_nodejs_compat_eol_v24`, etc. flags, thanks!

penalosa (approved):

PR #8316, opened 2025-03-01
feat: :sparkles: introduce ip param for wrangler login command, fixes [5937]

### LINKED ISSUE #5937: 🐛 BUG: `wrangler login` only listening `::1`

Wrangler core

3.57.2

v20.14.0

Ubuntu 22.04.4

### Observed behavior
when running `wrangler login` inside a devcontainer wrangler only listens to `[::1]:8976` which break vscode's port-forwarding. connections from the host's browser just freeze.

for `wrangler dev` this can be worked around by adding a rule on `wrangler.toml`
```toml
[dev]
ip = "127.0.0.1"
```
but there is nothing of the sort for `wrangler login`, and `--ip` is not accepted either.

```
curl -v http://localhost:8976
*   Trying 127.0.0.1:8976...
* connect to 127.0.0.1 port 8976 failed: Connection refused
*   Trying ::1:8976...
* Connected to localhost (::1) port 8976 (#0)
> GET / HTTP/1.1
> Host: localhost:8976
> User-Agent: curl/7.81.0
> Accept: */*
```
### Expected behavior

listen on both stacks, `127.0.0.1` and `::1`.

### Steps to reproduce
* on vscode create a ubuntu based devcontainer
* `pnpm create cloudflare something`
* `cd something`
* `pnpm i`
* `pnpm wrangler login`

https://github.com/amery/issue-wrangler-5937

_No response_

gnekich:
fix issue #5937 by providing option to user to pass ip for the temporary login server, enabling the use of wrangler login command in devcontainers

Fixes #5937.

I would also argue that the default should be '0.0.0.0' for the values, as it is temporary login server, available for a short period. Making DX a lot better than to add --ip '0.0.0.0' every time in devcontainers, but for now PR defaults to localhost as that was the default before.

---

- Tests
  - [ ] maybe for TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because: any regression would be captured by pre-existing tests in `__tests__/user.test.ts`
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/21737
- V3 wrangler backport: not necessary because new feature

### DIFF
--- a/.changeset/salty-ties-find.md
+++ b/.changeset/salty-ties-find.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+introduce callback-host and callback-port param for wrangler login command
--- a/packages/wrangler/src/__tests__/user.test.ts
+++ b/packages/wrangler/src/__tests__/user.test.ts
@@ -80,6 +80,45 @@ describe("User", () => {
 			});
 		});
 
+		it("should login a user when `wrangler login` is run with custom callbackHost param", async () => {
+			mockOAuthServerCallback("success");
+
+			let counter = 0;
+			msw.use(
+				http.post(
+					"*/oauth2/token",
+					async () => {
+						counter += 1;
+
+						return HttpResponse.json({
+							access_token: "test-access-token",
+							expires_in: 100000,
+							refresh_token: "test-refresh-token",
+							scope: "account:read",
+						});
+					},
+					{ once: true }
+				)
+			);
+
+			await runWrangler("login --callback-host='0.0.0.0'");
+
+			expect(counter).toBe(1);
+			expect(std.out).toMatchInlineSnapshot(`
+				"Attempting to login via OAuth...
+				Temporary login server listening on 0.0.0.0:8976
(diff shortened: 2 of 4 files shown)


petebacondarwin (changes requested):
Thanks for putting up this PR @gnekich.
I think we should use `host` rather than `ip` throughout (e.g. `localhost` is not an IP address).

I also think we should be more clear what these args are - host/ip and port are quite ambiguous and overloaded terms. When I first saw this PR I thought you were changing the host:port of our OAuth endpoint.
How about we use the terms `callback-host` and `callback-port`?

Finally, we will need a changeset (with a minor version bump) and a test.

gnekich:
Thanks for the feedback, @petebacondarwin! I've updated the PR and renamed the arguments to 'callback-host' and 'callback-port' for clarity. Also added a changeset with a minor version bump and a test.

petebacondarwin (approved):

petebacondarwin:
Thanks @gnekich - I'll follow up with the docs updates etc and merge this later this month after DevWeek.

emily-shen (approved):

PR #8338, opened 2025-03-04
Upload metafiles for Workers Assets

GregBrimble:
Reverts cloudflare/workers-sdk#8336


petebacondarwin (approved):
This is the exact same code changes that was approved and merged previously.
We reverted it because it appeared there were consistent C3 e2e failures.
But we have been unable to reproduce those failures locally, nor in this new PR.
So I am going to merge this and it can go out in Thursdays release.

Issue #7852, opened 2025-01-21
Improve support for debugging in Vite plugin


jamesopstad:
Confirm that VSCode can use breakpoints in Workers. Document how to enable this.

See https://github.com/flarelabs-net/vite-plugin-cloudflare/issues/74 for more context.

PR #8357, opened 2025-03-05
feat: add new `unsafeInspectorProxy` option to miniflare

dario-piotrowicz:
This PR adds a new `unsafeInspectorProxy` option to miniflare.

The main reason why I worked on this is to improve the debugging dx of tooling that rely on miniflare only (since wrangler already has its own inspector proxy worker), in particular our vite plugin.

<details><summary>vite-plugin</summary>

### worker

With a few very minor updates in the vite-plugin code the vscode debugging experience with a single
worker seems to work very nicely! 😄 🚀 

Screenshot 2025-03-11 at 09 56 30

As well as debugging using chrome 😄 

Screenshot 2025-03-11 at 10 00 56

### multi-worker

debugging the multi-worker playground app also seems to be working quite nicely in chrome

Screenshot 2025-03-11 at 10 07 47

Also in vscode, if an appropriate `.vscode/launch.json` file is provided 😄 

Screenshot 2025-03-11 at 10 12 19

</details> 

For more details check the changeset file in the PR.

> [!Note]
> To validate the changes in this PR you can have a play with the example apps here:
>  &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;  https://github.com/dario-piotrowicz/miniflare-unsafe-inspector-proxy-playground

Partially addresses: https://github.com/cloudflare/workers-sdk/issues/7852 (and DEVX-1661)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this is an experimental API which can be documented later when we commit to it


packages/miniflare/src/plugins/core/index.ts
```diff
@@ -186,6 +186,8 @@ export const CoreSharedOptionsSchema = z.object({
 	httpsCertPath: z.string().optional(),
 
 	inspectorPort: z.number().optional(),
+	inspectorProxy: z.boolean().optional(),
```
dario-piotrowicz, line 189: Should this maybe be `experimentalInspectorProxy` or something? 🤔

penalosa: Miniflare usually uses `unsafe` prefixes for things like that

dario-piotrowicz: yeah I saw that but `unsafe` feels to me like something for things that should not generally be used but are potentially stable (like `unsafeEval`), but anyways if that's the pattern, and we do want to make this as not ready, I'm happy to go with `unsafeInspectorProxy` 🙂

dario-piotrowicz: ok I've renamed this to `unsafeInspectorProxy` as I think it'd be safer to make this as experimental somehow (I'd also be happy to make this clearer with `experimentalInspectorProxy` although I'm usually a big fan of consistency 🙂 )

dario-piotrowicz:
Note: I haven't implemented anything akin to this `PROXY_CONTROLLER_AUTH_SECRET` check that we have in the `InspectorProxyWorker`: https://github.com/cloudflare/workers-sdk/blob/a3527988e8849eab92b66cfb3a30334bef706b34/packages/wrangler/templates/startDevWorker/InspectorProxyWorker.ts#L90
since I don't think it's needed nor even really applicable here (since any generic inspector client should be able to establish connections with miniflare's inspector proxy)

If anyone disagrees and thinks that something like that is needed here please let me know 🙂

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-relay.ts
```diff
@@ -0,0 +1,161 @@
+import assert from "node:assert";
+import WebSocket from "ws";
+import { isDevToolsEvent } from "./devtools";
+import type {
+	DevToolsCommandRequests,
+	DevToolsEvent,
+	DevToolsEvents,
+} from "./devtools";
+
+export class InspectorProxyRelay {
```
dario-piotrowicz, line 10: I must admit that I am not too convinced of this class name, I chose "relay" since this is relaying messages to specific workers (probably I should add some comments here...)

But I'm really not sure of it... maybe something like `WorkerInspectorProxy` would be better? 🤔

any thoughts?

packages/miniflare/src/index.ts
```diff
@@ -1445,6 +1479,19 @@ export class Miniflare {
 		// all of `requiredSockets` as keys.
 		this.#socketPorts = maybeSocketPorts;
 
+		if (this.#maybeInspectorProxy !== undefined) {
+			// Try to get inspector port for worker
```
petebacondarwin, line 1483: Should this be?

```suggestion
			// Try to get inspector port for the workers
```

In other words, this is the port for the internal inspector on the workerd instance, which will be wrapped by the inspector proxy?

packages/miniflare/src/index.ts
```diff
@@ -1529,6 +1576,8 @@ export class Miniflare {
 		if (disposing) return new URL("http://[100::]/");
 		// Make sure `dispose()` wasn't called in the time we've been waiting
 		this.#checkDisposed();
+		// if there is an inspector proxy let's wait for it to be ready
+		await this.#maybeInspectorProxy?.ready;
```
petebacondarwin, line 1580: Should this go above the `checkDisposed()` call?

packages/miniflare/src/index.ts
```diff
@@ -1903,6 +1956,9 @@ export class Miniflare {
 			// Remove from instance registry as last step in `finally`, to make sure
 			// all dispose steps complete
 			maybeInstanceRegistry?.delete(this);
+
+			// Close the inspector proxy server if there is one
+			await this.#maybeInspectorProxy?.dispose();
```
petebacondarwin, line 1961: Should this go above the `maybeInstanceRegistry?.delete(this)` call?
From the comment above that statement it suggests that we don't want to remove the instance from the registry until all the steps have completed.

packages/miniflare/src/plugins/core/inspector-proxy/index.ts
```diff
@@ -0,0 +1,269 @@
...
+		const deferredPromise = new DeferredPromise<void>();
+		this.#server.close((err) => {
+			if (err) {
+				deferredPromise.reject(err);
+			} else {
+				deferredPromise.resolve();
+			}
+		});
+		return deferredPromise;
+	}
```
petebacondarwin, line 254: Out of interest why not use a normal Promise here?

```suggestion
		return new Promise((resolve, reject) => {
			this.#server.close((err) => err ? reject(err) : resolve());
		});
```

packages/miniflare/src/plugins/core/inspector-proxy/index.ts
```diff
@@ -0,0 +1,269 @@
...
+		return ++this.#runtimeMessageCounter;
+	}
+
+	getInspectorURL(): URL {
+		return getWebsocketURL(this.userInspectorPort);
+	}
+
+	async updateConnection(runtimeInspectorPort: number) {
+		const workerdInspectorJson = (await fetch(
+			`http://127.0.0.1:${runtimeInspectorPort}/json`
```
petebacondarwin, line 207: Is the runtime inspector always on 127.0.0.1?

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-relay.ts
petebacondarwin: Or just `WorkerInspector`? TBH I think the current name is fine.

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-relay.ts
```diff
@@ -0,0 +1,161 @@
...
+		});
+
+		this.#devtoolsWs.on("message", (data) => {
+			const msg = JSON.parse(data.toString());
+			assert(this.#runtimeWs?.OPEN);
+			this.#sendMessageToRuntime(msg);
+		});
+	}
+
+	#runtimeMessageCounter = 1e8;
```
petebacondarwin, line 68: Out of interest, why not start at 1?

packages/miniflare/test/plugins/core/inspector-proxy/index.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+			],
+		});
+		t.teardown(() => mf.dispose());
+
+		await mf.ready;
+
+		// when the inspector proxy is created there are some ws message exchanges between it
+		// and the runtime, these can interfere with testing and make tests flaky, so we just
+		// wait a bit here to avoid such problems
+		await setTimeout(30);
```
petebacondarwin, line 184: 30 seems a very small amount.

packages/miniflare/test/plugins/core/inspector-proxy/index.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+	t.is(msg.method, "Runtime.consoleAPICalled");
+	t.is(msg.params.type, "log");
+	t.is(msg.params.args.length, 1);
+	t.is(msg.params.args[0].type, "string");
+	t.is(msg.params.args[0].value, LARGE_STRING);
+
+	const res = await resPromise;
+	t.is(await res.text(), `body:${LARGE_STRING}`);
+
+	console.log = originalConsoleLog;
```
petebacondarwin, line 439: Should be in a teardown?

```suggestion
	t.teardown(() => {
		console.log = originalConsoleLog;
	});
```

petebacondarwin (approved):
Cool stuff @dario-piotrowicz - very exciting to see this stuff land.
I left a bunch of comments - none of which are blocking I think.

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-relay.ts
dario-piotrowicz: I'm not sure about `WorkerInspector` since having `Proxy` somewhere in the name of the class seems quite valuable here (since essentially this is a proxy)

actually! maybe this should actually be called `InspectorProxy` (since it is proxying an inspector channel) while what I currently named `InspectorProxy` could be called `InspectorProxyController` (or something like that)

Meaning that we have a single `InspectorProxyController` that controls one or more `InspectorProxy`s 🤔 (one per worker)

I think that this would be more correct and would also make things a bit clearer to me... what do you think? 🤔

.changeset/little-spiders-sell.md
```diff
@@ -0,0 +1,51 @@
+---
+"miniflare": minor
+---
+
+feat: add new `unsafeInspectorProxy` option to miniflare
+
+add a new `unsafeInspectorProxy` option to miniflare that established a proxy
```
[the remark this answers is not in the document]
dario-piotrowicz: oops, thanks for spotting the typo 😓

.changeset/little-spiders-sell.md
```diff
@@ -0,0 +1,51 @@
...
+example:
+
+```js
+import { Miniflare } from "miniflare";
+
+const mf = new Miniflare({
+	// the inspector proxy will be accessible through port 9229
+	inspectorPort: 9229,
+	// enable the inspector proxy
+	unsafeInspectorProxy: true,
```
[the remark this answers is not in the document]
dario-piotrowicz: yes I think we could 🤔 

however can that be a bit unintuitive?

Also the current implementation allows users to set the global `unsafeInspectorProxy` to true but not set that in any worker, basically hiding all the workers from all client inspectors, this would not be possible if we inferred this 🤔  (but it's also probably not something that useful to be honest... 😅)

So I am not completely sure 🤔, but I'm also not against the idea, are you confident that this would make for a better API? (PS: given the unsafe prefix I'm also assuming that we're not committing right now to this API anyways and that we can change it in breaking ways in patches/minors?)

packages/miniflare/src/index.ts
```diff
@@ -760,6 +764,29 @@ export class Miniflare {
...
+				);
+			}
+
+			const workerNamesToProxy = new Set(
+				this.#workerOpts
+					.filter(
+						({ core: { unsafeInspectorProxy } }) => !!unsafeInspectorProxy
+					)
+					.map((w) => w.core.name ?? "")
+			);
```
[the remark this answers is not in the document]
dario-piotrowicz: I think this is the same conversation as above? 🙂 

https://github.com/cloudflare/workers-sdk/pull/8357/files#r1989332546

> Could we just use workerNamesToProxy.size > 0 to decide whether we are enabling the unsafe inspector proxy?

Yeah I think we can just do that if you want 🙂 

> Conversely what does it mean to enable the proxy at the top level but not to enable any of the actual workers?

As I mentioned it can be used to effectively hide all workers from inspector clients (those are otherwise exposed by workerd at the specified inspector port), not sure if there is an actual use case for this though 😅

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-relay.ts
dario-piotrowicz: I'm not actually sure, I simply followed the pattern already present in the inspector proxy worker:
https://github.com/cloudflare/workers-sdk/blob/5109fdd0bd5d57f22915fdf5c8114faf687cb0d2/packages/wrangler/templates/startDevWorker/InspectorProxyWorker.ts#L374

my guess is that this does distinguish our logs from standard/read ones coming from devtools that do start with 1... if you want I can change this to 1, I think it would work pretty much the same 🤔

packages/miniflare/src/plugins/core/inspector-proxy/index.ts
dario-piotrowicz: I think so, this does suggest so:
https://github.com/cloudflare/workers-sdk/blob/5109fdd0bd5d57f22915fdf5c8114faf687cb0d2/packages/miniflare/src/index.ts#L1567

packages/miniflare/src/index.ts
dario-piotrowicz: I think so? honestly I am not sure about the difference, but it's probably a good idea to move it 🙂👍

packages/miniflare/src/index.ts
dario-piotrowicz: Yes, thanks for the fix 🙂 

> In other words, this is the port for the internal inspector on the workerd instance, which will be wrapped by the inspector proxy?

Yes exactly 🙂

packages/miniflare/src/index.ts
dario-piotrowicz: yeah, I think that's a good call 👍

packages/miniflare/src/plugins/core/inspector-proxy/index.ts
dario-piotrowicz: yeah you're totally right 😅

I was just using deferred promises and I think I had the the pattern stuck in my head when I implemented this, thanks! 🙂

packages/miniflare/test/plugins/core/inspector-proxy/index.spec.ts
dario-piotrowicz: yes but it seems to be sufficient, at least for me locally, if this turns out to be an issue in CI we can increase it but as long as it doesn't introduce flakyness the smaller the time the better, right?

packages/miniflare/test/plugins/core/inspector-proxy/index.spec.ts
dario-piotrowicz: of course! 😓 🤦

packages/miniflare/src/index.ts
dario-piotrowicz: moved 🙂👍

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-relay.ts
dario-piotrowicz: I've updated the names to `InspectorProxyController` and `InspectorProxy` 🙂

packages/miniflare/src/index.ts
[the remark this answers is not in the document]
dario-piotrowicz: I've applied the suggested change to infer the option 🙂

.changeset/little-spiders-sell.md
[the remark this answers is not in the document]
dario-piotrowicz: well, why not, I've applied the suggested change 🙂

dario-piotrowicz:
@petebacondarwin thanks a lot for your (as usual) awesome review 🙏 🫶 

I've applied all your suggested changes, there are a few other improvement/fixes that I missed before and I've applied after your review, sorry about that 😓🙇

It'd be awesome if you could have another look 🙂 
(either way I think we're getting there 🥳 ) 

___

PS: I didn't mention it before, but you can find some logic/code duplication between the files I'm introducing and the [InspectorProxyWorker](https://github.com/cloudflare/workers-sdk/blob/5109fdd0bd5d57f22915fdf5c8114faf687cb0d2/packages/wrangler/templates/startDevWorker/InspectorProxyWorker.ts), I don't think we currently have a clean/simple way to share such code, and I am hoping that this miniflare implementation can at some point just replace the wrangler inspector proxy worker anyways, so hopefully this is not a concern here, but if it is I'm happy to address it however you feel appropriate 🙂

petebacondarwin:
> I am hoping that this miniflare implementation can at some point just replace the wrangler inspector proxy worker anyways, so hopefully this is not a concern here, but if it is I'm happy to address it however you feel appropriate 🙂

Yes, let's focus on doing that in Q2.

.changeset/little-spiders-sell.md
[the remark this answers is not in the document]
petebacondarwin: > Also the current implementation allows users to set the global unsafeInspectorProxy to true but not set that in any worker, basically hiding all the workers from all client inspectors, this would not be possible if we inferred this 🤔 (but it's also probably not something that useful to be honest... 😅)

I think that this is the counter-intuitive bit about this.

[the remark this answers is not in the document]
petebacondarwin: > well, why not, I've applied the suggested change 🙂

Do you need to update this changeset then?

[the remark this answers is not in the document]
petebacondarwin: Another alternative would be to have an array of worker names at the top level?

```
unsafeWorkerInspectors: ["worker-a", "worker-b"]
```

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-relay.ts
petebacondarwin: No don't change it. I was just curious.

.changeset/little-spiders-sell.md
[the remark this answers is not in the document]
dario-piotrowicz: yeah that's fair 🤔

packages/miniflare/src/index.ts
```diff
@@ -764,23 +781,15 @@ export class Miniflare {
 
 		this.#log = this.#sharedOpts.core.log ?? new NoOpLog();
 
-		if (this.#sharedOpts.core.unsafeInspectorProxy !== undefined) {
+		if (enableInspectorProxy) {
 			if (this.#sharedOpts.core.inspectorPort === undefined) {
 				throw new MiniflareCoreError(
 					"ERR_MISSING_INSPECTOR_PROXY_PORT",
-					"inspectorProxy option enabled without an inspectorPort specified"
+					"inspector proxy requested but without an inspectorPort specified"
```
petebacondarwin, line 788: I am not sure if you brought this up already, but should we just default to 9229 if not provided by the user?

petebacondarwin (approved):

.changeset/little-spiders-sell.md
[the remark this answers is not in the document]
dario-piotrowicz: > Do you need to update this changeset then?

yes I already did that 🙂 
https://github.com/cloudflare/workers-sdk/pull/8357/files#diff-fde07d0e0e977f6160d9e0f1f0cdb50918b8bd0856ac1aa7bac9be18d4eff1a4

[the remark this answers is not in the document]
dario-piotrowicz: > Another alternative would be to have an array of worker names at the top level?

Yes, but there we could end up with some edge case situations, like:
 - there could be a worker without a name
 - user could include in `unsafeProxyWorkerInspectors` names of non-existing workers (or make typos)
 - user could include in `unsafeProxyWorkerInspectors` duplicate names

So we'd need to add some extra validation etc... so I am not completely sold on it 🤔 

It does feel a bit more intuitive to me (but less robust because of what I mentioned above), if you think that this would make for a better DX I'm happy to make the change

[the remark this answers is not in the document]
petebacondarwin: Happy to leave it as it is.

packages/miniflare/src/index.ts
dario-piotrowicz: No I haven't brought this up

I'm not completely sure, I feel like the current implementation is ok and maybe I would avoid having additional inferred behavior that can potentially confuse users

My worry is that the logic we'd need here would be that if the user is requesting the proxy then `inspectorPort` would default to `9229` otherwise to `undefined`, this feels a bit too inferred/unclear to me 🤔

Unless you're suggesting to always default miniflare to 9229, but in that case:
 - I am not sure if that's desired (why wasn't it already done in the first place then?)
 - I'm not sure if that would still allow people to opt-out from having the workerd inspector set up (I think that's valuable, if for example there are bugs in workerd etc...)
 - I _think_ this could be considered a breaking change? (for example users could have a project comprised of some nodejs and miniflare, in which they use port 9229 for the nodejs portion and don't specify any inspector port in miniflare, if we were to start defaulting to 9229 and they would bump their miniflare dependency they would start encountering port collision issues 🤔)
 
So I am not completely sold on it, although it would avoid us from having to have this validation step (and down the line it could allow us to remove the 9229 default inspector port from wrangler and the vite plugin)

WDYT?

.changeset/little-spiders-sell.md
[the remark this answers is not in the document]
dario-piotrowicz: ok thanks 🙂 

Given the `unsafe` prefix I also assume that we can hopefully easily change course later on if we want 🙂

penalosa (approved):
Some tiny nits, but let's ship it!

Long term we should either pull this out into a common lib for Wrangler and Miniflare, or ditch Wrangler's implementation in favour of this.

packages/miniflare/src/index.ts
dario-piotrowicz: As we discussed we can keep the current behavior and default to 9229 later on if we want 🙂

packages/miniflare/test/plugins/core/inspector-proxy/index.spec.ts
```diff
@@ -0,0 +1,462 @@
...
+			"ws://localhost:9212/extra-worker-b"
+		);
+	}
+);
+
+test.serial(
+	"InspectorProxy: should allow debugging a single worker",
+	async (t) => {
+		const mf = new Miniflare({
+			inspectorPort: 9212,
```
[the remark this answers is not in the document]
dario-piotrowicz: this was totally spot on, thanks a lot for the suggestion! 🫶

packages/miniflare/package.json
```diff
@@ -46,6 +46,7 @@
 		"acorn": "8.14.0",
 		"acorn-walk": "8.3.2",
 		"exit-hook": "2.2.1",
+		"get-port": "^7.1.0",
```
penalosa, line 49: This should be a `devDependency`, so that miniflare correctly bundles it

packages/miniflare/scripts/build.mjs
```diff
@@ -181,7 +183,9 @@ async function buildPackage() {
 			"miniflare",
 			// Mark `dependencies` as external, but not `devDependencies` (we use them
 			// to signal single-use/small packages we want inlined in the bundle)
-			...getPackageDependencies(pkg),
+			...getPackageDependencies(pkg).filter(
+				(dep) => !packagesToBundle.has(dep)
+			),
```
penalosa, line 188: This logic isn't needed—instead `get-port` should be a dev dep

packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-controller.ts
```diff
@@ -20,18 +21,22 @@ export class InspectorProxyController {
...
+	#inspectorPort: number | Promise<number>;
 
 	constructor(
-		private userInspectorPort: number,
+		userInspectorPort: number,
 		private log: Log,
 		private workerNamesToProxy: Set<string>
 	) {
+		this.#inspectorPort =
+			userInspectorPort !== 0 ? userInspectorPort : getPort();
```
penalosa, line 34: Is this definitely needed? I _think_ we should be able to pass `0` all the way to workerd, and have workerd resolve it?

dario-piotrowicz: But this doesn't have anything to do with workerd, we create a nodejs server that handles this communication (and proxies things to workerd): https://github.com/cloudflare/workers-sdk/blob/5b91294699196fcc484edfbc2a456163abb5a4f6/packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-controller.ts#L58

packages/miniflare/scripts/build.mjs
dario-piotrowicz: oops, sorry my bad, fixed 🙂

packages/miniflare/package.json
dario-piotrowicz: sorry my bad, fixed 🙂👍

penalosa (approved):

petebacondarwin (approved):
as long as you fix the formatting issues 😄

PR #8365, opened 2025-03-05
vite-plugin: update unenv preset dependency

petebacondarwin:
Fixes #0000

This fixes an issue with the latest version of the unenv-preset and updates the vite plugin to use it.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal update
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 
  - [x] Not necessary because: vite plugin is not backported

### DIFF
--- a/packages/unenv-preset/package.json
+++ b/packages/unenv-preset/package.json
@@ -20,7 +20,9 @@
 		"directory": "packages/unenv-preset"
 	},
 	"license": "MIT OR Apache-2.0",
-	"sideEffects": false,
+	"sideEffects": [
+		"./dist/runtime/polyfill/**"
+	],
 	"type": "module",
 	"exports": {
 		".": {
--- a/packages/unenv-preset/package.json
+++ b/packages/unenv-preset/package.json
@@ -20,9 +20,7 @@
 		"directory": "packages/unenv-preset"
 	},
 	"license": "MIT OR Apache-2.0",
-	"sideEffects": [
-		"./dist/runtime/polyfill/**"
-	],
+	"sideEffects": false,
 	"type": "module",
 	"exports": {
 		".": {
--- a/packages/vite-plugin-cloudflare/package.json
+++ b/packages/vite-plugin-cloudflare/package.json
@@ -42,11 +42,11 @@
 		"test:watch": "vitest"
 	},
 	"dependencies": {
-		"@cloudflare/unenv-preset": "1.1.1",
+		"@cloudflare/unenv-preset": "workspace:*",
 		"@hattip/adapter-node": "^0.0.49",
 		"miniflare": "workspace:*",
 		"tinyglobby": "^0.2.12",
-		"unenv": "2.0.0-rc.1",
+		"unenv": "2.0.0-rc.15",
 		"wrangler": "workspace:*",
 		"ws": "catalog:default"
 	},
--- a/.changeset/cold-baths-knock.md
+++ b/.changeset/cold-baths-knock.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+update vite-plugin to use the latest unenv-preset
--- a/packages/vite-plugin-cloudflare/src/node-js-compat.ts
+++ b/packages/vite-plugin-cloudflare/src/node-js-compat.ts
@@ -61,8 +61,14 @@ export function injectGlobalCode(id: string, code: string) {
 		})
 		.join("\n");
 
+	// Some globals are not injected using the approach above but are added to globalThis via side-effect imports of polyfills from the unenv-preset.
+	const polyfillCode = env.polyfill
+		.map((polyfillPath) => `import "${polyfillPath}";\n`)
+		.join("");
+
 	const modified = new MagicString(code);
 	modified.prepend(injectedCode);
+	modified.prepend(polyfillCode);
 	return {
 		code: modified.toString(),
 		map: modified.generateMap({ hires: "boundary", source: id }),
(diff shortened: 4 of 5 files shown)


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -412,6 +413,15 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 				// Only configure this environment if it is a Worker using Node.js compatibility.
 				if (isNodeCompat(getWorkerConfig(name))) {
 					return {
+						build: {
+							rollupOptions: {
+								treeshake: {
+									// We need to ensure that all the Node.js polyfill files that are simply imported with no exports
+									// are considered to have side-effects to avoid them being tree-shaken.
+									moduleSideEffects: nodeJSCompatHasSideEffect,
```
jamesopstad, line 421: I wonder if this would still be necessary if the unenv preset had `sideEffects: true` in its `package.json`?

If not, I think it would be better to use an array of modules here rather than a function so that it can be merged with user config.

```suggestion
									moduleSideEffects: [...nodeCompatPolyfills],
```

petebacondarwin: I forgot about side effects in package.json!
If that doesn't work I'll go with an array. The function was only for debugging really

petebacondarwin: `sideEffects` fixed it!

jamesopstad (approved):

pnpm-workspace.yaml
```diff
@@ -14,6 +14,7 @@ catalog:
   rimraf: "^5.0.10"
   typescript: "^5.7.2"
   undici: "^5.28.5"
+  unenv: "2.0.0-rc.8"
```
vicb, line 17: we should not do that before unenv is stable (2.0)

packages/unenv-preset/package.json
```diff
@@ -20,7 +20,9 @@
 		"directory": "packages/unenv-preset"
 	},
 	"license": "MIT OR Apache-2.0",
-	"sideEffects": false,
+	"sideEffects": [
```
vicb, line 23: Is this needed (there is a package.json in that folder)

petebacondarwin: Vite didn't pick it up.

petebacondarwin: Vite doesn't appear to pick that up.

petebacondarwin: I think it is because, as far as this library is concerned `dist/runtime/polyfill` is not an entry-point to the package?
We have just a general export in the root of the package for `./*`.

packages/unenv-preset/src/runtime/polyfill/performance.ts
```diff
@@ -1,30 +1,27 @@
...
-	//       But it is a class in the polyfill, see `runtime/node/perf_hooks.ts`
 	Performance,
 	PerformanceEntry,
 	PerformanceMark,
 	PerformanceMeasure,
 	PerformanceObserver,
 	PerformanceObserverEntryList,
 	PerformanceResourceTiming,
-} from "node:perf_hooks";
+} from "unenv/node/perf_hooks";
```
petebacondarwin, line 10: I'm going to check again to see if this change is still needed...

pnpm-workspace.yaml
vicb: The thing here is that we would need to test both Vite + Wrangler when updating unenv and I'm afraid it might be too much constraint while things are still moving (fast).

Once unenv@2.0 lands, the changes should have a smaller extent and moving to a catalog at that point might make more sense. We can also try and opt-out if needed.

packages/unenv-preset/src/runtime/polyfill/performance.ts
vicb: At least it is not needed any more on the wrangler side.

.changeset/mean-dingos-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/unenv-preset": patch
+---
+
+fix: avoid circular reference to aliased import in Performance polyfill
```
vicb, line 5: This should be removed if the change is no more needed (with the update to RC 14)

packages/vite-plugin-cloudflare/package.json
```diff
@@ -42,11 +42,11 @@
 		"test:watch": "vitest"
 	},
 	"dependencies": {
-		"@cloudflare/unenv-preset": "1.1.1",
+		"@cloudflare/unenv-preset": "workspace:*",
```
vicb, line 45: As discussed, this should be a fixed version until we run the tests against EW.

vicb (approved):
LGTM, a few minor things to check.
Thanks!

packages/unenv-preset/package.json
pi0: TIL `sideEffects` can be an array too. For unenv we use a package.json inside dist ([this](https://unpkg.com/browse/unenv@2.0.0-rc.14/dist/runtime/polyfill/package.json)) and it worked well.

vicb (approved):

packages/unenv-preset/src/runtime/polyfill/performance.ts
petebacondarwin: Confirmed that this is no longer needed.

.changeset/mean-dingos-check.md
petebacondarwin: Removed.

PR #8367, opened 2025-03-06
Change id parameter to name for both bucket lifecycle and r2 bucket lock commands

jonesphillip:
Change id parameter to name for both bucket lifecycle and r2 bucket lock commands

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: test changes covered by unit tests
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/20566

### DIFF
--- a/.changeset/whole-dots-smile.md
+++ b/.changeset/whole-dots-smile.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: for both the r2 bucket lifecycle and r2 bucket lock commands --id parameter deprecated in favor of --name
--- a/.changeset/whole-dots-smile.md
+++ b/.changeset/whole-dots-smile.md
@@ -1,5 +1,5 @@
 ---
-"wrangler": patch
+"wrangler": minor
 ---
 
 fix: for both the r2 bucket lifecycle and r2 bucket lock commands --id parameter deprecated in favor of --name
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/__tests__/r2.test.ts
```diff
@@ -2074,7 +2074,7 @@ describe("r2", () => {
 						)
 					);
 					await runWrangler(
-						`r2 bucket lifecycle add ${bucketName} --id ${ruleId} --prefix ${prefix} --expire-days ${conditionValue}`
+						`r2 bucket lifecycle add ${bucketName} --name ${ruleId} --prefix ${prefix} --expire-days ${conditionValue}`
 					);
```
jkoe-cf, line 2078: would it be worth having a test to add with --id as the flag for name to make sure we keep backward compatibility?

jkoe-cf (approved):

elithrar (approved):

jkoe-cf (approved):

.changeset/whole-dots-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
```
edmundhung, line 2:
```suggestion
"wrangler": minor
```

I think this should be considered a feature rather than a fix..?

edmundhung (approved):

PR #8373, opened 2025-03-06
Add `CF-Cache-Status` to Workers Assets

WalshyDev:
Fixes N/A

Attach a CF-Cache-Status header to asset responses. We currently do not have a way to get this status from KV so we're assuming HIT if it's <= 100 ms. This will mark ~99.5% of requests today as cache HIT which is probably about right.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no change to behaviour - though we may changelog but not a blocker

### DIFF
--- a/.changeset/thirty-emus-find.md
+++ b/.changeset/thirty-emus-find.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-shared": minor
+---
+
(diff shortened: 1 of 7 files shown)


WillTaylorDev (approved):
This looks great! I think this is a big win :)

PR #8375, opened 2025-03-06
Email Workers  local dev and send_email binding

penalosa:
Fixes DEVX-1731, EMAIL-1147.
[DevWeek CR ](https://jira.cfdata.org/browse/CR-1147486)

This PR adds support for local dev for email handler by sending a POST request to `/cdn-cgi/handler/email` where the body contains a valid IMF message ([RFC 2822](https://www.rfc-editor.org/rfc/rfc2822.html#section-3.6.4) compliant) and accepts two search parameters `from` and `to` to simulate SMTP envelope from/to. 

It also adds support for the send_email binding.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: it will be announced later
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR:


fixtures/worker-ts/package.json
penalosa, line 1: This should be in a new fixture specifically for email workers

packages/miniflare/package.json
```diff
@@ -47,6 +47,7 @@
 		"acorn-walk": "8.3.2",
 		"exit-hook": "2.2.1",
 		"glob-to-regexp": "0.4.1",
+		"postal-mime": "^2.4.3",
```
penalosa, line 50: This should be a `devDependency` instead (so that it's bundled into the Miniflare release)

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,105 @@
...
+						internal: true,
+					},
+				],
+			});
+		}
+
+		const hasSendEmail =
+			args.options.email?.send_email?.length !== undefined &&
+			args.options.email?.send_email?.length > 0;
+		// we only want to insert on the first worker as it will be shared between them
```
penalosa, line 87: You shouldn't need the workerIndex check here, since the other workers won't have `hasSendEmail`, I don't think

packages/miniflare/src/workers/email/email.worker.ts
```diff
@@ -0,0 +1,26 @@
...
+	public constructor(
+		public readonly from: string,
+		public readonly to: string,
+		private raw: ReadableStream | string
+	) {
+		// @ts-expect-error This is a bit of hack! We need:
+		// - EmailMessage to be constructable (to match production)
+		// - EmailMessage to be able to be passed across JSRPC (to support e.g. message.reply(EmailMessage))
+		// - EmailMessage properties to be synchronously available (to match production). This rules out `RpcStub`
+		// The below is the only solution I could some up with that satisfies these contraints, but if the contrainsts change
```
penalosa, line 14:
```suggestion
		// The below is the only solution I could some up with that satisfies these constraints, but if the constraints change
```

packages/miniflare/src/workers/email/send_email.worker.ts
```diff
@@ -0,0 +1,84 @@
...
+		console.log(
+			`${blue("send_email binding called with the following message:")}\n  ${file}`
+		);
+	}
+}
+
+// create a new Ratelimit
+export default function (env: SendEmailEnv) {
+	return new SendEmailBinding(env);
+}
```
penalosa, line 84: Does this need to be a wrapped binding? It seems like a perfect use case for a JSRPC binding

packages/miniflare/src/workers/email/validators.worker.ts
```diff
@@ -0,0 +1,193 @@
+import { EmailMessage } from "cloudflare:email";
+import { blue } from "kleur/colors";
+import PostalMime, { Email } from "postal-mime";
+import { RAW_EMAIL } from "./constants";
+
+/**
+ *
+ * @param input Header value for Message-Id, References, In-Reply-To, and other header values that might contain Message-id style text.
+ */
+export function parseMessageIdStyleValue(input: string): string[] | undefined {
```
penalosa, line 10: What is this doing?

packages/miniflare/src/workers/email/validators.worker.ts
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		// if we ever reach here, means that there was some invalid input and we return undefined
+		return undefined;
+	}
+
+	return output;
+}
+
+export function isEmailRepliable(
```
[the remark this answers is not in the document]
penalosa: Could you add a comment explaining the logic here?

packages/miniflare/src/plugins/email/index.ts
LuisDuarte1: I'm not sure how multi worker environments work under miniflare, but my assumption was that the other workers can have the `send_email` binding, and it seemed reasonable to have the same constraint as the `EmailMessage` module

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -343,6 +347,143 @@ async function handleScheduled(
...
+		// up 1MiB. Headers will probably be the part that takes up most space, but it's rare for emails headers to be that big.
+		createForwardableEmailMessage(
+			from,
+			to,
+			new ReadableStream({
+				start(controller) {
+					// probably not the most efficient way to do this?
+					controller.enqueue(incomingEmailRaw);
+				},
+			}),
```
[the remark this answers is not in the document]
LuisDuarte1: The incoming request body was already consumed around line 387, but maybe it's possible to shorten this out by creating a response from the blob I would guess

packages/miniflare/src/workers/email/send_email.worker.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export class SendEmailBinding {
+	#env: SendEmailEnv;
+
+	constructor(env: SendEmailEnv) {
+		this.#env = env;
+	}
+
+	async send(emailMessage: EmailMessage): Promise<void> {
```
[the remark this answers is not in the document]
LuisDuarte1: atm from `worker-types` maybe I should change it to the miniflare one :sweat_smile:

packages/miniflare/src/workers/email/validators.worker.ts
LuisDuarte1: `postal-mime` doesn't fully parse the header values to `References` and `In-Reply-To`, so I had to parse it myself, for validation - usually they have one/multiple message-ids in the value in a specific format so a parser is kind of needed here.

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -706,6 +706,12 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 			bindings.hyperdrive?.map(hyperdriveEntry) ?? []
 		),
 		workflows: Object.fromEntries(bindings.workflows?.map(workflowEntry) ?? []),
+		email: {
+			// @ts-expect-error `type` is there, somehow?
```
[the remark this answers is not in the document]
LuisDuarte1: This was me ruling out stuff, maybe I can remove this now :)

patches/postal-mime.patch
```diff
@@ -0,0 +1,10 @@
+diff --git a/postal-mime.d.ts b/postal-mime.d.ts
+index bf8dc08b669a7f78cb33f719d45435a6940e8996..953d9f81d670430bb20ef467193f34bb37eafc97 100644
+--- a/postal-mime.d.ts
++++ b/postal-mime.d.ts
+@@ -1,4 +1,4 @@
+-export type RawEmail = string | ArrayBuffer | Uint8Array | Blob | Buffer | ReadableStream;
++export type RawEmail = string | ArrayBuffer | Uint8Array | Blob | ReadableStream;
```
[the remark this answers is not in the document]
LuisDuarte1: Although `postal-mime` says it works on the browser/workerd, `RawEmail` contains a `Buffer`, which is a node.js thing so TS wasn't too happy with this on the `.worker.ts` files if I remember it correctly. 

I can upstream a fix later, the fix probably isn't that trivial since people using node should have `Buffer` included?

packages/miniflare/src/workers/email/send_email.worker.ts
LuisDuarte1: I can try :) - I was just piggybacking on the rate-limit implementation

packages/miniflare/src/workers/email/validators.worker.ts
penalosa: Is there a simpler way to parse this (regex, maybe?)

packages/miniflare/src/workers/core/entry.worker.ts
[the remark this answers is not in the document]
penalosa: You should be able to clone the request body to make this work?

packages/miniflare/src/plugins/email/index.ts
LuisDuarte1: Now I use services instead of extensions, so it should be fine?

packages/miniflare/src/workers/email/send_email.worker.ts
LuisDuarte1: Done

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -181,7 +181,9 @@ export function printBindings(
 						value: addSuffix(
 							destination_address ||
 								allowed_destination_addresses?.join(", ") ||
-								"unrestricted"
+								"unrestricted",
+							// the send_email binding works in --remote but also in local dev, so we take the context.local value
+							{ isSimulatedLocally: context.local ?? true }
```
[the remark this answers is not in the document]
LuisDuarte1: agh, my bad, I misread that

packages/miniflare/src/workers/email/validators.worker.ts
LuisDuarte1: I don't think a regex can fit the bill here - it can match the input but cannot extract everything since capture groups only store the last match

emily-shen (commented):
just some stray comments

fixtures/email-worker/wrangler.toml
```diff
@@ -0,0 +1,18 @@
+name = "email-to-rss"
+main = "src/index.ts"
+compatibility_date = "2025-02-14"
+
+compatibility_flags = [ "nodejs_compat", "experimental" ]
```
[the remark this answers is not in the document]
LuisDuarte1: Yeah we can remove it :) - i'll just leave the nodejs_compat flag

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -216,6 +216,9 @@ export const CoreSharedOptionsSchema = z.object({
 	unsafeStickyBlobs: z.boolean().optional(),
 
 	unsafeEnableAssetsRpc: z.boolean().optional(),
+
+	// Enable directly triggering user-worker handlers with paths like `/cdn-cgi/handler/scheduled`
```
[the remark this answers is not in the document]
LuisDuarte1: Thank @penalosa :)

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,112 @@
...
+						internal: true,
+					},
+				],
+			});
+		}
+
+		const hasSendEmail =
+			args.options.email?.send_email?.length !== undefined &&
+			args.options.email?.send_email?.length > 0;
+		// we only want to insert on the first worker as it will be shared between them since it's a service?
```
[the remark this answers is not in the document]
LuisDuarte1: Yeah, I don't think this applies atm, just a left over from doing this from extensions

packages/miniflare/src/workers/email/email.worker.ts
```diff
@@ -0,0 +1,34 @@
+import { RAW_EMAIL } from "./constants";
```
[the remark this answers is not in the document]
LuisDuarte1: they have the `.worker.ts` prefix because I think that's how miniflare's `tsconfig.json` distinguish something that uses workerd or node, should I've done it differently?

penalosa (commented):
Pending tests, this looks mostly good to me.

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Welcome to Cloudflare Workers! This is your first worker.
+ *
+ * - Run `wrangler dev src/index.ts` in your terminal to start a development server
+ * - Open a browser tab at http://localhost:8787/ to see your worker in action
+ * - Run `wrangler deploy src/index.ts --name my-worker` to deploy your worker
+ *
+ * Learn more at https://developers.cloudflare.com/workers/
+ */
```
dario-piotrowicz, line 13: I'd be tempted to say that this should be removed? 🤔

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export default class extends WorkerEntrypoint<Env> {
+	async fetch(request: Request): Promise<Response> {
+		const url = new URL(request.url);
+		if (url.pathname === "/error") throw new Error("Hello Error");
+
+		if (url.pathname === "/send") {
+			const msg = createMimeMessage();
+			// msg.setHeader("In-Reply-To", message.headers.get("Message-ID"));
```
dario-piotrowicz, line 22: is this commented out line intentionally left here?

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -210,6 +210,10 @@ export const CoreSharedOptionsSchema = z.object({
 	unsafeModuleFallbackService: ServiceFetchSchema.optional(),
 	// Keep blobs when deleting/overwriting keys, required for stacked storage
 	unsafeStickyBlobs: z.boolean().optional(),
+
```
dario-piotrowicz, line 213:
```suggestion
```

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+			msg.setSender({ name: "GPT-4", addr: "sender@penalosa.cloud" });
+			msg.setRecipient("else@exmaple.com");
+			msg.setSubject("An email generated in a worker");
+			msg.addMessage({
+				contentType: "text/plain",
+				data: `Congratulations, you just sent an email from a worker.`,
```
dario-piotrowicz, line 28: Nit 😅 
```suggestion
				data: "Congratulations, you just sent an email from a worker.",
```

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+		msg.setHeader("In-Reply-To", message.headers.get("Message-ID")!);
+		msg.setSender({ name: "Sender", addr: "sender@penalosa.cloud" });
+		msg.setRecipient(message.from);
+		msg.setSubject("An email generated in a worker");
+		msg.addMessage({
+			contentType: "text/plain",
+			data: `Congratulations, you just sent an email from a worker.`,
+		});
+
+		var m = new EmailMessage(
```
dario-piotrowicz, line 51:
```suggestion
		const m = new EmailMessage(
```

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,67 @@
...
+			const msg = createMimeMessage();
+			// msg.setHeader("In-Reply-To", message.headers.get("Message-ID"));
+			msg.setSender({ name: "GPT-4", addr: "sender@penalosa.cloud" });
+			msg.setRecipient("else@exmaple.com");
+			msg.setSubject("An email generated in a worker");
+			msg.addMessage({
+				contentType: "text/plain",
+				data: `Congratulations, you just sent an email from a worker.`,
+			});
+			var m = new EmailMessage(
```
dario-piotrowicz, line 30:
```suggestion
			const m = new EmailMessage(
```

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,112 @@
...
+					worker: {
+						compatibilityDate: "2025-03-17",
+						modules: [
+							{
+								name: "send_email.mjs",
+								esModule: SEND_EMAIL_BINDING(),
+							},
+						],
+						bindings: [
+							...buildJsonBindings(config),
```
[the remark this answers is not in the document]
penalosa: I'm not really sure what you mean here—how is validation related to these bindings?

packages/miniflare/test/index.spec.ts
```diff
@@ -1420,6 +1420,273 @@ test("Miniflare: manually triggered scheduled events", async (t) => {
...
+		"Email worker rejected email with the following reason: I just don't like this email :("
+	);
+	t.is(res.status, 400);
+
+	res = await mf.dispatchFetch("http://localhost");
+	t.is(await res.text(), "true");
+});
+
+test("Miniflare: manually triggered email handler - forward does not throw", async (t) => {
+	const log = new TestLog(t);
```
penalosa, line 1510: Can you validate that the correct messages are logged using this test log?

fixtures/email-worker/src/index.ts
LuisDuarte1: Eh, it's just leftovers from the template - i can remove it

dario-piotrowicz: yeah it's no huge deal either way 😅👍

fixtures/email-worker/src/index.ts
LuisDuarte1: It was me just testing out stuff, I'll uncomment it - nice catch :)

packages/miniflare/src/plugins/email/index.ts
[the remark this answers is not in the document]
LuisDuarte1: I send `destination_address` and `allowed_destination_addresses` in the config - in the send_email worker I do the validation comparing these addresses to the email message.

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -1,3 +1,5 @@
+import { ForwardableEmailMessage } from "@cloudflare/workers-types";
+import { EmailMessage } from "cloudflare:email";
```
dario-piotrowicz, line 2: I think this import is unnecessary
```suggestion
```

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +356,156 @@ async function handleScheduled(
...
+	params: URLSearchParams,
+	request: Request,
+	service: Fetcher,
+	env: Env,
+	_ctx: ExecutionContext
+): Promise<Response> {
+	// `from` and `to` are the equivalent ones to the SMTP RCPT commands (not the ones in the email message itself)
+	const from = params.get("from") ?? "";
+	const to = params.get("to") ?? "";
+	// clone request as we need to the body to be cloned to forward it to the inside of the worker
```
dario-piotrowicz, line 390:
```suggestion
	// clone request as we need the body to be cloned to forward it to the inside of the worker
```

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +356,156 @@ async function handleScheduled(
...
+		const error = e as Error;
+		return new Response(
+			`Email could not be parsed: ${error.name}: ${error.message}`,
+			{ status: 400 }
+		);
+	}
+
+	if (parsedIncomingEmail.messageId === undefined) {
+		return new Response(
+			`Email could not be parsed: invalid or no message id provided`,
```
dario-piotrowicz, line 419:
```suggestion
			"Email could not be parsed: invalid or no message id provided",
```

packages/miniflare/test/index.spec.ts
LuisDuarte1: Hum I don't think I can do that, `log.logs` returns this and not the logs sent from the entry worker:  

```js
[
  [ 3, 'Ready on http://127.0.0.1:41313' ],
  [ 3, 'GET / 200 OK (2ms)' ],
  [ 3, 'GET / 200 OK (1ms)' ]
]
```

penalosa: Fair enough—sounds like an E2E test job then

packages/miniflare/src/workers/email/validators.worker.ts
```diff
@@ -0,0 +1,203 @@
...
+import PostalMime, { Email } from "postal-mime";
+import { RAW_EMAIL } from "./constants";
+import { type MiniflareEmailMessage as EmailMessage } from "./email.worker";
+
+/**
+ *
+ * @param input Header value for Message-Id, References, In-Reply-To, and other header values that might contain Message-id style text.
+ */
+export function parseMessageIdStyleValue(input: string): string[] | undefined {
+	// probably not quite RFC-2822 compliant because we don't explitly support comments,
```
dario-piotrowicz, line 11:
```suggestion
	// probably not quite RFC-2822 compliant because we don't explicitly support comments,
```

packages/miniflare/src/workers/email/validators.worker.ts
```diff
@@ -0,0 +1,203 @@
...
+
+	expectedReferences.push(parsedIncomingMessageId);
+
+	// if the email has the references header, we just validate them.
+	if (parsedReply.references !== undefined) {
+		const parsedReferences = parseMessageIdStyleValue(parsedReply.references);
+
+		if (
+			parsedReferences === undefined ||
+			parsedReferences !== expectedReferences
```
dario-piotrowicz, line 181: I don't think that this check is correct

One thing is that [`expectedReferences` is guaranteed not to be `undefined`](https://github.com/cloudflare/workers-sdk/pull/8375/files#diff-3cd2d33b2ad407cc27b9fdffafcb8cf8838892882b05dabd83f3de8c5fa26513R164), so the `parsedReferences === undefined` check is unnecessary

The second thing (unless I'm missing something) is that I think that `parsedReferences !== expectedReferences` is always going to be `true` since you're comparing two different arrays (by reference), and that doesn't work in js:
Screenshot 2025-03-21 at 16 39 48
Screenshot 2025-03-21 at 16 52 43

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export const EMAIL_PLUGIN = createPlugin({
+	options: z.object({ email: EmailOptionsSchema.optional() }),
+	getBindings(options): Worker_Binding[] {
+		if (!options.email?.send_email) {
+			return [];
+		}
+
+		const sendEmailBindings = options.email?.send_email ?? [];
```
dario-piotrowicz, line 52: doesn't the line above guarantee that this is truthy?

```suggestion
		const sendEmailBindings = options.email.send_email;
```

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,111 @@
...
+						esModule: EMAIL_MESSAGE(),
+						internal: true,
+					},
+				],
+			});
+		}
+
+		const hasSendEmail =
+			args.options.email?.send_email?.length !== undefined &&
+			args.options.email?.send_email?.length > 0;
```
dario-piotrowicz, line 83:
```suggestion
			args.options.email.send_email.length > 0;
```

packages/miniflare/src/workers/email/validators.worker.ts
```diff
@@ -0,0 +1,203 @@
+import { blue } from "kleur/colors";
+import PostalMime, { Email } from "postal-mime";
+import { RAW_EMAIL } from "./constants";
+import { type MiniflareEmailMessage as EmailMessage } from "./email.worker";
+
+/**
+ *
```
dario-piotrowicz, line 7:
```suggestion
```

packages/miniflare/src/workers/email/validators.worker.ts
LuisDuarte1: parsedReferences **can** be undefined, parsing the references value from the email might fail. 

Thanks about the array equality, JS is weird sometimes

packages/miniflare/src/workers/email/validators.worker.ts
dario-piotrowicz: What about something like this? 🤔 
```ts
function parseMessageIdStyleValue(str: string): string[] | undefined {
  if (!/^(?:\s*<([^>]+)>\s*)+$/.test(str)) {
    return undefined;
  }
  const match = str.match(/<[^>]+>/g);
  return match!.map((str) => str.slice(1, -1));
}
```

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		return new Response("Hello World!");
+	}
+	async email(message: ForwardableEmailMessage) {
+		const msg = createMimeMessage();
+		msg.setHeader("In-Reply-To", message.headers.get("Message-ID")!);
+		msg.setSender({ name: "Sender", addr: "sender@penalosa.cloud" });
+		msg.setRecipient(message.from);
+		msg.setSubject("An email generated in a worker");
```
CarmenPopoviciu, line 34:
```suggestion
		msg.setSubject("An email generated in a Worker");
```

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+		if (url.pathname === "/error") throw new Error("Hello Error");
+
+		if (url.pathname === "/send") {
+			const msg = createMimeMessage();
+			msg.setSender({ name: "GPT-4", addr: "sender@penalosa.cloud" });
+			msg.setRecipient("else@exmaple.com");
+			msg.setSubject("An email generated in a worker");
+			msg.addMessage({
+				contentType: "text/plain",
+				data: "Congratulations, you just sent an email from a worker.",
```
CarmenPopoviciu, line 17:
```suggestion
				data: "Congratulations, you just sent an email from a Worker.",
```

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+export default class extends WorkerEntrypoint<Env> {
+	async fetch(request: Request): Promise<Response> {
+		const url = new URL(request.url);
+		if (url.pathname === "/error") throw new Error("Hello Error");
+
+		if (url.pathname === "/send") {
+			const msg = createMimeMessage();
+			msg.setSender({ name: "GPT-4", addr: "sender@penalosa.cloud" });
+			msg.setRecipient("else@exmaple.com");
+			msg.setSubject("An email generated in a worker");
```
CarmenPopoviciu, line 14:
```suggestion
			msg.setSubject("An email generated in a Worker");
```

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -210,6 +210,8 @@ export const CoreSharedOptionsSchema = z.object({
 	unsafeModuleFallbackService: ServiceFetchSchema.optional(),
 	// Keep blobs when deleting/overwriting keys, required for stacked storage
 	unsafeStickyBlobs: z.boolean().optional(),
+	// Enable directly triggering user-worker handlers with paths like `/cdn-cgi/handler/scheduled`
```
CarmenPopoviciu, line 213:
```suggestion
	// Enable directly triggering user Worker handlers with paths like `/cdn-cgi/handler/scheduled`
```

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,111 @@
...
+					{
+						name: "cloudflare-internal:email",
+						esModule: EMAIL_MESSAGE(),
+						internal: true,
+					},
+				],
+			});
+		}
+
+		const hasSendEmail =
```
CarmenPopoviciu, line 81: we can simplify this whole construct with the following:
```
if(args.options?.email?.send_email?.length) {
    const sendEmailOptions = args.options.email?.send_email;

    ....
}
```

If `email` is always expected on `options`, then the above would become `if(args.options.email?.send_email?.length){....}`

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -398,8 +552,20 @@ export default <ExportedHandler<Env>>{
 		}
 
 		try {
-			if (url.pathname === "/cdn-cgi/mf/scheduled") {
-				return await handleScheduled(url.searchParams, service);
+			if (env[CoreBindings.TRIGGER_HANDLERS]) {
```
CarmenPopoviciu, line 555: will this break how `/cdn-cgi/mf/scheduled` used to work prior to this PR? I don't fully understand the impact of moving it behind this conditional

packages/miniflare/src/plugins/email/index.ts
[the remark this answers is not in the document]
CarmenPopoviciu: can't this validation live in wrangler? why does it have to live in Miniflare? what kind of validation is it?

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
 	});
 }
 
+function createForwardableEmailMessage(
```
CarmenPopoviciu, line 358: can we please extract these functions into a separate file? They are quite extensive and add to the cognitive load of this file 
🙏

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+}
+
+async function handleEmail(
+	params: URLSearchParams,
+	request: Request,
+	service: Fetcher,
+	env: Env,
+	_ctx: ExecutionContext
+): Promise<Response> {
+	// `from` and `to` are the equivalent ones to the SMTP RCPT commands (not the ones in the email message itself)
```
CarmenPopoviciu, line 386: "SMTP RCPT commands" I dunno what this means. Can we add a link or ref to docs to help readers have more context here

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+	const from = params.get("from") ?? "";
+	const to = params.get("to") ?? "";
+	// clone request as we need the body to be cloned to forward it to the inside of the worker
```
CarmenPopoviciu, line 389: forward it to the `inside of the worker`. I don't understand what this means. Which Worker? What's the inside of a Worker?

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+	if (!request.body || !clonedRequest.body) {
+		return new Response("Include an email in the body", { status: 400 });
+	}
+
+	// Get request blob as we need its size
+	const incomingEmailRaw = await (await request.blob()).bytes();
+
+	// Email worker limit is 25 MiB total
+	if (incomingEmailRaw.byteLength > 25 * 1024 * 1024) {
+		return new Response("Email is bigger than current limit - 25MiB.", {
```
CarmenPopoviciu, line 400:
```suggestion
		return new Response("Email contents is larger than the current limit of 25MiB.", {
```
would appreciate a double check on the wording here from a native speaker

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+	// clone request as we need the body to be cloned to forward it to the inside of the worker
+	const clonedRequest = request.clone();
+	if (!request.body || !clonedRequest.body) {
+		return new Response("Include an email in the body", { status: 400 });
+	}
+
+	// Get request blob as we need its size
+	const incomingEmailRaw = await (await request.blob()).bytes();
+
+	// Email worker limit is 25 MiB total
```
CarmenPopoviciu, line 398:
```suggestion
	// Email Routing does not support messages bigger than 25 MiB
	// see https://developers.cloudflare.com/email-routing/limits/#message-size
```

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+	if (parsedIncomingEmail.messageId === undefined) {
+		return new Response(
+			"Email could not be parsed: invalid or no message id provided",
+			{ status: 400 }
+		);
+	}
+
+	if (from !== parsedIncomingEmail.from.address) {
+		console.log(
+			`${yellow("Provided envelope from address doesn't match message From header")} - Envelope From: ${from}; Header From: ${parsedIncomingEmail.from.address}`
```
CarmenPopoviciu, line 425: I haven't run `wrangler dev` against this PR, so not sure how this will look printed, but are we sure this is the format we want for this log? Is this readable for users?

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+		);
+	} catch (e) {
+		const error = e as Error;
+		return new Response(
+			`Email could not be parsed: ${error.name}: ${error.message}`,
+			{ status: 400 }
+		);
+	}
+
+	let maybeClientError: string | undefined = undefined;
```
CarmenPopoviciu, line 448:
```suggestion
	let maybeClientError: string | undefined;
```

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+
+	// @ts-expect-error .email is not in the fetcher but it's a valid RPC call.
+	await service.email(
+		// FIXME(lduarte): I have a feeling that because we are using RPC to do this, from + to + headers + readableStream can only take
```
CarmenPopoviciu, line 452: are we tracking this `fixme` item? Also, can that fix land in a followup PR? Is that also part of the dev week deliverable? Just trying to understand

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+				maybeClientError = reason;
+			},
+			async function forward(rcptTo: string, headers?: Headers): Promise<void> {
+				console.log(
+					`${blue(".forward() called from Email Handler")} with\n  rcptTo: ${rcptTo}${headers ? `\n  headers:\n${[...headers.entries()].map(([k, v]) => `    ${k}: ${v}`).join("\n")}` : ""}`
+				);
+			},
+			async function reply(replyMessage: EmailMessage): Promise<void> {
+				if (!isEmailReplyable(parsedIncomingEmail, incomingEmailHeaders)) {
+					throw new Error("original email is not repliable");
```
CarmenPopoviciu, line 472:
```suggestion
					throw new Error("Original email is not replyable");
```

Will users know what this means? What the action here for them to fix this error?

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+				console.log(
+					`${blue(".reply() called from Email Handler with the following message:")}\n  ${file}`
+				);
+			}
+		)
+	);
+
+	if (maybeClientError !== undefined) {
+		return new Response(
+			`Email worker rejected email with the following reason: ${maybeClientError}`,
```
CarmenPopoviciu, line 498:
```suggestion
			`Email Worker rejected email with the following reason: ${maybeClientError}`,
```

packages/miniflare/src/workers/core/entry.worker.ts
CarmenPopoviciu, line 425: I find this log a bit confusing, and can't fully follow what it's trying to tell me. Is there a better wording we can use?

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+		// FIXME(lduarte): I have a feeling that because we are using RPC to do this, from + to + headers + readableStream can only take
+		// up 1MiB. Headers will probably be the part that takes up most space, but it's rare for emails headers to be that big.
+		createForwardableEmailMessage(
+			from,
+			to,
+			clonedRequest.body,
+			incomingEmailHeaders,
+			function setReject(reason: string): void {
+				console.log(
+					`${red(".setReject() called from Email Handler")} with the following reason: "${reason}"`
```
CarmenPopoviciu, line 461:
```suggestion
					`${red("Email handler rejected message")} with the following reason: "${reason}"`
```

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+				);
+				maybeClientError = reason;
+			},
+			async function forward(rcptTo: string, headers?: Headers): Promise<void> {
+				console.log(
+					`${blue(".forward() called from Email Handler")} with\n  rcptTo: ${rcptTo}${headers ? `\n  headers:\n${[...headers.entries()].map(([k, v]) => `    ${k}: ${v}`).join("\n")}` : ""}`
```
CarmenPopoviciu, line 467:
```suggestion
					`${blue("Email handler forwarded message")} with\n  rcptTo: ${rcptTo}${headers ? `\n  headers:\n${[...headers.entries()].map(([k, v]) => `    ${k}: ${v}`).join("\n")}` : ""}`
```

CarmenPopoviciu: also would be great if we could move that headers mapping out of the log block. It makes it very hard to read

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -351,6 +355,156 @@ async function handleScheduled(
...
+					"http://localhost/core/store-temp-file?extension=eml",
+					{
+						method: "POST",
+						body: finalReply,
+					}
+				);
+				const file = await resp.text();
+
+				console.log(
+					`${blue(".reply() called from Email Handler with the following message:")}\n  ${file}`
```
CarmenPopoviciu, line 490:
```suggestion
					`${blue("Email handler replied to sender with the following message:")}\n  ${file}`
```

packages/miniflare/src/workers/core/entry.worker.ts
LuisDuarte1: i tried to have error messages as closely matched as production, I admit that they aren't the best ones, but maybe it's more worth to have an environment close to prod?

packages/miniflare/src/workers/core/entry.worker.ts
LuisDuarte1: i tried to have error messages as closely matched as production, I admit that they aren't the best ones, but maybe it's more worth to have an environment close to prod?

packages/miniflare/src/workers/core/entry.worker.ts
LuisDuarte1: This a great question, this might be a breaking change - @penalosa would love your thoughts here since you started this part :)

penalosa: We actually broke it with `DevEnv` accidentally (it was being intercepted by the proxy worker)—this fixes `/cdn-cgi/mf/scheduled` to work properly again

packages/miniflare/src/workers/core/entry.worker.ts
penalosa: I'm going to add a warning about this limitation for users. It's not part of the dev week deliverable, and might not actually be solvable without runtime work.

packages/miniflare/src/workers/core/entry.worker.ts
penalosa: Should we not fix production as well, if that typo is there?

LuisDuarte1 (approved):
:ship: it, :rocket:

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -1026,6 +1026,8 @@ export async function buildMiniflareOptions(
 		liveReload: config.liveReload,
 		upstream,
 		unsafeProxySharedSecret: proxyToUserWorkerAuthenticationSecret,
+		unsafeTriggerHandlers: true,
+		logRequests: false,
```
penalosa, line 1030: `logRequests: false` <- that looks weird! So basically, the way we run Miniflare instances with `wrangler dev` is that there are two—one holding the proxy worker, and one holding the user worker. The issue with that setup is that end users would see _two_ sets of request logs from Miniflare, and so previously we'd been hiding _all_ of the user-worker Miniflare instance logs. Instead, we now specifically hide the request logs, allowing other logs to be shown to the user (such as details about emails being triggered)

packages/miniflare/src/workers/email/validate.ts
```diff
@@ -0,0 +1,136 @@
+import { red } from "kleur/colors";
+import PostalMime, { Email } from "postal-mime";
+import { RAW_EMAIL } from "./constants";
+import { type MiniflareEmailMessage as EmailMessage } from "./email.worker";
+
+// Email Routing has some limits on what emails can be responded to, documented at https://developers.cloudflare.com/email-routing/email-workers/reply-email-workers/
```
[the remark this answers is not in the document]
penalosa: Unfortunately this isn't in the runtime in production, it's in the backing email service, so I don't think it would necessarily make sense for it to be in workerd. We could open source bits of the email service, but it's all rust based so that would be a bit of an undertaking.

packages/miniflare/src/workers/email/send_email.worker.ts
```diff
@@ -0,0 +1,91 @@
...
+		try {
+			emailHeaders = new Headers(
+				parsedEmail.headers.map((header) => [header.key, header.value])
+			);
+		} catch (e) {
+			const error = e as Error;
+			throw new Error(`could not parse email: ${error.message}`);
+		}
+
+		if (emailMessage.from !== parsedEmail.from.address) {
```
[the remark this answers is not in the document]
LuisDuarte1: according to SMTP spec, it depends - the part before the `@` is up to email providers to define (I'm not sure what we do currently tbh), the domain part should be case-insensitive per domain names spec. 

I think it's fine to leave it like this, it's something very uncommon to happen - the "gentlement agreement" here is that emails should always be lowercase anyway

packages/wrangler/src/dev/miniflare.ts
penalosa: bcf083300

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+			contentType: "text/plain",
+			data: `Congratulations, you just sent an email from a Worker.`,
+		});
+
+		const m = new EmailMessage(message.to, message.from, msg.asRaw());
+		await message.forward(
+			"samuel@macleod.space",
+			new Headers({ hello: "world" })
+		);
+		// message.setReject("Rejection reason");
```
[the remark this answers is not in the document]
penalosa: d7c924a71

fixtures/email-worker/src/index.ts
```diff
@@ -0,0 +1,53 @@
...
+		await message.reply(m);
+	}
+
+	async scheduled() {
+		console.log("I'm scheduled!");
```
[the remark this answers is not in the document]
penalosa: d7c924a71

fixtures/email-worker/wrangler.toml
```diff
@@ -0,0 +1,11 @@
+name = "email-to-rss"
+main = "src/index.ts"
+compatibility_date = "2025-02-14"
+
+compatibility_flags = [ "nodejs_compat" ]
```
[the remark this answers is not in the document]
penalosa: d7c924a71

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -795,13 +807,8 @@ export function getGlobalServices({
 			name: SERVICE_ENTRY,
 			worker: {
 				modules: [{ name: "entry.worker.js", esModule: SCRIPT_ENTRY() }],
-				compatibilityDate: "2023-04-04",
-				compatibilityFlags: [
-					"nodejs_compat",
-					"service_binding_extra_handlers",
-					"brotli_content_encoding",
-					"rpc",
```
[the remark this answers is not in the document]
penalosa: Yep: https://developers.cloudflare.com/workers/configuration/compatibility-flags/#brotli-content-encoding-support & https://developers.cloudflare.com/workers/configuration/compatibility-flags/#durable-object-stubs-and-service-bindings-support-rpc, both enabled last year by default

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -210,6 +210,10 @@ export const CoreSharedOptionsSchema = z.object({
 	unsafeModuleFallbackService: ServiceFetchSchema.optional(),
 	// Keep blobs when deleting/overwriting keys, required for stacked storage
 	unsafeStickyBlobs: z.boolean().optional(),
+	// Enable directly triggering user Worker handlers with paths like `/cdn-cgi/handler/scheduled`
+	unsafeTriggerHandlers: z.boolean().optional(),
```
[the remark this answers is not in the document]
penalosa: Yeah—this is needed in order to _disable_ this behaviour in the ProxyWorker, so that triggering handlers gets through to the user-worker Miniflare instance

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,105 @@
...
+		name,
+		json: JSON.stringify(value),
+	}));
+}
+
+function createPlugin<O extends z.ZodType, S extends z.ZodType | undefined>(
+	pluginDefinition: Plugin<O, S>
+): Plugin<O, S> {
+	return pluginDefinition;
+}
```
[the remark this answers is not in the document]
penalosa: This PR started as a POC, and I was playing around with better ways to author plugins. I can remove this

fixtures/email-worker/wrangler.toml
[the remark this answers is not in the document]
penalosa: Oops—this is actually required for `mimetext`

packages/miniflare/src/plugins/email/index.ts
```diff
@@ -0,0 +1,105 @@
...
+						name: "cloudflare-internal:email",
+						esModule: EMAIL_MESSAGE(),
+						internal: true,
+					},
+				],
+			});
+		}
+
+		services.push(
+			...(args.options.email?.send_email ?? []).map(({ name, ...config }) => ({
```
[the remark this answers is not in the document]
penalosa: e0241926d

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -408,7 +407,9 @@ export default <ExportedHandler<Env>>{
 			}
 			response = maybeInjectLiveReload(response, env, ctx);
 			response = ensureAcceptableEncoding(clientAcceptEncoding, response);
+			// if (env[CoreBindings.TRIGGER_HANDLERS]) {
```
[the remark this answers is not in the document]
penalosa: 6e7f1dbf2

packages/miniflare/test/index.spec.ts
```diff
@@ -1406,19 +1406,305 @@ test("Miniflare: manually triggered scheduled events", async (t) => {
...
+		}
+	);
+	t.is(await res.text(), "Worker successfully processed email");
+	t.is(res.status, 200);
+
+	res = await mf.dispatchFetch("http://localhost");
+	t.is(await res.text(), "true");
+});
+
+test("Miniflare: manually triggered email handler - valid reply", async (t) => {
```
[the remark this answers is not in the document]
penalosa: I've actually got no idea—I'll remove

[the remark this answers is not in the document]
penalosa: Removed: 2c35e8011—this stuff is well tested in other tests

petebacondarwin:
Blocking this on CR with approval

petebacondarwin:
CR is approved

PR #8376, opened 2025-03-06
feat(miniflare): Implement Assets RPC Proxy Worker behaviour

CarmenPopoviciu:
will need some follow-up fixture changes once https://github.com/cloudflare/workerd/pull/3742 lands, but we will do that in a separate PR

Fixes DEVX-1643

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 
  - [x] Not necessary because: experimental feature

### DIFF
--- a/.changeset/real-jobs-own.md
+++ b/.changeset/real-jobs-own.md
@@ -1,0 +1,6 @@
+---
+"miniflare": patch
+"wrangler": patch
+---
+
+feat: Make local dev RPC behaviour on par with production for Workers with assets
--- a/fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
+++ b/fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
@@ -58,7 +58,8 @@ describe.each(devCmds)(
 		describe("Workers running in separate wrangler dev sessions", () => {
 			describe("Service binding to default export", () => {
 				// this currently incorrectly returns the User Worker response
-				// instead of the Asset Worker response
+				// instead of the Asset Worker response, unless `--x-assets-rpc`
+				// is provided
 				failsIf(!args.length)(
 					"should return Asset Worker response for routes that serve static content",
 					async ({ expect }) => {
@@ -130,7 +131,8 @@ describe.each(devCmds)(
 
 			describe("Service binding to default entrypoint", () => {
 				// this currently incorrectly returns the User Worker response
-				// instead of the Asset Worker response
+				// instead of the Asset Worker response, unless `--x-assets-rpc`
+				// is provided
 				failsIf(!args.length)(
 					"should return Asset Worker response for fetch requests for routes that serve static content",
 					async ({ expect }) => {
--- a/fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
+++ b/fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
@@ -6,6 +6,10 @@ import { runWranglerDev } from "../../shared/src/run-wrangler-long-lived";
 
 const devCmds = [{ args: [] }, { args: ["--x-assets-rpc"] }];
 
+const failsIf = (condition: boolean) => {
+	return condition ? it.fails : it;
+};
+
 describe.each(devCmds)(
 	"[wrangler dev $args][Workers + Assets] Service bindings to Worker with assets",
 	({ args }) => {
@@ -55,7 +59,7 @@ describe.each(devCmds)(
 			describe("Service binding to default export", () => {
 				// this currently incorrectly returns the User Worker response
 				// instead of the Asset Worker response
-				it.fails(
+				failsIf(!args.length)(
 					"should return Asset Worker response for routes that serve static content",
 					async ({ expect }) => {
 						let response = await fetch(`http://${ipWorkerA}:${portWorkerA}`);
@@ -127,8 +131,8 @@ describe.each(devCmds)(
 			describe("Service binding to default entrypoint", () => {
 				// this currently incorrectly returns the User Worker response
 				// instead of the Asset Worker response
-				it.fails(
-					"should return Asset Worker response for fetch requestsfor routes that serve static content",
+				failsIf(!args.length)(
+					"should return Asset Worker response for fetch requests for routes that serve static content",
 					async ({ expect }) => {
 						let response = await fetch(`http://${ipWorkerA}:${portWorkerA}`);
 						let text = await response.text();
@@ -239,6 +243,20 @@ describe.each(devCmds)(
 					);
 					expect(text).toContain(
 						'env.NAMED_ENTRYPOINT.busyBee("🐝") response: Hello busy 🐝s from worker-d busyBee(bee)'
+					);
+				});
+
+				it("should support promise pipelining", async ({ expect }) => {
+					// fetch URL is irrelevant here. workerA will internally call
+					// the appropriate fns on the service binding instead
+					let response = await fetch(`http://${ipWorkerA}:${portWorkerA}`);
+					let text = await response.text();
+					expect(response.status).toBe(200);
+					expect(text).toContain(
+						`env.NAMED_ENTRYPOINT.foo("🐙").bar.buzz() response: You made it! 🐙`
+					);
+					expect(text).toContain(
+						`env.NAMED_ENTRYPOINT.newBeeCounter().value response: 2`
 					);
 				});
 			});
--- a/fixtures/workers-with-assets-and-service-bindings/workerA/src/workerD.util.ts
+++ b/fixtures/workers-with-assets-and-service-bindings/workerA/src/workerD.util.ts
@@ -14,7 +14,7 @@ export async function getWorkerDResponses(request: Request, env) {
 	const buzzResult = await foo.bar.buzz();
 
 	// test RpcTarget + promise pipelining
-	const beeCounter = await env.NAMED_ENTRYPOINT.newBeeCounter();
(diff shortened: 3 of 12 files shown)


.changeset/real-jobs-own.md
```diff
@@ -0,0 +1,6 @@
+---
+"miniflare": patch
+"wrangler": patch
+---
+
+feat: Implement Proxy Worker behaviour we need for Workers+Assets
```
petebacondarwin, line 6: This is describing the implementation rather than the effect it has on the user's experience.

CarmenPopoviciu: true! will reword

packages/wrangler/e2e/assets-multiworker.test.ts
```diff
@@ -349,7 +349,9 @@ describe.each(
 				});
 			});
 
-			failsIf(style === "dev registry")(".fetch() existing asset", async () => {
+			failsIf(
+				style === "dev registry" && !wranglerDev.includes("--x-assets-rpc")
```
CarmenPopoviciu, line 353: this is good! this tells us that the Proxy Worker solved the issue for these use cases

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
```diff
@@ -55,7 +59,7 @@ describe.each(devCmds)(
 			describe("Service binding to default export", () => {
 				// this currently incorrectly returns the User Worker response
 				// instead of the Asset Worker response
-				it.fails(
+				failsIf(!args.length)(
```
CarmenPopoviciu, line 62: this is a good thing! this means that our Proxy Worker solved this use case

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
```diff
@@ -241,6 +245,20 @@ describe.each(devCmds)(
 						'env.NAMED_ENTRYPOINT.busyBee("🐝") response: Hello busy 🐝s from worker-d busyBee(bee)'
 					);
 				});
+
+				it("should support promise pipelining", async ({ expect }) => {
```
penalosa, line 249: Is this not going to fail?

fixtures/workers-with-assets-and-service-bindings/workerA/src/workerD.util.ts
```diff
@@ -9,13 +9,26 @@ export async function getWorkerDResponses(request: Request, env) {
 	// test named functions with parameters
 	const busyBeeResult = await env.NAMED_ENTRYPOINT.busyBee("🐝");
 
+	// test nested functions + promise pipelining
+	const foo = env.NAMED_ENTRYPOINT.foo("🐙");
+	const buzzResult = await foo.bar.buzz();
+
+	// test RpcTarget + promise pipelining
+	const beeCounter = await env.NAMED_ENTRYPOINT.newBeeCounter();
```
penalosa, line 17:
```suggestion
	const beeCounter = env.NAMED_ENTRYPOINT.newBeeCounter();
```
?

fixtures/workers-with-assets-and-service-bindings/workerB-with-default-export/src/index.ts
```diff
@@ -22,6 +22,17 @@ export default {
 		return `Hello busy ${bee}s from worker-b busyBee(bee)`;
 	},
 
+	/*
+	 * Nested functions
+	 */
```
penalosa, line 27: Could we add a test for properties, as well?

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -3,12 +3,53 @@ import type RouterWorker from "@cloudflare/workers-shared/asset-worker/src/index
...
+				 */
+				if (Reflect.has(target, prop)) {
+					return Reflect.get(target, prop);
+				}
+
+				/*
+				 * Otherwise, forward to the USER_WORKER and return its response
+				 */
+				return function (...args: Array<unknown>) {
+					// @ts-ignore to figure out
```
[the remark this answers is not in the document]
penalosa: It should be a `Service` or `Fetcher`, I think? I imagine this will need a ts-expect-error though

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -3,12 +3,53 @@ import type RouterWorker from "@cloudflare/workers-shared/asset-worker/src/index
...
+				 * return whatever `RPCProxyWorker[prop]` returns
+				 */
+				if (Reflect.has(target, prop)) {
+					return Reflect.get(target, prop);
+				}
+
+				/*
+				 * Otherwise, forward to the USER_WORKER and return its response
+				 */
+				return function (...args: Array<unknown>) {
```
penalosa, line 48: This doesn't work with properties, but I believe that'll be addressed after the workerd bump?

penalosa (approved):
LGTM

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
CarmenPopoviciu: not for named entrypoints, since the Proxy Worker is added to the pipeline only for default exports

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
CarmenPopoviciu: indeed. But I need to first see what I get back from the Runtime, once we have access to that workerd fix

fixtures/workers-with-assets-and-service-bindings/workerA/src/workerD.util.ts
CarmenPopoviciu: thx for catching that! yes indeed

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
dario-piotrowicz: should the comment above be updated as well?

adding something like `unless --x-assets-rpc is provided` at the end?

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
```diff
@@ -127,8 +131,8 @@ describe.each(devCmds)(
 			describe("Service binding to default entrypoint", () => {
 				// this currently incorrectly returns the User Worker response
 				// instead of the Asset Worker response
-				it.fails(
-					"should return Asset Worker response for fetch requestsfor routes that serve static content",
+				failsIf(!args.length)(
```
dario-piotrowicz, line 134: same as above 🙂

dario-piotrowicz: Also (VERY NIT! 🙈) I think something like this would be much more readable/easy to follow without having to check other parts of this file: `failsIf(!args.includes("--x-assets-rpc"))`

packages/miniflare/src/index.ts
```diff
@@ -1343,19 +1343,41 @@ export class Miniflare {
...
+					this.#sharedOpts.core.unsafeEnableAssetsRpc &&
+					workerOpts.assets.assets &&
+					entrypoint === "default"
+				) {
+					sockets.push({
+						name,
+						address,
+						service: {
+							name: `${RPC_PROXY_SERVICE_NAME}:${workerOpts.core.name}`,
+						},
```
dario-piotrowicz, line 1359: is the service the only thing that changes in the if vs else brackets? if so, wouldn't it be clearer to keep the call `sockets.push` the exact same but only conditionally swap services? something like this:
Screenshot 2025-03-19 at 15 48 58
or
Screenshot 2025-03-19 at 15 51 24

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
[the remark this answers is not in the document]
CarmenPopoviciu: ✅

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
CarmenPopoviciu: totally fair. We are removing the flag though this week, and once we do that, all these `failIf`s won't be necessary anymore

packages/miniflare/src/index.ts
CarmenPopoviciu: yeah sure, I can abs do that!

CarmenPopoviciu:
@dario-piotrowicz thank you so much for reviewing. I addressed all feedback comments [here](https://github.com/cloudflare/workers-sdk/pull/8376/commits/80891a833673203aef912f42f88a4f2a4270e3cd)

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
dario-piotrowicz: ah nice ok! 😄

dario-piotrowicz (approved):
Thanks a lot for the updates @CarmenPopoviciu 🫶 

I'm approving although I haven't been fully followed this RPC proxy worker implementation, but on a surface level the code looks good to me, and I trust Samuel's judgment 😄

PR #8382, opened 2025-03-06
feat(wrangler): Add Secrets Store support

jvaughan-cloudflare:
This PR introduces the Secrets Store API to the Wrangler CLI. It adds commands namespaces for `wrangler secrets-store` with sub-namespaces of `wrangler secrets-store store` and `wrangler secrets-store secret` for acting upon store and secret resources respectively. The intent here is to be able to fully manage the stores+secrets for an account's Secret Store, from wrangler. 

Changes:
- Added `wrangler secrets-store store list|create|delete` commands
- Added `wrangler secrets-store secret list|get|create|update|delete|duplicate` commands
- All commands default to local dev, need to pass `--remote` for acting on remote

Todo:
- Will implement local dev version of commands (specifically create/delete secret), in a subsequent change
- Fix e2e test runner, what do I need to do for API tokens to get the test runner to have the right permissions? Add new api token role to bach, and then update the api token in the github ci to include that role/permissions? 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Waiting on public documentation until the Secrets Store is released as a product.

### DIFF
--- a/packages/wrangler/e2e/secrets-store.test.ts
+++ b/packages/wrangler/e2e/secrets-store.test.ts
@@ -1,0 +1,153 @@
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
+import { generateResourceName } from "./helpers/generate-resource-name";
+import { normalizeOutput } from "./helpers/normalize";
+
+describe("secrets-store", async () => {
+	let cachedStoreId = "";
+	let cachedSecretId = "";
+
+	const storeName = generateResourceName("secrets-store-store");
+	const secretName = generateResourceName("secrets-store-secret");
+	const helper = new WranglerE2ETestHelper();
+
+	const originalColumns = process.stdout.columns;
+	beforeAll(() => {
+		process.stdout.columns = 180;
+	});
+
+	afterAll(() => {
+		process.stdout.columns = originalColumns;
+	});
+
+	const normalize = (str: string) => {
+		return normalizeOutput(str, {
+			[storeName]: "tmp-e2e-secrets-store-store",
+			[secretName]: "tmp-e2e-secrets-store-secret",
+			[process.env.CLOUDFLARE_ACCOUNT_ID as string]: "CLOUDFLARE_ACCOUNT_ID",
+		});
+	};
+
+	it("creates a store", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store store create ${storeName} --remote`
+		);
+
+		const regex = /ID:\s*(\w{32})/;
+		const match = output.stdout.match(regex);
+
+		if (match && match[1]) {
+			cachedStoreId = match[1];
+		} else {
+			throw new Error("No uuid for store found.");
+		}
+
+		expect(normalize(output.stdout)).toMatchInlineSnapshot(`
+			"🔐 Creating store... (Name: tmp-e2e-secrets-store-store-00000000-0000-0000-0000-000000000000)
+✅ Created store! (Name: tmp-e2e-secrets-store-store-00000000-0000-0000-0000-000000000000, ID: 00000000000000000000000000000000)"
+		`);
+	});
+
+	it("lists stores", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store store list --per-page 100 --remote`
+		);
+
+		expect(normalize(output.stdout)).toContain(
+			"tmp-e2e-secrets-store-store-00000000-0000-0000-0000-000000000000"
+		);
+	});
+
+	it("creates a secret", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store secret create ${cachedStoreId} --name ${secretName} --value shh --scopes workers --comment test --remote`
+		);
+
+		const regex = /ID:\s*(\w{32})/;
+		const match = output.stdout.match(regex);
+
+		if (match && match[1]) {
+			cachedSecretId = match[1];
+		}
+
+		expect(normalize(output.stdout)).toContain(
+			"🔐 Creating secret... (Name: tmp-e2e-secrets-store-secret-00000000-0000-0000-0000-000000000000, Value: REDACTED, Scopes: workers, Comment: test"
+		);
+		expect(normalize(output.stdout)).toContain(
+			"✅ Created secret! (ID: 00000000000000000000000000000000)"
+		);
+	});
+
+	it("gets a secret", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store secret get ${cachedStoreId} --secret-id ${cachedSecretId} --remote`
+		);
+
+		expect(normalize(output.stdout)).toContain(
+			"🔐 Getting secret... (ID: 00000000000000000000000000000000)"
+		);
+		expect(normalize(output.stdout)).toContain(
+			"tmp-e2e-secrets-store-secret-00000000-0000-0000-0000-000000000000"
+		);
+	});
+
+	it("updates a secret", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store secret update ${cachedStoreId} --secret-id ${cachedSecretId} --value shh --remote`
+		);
+
+		expect(normalize(output.stdout)).toContain(
+			"🔐 Updating secret... (ID: 00000000000000000000000000000000)"
+		);
+		expect(normalize(output.stdout)).toContain(
+			"✅ Updated secret! (ID: 00000000000000000000000000000000)"
+		);
+		expect(normalize(output.stdout)).toContain(
+			"tmp-e2e-secrets-store-secret-00000000-0000-0000-0000-00000000000"
+		);
+	});
+
+	it("deletes a secret", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store secret delete ${cachedStoreId} --secret-id ${cachedSecretId} --remote`
+		);
+
+		expect(normalize(output.stdout)).toMatchInlineSnapshot(`
+			"🔐 Deleting secret... (ID: 00000000000000000000000000000000)
+✅ Deleted secret! (ID: 00000000000000000000000000000000)"
+		`);
+	});
+
+	it("validates a secret is deleted", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store secret get ${cachedStoreId} --secret-id ${cachedSecretId} --remote`
+		);
+
+		expect(normalize(output.stdout)).toContain(
+			"🔐 Getting secret... (ID: 00000000000000000000000000000000)"
+		);
+		expect(normalize(output.stdout)).toContain("secret_not_found [code: 1001]");
+	});
+
+	it("deletes a store", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store store delete ${cachedStoreId} --remote`
+		);
+
+		expect(normalize(output.stdout)).toMatchInlineSnapshot(`
+			"🔐 Deleting store... (Name: 00000000000000000000000000000000)
+✅ Deleted store! (ID: 00000000000000000000000000000000)"
+		`);
+	});
+
+	it("validates a store is deleted", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store secret get ${cachedStoreId} --secret-id ${cachedSecretId} --remote`
+		);
+
+		expect(normalize(output.stdout)).toContain(
+			"🔐 Getting secret... (ID: 00000000000000000000000000000000)"
+		);
+		expect(normalize(output.stdout)).toContain("store_not_found [code: 1001]");
+	});
+});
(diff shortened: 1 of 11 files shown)


packages/wrangler/e2e/secrets-store.test.ts
```diff
@@ -0,0 +1,26 @@
...
+	const helper = new WranglerE2ETestHelper();
+
+	it("creates a store", async () => {
+		const normalize = (str: string) =>
+			normalizeOutput(str, {
+				[storeName]: "tmp-e2e-secrets-store-store",
+				[process.env.CLOUDFLARE_ACCOUNT_ID as string]: "CLOUDFLARE_ACCOUNT_ID",
+			});
+
+		const output = await helper.run(
```
jvaughan-cloudflare, line 18: Will add more e2e tests after figuring out what is needed for API token + account permissions to succeed

packages/wrangler/src/metrics/send-event.ts
```diff
@@ -82,7 +82,15 @@ export type EventNames =
...
-	| "provision resources";
+	| "provision resources"
+	| "create secrets-store store"
+	| "list secrets-store stores"
+	| "list secrets-store secrets"
+	| "get secrets-store secret"
+	| "delete secrets-store secret"
+	| "create secrets-store secret"
+	| "update secrets-store secret"
+	| "duplicate secrets-store secret";
```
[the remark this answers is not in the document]
jvaughan-cloudflare: Got it, didn't realize that! I'll remove the explicit metric calls

mitalirawat (approved):

packages/wrangler/src/secrets-store/commands.ts
```diff
@@ -0,0 +1,582 @@
...
+	},
+});
+
+export const secretsStoreStoreDeleteCommand = createCommand({
+	metadata: {
+		description: "Delete a store within an account",
+		status: "alpha",
+		owner: "Product: SSL",
+	},
+	positionalArgs: ["store-id"],
```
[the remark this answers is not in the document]
jvaughan-cloudflare: It could eventually, but currently the secrets store api endpoint is like `/accounts/:id/stores/:id` where the id is passed in the URL, so for now we have to look it reference it by ID and not name.

packages/wrangler/src/secrets-store/commands.ts
```diff
@@ -0,0 +1,582 @@
...
+			requiresArg: true,
+		},
+		remote: {
+			type: "boolean",
+			description: "Execute command against remote Secrets Store",
+			default: false,
+		},
+	},
+	async handler(args, _ctx) {
+		const config = readConfig(args);
```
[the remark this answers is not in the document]
jvaughan-cloudflare: Fixed!

packages/wrangler/src/secrets-store/commands.ts
```diff
@@ -0,0 +1,582 @@
...
+			demandOption: true,
+			requiresArg: true,
+		},
+		remote: {
+			type: "boolean",
+			description: "Execute command against remote Secrets Store",
+			default: false,
+		},
+	},
+	async handler(args, _ctx) {
```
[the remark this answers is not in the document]
jvaughan-cloudflare: Fixed!

packages/wrangler/src/secrets-store/commands.ts
```diff
@@ -0,0 +1,582 @@
...
+		"secret-id": {
+			describe: "ID of the secret to update",
+			type: "string",
+			requiresArg: true,
+			demandOption: true,
+		},
+		value: {
+			describe: "Updated value of the secret",
+			type: "string",
+		},
```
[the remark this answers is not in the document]
jvaughan-cloudflare: Hmm. It looks like the `put` command for the current secrets correlates to a `PUT` endpoint somewhere. We don't have a comparable endpoint in the secrets store that does an update or insert, we have two separate endpoints, which is why I created two different commands. Is it important to maintain consistency with the current `secret` commands? I don't know much about those tbh

packages/wrangler/src/secrets-store/commands.ts
```diff
@@ -0,0 +1,582 @@
...
+				Status: secret.status === "active" ? "active " : "pending",
+				Created: new Date(secret.created).toLocaleString(),
+				Modified: new Date(secret.modified).toLocaleString(),
+			},
+		];
+		logger.table(prettierSecret);
+	},
+});
+
+export const secretsStoreSecretCreateCommand = createCommand({
```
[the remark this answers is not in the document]
jvaughan-cloudflare: Hmm. It looks like the put command for the current secrets correlates to a PUT endpoint somewhere. We don't have a comparable endpoint in the secrets store that does an update or insert, we have two separate endpoints, which is why I created two different commands. Is it important to maintain consistency with the current secret commands? I don't know much about those tbh

packages/wrangler/src/secrets-store/commands.ts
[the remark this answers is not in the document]
jvaughan-cloudflare: I'll look into this, wasn't aware of the existing UX

packages/wrangler/src/secrets-store/commands.ts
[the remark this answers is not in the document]
penalosa: Totally fair if you want to have that explicit creation DX—just wanted to check that was an intentional choice to diverge from how existing worker secrets work

packages/wrangler/src/secrets-store/commands.ts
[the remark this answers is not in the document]
jvaughan-cloudflare: Are there any benefits to reading it in through stdin or a prompt? Or is this mostly to try and keep UX similar between this and the existing secrets?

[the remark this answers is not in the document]
penalosa: The primary reason is that these values are secrets, and so you probably don't want them in your terminal history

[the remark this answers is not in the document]
jvaughan-cloudflare: Got it, will add for sure then, since this is a security issue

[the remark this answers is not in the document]
jvaughan-cloudflare: Updated!

packages/wrangler/e2e/secrets-store.test.ts
```diff
@@ -0,0 +1,193 @@
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { readConfig } from "../src/config";
+import {
+	deleteStore,
+	listSecrets,
+	listStores,
+} from "../src/secrets-store/client";
```
penalosa, line 7: E2E tests shouldn't import directly from Wrangler's source. They're intended to test things as a user would (i.e. just by running the Wrangler CLI)

packages/wrangler/e2e/secrets-store.test.ts
```diff
@@ -0,0 +1,193 @@
...
+	it("validates a store is deleted", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store store list --remote`
+		);
+
+		expect(normalize(output.stdout)).toContain("🔐 Listing stores...");
+		expect(normalize(output.stderr)).toContain(
+			"List request returned no stores."
+		);
+	});
```
penalosa, line 192: This won't safely run in CI, since other E2E jobs might also be creating stores. What about trying to get a secret from the store? Would that throw an error that could be matched?

packages/wrangler/e2e/secrets-store.test.ts
```diff
@@ -0,0 +1,193 @@
...
+	};
+
+	await setup();
+
+	it("creates a store", async () => {
+		const output = await helper.run(
+			`wrangler secrets-store store create ${storeName} --remote`
+		);
+
+		expect(normalize(output.stdout)).toMatchInlineSnapshot(`
```
penalosa, line 78: The store ID should be extracted from the output here

packages/wrangler/e2e/secrets-store.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+		const output = await helper.run(
+			`wrangler secrets-store secret create ${id} --name ${secretName} --value 'shh!' --scopes 'workers' --comment 'test' --remote`
+		);
+
+		expect(normalize(output.stdout)).toContain(
+			"🔐 Creating secret... (Name: tmp-e2e-secrets-store-secret-00000000-0000-0000-0000-000000000000, Value: REDACTED, Scopes: workers, Comment: test"
+		);
+		expect(normalize(output.stdout)).toContain(
+			"✅ Created secret! (ID: 00000000000000000000000000000000)"
```
penalosa, line 105: The secret ID should be extracted from the output here

packages/wrangler/src/metrics/send-event.ts
```diff
@@ -82,7 +82,16 @@ export type EventNames =
...
+	| "provision resources"
+	| "create secrets-store store"
+	| "list secrets-store stores"
+	| "delete secrets-store store"
+	| "list secrets-store secrets"
+	| "get secrets-store secret"
+	| "delete secrets-store secret"
+	| "create secrets-store secret"
+	| "update secrets-store secret"
+	| "duplicate secrets-store secret";
```
penalosa, line 94: These events shouldn't be manually added—this type of telemetry is sent automatically

packages/wrangler/src/secrets-store/client.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export type CreateStore = {
+	name: string;
+};
+
+export async function createStore(
+	config: Config,
+	body: CreateStore
+): Promise<Store> {
+	const accountId = await requireAuth(config);
```
penalosa, line 23: nit, but these functions could take `accountId`, which would reduce the `requireAuth()` duplication in this file

packages/wrangler/src/secrets-store/client.ts
```diff
@@ -0,0 +1,189 @@
...
+			method: "GET",
+		}
+	);
+}
+
+export type CreateSecret = {
+	name: string;
+	value: string;
+	scopes: string[];
+	comment: string | null | undefined;
```
penalosa, line 103: Why are these `null | undefined` instead of optional fields (`comment?: string`)?

packages/wrangler/src/secrets-store/client.ts
```diff
@@ -0,0 +1,189 @@
...
+	const accountId = await requireAuth(config);
+	if (!body.value) {
+		delete body.value;
+	}
+	if (!body.scopes) {
+		delete body.scopes;
+	}
+	if (!body.comment) {
+		delete body.comment;
+	}
```
penalosa, line 145: This complexity seems unnecessary—these should just be optional fields

packages/wrangler/src/metrics/send-event.ts
jvaughan-cloudflare: Ah, so I can remove them from here entirely?

packages/wrangler/e2e/secrets-store.test.ts
jvaughan-cloudflare: Ah, you mean instead of the lookup I do? I can do that, just need to regex it out

packages/wrangler/src/secrets-store/client.ts
jvaughan-cloudflare: True, I'll clean that up

packages/wrangler/src/secrets-store/client.ts
jvaughan-cloudflare: Mostly because I haven't written typescript in a long time :laughing:  I'll change this to optional

packages/wrangler/e2e/secrets-store.test.ts
jvaughan-cloudflare: Yep, that would work!

jvaughan-cloudflare: Although your point brings up a larger issue, currently we cap each account to 1 store. So if multiple E2E jobs try to create a store on this account at the same time, one will fail... I need to think through this again.

packages/wrangler/src/secrets-store/client.ts
jvaughan-cloudflare: Although I guess that will just push the duplication to the commands files to make a `const accountId = config.account_id || (await getAccountId());` call

packages/wrangler/src/metrics/send-event.ts
penalosa: Yes 👍

packages/wrangler/e2e/secrets-store.test.ts
jvaughan-cloudflare: Okay, solved this one, can discuss off-thread

penalosa (approved):

PR #8383, opened 2025-03-06
fix(wrangler): Make kv bulk put --local respect base64:true

matthewdavidrodgers:
The bulk put api has an optional "base64" boolean property for each key. Before storing the key, the value should be decoded from base64.

For real (remote) kv, this is handled by the rest api. For local kv, it seems the base64 field was ignored, meaning encoded base64 content was stored locally rather than the raw values.

To fix, we need to decode each value before putting to the local miniflare namespace when base64 is true.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required (https://github.com/cloudflare/workers-sdk/pull/8400)
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because:  Fixes bug in behavior described in docs

### DIFF
--- a/.changeset/chatty-glasses-drum.md
+++ b/.changeset/chatty-glasses-drum.md
@@ -1,0 +1,15 @@
+---
+"wrangler": patch
+---
+
+Make kv bulk put --local respect base64:true
+
+The bulk put api has an optional "base64" boolean property for each key.
+Before storing the key, the value should be decoded from base64.
+
+For real (remote) kv, this is handled by the rest api. For local kv, it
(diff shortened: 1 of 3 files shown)


petebacondarwin (approved):
Nice fix! Thanks @matthewdavidrodgers

petebacondarwin:
E2E tests did pass on the internal draft PR (#8400): https://github.com/cloudflare/workers-sdk/actions/runs/13726216617

PR #8387, opened 2025-03-06
Support Text and Data module types

jamesopstad:
Fixes #000.

This builds on the support for CompiledWasm to add support for Text and Data module types.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 


packages/vite-plugin-cloudflare/src/shared.ts
```diff
@@ -1,5 +1,5 @@
-import { MODULE_TYPES } from "./constants";
+import { ADDITIONAL_MODULE_TYPES } from "./constants";
 
 export const UNKNOWN_HOST = "http://localhost";
 export const INIT_PATH = "/__vite_plugin_cloudflare_init__";
-export const MODULE_PATTERN = `__CLOUDFLARE_MODULE__(${MODULE_TYPES.join("|")})__(.*?)__`;
+export const ADDITIONAL_MODULE_PATTERN = `__CLOUDFLARE_MODULE__(${ADDITIONAL_MODULE_TYPES.join("|")})__(.*?)__`;
```
petebacondarwin, line 5: This pattern is only ever used to create Regexes.
I am not 100% sure whether V8 is able to cache such things but it feels wasteful to have to create a new regex for every usage, especially inside hooks like `renderChunk`.
How about we create the Regex once here and reuse that instead of the string?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -330,29 +333,31 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
 			async resolveId(source, importer) {
-				if (!source.endsWith(".wasm")) {
+				const additionalModuleType = matchAdditionalModule(source);
+
+				if (!additionalModuleType) {
 					return;
 				}
 
-				const resolved = await this.resolve(source, importer);
+				const resolved = await this.resolve(cleanUrl(source), importer);
```
petebacondarwin, line 342: I found when doing the Node.js compat stuff that manually cleaning a query off the end of an Id was a code smell that I was not really doing the right thing with how Vite expected things to work.

Can you confirm why this is needed?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -474,29 +474,36 @@ export function getDevMiniflareOptions(
...
-				// Cap'n Proto expects byte arrays for `:Data` typed fields from JSON
-				wasm: Array.from(source),
-			});
+			return MiniflareResponse.json(
+				{
+					CompiledWasm: { wasm: Array.from(contents) },
+					Data: { data: Array.from(contents) },
+					Text: { text: contents.toString() },
+				}[moduleType]
+			);
```
petebacondarwin, line 506: It seems a bit unnecessary to generate each of these object just to throw away all but one of them. Can you use a switch instead?

```suggestion
			switch(moduleType) {
				case "CompiledWasm":
					return new MiniflareResponse.json({ wasm: Array.from(contents) });
				case "Data":
					return new MiniflareResponse.json({ data: Array.from(contents) });
				case "Text":
					return new MiniflareResponse.json({ text: contents.toString() });
				default:
					return new MiniflareResponse.error();
			)
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -330,29 +333,31 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
 				assert(
```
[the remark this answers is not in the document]
jamesopstad: Yes, you're right that it could be a user error. I don't think we've always honoured that distinction about not using `assert` for user errors elsewhere in the codebase. I want to go through all the errors soon anyway so can check and change them. I'll change this to throw.

packages/vite-plugin-cloudflare/src/shared.ts
jamesopstad: I did originally have this similar to how you suggest but you then need to create a separate regex here for the global and non-global variants and remember to set `lastIndex = 0` when using the global one. Happy to change it back though.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Cleaning the query params isn't always a code smell. This `cleanUrl` function is actually copied from Vite and used a lot in their internal plugins. In this case, it is because the default rules include both `.wasm` and `.wasm?module`. Therefore, we want the match to include the query param but remove it before resolving the ID. I can add a comment to clarify this.

On a related note, I'm not 100% if we need to support `.wasm?module` going forward but included it so we don't deviate from Wrangler's behaviour.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: I would be amazed if this ever resulted in an observable difference but sure, I'll make the change. More significant, I think, is that I'm using the synchronous file system functions so I'll look at making the module fallback and `renderChunk` async.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: [cf4c860](https://github.com/cloudflare/workers-sdk/pull/8387/commits/cf4c860c2b85b35f7fb469a0eaedef2c7c1a1645)

packages/vite-plugin-cloudflare/src/index.ts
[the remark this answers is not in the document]
jamesopstad: [451f0f0](https://github.com/cloudflare/workers-sdk/pull/8387/commits/451f0f097d8965656456e63f919af8943667c22d)

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: [83ec0bb](https://github.com/cloudflare/workers-sdk/pull/8387/commits/83ec0bb8f5240228472fe14b0fa0bcbc6ff4e892)
[64af99c](https://github.com/cloudflare/workers-sdk/pull/8387/commits/64af99cb945d2b4513770277917b07a1b291703a)

packages/vite-plugin-cloudflare/src/shared.ts
jamesopstad: Just thinking about this. Now that I've made `renderChunk` async I think that reusing the same regex instance could lead to unexpected behaviour because of shared `lastIndex` state. Using a new regex instance inside `renderChunk` is a lot easier to reason about. I could still create a single regex instance to use for the non global usage elsewhere if you prefer.

petebacondarwin: Ah yes, of course. Definitely don't want to share the regex state for this. OK.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: Ignore my previous comment. You're right that it's wasteful because it's transforming the file contents. Didn't appreciate that difference before.

packages/vite-plugin-cloudflare/src/shared.ts
jamesopstad: Made some changes to reuse the regex for the non-global cases.

[18a609d](https://github.com/cloudflare/workers-sdk/pull/8387/commits/18a609d88e0a682659b3976221250766ca14eeaf)

jamesopstad:
Switched to draft because I've realised we need to add some additional handling to invalidate the modules when they change. Currently they are cached indefinitely so changes don't take effect until the dev server is restarted.

petebacondarwin (approved):

PR #8392, opened 2025-03-07
fix: retry zone and route lookup API calls

jahands:
Fixes Follow up to https://github.com/gzuidhof/zarr.js/pull/145

_Describe your change..._

In rare cases, looking up Zone or Route API calls may fail due to transient errors. This change improves the reliability of `wrangler deploy` when these errors occur.

Also fixes a rare issue where concurrent API requests may fail without correctly throwing an error which may cause a deployment to incorrectly appear successful.

I discovered this issue when I noticed that some API calls failed, but the deployment still completed:
https://gist.github.com/jahands/3b71d748c554b971cd99c76783b94c82
```
This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason: TypeError: fetch failed
```

This lead me down a rabbit hole where I discovered that the only way to reliably ensure p-queue errors are caught is to await the promise returned by `queue.add()`:

* https://github.com/sindresorhus/p-queue/issues/26
* https://github.com/gzuidhof/zarr.js/pull/145

The weird thing is that _most_ of the time, the errors still bubble up (I tried inserting errors in the same path and could not repro the original issue I saw). I suspect there is some kind of race condition where in rare cases, these promises don't bubble up properly due to lack of awaiting the rejecting promises.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Should be no functionality changes
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: No functionality changes

### DIFF
--- a/packages/wrangler/src/assets.ts
+++ b/packages/wrangler/src/assets.ts
@@ -118,6 +118,7 @@ export const syncAssets = async (
 	});
 
 	const queue = new PQueue({ concurrency: BULK_UPLOAD_CONCURRENCY });
+	const queuePromises: Array<Promise<void>> = [];
 	let attempts = 0;
 	const start = Date.now();
 	let completionJwt = "";
@@ -201,17 +202,19 @@ export const syncAssets = async (
 			}
 		};
 		// add to queue and run it if we haven't reached concurrency limit
-		void queue.add(() =>
-			doUpload().then((res) => {
-				completionJwt = res.jwt || completionJwt;
-			})
+		queuePromises.push(
+			queue.add(() =>
+				doUpload().then((res) => {
+					completionJwt = res.jwt || completionJwt;
+				})
+			)
 		);
 	}
 	queue.on("error", (error) => {
 		logger.error(error.message);
 		throw error;
 	});
-	await queue.onIdle();
+	await Promise.all(queuePromises);
 
 	// if queue finishes without receiving JWT from asset upload service (AUS)
 	// AUS only returns this in the final bucket upload response
--- a/packages/wrangler/src/assets.ts
+++ b/packages/wrangler/src/assets.ts
@@ -214,6 +214,8 @@ export const syncAssets = async (
 		logger.error(error.message);
 		throw error;
 	});
+	// using Promise.all() here instead of queue.onIdle() to ensure
+	// we actually throw errors that occur within queued promises.
 	await Promise.all(queuePromises);
 
 	// if queue finishes without receiving JWT from asset upload service (AUS)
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -1072,51 +1072,56 @@ async function publishRoutesFallback(
...
-					});
-				}
-				throw e;
-			}
-		});
-		await queue.onIdle();
+			})
+		);
 	}
+	await Promise.all(queuePromises);
```
jahands, line 1124: onIdle() was being called in the incorrect spot, preventing concurrency from working, so went ahead and moved it outside the for loop while I was here.

packages/wrangler/src/assets.ts
```diff
@@ -201,17 +202,19 @@ export const syncAssets = async (
...
+				})
+			)
 		);
 	}
 	queue.on("error", (error) => {
 		logger.error(error.message);
 		throw error;
 	});
-	await queue.onIdle();
+	await Promise.all(queuePromises);
```
petebacondarwin, line 217: Could you put a comment in here explaining why we are not using `onIdle()` to prevent someone from "optimizing" it back and then having to go down the same rabbit hole you did.

petebacondarwin (approved):
LGTM except for adding comments.

jrf0110:
Perhaps p-queue is the wrong library for this. Seems like https://github.com/sindresorhus/p-map or https://github.com/sindresorhus/p-all would be more appropriate

jahands:
> Perhaps p-queue is the wrong library for this. Seems like https://github.com/sindresorhus/p-map or https://github.com/sindresorhus/p-all would be more appropriate

I agree, but the ability to add dependencies in this repo was broken when I tried so I opted to stick with p-queue

jahands:
> LGTM except for adding comments.

@petebacondarwin done!

packages/wrangler/src/assets.ts
jahands: Done!

jahands:
rebased

jahands:
Tested the Workers Assets change by deploying a project with 10k new files and spot-checked that files got uploaded correctly

jahands:
Going to go ahead and merge this because the E2E tests that are failing are the same tests failing on `main`:

Screenshot 2025-03-07 at 1 46 08 PM

PR #8412, opened 2025-03-10
Fixes bug with transaction detection and removal

### LINKED ISSUE #5085: 🐛 BUG: Invalid BEGIN TRANSACTION detection

D1, Wrangler core

3.29.0 [wrangler]

v20.11.1

Ubuntu 22.04.3 LTS

### Observed behaviour
I tried to update a row in D1 by passing a string that contains the words _BEGIN TRANSACTION_ utilising the Wrangler D1 command `execute` and got the error below.
> ✘ [ERROR] Wrangler could not process the provided SQL file, as it contains several transactions.

### Expected behaviour
The row will be updated.

### Steps to reproduce
- No worker involved. I am running a Wrangler command against D1
- `wrangler.toml` is not relevant, since I am running a Wrangler command against D1, but here it is:
```
name = "ai"
main = "src/index.js"
compatibility_date = "2023-12-06"
```
- I am not running a local dev server
- To reproduce run from terminal the two bash commands below:
```bash
TEXT="You can''t refer to a table that doesn''t yet exist.\n* If you get \"cannot start a transaction within a transaction\", make sure you have removed \`BEGIN TRANSACTION\` and \`COMMIT\` from your dumped SQL statements."
npx wrangler d1 execute <any-valid-db-name> --local --command "UPDATE Nodes SET text=\"${TEXT}\" WHERE id=537;" 
```

_No response_

_No response_

FlareLine:
Fixes #5085 - bug when checking for transactions in SQL statements.

Tests to come...
---

- Tests
  - [X] TODO (before merge)
  - [X] Tests included
  - [X] Tests not necessary because:
  - [ ] I don't know
  - [X] Required
  - [ ] Not required because:
- Public documentation
  - [x] TODO (before merge)
  - [x] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: _this change is a trivial change._

### DIFF
--- a/packages/wrangler/src/d1/trimmer.ts
+++ b/packages/wrangler/src/d1/trimmer.ts
@@ -30,5 +30,5 @@ export function trimSqlQuery(sql: string): string {
 // sqlite may start an sql dump file with pragmas,
 // so we can't just use sql.startsWith here.
 export function mayContainTransaction(sql: string): boolean {
-	return sql.includes("BEGIN TRANSACTION");
+	return sql.includes("BEGIN TRANSACTION;");
 }


lambrospetrou:
This fix is only handling one of the cases that can break the current wrangler D1 code when parsing the statement. If the original bug report included the semicolon in their text value the issue would still occur.

We need to fix it properly by parsing the statements, or making it more robust with a more generic approach.

PR #8431, opened 2025-03-10
Support autogates via `MINIFLARE_WORKERD_AUTOGATES` env var

penalosa:
Allow for setting the `autogates` property of `workerd` config via an env var: `MINIFLARE_WORKERD_AUTOGATES`. This accepts a space separated list of strings that will be passed directly to `workerd`.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: this should be manually tested, to ensure it correctly enables autogates. cc @danlapid do you have any ideas for how this could be tested in an automated fashion?
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Miniflare-only change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal feature

### DIFF
--- a/.changeset/sweet-fans-arrive.md
+++ b/.changeset/sweet-fans-arrive.md
@@ -1,0 +1,5 @@
+---
+"miniflare": patch
+---
+
+Support `workerd` autogates via the `MINIFLARE_WORKERD_AUTOGATES` environment variable.
--- a/.changeset/sweet-fans-arrive.md
+++ b/.changeset/sweet-fans-arrive.md
@@ -1,5 +1,5 @@
 ---
-"miniflare": patch
+"miniflare": minor
 ---
 
 Support `workerd` autogates via the `MINIFLARE_WORKERD_AUTOGATES` environment variable.
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -2333,6 +2333,9 @@ export class Miniflare {
 			sockets,
 			extensions,
 			structuredLogging: this.#structuredWorkerdLogs,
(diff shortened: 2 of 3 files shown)


danlapid:
I guess you could replicate one of our wd tests that use this but it'll be really annoying because eventually we'll delete that autogate.
We could work really hard to create a test-specific autogate but that's probably not worth the hassle, a manual test is enough imho.

CarmenPopoviciu (approved):
pending a confirmation that this has been manually tested 🙏

penalosa:
Closing for now, since this is no longer high priority

workers-devprod:
- ✅ @cloudflare/wrangler

</details>

danlapid:
Rebased and tested locally and it works.
@penalosa can we merge please?

.changeset/sweet-fans-arrive.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": patch
```
dario-piotrowicz: could this be a `minor`? 🤔

dario-piotrowicz (approved):
Looks good to me, however I wonder if this should have some documentation too? 🤔

workers-devprod (approved):
Codeowners reviews satisfied

penalosa:
> however I wonder if this should have some documentation too? 🤔

This is an internal-only feature, so it doesn't need any

.changeset/sweet-fans-arrive.md
penalosa: It could be? It doesn't really matter though, Miniflare's versioning treats everything as a patch

PR #8432, opened 2025-03-10
Metafiles support in vite-plugin-cloudflare

GregBrimble:
Fixes WC-3285 and WC-3318.

More to come...

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: no coverage
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: will follow once everything else has landed


packages/vite-plugin-cloudflare/package.json
```diff
@@ -44,6 +44,7 @@
 	"dependencies": {
 		"@cloudflare/unenv-preset": "1.1.1",
 		"@hattip/adapter-node": "^0.0.49",
+		"chokidar": "^4.0.1",
```
jamesopstad, line 47: Vite has file watching built in so no need to install chokidar.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -197,12 +214,73 @@ export function getDevMiniflareOptions(
...
 	const assetsConfig =
 		resolvedPluginConfig.type === "assets-only"
 			? resolvedPluginConfig.config.assets
 			: entryWorkerConfig?.assets;
 
+	const headersFile = path.join(resolvedViteConfig.publicDir, HEADERS_FILENAME);
+	const redirectsFile = path.join(
+		resolvedViteConfig.publicDir,
+		REDIRECTS_FILENAME
+	);
```
jamesopstad, line 229: I think the `_headers` and `_redirects` files should be in the project root rather than the `public` directory. The `public` directory is only a small part of assets handling in Vite. They should then be emitted as part of the client bundle at build time but we can handle that in a separate PR.

```suggestion
	const headersFile = path.join(resolvedViteConfig.root, HEADERS_FILENAME);
	const redirectsFile = path.join(
		resolvedViteConfig.root,
		REDIRECTS_FILENAME
	);
```

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -197,12 +214,73 @@ export function getDevMiniflareOptions(
...
+
+	watch([headersFile, redirectsFile], {
+		persistent: true,
+		ignoreInitial: true,
+	}).on("all", async () => {
+		viteDevServer.restart();
+	});
+
```
jamesopstad, line 237: This file watching should be moved to the `handleHotUpdate` hook here:

https://github.com/cloudflare/workers-sdk/blob/869ec7b916487ec43b958a27bdfea13588c5685f/packages/vite-plugin-cloudflare/src/index.ts#L244-L248

I think you'll need to refactor things to move most of the new logic out of this file and then pass the `parsedHeaders` and `parsedRedirects` to the `getDevMiniflareOptions` function.

.changeset/deep-symbols-heal.md
jamesopstad, line 1: Let's remove the changeset as there will need to be subsequent PRs before we can consider this supported.

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/assets.spec.ts
```diff
@@ -20,3 +22,22 @@ test("returns the home page for not found routes", async () => {
 	const content = await page.textContent("h1");
 	expect(content).toBe("Vite + React");
 });
+
+test("applies _headers", async ({}) => {
+	const response = await fetch(viteTestUrl);
+	expect(response.headers.get("X-Header")).toBe("Custom-Value!!!");
+});
```
jamesopstad, line 29: Headers are only being applied to HTML files so, while this test passes, there are other tests with non-HTML assets that wouldn't.

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/assets.spec.ts
```diff
@@ -20,3 +22,22 @@ test("returns the home page for not found routes", async () => {
...
+
+test("applies _redirects", async ({}) => {
+	const response = await fetch(`${viteTestUrl}/foo`, { redirect: "manual" });
+	expect(response.status).toBe(302);
+	expect(response.headers.get("Location")).toBe("/bar");
+});
```
jamesopstad, line 35: Redirects are only being applied if the path doesn't match an existing non-HTML asset so this also wouldn't pass for other asset types.

jamesopstad (commented):
Thanks Greg. Your changes only address HTML files and paths that do not match existing assets. This is because other assets are currently handled by Vite's built in middleware before reaching the Asset Worker.

Happy to merge this once the changes have been addressed and then we can pick up the follow up work that's needed to make this feature fully functional. We're still having ongoing discussions about how to approach this as it poses various challenges.

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/assets.spec.ts
GregBrimble: Left a quick todo test for this

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/assets.spec.ts
GregBrimble: And this one too

packages/vite-plugin-cloudflare/src/miniflare-options.ts
GregBrimble: Not sure I agree on that. Most frameworks (e.g. [Astro](https://docs.astro.build/en/guides/integrations-guide/cloudflare/#headers)) expect users to author these in a public directory.

petebacondarwin: Yeah I think we have gone with these staying in the public directory for now. As tested in https://github.com/cloudflare/workers-sdk/pull/8496.

petebacondarwin (approved):
Let's land this and iterate. It provides basic support. Things to follow up on:

- [x] restarting Vite on change to those files
- [x] adding skipped tests for missing scenarios
- [ ] add experimental support for missing scenarios

Separately we need to add experimental support of `run_worker_first`.

GregBrimble:
> restarting Vite on change to those files

Should already be done here! :) https://github.com/cloudflare/workers-sdk/pull/8432/files#diff-0217c39a7be6fcdb410ac8dfea2ae2998b882a1aaba3f2e83ff2831a0beee7bfR283-R289

petebacondarwin (changes requested):
Ah just going to backtrack on this approval. The code itself is great and we should land it!

But as discussed internally, we are going to initially land any support for headers and redirects in dev mode (it already works in preview) behind an experimental flag.

This makes initial documentation simpler: "headers and redirects will be actioned in preview and production, not in dev" (for Vite).

So to avoid @GregBrimble from having to keep churning on this PR, I'll add a commit to put it behind a flag, and then land it.

petebacondarwin (approved):
But of course I'd approve, since I wrote some of this 😆

GregBrimble (commented):
✅ Can't approve my own PR, but LGTM

packages/vite-plugin-cloudflare/src/asset-config.ts
```diff
@@ -0,0 +1,128 @@
...
+						: {}),
+					...(entryWorkerConfig?.compatibility_flags
+						? { compatibility_flags: entryWorkerConfig?.compatibility_flags }
+						: {}),
+				};
+
+	if (!assetConfig) {
+		return compatibilityOptions;
+	}
+
```
jamesopstad, line 73: This needs to be removed. Users can have assets without having the `assets` field defined in the config.

```suggestion
```

packages/vite-plugin-cloudflare/src/asset-config.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/**
+ * Computes the assets config that will be passed to Miniflare,
+ * taking into account whether experimental _headers and _redirects support is on.
+ */
+export function getAssetsConfig(
+	resolvedPluginConfig: ResolvedPluginConfig,
+	entryWorkerConfig: Unstable_Config | undefined,
+	viteLogger: Log,
```
jamesopstad, line 47: Why are we importing `Log` from Miniflare and then calling it `viteLogger`. Can't we use the Vite logger directly?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -197,26 +195,15 @@ export function getDevMiniflareOptions(
 	viteDevServer: vite.ViteDevServer
 ): MiniflareOptions {
 	const resolvedViteConfig = viteDevServer.config;
+	const logger = new ViteMiniflareLogger(resolvedViteConfig);
```
jamesopstad, line 198: There may be good reason for moving this here but if it's possible to use `resolvedViteConfig.logger` directly then that would be preferable. Logging about the metafiles isn't really related to Miniflare.

packages/vite-plugin-cloudflare/src/asset-config.ts
```diff
@@ -0,0 +1,128 @@
...
+					compatibility_date: resolvedPluginConfig.config.compatibility_date,
+					compatibility_flags: resolvedPluginConfig.config.compatibility_flags,
+				}
+			: {
+					...(entryWorkerConfig?.compatibility_date
+						? { compatibility_date: entryWorkerConfig?.compatibility_date }
+						: {}),
+					...(entryWorkerConfig?.compatibility_flags
+						? { compatibility_flags: entryWorkerConfig?.compatibility_flags }
+						: {}),
```
jamesopstad, line 67:
```suggestion
					...(entryWorkerConfig?.compatibility_date
						? { compatibility_date: entryWorkerConfig.compatibility_date }
						: {}),
					...(entryWorkerConfig?.compatibility_flags
						? { compatibility_flags: entryWorkerConfig.compatibility_flags }
						: {}),
```

packages/vite-plugin-cloudflare/src/asset-config.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Computes the assets config that will be passed to Miniflare,
+ * taking into account whether experimental _headers and _redirects support is on.
+ */
+export function getAssetsConfig(
+	resolvedPluginConfig: ResolvedPluginConfig,
+	entryWorkerConfig: Unstable_Config | undefined,
+	viteLogger: Log,
+	resolvedConfig: ResolvedConfig
+) {
+	const assetConfig =
```
jamesopstad, line 50:
```suggestion
	const assetsConfig =
```

packages/vite-plugin-cloudflare/src/asset-config.ts
```diff
@@ -0,0 +1,128 @@
...
+						: {}),
+				};
+
+	if (!assetConfig) {
+		return compatibilityOptions;
+	}
+
+	const config = {
+		...compatibilityOptions,
+		...assetConfig,
```
jamesopstad, line 76:
```suggestion
		...assetsConfig,
```

.changeset/deep-symbols-heal.md
```diff
@@ -0,0 +1,19 @@
...
+cloudflare({
+	// ...
+	experimental: { headersAndRedirectsDevModeSupport: true },
+}),
+```
+
+Currently, in this experimental mode, requests that would result in an HTML response or a 404 response will take into account the \_headers and \_redirects settings.
+Known limitation: requests for existing static assets will be served directly by Vite without considering the \_headers or \_redirects settings.
+
+A production deployment or `vite preview` already accurately supports these settings.
```
jamesopstad, line 19: My main worry with including it in the changelog is that it invites people to try it out. If we put it in the changelog, can we get changesets to clearly mark this as unstable? Have a look at this React Router changelog for an example of what I mean - https://github.com/remix-run/react-router/blob/main/CHANGELOG.md#unstable-changes.

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/experimental-headers-and-redirects/assets.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+			expect(redirectBefore.status).toBe(302);
+
+			// We make both these changes at the same time because there is something strange about the test setup
+			// where fetches result in 500 errors, due to Miniflare stubs being reused after disposal.
+			writeFileSync(headersPath, "");
+			writeFileSync(redirectsPath, "");
+
+			// Wait for Vite to reload
+			await setTimeout(500);
+
```
jamesopstad, line 107: A little concerned that this could lead to the tests being quite flaky.

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/assets.spec.ts
```diff
@@ -1,22 +1,71 @@
-import { expect, test } from "vitest";
-import { page, viteTestUrl } from "../../__test-utils__";
+import { describe, expect, test } from "vitest";
+import { isBuild, page, viteTestUrl } from "../../__test-utils__";
 
-test("returns the correct home page", async () => {
-	const content = await page.textContent("h1");
-	expect(content).toBe("Vite + React");
-});
+describe("react-spa", { sequential: true, concurrent: false }, () => {
```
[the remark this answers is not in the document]
petebacondarwin: These might not be needed now. I was having to hack around a bit before. I'll try removing them.

packages/vite-plugin-cloudflare/src/asset-config.ts
petebacondarwin: OK

packages/vite-plugin-cloudflare/src/asset-config.ts
petebacondarwin: So strictly its a `ViteMiniflareLogger`. So it is a bit of both 😄 . Can rename.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
petebacondarwin: It's a different kind of logger with different methods, so it needs to be wrapped. The `ViteMiniflareLogger` is a wrapper around the Vite logger that Miniflare understands.

.changeset/deep-symbols-heal.md
petebacondarwin: I don't think changesets can do that. I can make this look more scary and indicate it might disappear.

But I feel like it needs a changeset of some kind, right?

jamesopstad: I'll let you decide what's best. Just flagging up the concern.

packages/vite-plugin-cloudflare/src/asset-config.ts
petebacondarwin: Is it possible to have a Vite project with no assets?
If so, how is that determined?

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/experimental-headers-and-redirects/assets.spec.ts
petebacondarwin: Ahah! There is a rather nice `expect.poll()` helper!

packages/vite-plugin-cloudflare/src/miniflare-options.ts
petebacondarwin: Oh! I just "actually" read your comment and I realise that you are right.

"Logging about the metafiles isn't really related to Miniflare." is the key sentence. We just need to wrap the vite logger to pass to the metafile loader, rather than a double wrap.

packages/vite-plugin-cloudflare/src/asset-config.ts
petebacondarwin: 5430df921ca18ed5497f4b19ec58307f08d89ace

packages/vite-plugin-cloudflare/src/miniflare-options.ts
petebacondarwin: 5430df921ca18ed5497f4b19ec58307f08d89ace

packages/vite-plugin-cloudflare/src/asset-config.ts
petebacondarwin: 5430df921ca18ed5497f4b19ec58307f08d89ace

packages/vite-plugin-cloudflare/src/asset-config.ts
petebacondarwin: 5430df921ca18ed5497f4b19ec58307f08d89ace

packages/vite-plugin-cloudflare/playground/react-spa/__tests__/experimental-headers-and-redirects/assets.spec.ts
petebacondarwin: 5430df921ca18ed5497f4b19ec58307f08d89ace

packages/vite-plugin-cloudflare/src/asset-config.ts
petebacondarwin: Or remove! 5430df921ca18ed5497f4b19ec58307f08d89ace

jamesopstad (approved):

PR #8441, opened 2025-03-11
Dario/vite plugin better debugging experience

dario-piotrowicz:

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 


packages/vite-plugin-cloudflare/README.md
```diff
@@ -482,6 +482,59 @@ Secrets can be provided to your Worker in local development using a [`.dev.vars`
...
+    }
+  ],
+  "compounds": [
+    {
+      "name": "Debug Workers",
+      "configurations": ["worker1"],
+      "stopAll": true
+    }
+  ]
+}
```
dario-piotrowicz, line 527: TODO: I need to validate this and see if all the fields are actually necessary

dario-piotrowicz: done 🙂

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -286,15 +311,32 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+						pluginConfig.persistState ?? true,
+						pluginConfig.inspectorPort
 					)
 				);
 
+				if (workerConfigs.length >= 1) {
+					addDebugToVitePrintUrls(vitePreviewServer);
+				}
+
+				const workerNames = workerConfigs.map((worker) => worker.name ?? "");
```
petebacondarwin, line 329: What does it mean for a worker to have no name here?

packages/vite-plugin-cloudflare/playground/cloudflare-env/vite.config.custom-mode.ts
```diff
@@ -3,5 +3,5 @@ import { defineConfig } from "vite";
 
 export default defineConfig({
 	mode: "custom-mode",
-	plugins: [cloudflare({ persistState: false })],
+	plugins: [cloudflare({ inspectorPort: 0, persistState: false })],
```
petebacondarwin, line 6: What effect does setting this to 0 have?
Does it ask the plugin to just choose its own free port, or does it disable debugging altogether.
If the former, is there a mechanism for disabling the inspector?

.changeset/red-cow-eats.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cloudflare/vite-plugin": minor
```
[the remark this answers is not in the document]
dario-piotrowicz: is this not a minor? :thinking:

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: mh.... I need to double check, in miniflare you can provide workers without names, but I actually think that the name is mandated in `readConfig`? so it might be that the name is always defined? :thinking:

jamesopstad (commented):
Looks great! I think that it would be nice to split the functionality to print the URL and serve the HTML into a separate Vite plugin (like we have for additional modules and node_compat) that adds its own middleware (normal Node one) at the front of the chain.  Also happy for us to refactor later if you would rather not do that now though.

dario-piotrowicz:
> I think that it would be nice to split the functionality to print the URL and serve the HTML into a separate Vite plugin (like we have for additional modules and node_compat) that adds its own middleware (normal Node one) at the front of the chain.

@jamesopstad thanks for the suggestion, yeah that's totally spot on, I started with that idea but then I just implemented the thing and forgot about it, I'll look into splitting it into its own plugin, thanks for reminding me! 🙏

packages/vite-plugin-cloudflare/README.md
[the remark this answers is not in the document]
dario-piotrowicz: oops! 😓👍

packages/vite-plugin-cloudflare/README.md
```diff
@@ -482,6 +482,62 @@ Secrets can be provided to your Worker in local development using a [`.dev.vars`
...
+      "resolveSourceMapLocations": null,
+      "attachExistingChildren": false,
+      "autoAttachChildProcesses": false,
+      "sourceMaps": true
+    }
+  ],
+  "compounds": [
+    {
+      "name": "Debug All Workers",
+      "configurations": ["Worker1"],
```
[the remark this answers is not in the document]
dario-piotrowicz: yes, I guess this doesn't make it clear? 😓

dario-piotrowicz:
> @jamesopstad thanks for the suggestion, yeah that's totally spot on, I started with that idea but then I just implemented the thing and forgot about it, I'll look into splitting it into its own plugin, thanks for reminding me! 🙏

Ah no, now I remember why I didn't went with that, because the fact that the inspectorPort is part of the main plugin made this looks less clear/beneficial, since there is already coupling between inspector port and the url 🤔 

So yeah not super beneficial, but I'm also not against the idea 🙂👍

jamesopstad:
> > @jamesopstad thanks for the suggestion, yeah that's totally spot on, I started with that idea but then I just implemented the thing and forgot about it, I'll look into splitting it into its own plugin, thanks for reminding me! 🙏
> 
> Ah no, now I remember why I didn't went with that, because the fact that the inspectorPort is part of the main plugin made this looks less clear/beneficial, since there is already coupling between inspector port and the url 🤔
> 
> So yeah not super beneficial, but I'm also not against the idea 🙂👍

Yeah, I see what you mean. I think you could see the inspector itself and the browser debugging as two separate features though. So enabling the inspector would be part of the main plugin but the browser debugging (with the printed URL etc.) is a separate plugin.

dario-piotrowicz:
@jamesopstad oof... this makes separating this logic in its own plugin pretty cumbersome! 😓 

https://github.com/cloudflare/workers-sdk/blob/62d5471eae9b5ed8cb31f025fa23ba3930b94317/packages/vite-plugin-cloudflare/src/index.ts#L307-L310

jamesopstad:
> @jamesopstad oof... this makes separating this logic in its own plugin pretty cumbersome! 😓
> 
> https://github.com/cloudflare/workers-sdk/blob/62d5471eae9b5ed8cb31f025fa23ba3930b94317/packages/vite-plugin-cloudflare/src/index.ts#L307-L310
> 
> I also wonder if messing with middlewares at this point can get in the way of the metadata stuff @petebacondarwin is implementing.... I'm starting to be pretty tempted to indeed keep this as is for now and maybe refactor it later on (during the freeze for example)

Good point. Yeah, I think keep it as it is. No need to spend more time on it.

dario-piotrowicz:
> Good point. Yeah, I think keep it as it is. No need to spend more time on it.

Sorry I just addressed the above, it's a bit awkward but moving the plugin at the top of the plugins array seems to work ok, I'm happy to revert this to the previous version if you prefer 🙂

jamesopstad:
> > Good point. Yeah, I think keep it as it is. No need to spend more time on it.
> 
> Sorry I just addressed the above, it's a bit awkward but moving the plugin at the top of the plugins array seems to work ok, I'm happy to revert this to the previous version if you prefer 🙂

You go with whichever you think makes more sense.

dario-piotrowicz:
> > > Good point. Yeah, I think keep it as it is. No need to spend more time on it.
> > 
> > 
> > Sorry I just addressed the above, it's a bit awkward but moving the plugin at the top of the plugins array seems to work ok, I'm happy to revert this to the previous version if you prefer 🙂
> 
> You go with whichever you think makes more sense.

honestly I don't have a huge preference, since I already have this implementation working we can progress with it if you're ok with it 🙂

packages/vite-plugin-cloudflare/README.md
```diff
@@ -374,6 +374,10 @@ It accepts an optional `PluginConfig` parameter.
   All requests are routed through your entry Worker.
   During the build, each Worker is output to a separate subdirectory of `dist`.
 
+- `inspectorPort?: number | false`
+
+  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. Defaults to `9229`.
```
dario-piotrowicz, line 379: I'm purposely not mentioning `0` as an option here as I've noticed that that is probably a bit buggy right now, I'll look into that in a follow up PR (if that's ok by you) and once it works correctly I'll update the text here.

packages/vite-plugin-cloudflare/README.md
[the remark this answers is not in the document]
dario-piotrowicz: done! 🙂

packages/vite-plugin-cloudflare/playground/cloudflare-env/vite.config.custom-mode.ts
dario-piotrowicz: Now we accept `false` too and I've updated the playground apps to use that 🙂👍

packages/vite-plugin-cloudflare/README.md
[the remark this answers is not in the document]
dario-piotrowicz: By the way, I also want to mention.... this configuration is a bit silly for a single worker... (since it gives you a configuration for "all workers" which only includes a single worker + the configuration for that single worker 😅)

The nice part is that it can be generalized for 1 or more workers (instead of the README being more verbose having separate instructions for one or multiple workers 😕)

.changeset/red-cow-eats.md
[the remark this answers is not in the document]
penalosa: No, because the vite plugin is still 0.x

.changeset/red-cow-eats.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+add `/__debug` path for better debugging
```
penalosa, line 5: It's unlikely, but there's a chance this could conflict with a user path. What about using `/cdn-cgi/debug`? The entire cdn-cgi path is blocked in production, and we use that path for things like triggering scheduled handlers already

packages/vite-plugin-cloudflare/README.md
```diff
@@ -482,6 +486,62 @@ Secrets can be provided to your Worker in local development using a [`.dev.vars`
 > [!NOTE]
 > The `vite build` command copies the relevant `.dev.vars[.env-name]` file to the output directory. This is only used when running `vite preview` and is not deployed with your Worker.
 
+## Debugging
+
+The Cloudflare Vite plugin allows you to conveniently debug your Worker code during local development.
+
+By default the inspector port used by the plugin is `9229`, you can customize such by providing a different port to the plugin's `inspectorPort` option.
```
penalosa, line 493:
```suggestion
By default the inspector port used by the plugin is `9229`, which can be customized by providing a different port to the plugin's `inspectorPort` option.
```

packages/vite-plugin-cloudflare/README.md
```diff
@@ -482,6 +486,62 @@ Secrets can be provided to your Worker in local development using a [`.dev.vars`
...
+
+There are two recommended ways of doing so, which we'll explore in this section.
+
+### Devtools
+
+When running `vite dev` or `vite preview` a convenient `/__debug` route in your local server will be made available which gives you access to [Cloudflare's implementation](/packages/chrome-devtools-patches) of [Chrome's DevTools](https://developer.chrome.com/docs/devtools/overview).
```
penalosa, line 499:
```suggestion
When running `vite dev` or `vite preview` a `/__debug` route in your local server will be made available which gives you access to [Cloudflare's implementation](/packages/chrome-devtools-patches) of [Chrome's DevTools](https://developer.chrome.com/docs/devtools/overview).
```

packages/vite-plugin-cloudflare/README.md
```diff
@@ -482,6 +486,62 @@ Secrets can be provided to your Worker in local development using a [`.dev.vars`
...
+
+Navigating to such route will open a devtools tab for each of the workers in your application (Note: In case of multiple workers you might need to allow your browser to open pop-ups).
```
penalosa, line 501:
```suggestion
Navigating to this route will open a devtools tab for each of the workers in your application (Note: in case of multiple workers you might need to allow your browser to open pop-ups).
```

packages/vite-plugin-cloudflare/README.md
```diff
@@ -482,6 +486,62 @@ Secrets can be provided to your Worker in local development using a [`.dev.vars`
...
+}
+```
+
+Where, `<NAME_OF_WORKER>` indicates the name of your worker as specified in your Wrangler configuration.
+
+Note: if you customized your `inspectorPort` you need to use that port in the `websocketAddress` field.
+
+If you have more than one worker you need add a configuration in the `configurations` field for each one and then include the configuration name in the `configurations` array in the compound configuration.
+
+Once you're `launch.json` file is ready, after running `vite dev` or `vite preview` you can select **Debug All Workers** at the top of the **Run & Debug** panel to attach debuggers to all the various Workers. Then you can add breakpoints to your code and start debugging.
```
penalosa, line 541:
```suggestion
Once your `launch.json` file is ready, after running `vite dev` or `vite preview` you can select **Debug All Workers** at the top of the **Run & Debug** panel to attach debuggers to all the various Workers. Then you can add breakpoints to your code and start debugging.
```

packages/vite-plugin-cloudflare/src/debugging.ts
```diff
@@ -0,0 +1,71 @@
...
+	};
+}
+
+/**
+ * Generate an HTML text that comprises of a single script that:
+ *  - redirects the page to the devtools for the debugging of the first available worker
+ *  - opens tags to the devtools for all the remaining workers if any
+ *
+ * Note: this works based on the miniflare inspector proxy logic (where workers are available via
+ * paths comprized of their names)
```
penalosa, line 43:
```suggestion
 * paths comprised of their names)
```

packages/vite-plugin-cloudflare/src/debugging.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param inspectorPort the inspector port that miniflare is using
+ * @returns the generated html
+ */
+export function getDebugPathHtml(workerNames: string[], inspectorPort: number) {
+	// this function should always be called only when there is at least one worker to debug
+	assert(workerNames.length >= 1, "no workers present to debug");
+
+	const workerDevtoolsUrls = workerNames.map((workerName) => {
+		const localHost = `localhost:${inspectorPort}/${workerName}`;
+		const devtoolsFrontendUrl = `https://devtools.devprod.cloudflare.dev/js_app?theme=systemPreferred&debugger=true&ws=${localHost}`;
```
penalosa, line 55: This should also have a `domain` parameter set to the name of the worker e.g. https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/src/dev/inspect.ts#L256-L257

packages/vite-plugin-cloudflare/src/debugging.ts
```diff
@@ -0,0 +1,71 @@
...
+		return devtoolsFrontendUrl;
+	});
+
+	return `
+		<script>
+			const workerUrls = [${workerDevtoolsUrls.map((str) => JSON.stringify(str)).join(", ")}];
```
penalosa, line 61: nit, but could this just be `JSON.stringify(workerDevtoolsUrls)`?

penalosa (approved):
Some nits, but LGTM!

.changeset/red-cow-eats.md
dario-piotrowicz: I think you're totally right, but I don't like `/cdn-cgi/debug` 🥹 

Also I chose `/__debug` following the [`vite-plugin-inspect`](https://www.npmjs.com/package/vite-plugin-inspect) plugin which uses `/__inspect`, making `/__debug` more consistent with the vite ecosystem (and making it not feel out of place)

So I would prefer to keep `/__debug` assuming that there is almost no chance for conflicts, but I am happy to reconsider if you or someone else on the team has a strong opinion on this

packages/vite-plugin-cloudflare/README.md
dario-piotrowicz: 🤦

packages/vite-plugin-cloudflare/src/debugging.ts
dario-piotrowicz: oh yeah... I think so! 😅

dario-piotrowicz: updated, thanks! 🙂👍

packages/vite-plugin-cloudflare/src/debugging.ts
dario-piotrowicz: Done 🙂 

Does this look good to you?
[f4b106e](https://github.com/cloudflare/workers-sdk/pull/8441/commits/f4b106e0abf688ecb6c300a9e87065e5acd77ab1)

Also, do you know what this does? I can't really notice any difference with or without it 🤔

.changeset/red-cow-eats.md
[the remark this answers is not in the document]
dario-piotrowicz: mh... ok updated 🙂

.changeset/polite-walls-bathe.md
```diff
@@ -0,0 +1,8 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+add `inspectorPort` option to plugin config
+
+add an `inspectorPort` option that allows developers to start a devTools inspector server
+that allows them to debug their workers (defaulting to `9229`)
```
jamesopstad, line 8:
```suggestion
add an `inspectorPort` option that allows developers to start a devTools inspector server to debug their workers (defaulting to `9229`)
```

packages/vite-plugin-cloudflare/package.json
```diff
@@ -45,6 +45,7 @@
 		"@cloudflare/unenv-preset": "workspace:*",
 		"@hattip/adapter-node": "^0.0.49",
 		"miniflare": "workspace:*",
+		"picocolors": "^1.1.1",
```
jamesopstad, line 48: I think this one should be dev dependency so that it's bundled. Following Vite here.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -60,6 +66,69 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 	let hasClientBuild = false;
 
 	return [
+		// Plugin that provides an __debug path for debugging the Cloudflare Workers.
```
jamesopstad, line 69: Can you try moving this plugin after the main plugin and then using the object form of `configureServer` and `configurePreviewServer` e.g.

```
configureServer: {
  order: 'pre',
  handler(viteDevServer) {
    // ...
  }
};
```

Should work but if it doesn't you can leave it as it is.

jamesopstad (approved):
Nice!

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Actually, you should just be able to set `enforce: 'pre'` on the plugin, no? Let me check.

jamesopstad: Yep, that works. So, yeah, if you can move the plugin after the main one and set `enfore: 'pre'` that would be great. Thanks!

dario-piotrowicz: I don't really like the fact that `enforce: 'pre'` then put this before all userland plugins 😕 (moving it to the top of our array is a change only scoped to us), but it's not a huge deal, if that's your preference I'm ok with it 👍

.changeset/red-cow-eats.md
penalosa: Totally fair—happy with __debug in that case

penalosa:
For posterity, merging with a flaky Linux Node 18 test, since Node 20 & 22 are passing, and the failing test i unrelated to this PR

PR #8451, opened 2025-03-12
Update Next.js version in the experimental template

vicb:
Update Next.js to ~15.2.2

Also:
- drop minify (for consistency with other templates + code already minified by OpenNext)
- simplify the config
- handle `next.config.ts` which is now generated by the Next CLI in TS mode (similar to what is done for Next on Pages)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: already updated

### DIFF
--- a/packages/create-cloudflare/templates-experimental/next/templates/wrangler.jsonc
+++ b/packages/create-cloudflare/templates-experimental/next/templates/wrangler.jsonc
@@ -5,7 +5,6 @@
   "compatibility_flags": [
     "nodejs_compat"
   ],
-  "minify": true,
   "assets": {
     "binding": "ASSETS",
     "directory": ".open-next/assets"
--- a/.changeset/real-clocks-ask.md
+++ b/.changeset/real-clocks-ask.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+Update the Next experimental template to use version ~15.2.2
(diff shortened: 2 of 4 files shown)


packages/create-cloudflare/templates-experimental/next/templates/wrangler.jsonc
```diff
@@ -5,7 +5,6 @@
   "compatibility_flags": [
     "nodejs_compat"
   ],
-  "minify": true,
```
dario-piotrowicz, line 8: why removing this? (is true the default)?

dario-piotrowicz (approved):

packages/create-cloudflare/templates-experimental/next/templates/wrangler.jsonc
vicb: The default is `false`.
The PR removes this to be consistent with other template and to help with inspecting the code.
I don't think there is any cons?

vicb: I forgot to mention that the code is minified by OpenNext unless the env var `OPEN_NEXT_DEBUG` is set to 1

PR #8455, opened 2025-03-12
test: add console log "grouping" to Wrangler e2e tests

petebacondarwin:
This should make it easier to see what is happening in CI.


dario-piotrowicz (approved):
Looks good to me 🙂 

Although I couldn't notice any difference in this PR wrangler e2es 🤔 

is this one of those changes that we can actually see only after the PR's been merged? 😕

PR #8477, opened 2025-03-13
wrangler deploy: Add containers configuration to PUT script

gabivlj:
Right now, container configuration only works on the PUT script path, but this will be fixed in further API changes.

This change makes sure that when container configurations are included in wrangler, it will include the `containers` configuration in wrangler deploy. These changes break `wrangler cloudchamber apply` users, but as the product is in early-stages and private only for now, we consider these breaking changes worth doing. 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: No e2e tests for containers yet.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Internal features

### DIFF
--- a/.changeset/common-paws-invite.md
+++ b/.changeset/common-paws-invite.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+wrangler deploy includes container configuration when uploading the script
--- a/packages/wrangler/src/config/environment.ts
+++ b/packages/wrangler/src/config/environment.ts
@@ -49,6 +49,12 @@ export type ContainerApp = {
 	name: string;
 	/* Number of application instances */
 	instances: number;
+
+	/* The user can set both image in here or under configuration */
+	image?: string;
+
+	class_name: string;
+
 	/* The scheduling policy of the application, default is regional */
 	scheduling_policy?: "regional" | "moon";
 	/* Configuration of the container */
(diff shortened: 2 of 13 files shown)


.changeset/common-paws-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
emily-shen, line 2: I think this should be a patch as cloudchamber is still experimental?

packages/wrangler/src/config/environment.ts
```diff
@@ -458,18 +464,7 @@ export interface EnvironmentNonInheritable {
...
-		 *
-		 * NOTE: This field is not automatically inherited from the top level environment,
-		 * and so must be specified in every named environment.
-		 *
-		 * @default {}
-		 * @nonInheritable
-		 */
-		app: ContainerApp[];
-	};
+	containers?: ContainerApp[];
```
emily-shen, line 467: should still have the @default and @nonInheritable tags i assume?

packages/wrangler/src/config/validation.ts
```diff
@@ -2210,47 +2207,45 @@ const validateContainerAppConfig: ValidatorFn = (
...
-	if (!("app" in value)) {
-		return true;
-	}
-
-	value = value.app;
 	if (!Array.isArray(value)) {
-		diagnostics.errors.push(
-			`"containers.app" should be an array, but got ${JSON.stringify(value)}`
-		);
+		diagnostics.errors.push(`"containers" should be an array`);
```
emily-shen, line 2211:
```suggestion
		diagnostics.errors.push(`"containers" should be an array`) but got ${JSON.stringify(value)}`
```

emily-shen (approved):

th0m (approved):

gabivlj:
thx for the review @emily-shen !

PR #8489, opened 2025-03-14
[vite-plugin]: make sure that `process.env` is correctly populated

### LINKED ISSUE #8457: vite-plugin: improve workerd process.env support
Currently we set `keepProcessEnv` to `false`:
https://github.com/cloudflare/workers-sdk/blob/869ec7b916487ec43b958a27bdfea13588c5685f/packages/vite-plugin-cloudflare/src/cloudflare-environment.ts#L183

meaning that vite replaces `process.env` code at build time

this prevents workerd from correctly populating `process.env`, we should update `keepProcessEnv` to be `true` instead

 

As part of this we should also explore and see if we can get rid of the `_VITE_ROOT` and `VITE_ENTRY_PATH_` environment variables as they would pollute the `process.env` object

dario-piotrowicz:
Fixes #8457

This PR does two things to make sure that the `process.env` object is correctly populated:
 - it updates `keepProcessEnv` to be `true` when the nodejs_compat and `nodejs_compat_populate_process_env` flags are set
 - removes the `__VITE_ROOT__` and `__VITE_ENTRY_PATH__` bindings that pollute that would otherwise populated `process.env`

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not a wrangler change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 


packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -125,6 +134,12 @@ export function createCloudflareEnvironmentOptions(
 	userConfig: vite.UserConfig,
 	environmentName: string
 ): vite.EnvironmentOptions {
+	const workerCompatFlags = workerConfig.compatibility_flags;
+
+	const workerdPopulatesEnv =
+		workerCompatFlags.includes("nodejs_compat") &&
+		workerCompatFlags.includes("nodejs_compat_populate_process_env");
```
threepointone, line 141: this should be true also if nodejs_compat is true and compat date >= 2025-04-01

dario-piotrowicz: ah, ok thanks 🙂 

I was looking for a compat date in the [workerd PR](https://github.com/cloudflare/workerd/pull/3610) but couldn't find one, I assumed this wouldn't be turned on by default automatically then 🤔

dario-piotrowicz: sorry I was multitasking and looking at the wrong PR! 🤦‍♂️ 😅 

https://github.com/cloudflare/workerd/blob/bd8abeabdf633bb274bf7e0e7f53a9e8656e465e/src/workerd/io/compatibility-date.capnp#L720

thanks again 🙂

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -118,3 +118,33 @@ function getNodeCompatEntries() {
...
+
+	if (!workerCompatFlags.includes("nodejs_compat")) {
+		return false;
+	}
+
+	if (workerCompatFlags.includes("nodejs_compat_populate_process_env")) {
+		return true;
+	}
+
+	const compatDateEnablingProcessEnvPopulation = new Date("2025-04-01");
```
dario-piotrowicz, line 139: Note: I couldn't add playground/e2e tests for this scenario, because we get a miniflare error when a future date is being used:
Screenshot 2025-03-14 at 16 36 16

I'll create an issue to add these tests after the first or april if that's ok with everyone

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
dario-piotrowicz: Updated, please have another look @threepointone 😄

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -118,3 +118,33 @@ function getNodeCompatEntries() {
 
 	return entries;
 }
+
+/**
+ * Discerns wether workerd would populate the `process.env` object a worker or not
```
petebacondarwin, line 123:
```suggestion
 * Discerns whether workerd would populate the `process.env` object in a worker or not
```

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -118,3 +118,33 @@ function getNodeCompatEntries() {
...
+ *
+ * @param workerConfig the worker configuration
+ * @returns true if workerd would populate the `process.env` given the provided configuration, false otherwise
+ */
+export function wouldWorkerdPopulateProcessEnv(workerConfig: WorkerConfig) {
+	const workerCompatFlags = workerConfig.compatibility_flags;
+
+	if (!workerCompatFlags.includes("nodejs_compat")) {
```
petebacondarwin, line 131: We should use the `isNodeCompat()` function here, since it is not as simple as just checking for this flag. E.g. someone might still be using `nodejs_compat_v2` etc.

petebacondarwin (changes requested):

packages/vite-plugin-cloudflare/src/node-js-compat.ts
dario-piotrowicz: mh... that's a good point... but I was working off of the [workerd implementation](https://github.com/cloudflare/workerd/blob/a36ee3c82e95b69560ac1d1491b6acc7e17806cc/src/workerd/io/compatibility-date.capnp#L718-L723) which doesn't seem to keep v2 into account

I checked this with wrangler dev and with `nodejs_compat` we do get the populated `process.env` but we don't with `nodejs_compat_v2`:
| `nodejs_compat` | `nodejs_compat_v2` |
|--------|--------|
| Screenshot 2025-03-17 at 10 12 41 | Screenshot 2025-03-17 at 10 13 14 |

So I think that my implementation is technically correct 🤔 , but I think we ought to check with the runtime team if this behavior is intentional (probably not, unless moving forward, for maintainability reasons they're not going to bother with v2?)

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -310,8 +310,6 @@ export function getDevMiniflareOptions(
 								unsafeEvalBinding: "__VITE_UNSAFE_EVAL__",
 								bindings: {
```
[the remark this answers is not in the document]
dario-piotrowicz: Sorry, what do you mean?

We do still need to pass the worker bindings through, are you suggesting to do
`bindings: workerOptions.bindings,`
instead of spreading?

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -83,15 +85,23 @@ export class CloudflareDevEnvironment extends vite.DevEnvironment {
...
+		workerConfig: WorkerConfig
+	) {
 		this.#worker = worker;
 
 		const response = await this.#worker.fetch(
 			new URL(INIT_PATH, UNKNOWN_HOST),
 			{
 				headers: {
+					[VITE_DEV_METADATA_HEADER]: JSON.stringify({
+						root: this.#viteResolvedConfig.root,
```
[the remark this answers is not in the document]
dario-piotrowicz: Ah yeah good suggestion! 🤔

packages/vite-plugin-cloudflare/playground/node-compat/vite.config.worker-process-populated-env.ts
```diff
@@ -0,0 +1,14 @@
+import { cloudflare } from "@cloudflare/vite-plugin";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+	build: {
+		outDir: "dist/worker-process",
```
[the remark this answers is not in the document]
dario-piotrowicz: good spot! 😓👍

packages/vite-plugin-cloudflare/src/miniflare-options.ts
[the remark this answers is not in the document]
jamesopstad: Because `...workerOptions` is included here, you only need to include a value again if something is being added.

https://github.com/cloudflare/workers-sdk/blob/3e3f23d62c4bbe4db91873daed37abe719faece0/packages/vite-plugin-cloudflare/src/miniflare-options.ts#L307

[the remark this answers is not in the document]
dario-piotrowicz: ahhh, of course! sorry, yeah thanks! 🙂👍

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -165,16 +166,17 @@ export function createWorkerEntrypointWrapper(
...
 		Wrapper.prototype[key] = async function (arg) {
-			const entryPath = this.env.__VITE_ENTRY_PATH__;
-
 			if (key === "fetch") {
 				const request = arg as Request;
 				const url = new URL(request.url);
 
 				if (url.pathname === INIT_PATH) {
+					const viteDevMetadata: { root: string; entryPath: string } =
+						JSON.parse(request.headers.get(VITE_DEV_METADATA_HEADER) ?? "null");
```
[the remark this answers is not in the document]
dario-piotrowicz: Done 🙂 

Actually also before I was considering adding some validation here and since you brought it up I thought I might as well add the full validation 🙂 (this could be a bit useless but it could come in handy in case we break the functionality during dev or something like that)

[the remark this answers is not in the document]
dario-piotrowicz: Please have a look 🙂

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -180,7 +190,9 @@ export function createCloudflareEnvironmentOptions(
 				],
 			},
 		},
-		keepProcessEnv: false,
+		// if workerd populates process.env then we need to let it do that,
+		// otherwise vite itself can own process.env
+		keepProcessEnv: wouldWorkerdPopulateProcessEnv(workerConfig),
```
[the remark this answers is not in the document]
dario-piotrowicz: My reasoning is, we decided to set this to false, meaning that vite would control `process.env`, we are now changing curse only because workerd can populate the `process.env` object, that's the only reason why we're changing this right? so it's also the only case that we're basically unhappy with, isn't it?

What I mean is that:
 - if workerd populates `process.env` then we want to leave control of `process.env` to workerd (because it needs to populate it obviously)
 - if workerd doesn't populate `process.env` then we can leave workerd to control it, since nodejs compat or not workerd is not going to do anything specific on `process.env` right? so there isn't really a benefit for it to own the object
 
 Or in other terms, what is the benefit in leaving `process.env` to be handled by workerd when only `nodejs_compat` is set? (meaning that workerd doesn't do anything specific with `process.env`)
 
 ___
 
Anyways that's the reason why I went with this approach, but I don't mind setting `keepProcessEnv` to true when nodejsCompat is enabled since there isn't a real downside to that (AFAICT), and also after the first of april the nodejs_compat_populate_process_env flag will be set by default anyways (meaning that the vast majority of people using nodejs_compat will also have this flag set)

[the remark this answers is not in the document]
dario-piotrowicz: updated to always set `keepProcessEnv` to `true` when nodejs compat is enabled 🙂

packages/vite-plugin-cloudflare/src/node-js-compat.ts
dario-piotrowicz: This is no longer applicable since I've removed this `wouldWorkerdPopulateProcessEnv` utility and am using `isNodeCompat` directly instead 🙂 

I'm making this as resolved then, I'll check with the runtime team regarding my v2 question 🙂 (but that's no longer related to this PR I think)

packages/vite-plugin-cloudflare/playground/module-resolution/__tests__/module-resolution.spec.ts
```diff
@@ -1,11 +1,11 @@
+// Note: this file should be kept in sync with ./nodejs-compat/module-resolution.spec.ts
```
dario-piotrowicz, line 1: this is annoying... but I don't think there's a way to do this better 😓 

If we were to copy the playground in a playground-tmp folder like we used to probably we could come up with some convention that could copy files over multiple variants, I can't remember why we removed the tmp folder, but I guess it's not worth reintroducing it for this

any thoughts? 🤔

dario-piotrowicz:
@petebacondarwin , @jamesopstad thanks for the helpful reviews 🙂 

I've updated the PR to address the various comments/suggestions, please have another look 🙏

packages/vite-plugin-cloudflare/playground/module-resolution/__tests__/module-resolution.spec.ts
petebacondarwin: If you think it is worth it, I'm not against running these tests in tmp directories (at least in CI), as long as it is still easy enough to cd into a playground and get it up and running locally without jumping through too many hoops.

dario-piotrowicz: I was just mentioning it, I am not really sure if it is worth it... do you remember why we removed the tmp directory? was it causing issues?

If we don't re-introduce a tmp directory, are you ok with this testing code duplication? and/or have any ideas on how to avoid this?

threepointone (approved):

jamesopstad (approved):

packages/vite-plugin-cloudflare/src/node-js-compat.ts
dario-piotrowicz: No longer applicable since we decided to just use `isNodeCompat` instead of relying on the compat date and flag directly

packages/vite-plugin-cloudflare/playground/module-resolution/__tests__/module-resolution.spec.ts
dario-piotrowicz: > To be honest adding complexity where we copy certain files between directories when creating the tmp projects for testing seems to be quite magical and not that obvious. So I think that a manual sync is acceptable.
> 
> What about using a symlink between the files?

I did consider symlinks but I don't think that vitest would cope with those, even if it did I am worried about other OSes (that we test in CI), are symlinks something that other OSes can understand? (sorry if it's a silly question)

> Or just importing the test cases between playgrounds?

I did though of that too thinking that it wouldn't work, but let me actually give it a quick try 👍

dario-piotrowicz: oh! 😮 

Surpisingly (to me at least 😅) this actually works: https://github.com/cloudflare/workers-sdk/pull/8489/commits/0cef2999d85143231193135eaff16897b738b4ca

This is awesome! I'm really happy with this solution, WDTY? 🙂

petebacondarwin: Marvellous!

petebacondarwin (approved):

PR #8492, opened 2025-03-14
Use a shared Vectorize index for the Vectorize E2E tests

penalosa:
Wrangler has E2E tests that ensure all resources can be accessed from both a local and remote dev session. Typically, these perform some sort of round trip operation on a binding to validate that it's both available and working as expected. For instance, for KV this might include writing a key and then reading it.

Implicitly, this means we're assuming a write operation followed by a read operation will reflect the data from the first wrte operation. This holds up for all bindings (even in practice KV) _except Vectorize_, which is [intentionally asynchronous](https://developers.cloudflare.com/vectorize/reference/client-api/#insert-vectors) here. As such, our Vectorize E2E test waited for the results of the write operation to be available before trying to read. This is _really slow_, and means that the Vectorize E2E tests often timeout (past 90s).

This PR changes that model for Vectorize. Instead, we write and read from a shared resource across all E2E test runs, which means that the read operations are reading values from a prior E2E test run's write. This is slightly less valuable as a test, but for our purposes it's perfect—we don't really care about the semantics of the Vectorize service, we just care that Wrangler has hooked up all the plumbing correctly. Any vaguely reasonable response from Vectorize validates that.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal testing change

### DIFF
--- a/packages/wrangler/e2e/dev-with-resources.test.ts
+++ b/packages/wrangler/e2e/dev-with-resources.test.ts
@@ -518,8 +518,12 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 		}
 	});
 
-	it("exposes Vectorize bindings", async () => {
-		const name = await helper.vectorize(32, "euclidean");
+	it.only("exposes Vectorize bindings", async () => {
+		const name = await helper.vectorize(
+			32,
+			"euclidean",
+			"well-known-vectorize"
+		);
 
 		await helper.seed({
 			"wrangler.toml": dedent`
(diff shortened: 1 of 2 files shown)


packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -518,8 +518,12 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 		}
 	});
 
-	it("exposes Vectorize bindings", async () => {
-		const name = await helper.vectorize(32, "euclidean");
+	it.only("exposes Vectorize bindings", async () => {
```
emily-shen, line 521:
```suggestion
	it("exposes Vectorize bindings", async () => {
```

packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -572,10 +574,11 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 			`wrangler dev ${flags} --port ${port} --inspector-port ${inspectorPort} --experimental-vectorize-bind-to-prod`
 		);
 		const { url } = await worker.waitForReady();
-		const res = await fetch(url);
+		await fetch(`${url}/insert`);
+		const res = await fetch(`${url}/query`);
```
emily-shen, line 578: can you leave a comment here saying its reading from another run's query?

emily-shen, line 578: also how does this work on the first run? or is the idea it'll inevitably fail on the very first run and we just need to rerun?

penalosa: Yeah, it'll fail on the first run when a new insert is added, but that's very rare

emily-shen (approved):

PR #8507, opened 2025-03-14
[vite-plugin] fix: make sure that internal variables are not exposed in the importable `env` object

### LINKED ISSUE #8458: [vite-plugin] Apply `stripInternalEnv` to importable env
We need to apply the `stripInternalEnv` logic: https://github.com/cloudflare/workers-sdk/blob/869ec7b916487ec43b958a27bdfea13588c5685f/packages/vite-plugin-cloudflare/src/runner-worker/env.ts#L13-L23
 to the importable env object (ref: https://github.com/cloudflare/workerd/pull/3610)

Note: we don't necessarily want to call `stripInternalEnv` on the process.env object since that should not include any value besides the ones set by the user (as per: https://github.com/cloudflare/workers-sdk/issues/8457)

___

Internal ref: DEVX-1744

dario-piotrowicz:
Fixes #8458

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: change not related to wrangler
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 


packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
```diff
@@ -77,6 +78,18 @@ export async function createModuleRunner(
...
+					const originalCloudflareWorkersModule = await import(
+						"cloudflare:workers"
+					);
+					return {
+						...originalCloudflareWorkersModule,
+						env: stripInternalEnv(
+							originalCloudflareWorkersModule.env as WrapperEnv
+						),
+					};
+				}
```
jamesopstad, line 91: Do you think it matters that you're returning a normal object rather than a module namespace object here? Not sure if this could lead to behaviour differences.

dario-piotrowicz: no I don't think it matters, I can't really imagine how this could effect user's code 🤔 do you have something in mind? or a different way of implementing this?

Also even if there could be behavioural differences I can't imagine them being very broad in scope (i.e. I think this could at most introduce some minor bug that would only apply to dev when you access the module, which we could look into and solve when reported), so this feels to me pretty safe overall 🤔

jamesopstad: I'm happy with that. Alternative would be to use a virtual module but this is nicer because it keeps everything in the runner worker. How about using `Object.freeze()` so the object is still immutable?

```suggestion
					return Object.freeze({
						...originalCloudflareWorkersModule,
						env: stripInternalEnv(
							originalCloudflareWorkersModule.env as WrapperEnv
						),
					});
				}
```

dario-piotrowicz: > How about using Object.freeze() so the object is still immutable?

Sure 👍, but what about doing `Object.seal()` instead?
[mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object) states that the module namespace objects are _sealed_ not _frozen_

jamesopstad (approved):

packages/vite-plugin-cloudflare/playground/importable-env/__tests__/env-entries.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+
+test("the importable env is accessible from outside the request handler", async () => {
+	expect(serverLogs.info.join()).toMatch(
+		/outside of request handler: importedEnv\["importable-env_VAR"\] === "my importable env variable"/
+	);
+});
+
+test("the fetch handler env contains the correct entries", async () => {
+	const json = await getJsonResponse();
+	expect(json?.["entires of the fetch handler env"]).toEqual([
```
jamesopstad, line 12:
```suggestion
	expect(json?.["entries of the fetch handler env"]).toEqual([
```

packages/vite-plugin-cloudflare/playground/importable-env/src/index.ts
```diff
@@ -0,0 +1,27 @@
...
+	`outside of request handler: importedEnv["${key}"] === "${(importedEnv as Env)[key]}"`
+);
+
+export default {
+	async fetch(_req, argEnv) {
+		const entriesOfArgEnv = getEntriesOf(argEnv);
+		const entriesOfImportedEnv = getEntriesOf(importedEnv);
+
+		return Response.json({
+			"entires of the fetch handler env": entriesOfArgEnv,
```
jamesopstad, line 16:
```suggestion
			"entries of the fetch handler env": entriesOfArgEnv,
```

jamesopstad (commented):
I wouldn't worry about running the CI again for these typos so I don't mind if we merge and fix them later.

packages/vite-plugin-cloudflare/playground/importable-env/__tests__/env-entries.spec.ts
dario-piotrowicz: oops! thanks! 🙂

PR #8534, opened 2025-03-17
Improve the error messaging when the user provides neither an entry point nor an asset directory

petebacondarwin:
- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: minor message change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/8535
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -2548,7 +2548,7 @@ addEventListener('fetch', event => {});`
 				`
 				[Error: Missing entry-point to Worker script or to assets directory
 
-				If you want to deploy a Worker that includes code, you can either:
+				If there is code to deploy, you can either:
 				- Specify an entry-point to your Worker script via the command line (ex: \`npx wrangler deploy src/index.ts\`)
 				- Or add the following to your "wrangler.toml" file:
 
@@ -2558,7 +2558,7 @@ addEventListener('fetch', event => {});`
 				\`\`\`
 
 
-				If you want to deploy only a directory of assets (with no server-side code), you can either:
+				If are uploading a directory of assets, you can either:
 				- Specify the path to the directory of assets via the command line: (ex: \`npx wrangler deploy --assets=./dist\`)
 				- Or add the following to your "wrangler.toml" file:
 
@@ -2576,7 +2576,7 @@ addEventListener('fetch', event => {});`
 				"[31mX [41;31m[[41;97mERROR[41;31m][0m [1mMissing entry-point to Worker script or to assets directory[0m
 
 
-				  If you want to deploy a Worker that includes code, you can either:
+				  If there is code to deploy, you can either:
 				  - Specify an entry-point to your Worker script via the command line (ex: \`npx wrangler deploy
 				  src/index.ts\`)
 				  - Or add the following to your \\"wrangler.toml\\" file:
@@ -2587,7 +2587,7 @@ addEventListener('fetch', event => {});`
 				  \`\`\`
 
 
-				  If you want to deploy only a directory of assets (with no server-side code), you can either:
+				  If are uploading a directory of assets, you can either:
 				  - Specify the path to the directory of assets via the command line: (ex: \`npx wrangler deploy
 				  --assets=./dist\`)
 				  - Or add the following to your \\"wrangler.toml\\" file:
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2543,15 +2543,60 @@ addEventListener('fetch', event => {});`
...
-				`[Error: Missing entry-point: The entry-point should be specified via the command line (e.g. \`wrangler deploy path/to/script\`) or the \`main\` config field.]`
+				`
+				[Error: Missing Worker entry-point or assets directory
+
+				If this Worker has code:
+				- Specify an entry-point via the command line (e.g. \`wrangler deploy path/to/script\`)
+				- Or add the following to your "wrangler.toml" file:
+
+				\`\`\`
+				name = "path/to/script"
```
emily-shen, line 2554:
```suggestion
				main = "path/to/script"
```

penalosa (approved):

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2536,22 +2537,70 @@ addEventListener('fetch', event => {});`
...
 			mockUploadWorkerRequest();
 			mockSubDomainRequest();
 			await expect(
 				runWrangler("deploy")
 			).rejects.toThrowErrorMatchingInlineSnapshot(
-				`[Error: Missing entry-point: The entry-point should be specified via the command line (e.g. \`wrangler deploy path/to/script\`) or the \`main\` config field.]`
+				`
+				[Error: Missing entry-point to Worker script or to assets directory
+
+				If you want to deploy a Worker that includes code, you can either:
```
emily-shen, line 2551: "includes code" is a bit confusing, someone might have "code" as an asset right?

maybe switch it the other way around, since if you have a worker with assets you'd want both right?
```
If you want to deploy only a directory of assets (with no server-side code), you can either:
etc.

// clunky phrasing but hopefully you get the point...
If you want to deploy a Worker as well or instead:
etc.
```

dario-piotrowicz: or simply instead of "includes code" you could phrase it as "includes server-side logic"? or something to that extent? 🤔

packages/wrangler/src/deployment-bundle/entry.ts
```diff
@@ -70,8 +72,47 @@ export async function getEntry(
 					"For Pages, please run `wrangler pages dev` instead."
 			);
 		}
+
+		const compatibilityDateStr = [
+			new Date().getFullYear(),
+			(new Date().getMonth() + 1 + "").padStart(2, "0"),
+			(new Date().getDate() + "").padStart(2, "0"),
```
dario-piotrowicz, line 79: Nit: you could create the date just once without doing it three times

dario-piotrowicz: Also isn't using the current date as the compat date problematic (it generally shows a warning)? I thought that for that reason in C3 we use the lastest compat date that we fetch from npm? 🤔

packages/wrangler/src/__tests__/deploy.test.ts
petebacondarwin: Hmm. This was the copy suggested by @korinne.

I'll push a commit with a further suggestion...

packages/wrangler/src/deployment-bundle/entry.ts
petebacondarwin: Yeah I realise that this is potentially tricky. But I was getting bogged down in scope creap.
Originally this just updating the text, then it was fixing up the package manager, then adding in name and compat date 😓 . I don't want to start re-implementing C3!
At the end of the day, this is just messaging, we are not automatically updating the user's code.
If the compat date issue is a problem, we could just leave a placeholder there - something like "<latest-compat-date>" but I felt like that was even more confusing and unhelpful.

packages/wrangler/src/__tests__/deploy.test.ts
petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/8534/commits/d6bc4ed75326aa13c93c5e70b6ac69d9388c1fce

PR #8551, opened 2025-03-18
Remove non-published packages from the v3 branch

penalosa:
Remove packages from the v3-maintenance branch which shouldn't be published


emily-shen (commented):
i don't suppose we could delete the tests that use the --x-assets-rpc flag, since the feature wont be added to v3?

emily-shen:
is there any reason we're only deleting some packages, and not everything other than miniflare and wrangler

emily-shen (approved):
wonder if this counts to your total diff on this repo 😄

petebacondarwin (approved):
Let's go!

PR #8553, opened 2025-03-18
add react-router c3 template

emily-shen:
---

- Tests
  - [x] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: c3
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 


penalosa (approved):

packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -370,9 +370,15 @@ const verifyBuildCfTypesScript = async (
 	expect(outputFileContentPreLines).toContain("// Generated by Wrangler");
 
 	// the file contains the env interface
-	expect(outputFileContentPreLines).toContain(
-		`interface ${envInterfaceName} {`,
+	// the file still contains the env interface
+	const hasEnvInterfacePre = outputFileContentPreLines.some(
+		(line) =>
+			// old type gen - some frameworkte templates pin older versions of wrangler
```
jamesopstad, line 376:
```suggestion
			// old type gen - some framework templates pin older versions of wrangler
```

packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -389,14 +395,14 @@ const verifyBuildCfTypesScript = async (
 	expect(outputFileContentPostLines).not.toContain("// Generated by Wrangler");
 
 	// the file still contains the env interface
-	const hasEnvInterface = outputFileContentPostLines.some(
+	const hasEnvInterfacePost = outputFileContentPostLines.some(
 		(line) =>
 			// old type gen - some frameworkte templates pin older versions of wrangler
```
jamesopstad, line 400:
```suggestion
			// old type gen - some framework templates pin older versions of wrangler
```

packages/create-cloudflare/src/frameworks/package.json
```diff
@@ -14,6 +14,7 @@
 		"create-qwik": "1.12.0",
 		"create-vite": "6.1.1",
 		"create-remix": "2.15.3",
+		"create-react-router": "7.3.0",
```
jamesopstad, line 17:
```suggestion
		"create-react-router": "7.4.0",
```

packages/create-cloudflare/templates/react-router/c3.ts
```diff
@@ -0,0 +1,51 @@
...
+		doneText: `${brandColor(`updated`)} ${dim("wrangler@latest")}`,
+	});
+};
+
+const config: TemplateConfig = {
+	configVersion: 1,
+	id: "react-router",
+	platform: "workers",
+	frameworkCli: "create-react-router",
+	displayName: "React Router",
```
jamesopstad, line 34:
```suggestion
	displayName: "React Router (formerly Remix)",
```

There was some discussion about making it clear that this is the successor to Remix and removing the Remix entry. I think we also decided to have `--framework=remix` direct to the React Router template if specified via the CLI.

@petebacondarwin Can you confirm?

petebacondarwin: Yes I believe that was the decision. No interactive way to get Remix, but still supporting the `--framework` approach for backward compatibility of CI-type jobs.

emily-shen: I'm a bit confused - should `--framework=remix` direct to remix or react router? if it directs react-router, does this mean the remix template is removed, since it can neither be accessed through cli or interactivity?

emily-shen:
also noticed that react-router requires node 20+ but all our ci runs in node 18. however, only yarn c3 tests are failing because of the node version??

packages/create-cloudflare/templates/react-router/c3.ts
jamesopstad: Yep, that's my understanding. We remove the Remix template and make `remix` an alias to `react-router` when specified in the CLI.

petebacondarwin: Sorry if I created confusion here. I don't believe that `--framework=remix` should alias to React Router. The main reason for keeping that option is for backward compatibility with users who are, for instance, running C3 as part of a non-interactive automated flow. In this case they should expect to get a Remix project; and may well be relying upon that. Removing the Remix template (equivalent to redirecting to React-Router) is arguably a breaking change.

So my understanding of what we should do is:

- hide the remix option from interactive diaglogs
- leave `--framework=remix` as-is

emily-shen: Done here [49ba99b](https://github.com/cloudflare/workers-sdk/pull/8553/commits/49ba99bf7a79964aa7ca82587c745c46c883a1c4)

packages/create-cloudflare/templates/react-router/c3.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const { npm } = detectPackageManager();
+
+const generate = async (ctx: C3Context) => {
+	await runFrameworkGenerator(ctx, [
+		ctx.project.name,
+		"--template",
+		"https://github.com/remix-run/react-router-templates/tree/main/cloudflare",
+		// to prevent asking about git twice, just let c3 do it
+		"--no-git-init",
```
jamesopstad, line 15: Should we also set `--no-install` and let C3 handle that?

```suggestion
		"--no-git-init",
		"--no-install"
```

jamesopstad (approved):
Looks good. I've added one comment about installing dependencies. If you do make that change, it will need a manual test with the prerelease before merging. Thanks!

PR #8556, opened 2025-03-18
vite dev support for SPA mode

GregBrimble:
Fixes WC-3319.

Implements `vite dev` support for Workers Assets' new SPA mode.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: will follow up once everything lands
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 

### DIFF
--- a/packages/miniflare/src/workers/core/entry.worker.ts
+++ b/packages/miniflare/src/workers/core/entry.worker.ts
@@ -95,6 +95,14 @@ function getUserRequest(
 	// https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#accept-encoding
 	request.headers.set("Accept-Encoding", "br, gzip");
 
+	// `miniflare.dispatchFetch(request)` strips any `sec-fetch-mode` header. This allows clients to
+	// send it over a `x-mf-sec-fetch-mode` header instead (currently required by `vite preview`)
+	const secFetchMode = request.headers.get("X-Mf-Sec-Fetch-Mode");
+	if (secFetchMode) {
+		request.headers.set("Sec-Fetch-Mode", secFetchMode);
+	}
+	request.headers.delete("X-Mf-Sec-Fetch-Mode");
+
 	if (rewriteHeadersFromOriginalUrl) {
 		request.headers.set("Host", url.host);
 	}
--- a/packages/vite-plugin-cloudflare/playground/spa-with-api/__tests__/spa-with-api.spec.ts
+++ b/packages/vite-plugin-cloudflare/playground/spa-with-api/__tests__/spa-with-api.spec.ts
@@ -1,5 +1,7 @@
+import { createConnection } from "node:net";
 import { expect, test } from "vitest";
 import { page } from "../../__test-utils__";
+import { viteTestUrl } from "../../vitest-setup";
 
 test("returns the correct home page", async () => {
 	const content = await page.textContent("h1");
@@ -17,4 +19,43 @@ test("returns the response from the API", async () => {
 	await responsePromise;
 	const contentAfter = await button.innerText();
 	expect(contentAfter).toBe("Name from API is: Cloudflare");
+});
+
+test("returns the home page even for 404-y pages", async () => {
+	await page.goto(`${viteTestUrl}/foo`);
+	const content = await page.textContent("h1");
+	expect(content).toBe("Vite + React");
+});
+
+test("returns the home page even for API-y pages", async () => {
+	await page.goto(`${viteTestUrl}/api/`);
+	const content = await page.textContent("h1");
+	expect(content).toBe("Vite + React");
+});
+
+test("requests made with/without explicit 'sec-fetch-mode: navigate' header to delegate correctly", async () => {
+	const responseWithoutHeader = await fetch(`${viteTestUrl}/foo`);
+	expect(responseWithoutHeader.status).toBe(404);
+	expect(await responseWithoutHeader.text()).toEqual("nothing here");
+
+	// can't make `fetch`es with `sec-fetch-mode: navigate` header, so we're doing it raw
+	const { hostname, port } = new URL(viteTestUrl);
+	const socket = createConnection(parseInt(port), hostname, () => {
+		socket.write(
+			`GET /foo HTTP/1.1\r\nHost: ${hostname}\r\nSec-Fetch-Mode: navigate\r\n\r\n`
+		);
+	});
+
+	let responseWithoutHeaderContentBuffer = "";
+	socket.on("data", (data) => {
+		responseWithoutHeaderContentBuffer += data.toString();
+	});
+
+	const responseWithoutHeaderContent = await new Promise((resolve) =>
+		socket.on("close", () => {
+			resolve(responseWithoutHeaderContentBuffer);
+		})
+	);
+
+	expect(responseWithoutHeaderContent).toContain("Vite + React");
 });
--- a/packages/vite-plugin-cloudflare/src/asset-workers/asset-worker.ts
+++ b/packages/vite-plugin-cloudflare/src/asset-workers/asset-worker.ts
@@ -1,15 +1,14 @@
 // @ts-ignore
 import AssetWorker from "@cloudflare/workers-shared/dist/asset-worker.mjs";
 import { UNKNOWN_HOST } from "../shared";
-import type { WorkerEntrypoint } from "cloudflare:workers";
 
 interface Env {
 	__VITE_ASSET_EXISTS__: Fetcher;
 	__VITE_FETCH_ASSET__: Fetcher;
 }
 
-export default class CustomAssetWorker extends (AssetWorker as typeof WorkerEntrypoint<Env>) {
-	override async fetch(request: Request): Promise<Response> {
+export default class CustomAssetWorker extends AssetWorker {
+	async fetch(request: Request): Promise<Response> {
 		const response = await super.fetch!(request);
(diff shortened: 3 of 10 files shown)


packages/vite-plugin-cloudflare/playground/spa-with-api/__tests__/spa-with-api.spec.ts
```diff
@@ -18,3 +19,27 @@ test("returns the response from the API", async () => {
 	const contentAfter = await button.innerText();
 	expect(contentAfter).toBe("Name from API is: Cloudflare");
 });
+
+test("returns the home page even for 404-y pages", async () => {
+	page.goto("/foo");
```
jamesopstad, line 24: These tests are failing at the moment because the navigations need full URLs and to be awaited:

```suggestion
	await page.goto(`${viteTestUrl}/foo`);
```

packages/vite-plugin-cloudflare/playground/spa-with-api/__tests__/spa-with-api.spec.ts
```diff
@@ -18,3 +19,27 @@ test("returns the response from the API", async () => {
...
+	const content = await page.textContent("h1");
+	expect(content).toBe("Vite + React");
+});
+
+test("returns the home page even for API-y pages", async () => {
+	page.goto("/api/");
```
jamesopstad, line 30:
```suggestion
	await page.goto(`${viteTestUrl}/api/`);
```

jamesopstad (commented):
Will have a proper look tomorrow. If you can get all the tests passing first that would great. Thanks!

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -536,6 +543,133 @@ export function getPreviewMiniflareOptions(
 		unstable_readConfig({ config: configPath })
 	);
 
+	const entryWorkerConfig = workerConfigs[0];
+
+	const assetWorkers: Array<WorkerOptions> =
```
GregBrimble, line 548: I am really, _really_ not a fan of how this works in Vite.

GregBrimble:
This second commit is garbage. Vite plugin people, please help me. Why are we relying on Vite's magic so much here? It doesn't represent reality at all. Is there a better way to do this?

packages/vite-plugin-cloudflare/src/utils.ts
```diff
@@ -16,6 +16,14 @@ export function getOutputDirectory(
 }
 
 export function toMiniflareRequest(request: Request): MiniflareRequest {
+	// something about how we dispatch this request strips the 'sec-fetch-mode: navigate' and replaces it with 'sec-fetch-mode: cors'
+	// current theory is that it is hattip or undici as it converts it for transport
+	// so, because this isn't a real security use-case,
+	// we can just set some arbitrary other header and convert that back into 'sec-fetch-mode: navigate' in our worker
```
jamesopstad, line 22: I've been looking at what's going on here. It's not `hattip`. If you log the request inside the `createMiddleware` function then you see that it contains `sec-fetch-mode: navigate`.

https://github.com/cloudflare/workers-sdk/blob/9609b9fd8a344257007233c3b8788f9e196a856d/packages/vite-plugin-cloudflare/src/index.ts#L271-L278

So it seems it's undici in Miniflare that is alters this. I found this issue, which is related - https://github.com/nodejs/undici/issues/1305.
There is also this comment in the linked PR, which doesn't seem very hopeful - https://github.com/nodejs/undici/pull/1829#issuecomment-1789885535.
And there's this comment in a Miniflare issue - https://github.com/cloudflare/miniflare/issues/139#issuecomment-1003755936.

Curious why this problem presents itself here but not in Wrangler or when using Miniflare directly. Any ideas?

petebacondarwin: The header is a bit of a red-herring, I think.

What I believe is happening is when one instantates a new Request object using another request as the `init` param in order to copy over its fields, the `mode` of this object is set to `cors`. The headers are copied as expected.

So if you look at the request it has the correct `navigate` header but the mode is now `cors`.

It is only then when Undici processes this request as part of a fetch it overwrites the Sec-Fetch-Mode header with the `mode` from the request.

jamesopstad: Right. I don't think you're permitted to explicitly set `mode: 'navigate'`. What do you think the solution is here? I guess we could add a special header as Greg has but then move the logic that uses this to set the `sec-fetch-mode` header into Miniflare. This doesn't seem like a Vite specific problem to solve.

petebacondarwin: The confusing this is that we are not seeing this in Wrangler. I think this maybe because the Proxy Worker (used in Wrangler) is written as a Worker and doesn't use Undici requests, so perhaps is not quite following the same logic?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
GregBrimble: Without it, `vite preview` was just always delegating 404s to the user worker, but I've now realized that this is just because of the same issue we saw in dev! I can remove all this now :)

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -95,6 +95,14 @@ function getUserRequest(
...
 	request.headers.set("Accept-Encoding", "br, gzip");
 
+	// `miniflare.dispatchFetch(request)` strips any `sec-fetch-mode` header. This allows clients to
+	// send it over a `x-mf-sec-fetch-mode` header instead (currently required by `vite preview`)
+	const secFetchMode = request.headers.get("X-Mf-Sec-Fetch-Mode");
+	if (secFetchMode) {
+		request.headers.set("Sec-Fetch-Mode", secFetchMode);
+	}
+	request.headers.delete("X-Mf-Sec-Fetch-Mode");
+
```
jamesopstad, line 105: Thanks for moving this here. This is great.

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -34,6 +34,7 @@ export interface AssetsOnlyConfig extends SanitizedWorkerConfig {
 	topLevelName: Defined<SanitizedWorkerConfig["topLevelName"]>;
 	name: Defined<SanitizedWorkerConfig["name"]>;
 	compatibility_date: Defined<SanitizedWorkerConfig["compatibility_date"]>;
+	compatibility_flags: Defined<SanitizedWorkerConfig["compatibility_flags"]>;
```
jamesopstad, line 37: Don't think this will be necessary as it's always defined in the returned config.

```suggestion
```

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -242,6 +242,12 @@ export function getDevMiniflareOptions(
 			],
 			bindings: {
 				CONFIG: {
+					...(entryWorkerConfig?.compatibility_date
+						? { compatibility_date: entryWorkerConfig?.compatibility_date }
```
jamesopstad, line 246:
```suggestion
						? { compatibility_date: entryWorkerConfig.compatibility_date }
```

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -242,6 +242,12 @@ export function getDevMiniflareOptions(
...
+						: {}),
+					...(entryWorkerConfig?.compatibility_flags
+						? { compatibility_flags: entryWorkerConfig.compatibility_flags }
+						: {}),
```
jamesopstad, line 250: These will only be added if there is a User Worker. I think that's correct because this behaviour is not relevant with assets only but just wanted to check.

packages/vite-plugin-cloudflare/src/utils.ts
```diff
@@ -22,6 +22,14 @@ export function getRouterWorker(miniflare: Miniflare) {
...
```
jamesopstad, line 28:
```suggestion
// Undici sets the `Sec-Fetch-Mode` header to `cors` so we capture it in a custom header to be converted back later.
```

jamesopstad (approved):
Looks great! Just added a few minor comments. Thanks a lot Greg.

packages/vite-plugin-cloudflare/src/asset-workers/asset-worker.ts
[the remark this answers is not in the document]
GregBrimble: `super.unstable_canFetch` needs me to not be casting `AssetWorker as WorkerEntrypoint<Env>`. Which then means I do need to be casking `this.env as typeof AssetWorker as { env: Env }` everywhere. I can remove the contentType optionality change though.

Totally agree on the exports. I took a stab at it ages ago but didn't land it. I'll try again in the near future, I promise.

packages/vite-plugin-cloudflare/src/miniflare-options.ts
GregBrimble: Nope, you're totally right. I've picked up these options from config now whether or not there is an user Worker.

PR #8557, opened 2025-03-18
Collapse C3 E2E jobs

penalosa:
Collapse C3 E2E jobs into a single job for both experimental & non-experimental variants. This should reduce CI contention by reducing the number of runners needed, and reduce CI overall time by removing the pnpm installation time for duplicate runners.

### DIFF
--- a/.github/workflows/c3-e2e.yml
+++ b/.github/workflows/c3-e2e.yml
@@ -29,7 +29,7 @@ jobs:
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}-${{ matrix.experimental }}
       cancel-in-progress: true
-    name: ${{ format('Run tests for {0}@{1} on {2} (experimental:{3})', matrix.pm.name, matrix.pm.version, matrix.os, matrix.experimental) }}
+    name: ${{ format('Run tests for {0}@{1} on {2}', matrix.pm.name, matrix.pm.version, matrix.os, matrix.experimental) }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/c3-e2e.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}-${{ matrix.experimental }}
       cancel-in-progress: true
-    name: ${{ format('Run tests for {0}@{1} on {2} (experimental:{3})', matrix.pm.name, matrix.pm.version, matrix.os, matrix.experimental) }}
+    name: ${{ format('Run tests for {0}@{1} on {2}', matrix.pm.name, matrix.pm.version, matrix.os, matrix.experimental) }}
```
vicb, line 32: remove 3 (`matrix.experimental`)?

vicb (approved):
LGTM, 1 nit

dario-piotrowicz (approved):
This looks good to me, it reduces parallelization though, since you're making the change I imagine you're confident that the tradeoff is worth it, I'm happy with that and we can always de-collapse the jobs later if this tradeoff is actually not worth it 👍

dario-piotrowicz:
> > it reduces parallelization though
> 
> We can always revert this if it turns out to be worse, but what I've been seeing is that C3 E2E jobs are queued on waiting for other C3 E2E jobs on the same PR to finish, which means that we're not getting many parallelisation benefits today anyway

Ah I see 😕, yeah I commented on this because generally the more parallization the better right? 😅 but yeah it's not this black and white with gh workflows unfortunately 😅

PR #8561, opened 2025-03-18
[do-not-merge] feat: add an `event-subscriptions` subcommand

sdnts:
Fixes #MQ-845

Adds an `event-subscriptions` subcommand that will be used to manage Event Subscriptions

CR-1140468

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 
  - [x] Not necessary because: this is a new feature

### DIFF
--- a/packages/wrangler/src/__tests__/event-subscriptions.test.ts
+++ b/packages/wrangler/src/__tests__/event-subscriptions.test.ts
@@ -1,0 +1,857 @@
+import assert from "node:assert";
+import { http, HttpResponse } from "msw";
+import * as products from "../event-subscriptions/products";
+import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
+import { mockConsoleMethods } from "./helpers/mock-console";
+import { clearDialogs } from "./helpers/mock-dialogs";
+import { useMockIsTTY } from "./helpers/mock-istty";
+import { msw } from "./helpers/msw";
+import { runInTempDir } from "./helpers/run-in-tmp";
+import { runWrangler } from "./helpers/run-wrangler";
+
+const testAccountId = "some-account-id";
+
+describe("wrangler event-subscriptions", () => {
+	mockAccountId({ accountId: testAccountId });
+	mockApiToken();
+	runInTempDir();
+
+	const std = mockConsoleMethods();
+
+	const { setIsTTY } = useMockIsTTY();
+	beforeEach(() => {
+		setIsTTY(true);
+	});
+
+	afterEach(() => {
+		msw.restoreHandlers();
+		vi.restoreAllMocks();
+		clearDialogs();
+	});
+
+	describe("event-subscription browse", async () => {
+		test("shows all defined product events", async () => {
+			msw.use(
+				http.get(
+					"*/accounts/:accountId/event_subscriptions/events",
+					async ({ params }) => {
+						expect(params.accountId).toEqual(testAccountId);
+						return HttpResponse.json({
+							success: true,
+							errors: [],
+							result: [
+								{
+									id: "slurper",
+									name: "Super Slurper",
+									events: [{ name: "e1" }, { name: "e2" }],
+								},
+								{
+									id: "workflows",
+									name: "Workflows",
+									events: [{ name: "e10" }, { name: "e20" }],
+								},
+							],
+						});
+					},
+					{ once: true }
+				)
+			);
+
+			const result = runWrangler("event-subscriptions browse");
+
+			await expect(result).resolves.toBeUndefined();
+			expect(std.out).toMatchInlineSnapshot(`
+				"┌───────────────┬──────────┐
+				│ service       │ events   │
+				├───────────────┼──────────┤
+				│ Super Slurper │ e1, e2   │
+				├───────────────┼──────────┤
+				│ Workflows     │ e10, e20 │
+				└───────────────┴──────────┘"
+			`);
+		});
+
+		test("ignores products that do not exist on both QBW and Wrangler", async () => {
+			msw.use(
+				http.get(
+					"*/accounts/:accountId/event_subscriptions/events",
+					async ({ params }) => {
+						expect(params.accountId).toEqual(testAccountId);
+						return HttpResponse.json({
+							success: true,
+							errors: [],
+							result: [
+								{
+									id: "workflows",
+									name: "Workflows",
+									events: [{ name: "e10" }, { name: "e20" }],
+								},
+							],
+						});
+					},
+					{ once: true }
+				)
+			);
+
+			const result = runWrangler("event-subscriptions browse");
+
+			await expect(result).resolves.toBeUndefined();
+			expect(std.out).toMatchInlineSnapshot(`
+				"┌───────────┬──────────┐
+				│ service   │ events   │
+				├───────────┼──────────┤
+				│ Workflows │ e10, e20 │
+				└───────────┴──────────┘"
+			`);
+		});
+
+		test("can filter product ids with --source", async () => {
+			msw.use(
+				http.get(
+					"*/accounts/:accountId/event_subscriptions/events",
+					async ({ params }) => {
+						expect(params.accountId).toEqual(testAccountId);
+						return HttpResponse.json({
+							success: true,
+							errors: [],
+							result: [
+								{ id: "slurper", name: "Super Slurper", events: [] },
+								{ id: "workflows", name: "Workflows", events: [] },
+							],
+						});
+					},
+					{ once: true }
+				)
+			);
+
+			const result = runWrangler("event-subscriptions browse --source=slurp");
+
+			await expect(result).resolves.toBeUndefined();
+			expect(std.out).toMatchInlineSnapshot(`
+				"┌───────────────┬────────┐
+				│ service       │ events │
+				├───────────────┼────────┤
+				│ Super Slurper │        │
+				└───────────────┴────────┘"
+			`);
+		});
+
+		test("can filter multiple product ids with --source", async () => {
+			msw.use(
+				http.get(
+					"*/accounts/:accountId/event_subscriptions/events",
+					async ({ params }) => {
+						expect(params.accountId).toEqual(testAccountId);
+						return HttpResponse.json({
+							success: true,
+							errors: [],
+							result: [
+								{ id: "slurper", name: "Super Slurper", events: [] },
+								{ id: "workflows", name: "Workflows", events: [] },
+								{ id: "workersAI", name: "Workers AI", events: [] },
+							],
+						});
+					},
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/event-subscriptions/products.ts
```diff
@@ -0,0 +1,74 @@
+import type { HandlerContext } from "../core/types";
+
+// Use the same `id` here as the one you used in queue-broker-worker
+export const PRODUCTS: ProductSpec[] = [
```
sdnts, line 4: So the idea is that CF products "onboard" on to the event hub by defining the events they're going to emit. They do so by sending a PR to the Queues codebase. However, there are UI-like things like a resource name that doesn't make sense to put inside Queues. I've got a similar array on the dashboard that basically lets us keep user-facing output here directly.

More importantly though, to create a subscription, Wrangler (and the dashboard) are supposed to send a request to the _product_, and so every product needs to tell Wrangler how to make this request.

This is why I maintain this static `PRODUCTS` list here.

packages/wrangler/src/event-subscriptions/products.ts
```diff
@@ -0,0 +1,74 @@
...
+	{
+		id: "clipboards",
```
sdnts, line 6: This is an entirely made-up product that I've been using to test things out before real teams onboard. I will remove this once the first team onboards -- we do not want to expose this to users because it is not real.

packages/wrangler/src/event-subscriptions/products.ts
```diff
@@ -0,0 +1,74 @@
...
+					}
+				),
+		},
+	},
+	{
+		id: "workersAI",
+		name: "Workers AI",
+		resource: null,
+		subscription: {
+			create: () => Promise.resolve({ id: "" }),
```
sdnts, line 33: These will be populated by teams. I'll remove entries for teams who do not onboard right now.

sdnts:
Do not merge yet

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+type EventSubscriptionEventSpec = {
+	id: string;
+	events: Array<{ name: string; description: string }>;
+};
+
+export const eventSubscriptionsNamespace = createNamespace({
+	metadata: {
+		owner: "Product: Queues",
+		status: "open-beta",
+		description: "Manage Event Subscriptions",
```
sdnts, line 29: @maheshwarip could use help with these descriptions, I'm not satisfied with what I came up with.

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+	events: string[];
+};
+
+type EventSubscriptionEventSpec = {
+	id: string;
+	events: Array<{ name: string; description: string }>;
+};
+
+export const eventSubscriptionsNamespace = createNamespace({
+	metadata: {
```
emily-shen, line 26: just add `hidden: true` to metadata

packages/wrangler/src/event-subscriptions/index.ts
maheshwarip: IMO it's ok - this looks like the format we're following for all the wrangler product descriptions?

packages/wrangler/src/event-subscriptions/index.ts
CarmenPopoviciu: how come? is this meant to be a hidden command?

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+	queue_name: string;
+	events: string[];
+};
+
+type EventSubscriptionEventSpec = {
+	id: string;
+	events: Array<{ name: string; description: string }>;
+};
+
+export const eventSubscriptionsNamespace = createNamespace({
```
CarmenPopoviciu, line 25: should this command name be post-fixed with `Command` as well, just like all other commands?
```suggestion
export const eventSubscriptionsNamespaceCommand = createNamespace({
```

CarmenPopoviciu: oh...nvm. Totally missed you're calling `createNamespace`. Ignore the comment

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+			demandOption: true,
+		},
+	},
+	validateArgs({ source }) {
+		const [productId, resourceId] = source.split(".");
+
+		const product = PRODUCTS.find((p) => p.id === productId);
+		if (!product) {
+			throw new CommandLineArgsError(
+				`Invalid source. TODO: Write a better / more helpful error message`
```
CarmenPopoviciu, line 197: [ ] TODO: Write a better / more helpful error message

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+			);
+		}
+
+		if (product.resource) {
+			if (!resourceId) {
+				throw new CommandLineArgsError(
+					`Invalid resource ID. TODO: Write a better / more helpful error message`
```
CarmenPopoviciu, line 204: [ ] TODO: Write a better / more helpful error message

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+		const accountId = await requireAuth(config);
+
+		const { name, source, destination: queueId, events } = args;
+
+		const [productId, resourceId] = source.split(".");
+		const product = PRODUCTS.find((p) => p.id === productId);
+
+		// Disabling ESLint and using `!` because we've validated the argument already
+		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+		const res = await product!.subscription.create(ctx, {
```
CarmenPopoviciu, line 221: > using `!` because we've validated the argument already

What guarantees that `PRODUCTS.find((p) => p.id === productId)` will always return a non-null result?

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+			demandOption: true,
+		},
+	},
+
+	async handler(args, ctx) {
+		const config = readConfig(args);
+		const accountId = await requireAuth(config);
+		const { id } = args;
+
+		const subscription = await ctx.fetchResult<EventSubscription>(
```
CarmenPopoviciu, line 144: what does the API return if no subscription with `id` is found? should that case be handled by the client here?

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+		},
+	},
+
+	async handler(args, ctx) {
+		const config = readConfig(args);
+		const accountId = await requireAuth(config);
+
+		const { id, name, events } = args;
+
+		const subscription = await ctx.fetchResult<EventSubscription>(
```
CarmenPopoviciu, line 271: what will the API return if subscription with `id` does not exist? Should this case be handled here?

packages/wrangler/src/event-subscriptions/index.ts
```diff
@@ -0,0 +1,323 @@
...
+		},
+	},
+
+	async handler(args, ctx) {
+		const config = readConfig(args);
+		const accountId = await requireAuth(config);
+
+		const { id } = args;
+
+		await ctx.fetchResult<EventSubscription[]>(
```
CarmenPopoviciu, line 316: same question regarding trying to delete a subscription that does not exist. How should that be handled?

packages/wrangler/src/event-subscriptions/products.ts
CarmenPopoviciu: can you please add a comment explaining what the purpose of `PRODUCTS` is :pray:. This will help folks outside of your team get some more context <3

packages/wrangler/src/__tests__/event-subscriptions.test.ts
```diff
@@ -0,0 +1,334 @@
...
+import { clearDialogs } from "./helpers/mock-dialogs";
+import { useMockIsTTY } from "./helpers/mock-istty";
+import { msw } from "./helpers/msw";
+import { runInTempDir } from "./helpers/run-in-tmp";
+import { runWrangler } from "./helpers/run-wrangler";
+import type { ProductSpec } from "../event-subscriptions/products";
+
+const testAccountId = "some-account-id";
+
+describe("wrangler", () => {
```
CarmenPopoviciu, line 13:
```suggestion
describe("wrangler event-subscriptions", () => {
```

packages/wrangler/src/__tests__/event-subscriptions.test.ts
```diff
@@ -0,0 +1,334 @@
...
+				{ once: true }
+			)
+		);
+
+		const result = runWrangler("event-subscriptions list");
+		await expect(result).resolves.toBeUndefined();
+		expect(std.out).toMatchSnapshot();
+	});
+
+	test("event-subscriptions get", async () => {
```
CarmenPopoviciu, line 244: would be great if `get`/`update`/`delete` tests would test what happens when the provided `id` does not exist

packages/wrangler/src/event-subscriptions/index.ts
sdnts: Ah this is not needed anymore. I had a thread going a week ago when we wanted to do this.

packages/wrangler/src/event-subscriptions/index.ts
sdnts: So up in the `validateArgs` function, I do the same lookup and throw an error if the product wasn't found. This code path would be unreachable if the product did not exist.

packages/wrangler/src/event-subscriptions/index.ts
sdnts: The API would return a 404 with an error message. I looked up what the `fetchResult` function was doing, and it looks like it handles that already?
But I think I'll  add a test for this case, why not

packages/wrangler/src/event-subscriptions/index.ts
sdnts: Same as above, will add a test!

sdnts:
This has changed significantly, I'll open up another PR

PR #8587, opened 2025-03-20
Add Wasm helper

jamesopstad:
Fixes #000.

This was just a bit of fun while I was waiting for CI. It adds support for `.wasm?init` imports, as documented by Vite (https://vite.dev/guide/features.html#webassembly). Not sure if it's something we should actually include or not so happy to close if you feel it's unnecessary.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: coming soon
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -341,6 +341,26 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 				});
 			},
 		},
+		// Plugin to support `.wasm?init` extension
+		{
+			name: "vite-plugin:wasm-helper",
+			enforce: "pre",
+			applyToEnvironment(environment) {
+				return getWorkerConfig(environment.name) !== undefined;
+			},
+			load(id) {
+				if (!id.endsWith(".wasm?init")) {
+					return;
+				}
+
+				return `
+					import wasm from "${cleanUrl(id)}";
+					export default function(opts = {}) {
+						return WebAssembly.instantiate(wasm, opts);
+					}
+				`;
+			},
+		},
 		// Plugin to support additional modules
 		{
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -316,6 +316,26 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+			applyToEnvironment(environment) {
+				return getWorkerConfig(environment.name) !== undefined;
+			},
+			load(id) {
+				if (!id.endsWith(".wasm?init")) {
+					return;
+				}
+
+				return `
+					import wasm from "${cleanUrl(id)}";
```
penalosa, line 332: I assume `cleanUrl()` here makes sure this isn't circular?

penalosa (approved):
Looks like a fun addition to me!

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: It trims off the query parameter so that the import is then resolved with the `.wasm` extension.

petebacondarwin (approved):
This is cool. Does the loader here ensure that the built-in Vite processing doesn't happen?

jamesopstad:
> This is cool. Does the loader here ensure that the built-in Vite processing doesn't happen?

Yep. I followed the example from built in plugin (https://github.com/vitejs/vite/blob/9827df2195905e5eb04b46dce357d12c3dff4876/packages/vite/src/node/plugins/wasm.ts#L49-L80) and this is inserted before it.

PR #8597, opened 2025-03-20
feat: Graduate experimental RPC support for Workers with assets in local dev

CarmenPopoviciu:
Fixes DEVX-1670

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: we are not calling this out in particular in our docs
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -63,10 +63,7 @@ import {
 	WorkerOptions,
 	WrappedBindingNames,
 } from "./plugins";
-import {
-	ROUTER_SERVICE_NAME,
-	RPC_PROXY_SERVICE_NAME,
-} from "./plugins/assets/constants";
+import { RPC_PROXY_SERVICE_NAME } from "./plugins/assets/constants";
 import {
 	CUSTOM_SERVICE_KNOWN_OUTBOUND,
 	CustomServiceKind,
@@ -1248,8 +1245,7 @@ export class Miniflare {
 							/*
 							 * If we are running multiple Workers in a single dev session,
 							 * and this is a binding to a Worker with assets, we want that
-							 * binding to point to the Router Worker or the Assets Proxy
-							 * Worker, if the `unsafeEnableAssetsRpc` flag is enabled
+							 * binding to point to the (assets) RPC Proxy Worker
 							 */
 							const maybeAssetTargetService = allWorkerOpts.find(
 								(worker) =>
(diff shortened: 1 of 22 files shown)


packages/miniflare/src/index.ts
```diff
@@ -1374,9 +1365,8 @@ export class Miniflare {
 			sharedOptions: sharedOpts.core,
 			allWorkerRoutes,
 			/*
-			 * - if Workers + Assets project but NOT Vitest, point to Router Worker or
-			 *   Proxy Worker depending on whether experimental flag `unsafeEnableAssetsRpc`
-			 *   was set
+			 * - if Workers + Assets project but NOT Vitest, point to the (assets) RPC Proxy
+			 *   Worker
 			 * - if Vitest with assets, the self binding on the test runner will point to
```
GregBrimble, line 1370: Should the self binding point to the RPC proxy?

CarmenPopoviciu: I'm confused abt that comment as well...I need to check with Emily which is the source of truth there...the comment or the code

GregBrimble (approved):

packages/miniflare/src/index.ts
GregBrimble: Specifically, this other comment seems contradictory: https://github.com/cloudflare/workers-sdk/pull/8597/files#diff-f46f54e70e03a1c9ddf8abc506eb10b5ff24424bc806f8c2c245aaaa07543239R272. But I haven't poked at the internals closely, so there might be a totally legit reason why its set in one place but not the other.

CarmenPopoviciu: to be clear, this was the original [comment](https://github.com/cloudflare/workers-sdk/pull/8288/files#diff-9ceb732889eaca4ea99209c2d4988f82a6ed60a853000e856d47fdcb687d67f2L1313-L1314), and this was the original [code](https://github.com/cloudflare/workers-sdk/pull/8288/files#diff-9ceb732889eaca4ea99209c2d4988f82a6ed60a853000e856d47fdcb687d67f2L1320). So this is pre our rpc work

CarmenPopoviciu: spoke with Emily, and turns out the code was correct, and the comment just a tad misleading. I updated the comment to correct that, and posting here a diagram for posterity, that highlights the two use cases 
Screenshot 2025-03-21 at 11 53 30

GregBrimble: Makes sense! Ship it!

penalosa (approved):

PR #8612, opened 2025-03-20
chore(fixtures): Add fixture tests for Workers + Assets and WebSockets

GregBrimble:
Fixes n/a

Adds fixture tests for Workers + Assets and WebSockets

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: adds fixture tests
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: adds fixture tests
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: adds fixture tests

### DIFF
--- a/fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
+++ b/fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
@@ -7,20 +7,22 @@ import { runWranglerDev } from "../../shared/src/run-wrangler-long-lived";
 describe("[Workers + Assets] Service bindings to Worker with assets", () => {
 	let ipWorkerA: string,
 		portWorkerA: number,
-		stopWorkerA: (() => Promise<unknown>) | undefined;
+		stopWorkerA: (() => Promise<unknown>) | undefined,
+		getOutputWorkerA: () => string;
 	let stopWorkerB: (() => Promise<unknown>) | undefined,
 		getOutputWorkerB: () => string;
 	let stopWorkerC: (() => Promise<unknown>) | undefined,
 		getOutputWorkerC: () => string;
 	let stopWorkerD: (() => Promise<unknown>) | undefined;
+	let stopWorkerWS: (() => Promise<unknown>) | undefined;
 
 	beforeAll(async () => {
-		({ getOutput: getOutputWorkerB } = await runWranglerDev(
+		({ stop: stopWorkerB, getOutput: getOutputWorkerB } = await runWranglerDev(
 			resolve(__dirname, "..", "workerB-with-default-export"),
 			["--port=0", "--inspector-port=0"]
 		));
 
-		({ getOutput: getOutputWorkerC } = await runWranglerDev(
+		({ stop: stopWorkerC, getOutput: getOutputWorkerC } = await runWranglerDev(
 			resolve(__dirname, "..", "workerC-with-default-entrypoint"),
 			["--port=0", "--inspector-port=0"]
 		));
@@ -30,10 +32,16 @@ describe("[Workers + Assets] Service bindings to Worker with assets", () => {
 			["--port=0", "--inspector-port=0"]
 		));
 
+		({ stop: stopWorkerWS } = await runWranglerDev(
+			resolve(__dirname, "..", "workerWS"),
+			["--port=0", "--inspector-port=0"]
+		));
+
 		({
 			ip: ipWorkerA,
 			port: portWorkerA,
 			stop: stopWorkerA,
+			getOutput: getOutputWorkerA,
 		} = await runWranglerDev(resolve(__dirname, "..", "workerA"), [
 			"--port=0",
 			"--inspector-port=0",
@@ -45,13 +53,11 @@ describe("[Workers + Assets] Service bindings to Worker with assets", () => {
 		await stopWorkerB?.();
 		await stopWorkerC?.();
 		await stopWorkerD?.();
+		await stopWorkerWS?.();
 	});
 
 	describe("Workers running in separate wrangler dev sessions", () => {
 		describe("Service binding to default export", () => {
-			// this currently incorrectly returns the User Worker response
-			// instead of the Asset Worker response, unless `--x-assets-rpc`
-			// is provided
 			it("should return Asset Worker response for routes that serve static content", async ({
 				expect,
 			}) => {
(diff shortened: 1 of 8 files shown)


CarmenPopoviciu:
I can take this over Greg <3

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
```diff
@@ -0,0 +1,20 @@
+import { WorkerEntrypoint } from "cloudflare:workers";
+
+export default class extends WorkerEntrypoint {
```
CarmenPopoviciu, line 3: something in this Worker's logic throws 👇 

```
✘ [ERROR] Uncaught (in response) Error: The script will never generate a response.
```

but for the life in me I'm not sure what Im doing wrong. We don't seem to be doing anything different than https://developers.cloudflare.com/workers/examples/websockets/, but I dunno. It doesn't break tests or kill the `wrangler dev` session, but would be nice to have stdout error free ¯\_(ツ)_/¯

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
```diff
@@ -45,13 +53,11 @@ describe("[Workers + Assets] Service bindings to Worker with assets", () => {
 		await stopWorkerB?.();
 		await stopWorkerC?.();
 		await stopWorkerD?.();
+		await stopWorkerWS?.();
 	});
 
 	describe("Workers running in separate wrangler dev sessions", () => {
 		describe("Service binding to default export", () => {
-			// this currently incorrectly returns the User Worker response
```
CarmenPopoviciu, line 52: this is an outdated comment

fixtures/workers-with-assets-and-service-bindings/tests/index.test.ts
```diff
@@ -156,9 +161,6 @@ describe("[Workers + Assets] Service bindings to Worker with assets", () => {
 		});
 
 		describe("Service binding to default entrypoint", () => {
-			// this currently incorrectly returns the User Worker response
```
CarmenPopoviciu, line 159: this is an outdated comment

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+	fetch(req: Request) {
+		if (!req.headers.get("Upgrade")?.includes("websocket")) {
+			return new Response("Hello from worker-ws fetch()");
+		}
+
+		const [server, client] = Object.values(new WebSocketPair());
```
vicb, line 9:
```ts
  const client = webSocketPair[0],
    server = webSocketPair[1];
```

That's what you are doing differently ;)

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
GregBrimble: I've never seen that happen before! Not at all sure what could be causing that... it's very clearly returning a response in all situations!

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
CarmenPopoviciu: the docs example will work towards what I have in my code

```
The WebSocketPair constructor returns an Object, with the 0 and 1 keys each
holding a WebSocket instance as its value. It is common to grab the two WebSockets 
from this pair using Object.values ↗ and ES6 destructuring ↗, as seen in the below 
example.
```

GregBrimble (commented):
Can't approve my own PR, but yup, LGTM!

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+		server.accept();
```
petebacondarwin, line 11: Should this go after you have setup the event listener?

petebacondarwin: I guess it doesn't matter since the response hasn't been returned to the client yet.

petebacondarwin: Is there a problem with this worker dying before the websocket ping-pong completes?
Do we need a `waitUntil` here?

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+		const textEncoder = new TextEncoder();
+		server.addEventListener("message", (event) => {
+			server.send(textEncoder.encode(`pong: ${event.data}`));
```
petebacondarwin, line 15: should we then call `server.close()` here?

petebacondarwin (approved):

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
CarmenPopoviciu: I wondered about that as well. I'm basically following the steps in https://developers.cloudflare.com/workers/examples/websockets/, which does not mention closing the server. Perhaps in our case it makes more sense since we know the client won't send any other messages after the initial one. Happy to add it

CarmenPopoviciu: added in https://github.com/cloudflare/workers-sdk/pull/8612/commits/c2a8365a88aaa16b8861957bc6d7cd0952cf0e70

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
CarmenPopoviciu: looks like adding `server.close()` solved this issue. Err is gone

fixtures/workers-with-assets-and-service-bindings/workerWS/src/index.ts
CarmenPopoviciu: no idea tbh. is this something we need to worry about in the context of running the fixture tests? Feels like not, but maybe I'm wrong

PR #8616, opened 2025-03-21
chore: define wrapper for cloudflare vite-plugin for the playground apps

dario-piotrowicz:
Quick idea I had so that we don't have to remember to always set `inspectorPort` and `persistState` to `false` when configuring a playground app

This could also be nice for future proofing, in care we introduce other properties that should be defaulted to something in all playground apps

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: is a testing utility change
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


jamesopstad:
I'm not super keen on this because it changes the way the plugin is resolved. Now it comes from a parent package rather than the individual playground. It also makes the playgrounds less useful as references.

PR #8617, opened 2025-03-21
chore: app option to run the vite-plugin playground serve tests in parallel

dario-piotrowicz:
I think that the only vite-plugin playground tests needing to run sequentially are the build
ones (because of `.wrangler/deploy/config.json`) so I think we can just as well run the dev ones in parallel
to save time
(PS: my only small concern is prebundling since the `node_modules/.vite` directory is still shared across different
variants, but I don't think we ever encountered any issue with that)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: minor chore change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not wrangler related
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: minor chore change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/playground/package.json
+++ b/packages/vite-plugin-cloudflare/playground/package.json
@@ -5,10 +5,13 @@
 	"type": "module",
 	"scripts": {
 		"check:types": "tsc --build",
-		"test:build": "cross-env VITE_TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
-		"pretest:ci": "pnpm playwright install chromium",
-		"test:ci": "pnpm test:serve && pnpm test:build",
-		"test:serve": "vitest run -c vitest.config.e2e.ts"
+		"playwright:install": "pnpm playwright install chromium",
+		"pretest:ci": "pnpm playwright:install",
+		"test:ci": "pnpm test:ci:serve && pnpm test:ci:build",
+		"test:ci:build": "cross-env VITE_TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
+		"test:ci:serve": "vitest run -c vitest.config.e2e.ts",
+		"pretest:serve": "pnpm playwright:install",
+		"test:serve": "pnpm test:ci:serve --poolOptions.forks.singleFork=false"
 	},
 	"devDependencies": {
 		"@cloudflare/vite-plugin": "workspace:*",


dario-piotrowicz:
oof... by looking at the CI this seems not to work well in the prisma app on windows 😓

dario-piotrowicz:
I'm skipping the prisma test in windows then, I think this is an ok tradeoff here since we still run those for the other OSes anyways, is someone disagree I'm happy to drop the effort here, or of course consider a different possible solution if anyone has any idea 🙂

packages/vite-plugin-cloudflare/playground/prisma/__tests__/prisma.spec.ts
```diff
@@ -9,7 +9,11 @@ import { getJsonResponse } from "../../__test-utils__";
 //
 // We do this in the `preServe()` hook, in `serve.ts`, that is called from the `packages/vite-plugin-cloudflare/playground/vitest-setup.ts` file.
 
-test("runs D1 query using Prisma", async () => {
+test.skipIf(
+	// for some reason this seems to be problematic on windows
+	// when running the various playground tests in parallel
+	process.platform === "win32"
+)("runs D1 query using Prisma", async () => {
```
petebacondarwin, line 16: What if we just run the tests in sequence in Windows, rather than skipping this test?

dario-piotrowicz: ah sure, that didn't occur to me! 🙂👍

dario-piotrowicz: updated, thanks for the suggestion :slightly_smiling_face:

jamesopstad:
Not sure how this will play out with https://github.com/cloudflare/workers-sdk/pull/8652 and reusing the same Miniflare instance so let's rebase once that PR is merged and see if there are any issues.

jamesopstad:
I think we should put this PR on hold for the time being as it may make it harder to diagnose flakiness in the tests.

dario-piotrowicz:
> I think we should put this PR on hold for the time being as it may make it harder to diagnose flakiness in the tests.

I am not really sure if there is indeed a flakiness issue with the tests 🤔 

I'm checking them out here: https://github.com/cloudflare/workers-sdk/pull/8699 and they don't really seem flaky 😕 

I am experiencing flakiness here: https://github.com/cloudflare/workers-sdk/pull/8685 but either way I need to resolve that before merging the PR anyways

Have you seen flakiness in the tests before? (I did see some 8685 that seemed unrelated to the PR, but given the lack of flakiness in 8699 I am having second thoughts)

dario-piotrowicz:
PS: if there are concerns regarding CI, I'd also be ok to have the tests run sequentially in CI, I'm proposing this change mainly because the sequential run during local development is unnecessarily slow

petebacondarwin (changes requested):
You can get the same effect by adding `--poolOptions.forks.singleFork=false` to the command line. So I would propose we don't actually update the vitest.config.ts files here.
Instead how about we update the package.json scripts to be:

```
	"scripts": {
		"check:types": "tsc --build",
		"test:ci:build": "cross-env VITE_TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
		"pretest:ci": "pnpm playwright install chromium",
		"test:ci": "pnpm test:ci:serve && pnpm test:ci:build",
		"test:ci:serve": "vitest run -c vitest.config.e2e.ts",
		"test:serve": "pnpm test:ci:serve --poolOptions.forks.singleFork=false",
	},
```

Then you can cd into the playground and run `pnpm test:serve`

packages/vite-plugin-cloudflare/playground/package.json
```diff
@@ -5,10 +5,11 @@
...
 	"scripts": {
 		"check:types": "tsc --build",
-		"test:build": "cross-env VITE_TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
 		"pretest:ci": "pnpm playwright install chromium",
-		"test:ci": "pnpm test:serve && pnpm test:build",
-		"test:serve": "vitest run -c vitest.config.e2e.ts"
+		"test:ci": "pnpm test:ci:serve && pnpm test:ci:build",
+		"test:ci:build": "cross-env VITE_TEST_BUILD=1 vitest run -c vitest.config.e2e.ts",
+		"test:ci:serve": "vitest run -c vitest.config.e2e.ts",
+		"test:serve": "pnpm test:ci:serve --poolOptions.forks.singleFork=false"
```
petebacondarwin, line 12: Do we need a `pretest:serve` too?

petebacondarwin (approved):

packages/vite-plugin-cloudflare/playground/package.json
dario-piotrowicz: done 🙂

PR #8623, opened 2025-03-21
feat: add bulk gets for kv in miniflare

teresalves:
Fixes KV-1435
Adding new feature for bulk gets in miniflare.
It has already been added in workerd.

We are injecting a timestamp when we read/write our keys. So `key1` becomes `1742848988250_122630884213439key1`. This never influenced our tests because we never got the key back, and since the keys always matched in the write and read, it didn't matter. Now we had to clean that up when returning it back to the test result.

CR-1140001

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: We are just adding a new component in miniflare.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: The docs will be done separately.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: adds new feature


penalosa:
@teresalves this will need to rebase on main to pull in the latest workerd

packages/miniflare/test/test-shared/miniflare.ts
```diff
@@ -38,7 +38,23 @@ export function namespace<T>(ns: string, binding: T): Namespaced<T> {
...
 					if (Array.isArray(keys)) keys = keys.map((key) => ns + key);
-					return (value as (...args: unknown[]) => unknown)(keys, ...args);
+					const result = (value as (...args: unknown[]) => unknown)(
+						keys,
+						...args
+					);
+					if (result instanceof Promise) {
+						return result.then((res) => {
+							if (res instanceof Map) {
+								const newResult = new Map<string, unknown>();
```
radev, line 48: I think we should put a comment here explaining what it going on:
```
// KV.get([a,b,c]) would be prefixed with ns, so we strip this prefix from response.
// Map keys => [ns-a, ns-b, ns-c] -> [a,b,c]
```

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +75,41 @@ function secondsToMillis(seconds: number): number {
 	return seconds * 1000;
 }
 
+async function processKeyValue(
+	obj: KeyValueEntry<unknown> | null,
+	type: string = "text",
+	withMetadata: boolean = false
+) {
+	const decoder = new TextDecoder();
+	let r = "";
```
pmiguel, line 84: A bit nitpicky, but use a better name.

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -81,13 +118,44 @@ export class KVNamespaceObject extends MiniflareDurableObject {
 	}
 
 	@GET("/:key")
+	@POST("/bulk/get")
 	get: RouteHandler<KVParams> = async (req, params, url) => {
+		if (req.method === "POST" && req.body != null) {
+			let r = "";
```
pmiguel, line 124: Same as above.

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +75,41 @@ function secondsToMillis(seconds: number): number {
...
+			"At least of of the requested keys corresponds to a non-JSON value"
+		);
+	}
+	if (val == null) {
+		return null;
+	}
+	if (withMetadata) {
+		return {
+			value: val,
+			metadata: obj?.metadata ? JSON.stringify(obj?.metadata) : null,
```
ferhatelmas, line 107: metadata isn't string but json, and we need a metadata test

CarmenPopoviciu:
could we please get a KV approval on this PR 🙏 ?

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +76,46 @@ function secondsToMillis(seconds: number): number {
...
+		val =
+			obj?.value == null
+				? null
+				: type === "json"
+					? JSON.parse(decodedValue)
+					: decodedValue;
+	} catch (err: any) {
+		throw new HttpError(
+			400,
+			"At least of of the requested keys corresponds to a non-JSON value"
```
CarmenPopoviciu, line 104:
```suggestion
			"At least one of the requested keys corresponds to a non-JSON value"
```
?

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -81,13 +124,44 @@ export class KVNamespaceObject extends MiniflareDurableObject {
...
+			const decoder = new TextDecoder();
+			for await (const chunk of req.body) {
+				decodedBody += decoder.decode(chunk, { stream: true });
+			}
+			decodedBody += decoder.decode();
+			const parsedBody = JSON.parse(decodedBody);
+			const keys: string[] = parsedBody.keys;
+			const type = parsedBody?.type;
+			if (type && type !== "text" && type !== "json") {
+				return new Response("", { status: 400 });
```
CarmenPopoviciu, line 140: we sure abt returning an empty response here?

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -81,13 +124,44 @@ export class KVNamespaceObject extends MiniflareDurableObject {
...
+			}
+			const obj: { [key: string]: any } = {};
+			if (keys.length > MAX_BULK_GET_KEYS) {
+				return new Response("", { status: 400 });
```
CarmenPopoviciu, line 144: same empty resp question here

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +76,46 @@ function secondsToMillis(seconds: number): number {
 	return seconds * 1000;
 }
 
+async function processKeyValue(
+	obj: KeyValueEntry<unknown> | null,
+	type: string = "text",
```
CarmenPopoviciu, line 81: what are the possible values for this type? Can this really be any string? If not perhaps a union type would be a better fit?

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +76,46 @@ function secondsToMillis(seconds: number): number {
...
+	withMetadata: boolean = false
```
CarmenPopoviciu, line 82:
```suggestion
	withMetadata = false
```

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -20,6 +22,7 @@ import {
 	validatePutOptions,
 } from "./validator.worker";
 
+const MAX_BULK_GET_KEYS = 100;
```
ferhatelmas, line 25: I was meaning more sharing it with wrangler commands since they should be same and changed at the same time

packages/miniflare/src/workers/kv/namespace.worker.ts
ferhatelmas: text | json

packages/miniflare/src/workers/kv/namespace.worker.ts
ferhatelmas: similar type could be `type: "text" | "json" = "text"`

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -20,6 +22,7 @@ import {
...
 interface KVParams {
```
ferhatelmas, line 26: not exactly related to this PR but why do we redefine this?

.changeset/funny-eggs-divide.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": minor
+---
+
+Add bulk gets for Workers KV
```
pmiguel, line 5:
```suggestion
Add Miniflare Workers KV bulk get support
```

packages/miniflare/src/workers/kv/namespace.worker.ts
pmiguel: Command doesn't specify the max.

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +76,46 @@ function secondsToMillis(seconds: number): number {
...
+		for await (const chunk of obj?.value) {
+			decodedValue += decoder.decode(chunk, { stream: true });
+		}
+		decodedValue += decoder.decode();
+	}
+
+	let val = null;
+	try {
+		val =
+			obj?.value == null
```
CarmenPopoviciu, line 96: this [equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality) check can return an unintended result, since it attempts to convert and compare operands that are of different types. Did you mean to use a [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality) check perhaps `===`? Or maybe a [nullish coalescing assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment) `??=`?

packages/miniflare/src/workers/kv/namespace.worker.ts
CarmenPopoviciu: having read through the code once, I'm wondering if this should be smth like `type: "text" | "json"` 🤔

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +76,46 @@ function secondsToMillis(seconds: number): number {
...
+				: type === "json"
+					? JSON.parse(decodedValue)
+					: decodedValue;
+	} catch (err: any) {
+		throw new HttpError(
+			400,
+			"At least of of the requested keys corresponds to a non-JSON value"
+		);
+	}
+	if (val == null) {
```
CarmenPopoviciu, line 107: same [comment](https://github.com/cloudflare/workers-sdk/pull/8623/files#r2016037909) here

packages/miniflare/src/workers/kv/namespace.worker.ts
ferhatelmas: delegates to API?

maintaining a constant not ideal but could save a round trip

packages/miniflare/test/test-shared/miniflare.ts
```diff
@@ -38,7 +38,25 @@ export function namespace<T>(ns: string, binding: T): Namespaced<T> {
...
+						...args
+					);
+					if (result instanceof Promise) {
+						return result.then((res) => {
+							// KV.get([a,b,c]) would be prefixed with ns, so we strip this prefix from response.
+							// Map keys => [ns-a, ns-b, ns-c] -> [a,b,c]
+							if (res instanceof Map) {
+								const newResult = new Map<string, unknown>();
+								for (const [key, value] of res) {
+									newResult.set(key.slice(ns.length), value);
```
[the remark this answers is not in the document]
ferhatelmas: `-` is from misleading doc, actually key is `{ns}{key}`

packages/miniflare/test/plugins/kv/index.spec.ts
```diff
@@ -122,6 +122,107 @@ test("get: returns value", async (t) => {
 	const result = await kv.get("key");
 	t.is(result, "value");
 });
+
+test("bulk get: returns value", async (t) => {
```
CarmenPopoviciu, line 126: tests yay!!! ♥️

.changeset/funny-eggs-divide.md
ferhatelmas: not necessary I guess since it will be release notes of miniflare package

packages/miniflare/test/test-shared/miniflare.ts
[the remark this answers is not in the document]
CarmenPopoviciu: oh I see..we should update the comment then to reflect that

packages/miniflare/test/plugins/kv/index.spec.ts
```diff
@@ -122,6 +122,107 @@ test("get: returns value", async (t) => {
...
+	]);
+
+	t.deepEqual(result, expectedResult);
+});
+
+test("bulk get: check max keys", async (t) => {
+	const { kv } = t.context;
+	await kv.put("key1", "value1");
+	const keyArray = [];
+	for (let i = 0; i <= 100; i++) {
```
ferhatelmas, line 142: can use the constant now

.changeset/funny-eggs-divide.md
CarmenPopoviciu: I think it's a valid suggestion

rts-rob (approved):
Approving as is; we will do a fast follow to tighten up some documentation.

packages/miniflare/src/workers/kv/namespace.worker.ts
teresalves: We are not using this text in workerd tbh (yes we should). I updated this anyway, for future use, but the tests validate that the error thrown on the binding side is still `KV GET_BULK failed: 400 Bad Request`

.changeset/funny-eggs-divide.md
teresalves: yeah this file is specifically for the miniflare package

packages/miniflare/src/workers/kv/namespace.worker.ts
teresalves: No idea honestly, I was checking it now and it is in fact weird. Should we create a ticket to review this?
I also wanted to create one to decouple our things from R2 so maybe we can just have a "cleaning up miniflare" ticket

ferhatelmas: sounds good to make a pass afterwards to clean up but also note that this will be dead code eventually with JSRPC so I wouldn't rush

packages/miniflare/src/workers/kv/namespace.worker.ts
teresalves: It can be undefined so i opted to do the comparison with (!obj?.value) and it works for our scenarios

ferhatelmas (approved):

packages/miniflare/src/workers/kv/namespace.worker.ts
```diff
@@ -73,6 +75,45 @@ function secondsToMillis(seconds: number): number {
...
+			: type === "json"
+				? JSON.parse(decodedValue)
+				: decodedValue;
+	} catch (err: any) {
+		throw new HttpError(
+			400,
+			"At least one of the requested keys corresponds to a non-JSON value"
+		);
+	}
+	if (val === null) {
```
CarmenPopoviciu, line 105: this is a nit but you can basically rewrite
```
if (val === null) {
	return null;
}
if (withMetadata) {
	return {
		value: val,
		metadata: obj?.metadata ?? null,
	};
}

return val;
```

to 

```
if (val && withMetadata) {
	return {
		value: val,
		metadata: obj?.metadata ?? null,
	};
}

return val;
```

CarmenPopoviciu (approved):
great work on this 🚀

packages/miniflare/src/workers/kv/namespace.worker.ts
teresalves: Thanks for the suggestion! That does look cleaner, added it!

PR #8625, opened 2025-03-21
Provide more informative error messages for Workers CI tag validation

WalshyDev:
Fixes CUSTESC-49792

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal only change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Workers CI internal usage change

### DIFF
--- a/packages/wrangler/src/environment-variables/misc-variables.ts
+++ b/packages/wrangler/src/environment-variables/misc-variables.ts
@@ -93,18 +93,18 @@ export const getOutputFilePathFromEnv = getEnvironmentVariableFactory({
 });
 
 /**
- * `WRANGLER_CI_MATCH_TAG` specifies a worker tag
+ * `WRANGLER_CI_MATCH_TAG` specifies a Worker tag
  *
- * If this is set, Wrangler will ensure the worker being targeted has this tag
+ * If this is set, Wrangler will ensure the Worker being targeted has this tag
  */
 export const getCIMatchTag = getEnvironmentVariableFactory({
 	variableName: "WRANGLER_CI_MATCH_TAG",
 });
 
 /**
- * `WRANGLER_CI_OVERRIDE_TAG` specifies a worker tag
+ * `WRANGLER_CI_OVERRIDE_NAME` specifies a Worker name
  *
- * If this is set, Wrangler will override the worker name with this tag
+ * If this is set, Wrangler will override the Worker name with this one
  */
 export const getCIOverrideName = getEnvironmentVariableFactory({
 	variableName: "WRANGLER_CI_OVERRIDE_NAME",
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -102,9 +102,9 @@ export const getCIMatchTag = getEnvironmentVariableFactory({
 });
 
 /**
- * `WRANGLER_CI_OVERRIDE_TAG` specifies a worker tag
+ * `WRANGLER_CI_OVERRIDE_NAME` specifies a worker name
```
CarmenPopoviciu, line 105:
```suggestion
 * `WRANGLER_CI_OVERRIDE_NAME` specifies a Worker name
```

packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -102,9 +102,9 @@ export const getCIMatchTag = getEnvironmentVariableFactory({
...
  *
- * If this is set, Wrangler will override the worker name with this tag
+ * If this is set, Wrangler will override the worker name with this one
```
CarmenPopoviciu, line 107:
```suggestion
 * If this is set, Wrangler will override the Worker name with this one
```

packages/wrangler/src/match-tag.ts
```diff
@@ -48,9 +49,17 @@ export async function verifyWorkerMatchesCITag(
 			throw new FatalError(
 				`The name in your ${configFileName(configPath)} file (${workerName}) must match the name of your Worker. Please update the name field in your ${configFileName(configPath)} file.`
 			);
+		} else if (e instanceof APIError) {
+			throw new FatalError(
+				"An error occurred while trying to verify the Worker.\n" +
```
CarmenPopoviciu, line 54: `while trying to verify the Worker` what does verify mean in this context? Will users understand what this is about?

WalshyDev: Yeah good point, I thought I was copying an existing error but now I don't know where I saw it haha

Updating the wording to match the other one which is clearer

packages/wrangler/src/match-tag.ts
```diff
@@ -48,9 +49,17 @@ export async function verifyWorkerMatchesCITag(
 			throw new FatalError(
 				`The name in your ${configFileName(configPath)} file (${workerName}) must match the name of your Worker. Please update the name field in your ${configFileName(configPath)} file.`
 			);
+		} else if (e instanceof APIError) {
+			throw new FatalError(
+				"An error occurred while trying to validate that the Worker name matches what is expected by the build system.\n" +
+					e.message +
```
CarmenPopoviciu, line 55: will this `message` contain enough information for users to know what they need to action? Not sure we ever improved these err messages we get, but they used to be rather cryptic. If that's still the case, would be good to suggest an action on the client side the

CarmenPopoviciu (approved):
❤️ thx Walshy

PR #8601, opened 2025-03-20
chore: add missing `inspectorPort: false` in vite-plugin playground app

dario-piotrowicz:
In all the vite-plugin playground apps we need to set the `inspectorPort` option to `false` otherwise the CI doesn't seem to properly cope, one (from https://github.com/cloudflare/workers-sdk/pull/8489) is missing (simply because https://github.com/cloudflare/workers-sdk/pull/8441 got merged when out of date with `main`) causing the following issue:
https://github.com/cloudflare/workers-sdk/actions/runs/13974580123/job/39124949648?pr=8596

This PR addresses that

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: ci chore
- Wrangler E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not wrangler related
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: ci chore
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a Wrangler change

PR #8627, opened 2025-03-21
Turn off inspector port in importable-env fixture

penalosa:
Duplicate of https://github.com/cloudflare/workers-sdk/pull/8601 for a new fixture


penalosa:
For posterity—merging without CI passing because CI is currently broken on main due to this issue, and is incredibly backed up. I've verified this locally.

PR #8639, opened 2025-03-22
Implement support for ctx.props.

kentonv:
In `wrangler.toml` you can declare a service binding with props:

```
[[services]]
binding = "MY_SERVICE"
service = "some-worker"
props = { foo = 123, bar = 456 }
```

Unfortunately, it currently doesn't work when wrangler is running different workers in different processes, as `props` cannot be sent to a remote `workerd`. Fixing this would require `workerd` changes, but since we are moving away from the multi-process mode anyway, perhaps it's not worth it.

Note: I used Claude Code to find my way around the codebase. It wrote most of the PR. It was not able to figure out that changes were needed under `packages/miniflare`, and I myself had a pretty hard time figuring out where to edit.

### DIFF
--- a/packages/miniflare/src/plugins/core/index.ts
+++ b/packages/miniflare/src/plugins/core/index.ts
@@ -251,6 +251,7 @@ function getCustomServiceDesignator(
 ): ServiceDesignator {
 	let serviceName: string;
 	let entrypoint: string | undefined;
+	let props: any;
 	if (typeof service === "function") {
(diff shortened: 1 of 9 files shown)


kentonv:
Apparently I deleted a bunch of boilerplate from the PR description without reading the explicit instructions that I should not delete them.

Fine I will close and re-open this PR with the boilerplate.

kentonv:
See #8640

Issue #8558, opened 2025-03-18
🚀 Feature Request: Support `ctx.props`


penalosa:
### Describe the solution

Support `ctx.props` on service bindings

PR #8640, opened 2025-03-22
Implement support for ctx.props.

kentonv:
Fixes https://github.com/cloudflare/workers-sdk/issues/8558

In `wrangler.toml` you can declare a service binding with props:

```
[[services]]
binding = "MY_SERVICE"
service = "some-worker"
props = { foo = 123, bar = 456 }
```

Unfortunately, it currently doesn't work when wrangler is running different workers in different processes, as `props` cannot be sent to a remote `workerd`. Fixing this would require `workerd` changes, but since we are moving away from the multi-process mode anyway, perhaps it's not worth it.

Note: I used Claude Code to find my way around the codebase. It wrote most of the PR. It was not able to figure out that changes were needed under `packages/miniflare`, and I myself had a pretty hard time figuring out where to edit.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: ctx.props is a cross-cutting feature and will be documented when it is done
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: It's a new feature, people should upgrade to wrangler 4 if they want to use it.

### DIFF
--- a/.changeset/curly-groups-scream.md
+++ b/.changeset/curly-groups-scream.md
@@ -1,0 +1,32 @@
+---
+"miniflare": minor
+"wrangler": minor
+---
+
+Add support for defining `props` on a Service binding.
+
+In your configuration file, you can define a service binding with props:
+
+```json
+{
+	"services": [
+		{
+			"binding": "MY_SERVICE",
+			"service": "some-worker",
+			"props": { "foo": 123, "bar": "value" }
+		}
+	]
+}
+```
+
+These can then be accessed by the callee:
+
+```ts
+import { WorkerEntrypoint } from "cloudflare:workers";
+
+export default class extends WorkerEntrypoint {
+	fetch() {
+		return new Response(JSON.stringify(this.ctx.props));
+	}
+}
+```
--- a/packages/miniflare/src/plugins/core/services.ts
+++ b/packages/miniflare/src/plugins/core/services.ts
@@ -90,6 +90,7 @@ export const ServiceDesignatorSchema = z.union([
 	z.object({
 		name: z.union([z.string(), z.literal(kCurrentWorker)]),
 		entrypoint: z.ostring(),
+		props: z.record(z.unknown()).optional(),
 	}),
 	z.object({ network: NetworkSchema }),
 	z.object({ external: ExternalServerSchema }),
--- a/packages/wrangler/e2e/multiworker-dev.test.ts
+++ b/packages/wrangler/e2e/multiworker-dev.test.ts
@@ -68,6 +68,10 @@ describe("multiworker", () => {
 							await counter.increment(3)
 							return new Response(String(await counter.value))
                         }
+                        if (url.pathname === "/props") {
+                            const props = await env.COUNTER.getProps()
+                            return new Response(JSON.stringify(props))
+                        }
 						return env.BEE.fetch(req);
 					},
 				};
@@ -128,6 +132,9 @@ describe("multiworker", () => {
 				export class CounterService extends WorkerEntrypoint {
 					async newCounter() {
 						return new Counter();
+					}
+					async getProps() {
+						return this.ctx.props;
 					}
 				}
 				export default{
@@ -193,6 +200,7 @@ describe("multiworker", () => {
 						binding = "COUNTER"
 						service = '${workerName2}'
 						entrypoint = 'CounterService'
+						props = { foo = 123, bar = { baz = "hello from props" } }
 				`,
 			});
 		});
@@ -228,6 +236,26 @@ describe("multiworker", () => {
 
 			await vi.waitFor(
 				async () => await expect(fetchText(`${url}/count`)).resolves.toBe("6"),
+				{ interval: 1000, timeout: 10_000 }
+			);
+		});
+
+		it("can access service props through a binding", async () => {
+			const workerA = helper.runLongLived(
+				`wrangler dev -c wrangler.toml -c ${b}/wrangler.toml`,
+				{ cwd: a }
+			);
+			const { url } = await workerA.waitForReady(5_000);
+
+			await vi.waitFor(
+				async () => {
+					const response = await fetch(`${url}/props`);
+					const props = await response.json();
+					expect(props).toEqual({
+						foo: 123,
+						bar: { baz: "hello from props" },
+					});
+				},
 				{ interval: 1000, timeout: 10_000 }
 			);
 		});
--- a/packages/wrangler/src/dev/miniflare.ts
+++ b/packages/wrangler/src/dev/miniflare.ts
@@ -443,6 +443,7 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 			serviceBindings[service.binding] = {
 				name: service.service,
 				entrypoint: service.entrypoint,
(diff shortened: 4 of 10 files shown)


kentonv:
It's not immediately obvious to me what this "changeset" thing is about (the links don't help). Can someone just push whatever is needed into this branch? Thanks.

kentonv:
I didn't answer the Wrangler v3 backport question because I have no idea what it's about.

GregBrimble:
I'd say the back port is not necessary because this is a new feature. The back port is just about our maintenance commitment to Wrangler v3 and copying over any bug fixes.

Changesets are the way we auto-generate release notes and bump versions of the packages here in workers-sdk. You can run `npx changeset` in the root of the repo and it will guide you through creating one. I've drafted one for you: https://github.com/cloudflare/workers-sdk/pull/8640/commits/0078cdc3c47e73300ca2ed33b546fbac051afde3.

GregBrimble (approved):

packages/miniflare/src/plugins/core/services.ts
```diff
@@ -84,12 +84,17 @@ export const ServiceFetchSchema = z.custom<
 	(request: Request, mf: Miniflare) => Awaitable<Response>
 >((v) => typeof v === "function");
 
+export const PropsSchema = z.object({
+	json: z.string(),
+});
```
penalosa, line 89: Is this schema correct? It seems to be saying that props are a string—should they not be an object?

packages/wrangler/e2e/multiworker-dev.test.ts
```diff
@@ -232,6 +248,33 @@ describe("multiworker", () => {
 			);
 		});
 
+		it("can access service props through a binding", async () => {
+			// TODO: We currently test `ctx.props` support with the worker calling
+			// back to itself, not to another worker, because the multi-process
+			// dev environment doesn't support ctx.props, so only self-bindings
+			// actually work. When we move to the single-process dev approach,
+			// `ctx.props` should then just work cross-worker, and we should update
+			// this test to test that.
```
penalosa, line 257: I don't think this is correct. This test file is for the single process dev approach, and so multi-worker props should work

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -442,6 +442,9 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 			serviceBindings[service.binding] = {
 				name: service.service,
 				entrypoint: service.entrypoint,
+				props: service.props
+					? { json: JSON.stringify(service.props) }
+					: undefined,
```
penalosa, line 447: I don't think this is the right place to do this. IMO props should be passed as an object to Miniflare and then stringified before they're sent to workerd

packages/miniflare/src/plugins/core/services.ts
kentonv: It's an object that has been JSON-encoded, so it's a string.

This schema matches what the workerd config wants. But I suppose this layer doesn't necessarily have to match workerd config. We could keep it as an arbitrary object at this layer, and JSON-encode it only when building the final workerd config at a lower layer.

I don't really have an opinion on this, but the current approach seems to work fine.

packages/wrangler/e2e/multiworker-dev.test.ts
kentonv: Ah, perhaps I got confused because I initially had a different bug that prevented props from getting through to workerd config, and I thought I was hitting this problem. I'll try changing it back to cross-worker.

.changeset/curly-groups-scream.md
```diff
@@ -0,0 +1,27 @@
...
+Add support for defining `props` on a Service binding.
+
+In your configuration file, you can define a service binding with props:
+
+```toml
+[[services]]
+binding = "MY_SERVICE"
+service = "some-worker"
+props = { foo = 123, bar = "value" }
+```
```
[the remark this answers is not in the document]
kentonv: I am not familiar with that format, could you suggest an edit?

kentonv:
Finally got around to updating this:
* Updated multiworker-dev.test.ts to actually test this using a cross-worker binding, it does in fact work.
* Pushed the JSON stringification step deeper into miniflare.

kentonv:
(Rebased to `main`.)

kentonv:
Ping?

.changeset/curly-groups-scream.md
penalosa, line 15:
```suggestion
```json
{
  "services": [{
    "binding": "MY_SERVICE",
    "service": "some-worker",
    "props": { "foo": 123, "bar": "value" }
  }]
}
```
```

File: packages/miniflare/src/plugins/core/services.ts [hunk 5]
```diff
@@ -90,6 +90,7 @@ export const ServiceDesignatorSchema = z.union([
 	z.object({
 		name: z.union([z.string(), z.literal(kCurrentWorker)]),
 		entrypoint: z.ostring(),
+		props: z.record(z.any()).optional(),
```
penalosa, line 93: Nit, but this might be better:
```suggestion
		props: z.record(z.unknown()).optional(),
```

penalosa (approved):

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -506,6 +507,9 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 				}
 			}
 
+			// BUG: We have no way to pass `props` across an external socket, so we
+			// drop them. We are planning to move away from the multi-process model
```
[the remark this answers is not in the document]
kentonv: The original way that `wrangler dev` supported service bindings is you would run separate instances of `wrangler dev` in each service's directory, and they would magically find each other and talk to each other. But this isn't the way workerd was designed to be used.

The new single-process model, which Samuel has implemented and can explain in more detail, arranges to pull all of your services into a single workerd instance with direct service bindings between them defined in its config. As I understand it it's still considered experimental, but we should really be pushing people over to this model as it is how workerd was intended to be used. The multi-process model has led to friction with JSRPC, ctx.props, etc.

kentonv:
Changes since code review:

https://github.com/cloudflare/workers-sdk/compare/0164a235a4bae9c3ff252575a2e56758298e11e8..c2b521afd13da6d516df1186900947db4def1d21

kentonv:
What does this mean?

Screenshot from 2025-04-22 09-56-21

kentonv:
Re-running the E2E mac test which timed out. I see it's not required -- how hard should I try to make it work?

petebacondarwin:
> Re-running the E2E mac test which timed out. I see it's not required -- how hard should I try to make it work?

Looks like it might pass this time; if not the vite ones are green and cached, and the Wrangler ones are incrementally cached, so will not take long to re-run if any fail. I can guide it home.

PR #8666, opened 2025-03-25
Remove `NodeJSCompatModule`

penalosa:
Fixes n/a

This will soon be removed in `workerd`, and was never fully supported for deployed Workers. As such, it's being entirely removed.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: This passes the existing tests (with some modifications), which should be sufficient
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: never exposed to users
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/8683
  - [ ] Not necessary because:

### DIFF
--- a/packages/miniflare/test/plugins/core/modules.spec.ts
+++ b/packages/miniflare/test/plugins/core/modules.spec.ts
@@ -20,7 +20,7 @@ test("Miniflare: accepts manually defined modules", async (t) => {
 	// Check with just `path`
 	const mf = new Miniflare({
 		compatibilityDate: "2023-08-01",
-		compatibilityFlags: ["nodejs_compat"],
+		compatibilityFlags: ["nodejs_compat_v2"],
 		// TODO(soon): remove `modulesRoot` once https://github.com/cloudflare/workerd/issues/1101 fixed
 		//  and add separate test for that
 		modulesRoot: ROOT,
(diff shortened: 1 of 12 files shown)


CarmenPopoviciu (approved):

CarmenPopoviciu:
backport PR isn't created due to conflicts, so we'll need to create manually

emily-shen (approved):

packages/miniflare/test/plugins/core/modules.spec.ts
```diff
@@ -207,13 +207,13 @@ test("Miniflare: cannot automatically collect modules from dynamic import expres
 		modulesRoot: ROOT,
 		modulesRules: [
 			// Implicitly testing default module rules for `ESModule` and `CommonJS`
```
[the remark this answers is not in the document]
CarmenPopoviciu: thx for catching this

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -372,11 +372,8 @@ function buildProjectWorkerOptions(
...
 	);
 
-	if (mode !== "v1" && mode !== "v2") {
-		runnerWorker.compatibilityFlags.push(
-			"nodejs_compat",
-			"no_nodejs_compat_v2"
-		);
+	if (mode !== "v2") {
+		runnerWorker.compatibilityFlags.push("nodejs_compat_v2");
 	}
```
dario-piotrowicz, line 377: Could someone clarify this logic here? why are we adding the flag if not present?

CarmenPopoviciu: maybe @petebacondarwin ? This code path is very unfamiliar to me. Worth noting tho that this PR doesn't change the logic by which these flags are applied, so if the original code was doing something wrong I would rather fix that in a followup PR, rather than block this PR on it?

PR #8680, opened 2025-03-26
[vite-plugin] improve inspector port handling

### LINKED ISSUE #8670: 🐛 BUG: The vite plugin errors when the inspector port 9229 is already used

Vite Plugin

vite-plugin@0.1.15

Mac

_No response_

### Describe the Bug

If try to run two different vite dev (or preview) servers using the Cloudflare plugin, I will get an error stating that :::9229 is being already in use (9229 the default port we set for the miniflare inspector), this should not happen
 
this should be changed so that:
 - if I don't specify an inspector port, 9229 is used, but if the port is already taken a new one is picked up instead
 - if I specify an inspector port (such as 9229) and that port is used then the above error should indeed crash their vite server
 
(PS: note that /__debug also does not really care about the port being used so that would still work in the same exact way even if a random port is used)

_No response_

dario-piotrowicz:
Fixes #8670

this change updates the plugin so that if an inspector port is not specified and the
default inspector port (9229) is not available it keeps looking for other available
port instead of crashing

this also makes sure that users can specify `0` to use a random port

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: there is not clean/simple way to test this functionality
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: readme updated
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/README.md
+++ b/packages/vite-plugin-cloudflare/README.md
@@ -376,7 +376,7 @@ It accepts an optional `PluginConfig` parameter.
 
 - `inspectorPort?: number | false`
 
-  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. By default the vite plugin will attempt to use the first available port it can find starting at `9229`.
+  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. If the value is `0` a random inspector port is chosen. By default the vite plugin will attempt to use the first available port it can find starting at `9229`.
 
 > [!NOTE]
 > When running `wrangler deploy`, only your main (entry) Worker will be deployed.
--- a/packages/vite-plugin-cloudflare/README.md
+++ b/packages/vite-plugin-cloudflare/README.md
@@ -376,7 +376,7 @@ It accepts an optional `PluginConfig` parameter.
 
 - `inspectorPort?: number | false`
 
-  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. If the value is `0` a random inspector port is chosen. By default the vite plugin will attempt to use the first available port it can find starting at `9229`.
+  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). By default the vite plugin will attempt to use the first available port it can find starting at `9229`.
 
 > [!NOTE]
 > When running `wrangler deploy`, only your main (entry) Worker will be deployed.
--- a/packages/vite-plugin-cloudflare/README.md
+++ b/packages/vite-plugin-cloudflare/README.md
@@ -376,7 +376,7 @@ It accepts an optional `PluginConfig` parameter.
 
 - `inspectorPort?: number | false`
 
-  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. Defaults to `9229`.
+  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. By default the vite plugin will attempt to use the first available port it can find starting at `9229`.
 
 > [!NOTE]
 > When running `wrangler deploy`, only your main (entry) Worker will be deployed.
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -5,6 +5,7 @@ import * as path from "node:path";
 import { createMiddleware } from "@hattip/adapter-node";
 import MagicString from "magic-string";
 import { Miniflare } from "miniflare";
+import colors from "picocolors";
 import * as vite from "vite";
 import {
 	createModuleReference,
@@ -316,16 +317,17 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					"Unexpected error: No Vite HTTP server"
 				);
 
-				const inspectorPort =
-					pluginConfig.inspectorPort ??
-					(await getFirstAvailablePort(DEFAULT_INSPECTOR_PORT));
+				const inputInspectorPort = await getInputInspectorPortOption(
+					pluginConfig,
+					viteDevServer
+				);
 
 				if (miniflare) {
 					await miniflare.setOptions(
 						getDevMiniflareOptions(
 							resolvedPluginConfig,
 							viteDevServer,
-							inspectorPort
+							inputInspectorPort
 						)
 					);
 				} else {
@@ -333,7 +335,7 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 						getDevMiniflareOptions(
 							resolvedPluginConfig,
 							viteDevServer,
-							inspectorPort
+							inputInspectorPort
 						)
 					);
 				}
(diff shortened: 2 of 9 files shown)


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -53,6 +54,9 @@ let workersConfigsWarningShown = false;
 
 let miniflare: Miniflare | undefined;
 
+/** The inspector port (or false if inspecting is disabled) to use for the vite preview command */
+let previewInspectorPort: number | false | undefined;
```
dario-piotrowicz, line 58: using a variable like this is really ugly.... if anyone has a better suggestion I'm all ears 🥲

CarmenPopoviciu (approved):

packages/vite-plugin-cloudflare/README.md
```diff
@@ -376,7 +376,7 @@ It accepts an optional `PluginConfig` parameter.
 
 - `inspectorPort?: number | false`
 
-  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. Defaults to `9229`.
+  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). Can be set to `false` to disable the debugging inspector altogether. If the value is `0` a random inspector port is chosen. By default the vite plugin will attempt to use the first available port it can find starting at `9229`.
```
jamesopstad, line 379: Don't think there's any need to mention the `0` value. Would there ever be a reason to use it with the new default behaviour?

```suggestion
  Optional inspector port to use for debugging your workers, for more details on debugging see the [devtools section](#devtools). By default the vite plugin will attempt to use the first available port it can find starting at `9229`.
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -53,6 +54,9 @@ let workersConfigsWarningShown = false;
 
 let miniflare: Miniflare | undefined;
 
+/** The resolved inspector port (or undefined if inspecting is disabled) */
+let resolvedInspectorPort: number | undefined;
```
jamesopstad, line 58: Does this need to be in the module scope rather than the `cloudflare` function scope?

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -82,14 +83,16 @@ function workerNameToEnvironmentName(workerName: string) {
...
 	userConfig: vite.UserConfig,
 	viteEnv: vite.ConfigEnv
-): ResolvedPluginConfig {
+): Promise<ResolvedPluginConfig> {
 	const configPaths = new Set<string>();
 	const persistState = pluginConfig.persistState ?? true;
-	const inspectorPort = pluginConfig.inspectorPort ?? DEFAULT_INSPECTOR_PORT;
+	const inspectorPort =
+		pluginConfig.inspectorPort ??
+		(await getFirstAvailablePort(DEFAULT_INSPECTOR_PORT));
```
jamesopstad, line 95: I would rather that we move this into the `configureServer` hook. It's not needed at build time and currently it's the only thing making this function and the `config` hook async.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -316,6 +320,11 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					);
 				}
 
+				if (resolvedPluginConfig.inspectorPort !== false) {
+					const miniflareInspectorUrl = await miniflare.getInspectorURL();
+					resolvedInspectorPort = Number.parseInt(miniflareInspectorUrl.port);
+				}
```
jamesopstad, line 326: Looks like this is only used inside the debug middleware, in which case it could just be moved there and the `resolvedInspectorPort` variable could be local.

I'm also wondering though, if a user is connecting manually (e.g. with VSCode) how will they know what port to connect to?

packages/vite-plugin-cloudflare/README.md
dario-piotrowicz: yeah, that's a good point 👍

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: ah, no you're right, sorry my bad 👍

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: > Looks like this is only used inside the debug middleware, in which case it could just be moved there and the resolvedInspectorPort variable could be local.

Good call, done 👍

dario-piotrowicz: mh... thinking about it, at least the warning feels necessary as people might really get frustrated if the thing they can debug their worker on 9229 but it turns out they can't 😕

dario-piotrowicz: > Looks like this is only used inside the debug middleware, in which case it could just be moved there and the `resolvedInspectorPort` variable could be local.

No we can't to that, is this what you had in mind? https://github.com/cloudflare/workers-sdk/pull/8680/commits/8382228e9faeb2f87ac359da6b06da19b30103fc

That doesn't work because the debug plugin needs to be run before the main one so `configureServer` in the debug plugin there runs before `configureServer` of the main one where we actually instantiate miniflare

If feels a bit ugly but if you want I guess we can add the logic inside the debug plugin middleware itself so that it does that on demand 😕 , like so:
```js
viteDevServer.middlewares.use(async (req, res, next) => {
	const resolvedInspectorPort =
		await getResolvedInspectorPort(pluginConfig);
	if (req.url === debuggingPath && resolvedInspectorPort) {
		const html = getDebugPathHtml(workerNames, resolvedInspectorPort);
		res.setHeader("Content-Type", "text/html");
		return res.end(html);
	}
	next();
});
```

I'm not a huge fan of performing that plugin resolution all the time... although it's quite cheap and not a huge deal, I'm happy to go with that if that's your preference

jamesopstad: Might be worth getting @petebacondarwin's opinion on this too. I'm wondering now if finding another port is actually the right approach for the inspector. Another option would be to disable the inspector if the default port is already in use and display `-> Debug: port 9229 is already in use` dimmed (or something to that effect). Sorry to be indecisive.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -83,7 +85,7 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 			name: "vite-plugin-cloudflare",
 			// This only applies to this plugin so is safe to use while other plugins migrate to the Environment API
 			sharedDuringBuild: true,
-			config(userConfig, env) {
+			async config(userConfig, env) {
```
jamesopstad, line 88:
```suggestion
			config(userConfig, env) {
```

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: As we agreed async let's keep the implementation as is for now 😄

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -315,13 +317,26 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					"Unexpected error: No Vite HTTP server"
 				);
 
+				const inputInspectorPort = await getInputInspectorPortOption(
+					pluginConfig,
+					viteDevServer
+				);
+
```
jamesopstad, line 324: Not sure it's right to run this on restarts i.e. when the Worker config changes etc.. Isn't there a risk that it assigns a new port each time because it's already in use? Could you use `getInputInspectorPortOption` in the `if` block and reuse the returned value in the `else` block?

jamesopstad (approved):
Question about server restarts (I think this needs to be manually tested if you haven't already) but otherwise looks good.

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: oops, thanks 🙂👍

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: I don't think that that's the right call 🤔 

Note that this runs also when you make changes in the `vite.config.ts`, if I were to make a change there during development to the `inspectorPort` itself (say from it being undifined to a specific port for example) I would definitely want this change to picked up and reflected in my dev server, wouldn't I?
(PS: the above doesn't even work nicely right now 😢, I'm looking into why that is)

dario-piotrowicz: oof.... the above does not work properly because miniflare does not update the inspector port on `.setOptions` 😓  I will look into that, but if that's ok by you I would do it in a followup PR? 🙏

jamesopstad: Good thinking about needing to support changing the `inspectorPort` in the Vite config. Yes, happy for this to be merged as is and for that to be investigated in a follow-up PR.

PR #8682, opened 2025-03-26
fix: include asset binding when printing binding summary

CarmenPopoviciu:


CarmenPopoviciu:
closing as we decided we're not backporting patches for beta features

PR #8685, opened 2025-03-26
[vite-plugin] make sure that `.dev.vars` files trigger a full reload

### LINKED ISSUE #8669: 🐛 BUG: The vite plugin does not watch .dev.vars

Vite Plugin

vite-plugin@0.1.15

Mac

_No response_

### Describe the Bug

When running vite dev using the Cloudflare vite plugin, when changes are made in the `.dev.vars` files the vite dev server should restarts (in the same way it does when changes are made in wrangler config files) but it doesn't

_No response_

dario-piotrowicz:
Fixes #8669

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -681,3 +678,67 @@ function getDotDevDotVarsContent(
...
 	return null;
 }
+
+/**
+ * Returns a function that checks if a given resolved path to a `changedFile` matches
+ * some of the files that Cloudflare cares about (and that require the vite server to
+ * reload)
+ *
+ * Note: the function is cached based on the resolvedPluginConfig just so that
+ *       the extra paths resolutions are only done once
```
dario-piotrowicz, line 688: arguably if I have to do this I could as well resolve all the paths and drop them onto `resolvedPluginConfig` when we create the object 🤔

I can give that alternative a shot if people dislike this approach 🙂

petebacondarwin (commented):
Most of this PR appears to be optimising for performance by caching the paths and functions. I am a bit worried that the module level `getHasCloudflareFileChangedMap` could result in a memory leak after lots of updates. But more importantly, did you do this because there was a noticeable performance issue? When I did the original work to support the _headers and _redirects files, I couldn't tell any difference to how long it took for that code to run. So I wonder if all the complexity is even needed?

dario-piotrowicz:
> Most of this PR appears to be optimising for performance by caching the paths and functions. I am a bit worried that the module level `getHasCloudflareFileChangedMap` could result in a memory leak after lots of updates. But more importantly, did you do this because there was a noticeable performance issue? When I did the original work to support the _headers and _redirects files, I couldn't tell any difference to how long it took for that code to run. So I wonder if all the complexity is even needed?

it might be not needed, I didn't actually see any performance issue, I figured it would be beneficial to make this more performance since the `hotUpdate` function can be called many times, I was trying to prevent perf issues rather than addressing anything, I'm totally happy to remove any optimization logic and address any potential optimization need later when/if needed 👍 

(PS: I am also starting to wonder if the current caching logic is what's causing the linux/windows CI flakiness for some reason 😕 (as I can't think of anything else that could cause that))

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -681,3 +682,22 @@ function getDotDevDotVarsContent(
...
+	resolvedPluginConfig: ResolvedPluginConfig,
+	changedFile: string
+) {
+	const dotDevDotVarsPaths = new Set(
+		[...resolvedPluginConfig.configPaths].map((configPath) =>
+			path.join(path.dirname(configPath), ".dev.vars")
+		)
+	);
+
+	// Note that we must "resolve" the changed file since the path from Vite will not match Windows backslashes.
```
petebacondarwin, line 699: We can resolve the `changedFile` in the hotUpdate hook and the pass the resolved one to  each of the checks.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -681,3 +682,22 @@ function getDotDevDotVarsContent(
...
+/**
+ * Returns true if the `changedFile` matches one of a potential .dev.vars file.
+ */
+function hasDotDevDotVarsFileChanged(
+	resolvedPluginConfig: ResolvedPluginConfig,
+	changedFile: string
+) {
+	const dotDevDotVarsPaths = new Set(
+		[...resolvedPluginConfig.configPaths].map((configPath) =>
+			path.join(path.dirname(configPath), ".dev.vars")
```
petebacondarwin, line 695: Do we need to handle `.dev.vars.<env>`  type files too?

dario-piotrowicz: oh.... that's a good question! 🤔  I need to check, I wonder if we support them in the first place

dario-piotrowicz: yes, we do! great call, I'll address this, thanks! 🫶

dario-piotrowicz:
I've converted the PR to draft as I need to:
 - [ ] handle `.dev.vars.<env>` files too
 - [ ] make sure the changes don't make the tests flaky
        (they currently are very flaky in CI for linux and windows, I haven't managed to figured out why...)

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: done 🙂

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -308,13 +308,16 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 				}
 			},
 			hotUpdate(options) {
+				// Note that we must "resolve" the changed file since the path from Vite will not match Windows backslashes.
+				const changedFilePath = path.resolve(options.file);
+
 				if (
-					// Vite normalizes `options.file` so we use `path.resolve` for Windows compatibility
-					resolvedPluginConfig.configPaths.has(path.resolve(options.file)) ||
+					resolvedPluginConfig.configPaths.has(path.resolve(changedFilePath)) ||
```
jamesopstad, line 315:
```suggestion
					resolvedPluginConfig.configPaths.has(changedFilePath) ||
```

jamesopstad (approved):
Looks great!

Out of interest, what was the reason for replacing `.toBe` with `.toMatch` in some of the tests? Was this leading to flakes?

packages/vite-plugin-cloudflare/src/index.ts
dario-piotrowicz: 🤦 👍

PR #8688, opened 2025-03-26
Ensure that Node.js polyfills are pre-optimized before the first request

petebacondarwin:
Fixes [DEVX-1769](https://jira.cfdata.org/browse/DEVX-1769)

Previously, these polyfills were only optimized on demand when Vite became aware of them.
This was either because Vite was able to find an import to a polyfill when statically analysing the import tree of the entry-point,
or when a polyfilled module was dynamically imported as part of a executing code to handle a request.

In the second case, the optimizing of the dynamically imported dependency causes a reload of the Vite server, which can break applications that are holding state in modules during the request.
This is the case of most React type frameworks, in particular React Router.

Now, we pre-optimize all the possible Node.js polyfills when the server starts before the first request is handled.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: bug fix / refactor
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: vite plugin is not backported

### DIFF
--- a/packages/vite-plugin-cloudflare/e2e/global-setup.ts
+++ b/packages/vite-plugin-cloudflare/e2e/global-setup.ts
@@ -3,7 +3,7 @@ import os from "node:os";
 import path from "node:path";
 import util from "node:util";
 import { startMockNpmRegistry } from "@cloudflare/mock-npm-registry";
-import type { GlobalSetupContext } from "vitest/node";
+import type { TestProject } from "vitest/node";
 
 declare module "vitest" {
 	export interface ProvidedContext {
@@ -13,8 +13,7 @@ declare module "vitest" {
 
 // Using a global setup means we can modify tests without having to re-install
 // packages into our temporary directory
-// Typings for the GlobalSetupContext are augmented in `global-setup.d.ts`.
-export default async function ({ provide }: GlobalSetupContext) {
+export default async function ({ provide }: TestProject) {
 	const stopMockNpmRegistry = await startMockNpmRegistry(
 		"@cloudflare/vite-plugin"
 	);
@@ -28,7 +27,9 @@ export default async function ({ provide }: GlobalSetupContext) {
 	return async () => {
 		await stopMockNpmRegistry();
 
-		console.log("Cleaning up temporary directory...");
-		await fs.rm(root, { recursive: true, maxRetries: 10 });
+		if (!process.env.CLOUDFLARE_VITE_E2E_KEEP_TEMP_DIRS) {
+			console.log("Cleaning up temporary directory...", root);
+			await fs.rm(root, { recursive: true, maxRetries: 10 });
+		}
 	};
 }
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -29,6 +29,7 @@ import {
 import {
 	injectGlobalCode,
 	isNodeCompat,
+	nodeCompatEntries,
 	nodeCompatExternals,
 	NODEJS_MODULES_RE,
 	nodejsBuiltins,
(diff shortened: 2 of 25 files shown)


packages/vite-plugin-cloudflare/e2e/global-setup.ts
```diff
@@ -3,7 +3,7 @@ import os from "node:os";
 import path from "node:path";
 import util from "node:util";
 import { startMockNpmRegistry } from "@cloudflare/mock-npm-registry";
-import type { GlobalSetupContext } from "vitest/node";
+import type { TestProject } from "vitest/node";
```
petebacondarwin, line 6: GlobalSetupContext is deprecated

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -574,6 +566,30 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+											result.resolved
+										);
+									return registration?.processing;
+								}
+							}
+						);
+						await Promise.all(processingPromises);
+					}
+				}
+			},
```
jamesopstad, line 592: I've just tried this out and unfortunately it seems to prevent the normal dependency crawling and optimisation from executing. To reproduce, try these steps in the `prisma` playground:

- Run `pnpm dev` without making a request
- Notice that the `.vite` directory contains all the `unenv` polyfills but is missing the `prisma` imports.
- Make a request and notice that the `@prisma/adapter-d1` and `@prisma/client-generated` dependencies are now optimised
- Comment out the `configureServer` hook shown here and remove the `.vite` directory
- Run `pnpm dev` without making a request
- Notice that the `.vite` directory contains the `prisma` imports but not the `unenv` polyfills

I feel like the `configureServer` hook is the wrong place to be executing this logic. I also have doubts about whether we should use `depsOptimizer.registerMissingImport` at all as it seems more like an internal rather than public API.

petebacondarwin: Thanks for this feedback @jamesopstad - I'll investigate.

petebacondarwin: We needed to wait for the dependency optimizer to initialize before registering these other deps. Otherwise it appears that registering imports will cancel the crawl of the static imports of the entry-point.

Fix with test inbound.

petebacondarwin: See https://github.com/cloudflare/workers-sdk/pull/8688/files#diff-0217c39a7be6fcdb410ac8dfea2ae2998b882a1aaba3f2e83ff2831a0beee7bfR582

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -579,6 +571,37 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+											result.resolved
+										);
+									return registration?.processing;
+								}
+							}
+						);
+						// Wait for all the additional deps to be optimized before continuing to serve requests.
+						await Promise.all(processingPromises);
+					}
+				}
```
jamesopstad, line 603: Just a suggestion. We could wrap all of this in a `Promise.all` so that more work can be done in parallel.

```suggestion
				await Promise.all(
					Object.values(viteDevServer.environments).flatMap(
						async (environment) => {
							const workerConfig = getWorkerConfig(environment.name);
							if (isNodeCompat(workerConfig)) {
								// Make sure that the dependency optimizer has been initialized.
								// This ensures that its standard static crawling to identify libraries to optimize still happens.
								// If you don't call `init()` then the calls to `registerMissingImport()` appear to cancel the static crawling.
								await environment.depsOptimizer?.init();

								// Register every unenv-preset entry-point with the dependency optimizer upfront before the first request.
								// Without this the dependency optimizer will try to bundle them on-the-fly in the middle of the first request.
								// That can potentially cause problems if it causes previously optimized bundles to become stale and need to be bundled.
								return Array.from(nodeCompatEntries).map((entry) => {
									const result = resolveNodeJSImport(entry);
									if (result) {
										const registration =
											environment.depsOptimizer?.registerMissingImport(
												result.unresolved,
												result.resolved
											);
										return registration?.processing;
									}
								});
							}
						}
					)
				);
```

jamesopstad (approved):
Looks great! Thanks for figuring this out.

jamesopstad:
This comment is no longer valid now that you've removed the `configureServer` hook in `vite-plugin-cloudflare:nodejs-compat-warnings`. Is the logic still correct?

petebacondarwin:
> This comment is no longer valid now that you've removed the `configureServer` hook in `vite-plugin-cloudflare:nodejs-compat-warnings`. Is the logic still correct or is there a chance the assertion will fail?
> 
> https://github.com/cloudflare/workers-sdk/blob/49540547369a5ae5fca9345d50dcdbdc80b1d502/packages/vite-plugin-cloudflare/src/index.ts#L691

Yep - good spot. I updated the comment: it is now the resolveId hook that sets this element. The assertion is still valid.

PR #8698, opened 2025-03-27
remove BETA label from Workers+Assets templates and hide Pages templates

petebacondarwin:
Fixes [DEVX-1707](https://jira.cfdata.org/browse/DEVX-1707)

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no test facing change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Not Wrangler nor Vite plugin related
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/21331
  - [ ] Documentation not necessary because: Covered by new Workers + Assets docs
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  C3 is not backported

### DIFF
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -472,17 +472,25 @@ export const createContext = async (
 				label: "platform",
 				question: "Select your deployment platform",
 				options: [
-					{
-						label: "Workers with Assets (BETA)",
-						value: "workers",
-						description:
-							"Take advantage of the full Developer Platform, including R2, Queues, Durable Objects and more.",
-					},
-					{
-						label: "Pages",
-						value: "pages",
-						description: "Great for simple websites and applications.",
-					},
+					...(frameworkConfig.platformVariants.workers.hidden
+						? []
+						: [
+								{
+									label: "Workers with Assets",
+									value: "workers",
+									description:
+										"Take advantage of the full Developer Platform, including R2, Queues, Durable Objects and more.",
+								},
+							]),
+					...(frameworkConfig.platformVariants.pages.hidden
+						? []
+						: [
+								{
+									label: "Pages",
+									value: "pages",
+									description: "Great for simple websites and applications.",
+								},
+							]),
 					backOption,
 				],
 				defaultValue: "workers",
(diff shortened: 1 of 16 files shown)


packages/create-cloudflare/src/templates.ts
```diff
@@ -444,17 +444,25 @@ export const createContext = async (
...
-						value: "workers",
-						description:
-							"Take advantage of the full Developer Platform, including R2, Queues, Durable Objects and more.",
-					},
-					{
-						label: "Pages",
-						value: "pages",
-						description: "Great for simple websites and applications.",
-					},
+					...(frameworkConfig.platformVariants.workers.hidden
```
emily-shen, line 447: we need to still return the hidden options if `args.platform` has been provided, otherwise using `--platform=pages` shows up as 

`├ Select your deployment platform (skipped)`

petebacondarwin:
Also I need to remove the choices.

packages/create-cloudflare/src/templates.ts
petebacondarwin: I think I've fixed this... 3bd4c1325321cf9eff13e1891b4d18fe4a4d8477

emily-shen (approved):
manually tested as well :)

i think this hasn't been rebased since nextjs graduated, could you rebase and hide the next on pages template?

emily-shen:
https://github.com/cloudflare/cloudflare-docs/pull/21331 docs

> * [x]  I need to skip the prompt when there is only one option

up to you, but maybe it would be helpful to still show the prompt so you know which platform you've ended up with?

petebacondarwin:
> > * [x]   I need to skip the prompt when there is only one option
> 
> up to you, but maybe it would be helpful to still show the prompt so you know which platform you've ended up with?

The prompt is shown but the question is skipped:

image

petebacondarwin:
> qwik has both options workers and pages, but if you try analog which only has pages. i guess this is the same behaviour its always had so it really doesn't have to be resolved here, just never really noticed it until i was trying a bunch of frameworks in a row.

Ah right. We should add Workers support after Dev Week and then this will resolve itself 😄 

Thanks for the testing and reviewing!

PR #8665, opened 2025-03-25
thomasgauvin: add logs for remote/local for kv operations

thomasgauvin:
Fixes https://discord.com/channels/595317990191398933/1354056005637636107/1354132416918196254

PR #8746, opened 2025-04-01
Log whether a command is operating on a local or remote resource

emily-shen:
Added to KV and R2 for now.

skipping d1 since its commands haven't been refactored to use createCommand yet.

Not added to KV list or get commands, since they seem to implicitly expect the relevant values to be returned (e.g. printBanner = false)

replaces #8665 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: messaging update
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not relevant to v3 (local by default happened in v4)

### DIFF
--- a/packages/create-cloudflare/templates/astro/workers/templates/ts/wrangler.jsonc
+++ b/packages/create-cloudflare/templates/astro/workers/templates/ts/wrangler.jsonc
@@ -11,5 +11,8 @@
   },
   "observability": {
     "enabled": true
-  }
+  },
+	"vars":{
+		"TBD": "<TBD>"
+	}
 }
--- a/packages/wrangler/src/core/register-yargs-command.ts
+++ b/packages/wrangler/src/core/register-yargs-command.ts
@@ -1,4 +1,3 @@
-import assert from "node:assert";
 import { fetchResult } from "../cfetch";
 import { readConfig } from "../config";
 import { defaultWranglerConfig } from "../config/config";
--- a/packages/wrangler/src/core/register-yargs-command.ts
+++ b/packages/wrangler/src/core/register-yargs-command.ts
@@ -1,10 +1,11 @@
+import assert from "node:assert";
 import { fetchResult } from "../cfetch";
 import { readConfig } from "../config";
 import { defaultWranglerConfig } from "../config/config";
 import { FatalError, UserError } from "../errors";
 import { run } from "../experimental-flags";
 import { logger } from "../logger";
-import { printResourceLocation } from "../utils/is-local";
+import { isLocal, printResourceLocation } from "../utils/is-local";
 import { printWranglerBanner } from "../wrangler-banner";
 import { demandSingleValue } from "./helpers";
 import type { CommonYargsArgv, SubHelp } from "../yargs-types";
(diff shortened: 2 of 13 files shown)


packages/wrangler/src/core/types.ts
```diff
@@ -123,6 +123,13 @@ export type CommandDefinition<
 		 * If true, then look for a redirect file at `.wrangler/deploy/config.json` and use that to find the Wrangler configuration file.
 		 */
 		useConfigRedirectIfAvailable?: boolean;
+
+		/**
+		 * If true, print a message about whether the command is operating on a local or remote resource
+		 */
+		printLocalResourceWarning?:
```
CarmenPopoviciu, line 130: the naming is a bit confusing since it actually prints whether the resource is local or remote. Should we rename to `printResourceLocationWarning` instead?

packages/wrangler/src/core/register-yargs-command.ts
```diff
@@ -90,6 +91,23 @@ function createHandler(def: CommandDefinition) {
 			// TODO(telemetry): send command started event
 
 			await def.validateArgs?.(args);
+
+			if (def.behaviour?.printLocalResourceWarning) {
+				// if remote is undefined, it defaults to local
+				let remote = "remote" in args ? args.remote : undefined;
+				if ("local" in args && args.local === false) {
+					remote = false;
```
CarmenPopoviciu, line 99: so `args.local === false` does not imply `remote = true`?

emily-shen: ugh i meant `"local" in args && args.local === undefined` (so if both are undefined, `remote = false`, ie default to local)

packages/wrangler/src/core/register-yargs-command.ts
```diff
@@ -90,6 +91,28 @@ function createHandler(def: CommandDefinition) {
 			// TODO(telemetry): send command started event
 
 			await def.validateArgs?.(args);
+
+			if (def.behaviour?.printResourceLocation) {
```
penalosa, line 95: The types indicate this could also be a function?

penalosa (commented):
I think we need to unwrap the passed function

CarmenPopoviciu (approved):

penalosa (approved):

pmiguel (approved):
KV-size, LGTM.

packages/create-cloudflare/templates/astro/workers/templates/ts/wrangler.jsonc
```diff
@@ -11,5 +11,8 @@
   },
   "observability": {
     "enabled": true
-  }
+  },
+	"vars":{
+		"TBD": "<TBD>"
```
petebacondarwin, line 16: @emily-shen - did this mean to be here?

emily-shen: nope D: 

https://github.com/cloudflare/workers-sdk/pull/8779

PR #8747, opened 2025-04-01
[miniflare] fix: make sure that `miniflare#setOptions` allows the update of inspector ports

dario-piotrowicz:
Currently if I use an `inspectorPort` in a miniflare instance and then try to change such
port via `setOptions` such change is not actually applied, this PR fixes that

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: bugfix of an undocumented feature
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: this is a bugfix of a miniflare feature not used by wrangler v3

### DIFF
--- a/packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-controller.ts
+++ b/packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-controller.ts
@@ -60,7 +60,12 @@ export class InspectorProxyController {
 
 		this.#initializeWebSocketServer(server);
 
+		const listeningPromise = new Promise<void>((resolve) =>
+			server.once("listening", resolve)
+		);
 		server.listen(await this.#inspectorPort);
+
+		await listeningPromise;
 
 		return server;
 	}
@@ -70,7 +75,11 @@ export class InspectorProxyController {
 		await new Promise<void>((resolve, reject) => {
 			server.close((err) => (err ? reject(err) : resolve()));
 		});
+		const listeningPromise = new Promise<void>((resolve) =>
+			server.once("listening", resolve)
+		);
 		server.listen(await this.#inspectorPort);
+		await listeningPromise;
 	}
 
 	#initializeWebSocketServer(server: Server) {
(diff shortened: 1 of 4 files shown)


packages/miniflare/src/plugins/core/inspector-proxy/inspector-proxy-controller.ts
```diff
@@ -55,11 +60,29 @@ export class InspectorProxyController {
...
+	async #restartServer() {
+		const server = await this.#server;
+		server.closeAllConnections();
+		await new Promise<void>((resolve, reject) => {
+			server.close((err) => (err ? reject(err) : resolve()));
+		});
+		const listeningPromise = new Promise<void>((resolve) =>
+			server.once("listening", resolve)
+		);
+		server.listen(await this.#inspectorPort);
```
penalosa, line 82: Do we need to re-initialise the websocket server here?

dario-piotrowicz: oh, that's a great call out! I need to double check that, we probably need to! thanks! 🫶 

(I wanted to keep things simple and didn't add a test that actually uses the websocket, I probably ought to add one!)

dario-piotrowicz: (mh... but on second thought this worked with fine with my manual testing (where I also tested the devtools page)... mh.... 🤔 maybe this is something that we should do for cleanliness but not completely necessary? 🤔)

dario-piotrowicz: I re-checked it does seem to work just fine 🙂 

I think `WebSocketServer` is simply coping with the underlying server stopping and re-starting 🙂

dario-piotrowicz: I've added a test to make sure that this works as intended 🙂 : https://github.com/cloudflare/workers-sdk/pull/8747/commits/a91fd77cc2b6bfbcfc9d5e38e9760d55eb4a7120

penalosa (approved):

PR #8754, opened 2025-04-01
improve turbo settings and vite e2e test config

petebacondarwin:

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: test and build config changes
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: test and build config changes
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  test and build config changes

### DIFF
--- a/packages/vite-plugin-cloudflare/playground/node-compat/__tests__/worker-warnings/warnings.spec.ts
+++ b/packages/vite-plugin-cloudflare/playground/node-compat/__tests__/worker-warnings/warnings.spec.ts
@@ -1,15 +1,18 @@
 import dedent from "ts-dedent";
 import { expect, test, vi } from "vitest";
-import { serverLogs } from "../../../__test-utils__";
+import { isBuild, serverLogs } from "../../../__test-utils__";
 
-test("should display warnings if nodejs_compat is missing", async () => {
-	await vi.waitFor(async () => {
-		expect(serverLogs.warns.join("").replaceAll("\\", "/")).toContain(
-			dedent`
+test.skipIf(isBuild)(
+	"should display warnings if nodejs_compat is missing",
+	async () => {
+		await vi.waitFor(async () => {
+			expect(serverLogs.warns.join("").replaceAll("\\", "/")).toContain(
+				dedent`
 				Unexpected Node.js imports for environment "worker". Do you need to enable the "nodejs_compat" compatibility flag? Refer to https://developers.cloudflare.com/workers/runtime-apis/nodejs/ for more details.
 				 - "node:assert/strict" imported from "worker-warnings/index.ts"
 				 - "perf_hooks" imported from "worker-warnings/index.ts"
 				`
-		);
-	});
-});
+			);
+		});
+	}
+);
--- a/packages/wrangler/package.json
+++ b/packages/wrangler/package.json
@@ -64,7 +64,7 @@
 		"test": "dotenv -- pnpm run assert-git-version && dotenv -- vitest",
 		"test:ci": "pnpm run test run",
 		"test:debug": "pnpm run test --silent=false --verbose=true",
-		"test:e2e": "dotenv -- vitest -c ./e2e/vitest.config.mts",
+		"test:e2e": "dotenv -- vitest run -c ./e2e/vitest.config.mts",
 		"test:watch": "pnpm run test --testTimeout=50000 --watch",
 		"type:tests": "tsc -p ./src/__tests__/tsconfig.json && tsc -p ./e2e/tsconfig.json"
 	},
--- a/packages/wrangler/turbo.json
+++ b/packages/wrangler/turbo.json
@@ -61,12 +61,16 @@
 		},
 		"test:ci": {
 			"inputs": ["!*/**/.wrangler/**", "**/__tests__/**"],
-			"dependsOn": ["build"]
+			"dependsOn": ["build"],
+			"env": ["VITEST", "NODE_DEBUG", "MINIFLARE_WORKERD_PATH"]
 		},
 		"test:e2e": {
 			"inputs": ["e2e/**"],
 			"dependsOn": ["build"],
 			"env": [
+				"VITEST",
+				"NODE_DEBUG",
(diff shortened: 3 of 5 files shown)


turbo.json
```diff
@@ -3,7 +3,7 @@
 	"remoteCache": {
 		"signature": true
 	},
-	"globalEnv": [
+	"globalPassThroughEnv": [
```
penalosa, line 6: Is this correct? This will mean that these values won't contribute to cache keys, which I'm pretty sure we don't want (`CI_OS` should, at the very least)

packages/vite-plugin-cloudflare/playground/node-compat/__tests__/worker-warnings/warnings.spec.ts
```diff
@@ -1,15 +1,18 @@
 import dedent from "ts-dedent";
 import { expect, test, vi } from "vitest";
-import { serverLogs } from "../../../__test-utils__";
+import { isBuild, serverLogs } from "../../../__test-utils__";
 
-test("should display warnings if nodejs_compat is missing", async () => {
-	await vi.waitFor(async () => {
-		expect(serverLogs.warns.join("").replaceAll("\\", "/")).toContain(
-			dedent`
+test.skipIf(isBuild)(
```
petebacondarwin, line 5: We I think is that because the node compat worker-warnings playground doesn't have nodejs_compat enabled (in order to trigger the warnings) it does build successfully by externalising the node compat imports. This means that when the build output is previewed, Miniflare fails to start because the node imports can't be resolved.

Something to do with our testing setup means that this doesn't happen consistently so we are going to skip this test for builds for now.

packages/wrangler/turbo.json
```diff
@@ -21,55 +21,61 @@
 				"config-schema.json"
 			],
 			"env": [
```
CarmenPopoviciu, line 23: this is the list of environment variables a task depends on. Any change to these environment variables will cause all tasks to miss cache

packages/wrangler/turbo.json
```diff
@@ -21,55 +21,61 @@
...
-				"PATH",
-				"WRANGLER_LOG",
-				"EXPERIMENTAL_MIDDLEWARE",
-				"FORMAT_WRANGLER_ERRORS",
+				"NODE_ENV",
+				"SOURCEMAPS",
+				"SPARROW_SOURCE_KEY",
+				"SENTRY_DSN"
+			],
+			"passThroughEnv": [
```
CarmenPopoviciu, line 31: A list of environment variables that are made available to tasks. Any change to these environment variables will not affect how tasks miss/hit the cache

CarmenPopoviciu (approved):

packages/wrangler/package.json
```diff
@@ -64,7 +64,7 @@
 		"test": "dotenv -- pnpm run assert-git-version && dotenv -- vitest",
 		"test:ci": "pnpm run test run",
 		"test:debug": "pnpm run test --silent=false --verbose=true",
-		"test:e2e": "dotenv -- vitest -c ./e2e/vitest.config.mts",
+		"test:e2e": "dotenv -- vitest run -c ./e2e/vitest.config.mts",
```
petebacondarwin, line 67: Just a quality of life improvement for local development. Without this the test:e2e goes into interactive mode and never exists.

packages/wrangler/turbo.json
petebacondarwin: The ones listed here in `env` are actually used by the `scripts/bundle.ts` file to inject code into the Wrangler build output. So it is important that these are part of the cache key.

packages/wrangler/turbo.json
petebacondarwin: The env vars here are only actually used at runtime (i.e. when you run `wrangler` the CLI tool). So they do not affect the build output at all, and therefore there is no need to include these in the cache key.

Notably we previously had `PWD` in the cache key, which meant that just changing directory and running the wrangler build again would miss the cache.

packages/wrangler/turbo.json
```diff
@@ -21,55 +21,61 @@
...
+				"WRANGLER_LOG",
+				"WRANGLER_SEND_METRICS",
+				"WRANGLER_WORKER_REGISTRY_PORT"
 			]
 		},
 		"test:ci": {
 			"inputs": ["!*/**/.wrangler/**", "**/__tests__/**"],
-			"dependsOn": ["build"]
+			"dependsOn": ["build"],
+			"env": ["VITEST", "NODE_DEBUG", "MINIFLARE_WORKERD_PATH"]
```
petebacondarwin, line 65: These env vars affect the running of the tests. They need to be in the cache key but also if they were not here they would not be provided to the underlying test process so would have no effect.

packages/wrangler/turbo.json
```diff
@@ -21,55 +21,61 @@
...
 		},
 		"test:e2e": {
 			"inputs": ["e2e/**"],
-			"dependsOn": ["build"]
+			"dependsOn": ["build"],
+			"env": [
```
petebacondarwin, line 70: These vars affect the wrangler e2e tests and need to be passed through to the vitest process and be part of the cache key.

turbo.json
```diff
@@ -3,14 +3,8 @@
...
 	},
-	"globalEnv": [
-		"CI_OS",
-		"NODE_VERSION",
-		"VITEST",
-		"NODE_DEBUG",
-		"NODE_EXTRA_CA_CERTS",
-		"MINIFLARE_WORKERD_PATH"
-	],
+	"globalEnv": ["CI_OS", "NODE_VERSION"],
```
petebacondarwin, line 6: Let's ensure that we cache separately for each OS and Node.js version.
Particularly relevant for our shared CI cache.

PR #8759, opened 2025-04-02
fix: Minor refactor for the R2 catalog commands

garvit-gupta:
fix: Minor refactor in the R2 data catalog wrangler commands

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required (https://github.com/cloudflare/workers-sdk/pull/8761)
  - [ ] Not required because:
- Public documentation
  - [] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/issues/21030
  - [ ] Documentation not necessary because:
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: patching commands for a new product that was only added to wrangler v4

PR #8761, opened 2025-04-02
Run CI on behalf of #8759

workers-devprod:
This PR is created to run CI on behalf of \#8759. It can be closed after the CI run is complete.


petebacondarwin:
For future reference the tests here did pass

PR #8768, opened 2025-04-03
No longer warn if the user sets `upload_source_maps` in the Worker config.

jamesopstad:
Fixes #000.

No longer warn if the user sets `upload_source_maps` in the Worker config.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/old-guests-smile.md
+++ b/.changeset/old-guests-smile.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+No longer warn if the user sets `upload_source_maps` in the Worker config.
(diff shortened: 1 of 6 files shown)


emily-shen:
does this mean wrangler can now upload source maps (which get generated by vite)?

jamesopstad:
> does this mean wrangler can now upload source maps (which get generated by vite)?

Yes (sorry, I forgot this is relevant to your docs PR). Previously, we were thinking about doing this implicitly based on whether the user sets Vite to output source maps. We decided in the end, however, that there might be scenarios where the user wants to preview a production build with source maps but not have them uploaded. That's why we're not warning on this field anymore.

petebacondarwin (approved):

PR #8774, opened 2025-04-03
Remove sourcemaps from NPM release

### LINKED ISSUE #8770: sourcemaps in published packages
Do we need to publish sourcemaps for the workers packages?

for example, `wrangler` is a CLI, so is there a use case outside of contributors to this project who actually care about having the sourcemaps?

The code isn't minified either, so is still readable if a contributor did need to debug it, but then surely they can just link it up locally where sourcemaps do exist.

We could drop at least ~10MB from wrangler's published package by removing those.

I suspect we could do the same for other packages in this repo, too.

penalosa:
Fixes #8770

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: change to release process
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: change to release process
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: change to release process
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: change to release process

### DIFF
--- a/packages/wrangler/scripts/bundle.ts
+++ b/packages/wrangler/scripts/bundle.ts
@@ -96,7 +96,7 @@ const embedWorkersPlugin: Plugin = {
 					format: "esm",
 					target: "esnext",
 					bundle: true,
-					sourcemap: true,
+					sourcemap: process.env.SOURCEMAPS !== "false",
 					sourcesContent: false,
 					metafile: true,
 					entryPoints: [args.path],
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -33,6 +33,7 @@ jobs:
         run: pnpm run build
         env:
           CI_OS: ${{ runner.os }}
+          SOURCEMAPS: "false"
 
       - name: Create Version PR or Publish to NPM
         id: changesets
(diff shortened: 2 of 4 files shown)


packages/wrangler/scripts/bundle.ts
```diff
@@ -96,7 +96,7 @@ const embedWorkersPlugin: Plugin = {
 					format: "esm",
 					target: "esnext",
 					bundle: true,
-					sourcemap: true,
+					sourcemap: process.env.SOURCEMAPS !== "false",
```
petebacondarwin, line 99: LOL we were setting SOURCEMAP to false but then ignoring it here!

petebacondarwin (approved):
What about create-cloudflare? Does that have sourcemaps?

penalosa:
> What about create-cloudflare? Does that have sourcemaps?

It doesn't, no

43081j:
`miniflare` has sourcemaps FYI so maybe that's worth doing in a separate PR if it makes sense to

PR #8806, opened 2025-04-07
Replace assertion in node-js compat transform hook with early return

jamesopstad:
Fixes #000.

Replace assertion in vite-plugin-cloudflare:nodejs-compat plugin transform hook with early return. This prevents an error from being logged when building with React Router and TailwindCSS.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/big-trees-drum.md
+++ b/.changeset/big-trees-drum.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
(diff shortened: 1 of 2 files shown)


dario-piotrowicz (approved):
Looks good to me, although I would add a TODO comment there saying why we early return instead of asserting and that we should hopefully reinstate the assert and address the issue more properly (if we do think that there might a better solution)

PR #8823, opened 2025-04-07
fix: replace `process.env.NODE_ENV` for nodejs_compat builds

dario-piotrowicz:
This PR makes sure that occurrences of `process.env.NODE_ENV` are replaced with the
current `process.env.NODE_ENV` value or `"production"` on builds that include
the `nodejs_compat` flag, this enables libraries checking such value
(e.g. `react-dom`) to be properly treeshaken

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/little-cities-yell.md
+++ b/.changeset/little-cities-yell.md
@@ -1,0 +1,10 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+fix: replace `process.env.NODE_ENV` for nodejs_compat builds
+
+make sure that occurrences of `process.env.NODE_ENV` are replaced with the
+current `process.env.NODE_ENV` value or `"production"` on builds that include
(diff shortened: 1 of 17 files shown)


jamesopstad (approved):
Great! Thanks Dario.

jamesopstad:
Just tried running the prerelease with React Router and can confirm that it solves the issue. This is logged during build:

[plugin replace] @rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.

Is it worth making that change so the user doesn't see this?

PR #8845, opened 2025-04-08
Refactor Turbo + CI jobs

penalosa:
This PR is a semi-drastic approach to improving our CI situation. It:
* fixes up some turbo settings to make caching work better (see individual comments)
* Only runs multi-OS tests on VP, which will hopefully drastically reduce the pain of getting PRs in at the cost of perhaps more frequent reverts/fixups when we only detect an issue after it's gotten to Version Packages. However, those reverts/fixups should be really easy to land, which is I think a tradeoff worth at least experimenting with
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: ci refactor
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: ci related
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: ci refactor


.github/pr-labeler.config.yml
```diff
@@ -7,5 +7,5 @@ e2e:
 c3-e2e:
   - any:
       - changed-files:
-          - any-glob-to-any-file: "packages/create-cloudflare/**/*"
+          - any-glob-to-any-file: "packages/create-cloudflare/e2e-tests/*"
```
penalosa, line 10: Not all C3 changes need E2E tests to be run—C3 has unit tests that provide some coverage

.github/workflows/create-pullrequest-prerelease.yml
```diff
@@ -19,6 +19,12 @@ jobs:
 
       - name: Install Dependencies
         uses: ./.github/actions/install-dependencies
+        with:
+          turbo-api: ${{ secrets.TURBO_API }}
+          turbo-team: ${{ secrets.TURBO_TEAM }}
+          turbo-token: ${{ secrets.TURBO_TOKEN }}
+          turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
+          node-version: ${{ matrix.node_version }}
```
penalosa, line 27: Make sure pre-releases can use cached build artifacts

.github/workflows/e2e.yml
```diff
@@ -6,6 +6,84 @@ on:
...
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.node }}
+      cancel-in-progress: true
+    timeout-minutes: 40
+    if: github.event_name == 'pull_request' && contains(github.event.*.labels.*.name, 'e2e' ) && github.event.pull_request.head.repo.owner.login == 'cloudflare' && github.head_ref == 'changeset-release/main'
+    name: "E2E Test"
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-2022, ubuntu-22.04]
```
penalosa, line 19: Run windows + ubuntu Wrangler E2E tests on VP

.github/workflows/e2e.yml
```diff
@@ -16,7 +94,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, windows-2022, ubuntu-22.04]
+        os: [macos-13]
```
penalosa, line 97: ...but only run macos e2e tests on PRs

.github/workflows/test-and-check.yml
```diff
@@ -78,19 +78,28 @@ jobs:
           turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
 
       - name: Build
-        run: pnpm run build
+        run: pnpm run build --summarize
```
penalosa, line 81: Add turbo debug logs so we can check how caching is set up after the fact

.github/workflows/test-and-check.yml
```diff
@@ -78,19 +78,28 @@ jobs:
...
 
-  test:
+      - name: Upload turbo logs
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: turbo-runs-${{ matrix.os }}-${{ matrix.node }}
+          path: .turbo/runs
+
+  test-vp:
```
penalosa, line 101: Run tests on windows + ubuntu on VP

.github/workflows/test-and-check.yml
```diff
@@ -155,3 +160,67 @@ jobs:
...
+    timeout-minutes: 30
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.filter }}-${{ matrix.node_version }}-test
+      cancel-in-progress: true
+
+    name: ${{ format('Tests ({0})', matrix.os_name) }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
```
penalosa, line 181: ...but only run macos tests on PRs

packages/create-cloudflare/turbo.json
```diff
@@ -3,18 +3,20 @@
 	"extends": ["//"],
 	"tasks": {
 		"build": {
+			"passThroughEnv": ["CI"],
```
penalosa, line 6: Make sure the CI env is available, but not part of the cache key

packages/create-cloudflare/turbo.json
```diff
@@ -3,18 +3,20 @@
...
 			"env": [
 				"TEST_PM",
 				"TEST_PM_VERSION",
 				"npm_config_user_agent",
-				"CI",
 				"CREATE_CLOUDFLARE_TELEMETRY_DISABLED",
 				"CREATE_CLOUDFLARE_TELEMETRY_DEBUG",
 				"SPARROW_SOURCE_KEY"
 			],
-			"inputs": ["src/**", "templates*/**", "scripts/**", "*", "vitest*"],
```
penalosa, line 15: This includes most folders, so instead just use the turbo default of all git-tracked files

edmundhung (approved):

.github/pr-labeler.config.yml
petebacondarwin: Hmm, not sure about this one...

petebacondarwin: what if we changed a c3.ts file?

.github/workflows/c3-e2e.yml
```diff
@@ -19,12 +19,11 @@ jobs:
...
         run: |
           npx turbo-ignore create-cloudflare --task="test:e2e" --fallback=$(git merge-base origin/main HEAD)
-
-  e2e:
+  e2e-vp:
     # Note: please keep this job in sync with the e2e-only-dependabot-bumped-framework one
     #       in .github/workflows/c3-e2e-dependabot.yml
     needs: ["turbo-ignore"]
-    if: ${{ needs.turbo-ignore.outputs.skip == 'failure' }}
+    if: needs.turbo-ignore.outputs.skip == 'failure' && github.head_ref == 'changeset-release/main'
```
petebacondarwin, line 26: should we check the org is cloudflare too?

.github/workflows/c3-e2e.yml
```diff
@@ -34,12 +33,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        pm:
-          [
-            { name: npm, version: "0.0.0" },
-            { name: pnpm, version: "9.12.0" },
-            { name: yarn, version: "1.0.0" },
-          ]
+        pm: [{ name: npm, version: "0.0.0" }, { name: yarn, version: "1.0.0" }]
```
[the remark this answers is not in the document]
petebacondarwin: i.e. we don't need to run pnpm because that is covered by the other job?

.github/workflows/c3-e2e.yml
```diff
@@ -91,3 +85,67 @@ jobs:
           experimental: false
           accountId: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
           apiToken: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
+
+  e2e:
+    # Note: please keep this job in sync with the e2e-only-dependabot-bumped-framework one
```
petebacondarwin, line 90: Can we move some of this to an action to reduce duplication?

.github/workflows/e2e.yml
```diff
@@ -6,6 +6,84 @@ on:
...
+          turbo-api: ${{ secrets.TURBO_API }}
+          turbo-team: ${{ secrets.TURBO_TEAM }}
+          turbo-token: ${{ secrets.TURBO_TOKEN }}
+          turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
+
+      - name: Run builds
+        run: pnpm run build
+        env:
+          NODE_ENV: "production"
+          CI_OS: ${{ runner.os }}
```
petebacondarwin, line 41: I don't think this should be needed since it will be built by the `test:e2e` turbo task it the deps are setup correctly.

.github/workflows/e2e.yml
```diff
@@ -6,6 +6,84 @@ on:
...
+        with:
+          name: e2e-test-debug-logs-${{ matrix.os }}-${{ matrix.node }}
+          path: ${{ runner.temp }}/wrangler-debug-logs/
+
+      - name: Upload test report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-test-report-${{ matrix.os }}-${{ matrix.node }}
+          path: ${{ runner.temp }}/test-report/
```
petebacondarwin, line 78: Now we write failed debug logs to the CI output I don't know if there is much value in these anymore.

packages/create-cloudflare/turbo.json
petebacondarwin: I wonder if this should be set in the root turbo.json?

packages/create-cloudflare/turbo.json
```diff
@@ -3,18 +3,20 @@
...
 			"outputs": ["dist/**"]
 		},
+		"test:ci": {
+			"dependsOn": ["build"]
+		},
```
petebacondarwin, line 19: This is already covered in the root turbo.json, right?

packages/miniflare/turbo.json
```diff
@@ -3,12 +3,10 @@
 	"extends": ["//"],
 	"tasks": {
 		"build": {
-			"inputs": ["src/**", "scripts/**", "types/**", "*.mjs", "*.js", "*.json"],
```
petebacondarwin, line 6: The only slightly annoying thing about this is that miniflare is a little slow to build and if you are just making changes to tests the build will have to re-run each time. But I can live with that.

petebacondarwin (approved):
General approval - a few nits.

.github/workflows/c3-e2e.yml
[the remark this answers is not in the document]
penalosa: Yeah (same with the Wrangler E2E tests). The other job runs unconditionally

.github/workflows/c3-e2e.yml
penalosa: I don't think we need to? The C3 tests shouldn't depend on any actions secrets, I don't think (although I may be mistaken there?)

packages/create-cloudflare/turbo.json
penalosa: Yeah, I'm just adding it here to be a bit more explicit. Happy to remove if you'd prefer

.github/pr-labeler.config.yml
penalosa: I've added the templates folder to this too, which should cover that: faa956814

packages/create-cloudflare/turbo.json
penalosa: 157bf0f21

packages/miniflare/turbo.json
penalosa: I've refactored this slightly so that changes to only test files shouldn't cause the build to re-run

turbo.json
```diff
@@ -4,7 +4,7 @@
 		"signature": true
 	},
 	"globalEnv": ["CI_OS", "NODE_VERSION"],
-	"globalPassThroughEnv": ["NODE_EXTRA_CA_CERTS"],
+	"globalPassThroughEnv": ["NODE_EXTRA_CA_CERTS", "CI"],
```
petebacondarwin, line 7: I think we run some different tests based on `CI`...

penalosa: I can make it a global env, but then we'd need to provide it locally for caches to be hit—I suppose that's okay?

penalosa: I've enabled the one test that was disabled locally, so we should be in a place now where local tests are a superset of CI tests, which feels pretty safe

.github/workflows/c3-e2e.yml
```diff
@@ -91,3 +85,67 @@ jobs:
           experimental: false
           accountId: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
           apiToken: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
+
+  e2e:
```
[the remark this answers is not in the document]
emily-shen: this does mean gatsby + next on pages tests wont run on PRs. I suppose its unlikely we'll break either of those.

PR #8885, opened 2025-04-10
fix(wrangler): Disambiguate the "No files to upload. Proceeding with deployment..." message

CarmenPopoviciu:
Some folks are confused about the "No files to upload. Proceeding with deployment..." message when deploying Workers with Assets, especially when lacking the context that wrangler is diffing the local files with the asset manifest and only uploading the changes. This PR disambiguates this messages to something hopefully more meaningful

Fixes DEVX-1572

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: copy change
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: copy change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: assets is a beta feature in wrangler v3. we don't backport patches to beta features

### DIFF
--- a/packages/wrangler/src/assets.ts
+++ b/packages/wrangler/src/assets.ts
@@ -89,7 +89,9 @@ export const syncAssets = async (
 				{ telemetryMessage: true }
 			);
 		}
-		logger.info(`No files to upload. Proceeding with deployment...`);
+		logger.info(
+			`No updated asset files to upload. Proceeding with deployment...`
+		);
 		return initializeAssetsResponse.jwt;
 	}
 
--- a/.changeset/mean-badgers-battle.md
+++ b/.changeset/mean-badgers-battle.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Disambiguate the "No files to upload. Proceeding with deployment..." message
--- a/packages/wrangler/e2e/deployments.test.ts
+++ b/packages/wrangler/e2e/deployments.test.ts
@@ -274,7 +274,7 @@ Current Version ID: 00000000-0000-0000-0000-000000000000`);
 				[
 					"Uploaded 1 of 1 assets",
 					"Uploaded 1 of 2 assets",
-					"No files to upload.",
+					"No updated asset files to upload.",
 				].some((s) => output.includes(s))
 			).toBeTruthy();
 		},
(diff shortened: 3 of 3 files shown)


packages/wrangler/src/assets.ts
```diff
@@ -89,7 +89,9 @@ export const syncAssets = async (
 				{ telemetryMessage: true }
 			);
 		}
-		logger.info(`No files to upload. Proceeding with deployment...`);
+		logger.info(
+			`No changed asset files found to upload. Proceeding with deployment...`
```
CarmenPopoviciu, line 93: It would be great if we can somehow signal to users that the changes are wrt previous deployments, but not sure how to capture that in one sentence. Ideas welcome 🙏

emily-shen: No assets changed since the last deployment. Proceeding with deployment...

maybe?

CarmenPopoviciu: the thing is it's not just last deployment right? It's *any* previous deployment

emily-shen: i think that's effectively the same thing right? each deployment updates the asset state, so its just current vs last that matters?
tbh i dont know EWC handles things like asset deletion super well, i'm pretty sure it doesn't keep that asset around?

CarmenPopoviciu: ok, so I just tested this and here is what I'm seeing:

Say we have a project with one asset file `index.html`. The contents of the file is `Hello`. 
- `wrangler deploy`
- change `index.html` contents to `Hello1` + wrangler deploy`
- change `index.html` contents to `Hello2` + wrangler deploy`
- change `index.html` contents to `Hello` + wrangler deploy`. Stdout shows `No files to upload. Proceeding with deployment...`

So we changed `index.html`'s contents to that from 3 version before. So EWC doesn't compare crr vs last, but crr vs any previous deployment

I'm guessing `No assets changed since previous deployments. Proceeding with deployment...` sounds shady right?

CarmenPopoviciu: > i think that's probably fine - why does it seem shady to you?

strictly speaking from a grammar pow...wasn't sure if that's proper English 😅 

kk...in that case let's go with `No updated asset files to upload. Proceeding....` and take it from there.

Thx so much for your input Emily 🙏 🧡

edmundhung (approved):

emily-shen (approved):

PR #8886, opened 2025-04-10
use generated types in c3

emily-shen:
Fixes DEVX-1613

This PR updates c3 to use types generated by `wrangler types` instead of installing `@cloudflare/workers-types`.

- Removes code that installs and updates config for `@cloudflare/workers-types from most frameworks. 
- deletes `worker-configuration.d.ts` etc. from framework templates, as this now gets generated automatically.

A few frameworks (remix and qwik) pin wrangler 3, which generate runtime types in a different way, so for simplicity we continue installing workers-types for those frameworks. This can be cleaned up in a followup if/when those frameworks upgrade to wrangler v4. 

Testing for framework templates has also been updated to always check types based on the template config (ie verifyBuildCfTypes is no longer necessary), and checks the immediate output of c3 instead of the output after re-running typegen.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/20082
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/create-cloudflare/src/workers.ts
+++ b/packages/create-cloudflare/src/workers.ts
@@ -73,13 +73,13 @@ const maybeInstallNodeTypes = async (ctx: C3Context, npm: string) => {
 		parsedConfig = TOML.parse(wranglerTomlStr);
 	}
 
-	const compatibility_flags = Array.isArray(parsedConfig["compatibility_flags"])
+	const compatibilityFlags = Array.isArray(parsedConfig["compatibility_flags"])
 		? parsedConfig["compatibility_flags"]
 		: [];
 
 	if (
-		compatibility_flags.includes("nodejs_compat") ||
-		compatibility_flags.includes("nodejs_compat_v2")
+		compatibilityFlags.includes("nodejs_compat") ||
+		compatibilityFlags.includes("nodejs_compat_v2")
 	) {
 		await installPackages(["@types/node"], {
 			dev: true,


packages/create-cloudflare/templates/gatsby/workers/c3.ts
emily-shen, line 1: ssg frameworks like gatsby used to get workers types installed - now they don't get any runtime types. could that cause any issues...? would it be safer to just generate types for these too?

packages/create-cloudflare/src/cli.ts
```diff
@@ -168,6 +167,10 @@ const configure = async (ctx: C3Context) => {
 	addWranglerToGitIgnore(ctx);
 
 	await updatePackageScripts(ctx);
+	if (ctx.template.installWorkersTypes) {
```
emily-shen, line 170: qwik and remix only.

packages/create-cloudflare/src/workers.ts
```diff
@@ -1,74 +1,103 @@
...
 import { detectPackageManager } from "helpers/packageManagers";
 import { installPackages } from "helpers/packages";
 import * as jsonc from "jsonc-parser";
-import type { C3Context } from "types";
+import type { C3Context, PackageJson } from "types";
 
 /**
- * Installs the latest version of the `@cloudflare/workers-types` package
- * and updates the .tsconfig file to use the latest entrypoint version.
+ * Generate types using `wrangler types` and update tsconfig
```
vicb, line 14:
```suggestion
 * Generate types using the `cf-typegen` script and update tsconfig
```

packages/create-cloudflare/src/workers.ts
```diff
@@ -1,74 +1,103 @@
...
-	const { npm } = detectPackageManager();
 
+export async function generateWorkersTypes(ctx: C3Context) {
 	if (!usesTypescript(ctx)) {
 		return;
 	}
+	const packageJsonPath = resolve("package.json");
+	const packageManifest = readJSON(packageJsonPath) as PackageJson;
+	if (!Object.keys(packageManifest.scripts ?? {}).includes("cf-typegen")) {
+		return;
```
vicb, line 24: Do we want to return a true/false status?

packages/create-cloudflare/src/workers.ts
```diff
@@ -1,74 +1,103 @@
...
+	}
+
+	delete packageManifest["devDependencies"]?.["@cloudflare/workers-types"];
+
+	writeFile(packageJsonPath, JSON.stringify(packageManifest, null, 2));
+	await updateTsConfig(ctx);
 }
 
-export async function addWorkersTypesToTsConfig(ctx: C3Context) {
+export async function updateTsConfig(ctx: C3Context) {
```
vicb, line 56: Would be nice to add a comment about what the function does

packages/create-cloudflare/src/templates.ts
```diff
@@ -142,10 +142,19 @@ export type TemplateConfig = {
 	/** The key of the package.json "scripts" entry for previewing the project. Defaults to undefined (there might not be such script) */
 	previewScript?: string;
 
+	/** The file path to the generated types file. Defaults to worker-configuration.d.ts*/
+	typesPath?: string;
```
vicb, line 146: IIUC `typesPath` and `cf-typegen` have to be in sync which will probably cause bugs down the road.

IMO we should either extract `typesPath` from the script of generate the script from the path

packages/create-cloudflare/src/workers.ts
```diff
@@ -1,74 +1,103 @@
...
-		doneText: `${brandColor("installed")} ${dim(`via ${npm}`)}`,
+	const { npm } = detectPackageManager();
+
+	const typesCmd = [npm, "run", "cf-typegen"];
+
+	await runCommand(typesCmd, {
+		cwd: ctx.project.path,
+		silent: true,
+		env: {
+			CLOUDFLARE_ACCOUNT_ID: ctx.account?.id,
```
vicb, line 35: question: why is the account needed here?

emily-shen:
> Added a few comments.
> 
> Is there any reason why i.e. Next is not in this PR?

next + workers was in the pr, but i seem to have left out next on pages by accident - ac7026a

packages/create-cloudflare/templates/hello-world-with-assets/c3.ts
```diff
@@ -21,5 +21,6 @@ const config: TemplateConfig = {
 			},
 		},
 	},
+	compatibilityFlags: ["nodejs_compat"],
```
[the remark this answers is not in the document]
emily-shen: the flag was already there (in the wrangler configs), just not in the c3 templateConfig here. Its not great that they can be out of sync like this, but this sort of thing happens all over c3 and is probably beyond the scope of this PR to fix. (I'll make sure I don't add to the problem though...)

packages/create-cloudflare/src/workers.ts
emily-shen: i don't think this is necessary, we don't need to do anything with the return value currently.

packages/create-cloudflare/src/templates.ts
emily-shen: i completely agree in principle, but c3 is totally covered in this pattern (see `previewScript` right above), and its probably not worth fixing it here. its non-trivial to parse the command to extract the typesPath from that, and we can't just generate the command from the typesPath because there are other options that may be passed in.

fwiw we haven't had any problems so far, and this is well tested, so I think its not a big risk.

packages/create-cloudflare/templates/docusaurus/pages/c3.ts
```diff
@@ -27,5 +27,6 @@ const config: TemplateConfig = {
 	devScript: "preview",
 	deployScript: "deploy",
 	previewScript: "preview",
+	skipWranglerTypegen: false,
```
vicb, line 30: Do we want this as this is the default value?

packages/create-cloudflare/src/cli.ts
```diff
@@ -168,6 +167,12 @@ const configure = async (ctx: C3Context) => {
...
```
vicb, line 170: would `installWorkersTypes === true` and `skipWranglerTypegen === false` be valid? (i.e. both install and generate) ?

if not we could have have `workersTypes === "generated" | "installed# | undefined`

packages/create-cloudflare/src/templates.ts
vicb: > we can't just generate the command from the typesPath because there are other options that may be passed in.

Are we using any options beside `--env-interface` in this package ?

As of today we have 

```ts
			verifyBuildCfTypes: {
				outputFile: "env.d.ts",
				envInterfaceName: "CloudflareEnv",
			},
```			

in the e2e config, which is one more thing that should we should sync with the script.

We could move those 2 props from the e2e to the template config and generate the command.

If needed later we could add more options.

packages/create-cloudflare/templates/docusaurus/pages/c3.ts
emily-shen: think that was a typo, superseded by [746c163](https://github.com/cloudflare/workers-sdk/pull/8886/commits/746c16364eea22c8143fed0754e409881a347f95) anyway

packages/create-cloudflare/src/templates.ts
emily-shen: [5b80895](https://github.com/cloudflare/workers-sdk/pull/8886/commits/5b80895f1f98d82cc1cefce14562bca90bbf9504)

packages/create-cloudflare/templates/hello-world-with-assets/c3.ts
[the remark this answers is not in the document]
vicb: I looked at it when working on https://github.com/cloudflare/workers-sdk/pull/9147

This flag is only used by pages (9147 updates the description).

So I think it can be removed from here

packages/create-cloudflare/src/templates.ts
```diff
@@ -142,10 +142,18 @@ export type TemplateConfig = {
...
+	/** The name of the Env type generated by wrangler types. Defaults to `Env`*/
+	envInterfaceName?: string;
+
 	/** The file path of the template. This is used internally and isn't a user facing config value.*/
 	path?: string;
 
 	bindings?: Record<string, unknown>;
+
+	/** Default false. To accomodate SSG frameworks etc. */
```
vicb, line 155:
```suggestion
	/** Defaults to `"none"`. To accomodate SSG frameworks etc. */
```

Comments:
- Maybe we can quickly document each value?
- I'm not sure about what "To accomodate SSG frameworks etc." means here, do we need to keep this

vicb: IIUC the default is actually "generated"

packages/create-cloudflare/templates/hello-world-with-assets/c3.ts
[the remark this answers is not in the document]
emily-shen: we use`compatibilityFlags` in this PR [here](https://github.com/cloudflare/workers-sdk/pull/8886/files#diff-bf88904f9918671fd070ddc546a8a5b005a1a572b68e3a9ee2a2df5ef0d1c754R38) to make sure `@types/node` is installed. i can update the comment to reflect this.

packages/create-cloudflare/src/templates.ts
```diff
@@ -605,6 +613,10 @@ export const createContext = async (
 	const directory = dirname(path);
 	const originalCWD = process.cwd();
 
+	// set default to generate types
```
vicb, line 616: As this function is quite long I think a better pattern would be:

- compute the template value
- use the template

meaning moving this code up to line 567.

You could also use `template = {...default, ...template}` there.

The current code is:

- compute the template value
- use some properties
- add some default

Which makes the code harder to follow and more error prone

packages/create-cloudflare/src/workers.ts
```diff
@@ -1,74 +1,111 @@
...
-export async function installWorkersTypes(ctx: C3Context) {
-	const { npm } = detectPackageManager();
 
+export async function generateWorkersTypes(ctx: C3Context) {
 	if (!usesTypescript(ctx)) {
 		return;
 	}
+	const packageJsonPath = resolve("package.json");
+	const packageManifest = readJSON(packageJsonPath) as PackageJson;
+	if (!Object.keys(packageManifest.scripts ?? {}).includes("cf-typegen")) {
```
vicb, line 23:
```suggestion
	if (!packageManifest.scripts?.["cf-typegen"]) {
```

packages/create-cloudflare/src/workers.ts
```diff
@@ -1,74 +1,111 @@
...
+ * Generate types using the `cf-typegen` script and update tsconfig
  */
-export async function installWorkersTypes(ctx: C3Context) {
-	const { npm } = detectPackageManager();
 
+export async function generateWorkersTypes(ctx: C3Context) {
 	if (!usesTypescript(ctx)) {
 		return;
 	}
+	const packageJsonPath = resolve("package.json");
```
vicb, line 21: Should we use something similar to line 60?

```ts
const tsconfigPath = join(ctx.project.path, "tsconfig.json");
```

packages/create-cloudflare/src/workers.ts
```diff
@@ -1,74 +1,111 @@
...
+				typegen.some((line) =>
+					line.includes("// Runtime types generated with workerd"),
+				)
+			) {
+				newTypes = newTypes.filter(
+					(t: string) => !t.startsWith("@cloudflare/workers-types"),
+				);
+			}
+			// add node types if nodejs_compat is enabled
+			if (ctx.template.compatibilityFlags?.includes("nodejs_compat")) {
```
vicb, line 102: Do we want to also handled `nodejs_compat_v2` ?

vicb (commented):
Added inline comments.

The PR contains quite a lot of changes.
I think it would help reviewing if the description of the PR lists what is changed in the PR - something to keep in mind for future PRs.

Thanks!

emily-shen:
> Added inline comments.
> 
> The PR contains quite a lot of changes. I think it would help reviewing if the description of the PR lists what is changed in the PR - something to keep in mind for future PRs.
> 
> Thanks!

Ah great feedback, i've updated it now for posterity (and good habits!)

packages/create-cloudflare/templates/hello-world-with-assets/c3.ts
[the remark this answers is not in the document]
emily-shen: actually to get around the issue of this getting out of sync, and to support remote templates, that bit should probably read the wrangler config instead - i'll update that.

packages/create-cloudflare/e2e-tests/frameworks.test.ts
```diff
@@ -352,57 +355,53 @@ const verifyPreviewScript = async (
...
 		(line) =>
 			// old type gen - some framework templates pin older versions of wrangler
 			line === `interface ${envInterfaceName} {` ||
 			// new after importable env change
 			line === `interface ${envInterfaceName} extends Cloudflare.Env {}`,
 	);
-	expect(hasEnvInterfacePre).toBe(true);
-
-	// Run the `cf-typegen` script to generate types for bindings in fixture
-	const buildTypesProc = spawnWithLogging(
```
emily-shen, line 384: I don't think its useful here to test `wrangler types` works, so instead this now tests the immediate output of c3 is as expected.

packages/create-cloudflare/src/workers.ts
```diff
@@ -2,73 +2,147 @@ import { existsSync } from "fs";
...
+			allowTrailingComma: true,
+		});
+	} else if (wranglerTomlExists(ctx)) {
+		const wranglerTomlStr = readWranglerToml(ctx);
+		parsedConfig = TOML.parse(wranglerTomlStr);
 	}
 
-	const s = spinner();
-	s.start("Adding latest types to `tsconfig.json`");
+	const compatibility_flags = Array.isArray(parsedConfig["compatibility_flags"])
```
vicb, line 68:
```suggestion
	const compatibilityFlags = Array.isArray(parsedConfig["compatibility_flags"])
```

vicb (approved):
LGTM,
Thanks Emily!

packages/create-cloudflare/templates/next/pages/env.d.ts
petebacondarwin, line 1: Deleting this file has broken the CI in Version Packages @emily-shen

petebacondarwin: Repro:

```
npx create-cloudflare@beta --platform=pages --framework=next -- --yes
```

It is because we need to also remove the bit in c3.ts that is trying to copy it.

petebacondarwin: We missed it in the PR CI jobs because next-on-pages was excluded from pnpm runs.

petebacondarwin: Fix in https://github.com/cloudflare/workers-sdk/pull/9182

PR #8786, opened 2025-04-04
fix: make sure users can change inspector port when running vite dev

dario-piotrowicz:
currently when a user starts a dev server with `vite dev` the inspector port that the Cloudflare plugin will use will always be the initial one, even if the user sets a specific port in the plugin's options inside the vite config file, the changes here make sure that such config updates are instead actually reflected

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: this functionality is not really easily testable
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: bugfix
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a Wrangler change

PR #8901, opened 2025-04-11
chore: add `--force` and `--non-interactive` flags to yarn install command in the vite-plugin e2es

dario-piotrowicz:
Any vite-plugin e2e run is causing the GitHub CI to get stuck (see https://github.com/cloudflare/workers-sdk/pull/8786, https://github.com/cloudflare/workers-sdk/pull/8895)

this happens when we run `yarn install` for setting up the e2e fixtures, by testing it I've seen that passing the
`--force`  and `--non-interactive` flags seems to fix the issue, so that's what I'm adding here

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: fixing problematic e2es
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal fix
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/e2e/helpers.ts
+++ b/packages/vite-plugin-cloudflare/e2e/helpers.ts
@@ -49,10 +49,7 @@ export const test = baseTest.extend<{
 			if (pm === "yarn") {
 				// Note: we use --force and --non-interactive because
 				//       otherwise yarn gets stuck during installation in CI
-				runCommand(
-					`${pm} install --verbose --force --non-interactive`,
-					projectPath
-				);
+				runCommand(`${pm} install --force --non-interactive`, projectPath);
 			} else {
 				runCommand(`${pm} install`, projectPath);
 			}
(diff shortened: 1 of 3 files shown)


packages/vite-plugin-cloudflare/e2e/helpers.ts
```diff
@@ -46,7 +46,9 @@ export const test = baseTest.extend<{
 			debuglog("Fixture copied to " + projectPath);
 			await updateVitePluginVersion(projectPath);
 			debuglog("Updated vite-plugin version in package.json");
-			runCommand(`${pm} install`, projectPath);
+			// Note: we use --force because otherwise yarn
+			//       gets stuck during installation in CI
+			runCommand(`${pm} install --force`, projectPath);
```
dario-piotrowicz, line 51: Note: alternatively I could set the flag conditionally based on `process.env.CI`, but I figured we might as well keep the same behavior locally to avoid potential confusion later on (e.g. if for example something breaks when `--force` is used)

PR #8936, opened 2025-04-14
chore: cleanup e2e test certs

edmundhung:
Fixes n/a.

This make sure no stale certificates are left after running the e2e tests.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/e2e/cert.test.ts
+++ b/packages/wrangler/e2e/cert.test.ts
@@ -125,8 +125,8 @@ describe("cert", () => {
 	const { certificate: caCert } = generateRootCaCert();
 
 	// Generate filenames for concurrent e2e test environment
-	const mtlsCertName = `tmp-e2e-mtls_cert_${randomUUID()}`;
-	const caCertName = `tmp-e2e-ca_cert_${randomUUID()}`;
+	const mtlsCertName = `tmp-e2e-mtls-cert-${randomUUID()}`;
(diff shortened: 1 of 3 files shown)


packages/wrangler/e2e/cert.test.ts
```diff
@@ -125,8 +125,8 @@ describe("cert", () => {
 	const { certificate: caCert } = generateRootCaCert();
 
 	// Generate filenames for concurrent e2e test environment
-	const mtlsCertName = `mtls_cert_${randomUUID()}`;
-	const caCertName = `ca_cert_${randomUUID()}`;
+	const mtlsCertName = `tmp-e2e-mtls_cert_${randomUUID()}`;
```
emily-shen, line 128: this means we won't delete existing ones though. given this is the testing account and we should never have certs been used on here, i think we don't need this and we can just delete if the name has ca_cert/mtls_cert?

edmundhung: I have deleted most of them in the test account yesterday. So there shouldn't be much left there. But I think having a prefix to signal that would still be helpful. 👍🏼

PR #8998, opened 2025-04-17
feat(wrangler): Error if worker-level fields are configured in a static-site only Worker

CarmenPopoviciu:
Fixes DEVX-1560.

Some configuration fields don't make sense for assets-only Workers, and are silently dropped on deploy. This PR throws an error if such fields are found in the `wrangler.toml` file.

Screenshot 2025-04-17 at 13 34 13

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not docummented
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -4442,7 +4442,6 @@ addEventListener('fetch', event => {});`
 				.toThrowErrorMatchingInlineSnapshot(`
 				[Error: Assets-only Workers do not support the following configuration keys:
 
-				⋅ "send_metrics"
 				⋅ "vars"
 				⋅ "kv_namespaces"
 				⋅ "services"
@@ -4474,7 +4473,6 @@ addEventListener('fetch', event => {});`
 				.toThrowErrorMatchingInlineSnapshot(`
 				[Error: Assets-only Workers do not support the following configuration keys:
 
-				⋅ "send_metrics"
 				⋅ "vars"
 				⋅ "kv_namespaces"
 				⋅ "services"
@@ -4508,7 +4506,6 @@ addEventListener('fetch', event => {});`
 					.toThrowErrorMatchingInlineSnapshot(`
 				[Error: Assets-only Workers do not support the following configuration keys:
 
-				⋅ "send_metrics"
 				⋅ "vars"
 				⋅ "kv_namespaces"
 				⋅ "services"
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -4416,6 +4416,108 @@ addEventListener('fetch', event => {});`
 			`);
 		});
 
+		it("should error when deploying an assets-only Worker with unsupported configuration for assets-only", async () => {
+			writeWranglerConfig({
+				assets: {
+					directory: "assets",
+				},
+				send_metrics: true,
+				vars: {
+					MY_VAR: "my var",
+				},
+				services: [
+					{
+						binding: "BINDING_NAME",
+						service: "WORKER_NAME",
+					},
+				],
+				kv_namespaces: [
+					{
+						id: "123",
+						binding: "KV_BINDING_NAME",
+					},
+				],
+			});
+			await expect(runWrangler("deploy")).rejects
+				.toThrowErrorMatchingInlineSnapshot(`
+				[Error: Assets-only Workers do not support the following configuration keys:
+
+				⋅ "send_metrics"
+				⋅ "vars"
+				⋅ "kv_namespaces"
+				⋅ "services"
+
+				Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.]
+			`);
+		});
+
+		it("should error when deploying an assets-only Worker via the --assets flag, with unsupported configuration for assets-only", async () => {
+			writeWranglerConfig({
+				send_metrics: true,
+				vars: {
+					MY_VAR: "my var",
+				},
+				services: [
+					{
+						binding: "BINDING_NAME",
+						service: "WORKER_NAME",
+					},
+				],
+				kv_namespaces: [
+					{
+						id: "123",
+						binding: "KV_BINDING_NAME",
+					},
+				],
+			});
+			await expect(runWrangler("deploy --assets assets")).rejects
+				.toThrowErrorMatchingInlineSnapshot(`
+				[Error: Assets-only Workers do not support the following configuration keys:
+
+				⋅ "send_metrics"
+				⋅ "vars"
+				⋅ "kv_namespaces"
+				⋅ "services"
+
+				Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.]
+			`);
+		});
+
+		it.fails(
+			"should error when deploying an assets-only Worker with unsupported flags for assets-only set",
+			async () => {
+				writeWranglerConfig({
+					assets: {
+						directory: "xyz",
+					},
+					send_metrics: true,
+					services: [
+						{
+							binding: "BINDING_NAME",
+							service: "WORKER_NAME",
+						},
+					],
+					kv_namespaces: [
+						{
+							id: "123",
+							binding: "KV_BINDING_NAME",
+						},
+					],
+				});
+				await expect(runWrangler("deploy --var MY_VAR:13")).rejects
+					.toThrowErrorMatchingInlineSnapshot(`
+				[Error: Assets-only Workers do not support the following configuration keys:
+
+				⋅ "send_metrics"
+				⋅ "vars"
+				⋅ "kv_namespaces"
+				⋅ "services"
+
+				Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.]
+			`);
+			}
+		);
+
 		it("should warn when using smart placement with Worker first", async () => {
 			const assets = [
 				{ filePath: ".assetsignore", content: "*.bak\nsub-dir" },
--- a/packages/wrangler/src/__tests__/dev.test.ts
+++ b/packages/wrangler/src/__tests__/dev.test.ts
@@ -1537,7 +1537,6 @@ describe.sequential("wrangler dev", () => {
 				.toThrowErrorMatchingInlineSnapshot(`
 				[Error: Assets-only Workers do not support the following configuration keys:
 
-				⋅ "send_metrics"
 				⋅ "vars"
 				⋅ "kv_namespaces"
 				⋅ "services"
@@ -1570,7 +1569,6 @@ describe.sequential("wrangler dev", () => {
 				.toThrowErrorMatchingInlineSnapshot(`
 				[Error: Assets-only Workers do not support the following configuration keys:
 
-				⋅ "send_metrics"
 				⋅ "vars"
 				⋅ "kv_namespaces"
 				⋅ "services"
@@ -1605,7 +1603,6 @@ describe.sequential("wrangler dev", () => {
 					.toThrowErrorMatchingInlineSnapshot(`
 				[Error: Assets-only Workers do not support the following configuration keys:
 
-				⋅ "send_metrics"
 				⋅ "vars"
 				⋅ "kv_namespaces"
 				⋅ "services"
--- a/packages/wrangler/src/__tests__/dev.test.ts
+++ b/packages/wrangler/src/__tests__/dev.test.ts
@@ -1474,6 +1474,27 @@ describe.sequential("wrangler dev", () => {
 			);
 		});
 
+		it("should error if config.assets and --site are used together", async () => {
+			writeWranglerConfig({
+				main: "./index.js",
+				assets: {
+					directory: "assets",
+				},
+			});
+			fs.writeFileSync("index.js", `export default {};`);
+			fs.mkdirSync("assets");
+			fs.mkdirSync("xyz");
+
+			await expect(
+				runWrangler("dev --site xyz")
+			).rejects.toThrowErrorMatchingInlineSnapshot(
+				`
+				[Error: Cannot use assets and Workers Sites in the same Worker.
+				Please remove either the \`site\` or \`assets\` field from your configuration file.]
+			`
+			);
+		});
+
 		it("should error if an ASSET binding is provided without a user Worker", async () => {
 			writeWranglerConfig({
 				assets: { directory: "assets", binding: "ASSETS" },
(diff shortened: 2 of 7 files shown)


packages/wrangler/src/assets.ts
```diff
@@ -490,45 +489,27 @@ export function getAssetsOptions(
...
  * and throw an appropriate error if invalid.
  */
 export function validateAssetsArgsAndConfig(
-	args: Pick<StartDevWorkerOptions, "legacy" | "assets" | "entrypoint">
-): void;
-export function validateAssetsArgsAndConfig(
-	args:
-		| Pick<StartDevOptions, "site" | "assets" | "script">
-		| Pick<DeployArgs, "site" | "assets" | "script">,
+	args: {
```
CarmenPopoviciu, line 497: IMHO, unless I'm missing smth, the complexity we had before is unnecessary. We practically only ever need the `site`, `assets`, and `script` args passed into this function, so why make things more complicated than they have to be ¯\_(ツ)_/¯?

packages/wrangler/src/assets.ts
```diff
@@ -538,18 +519,65 @@ export function validateAssetsArgsAndConfig(
...
+					`${keys}\n\n` +
+					`Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.`
+			);
+		}
+	}
 }
 
-const WORKER_JS_FILENAME = "_worker.js";
+function getConfigKeysUnsupportedByAssetsOnly(config: Config): Array<string> {
+	const supportedAssetsOnlyConfigKeys = new Set([
```
CarmenPopoviciu, line 551: @GregBrimble I had a look in Dash, and it would seem that `compatibility_date` and `compatibility_flags` are also disabled for assets-only. Does that sound right to you?

CarmenPopoviciu, line 551: @GregBrimble what do we wanna do abt args here? Some config fields, such as `vars` for example, can also be configured via flags (see `--var`). What's the expected behaviour we're looking for? IMHO it's kinda weird that configuring `[vars]` throws, while `--var` does not. At the same time, we need to consider that this function runs for 3 diff commands `dev`, `deploy`, and `versions upload`, so that's a pretty long list of flags we need to check against. The list of whitelisted config keys is very small, but that won't be the case for args, unless we can use some sort of subset?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4416,6 +4416,108 @@ addEventListener('fetch', event => {});`
...
+				⋅ "send_metrics"
+				⋅ "vars"
+				⋅ "kv_namespaces"
+				⋅ "services"
+
+				Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.]
+			`);
+		});
+
+		it.fails(
```
CarmenPopoviciu, line 4486: related to https://github.com/cloudflare/workers-sdk/pull/8998/files#r2048784806

packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -1489,6 +1510,111 @@ describe.sequential("wrangler dev", () => {
...
+				⋅ "send_metrics"
+				⋅ "vars"
+				⋅ "kv_namespaces"
+				⋅ "services"
+
+				Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.]
+			`);
+		});
+
+		it.fails(
```
CarmenPopoviciu, line 1582: related to https://github.com/cloudflare/workers-sdk/pull/8998/files#r2048784806

packages/wrangler/src/assets.ts
```diff
@@ -538,18 +519,65 @@ export function validateAssetsArgsAndConfig(
...
+	if (
+		(args.assets || config.assets?.directory) &&
+		!(args.script || config.main)
+	) {
+		const unsupportedConfigKeys = getConfigKeysUnsupportedByAssetsOnly(config);
+
+		if (unsupportedConfigKeys.length > 0) {
+			const keys = unsupportedConfigKeys.map((key) => `⋅ "${key}"`).join("\n");
+
+			throw new UserError(
```
CarmenPopoviciu, line 541: are we abs sure we want to throw instead of warn?

packages/wrangler/src/assets.ts
GregBrimble: No. Compat flags and dates should be enabled. If they're disabled in the dash, that's a bug.

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -4416,6 +4416,108 @@ addEventListener('fetch', event => {});`
...
+						id: "123",
+						binding: "KV_BINDING_NAME",
+					},
+				],
+			});
+			await expect(runWrangler("deploy")).rejects
+				.toThrowErrorMatchingInlineSnapshot(`
+				[Error: Assets-only Workers do not support the following configuration keys:
+
+				⋅ "send_metrics"
```
GregBrimble, line 4445: send_metrics is a Wrangler feature which logs usage, no? So that does still feel relevant to static sites.

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -325,7 +325,14 @@ async function resolveConfig(
 		);
 	}
 
-	validateAssetsArgsAndConfig(resolved);
+	validateAssetsArgsAndConfig(
+		{
+			site: resolved.legacy.site?.bucket,
+			assets: input.assets,
+			script: input.entrypoint,
```
penalosa, line 332: I don't think this is right. We were previously passing in the _resolved_ config, which comes from the config file, but now we're just passing in the input config (cli args)

packages/wrangler/src/assets.ts
```diff
@@ -490,45 +489,27 @@ export function getAssetsOptions(
...
-			: (args.assets || config?.assets) && (args.site || config?.site)
-	) {
+	if ((args.assets || config.assets) && (args.site || config.site)) {
 		throw new UserError(
 			"Cannot use assets and Workers Sites in the same Worker.\n" +
 				"Please remove either the `site` or `assets` field from your configuration file."
 		);
 	}
 
-	const noOpEntrypoint = path.resolve(
```
penalosa, line 522: Why is this being removed?

packages/wrangler/src/assets.ts
```diff
@@ -538,18 +519,66 @@ export function validateAssetsArgsAndConfig(
...
+					`${keys}\n\n` +
+					`Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.`
+			);
+		}
+	}
 }
 
-const WORKER_JS_FILENAME = "_worker.js";
+function getConfigKeysUnsupportedByAssetsOnly(config: Config): Array<string> {
+	const supportedAssetsOnlyConfigKeys = new Set([
```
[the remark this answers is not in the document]
CarmenPopoviciu: yeah, that's not right. Will fix

packages/wrangler/src/api/startDevWorker/ConfigController.ts
CarmenPopoviciu: that's because we changed the signature of this function to expect the config and cli args separately. This is why we don't want to pass in the resolved config anymore, but rather the config and cli args separately

penalosa: Didn't the function optionally take config before? I might be missing something, but if we already have the resolved config here, why not just pass it into the function? Why do we need to resolve again?

packages/wrangler/src/assets.ts
CarmenPopoviciu: The whole setup of `validateAssetsArgsAndConfig` was a bit convoluted IMO, mainly - I *think* - because it is called in 3 diff commands `dev`, `deploy`, and `versions upload`, and so, it tried to be generic enough so as to support the diff types of args those commands would provide. So for example, for `wrangler dev` this fn would be called with `args: Pick<StartDevWorkerOptions, "legacy" | "assets" | "entrypoint">` whereas for `wrangler deploy` it would be called with `args: Pick<DeployArgs, "site" | "assets" | "script">` and `config: Config`.

`noOpEntrypoint` was only ever used in the if conditional if we detected `"legacy" in args`. However, having changed `validateAssetsArgsAndConfig`'s signature to separate config from cli args params, the `"legacy" in args` check doesn't need to be performed anymore, and therefore `noOpEntrypoint` became redundant.

AFAIK we have extensive tests that cover this validation, and I've added some more tests. If my changes would have broken anything I'd have expected tests to start failing

packages/wrangler/src/assets.ts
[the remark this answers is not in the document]
CarmenPopoviciu: fixed in https://github.com/cloudflare/workers-sdk/pull/8998/commits/651a8baa3cdd65a600f75d82d9e585d740855212

packages/wrangler/src/deploy/index.ts
```diff
@@ -235,7 +235,14 @@ export const deployCommand = createCommand({
 
 		const entry = await getEntry(args, config, "deploy");
 
-		validateAssetsArgsAndConfig(args, config);
+		validateAssetsArgsAndConfig(
+			{
+				site: args.site,
+				assets: args.assets,
+				script: args.script,
+			},
```
petebacondarwin, line 243: Why this change?

packages/wrangler/src/assets.ts
```diff
@@ -538,18 +519,67 @@ export function validateAssetsArgsAndConfig(
...
+		"name",
+		"compatibility_date",
+		"compatibility_flags",
+		"assets",
+		"build",
```
petebacondarwin, line 556: why is build supported?

GregBrimble: build being supported feels very reasonable to me. it's a Wrangler-only feature (doesn't make it to EWC), and would be useful if you want to build a static site, for example.

packages/wrangler/src/assets.ts
GregBrimble: We lock down these fields in the dash. EWC does accept them, but it also would accept `{ "complete": "junk" }` keys (we're not "strict").

I'd say error sets people up for the most success, but I leave the final decision to ANT. If you'd prefer warning, go for it.

packages/wrangler/src/assets.ts
GregBrimble: I think, it's kinda what Samuel was getting at earlier in a different comment: why isn't this function just receiving a fully-resolved config object? By this point, surely, there's nothing that assets validation cares about which would be specific to args or config: it only cares about which features are enabled/disabled. Whether an arg was used or config was used to enable something that isn't supported, it doesn't matter.

Is there any sane way we can combine them together before calling this?

packages/wrangler/src/assets.ts
```diff
@@ -490,45 +489,27 @@ export function getAssetsOptions(
...
-): void;
-export function validateAssetsArgsAndConfig(
-	args:
-		| Pick<StartDevOptions, "site" | "assets" | "script">
-		| Pick<DeployArgs, "site" | "assets" | "script">
-		| Pick<StartDevWorkerOptions, "legacy" | "assets" | "entrypoint">,
-	config?: Config
 ): void {
-	if (
-		"legacy" in args
```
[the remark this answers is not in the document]
CarmenPopoviciu: this is confused me as well, but this check is not related to legacy assets. It's related to legacy sites

packages/wrangler/src/api/startDevWorker/ConfigController.ts
CarmenPopoviciu: right....so I think conceptually there are two ways we can think abt this function:

1. the function receives a resolved configuration object, where the callee of `validateAssetsArgsAndConfig` is responsible of the `args` <> `config` resolution. 
2. the function receives both the `args` and `config` objects, and is itself responsible of the `args` <> `config` resolution. 

Currently (prior to my changes) we actually do a combination of both, meaning that `validateAssetsArgsAndConfig` can sometimes take a resolved config, and sometimes both `args` and `config`, which to me feels less than ideal, confusing, and unnecessarily complicated. But I can accept that maybe that's just me. 

In my ideal world, I'd like this function to be straightforward and take either one or the other. That way I don't have to worry about doing things like 

```
if ("legacy" in args
	? args.assets && args.legacy.site
	: (args.assets || config?.assets) && (args.site || config?.site)
) {...}
```

when I could instead be doing
```
if(resolvedConfig.assets && resolvedConfig.legacy.site) {....}
```
or
```
if((args.assets || normalizedConfig?.assets) && (args.site || normalizedConfig?.site)) {...}
```
depending on which of the two options we went with. And it's not just that. For the kind of validation that I had to put in place, having these different fn signatures complicates things even more. For one, configuration key naming is not persistent across the resolved config vs the normalized config (`sendMetrics` vs `send_metrics` for example), or at least, it isn't in the current implementation. That means I can't simply traverse my array of whitelisted config keys and check `if(normalizedConfig[key] || resolvedConfig[key])`, because the key name is not the same. Of course these are all problems I can solve, but with added complexity which IMO is not necessary.

Historically we seem to have always taken the second approach, see [here](https://github.com/cloudflare/workers-sdk/blob/508a1a31a039a5f4700efbc7535a165d79b22cb9/packages/wrangler/src/deploy/index.ts#L236) or [here](https://github.com/cloudflare/workers-sdk/blob/508a1a31a039a5f4700efbc7535a165d79b22cb9/packages/wrangler/src/deploy/index.ts#L262) or [here](https://github.com/cloudflare/workers-sdk/blob/508a1a31a039a5f4700efbc7535a165d79b22cb9/packages/wrangler/src/assets.ts#L392), and this is kind of how I was looking at the problem when I created this PR. There are certain limitations to my implementation as well, and if we wanted to do args validation too, which I 100% agree with Greggs we should do, then we'd prolly need to make the type of that `args` param more generic.

Now y'all know that I'm not one to often die on any code hill, so if folks would rather pass a resolved config as a param, or keep the fn signatures as is, and add the validation on top of that, then sure ¯\_(ツ)_/¯. It personally took me a bit to wrap my head around this function, and saw an opportunity to simplify :)

packages/wrangler/src/assets.ts
CarmenPopoviciu: https://github.com/cloudflare/workers-sdk/pull/8998/files#r2068920006

CarmenPopoviciu: yeah, I thought so. thx for confirming

vicb:
@GregBrimble I am lacking context here but I have a question about why this PR is needed.

When I see:

```text
				[Error: Assets-only Workers do not support the following configuration keys:
				⋅ "vars"
				⋅ "kv_namespaces"
				⋅ "services"
				Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.]
```

I wonder why we need to error here. None of vars, KV, services incur cost for users.

So I'm wondering why we should error here? 

The reasoning is that we do not error/warn if we define a KV that is unused in a worker.
My concern here is that this PR introduces code that we will to maintain (when we add more binding types, asset features, ...) and we should only do this if it solves an actual issue.

GregBrimble:
It's about better defining that "pit of success" for users. Here, someone might configure a variable, expecting it to be populated through to their CRA or whatever. Or for a binding to be magically available.

Since these features won't work in a static-only site, rather we warn them early about this misconfiguration rather than let them deploy and then discover it doesn't work.

packages/wrangler/src/assets.ts
korinne: For this case, I think a warning would be better -- we can warn that these fields will be locked down in the Dashboard (assuming this means non-editable). @penalosa what do you think?

packages/wrangler/src/assets.ts
```diff
@@ -538,18 +519,67 @@ export function validateAssetsArgsAndConfig(
...
+	) {
+		const unsupportedConfigKeys = getConfigKeysUnsupportedByAssetsOnly(config);
+
+		if (unsupportedConfigKeys.length > 0) {
+			const keys = unsupportedConfigKeys.map((key) => `⋅ "${key}"`).join("\n");
+
+			throw new UserError(
+				`Assets-only Workers do not support the following configuration keys:\n\n` +
+					`${keys}\n\n` +
+					`Please remove these fields from your configuration file, or configure the "main" field if you are trying to deploy a Worker with assets.`
```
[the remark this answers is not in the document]
korinne: Hmm here's me preferred language (and I'd rather a warning than an error). This might be a bit too long, so would love to see a screenshot: 

**Warning:** :Your [Wrangler configuration file](https://developers.cloudflare.com/workers/wrangler/configuration/) uses server-side configurations (${keys}) without a defined "main" Worker. These settings are inactive if you're only deploying [static assets (https://developers.cloudflare.com/workers/static-assets/). 
  - To deploy just static assets: Please remove the listed keys.
  - To deploy a Worker with executable code: Define the "main" field in Wrangler config, like 'main = "src/worker.ts"'.

CarmenPopoviciu:
going to go ahead and close. The PR is pretty stale by now. If we decide to want this feature after all, we can always revisit

PR #8983, opened 2025-04-16
Removes open-beta disclaimer from workflows commands

Caio-Nogueira:
Fixes WOR-637

This PR removes the open-beta disclaimer that is now outdated since Workflows is GA.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: No new feature was added. This merely removes an outdated disclaimer.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: workflows is open beta in v3

PR #8999, opened 2025-04-17
[V3 Backport] Removes open-beta disclaimer from workflows commands

Caio-Nogueira:
Backport PR for [#8983](https://github.com/cloudflare/workers-sdk/pull/8983).


CarmenPopoviciu:
this backport PR is not necessary since we do not backport patches to experimental features. Furthermore Workflows was graduated to stable only in v4, so technically it is still in open beta in v3. Changing the messaging does not make sense here. I'll go ahead and close

PR #9002, opened 2025-04-17
feat(wrangler): Require user confirmation before running remote dev with assets

CarmenPopoviciu:
Fixes DEVX-1562

This PR adds a prompt when running `wrangler dev --remote` with assets, informing users of possible consequences, and requiring user confirmation to proceed 

Screenshot 2025-04-17 at 14 37 21

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/src/dev.ts
+++ b/packages/wrangler/src/dev.ts
@@ -11,7 +11,7 @@ import {
 	extractBindingsOfType,
 } from "./api/startDevWorker/utils";
 import { getAssetsOptions } from "./assets";
-import { configFileName, formatConfigSnippet } from "./config";
+import { configFileName, formatConfigSnippet, readConfig } from "./config";
 import { createCommand } from "./core/create-command";
 import { validateRoutes } from "./deploy/deploy";
 import { validateNodeCompatMode } from "./deployment-bundle/node-compat";
@@ -19,6 +19,7 @@ import { devRegistry, getBoundRegisteredWorkers } from "./dev-registry";
 import { getVarsForDev } from "./dev/dev-vars";
 import registerDevHotKeys from "./dev/hotkeys";
 import { maybeRegisterLocalWorker } from "./dev/local";
+import { confirm } from "./dialogs";
 import { UserError } from "./errors";
 import isInteractive from "./is-interactive";
 import { logger } from "./logger";
@@ -598,6 +599,20 @@ export async function startDev(args: StartDevOptions) {
 	try {
 		if (args.logLevel) {
 			logger.loggerLevel = args.logLevel;
+		}
+
+		if (args.remote) {
+			const config = readConfig(args);
+
+			if (args.assets || config.assets) {
+				const ok = await confirm(
+					"Using `wrangler dev --remote` with assets requires a new Worker to be uploaded to the server. Do you want to continue?"
+				);
+
+				if (!ok) {
+					throw new UserError("Command aborted by user");
+				}
+			}
 		}
 
 		const authHook: AsyncHook<CfAccount, [Pick<Config, "account_id">]> = async (
--- a/.changeset/cold-pans-crash.md
+++ b/.changeset/cold-pans-crash.md
@@ -1,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Require user confirmation before running a remote development session with Workers + assets
--- a/packages/wrangler/src/__tests__/dev.test.ts
+++ b/packages/wrangler/src/__tests__/dev.test.ts
@@ -15,6 +15,7 @@ import { logger } from "../logger";
 import { sniffUserAgent } from "../package-manager";
 import { mockAccountId, mockApiToken } from "./helpers/mock-account-id";
 import { mockConsoleMethods } from "./helpers/mock-console";
+import { mockConfirm } from "./helpers/mock-dialogs";
 import { useMockIsTTY } from "./helpers/mock-istty";
 import {
 	msw,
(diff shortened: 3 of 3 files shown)


packages/wrangler/src/dev.ts
```diff
@@ -600,6 +601,20 @@ export async function startDev(args: StartDevOptions) {
 			logger.loggerLevel = args.logLevel;
 		}
 
+		if (args.remote) {
+			const config = readConfig(args);
```
CarmenPopoviciu, line 605: if there is a better way and we can avoid reading the config I'm all years. But unless we move this conditional at a deeper internal wrangler layer, which I'd really want to avoid, I dunno how.

penalosa: Could we put it in the ConfigController?

packages/wrangler/src/dev.ts
```diff
@@ -600,6 +601,20 @@ export async function startDev(args: StartDevOptions) {
...
+
+			if (args.assets || config.assets) {
+				const ok = await confirm(
+					"Using `wrangler dev --remote` with assets requires a new Worker to be uploaded to the server. Do you want to continue?"
```
penalosa, line 609: What does this actually mean? `uploaded to the server` doesn't tell me much. Is it just for new sessions? Will it deploy the current state of the codebase to production?

CarmenPopoviciu: yeah, the copy here prolly needs some rephrasing. I basically took what I had in the ticket. Maybe @korinne or @GregBrimble want to chime in here?

packages/wrangler/src/dev.ts
CarmenPopoviciu: I think that's what William tried in the first stab at this and he was getting some weird behaviour. I don't have the context anymore, but can give it a try. I also had mixed feelings abt having this sort of interactivity live at controller level, but it's not a hill I'd die on

penalosa: That's fair, and if we can keep this interactivity out of the controllers that'd be ideal, but I feel slightly uneasy about this double-read of config that's needed here.

Is there more info somewhere on the motivation for this?

CarmenPopoviciu: I'm abs with you on this, on both reading the config twice and having the interactivity live at controller level. Between the two, reading the config twice felt like the lesser evil because:

- historically command interactivity lived at handler level
- the controller codepaths are hit in multiple flows (cmd startup, config changes, etc), which we might or might not want to target with our interactivity logic

I did go ahead and try your suggestion out, and it does work fine. The only issue is that users now have to confirm on every config update, which is very annoying IMO, so I'll need to put a mechanism in place that makes sure we only ask for confirmation once per dev session. This sort of "side-effect" goes back to the second point I was making earlier

> Is there more info somewhere on the motivation for this?

I had a similar question...Assets-only is one use case that stuck in my memory, where users might find it unexpected that a Worker was deployed on their account, and that traffic to that Worker will count towards their monthly usage

packages/wrangler/src/dev.ts
GregBrimble, line 609: This should only be shown if a Worker with this name does not already exist on the user's account. If a Worker with this name does not already exist, uploading assets will implicitly create a placeholder Worker with that name. We'd expect users to, `dev --remote` around for a while and then likely `wrangler deploy` (replacing that placeholder) so it's not a big deal. We made the decision a while ago that this wasn't worth cleaning up after a `dev --remote` session.

```suggestion
      // idk if config.name is environment aware?
      const workerExists = fetch("/workers/services/${config.name}").then(resp => resp.status !== 404);
			if ((args.assets || config.assets) && !workerExists) {
				const ok = await confirm(
					"As one does not already exist, using `wrangler dev --remote` with assets will deploy a temporary Worker with the name ${parsedConfig.name} to your account. It will be automatically replaced if you ever run `wrangler deploy`. Do you want to continue?"
```

packages/wrangler/src/dev.ts
GregBrimble: I'd definitely say read twice. Because, in https://github.com/cloudflare/workers-sdk/pull/9002/files#r2067104992, I'm suggesting you make a fetch afterwards too.

> Assets-only is one use case that stuck in my memory, where users might find it unexpected that a Worker was deployed on their account, and that traffic to that Worker will count towards their monthly usage

Nothing like that. It's just because we need a place to upload assets for a `dev --remote` session. If there is no Worker yet, we need to create one! The API already automatically implicitly does create one, so we just want to warn users before they end up with these temporary placeholder Workers if they are doing something really weird and trying out 20 differently named Workers and never deploying any of them.

penalosa: > we just want to warn users before they end up with these temporary placeholder Workers if they are doing something really weird and trying out 20 differently named Workers and never deploying any of them.

Yeah, that's the thing I'm hung up on. _Why_ do we want to warn users? Is this a complaint we've been hearing? This type of thing really messes up Wrangler's data model and I'd love to be absolutely sure that we want to interactively prompt. We don't do this for anything else right now, and for instance uploading a secret to a non-existent Worker will also create the Worker in the background.

CarmenPopoviciu: didn't realise that either. Thx both for chiming in 🙏

Issue #9006, opened 2025-04-19
vite-plugin: secrets store bindings throws Secret "[SOME_SECRET_KEY]" not found


woodybriggs:
### Which Cloudflare product(s) does this pertain to?

Vite Plugin

### What versions & operating system are you using?

wrangler:4.12.0, @cloudflare/vite-plugin:1.0.8

### Please provide a link to a minimal reproduction

https://github.com/woodybriggs/lucky-term-181e

### Describe the Bug

Secrets store is unable to find bound secret when using @cloudflare/vite-plugin in development.
I have dug around, and as far as i can get is that the miniflare instance that holds the secret_store does not have a bound store
`packages/miniflare/src/workers/secrets-store/secret.worker.ts line:14`
`this.env.store` is an empty KvNamespace; In that the response of this.env.store.list() returns an empty list using vite dev but with wrangler dev it returns the correct list of secrets.

This seems like it could be a miniflare configuration issue that is caused by the vite-plugin.

thelovekesh:
Hey, any updates here?

PR #9007, opened 2025-04-19
[vite-plugin-cloudflare] fix: secrets store persist path

woodybriggs:
Fixes [#9006](https://github.com/cloudflare/workers-sdk/issues/9006).

_add persist path for secrets store for vite-plugin_

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/vite-plugin-cloudflare/src/miniflare-options.ts
+++ b/packages/vite-plugin-cloudflare/src/miniflare-options.ts
@@ -62,7 +62,7 @@ function getPersistence(
 		kvPersist: path.join(persistPath, "kv"),
 		r2Persist: path.join(persistPath, "r2"),
 		workflowsPersist: path.join(persistPath, "workflows"),
-		secretsStorePersist: path.join(persistPath, "secrets-store"),
+		secretsStorePersist: path.join(persistPath, "kv"),
 	};
 }
 
--- a/packages/vite-plugin-cloudflare/src/miniflare-options.ts
+++ b/packages/vite-plugin-cloudflare/src/miniflare-options.ts
@@ -37,6 +37,7 @@ type PersistOptions = Pick<
 	| "kvPersist"
 	| "r2Persist"
 	| "workflowsPersist"
+	| "secretsStorePersist"
 >;
 
 function getPersistence(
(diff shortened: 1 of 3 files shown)


woodybriggs:
Really not sure if using the KV persist path is the right call. I know that SecretsStore uses KV internally in development mode, so it defaults to storing local secrets in the KV persist path as a seperate db, when using the wrangler cli to create local secrets.

happy to try and change this so that secrets store has its own sub dir as a sibling to .wrangler/state/v3/kv i.e. .wrangler/state/v3/secrets-store

woodybriggs:
I have spent quite a while on this today.

I have run into an issues while trying to use a separate directory for the persist path of secret store. To summarize the issue I'm having:

When trying to use both kv and secrets store at the same time (With different persist paths as config.) secrets store ends up using kv persist path instead of config provided persist path.

This is where I have spent most of my time debugging today, but am unable to pinpoint where the config gets overridden for secretsStorePersist.

I have managed to create a test that reproduces this issue. I will push as part of this PR.

woodybriggs:
I believe I understand the issue now.

Both kv and secrets-store plugins produce a kv:storage service via plugin.getServices call. One of these kv:storage services gets deduped. I believe that the kv:storage service emitted by secrets-service is the one being ommited. As it is defined in the PLUGIN object list after the kv plugin [here](packages/miniflare/src/plugins/index.ts). This is actually a miniflare issue and not a vite-plugin issue.

My current stance on this PR is that... The changes in the PR do "fix" the issue, but a decision needs to be made about whether or not this fix is okay for the long term.

woodybriggs:
@edmundhung no worries, thanks for taking the time to look at this bug and my related PRs. I wish i could have helped more. But it seemed like an internal decision was required so I felt like i couldnt help more.

PR #9024, opened 2025-04-22
Add undefined/null checks to containers list

IRCody:
Provide helpful error message when user lists containers but has none to list instead of an obscure error.

_Describe your change..._
Fixes an error where a user doing:
`wrangler containers list` but had no containers would get an error message like:

"Cannot read properties of undefined..." 
Because the returned applications list was empty.

This adds checks for undefined/null/length 0 and provides a reasonable error message with a link to the containers dash which has pointers to 
---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: This is a change to an interactive command which does not have tests yet.
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: No e2e exists for containers/cloudchamber.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: No behavior change.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: The containers subcommand is only present in v4.

### DIFF
--- a/.changeset/cold-aliens-learn.md
+++ b/.changeset/cold-aliens-learn.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Correctly handle 0 length responses to wrangler containers list.
--- a/packages/wrangler/src/containers/containers.ts
+++ b/packages/wrangler/src/containers/containers.ts
@@ -175,6 +175,18 @@ async function listCommandHandle(
 			return;
 		}
 
+		// If we don't get multiple applications for any reason exit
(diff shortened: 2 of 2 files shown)


petebacondarwin (approved):
I wasn't able to test this locally because the accounts I have available don't have Cloudchamber authorisation right now. But the code looks safe enough.

PR #9026, opened 2025-04-22
fix(miniflare): kv services name should be unique per persist path

edmundhung:
Fixes n/a.

This PR improves the KV store service name so that it could have several storage and object service with a different persist key.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: v4 only feature

### DIFF
--- a/.changeset/cool-socks-stay.md
+++ b/.changeset/cool-socks-stay.md
@@ -1,0 +1,7 @@
+---
+"miniflare": patch
+---
+
+fix: kv services name should be unique per persist path
+
+This fix ensures that the KV and Secret Store plugin will create a dedicated object and storage service when they have a different persist path.
--- a/packages/miniflare/src/plugins/kv/index.ts
+++ b/packages/miniflare/src/plugins/kv/index.ts
@@ -1,11 +1,7 @@
 import fs from "fs/promises";
 import SCRIPT_KV_NAMESPACE_OBJECT from "worker:kv/namespace";
 import { z } from "zod";
-import {
-	Service,
-	Worker_Binding,
-	Worker_Binding_DurableObjectNamespaceDesignator,
-} from "../../runtime";
+import { Service, Worker_Binding } from "../../runtime";
 import { PathSchema } from "../../shared";
 import { SharedBindings } from "../../workers";
 import {
@@ -15,6 +11,7 @@ import {
 	namespaceEntries,
 	namespaceKeys,
 	objectEntryWorker,
+	Persistence,
 	PersistenceSchema,
 	Plugin,
 	ProxyNodeBinding,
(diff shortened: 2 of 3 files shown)


woodybriggs:
this may fix https://github.com/cloudflare/workers-sdk/issues/9006 and solve my issues in https://github.com/cloudflare/workers-sdk/pull/9007

packages/miniflare/src/plugins/kv/index.ts
```diff
@@ -96,9 +99,23 @@ export const KV_PLUGIN: Plugin<
 	}) {
 		const persist = sharedOptions.kvPersist;
 		const namespaces = namespaceEntries(options.kvNamespaces);
+		const KV_OBJECT_SERVICE_NAME = addPersistPathSuffix(
+			`${KV_PLUGIN_NAME}:object`,
```
edmundhung, line 103: I have changed the service name from `kv:ns` to `kv:object:{persistPath}` just to avoid conflict with the namespace service name (`kv:ns:{namespace}`)

packages/miniflare/src/plugins/kv/index.ts
```diff
@@ -41,19 +38,25 @@ export const KVSharedOptionsSchema = z.object({
...
 ): options is SitesOptions {
 	return options.sitePath !== undefined;
 }
 
+function addPersistPathSuffix(
+	baseName: string,
+	persistPath: Persistence
+): string {
+	if (typeof persistPath === "string") {
+		return `${baseName}:${persistPath}`;
```
edmundhung, line 54: Is there any reserved characters for service name?

packages/miniflare/src/plugins/kv/index.ts
```diff
@@ -96,9 +99,23 @@ export const KV_PLUGIN: Plugin<
 	}) {
 		const persist = sharedOptions.kvPersist;
 		const namespaces = namespaceEntries(options.kvNamespaces);
+		const KV_OBJECT_SERVICE_NAME = addPersistPathSuffix(
+			`${KV_PLUGIN_NAME}:object`,
+			persist
+		);
+		const KV_STORAGE_SERVICE_NAME = addPersistPathSuffix(
```
pmiguel, line 106: Won't changing the persisPath break compatibility with existing local kv namespaces?

.changeset/cool-socks-stay.md
```diff
@@ -0,0 +1,7 @@
+---
+"miniflare": patch
+---
+
+fix: kv services name should be unique per persist path
+
+This fix ensures that the KV and Secret Store plugin will create a dedicated object and storage service when they have a different persist path.
```
pmiguel, line 7: Why do these two separate plugins colide? Is KV plugin a dependency?

woodybriggs: Secrets store uses KV in its underlying implementation when running in development mode.

See my issue #9006 and pr #9007 for full details.

Summary is that both kv and secrets store emit a kv:storage service and one of them gets deduped (the secrets store one). If they are named different then  they won't be deduped.

packages/miniflare/src/plugins/kv/index.ts
edmundhung: I don't think so. This doesn't change the service name of the namespace service but just the underlying object and storage service.

packages/miniflare/src/plugins/kv/index.ts
penalosa: url encode

packages/miniflare/src/plugins/kv/index.ts
pmiguel: Do you mind performing an hand-over-hand test where you write data into a namespace with the current wrangler build, and then try to read it with this version? I'm afraid this might change the underlying storage path, and data becomes unaccessible between versions. If that's the case, this might require a major bump.

edmundhung: Sure. Happy to give it a manual test to confirm the behavior. Definitely don't want to break existing users because of it. I will update here after the test.

edmundhung: @pmiguel Here is an example app that saves a random number in a KV namespace: https://github.com/edmundhung/kv-test

You can then update it to the pre-release version by running 
```sh
npm install --save-dev https://prerelease-registry.devprod.cloudflare.dev/workers-sdk/runs/14644497405/npm-package-wrangler-9026
```

The same number should be shown.

edmundhung:
Superseded by #9106

PR #9033, opened 2025-04-23
chore: replace most `wrangler.toml` files in monorepo with `wrangler.jsonc` ones

dario-piotrowicz:
This PR replaces most of the existing `wrangler.toml` files in the monorepo to `wrangler.jsonc` ones

I converted them using a quick script (included below) I wrote, I then spot checked the various files, adjusted what needed and copy-pasted comments when present (since my script doesn't account for those)

<details><summary>Script</summary>
<p>

```js
const { glob, readFileSync, rmSync, writeFileSync } = require('node:fs');
const TOML = require('@ltd/j-toml');

const absolutePathDir = '/Users/dario/Repos/workers-sdk';

glob(`${absolutePathDir}/**/*.toml`, (err, matches) => {
    if(err) throw err;

    for (const file of matches) {
        const fileContent = readFileSync(file, 'utf8');
        const config = TOML.parse(fileContent, 1, '', false);
        const outputFile = file.replace(/\.toml$/, '.jsonc');
        writeFileSync(outputFile, JSON.stringify(config, undefined, 2), 'utf8');
        rmSync(file);
    }

    console.log(`\x1b[35m converted: \x1b[0m`);
    console.log(matches);
    console.log('');
});
```

</p>
</details> 

> [!WARNING]
> This update is pretty comprehensive since the script should have caught all the toml file present, but it 
> does not include every possible toml reference in the monorepo, there are for example tests that write
> and read toml files which I did not update, this PR is quite massive already so I would prefer to land it
> as soon as possible as is and deal with the remaining references as a followup

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: chore change
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not something needing documentation
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a Wrangler change?


packages/create-cloudflare/templates/remix/workers/templates/wrangler.toml
petebacondarwin, line 1: How come this file was not converted?

packages/miniflare/test/index.spec.ts
```diff
@@ -1327,7 +1327,7 @@ test("Miniflare: modules in sub-directories", async (t) => {
 	t.is(await res.text(), "123");
 });
 
-test("Miniflare: python modules", async (t) => {
+test.skip("Miniflare: python modules", async (t) => {
```
petebacondarwin, line 1330: What?

packages/create-cloudflare/templates/remix/workers/templates/wrangler.toml
dario-piotrowicz: this was causing some C3 failures, instead of investigating and prolonging the time this PR needs to be open I thought it'd be better to just skip this and convert it as a followup, I'm happy to do it here if you think we should include such change here

packages/miniflare/test/index.spec.ts
dario-piotrowicz: oops, thanks a lot for noticing it, I mistakenly committed this 🙇

dario-piotrowicz: removed, thanks a lot for spotting it! 🙏

packages/create-cloudflare/src/workers.ts
```diff
@@ -52,7 +52,7 @@ export async function addWorkersTypesToTsConfig(ctx: C3Context) {
 	const typesEntrypoint = `@cloudflare/workers-types/${entrypointVersion}`;
 
 	try {
-		const config = jsonc.parse(tsconfig);
+		const config = jsonc.parse(tsconfig, [], { allowTrailingComma: true });
```
penalosa, line 55: Is this just an unrelated bugfix?

packages/vitest-pool-workers/CHANGELOG.md
```diff
@@ -1372,7 +1372,7 @@
   		poolOptions: {
   			workers: {
   				wrangler: {
-  					configPath: "./wrangler.toml",
+  					configPath: "./wrangler.jsonc",
```
penalosa, line 1375: We shouldn't be modifying the changelog like this. This can (and should) stay as `wrangler.toml`

packages/workers-shared/router-worker/wrangler.jsonc
```diff
@@ -0,0 +1,85 @@
+/*
+  Configuration file for the Router Worker
+
+  Please note that wrangler has a dependency on this file, and will
+  attempt to read it as part of setting up a new Miniflare instance
+  in developemnt mode. We should ensure that any configuration changes
```
penalosa, line 6:
```suggestion
  in development mode. We should ensure that any configuration changes
```

packages/vitest-pool-workers/CHANGELOG.md
dario-piotrowicz: sorry, this was just a search-and-replace unintentional change, thanks a bunch for spotting it, I'll revert it 👍

packages/create-cloudflare/src/workers.ts
dario-piotrowicz: yes, I think this is a valid change but I do think that I should not include it here

I think I just added it when I was hitting issues with trailing comma issues before, but after I removed all the create-cloudflare toml->jsonc changes so this should not be needed in this PR anymore, I'll remove this here and see if CI is happy, I'll also reintroduce this in a followup PR if that works for you 🙂

packages/vitest-pool-workers/CHANGELOG.md
dario-piotrowicz: removed 🙂

packages/workers-shared/package.json
```diff
@@ -58,7 +58,7 @@
 	},
 	"devDependencies": {
 		"@cloudflare/eslint-config-worker": "workspace:*",
-		"@cloudflare/vitest-pool-workers": "^0.5.41",
+		"@cloudflare/vitest-pool-workers": "^0.7.0",
```
CarmenPopoviciu, line 61: why not `^0.8.19` like in all the other changes?

CarmenPopoviciu (approved):

packages/workers-shared/package.json
dario-piotrowicz: I did do that but I had to downgrade it (see https://github.com/cloudflare/workers-sdk/pull/9033/commits/a706a0a4ab05e88c57b4c59120aa03321894a569) as the newer version was causing some wrangler incompatibility issues

I figured that it'd be better not to scope creep things here and just dump the dependency to the higher minor it could go

packages/create-cloudflare/templates/remix/workers/templates/wrangler.toml
dario-piotrowicz: I was looking into this I think that the issue is that the remix workers template itself uses a toml file: https://github.com/remix-run/remix/tree/main/templates/cloudflare-workers

This is problematic because we then end up with two different wrangler config files and specifically only the toml one is updated ([code](https://github.com/cloudflare/workers-sdk/blob/1be56441dec5e8faea3b6c007c1a347bab1a4029/packages/create-cloudflare/e2e-tests/frameworks.test.ts#L250-L256)), so this would need to be fixed upstream first, I'm happy to open a PR in the remix repo, although given their move to react-router I don't really know this would be worth the effort, WDYT?

petebacondarwin: OK, so we have to take the lead from the framework I guess.
One thing we could do is in our `c3.ts` code, replace the wrangler.toml that is generated with our own one. Or we could just leave this as-is until they get round to updating their template.

packages/create-cloudflare/src/workers.ts
dario-piotrowicz: I've been looking into this and maybe this change is not as valid as I thought it was. The C3 wrangler files are not part of the monorepo's formatting: https://github.com/cloudflare/workers-sdk/blob/d34ef3d5b0e45470e6c06a7bd24a498f85b98af8/.prettierignore#L16-L19 so they don't necessarily need to include trailing commas, including them in the monorepo formatting doesn't seem really necessary, so for the time being I think that it's fine keeping the code as is and update this later when/if needed 🙂

PR #9044, opened 2025-04-24
Add support for `run_worker_first` flag in Miniflare options

### LINKED ISSUE #8430: 🐛 BUG: @cloudflare/vite-plugin ignore `run_worker_first`

Vite Plugin

0.1.9 [@cloudflare/vite-plugin]

Mac Sequoia 15.3.1

_No response_

### Describe the Bug

Using the `create cloudflare@latest --template react-swc-ts` and adding the following `assets` config to the wrangler.jsonc
```
"assets": {
        "not_found_handling": "single-page-application",
	"binding": "ASSETS",
	"run_worker_first": true
	},
```

Then removing the following line from the worker
```
return Response.json({
      message: "Hello world",
    });

    // return env.ASSETS.fetch(request);
```

You will find it still continues to serve the index.html rather than returning "Hello world"

_No response_

0xbad0c0d3:
This update introduces support for the `run_worker_first` flag when configuring Miniflare options. A new test suite ensures proper handling of this flag, including scenarios where it is undefined. Additionally, the `miniflare-options` logic was updated to pass the flag to the worker configuration when specified.

Fixes #8430.

---

- Tests
  - [ ] TODO (before merge)
  - [X ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


petebacondarwin:
Hey thanks for submitting this change @0xbad0c0d3. Unfortunately I think it is a bit more involved to make this work fully.
At the moment Vite will handle all non-HTML assets before it ever gets to the Worker, so whatever the configuration of Miniflare, the Worker will not get to intercept such requests.

I have a draft Worker that solves this problem more fully - https://github.com/cloudflare/workers-sdk/pull/8494 - but due to the complexity it adds to the implementation, we want land it behind a flag. I hope to have that ready (at least the run_worker_first bits) by the end of this week.

0xbad0c0d3:
@petebacondarwin sorry. I didn't get your words:

> At the moment Vite will handle all non-HTML assets before it ever gets to the Worker

Isn't it must be so? How other way the worker will know about request? If vite-server is the entry-point of the request...
Sorry for the, probably, dummy question ))

petebacondarwin:
The Vite dev server is basically like an express server that has middleware. The Worker gets triggered by a middleware at the very end of the chain after all the other middlewares have had a chance to deal with the request. We have turned off most of the built in middleware so that we can handle HTML requests. But there is still middleware that will identify requests that match assets (either static or generated client side code) and will just respond with that content instead of passing it through to the other middlewares, such as the one that triggers the Worker.

petebacondarwin:
Actually, I was just going through my bigger PR and I realised that just adding this to miniflare options is enough to get it working in `vite preview` mode (but not `vite dev` mode). I have a playground fixture in my PR that tests this so I will push a commit here and I think we can land this PR, with a few tweaks, to support in preview mode at least.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -359,8 +359,13 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 
 				await initRunners(resolvedPluginConfig, viteDevServer, miniflare);
 
+				const viteHandle =
+					/^https?:\/\/[^/]+\/(?:(?:src|node_modules[^\/]*?\/)|@\w+)|\?import$/i;
```
petebacondarwin, line 363: I don't believe this is sufficient to handle all the scenarios where the Worker should not intercept the request, and even if it worked now (which I don't think it would) it would be brittle to future changes to Vite and also to other plugins that are used alongside this one.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -371,18 +376,14 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					{ alwaysCallNext: false }
 				);
 
+				viteDevServer.middlewares.use(middleware);
```
petebacondarwin, line 379: Following on from my comment on the passthrough above, I don't believe we can do this here. See my other PR for how we think we an solve this more correctly (although still with potential dangers when interoperating with other plugins) - https://github.com/cloudflare/workers-sdk/pull/8494

petebacondarwin (changes requested):
I don't think we can solve the `vite dev` mode issue so simply. I propose we change this PR to only add support for `vite preview`, which is quite straightforward, and then we can look into `vite dev` separately.

petebacondarwin:
Pushing a commit now...

petebacondarwin:
This tweaks the way we update the miniflare config to be a bit simpler, and adds back in the fixture but only tests it in `vite preview` mode. I also tweaked the changeset to be more user facing and less implementation oriented.

packages/vite-plugin-cloudflare/src/__tests__/get-miniflare-options.spec.ts
petebacondarwin, line 1: The playground tests cover this behaviour, so this unit tests doesn't really add any value.
Further it requires a bunch of type hacks because it doesn't follow the exact requirements of the params passed to `getMiniflareOptions()`.
So I propose we just remove it, which I'll do in a commit now.

petebacondarwin (approved):

jamesopstad (changes requested):

packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -219,6 +219,8 @@ export function getDevMiniflareOptions(
 			bindings: {
 				CONFIG: {
 					has_user_worker: resolvedPluginConfig.type === "workers",
+					invoke_user_worker_ahead_of_assets:
+						assetsConfig?.run_worker_first ?? false,
```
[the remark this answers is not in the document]
petebacondarwin: Oh... let me check. I thought this was needed. Sorry.

PR #9048, opened 2025-04-24
fix: Validate input file for Vectorize inserts

garvit-gupta:
This PR adds logic to validate input files for Vectorize inserts/upserts. Only valid file paths, with non-empty files would be accepted with this change. Since the validation check only checks the file metadata, and does not read/parse the file, it should have no impact on command performance/resource consumption.

Context: As of now, customers can supply invalid/empty files as part of the Vectorize insert/upsert commands in Wrangler. The Vectorize commands display an error while parsing the file, but then proceed with processing an invalid input file (the command then behaves like a no-op). This PR improves user experience by carrying out a validation check before parsing the file, returns an error message on finding an invalid file, and also returns early. 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: This PR just adds a validation check to existing commands.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/9093
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/vectorize/insert.ts
+++ b/packages/wrangler/src/vectorize/insert.ts
@@ -7,6 +7,7 @@ import { insertIntoIndex, insertIntoIndexV1 } from "./client";
 import {
 	deprecatedV1DefaultFlag,
 	getBatchFromFile,
+	isValidFile,
 	VECTORIZE_MAX_BATCH_SIZE,
 	VECTORIZE_MAX_UPSERT_VECTOR_RECORDS,
 	VECTORIZE_UPSERT_BATCH_SIZE,
@@ -54,6 +55,11 @@ export const vectorizeInsertCommand = createCommand({
 	},
 	positionalArgs: ["name"],
 	async handler(args, { config }) {
+		if (!(await isValidFile(args.file))) {
+			logger.error(`🚨 Cannot read invalid or empty file: ${args.file}.`);
+			return;
+		}
+
 		const rl = createInterface({ input: createReadStream(args.file) });
 
 		if (
--- a/packages/wrangler/src/vectorize/insert.ts
+++ b/packages/wrangler/src/vectorize/insert.ts
@@ -2,6 +2,7 @@ import { createReadStream } from "node:fs";
 import { createInterface } from "node:readline";
 import { File, FormData } from "undici";
 import { createCommand } from "../core/create-command";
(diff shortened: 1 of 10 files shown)


sejoker (approved):

packages/wrangler/src/vectorize/insert.ts
```diff
@@ -54,6 +55,11 @@ export const vectorizeInsertCommand = createCommand({
 	},
 	positionalArgs: ["name"],
 	async handler(args, { config }) {
+		if (!(await isValidFile(args.file))) {
+			logger.error(`🚨 Cannot read invalid or empty file: ${args.file}.`);
```
CarmenPopoviciu, line 59: should this throw a `UserError` instead?

packages/wrangler/src/vectorize/upsert.ts
```diff
@@ -45,6 +46,11 @@ export const vectorizeUpsertCommand = createCommand({
 	},
 	positionalArgs: ["name"],
 	async handler(args, { config }) {
+		if (!(await isValidFile(args.file))) {
+			logger.error(`🚨 Cannot read invalid or empty file: ${args.file}.`);
```
CarmenPopoviciu, line 50: should this throw a `UserError` instead?
```
if (!(await isValidFile(args.file))) {
  throw new UserError("Could not read invalid or empty file: ${args.file}");
}
```

garvit-gupta: Now throwing a UserError for this. The latest commit now updates the behavior for all Vectorize commands to throw UserErrors instead of logging errors and returning early.

packages/wrangler/src/vectorize/insert.ts
garvit-gupta: Now throwing a UserError for this. The latest commit now updates the behavior for all Vectorize commands to throw UserErrors instead of logging errors and returning early.

CarmenPopoviciu:
@penalosa as next on-call, can you please review the backport PR 🙏 ? I can't review myself since I created it

PR #9096, opened 2025-04-29
Backport #9095 -  wrangler login put custom callback host and port into the auth URL

petebacondarwin:
Backport of #9095


petebacondarwin:
Oops. Just realised that the `callbackHost` and `callbackPort` were only added in V4, so no need for a backport.

PR #9098, opened 2025-04-29
Migrate Workers Containers commands to Containers API Endpoints

jseba:
Workers Containers

The Workers Containers API has new endpoints and auth tokens to enforce its scope. This change moves the Wrangler interaction with the API to those new endpoints and uses the new token for them.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included: no new tests added, existing workers container E2E tests should continue functioning as-is
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Internal change to API endpoints
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/wrangler/src/containers/index.ts
```diff
@@ -26,35 +30,41 @@ export const containers = (
...
 			"push a tagged image to a Cloudflare managed registry, which is automatically integrated with your account",
 			(args) => pushYargs(args),
-			(args) => handleFailure(pushCommand)(args)
+			(args) => handleFailure(pushCommand, containersScope)(args)
+		)
+		.command(
+			"apply",
+			"apply the changes in the container applications to deploy",
+			(args) => applyCommandOptionalYargs(args),
+			(args) => handleFailure(applyCommand, containersScope)(args)
```
penalosa, line 45: Why is this PR adding a new command?

jseba: ah I think that's from when I using an in-progress branch as the base for these changes, I'll remove that

packages/wrangler/src/user/user.ts
```diff
@@ -360,6 +360,7 @@ const DefaultScopes = {
 
 const OptionalScopes = {
 	"cloudchamber:write": "Manage Cloudchamber",
+	"containers:write": "Manage Workers Containers",
```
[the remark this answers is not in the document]
jseba: I think that's the plan once we're out of private beta afaik

.changeset/rude-kings-teach.md
```diff
@@ -0,0 +1,8 @@
+---
+"wrangler": minor
```
petebacondarwin: Since there is no new behaviour here, it is just an internal refactor, then this should be

```suggestion
"wrangler": patch
```

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -177,16 +196,14 @@ export async function fillOpenAPIConfiguration(config: Config, json: boolean) {
 	}
 
 	const scopes = getScopes();
-	const needsCloudchamberToken = !scopes?.find(
-		(scope) => scope === "cloudchamber:write"
-	);
-	const cloudchamberScope: Scope[] = ["cloudchamber:write"];
+	const needsToken = !scopes?.find((s) => s === scope);
```
petebacondarwin: Using `.some()` returns a boolean already.

```suggestion
	const needsToken = !scopes?.some((s) => s === scope);
```

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -177,16 +196,14 @@ export async function fillOpenAPIConfiguration(config: Config, json: boolean) {
...
+	const neededScopes: Scope[] = [scope];
 	const scopesToSet: Scope[] =
 		scopes == undefined
-			? cloudchamberScope.concat(DefaultScopeKeys)
-			: cloudchamberScope.concat(scopes);
+			? neededScopes.concat(DefaultScopeKeys)
+			: neededScopes.concat(scopes);
```
petebacondarwin: It is possible that `scopes` already contains `scope` so this could result in a duplication. Is that what you want?

petebacondarwin (commented):
LGTM - can you change the changeset and then happy to approve.

packages/wrangler/src/cloudchamber/common.ts
jseba: added in update

.changeset/rude-kings-teach.md
jseba: gotcha, reflected in the update

packages/wrangler/src/cloudchamber/common.ts
jseba: this is kinda convoluted but it should be that if the token has the scope already (`needsToken == false`) then the re-auth calls won't happen

packages/wrangler/src/user/user.ts
```diff
@@ -361,23 +361,16 @@ const DefaultScopes = {
...
-	...OptionalScopes,
-};
-
 /**
  * The possible keys for a Scope.
  *
  * "offline_access" is automatically included.
  */
-export type Scope = keyof typeof AllScopes;
+export type Scope = keyof typeof DefaultScopes;
```
CarmenPopoviciu, line 373: just double checking...so `"cloudchamber:write"` is intentionally excluded from this scopes list correct?

packages/containers-shared/src/client/core/request.ts
```diff
@@ -197,6 +213,13 @@ const getRequestBody = (options: ApiRequestOptions): any => {
 	return undefined;
 };
 
+const isResponseSchemaV4 = (
+	config: OpenAPIConfig,
+	_options: ApiRequestOptions
```
vicb: remove unused `_options`?

packages/containers-shared/src/client/core/request.ts
```diff
@@ -319,16 +342,28 @@ export const request = <T>(
...
-				catchErrorCodes(options, result);
+				const result: ApiResult = isResponseSchemaV4(config, options)
+					? parseResponseSchemaV4(responseBody)
+					: {
+							url,
+							ok: response.ok,
+							status: response.status,
+							statusText: response.statusText,
+							body: responseHeader ?? responseBody,
+						};
```
vicb: This code will create a function (`parseResponseSchemaV4`) each time it is executed. Also IMO it's harder to read than just:

```
let result: ApiResult;

if (isResponseSchemaV4(config)) {
  result = ...; // inline your code
} else {
  result = { /* l359-363 */}
}
```

vicb (approved):

packages/wrangler/src/user/user.ts
jseba: it should be included, @penalosa removed the split `AllScopes` and `OptionalScopes` and it's just `DefaultScopes` which includes both tokens now

packages/containers-shared/src/client/core/request.ts
jseba: ah, I'm used to other languages where if there's no closure it won't recreate the function body every invocation and statically compile it, I'll make it a free function instead

I can go whichever way with it, the way it's written is more rust-like to me but I don't know idiomatic JS/TS

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -49,8 +39,17 @@ import type {
 	UserDeploymentConfiguration,
 } from "@cloudflare/containers-shared";
 
+export const cloudchamberScope = "cloudchamber:write" as const;
+
 export type CommonCloudchamberConfiguration = { json: boolean };
 
+const containerIdRe = /[^/]{36}/;
```
CarmenPopoviciu:
```suggestion
const containerIdRegexp = /[^/]{36}/;
```

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -49,8 +39,17 @@ import type {
...
+
+export function isValidContainerID(value: string): boolean {
+	const matches = value.match(containerIdRe);
```
CarmenPopoviciu:
```suggestion
	const matches = value.match(containerIdRegexp);
```

CarmenPopoviciu (approved):
re-approving

PR #9106, opened 2025-04-30
fix(miniflare): decouple KV plugin from Secrets Store plugin to avoid service name conflicts

### LINKED ISSUE #9006: vite-plugin: secrets store bindings throws Secret "[SOME_SECRET_KEY]" not found

edmundhung:
Fixes #9006.

We were relying on the KV plugin to setup the object and storage service for the internal KV namespaced used by the local sercets store binding. However, this creates an issue when the users have both a KV namespace and secret binding defined with each creating a storage service sharing the same name while having a different persist path.

This decouples kv plugin from the sercets store plugin by copying the implementation in the kv bindings with the service named customzied for the secret store plugin.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: v4 feature

### DIFF
--- a/packages/wrangler/e2e/dev-with-resources.test.ts
+++ b/packages/wrangler/e2e/dev-with-resources.test.ts
@@ -370,6 +370,10 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 				name = "${workerName}"
 				main = "src/index.ts"
 				compatibility_date = "2025-01-01"
+				# Regression test for https://github.com/cloudflare/workers-sdk/issues/9006
+				kv_namespaces = [
+					${isLocal ? `{ binding = "KV", id = "LOCAL_ONLY" }` : ""}
+				]
 				secrets_store_secrets = [
 					{ binding = "SECRET", store_id = "${storeId}", secret_name = "${secret_name}" }
 				]
--- a/.changeset/silver-bats-brush.md
+++ b/.changeset/silver-bats-brush.md
@@ -1,0 +1,7 @@
+---
+"miniflare": patch
+---
+
+fix: decouple KV plugin from secrets store plugin
+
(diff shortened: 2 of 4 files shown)


dario-piotrowicz (approved):
Looks good to me 🙂 

I think this could use a test where you do check that users can use the secrets store alongside a KV binding, but maybe that's a bit of an overkill 🤔

packages/wrangler/e2e/dev-with-resources.test.ts
```diff
@@ -370,6 +370,10 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 				name = "${workerName}"
 				main = "src/index.ts"
 				compatibility_date = "2025-01-01"
+				# Regression test for https://github.com/cloudflare/workers-sdk/issues/9006
+				kv_namespaces = [
+					${isLocal ? `{ binding = "KV", id = "LOCAL_ONLY" }` : ""}
+				]
```
edmundhung, line 376: I ran the same test without the fix here: https://github.com/cloudflare/workers-sdk/actions/runs/15183927642/job/42699817229?pr=9338#step:5:4996

PR #9108, opened 2025-04-30
Support running Vite in middleware mode

jamesopstad:
Fixes https://github.com/cloudflare/workers-sdk/issues/8892

Adds support for running Vite in middleware mode. This enables using Storybook with the Vite plugin, which would previously crash. WebSocket connections to Workers are not supported when in middleware mode.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/social-baboons-jog.md
+++ b/.changeset/social-baboons-jog.md
@@ -1,5 +1,5 @@
 ---
-"@cloudflare/vite-plugin": patch
+"@cloudflare/vite-plugin": minor
 ---
 
 Add support for running Vite in middleware mode. This enables using Storybook with the Vite plugin, which would previously crash. WebSocket connections to Workers are not supported when in middleware mode.
--- a/.changeset/social-baboons-jog.md
+++ b/.changeset/social-baboons-jog.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+Add support for running Vite in middleware mode. This enables using Storybook with the Vite plugin, which would previously crash. WebSocket connections to Workers are not supported when in middleware mode.
--- a/packages/vite-plugin-cloudflare/playground/vitest-setup.ts
+++ b/packages/vite-plugin-cloudflare/playground/vitest-setup.ts
@@ -181,7 +181,11 @@ beforeAll(async (s) => {
 }, 15_000);
 
 beforeEach(async () => {
-	await page.goto(viteTestUrl);
+	try {
+		await page.goto(viteTestUrl);
+	} catch (error) {
+		// Expected to error for middleware-mode test as `viteTestUrl` will not be valid
+	}
 });
 
 export async function loadConfig(configEnv: ConfigEnv) {
--- a/packages/vite-plugin-cloudflare/playground/vitest-setup.ts
+++ b/packages/vite-plugin-cloudflare/playground/vitest-setup.ts
@@ -153,6 +153,7 @@ beforeAll(async (s) => {
 				if (serve) {
 					server = (await serve()) ?? server;
 					viteServer = mod.viteServer ?? viteServer;
+					viteTestUrl = mod.viteTestUrl ?? viteTestUrl;
 				}
 			} else {
 				await startDefaultServe();
@@ -181,11 +182,7 @@ beforeAll(async (s) => {
 }, 15_000);
 
 beforeEach(async () => {
-	try {
-		await page.goto(viteTestUrl);
-	} catch (error) {
-		// Expected to error for middleware-mode test as `viteTestUrl` will not be valid
-	}
+	await page.goto(viteTestUrl);
 });
 
 export async function loadConfig(configEnv: ConfigEnv) {
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -366,6 +366,7 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					{ alwaysCallNext: false }
 				);
 
+				// The HTTP server is not available in middleware mode
 				if (viteDevServer.httpServer) {
 					handleWebSocket(viteDevServer.httpServer, async () => {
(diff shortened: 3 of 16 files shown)


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -371,12 +366,14 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+				if (viteDevServer.httpServer) {
+					handleWebSocket(viteDevServer.httpServer, async () => {
+						assert(miniflare, `Miniflare not defined`);
+						const routerWorker = await getRouterWorker(miniflare);
 
-					return routerWorker.fetch;
-				});
+						return routerWorker.fetch;
+					});
+				}
```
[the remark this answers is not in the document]
dario-piotrowicz: could have show the warning if the application actually tries to establish a websocket connection?

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
```diff
@@ -181,7 +181,11 @@ beforeAll(async (s) => {
 }, 15_000);
 
 beforeEach(async () => {
-	await page.goto(viteTestUrl);
+	try {
+		await page.goto(viteTestUrl);
+	} catch (error) {
+		// Expected to error for middleware-mode test as `viteTestUrl` will not be valid
+	}
```
dario-piotrowicz, line 188: could we instead avoid running the `page.goto` in the first place for the middleware-mode app?

jamesopstad: Is there a way to opt out of the `beforeEach` in a single test file?

packages/vite-plugin-cloudflare/src/index.ts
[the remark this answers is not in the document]
jamesopstad: The issue is that Storybook (and probably other similar tools that use middleware mode) create their own WebSockets and we can't tell if these are created by the user or not. WebSockets do work more generally, but not ones intended for Workers.

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
dario-piotrowicz: I was thinking of something along these lines:
```js
beforeEach(async () => {
	if (rootDir.includes("middleware-mode")) {
	    // there is no `ivteTestUrl` to navigate to for middleware-mode
		return;
	}
	
	await page.goto(viteTestUrl);
});
```

it is a bit hardcoded but it might be better than swallowing all potential exceptions here? 🤔

dario-piotrowicz (approved):

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
jamesopstad: Ah, nice idea. I'll try that.

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
```diff
@@ -181,7 +181,11 @@ beforeAll(async (s) => {
 }, 15_000);
 
 beforeEach(async () => {
-	await page.goto(viteTestUrl);
+	try {
+		await page.goto(viteTestUrl);
+	} catch (error) {
+		// Expected to error for middleware-mode test as `viteTestUrl` will not be valid
```
petebacondarwin, line 187: Could we make `viteTestUrl` valid somehow?

packages/vite-plugin-cloudflare/src/index.ts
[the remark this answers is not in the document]
petebacondarwin: Perhaps a debug log?

petebacondarwin (approved):
Sweet!

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
jamesopstad: I found a way to do this by exporting `viteTestUrl` from the custom serve file and then capturing that. Much nicer now.

[f469040](https://github.com/cloudflare/workers-sdk/pull/9108/commits/f469040092e02c0e1fc93cacc1b526a17c46eb64)

.changeset/social-baboons-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
```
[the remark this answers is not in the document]
jamesopstad: [6f3b74f](https://github.com/cloudflare/workers-sdk/pull/9108/commits/6f3b74f49b5095bc58c8b2e049aec4829b66ab7b)

packages/vite-plugin-cloudflare/src/index.ts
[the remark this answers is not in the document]
jamesopstad: I don't think Vite has any built in way for plugins to add debug logs. Internally they use [debug](https://www.npmjs.com/package/debug) so we could potentially import that and use it too. I wonder if it would make sense for Vite to export their [createDebugger](https://github.com/vitejs/vite/blob/b040d547a17c4bfe8aba44534228667a50612318/packages/vite/src/node/utils.ts#L199) utility for plugins to use.

For now, I've added an extra comment to clarify the intention in case we want to add debug logging in future.

[d28a115](https://github.com/cloudflare/workers-sdk/pull/9108/commits/d28a11529426f006ef310cda421c9d2c6c131cd2)

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
jamesopstad: See https://github.com/cloudflare/workers-sdk/pull/9108#discussion_r2069988230

PR #9112, opened 2025-04-30
Bump to Node.js 20

penalosa:
Following our [Node.js version support policy](https://developers.cloudflare.com/workers/wrangler/install-and-update/), Wrangler no longer supports running with Node.js 18 now that it's EOL. This PR updates workers-sdk to use `@types/node@20`, run CI against Node 20, and updates Wrangler to fail if run with Node 18.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Existing tests should pass
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Already documented
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a backportable change

### DIFF
--- a/packages/wrangler/bin/wrangler.js
+++ b/packages/wrangler/bin/wrangler.js
@@ -2,15 +2,15 @@
 const { spawn } = require("child_process");
 const path = require("path");
 
+const ERR_NODE_VERSION = "18.0.0";
 const MIN_NODE_VERSION = "20.0.0";
-
 let wranglerProcess;
 
 /**
  * Executes ../wrangler-dist/cli.js
  */
 function runWrangler() {
-	if (semiver(process.versions.node, MIN_NODE_VERSION) < 0) {
+	if (semiver(process.versions.node, ERR_NODE_VERSION) < 0) {
 		// Note Volta and nvm are also recommended in the official docs:
 		// https://developers.cloudflare.com/workers/get-started/guide#2-install-the-workers-cli
 		console.error(
--- a/packages/wrangler/bin/wrangler.js
+++ b/packages/wrangler/bin/wrangler.js
@@ -2,7 +2,7 @@
 const { spawn } = require("child_process");
 const path = require("path");
 
-const MIN_NODE_VERSION = "18.0.0";
+const MIN_NODE_VERSION = "20.0.0";
 
 let wranglerProcess;
 
--- a/.changeset/young-trainers-do.md
+++ b/.changeset/young-trainers-do.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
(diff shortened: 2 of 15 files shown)


packages/wrangler/bin/wrangler.js
```diff
@@ -2,7 +2,7 @@
 const { spawn } = require("child_process");
 const path = require("path");
 
-const MIN_NODE_VERSION = "18.0.0";
+const MIN_NODE_VERSION = "20.0.0";
```
petebacondarwin, line 5: I am not sure we should hard fail on node 18 quite yet. How about a big warning for the next few months?

penalosa: I've made it a warning in [612e8f8](https://github.com/cloudflare/workers-sdk/pull/9112/commits/612e8f887af648c2b58ad26ca931b85c2c87d4dd). The warning is in the Wrangler banner (rather than `wrangler.js`) so that it doesn't get shown on commands that need clean JSON output.

packages/wrangler/bin/wrangler.js
```diff
@@ -2,15 +2,15 @@
...
+const ERR_NODE_VERSION = "18.0.0";
+const MIN_NODE_VERSION = "20.0.0";
 let wranglerProcess;
 
 /**
  * Executes ../wrangler-dist/cli.js
  */
 function runWrangler() {
-	if (semiver(process.versions.node, MIN_NODE_VERSION) < 0) {
+	if (semiver(process.versions.node, ERR_NODE_VERSION) < 0) {
```
dario-piotrowicz, line 13: This change doesn't seem correct to me, is it?

Why changing this to `ERR_NODE_VERSION`?

penalosa: I'm changing this to only _error_ if below 18, but tell people to upgrade to 20 in the messaging

dario-piotrowicz: I'm not sure I'm following this... so if I were to use node 19 (which we don't support), why wouldn't we want to error here?

(also, is `< 0` correct? because this would not show the error if I were to use node precisely `18.0.0`)

penalosa: This is a tempory measure for the next few months to avoid being too loud about version support changes. We only stopped supporting Node.js 18 a couple of days ago, and this is the first change in Node.js version support we've made with the new v4 support guarantee.

PR #9115, opened 2025-05-01
Don't crash on non-existent tails in Vite

penalosa:
In Vite, only connect the tail consumers that represent Workers that are defined in the Vite config. Warn that a tail will be omitted otherwise. This _differs from service bindings_ because tail consumers are "optional" in a sense, and shouldn't affect the runtime behaviour of a Worker. 

Fixes issues like https://github.com/getsentry/sentry-mcp/pull/134#issuecomment-2842683985
	

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: covered by playgrounds
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: fix to a feature introduced in v4

### DIFF
--- a/packages/vite-plugin-cloudflare/src/miniflare-options.ts
+++ b/packages/vite-plugin-cloudflare/src/miniflare-options.ts
@@ -211,7 +211,7 @@ function filterTails(
 			// Don't interfere with network-based tail connections (e.g. via the dev registry), or kCurrentWorker
 			return true;
 		}
-		const found = !!userWorkers.find((w) => w.name === name);
+		const found = userWorkers.some(w => w.name === name);
 
 		if (!found) {
 			log(
--- a/packages/vite-plugin-cloudflare/src/miniflare-options.ts
+++ b/packages/vite-plugin-cloudflare/src/miniflare-options.ts
@@ -9,6 +9,7 @@ import {
 	LogLevel,
 	Response as MiniflareResponse,
 } from "miniflare";
+import colors from "picocolors";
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-cloudflare/src/miniflare-options.ts
```diff
@@ -189,6 +190,43 @@ function getEntryWorkerConfig(
...
+			typeof tailService === "object" &&
+			"name" in tailService &&
+			typeof tailService.name === "string"
+		) {
+			name = tailService.name;
+		} else {
+			// Don't interfere with network-based tail connections (e.g. via the dev registry), or kCurrentWorker
+			return true;
+		}
+		const found = !!userWorkers.find((w) => w.name === name);
```
jamesopstad, line 214:
```suggestion
		const found = userWorkers.some(w => w.name === name);
```

jamesopstad (approved):
Looks good. Does this need a changeset?

packages/vite-plugin-cloudflare/src/miniflare-options.ts
jamesopstad: Sorry, my bad. I think the checks are failing because I put `w` instead of `(w)`.

PR #9120, opened 2025-05-01
Add `experimental_startMixedModeSession` no-op utility

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1836

Add `experimental_startMixedModeSession` no-op utility

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no-op implementation for the experimental utility[^1]
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no-op implementation for the experimental utility
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: unreleased experimental feature
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: experimental new feature

[^1]: I considered adding some tests to make sure that the values returned by the utility are those that we expect, but I figured that the correctness of the returned value is already guaranteed by the utility's types so I didn't think that adding tests for the utility at this stage would really add any value here

### DIFF
--- a/packages/wrangler/src/api/mixedMode/index.ts
+++ b/packages/wrangler/src/api/mixedMode/index.ts
@@ -1,9 +1,13 @@
 import path from "node:path";
 import { getBasePath } from "../../paths";
+import { requireApiToken, requireAuth } from "../../user";
 import { startWorker } from "../startDevWorker";
 import type { StartDevWorkerInput, Worker } from "../startDevWorker/types";
 
-type MixedModeSession = Pick<Worker, "ready" | "setConfig"> & {
+type BindingsOpt = StartDevWorkerInput["bindings"];
+
+type MixedModeSession = Pick<Worker, "ready" | "dispose"> & {
+	["setConfig"]: (bindings: BindingsOpt) => Promise<void>;
 	["mixedModeConnectionString"]: MixedModeConnectionString;
 };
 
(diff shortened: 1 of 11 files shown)


packages/wrangler/src/api/mixedMode/index.ts
```diff
@@ -0,0 +1,35 @@
...
+		entrypoint: proxyServerWorkerEntrypoint,
+		dev: {
+			remote: true,
+		},
+		bindings,
+	});
+
+	return {
+		ready,
+		setConfig,
```
penalosa, line 32: This shouldn't be the actual `setConfig()`, it should be a function that only accepts the `bindings` subset of `setConfig()`

penalosa (commented):
Could you add a fixture/e2e test for this? Something along the lines of calling `startMixedModeSession()` and then verifying that a fetch to the url returns the right response?

packages/wrangler/src/api/mixedMode/index.ts
dario-piotrowicz: oh right! good call, sorry I missed that! 😓👍

dario-piotrowicz:
> Could you add a fixture/e2e test for this? Something along the lines of calling `startMixedModeSession()` and then verifying that a fetch to the url returns the right response?

good call, sure I'm on it 👍

packages/wrangler/src/api/mixedMode/index.ts
dario-piotrowicz: addressed 🙂

dario-piotrowicz:
> Could you add a fixture/e2e test for this? Something along the lines of calling `startMixedModeSession()` and then verifying that a fetch to the url returns the right response?

Added 🙂👍

That was a nice suggestion, it made me notice that I was missing a few bits in the `startMixedModeSession()` implementation, thanks! 😄

PR #9125, opened 2025-05-02
Remove unused `capnp-ts` patch

penalosa:
We no longer depend on `capnp-ts`, so remove the stale patch


anonrig (approved):

vicb:
@penalosa could you please update the lock file

anonrig (approved):

PR #9126, opened 2025-05-02
V3 Backport [#9124]: make that `unstable_startWorker` can correctly throw configuration errors

dario-piotrowicz:
v3 backport for https://github.com/cloudflare/workers-sdk/pull/9124

### DIFF
--- a/.changeset/metal-clubs-vanish.md
+++ b/.changeset/metal-clubs-vanish.md
@@ -1,0 +1,14 @@
+---
+"wrangler": patch
+---
+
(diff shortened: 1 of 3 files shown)


penalosa:
I don't think we need to backport this, as it's an experimental API?

dario-piotrowicz:
Yeah good call @penalosa , sorry I didn't think of SDW as being experimental 🤔 (which clearly is given its full name 😅) 

I'm happy not to backport this 👍

PR #9147, opened 2025-05-06
[c3] Add the `global_fetch_strictly_public` flag to workers templates

vicb:
`global_fetch_strictly_public` allows fetching from the worker - it makes sense to use it when there is no origin server which is the case for frameworks.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Update the compatibility flags
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: The flag is documented
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: c3 templates


packages/create-cloudflare/src/templates.ts
```diff
@@ -132,7 +132,7 @@ export type TemplateConfig = {
 		ctx: C3Context,
 	) => Promise<Record<string, string | object>>;
 
-	/** An array of compatibility flags to be specified when deploying to pages or workers.*/
+	/** An array of compatibility flags to be specified when deploying to pages (unused for workers) */
```
petebacondarwin, line 135: Good spot. I see that a couple of other Workers based c3.ts files still include this property. We should remove those as well.
E.g. 
- packages/create-cloudflare/templates-experimental/solid/c3.ts
- packages/create-cloudflare/templates/hello-world-with-assets/c3.ts

petebacondarwin (changes requested):
Can you clean up the last of the Workers that use that property too.

packages/create-cloudflare/src/templates.ts
petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/9210

petebacondarwin (approved):

Cherry:
As someone who uses C3 to bootstrap Workers for both origin and origin-less setups, and almost always uses the default 0Hello World example (which today is with Assets), enabling this compat flag by default seems really interesting.

Cloudflare has historically enabled new compat flags by default after a certain date, but never in a manner like this (outside of node compat) for projects bootstrapped via C3 to my knowledge. Looking at https://github.com/cloudflare/workers-sdk/tree/main/packages/create-cloudflare/templates, I see it was enabled in a few frameworks but not all of them, so was just wondering how this is planned to be enabled moving forward if there's any insight you could share?

petebacondarwin:
Thanks for the comment @Cherry.

I believe only the Pages templates did not have this flag added, since Pages already has customized handling that effectively makes this flag unnecessary. If we missed any Workers templates please let us know.

Regarding flag handling, it is true that we often have a compat date when flags become on by default, but this is not mandatory and there are a few flags that have always been deemed as optional user choice: `nodejs_compat`, `nodejs_als`, `queue_consumer_no_wait_for_wait_until`, `disallow_importable_env`, `request_signal_passthrough`.

I don't think the `global_fetch_strictly_public` is likely to become default on any time soon. But it is actually quite useful for Workers projects that actually act as the origin, which is becoming more and more common, rather than those that sit in-front of an external origin.

One thing about putting this into the C3 templates is that it helps raise awareness of this flag, and is easy to remove if not required.

PR #9149, opened 2025-05-06
Implement Mixed Mode proxy server & client

penalosa:
Fixes [DEVX-1838](https://jira.cfdata.org/browse/DEVX-1838) & [DEVX-1839](https://jira.cfdata.org/browse/DEVX-1839)

This PR continues the Mixed Mode implementation work by adding a client & server, as well as plumbing through the first binding type, AI.

The Mixed Mode server Worker is deployed using the `startMixedModeSession()` API as a realish preview Worker. It does one thing—proxy requests from binding implementations to the production binding implementation (i.e. from the KV implementation in `workerd` to the underlying KV implementation in production). 

The Mixed Mode client Worker is the immediate target of those requests from bindings. It then massages the requests slightly so they can be sent over the wire (putting the original URL in a header, prefixing original headers so they're not stripped by/don't interfere with the network call, recording the target binding name), and then sends them on to the server Worker.

An AI binding plugin has been added to Miniflare that purely supports Mixed Mode. Future bindings will be enabled for Mixed Mode in a followup PR (pending backend changes)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: covered by fixture
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/.github/workflows/test-and-check.yml
+++ b/.github/workflows/test-and-check.yml
@@ -226,8 +226,8 @@ jobs:
         if: steps.changes.outputs.everything_but_markdown == 'true'
         run: pnpm run test:ci --concurrency 1 ${{ matrix.filter }}
         env:
-          TMP_CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-          TMP_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
           NODE_OPTIONS: "--max_old_space_size=8192"
           WRANGLER_LOG_PATH: ${{ runner.temp }}/wrangler-debug-logs/
           TEST_REPORT_PATH: ${{ runner.temp }}/test-report/index.html
--- a/.github/workflows/test-and-check.yml
+++ b/.github/workflows/test-and-check.yml
@@ -224,7 +224,7 @@ jobs:
 
       - name: Run tests
         if: steps.changes.outputs.everything_but_markdown == 'true'
-        run: pnpm run test:ci --concurrency 1 ${{ matrix.filter }}
+        run: pnpm run test:ci --log-order=stream --concurrency 1 ${{ matrix.filter }}
         env:
           TEST_CLOUDFLARE_API_TOKEN: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
           TEST_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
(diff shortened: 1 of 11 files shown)


.github/workflows/test-and-check.yml
```diff
@@ -225,8 +225,8 @@ jobs:
         if: steps.changes.outputs.everything_but_markdown == 'true'
         run: pnpm run test:ci --concurrency 1 ${{ matrix.filter }}
         env:
-          TMP_CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-          TMP_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          TEST_CLOUDFLARE_API_TOKEN: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
+          TEST_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
```
dario-piotrowicz, line 229: out of curiosity, was `TMP_CLOUDFLARE_API_TOKEN` doing anything before? 🤔

dario-piotrowicz (approved):
Looks good to me 🙂

.github/workflows/test-and-check.yml
penalosa: Not as far as I can tell, no

packages/wrangler/src/api/mixedMode/index.ts
```diff
@@ -24,10 +26,7 @@ export async function startMixedModeSession(
 		config: proxyServerWorkerWranglerConfig,
 		dev: {
 			remote: true,
-			auth: {
-				accountId: await requireAuth({}),
-				apiToken: requireApiToken(),
-			},
+			auth: options?.auth,
```
dario-piotrowicz, line 29: why? wouldn't it make sense for `startMixedModeSession` to also handle the authentication? 🤔

packages/wrangler/templates/mixedMode/proxyServerWorker/index.ts
```diff
@@ -1,5 +1,22 @@
...
+			}
+			return env[targetBinding].fetch(
+				request.headers.get("MF-URL")!,
+				new Request(request, {
+					redirect: "manual",
+					headers: originalHeaders,
+				})
+			);
+		}
 		return new Response("no-op mixed-mode proxyServerWorker");
```
dario-piotrowicz, line 20: shall we update this to a 400 response? 🤔

dario-piotrowicz (approved):
looks good to me 🙂

packages/wrangler/src/api/mixedMode/index.ts
penalosa: This is just allowing the user to override the authentication. If they don't pass anything in, the `ConfigController` will provide a default value (which is the same as the one we had specified here)

dario-piotrowicz: ah ok, nice! thanks 🙂

Issue #9000, opened 2025-04-17
Cloudflare Vite plugin crashes when using https


li-kai:
### Which Cloudflare product(s) does this pertain to?

Vite Plugin

### What versions & operating system are you using?

Node v22, macOS, v1.0.5-1.0.8 of @cloudflare/vite-plugin

### Please provide a link to a minimal reproduction

https://github.com/li-kai/cloudflare-vite-https

### Describe the Bug

When using the [server https vite config](https://vite.dev/config/server-options.html#server-https), accessing the website crashes in node v22. I've provided a minimal repro with [vite's basic-ssl plugin](https://github.com/vitejs/vite-plugin-basic-ssl), which auto-provisions certs.

I can also reproduce it with my own certs if you wish not to use the plugin.
```
    server: {
      https: {
        key: path.resolve(__dirname, "./key.pem"),
        cert: path.resolve(__dirname, "./cert.pem"),
      },
```

### Please provide any relevant error logs

```
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/http2/util:656
      throw new ERR_HTTP2_INVALID_CONNECTION_HEADERS(key);
            ^

TypeError [ERR_HTTP2_INVALID_CONNECTION_HEADERS]: HTTP/1 Connection specific headers are forbidden: "transfer-encoding"
    at mapToHeaders (node:internal/http2/util:656:13)
    at ServerHttp2Stream.respond (node:internal/http2/core:2786:25)
    at [begin-send] (node:internal/http2/compat:894:19)
    at Http2ServerResponse.writeHead (node:internal/http2/compat:769:21)
    at Http2ServerResponse.write (node:internal/http2/compat:808:12)
    at Readable.ondata (node:internal/streams/readable:1009:22)
    at Readable.emit (node:events:518:28)
    at Readable.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3) {
  code: 'ERR_HTTP2_INVALID_CONNECTION_HEADERS'
```

PR #9152, opened 2025-05-06
Support HTTPS and HTTP/2

jamesopstad:
Fixes https://github.com/cloudflare/workers-sdk/issues/9000.

This adds support for HTTPS so that configuring [`server.https`](https://vite.dev/config/server-options#server-https) and/or [`preview.https`](https://vite.dev/config/preview-options#preview-https) in your Vite config now works as expected.

Vite uses HTTP/2 by default when HTTPS is enabled. This was previously broken with our plugin because Undici uses HTTP/1.1 and includes the `transfer-encoding: chunked` header for streamed responses. The presence of this header would cause the dev server to crash. The changes here detect if HTTP/2 is being used and, if so, remove the `transfer-encoding` header.

As part of this PR I've replaced `@hattip/adapter-node` with `@mjackson/node-fetch-server` as it has better compatibility with HTTP/2 and has greater usage. I've also added a new 'streaming' playground so that we have test coverage for streaming responses (with and without HTTPS enabled).

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because: N/A
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -369,23 +369,27 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 				}
 
 				return () => {
-					viteDevServer.middlewares.use(async (req, res) => {
+					viteDevServer.middlewares.use(async (req, res, next) => {
 						assert(miniflare, `Miniflare not defined`);
 						const routerWorker = await getRouterWorker(miniflare);
 
-						const request = createRequest(req, res);
-						const response = await routerWorker.fetch(
-							toMiniflareRequest(request),
-							{
-								redirect: "manual",
+						try {
+							const request = createRequest(req, res);
+							const response = await routerWorker.fetch(
+								toMiniflareRequest(request),
+								{
+									redirect: "manual",
+								}
+							);
+
+							if (req.httpVersionMajor === 2) {
+								response.headers.delete("transfer-encoding");
 							}
-						);
 
-						if (req.httpVersionMajor === 2) {
-							response.headers.delete("transfer-encoding");
+							await sendResponse(res, response as any);
+						} catch (error) {
+							next(error);
 						}
-
-						await sendResponse(res, response as any);
 					});
 				};
 			},
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -378,7 +378,9 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 								}
 							);
 
+							// Vite uses HTTP/2 when `server.https` is enabled
 							if (req.httpVersionMajor === 2) {
+								// HTTP/2 disallows use of the `transfer-encoding` header
 								response.headers.delete("transfer-encoding");
 							}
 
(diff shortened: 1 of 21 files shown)


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -399,23 +406,28 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+					try {
+						const request = createRequest(req, res);
+						const response = await miniflare.dispatchFetch(
+							toMiniflareRequest(request),
+							{ redirect: "manual" }
+						);
+
+						if (req.httpVersionMajor === 2) {
+							response.headers.delete("transfer-encoding");
+						}
```
dario-piotrowicz, line 425: I think that it would be pretty valuable to add a comment here explaining why this is needed

dario-piotrowicz (commented):
Out of curiosity, do you know if this is an expected undici behavior? Because based on this comment: https://github.com/nodejs/undici/blob/14e62db0d0cff4bea27357aa5bd14881459b27c7/lib/web/fetch/index.js#L1747-L1749 
It seems like undici does take into account that the header is incompatbile wth HTTP/2 to some degree 🤔

jamesopstad:
> Out of curiosity, do you know if this is an expected undici behavior? (is it an undici bug?) Because based on this comment: https://github.com/nodejs/undici/blob/14e62db0d0cff4bea27357aa5bd14881459b27c7/lib/web/fetch/index.js#L1747-L1749 It seems like undici does take into account that the header is incompatbile wth HTTP/2 to some degree 🤔
> 
> (What I'm trying to figure is whether we can make undici not add the header in the first place)

Hmm, good find. Yes, I agree it would be good to fix this elsewhere if possible (either in Undici or Miniflare). Do you know how Undici determines if the connection is HTTP/2 or not? I wonder if this doesn't apply because we're just using the `Request` and `Response` objects rather than `fetch`.

dario-piotrowicz:
> Hmm, good find. Yes, I agree it would be good to fix this elsewhere if possible (either in Undici or Miniflare). 

Yeah agreed 🙂 
I'm also happy with your solution but I would love to understand if it's actually necessary

> Do you know how Undici determines if the connection is HTTP/2 or not? 

I'm not really sure 😓 this maybe? 🤷 https://github.com/nodejs/undici/blob/14e62db0d0cff4bea27357aa5bd14881459b27c7/lib/dispatcher/client.js#L441

> I wonder if this doesn't apply because we're just using the `Request` and `Response` objects rather than `fetch`.

The comment I shared only talks about the request, but you're getting the header in the response, so maybe undici is handling one but not the other? 🤔

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: And above in the dev server

petebacondarwin (approved):
Really nice clean PR and I love the use of the new library.
Agree with @dario-piotrowicz about adding the comment.
But otherwise LGTM!

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: [f3d678f](https://github.com/cloudflare/workers-sdk/pull/9152/commits/f3d678f3114c97417ad45a63ea522fd84c7b0c55)

jamesopstad:
@dario-piotrowicz I'm going to keep the fix in the plugin for now but if there's a better way, feel free to update.

dario-piotrowicz:
> @dario-piotrowicz I'm going to keep the fix in the plugin for now but if there's a better way, feel free to open a follow up PR.

Sounds good, as I mentioned I'm fine with this solution 🙂👍

PR #9155, opened 2025-05-06
Cloudchamber update

gpanders:
Fixes CC-5141.

The Cloudchamber API was updated to prefer typed values for things like memory, disk, etc (while still supporting the older string values). Update Wrangler to use these fields when interacting with the API, leaving the string <-> value conversion to be done in the UI (rather than in the API).

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: The user interface to Wrangler (CLI flags and wrangler.toml) is unchanged, this only impacts how Wrangler interacts with the Cloudchamber API.
- Wrangler V3 Backport
  - [x] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/wrangler/src/cloudchamber/modify.ts
```diff
@@ -229,11 +236,17 @@ async function handleModifyCommand(
 		true
 	);
 
+	const memory = args.memory ?? config.cloudchamber.memory;
```
emily-shen, line 239: nit: this is repeated quite a few times, maybe chuck it in a util?

emily-shen (approved):

packages/wrangler/src/cloudchamber/client/models/AccountDefaults.ts
```diff
@@ -9,5 +9,10 @@ import type { MemorySizeWithUnit } from "./MemorySizeWithUnit";
  */
 export type AccountDefaults = {
 	vcpus: number;
+	memory_mib?: number;
+	/**
+	 * DEPRECATED
```
dario-piotrowicz, line 14:
```suggestion
```

packages/wrangler/src/cloudchamber/client/models/AccountDefaults.ts
```diff
@@ -9,5 +9,10 @@ import type { MemorySizeWithUnit } from "./MemorySizeWithUnit";
...
+	 * @deprecated
```
dario-piotrowicz, line 15: Could the deprecation (here and for the rest of the files) include a reason for the deprecation and/or what alternative to use?

packages/wrangler/src/cloudchamber/client/models/AccountLimit.ts
```diff
@@ -14,10 +14,23 @@ import type { NodeGroup } from "./NodeGroup";
...
 	memory_per_deployment: MemorySizeWithUnit;
+	memory_mib_per_deployment?: number;
+	/**
+	 * @deprecated
+	 */
 	disk_per_deployment: DiskSizeWithUnit;
+	disk_mb_per_deployment?: number;
 	total_vcpu: number;
+	/**
+	 * DEPRECATED
```
dario-piotrowicz, line 29:
```suggestion
```

emily-shen (commented):
also this probably shouldn't have a changeset since there are no user-impacting changes

packages/wrangler/src/__tests__/parse.test.ts
```diff
@@ -443,3 +444,37 @@ describe("searchLocation", () => {
...
+			["1.3kib", 1331],
+			["42MB", 42_000_000],
+			["42mib", 44_040_192],
+			["0.8MB", 800_000],
+			["0.8MiB", 838_860],
+			["2 GB", 2_000_000_000],
+
+			// If the b/ib suffix is omitted, assume non-binary units
+			["2G", 2_000_000_000],
+			["2 giB", 2_147_483_648],
```
dario-piotrowicz, line 463: this entry does end with `iB` right? so, is it in the correct section of cases? (or should it be above the `b/ib` comment?)

packages/wrangler/src/__tests__/parse.test.ts
```diff
@@ -443,3 +444,37 @@ describe("searchLocation", () => {
 		});
 	});
 });
+
+describe("parseByteSize", () => {
+	it("should calculate valid byte sizes", () => {
+		const cases: [string, number][] = [
+			["3", 3],
+			["3B", 3],
```
dario-piotrowicz, line 452:
```suggestion
			["3B", 3],
			["3b", 3],
```

packages/wrangler/src/parse.ts
```diff
@@ -359,3 +359,39 @@ export function parseNonHyphenedUuid(uuid: string | null): string | null {
 
 	return hyphenated.slice(0, 36);
 }
+
+export function parseByteSize(
```
dario-piotrowicz, line 363: I think that this function doesn't correctly parse values that incorrectly end with just `ib`
Screenshot 2025-05-23 at 15 35 38

shouldn't these result in `NaN`s?

packages/wrangler/src/__tests__/parse.test.ts
gpanders: Yes thank you! Must have added or modified this test at some point after making the comment. Moved as suggested.

packages/wrangler/src/parse.ts
gpanders: Good catch, thanks! Updated the function and added some test cases.

packages/wrangler/src/cloudchamber/modify.ts
gpanders: Done in 7a40af5a43c284a5d7c13a3ecb3d5a6e7ee8c801

gpanders:
> also this probably shouldn't have a changeset since there are no user-impacting changes

Removed

dario-piotrowicz (approved):
Looks great to me, thanks for the changes @gpanders 🫶

PR #9165, opened 2025-05-07
fix(r2): validate bucket names

### LINKED ISSUE #9148: [FEATURE] Improve `wrangler r2` UX
We have found several issue with the `wrangler r2` commands when developing the OpenNext adapters - using wrangler v4.6

### local

**L.1: it is possible to put when the bucket does not exist:**

```
-> % wrangler r2 object put "unexisting-bucket/key" --file LICENSE --local 

 ⛅️ wrangler 4.6.0 (update available 4.14.2)
------------------------------------------------------

Creating object "key" in bucket "unexisting-bucket".
▲ [WARNING] By default, `wrangler r2` commands access a local simulator of your R2 bucket, the same as that used by `wrangler dev`. To access your remote R2 bucket, re-run the command with the --remote flag


Upload complete.
```

**L.2:  bucket names are not validated (they should be lowercase)**

```
-> % wrangler r2 object put "INVALID/key" --file LICENSE --local 

 ⛅️ wrangler 4.6.0 (update available 4.14.2)
------------------------------------------------------

Creating object "key" in bucket "INVALID".
▲ [WARNING] By default, `wrangler r2` commands access a local simulator of your R2 bucket, the same as that used by `wrangler dev`. To access your remote R2 bucket, re-run the command with the --remote flag


Upload complete.
```

Here we should error saying that the bucket name is invalid

### remote

**R.1: invalid bucket names generate not very helpful error message**

```
-> % wrangler r2 object put "INVALID/key" --file LICENSE --remote

 ⛅️ wrangler 4.6.0 (update available 4.14.2)
------------------------------------------------------

Creating object "key" in bucket "INVALID".

✘ [ERROR] Failed to fetch /accounts/0a32994c9e7d81ae884d4b40873dd43f/r2/buckets/INVALID/objects/key - 400: Bad Request);


If you think this is a bug then please create an issue at https://github.com/cloudflare/workers-sdk/issues/new/choose
Note that there is a newer version of Wrangler available (4.14.2). Consider checking whether upgrading resolves this error.
✔ Would you like to report this error to Cloudflare? Wrangler's output and the error details will be shared with the Wrangler team to help us diagnose and fix the issue. … no
```

We should probably have an error message like this one:

```
-> % wrangler r2 bucket create "INVALID"

 ⛅️ wrangler 4.6.0 (update available 4.14.2)
------------------------------------------------------


✘ [ERROR] The bucket name "INVALID" is invalid. Bucket names must begin and end with an alphanumeric and can only contain letters (a-z), numbers (0-9), and hyphens (-).
```

/cc @conico974 @james-elicx

vicb:
Fixes #9148

Before this PR, bucket names were only validated when creating a bucket (via `wrangler r2 create bucket_name`)

This PR adds validation to:
- wrangler config
- object commands - when trying to put, get, delete an object
- pages dev CLI command
- pipeline config

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: names are documented
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/9172
  - [ ] Not necessary because: patch

### DIFF
--- a/fixtures/pages-functions-app/functions/r2/create.ts
+++ b/fixtures/pages-functions-app/functions/r2/create.ts
@@ -1,8 +1,8 @@
 type Env = {
-	BUCKET: R2Bucket;
+	bucket: R2Bucket;
 };
 
 export const onRequestPut: PagesFunction<Env> = async ({ env }) => {
-	const object = await env.BUCKET.put("test", "Hello world!");
+	const object = await env.bucket.put("test", "Hello world!");
 	return new Response(JSON.stringify(object));
 };
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -3,6 +3,7 @@ import path from "node:path";
 import { dedent } from "ts-dedent";
 import { UserError } from "../errors";
 import { getFlag } from "../experimental-flags";
+import { isValidR2BucketName } from "../r2/helpers";
 import { friendlyBindingNames } from "../utils/print-bindings";
 import { Diagnostics } from "./diagnostics";
 import {
@@ -2561,6 +2562,18 @@ const validateR2Binding: ValidatorFn = (diagnostics, field, value) => {
 		);
 		isValid = false;
 	}
+	if (
+		isValid &&
+		hasProperty(value, "bucket_name") &&
+		!isValidR2BucketName(value.bucket_name)
+	) {
+		diagnostics.errors.push(
+			`"${field}.bucket_name" must begin and end with an alphanumeric, only contain letters (a-z), numbers (0-9), and hyphens (-)," +
+			" and be between 3 and 63 characters long. but got ${JSON.stringify(value.bucket_name)}.`
+		);
+		isValid = false;
+	}
+
 	if (!isOptionalProperty(value, "preview_bucket_name", "string")) {
 		diagnostics.errors.push(
 			`"${field}" bindings should, optionally, have a string "preview_bucket_name" field but got ${JSON.stringify(
@@ -2569,6 +2582,18 @@ const validateR2Binding: ValidatorFn = (diagnostics, field, value) => {
 		);
 		isValid = false;
 	}
+	if (
+		isValid &&
+		hasProperty(value, "preview_bucket_name") &&
+		!isValidR2BucketName(value.preview_bucket_name)
+	) {
+		diagnostics.errors.push(
+			`"${field}.preview_bucket_name" must begin and end with an alphanumeric, only contain letters (a-z), numbers (0-9), and hyphens (-)," +
+			" and be between 3 and 63 characters long. but got ${JSON.stringify(value.preview_bucket_name)}.`
+		);
+		isValid = false;
+	}
+
 	if (!isOptionalProperty(value, "jurisdiction", "string")) {
 		diagnostics.errors.push(
 			`"${field}" bindings should, optionally, have a string "jurisdiction" field but got ${JSON.stringify(
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -3,7 +3,7 @@ import path from "node:path";
 import { dedent } from "ts-dedent";
 import { UserError } from "../errors";
 import { getFlag } from "../experimental-flags";
-import { isValidR2BucketName } from "../r2/helpers";
+import { bucketFormatMessage, isValidR2BucketName } from "../r2/helpers";
 import { friendlyBindingNames } from "../utils/print-bindings";
 import { Diagnostics } from "./diagnostics";
 import {
@@ -2568,8 +2568,7 @@ const validateR2Binding: ValidatorFn = (diagnostics, field, value) => {
 		!isValidR2BucketName(value.bucket_name)
 	) {
 		diagnostics.errors.push(
-			`"${field}.bucket_name" must begin and end with an alphanumeric, only contain letters (a-z), numbers (0-9), and hyphens (-)," +
-			" and be between 3 and 63 characters long. but got ${JSON.stringify(value.bucket_name)}.`
+			`${field}.bucket_name=${JSON.stringify(value.bucket_name)} is invalid. ${bucketFormatMessage}`
 		);
 		isValid = false;
 	}
@@ -2588,8 +2587,7 @@ const validateR2Binding: ValidatorFn = (diagnostics, field, value) => {
 		!isValidR2BucketName(value.preview_bucket_name)
 	) {
 		diagnostics.errors.push(
-			`"${field}.preview_bucket_name" must begin and end with an alphanumeric, only contain letters (a-z), numbers (0-9), and hyphens (-)," +
-			" and be between 3 and 63 characters long. but got ${JSON.stringify(value.preview_bucket_name)}.`
+			`${field}.preview_bucket_name= ${JSON.stringify(value.preview_bucket_name)} is invalid. ${bucketFormatMessage}`
 		);
 		isValid = false;
 	}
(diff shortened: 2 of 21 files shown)


penalosa:
@vicb this needs a backport since it's a Wrangler patch

vicb:
> @vicb this needs a backport since it's a Wrangler patch

I don't think we want to backport that?

vicb:
> > @vicb this needs a backport since it's a Wrangler patch
> 
> I don't think we want to backport that?

Discussed offline, I'll create a v3 PR when this one is green and approved

fixtures/pages-functions-app/functions/r2/create.ts
```diff
@@ -1,8 +1,8 @@
 type Env = {
-	BUCKET: R2Bucket;
+	bucket: R2Bucket;
```
petebacondarwin, line 2: Seems like an unnecessary change, or am I missing something?

packages/wrangler/src/config/validation.ts
```diff
@@ -2561,6 +2562,18 @@ const validateR2Binding: ValidatorFn = (diagnostics, field, value) => {
...
 	}
+	if (
+		isValid &&
+		hasProperty(value, "bucket_name") &&
+		!isValidR2BucketName(value.bucket_name)
+	) {
+		diagnostics.errors.push(
+			`"${field}.bucket_name" must begin and end with an alphanumeric, only contain letters (a-z), numbers (0-9), and hyphens (-)," +
+			" and be between 3 and 63 characters long. but got ${JSON.stringify(value.bucket_name)}.`
+		);
```
petebacondarwin, line 2573: This message is repeated six times in the code. Could it just be added to the `isValidR2BucketName()` function to avoid the duplication?

packages/wrangler/src/pages/dev.ts
```diff
@@ -1269,6 +1270,16 @@ function getBindingsFromArgs(args: typeof pagesDevCommand.args): Partial<
...
 				}
 
+				const bucketName = ref || binding.toString();
+
+				if (!isValidR2BucketName(bucketName)) {
+					logger.error(
+						`The bucket name "${bucketName}" is invalid. ` +
+							"Bucket names must begin and end with an alphanumeric, only contain letters (a-z), numbers (0-9), and hyphens (-), and be between 3 and 63 characters long."
+					);
+					return;
```
petebacondarwin, line 1280: Why the quiet exit here?
If you throw a UserError then the process will exit with a non-zero exit code.

packages/wrangler/src/config/validation.ts
```diff
@@ -2561,6 +2562,18 @@ const validateR2Binding: ValidatorFn = (diagnostics, field, value) => {
 		);
 		isValid = false;
 	}
+	if (
+		isValid &&
+		hasProperty(value, "bucket_name") &&
+		!isValidR2BucketName(value.bucket_name)
+	) {
+		diagnostics.errors.push(
+			`"${field}.bucket_name" must begin and end with an alphanumeric, only contain letters (a-z), numbers (0-9), and hyphens (-)," +
```
emily-shen, line 2571: looks like uppercase letters are invalid too, even though the api error message gives zero indication about this :/

```suggestion
			`"${field}.bucket_name" must begin and end with an alphanumeric character, only contain lowercase letters (a-z), numbers (0-9), and hyphens (-)," +
```

packages/wrangler/src/pipelines/cli/create.ts
```diff
@@ -157,6 +158,10 @@ export const pipelinesCreateCommand = createCommand({
 
 	async handler(args, { config }) {
 		const bucket = args.r2Bucket;
+		if (!isValidR2BucketName(bucket)) {
+			`The bucket name "${bucket}" is invalid. ` +
```
emily-shen, line 162: this doesn't seem to be throwing an error

fixtures/pages-functions-app/functions/r2/create.ts
vicb: The name should be lowercase

packages/wrangler/src/config/validation.ts
vicb: We can do that, probably renaming the function to `validateR2BucketName` returning `{isValid, errorMsg?}`

packages/wrangler/src/pages/dev.ts
vicb: Because this is similar to line 1270 and I don't any `UserError` here.

packages/wrangler/src/pipelines/cli/create.ts
vicb: Good catch, it's missing a `throw UserError(...)` I guess

fixtures/pages-functions-app/functions/r2/create.ts
petebacondarwin: But this is not the bucket name, it is the binding name, which doesn't have the same validation requirements.

vicb:
@emily-shen @petebacondarwin thanks for the reviews, I have addressed the feedback, PTAL

- Factor and clarify the format message into a single const
- Add a missing `throw new UserError(...)`

packages/wrangler/src/config/validation.ts
```diff
@@ -2569,6 +2581,17 @@ const validateR2Binding: ValidatorFn = (diagnostics, field, value) => {
 		);
 		isValid = false;
 	}
+	if (
+		isValid &&
```
emily-shen, line 2585: with this `isValid` check, if you have invalid  bucket_name and invalid preview_bucket_name, only the error for bucket_name will show up I think?

vicb: That's right.

I felt like this was good enough (i.e. you'll get the other error the next time if only one is fixed) vs making the code more complicated (to only validate when there is a name)

But if you feel strongly I can change that, LMK

emily-shen: i think it we usually try our best to avoid that pattern - fixing something and exposing another error, although the error is the same here so its probably not a huge deal.

emily-shen (approved):

packages/wrangler/src/config/validation.ts
vicb: I understand but at the same time I don't really understand why we have a `preview_bucket_name` when we have environments and I don't think other bindings have a similar thing.

arcticmatt:
would be nice if there was a flag/option to disable this behavior, since it's not backwards compatible with all `wrangler.jsonc` files that were working before this patch

edmundhung:
@arcticmatt We are relaxing R2 bucket validation for pages dev commands in #9257 based on the feedbacks and will release it very soon.

arcticmatt:
@edmundhung I'm not using pages dev

Here's an example of the issue I ran into:

```
pnpm wrangler d1 migrations apply DB_NAME --local

 ⛅️ wrangler 4.15.1
-------------------


✘ [ERROR] Processing wrangler.jsonc configuration:

    - r2_buckets[0].bucket_name="REPLACEME_R2_BUCKET_NAME" is invalid. Bucket names must begin and
  end with an alphanumeric character, only contain lowercase letters, numbers, and hyphens, and be
  between 3 and 63 characters long.
```

I have some existing code that uses all caps for bucket names, and then I replace the bucket name with a valid one once deploying. It's not too bad to start using lowercase for the placeholder name, but I didn't expect this to break when upgrading to a new minor version of Wrangler

Cherry:
Yeah, this being validated at wrangler config level, which happens for (almost?) every command in `wrangler` can be quite problematic when this wasn't previous enforced this way. Template or multi-stage-like projects where bindings are inserted at build time, etc. are going to suffer with errors now when upgrading.

If this is to be validated, I'd recommend it's only on operations where it fundamentally needs to be - like `r2` commands, deploying a worker, etc. Running local d1 migrations for example in a project with a `wrangler.json` that has a template like `<BINDING_NAME>` similar to the docs really shouldn't start causing commands to throw errors when upgrading to a new minor wrangler version.

PR #9179, opened 2025-05-08
Rename `Cloudflare` namespace in generated types to `CF`

emily-shen:
Fixes DEVX-1843

the `Cloudflare` namespace was conflicting with the `Cloudflare` global when both generated types and @cloudflare/workers-types existed. This renames the namespace (not user facing) to `CF` to prevent this.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not a public facing change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: `Cloudflare` namespace doesn't exist in v3 types output


penalosa (commented):
This will need a corresponding workerd PR and release before landing, but LGTM ✅

emily-shen:
> the `Cloudflare` namespace was conflicting with the `Cloudflare` global when both generated types and @cloudflare/workers-types existed. This renames the namespace (not user facing) to `CF` to prevent this.

https://github.com/cloudflare/workerd/pull/4110

Cherry (commented):
If anyone is using something like: `export default {...} satisfies ExportedHandler<Cloudflare.Env>` won't this be broken with this change?

There's also code in the wild that is using this: https://github.com/redwoodjs/sdk/blob/f4596a9c5d1721bd105b6db33aa72e75cce9e3dc/sdk/src/runtime/lib/realtime/worker.ts#L12

emily-shen:
> If anyone is using something like: `export default {...} satisfies ExportedHandler<Cloudflare.Env>` won't this be broken with this change?
> 
> There's also code in the wild that is using this for popular frameworks: https://github.com/redwoodjs/sdk/blob/f4596a9c5d1721bd105b6db33aa72e75cce9e3dc/sdk/src/runtime/lib/realtime/worker.ts#L12

Argh, good catch @Cherry - we'll revert this. Thank you 🧡

PR #9183, opened 2025-05-08
add `remote` option to initial bindings

dario-piotrowicz:
Fixes (partially) https://jira.cfdata.org/browse/DEVX-1852

This PR adds the `remote` config option to some of the binding configs

(this PR does not include all bindings, I think it makes sense to add the others as we finalize the API/implement support for them?)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: undocumented experimental feature
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: new experimental feature

### DIFF
--- a/.changeset/orange-ads-judge.md
+++ b/.changeset/orange-ads-judge.md
@@ -1,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+add `remote` option to initial bindings
+
+add the `remote` option (initial implementation
+gated behind `--x-mixed-mode`) for the following
+bindings: `service`, `kv`, `r2`, `d1`, `queue` and `workflow`
--- a/packages/wrangler/src/dev/miniflare.ts
+++ b/packages/wrangler/src/dev/miniflare.ts
@@ -414,10 +414,6 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 } {
 	const bindings = config.bindings;
 
-	// TODO(DEVX-1853, DEVX-1854): call `startMixedModeSession` here so that we get mixedModeConnectionStrings that we can
-	//       pass to miniflare (Note: this function is used from `LocalRuntimeController`, `MultiworkerRuntimeController` and
-	//       `getPlatformProxy`)
-
 	// Setup blob and module bindings
 	// TODO: check all these blob bindings just work, they're relative to cwd
 	const textBlobBindings = { ...bindings.text_blobs };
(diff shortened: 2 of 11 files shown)


.changeset/orange-ads-judge.md
```diff
@@ -0,0 +1,9 @@
+---
+"wrangler": patch
+---
+
+add `remote` option to initial bindings
+
+add the `remote` option (initial implementation
+gated behind `--x-mixed-mode`) for the following
+bindings: `service`, `kv`, `r2`, `d1`, `queue` and `workflow`
```
penalosa, line 9: How was this subset chosen? I think we probably also want Images, Vectorize, Browser Rendering, and mTLS?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -414,6 +414,10 @@ export function buildMiniflareBindingOptions(config: MiniflareBindingsConfig): {
 } {
 	const bindings = config.bindings;
 
+	// TODO(DEVX-1853, DEVX-1854): call `startMixedModeSession` here so that we get mixedModeConnectionStrings that we can
+	//       pass to miniflare (Note: this function is used from `LocalRuntimeController`, `MultiworkerRuntimeController` and
+	//       `getPlatformProxy`)
+
```
penalosa, line 420: I don't think we'd want to be calling it here, since this function is called on every reload. I think we'd probably want to call it directly in the controllers?

penalosa (approved):

.changeset/orange-ads-judge.md
penalosa: Those aren't—all of them can be remote or local (admittedly the local experience doesn't work yet for most of them...)

dario-piotrowicz: ok yeah, but again these feel a bit controversial/unclear to me right now, so that's why I am thinking of skipping them for now and deal with them later when we clarified how we want our API to look like

would you be ok with this? 🙂

penalosa: Sure, sounds good

penalosa (approved):

PR #9198, opened 2025-05-10
fix: start transferred MessagePort in fetch-sync worker

kylecarbs:
This was causing running Miniflare in Bun tests to hang indefinitely.

Fixes https://github.com/oven-sh/bun/issues/16240

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: no effect under node.js
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no effect under node.js
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no public change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/9211
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/stupid-files-jog.md
+++ b/.changeset/stupid-files-jog.md
@@ -1,0 +1,9 @@
+---
+"miniflare": patch
+---
+
+fix: ensure the fetch proxy message port is started
+
+While Node.js will start the message port automatically when a `message` event listener is added,
+this diverges from the standard Web API for message ports, which require you to explicitly start
+listening on the port.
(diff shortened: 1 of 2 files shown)


petebacondarwin (approved):
Cool. Thanks for adding this.
In Node.js is appears that this is unnecessary as it is [implied by attaching a `message` listener](https://nodejs.org/api/worker_threads.html#class-messageport:~:text=When%20using%20this%20port%20as%20an%20event%20emitter%2C%20this%20is%20called%20automatically%20once%20%27message%27%20listeners%20are%20attached.).
But I can see that is where Node.js diverges from the [Web standard API](https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/start#:~:text=Another%20option%20would%20be%20to%20do%20this%20using%20EventTarget.addEventListener%2C%20however%2C%20when%20this%20method%20is%20used%2C%20you%20need%20to%20explicitly%20call%20start()%20to%20begin%20the%20flow%20of%20messages%20to%20this%20document%3A). 

I'll add a changeset for good measure. Since this has no effect on Node.js and we don't officially support Bun, I don't see any value in adding an extra test for this functionality.

PR #9219, opened 2025-05-13
bump unenv to 2.0.0-rc.17

vicb:
Thanks @pi0 for unenv 2.0.0-rc.17

@thomasgauvin This will pull debug named exports - should be in wrangler at the end of the week

/cc @anonrig 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: version bump
- Wrangler V3 Backport
  - [] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: we be handled as a follow up, see https://github.com/cloudflare/workers-sdk/issues/9222

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #9221 
- <kbd>&nbsp;2&nbsp;</kbd> #9220 
- <kbd>&nbsp;1&nbsp;</kbd> #9219 👈

PR #8939, opened 2025-04-14
add missing fns to unenv-preset

anonrig:
Adds the missing fns and modules to unenv-preset

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Existing tests are enough
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Documentation not necessary because:
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:

PR #9220, opened 2025-05-13
Sync unenv-preset with workerd

vicb:
This PR supersedes #8939 by @anonrig. The current PR contain the code only, tests will be merged once the preset is bumped into wrangler and are in the follow up PRs.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: need version update, added in the next PR
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: bump
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

---
This is **part 2 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #9221 
- <kbd>&nbsp;2&nbsp;</kbd> #9220 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #9219

### DIFF
--- a/packages/unenv-preset/package.json
+++ b/packages/unenv-preset/package.json
@@ -45,7 +45,7 @@
 		"test:watch": "vitest"
 	},
 	"devDependencies": {
-		"@types/node-unenv": "npm:@types/node@^22.13.9",
+		"@types/node-unenv": "npm:@types/node@^22.14.0",
 		"typescript": "catalog:default",
 		"unbuild": "^3.2.0",
 		"undici": "catalog:default",
@@ -54,7 +54,7 @@
 	},
 	"peerDependencies": {
 		"unenv": "2.0.0-rc.17",
-		"workerd": "^1.20250320.0"
+		"workerd": "^1.20250508.0"
 	},
 	"peerDependenciesMeta": {
 		"workerd": {
--- a/.changeset/five-peaches-cough.md
+++ b/.changeset/five-peaches-cough.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/unenv-preset": patch
+---
+
+Sync unenv-preset with workerd
--- a/packages/unenv-preset/src/preset.ts
+++ b/packages/unenv-preset/src/preset.ts
@@ -15,6 +15,8 @@ const nodeCompatModules = [
 	"_stream_readable",
 	"_stream_transform",
 	"_stream_writable",
+	"_tls_common",
+	"_tls_wrap",
 	"assert",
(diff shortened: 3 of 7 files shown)


packages/unenv-preset/package.json
```diff
@@ -45,7 +45,7 @@
 		"test:watch": "vitest"
 	},
 	"devDependencies": {
-		"@types/node-unenv": "npm:@types/node@^22.13.9",
+		"@types/node-unenv": "npm:@types/node@^22.14.0",
```
vicb, line 48: Sync with https://github.com/unjs/unenv/blob/fa3e8c37c2a2230a75d98925a8ed3a9813b0bd7e/.nvmrc

petebacondarwin (commented):
Does the unenv-preset dependency on workerd get bumped by our dependabot PRs?
How do we align things between unenv-preset and miniflare if they depend upon different versions of workerd?

vicb:
> Does the unenv-preset dependency on workerd get bumped by our dependabot PRs? 

No the `workerd` version in `unenv-preset` is whatever is needed for the the preset. 
Meaning this could be an earlier version here but I picked the version I tested.

> How do we align things between unenv-preset and miniflare if they depend upon different versions of workerd?

`unenv-preset` uses the minimum workerd version it requires. `wrangler`/`miniflare` can use a more recent version.

petebacondarwin:
> `unenv-preset` uses the minimum workerd version it requires. `wrangler`/`miniflare` can use a more recent version.

So we need to make sure that we bump the workerd version in Miniflare before we bump to a new version of unenv-preset that has a minimum on a new workerd version?

anonrig (approved):

packages/unenv-preset/package.json
```diff
@@ -54,7 +54,7 @@
 	},
 	"peerDependencies": {
 		"unenv": "2.0.0-rc.17",
```
petebacondarwin, line 56: We'll need to update the vite-plugin to depend upon this version too I think

petebacondarwin (approved):
I think this is fine but I am surprised that there is no peer dependency problem in the vite-plugin e2e tests...

packages/unenv-preset/package.json
vicb: https://github.com/cloudflare/workers-sdk/pull/9228

PR #9223, opened 2025-05-13
clarify the description of `--keep-vars`

vicb:
The original description is:

```text
Stop Wrangler from deleting vars that are not present in the Wrangler configuration file
By default Wrangler will remove all vars and replace them with those found in the Wrangler configuration.
If your development approach is to modify vars after deployment via the dashboard you may wish to set this flag.
```

I think this is confusing:

```text
Stop Wrangler from deleting vars that are not present in the Wrangler configuration file
```

Wrangler actually deletes all the variables (the ones in the configuration file are uploaded again but they are still deleted in the first place - see the second sentence)

```text
If your development approach is to modify vars after deployment via the dashboard you may wish to set this flag.
```

I find this confusing for 2 reasons:

- if you update the vars "after deployment" then it should not matter wether they are deleted or not
- "set this flag" is confusing (i.e. set to true/false?)

Also we mention in the docs that [secrets are environment variables](https://developers.cloudflare.com/workers/configuration/secrets/#compare-secrets-and-environment-variables) so it's nice to add what happens to secrets

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: Text update
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/22386
  - [ ] Documentation not necessary because: not affected
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: text update


packages/wrangler/src/deploy/index.ts
```diff
@@ -176,9 +176,9 @@ export const deployCommand = createCommand({
 		},
 		"keep-vars": {
 			describe:
-				"When set to true, the environment variables are not deleted before the deployment.\n" +
-				"When set to false (default) Wrangler will delete all vars before settings those found in the Wrangler configuration.\n" +
-				"If you set variables via the dashboard you probably want to set this flag to true.\n" +
+				"When not used, Wrangler will delete all vars before settings those found in the Wrangler configuration.\n" +
+				"When used, the environment variables are not deleted before the deployment.\n" +
```
petebacondarwin, line 180:
```suggestion
				"When used (and set to true), the environment variables are not deleted before the deployment.\n" +
```

packages/wrangler/src/deploy/index.ts
```diff
@@ -176,7 +176,10 @@ export const deployCommand = createCommand({
 		},
 		"keep-vars": {
 			describe:
-				"Stop Wrangler from deleting vars that are not present in the Wrangler configuration file\nBy default Wrangler will remove all vars and replace them with those found in the Wrangler configuration.\nIf your development approach is to modify vars after deployment via the dashboard you may wish to set this flag.",
+				"When not used, Wrangler will delete all vars before setting those found in the Wrangler configuration.\n" +
```
CarmenPopoviciu:
```suggestion
				"When not used (or set to false), Wrangler will delete all vars before setting those found in the Wrangler configuration.\n" +
```

PR #9230, opened 2025-05-13
Support --tag and --message in wrangler deploy

WalshyDev:
Fixes N/A

Adds support for specifying `--tag` and/org `--message` in `wrangler deploy` not only `wrangler versions upload`

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [x] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/easy-towns-fall.md
+++ b/.changeset/easy-towns-fall.md
@@ -1,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+chore: Added support for --tag and --message in `wrangler deploy` when using the verisons/deployments API (essentially when not using Service Worker format and Durable Objects).
+
+This lets you specify a tag and message for the deployment, which is useful for tracking changes and managing deployments more effectively. This was previously only available on the `wrangler versions upload` command.
--- a/packages/wrangler/src/deploy/deploy.ts
+++ b/packages/wrangler/src/deploy/deploy.ts
@@ -112,6 +112,8 @@ type Props = {
 	dispatchNamespace: string | undefined;
 	experimentalAutoCreate: boolean;
 	metafile: string | boolean | undefined;
+	tag: string | undefined;
+	message: string | undefined;
 };
 
 export type RouteObject = ZoneIdRoute | ZoneNameRoute | CustomDomainRoute;
@@ -710,12 +712,42 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 						}
 					: undefined,
 			observability: config.observability,
+			annotations:
+				props.message || props.tag
+					? {
+							"workers/message": props.message,
+							"workers/tag": props.tag,
+						}
+					: undefined,
 		};
 
 		sourceMapSize = worker.sourceMaps?.reduce(
 			(acc, m) => acc + m.content.length,
 			0
 		);
+
+		// We can use the new versions/deployments APIs if we:
+		// * are uploading a worker that already exists
+		// * aren't a dispatch namespace deploy
+		// * aren't a service env deploy
+		// * aren't a service Worker
+		// * we don't have DO migrations
+		// * we aren't an fpw
+		// * not a container worker
+		const canUseNewVersionsDeploymentsApi =
+			workerExists &&
+			props.dispatchNamespace === undefined &&
+			prod &&
+			format === "modules" &&
+			migrations === undefined &&
+			!config.first_party_worker &&
+			config.containers === undefined;
+
+		if (!canUseNewVersionsDeploymentsApi && worker.annotations !== undefined) {
+			logger.warn(
+				"Worker does not support Worker Versions, ignoring tag and message."
+			);
+		}
 
 		await printBundleSize(
 			{ name: path.basename(resolvedEntryPointPath), content: content },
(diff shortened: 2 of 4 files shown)


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -710,13 +712,43 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
+			workerExists &&
+			props.dispatchNamespace === undefined &&
+			prod &&
+			format === "modules" &&
+			migrations === undefined &&
+			!config.first_party_worker &&
+			config.containers === undefined;
+
+		if (!canUseNewVersionsDeploymentsApi && worker.annotations !== undefined) {
+			logger.warn(
```
WalshyDev, line 747: move this up so we were above the bundle size print

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -710,13 +712,43 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
+				"Not using the new versions/deployments API, so annotations will not be set."
```
WalshyDev, line 748: copy tbd, not fully sure what we want to say here

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -710,13 +712,43 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
+			props.dispatchNamespace === undefined &&
+			prod &&
+			format === "modules" &&
+			migrations === undefined &&
+			!config.first_party_worker &&
+			config.containers === undefined;
+
+		if (!canUseNewVersionsDeploymentsApi && worker.annotations !== undefined) {
+			logger.warn(
+				"Worker does not support Worker Versions, ignoring tag and message."
```
WillTaylorDev, line 748: From a user perspective, it might be a little confusing on why we're talking about worker version in a deploy command. Maybe something like "Worker does not support annotations. Ignoring tag and message" -- and/or maybe we could check the props to see if tag/annotations are provided at all? Non blocker for me either way.

WalshyDev: > From a user perspective, it might be a little confusing on why we're talking about worker version in a deploy command. 

Agreed, I just can't think of better wording 😅 

> maybe we could check the props to see if tag/annotations are provided at all?

that's what the annotations !== undefined is doing

.changeset/easy-towns-fall.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+chore: Added support for --tag and --message in `wrangler deploy` when using the verisons/deployments API (essentially when not using Service Worker format and Durable Objects).
```
penalosa, line 5:
```suggestion
Added support for --tag and --message in `wrangler deploy` when using the versions/deployments API (essentially when not using Service Worker format and Durable Objects).
```

Does this mean service worker _and_ DOs, or service worker _or_ DOs?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -710,13 +712,43 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 						}
 					: undefined,
 			observability: config.observability,
+			annotations:
+				props.message || props.tag
+					? {
+							"workers/message": props.message,
+							"workers/tag": props.tag,
+						}
+					: undefined,
```
penalosa, line 721: Why are we sending both when only one is specified?

packages/wrangler/src/deploy/deploy.ts
penalosa: @korinne would love a product perspective on this. I'd find this super confusing as a user, and I'd want to know _why_ my Worker doesn't support Worker Versions, and what I can do about it. Most of the reasons are old and deprecated, so that seems fair enough, but what about DOs? That's a core part of our platform that really should work with Versions, I think

packages/wrangler/src/deploy/index.ts
```diff
@@ -211,6 +211,17 @@ export const deployCommand = createCommand({
...
+			describe: "A tag for this Worker Gradual Rollouts Version",
+			type: "string",
+			requiresArg: true,
+		},
+		message: {
+			describe:
+				"A descriptive message for this Worker Gradual Rollouts Version",
+			type: "string",
+			requiresArg: true,
+		},
```
penalosa, line 224: The product name is gradual deployments, I think (https://developers.cloudflare.com/workers/configuration/versions-and-deployments/gradual-deployments/). However, given this is for `deploy`, why the mention of versions?

penalosa (changes requested):

lrapoport-cf:
hey @WalshyDev are you still planning on trying to land this PR? if so, can you please follow up on @penalosa 's comments?

penalosa:
Closing for now. @WalshyDev if you're still planning to land this feel free to open a new PR

PR #9241, opened 2025-05-14
test: fail e2e test if peer dependencies are broken

petebacondarwin:
Changes the vite-plugin e2e tests so that they fail if there is a mismatch in a peer dependency.
This is to catch problems with the relationship between vite-plugin and unenv-preset.

### DIFF
--- a/packages/vite-plugin-cloudflare/e2e/helpers.ts
+++ b/packages/vite-plugin-cloudflare/e2e/helpers.ts
@@ -32,6 +32,7 @@ const testEnv = {
 const strictPeerDeps = {
 	pnpm: "--strict-peer-dependencies",
 	npm: "--strict-peer-deps",
+	// yarn does not have an option for strict checks
 	yarn: "",
 };
 
--- a/packages/vite-plugin-cloudflare/e2e/helpers.ts
+++ b/packages/vite-plugin-cloudflare/e2e/helpers.ts
@@ -29,6 +29,12 @@ const testEnv = {
 	VITEST: undefined,
 };
 
+const strictPeerDeps = {
+	pnpm: "--strict-peer-dependencies",
+	npm: "--strict-peer-deps",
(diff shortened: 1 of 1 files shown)


packages/vite-plugin-cloudflare/e2e/helpers.ts
```diff
@@ -29,6 +29,12 @@ const testEnv = {
 	VITEST: undefined,
 };
 
+const strictPeerDeps = {
+	pnpm: "--strict-peer-dependencies",
+	npm: "--strict-peer-deps",
+	yarn: "",
+};
```
vicb, line 36:
```suggestion
const strictPeerDeps = {
  // yarn does not have an option for strict checks
	pnpm: "--strict-peer-dependencies",
	npm: "--strict-peer-deps",	
};
```

packages/vite-plugin-cloudflare/e2e/helpers.ts
```diff
@@ -42,7 +48,9 @@ export function seed(fixture: string, pm: "pnpm" | "yarn" | "npm") {
 		debuglog("Fixture copied to " + projectPath);
 		await updateVitePluginVersion(projectPath);
 		debuglog("Updated vite-plugin version in package.json");
-		runCommand(`${pm} install`, projectPath, { attempts: 2 });
+		runCommand(`${pm} install ${strictPeerDeps[pm]}`, projectPath, {
```
vicb, line 51:
```suggestion
		runCommand(`${pm} install ${strictPeerDeps[pm] ?? ""}`, projectPath, {
```

vicb (approved):

packages/vite-plugin-cloudflare/e2e/helpers.ts
petebacondarwin: Added the comment but not removed the option

PR #9245, opened 2025-05-14
Support Mixed Mode Dispatch Namespaces

penalosa:
Continuation of DEVX-1859

Support Dispatch Namespaces in Mixed Mode dev. These are _similar_ to service bindings, but have a synchronous `.get()` method that a user calls first to retrieve the `Fetcher`. As such, they've been special cased so that the `.get()` is implemented as a Wrapped Binding that attaches a special header encoding the `.get()` arguments. The Mixed Mode Server uses that to call `.get()` on the actual Dispatch Namespace binding when the `Fetcher` is used.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: covered by fixture
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a patch

### DIFF
--- a/packages/miniflare/src/plugins/dispatch-namespace/index.ts
+++ b/packages/miniflare/src/plugins/dispatch-namespace/index.ts
@@ -1,0 +1,101 @@
+import assert from "node:assert";
+import LOCAL_DISPATCH_NAMESPACE from "worker:dispatch-namespace/dispatch-namespace";
+import { z } from "zod";
+import { Worker_Binding } from "../../runtime";
+import {
+	mixedModeClientWorker,
+	MixedModeConnectionString,
+	Plugin,
+	ProxyNodeBinding,
+} from "../shared";
+
+export const DispatchNamespaceOptionsSchema = z.object({
+	dispatchNamespaces: z
+		.record(
+			z.object({
+				namespace: z.string(),
+				mixedModeConnectionString: z
+					.custom<MixedModeConnectionString>()
+					.optional(),
+			})
+		)
+		.optional(),
+});
+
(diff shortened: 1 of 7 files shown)


packages/miniflare/src/workers/dispatch-namespace/dispatch-namespace.worker.ts
```diff
@@ -0,0 +1,30 @@
+interface Env {
+	fetcher: Fetcher;
+}
+class LocalDispatchNamespace implements DispatchNamespace {
```
dario-piotrowicz, line 4:
```suggestion

class LocalDispatchNamespace implements DispatchNamespace {
```

packages/miniflare/src/plugins/dispatch-namespace/index.ts
```diff
@@ -0,0 +1,101 @@
...
+							name
+						),
+					};
+				}
+			),
+			extensions: [
+				{
+					modules: [
+						{
+							name: `${DISPATCH_NAMESPACE_PLUGIN_NAME}:local-namespace`,
```
dario-piotrowicz, line 92: ultra nit 😅 

```suggestion
							name: `${DISPATCH_NAMESPACE_PLUGIN_NAME}:local-dispatch-namespace`,
```
maybe?

dario-piotrowicz (approved):
Awesome! 🤩 🚀

PR #9257, opened 2025-05-15
Revert R2 bucket validation for `pages dev` commands

penalosa:
Fixes https://github.com/cloudflare/workers-sdk/issues/9253

Relax R2 bucket validation for `pages dev` commands

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: already documented like this
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: fix to a v3 change

PR #9259, opened 2025-05-15
V3 Backport [#9257]: Revert R2 bucket validation for  commands

workers-devprod:
This is an automatically opened PR to backport patch changes from #9257 to Wrangler v3


edmundhung:
We backported #9172 to v3. So we will need to backport this too.

edmundhung (approved):

PR #9234, opened 2025-05-13
add no-op `ctx.props` to getPlatformProxy to fix types mismatch

emily-shen:
Fixes #9233

I don't think `props` can be meaningfully used with getPlatformProxy? So this just fixes the types issue. 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: covered by fixtures
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: bugfix
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/9262
  - [ ] Not necessary because:

PR #9262, opened 2025-05-15
V3 Backport [#9234]: add no-op  to getPlatformProxy to fix types mismatch

workers-devprod:
This is an automatically opened PR to backport patch changes from #9234 to Wrangler v3


emily-shen (approved):

edmundhung:
@emily-shen Do we confirm if the workerd version used in v3 has `ctx.props` defined?

PR #9277, opened 2025-05-15
Support Mixed Mode for more binding types

penalosa:
Fixes [DEVX-1859](https://jira.cfdata.org/browse/DEVX-1859)

Support Mixed Mode in Miniflare for more bindings: KV, R2, D1, Images, Vectorize. This also involves supporting "raw" mode for bindings when using `startMixedModeSession()`, and adding a new `remote: "minimal"` mode (which doesn't do anything today, but might in future).

There are some things to note:
- Images local mode has been moved to Miniflare, but Wrangler still injects it's own via a wrapped binding. This can be addressed in a followup. In contrast to the Wrangler implementation, Miniflare installs `sharp` by default, because Miniflare & Wrangler no longer support Node.js v16 (which `sharp` does not support, which necessitated dynamic loading in Wrangler)
- Some groundwork has been done for Queues, Workflows, and mTLS bindings (raw mode support), but they will require followup work in Miniflare
- Integration with Wrangler is not part of this PR
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


dario-piotrowicz (commented):
First round of reviews 🙂 

I think we should keep track of the DevProd Testing remote resources that we rely on in the e2es, I've created https://jira.cfdata.org/browse/DEVX-1902 for that, please let me know if I've missed anything

Also I think we should discuss the API changes 🥲

packages/miniflare/src/plugins/images/fetcher.ts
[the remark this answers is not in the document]
penalosa: This code duplication shouldn't stick around for very long—Wrangler will use the Miniflare version of the local images binding in a followup (I just wanted to reduce the scope of this PR). I've added a ticket so we don't forget: https://jira.cfdata.org/browse/DEVX-1904

fixtures/mixed-mode-node-test/tests/startMixedModeSession.test.js
```diff
@@ -165,9 +165,271 @@ baseDescribe("startMixedModeSession", () => {
...
+    await env.KV_BINDING.put('KEY', 'VALUE');
+
+	// read a key-value pair
+	const value = await env.KV_BINDING.get('KEY');
+
+	// list all key-value pairs
+	const allKeys = await env.KV_BINDING.list();
+
+	// delete a key-value pair
+	// await env.KV_BINDING.delete('KEY');
```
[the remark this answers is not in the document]
penalosa: It's not being asserted, but it is running—and the key thing I was trying to go for here was that you could use the KV API surfaces without errors being thrown (because the mixed mode connection wasn't configured properly)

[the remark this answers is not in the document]
dario-piotrowicz: I see 🙂, it sounds a bit overkill to me since if `KV_BINDING.get` works then I can't imagine how the other method could not 🤔, but it doesn't really cause much harm so I am happy to leave it in if you prefer it that way 🙂

packages/wrangler/e2e/seed-files/mixed-mode-workers/images.js
```diff
@@ -0,0 +1,16 @@
+export default {
+	async fetch(request, env) {
+		const image = await fetch(
+			// There's nothing special about this image—it's just a random publicly accessible PNG
+			"https://cf-assets.www.cloudflare.com/dzlvafdwdttg/73IDGdh9xNkGVXija5vC3b/04cbc66d20d6bade310f270f0f1bd834/Frame_1.png"
+		);
+
+		const response = (
+			await env.IMAGES.input(image.body)
+				// Without Mixed mode, this fails with: Error: IMAGES_INFO_ERROR 9520: ERROR: Unsupported image type
```
dario-piotrowicz, line 10: 🚀

packages/wrangler/e2e/seed-files/mixed-mode-workers/dispatch-namespace.js
```diff
@@ -0,0 +1,13 @@
+export default {
+	async fetch(request, env) {
+		try {
+			const worker = env.DISPATCH.get("mixed-mode-test-customer-worker");
+			return Response.json({
+				worker: await (await worker.fetch("http://example.com")).text(),
+			});
+		} catch (e) {
+			console.log(e);
```
dario-piotrowicz, line 9: is the `console.log` intentional?

packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
```diff
@@ -0,0 +1,355 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
dario-piotrowicz, line 1: Disabling the eslint rule for the whole file feels a bit weird to me, since this is only needed for the `setupResult` variable, what about doing this instead?

```diff
--- a/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
+++ b/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
@@ -1,5 +1,5 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
 /* eslint-disable @typescript-eslint/consistent-type-imports */
+import assert from "node:assert";
 import path from "node:path";
 import dedent from "ts-dedent";
 import {
@@ -308,9 +308,11 @@ describe.each(testCases)("Mixed Mode for $name", (testCase) => {
                );
                const setupResult = await testCase.setup?.(helper);
 
+               assert(setupResult);
+
                const mixedModeSession = await experimental_startMixedModeSession(
                        typeof testCase.mixedModeSessionConfig === "function"
-                               ? testCase.mixedModeSessionConfig(setupResult!)
+                               ? testCase.mixedModeSessionConfig(setupResult)
                                : testCase.mixedModeSessionConfig
                );
 
@@ -322,7 +324,7 @@ describe.each(testCases)("Mixed Mode for $name", (testCase) => {
                        modulesRoot: helper.tmpPath,
                        ...testCase.miniflareConfig(
                                mixedModeSession.mixedModeConnectionString,
-                               setupResult!
+                               setupResult
                        ),
                });
                expect(await (await mf.dispatchFetch("http://example.com")).text()).toEqual(
```

packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
```diff
@@ -0,0 +1,355 @@
...
+/* eslint-disable @typescript-eslint/consistent-type-imports */
```
dario-piotrowicz, line 2: why? why not just auto-fixing the imports? 🤔 

(I'm getting the feeling that maybe you committed these two `eslint-disable` rules by mistake 🤔)

dario-piotrowicz (approved):
Wow! Amazing work!!! 🫶 🚀

packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
penalosa: That won't work, because the setupResult can sometimes be undefined (it's just that in that case the test functions don't expect it). I tried to fudge around with TS to get this to work, but since it was in test code I didn't want to spend too much time on it. I can add the eslint rule to the specific lines, maybe?

packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
penalosa: Oops, yeah, the consistent-type-imports rule can be removed, I think

PR #9280, opened 2025-05-15
Add support for static routing (i.e. \_routes.json) to Workers Assets

matthewdavidrodgers:
For WC-3583

Adds support for static routing (i.e. \_routes.json) to Workers Assets

Implements the proposal noted here https://github.com/cloudflare/workers-sdk/discussions/9143

In brief: when static routing is present for a Worker with assets, routing via those static rules takes precedence. When a request is evaluated in the Router Worker, the request path is first compared to the "exclude" rules. If any match, the request is forwarded directly to the Asset Worker. If instead any "include" rules match, the request is forwarded directly to the User Worker. If neither match (or static routing was not provided), the existing behavior takes over.

As part of this explicit routing, when static routing is present, the check against "Sec-Fetch-Mode: navigate" (to guess if this should serve an asset or go to the User Worker for not\_found\_handling) is disabled. Routing can be controlled by uploading a \_routes.json, and asset serving (including when an index.html or 404.html page is served) will be more simple.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not a wrangler/vite change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: will be handled separately
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Not a wrangler change

### DIFF
--- a/packages/workers-shared/asset-worker/src/utils/rules-engine.ts
+++ b/packages/workers-shared/asset-worker/src/utils/rules-engine.ts
@@ -28,6 +28,43 @@ export const replacer = (str: string, replacements: Replacements) => {
 	return str;
 };
 
+export const generateGlobOnlyRuleRegExp = (rule: string) => {
+	// Escape all regex characters other than globs (the "*" character) since that's all that's supported.
+	rule = rule.split("*").map(escapeRegex).join(".*");
+
+	// Wrap in line terminators to be safe.
+	rule = "^" + rule + "$";
+
+	return RegExp(rule);
+};
+
+export const generateRuleRegExp = (rule: string) => {
+	// Create :splat capturer then escape.
+	rule = rule.split("*").map(escapeRegex).join("(?<splat>.*)");
+
+	// Create :placeholder capturers (already escaped).
+	// For placeholders in the host, we separate at forward slashes and periods.
+	// For placeholders in the path, we separate at forward slashes.
+	// This matches the behavior of URLPattern.
+	// e.g. https://:subdomain.domain/ -> https://(here).domain/
+	// e.g. /static/:file -> /static/(image.jpg)
+	// e.g. /blog/:post -> /blog/(an-exciting-post)
+	const host_matches = rule.matchAll(HOST_PLACEHOLDER_REGEX);
+	for (const host_match of host_matches) {
+		rule = rule.split(host_match[0]).join(`(?<${host_match[1]}>[^/.]+)`);
+	}
+
+	const path_matches = rule.matchAll(PLACEHOLDER_REGEX);
+	for (const path_match of path_matches) {
+		rule = rule.split(path_match[0]).join(`(?<${path_match[1]}>[^/]+)`);
+	}
+
+	// Wrap in line terminators to be safe.
+	rule = "^" + rule + "$";
+
+	return RegExp(rule);
+};
+
 export const generateRulesMatcher = <T>(
 	rules?: Record<string, T>,
 	replacerFn: (match: T, replacements: Replacements) => T = (match) => match
@@ -40,31 +77,8 @@ export const generateRulesMatcher = <T>(
 		.map(([rule, match]) => {
 			const crossHost = rule.startsWith("https://");
 
-			// Create :splat capturer then escape.
-			rule = rule.split("*").map(escapeRegex).join("(?<splat>.*)");
-
-			// Create :placeholder capturers (already escaped).
-			// For placeholders in the host, we separate at forward slashes and periods.
-			// For placeholders in the path, we separate at forward slashes.
-			// This matches the behavior of URLPattern.
-			// e.g. https://:subdomain.domain/ -> https://(here).domain/
-			// e.g. /static/:file -> /static/(image.jpg)
-			// e.g. /blog/:post -> /blog/(an-exciting-post)
-			const host_matches = rule.matchAll(HOST_PLACEHOLDER_REGEX);
-			for (const host_match of host_matches) {
-				rule = rule.split(host_match[0]).join(`(?<${host_match[1]}>[^/.]+)`);
-			}
-
-			const path_matches = rule.matchAll(PLACEHOLDER_REGEX);
-			for (const path_match of path_matches) {
-				rule = rule.split(path_match[0]).join(`(?<${path_match[1]}>[^/]+)`);
-			}
-
-			// Wrap in line terminators to be safe.
-			rule = "^" + rule + "$";
-
 			try {
-				const regExp = new RegExp(rule);
+				const regExp = generateRuleRegExp(rule);
 				return [{ crossHost, regExp }, match];
 			} catch {}
 		})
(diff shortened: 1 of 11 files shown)


emily-shen (approved):

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1325,41 +1325,56 @@ describe("[Asset Worker] `canFetch`", () => {
...
+								hasStaticRouting[1],
+						});
+					}
 				}
 			}
 		}
 
 		it.each(matrix)(
-			"compatibility_date $compatibilityDate, compatibility_flags $compatibilityFlags, not_found_handling $notFoundHandling, headers: $headers -> $expected",
+			"compatibility_date $compatibilityDate, compatibility_flags $compatibilityFlags, not_found_handling $notFoundHandling, headers: $headers, hasStaticRouting: hasStaticRouting -> $expected",
```
GregBrimble, line 1356:
```suggestion
			"compatibility_date $compatibilityDate, compatibility_flags $compatibilityFlags, not_found_handling $notFoundHandling, headers: $headers, hasStaticRouting $hasStaticRouting -> $expected",
```

packages/workers-shared/asset-worker/src/utils/rules-engine.ts
```diff
@@ -131,3 +135,19 @@ export const generateRedirectsMatcher = (
...
 		})
 	);
+
+export const generateStaticRoutingRuleMatcher =
+	(rules: string[]) =>
+	({ request }: { request: Request }) => {
+		const { pathname } = new URL(request.url);
+		for (const rule of rules) {
+			try {
+				const regExp = generateRuleRegExp(rule);
```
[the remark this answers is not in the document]
matthewdavidrodgers: good call - other characters escaped

packages/workers-shared/asset-worker/tests/rules-engine.test.ts
```diff
@@ -115,3 +119,83 @@ describe("replacer", () => {
 		);
 	});
 });
+
+describe("static routing rules", () => {
```
GregBrimble, line 123: If you wanted a bunch of pre-built tests, you could grab these: https://cflare.co/url-glob-test. I think that's what Pages' stuff actually runs against today anyway.

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -75,6 +77,69 @@ export default {
...
+					config.static_routing.exclude ?? []
+				);
+				if (
+					excludeRulesMatcher({
+						request,
+					})
+				) {
+					// direct to asset worker
+					analytics.setData({
+						dispatchtype: DISPATCH_TYPE.WORKER,
```
GregBrimble, line 92:
```suggestion
						dispatchtype: DISPATCH_TYPE.ASSETS,
```

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -75,6 +77,69 @@ export default {
 
 			const maybeSecondRequest = request.clone();
 
+			if (config.static_routing) {
```
GregBrimble, line 80: Do we want to time/trace static routing evaluation?

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -75,6 +77,69 @@ export default {
...
+					excludeRulesMatcher({
+						request,
+					})
+				) {
+					// direct to asset worker
+					analytics.setData({
+						dispatchtype: DISPATCH_TYPE.WORKER,
+						staticRoutingDecision: STATIC_ROUTING_DECISION.EXCLUDE,
+					});
+					return await env.JAEGER.enterSpan("dispatch_worker", async (span) => {
```
GregBrimble, line 95:
```suggestion
					return await env.JAEGER.enterSpan("dispatch_assets", async (span) => {
```

I'm not sure what the purpose of this + `dispatchType` tag is, but whatever, it seems to be how we're doing asset dispatches down below anyway.

matthewdavidrodgers: i agree - the span name and the label are redundant, but yeah i was just mirroring existing behavior

packages/workers-shared/router-worker/src/index.ts
matthewdavidrodgers: i don't think that's necessary, since it should only be on the order of milliseconds. at the very least we'd be able to infer static routing cost between the start of the request context and any dispatch span. but if you think it'd be an asset i'm happy to add it

packages/workers-shared/asset-worker/tests/rules-engine.test.ts
matthewdavidrodgers: nice - not able to steal all of them due to the fact that the matcher only matches on Requests, but i stole a big chunk

packages/workers-shared/utils/types.ts
```diff
@@ -6,8 +6,15 @@ const InternalConfigSchema = z.object({
 	debug: z.boolean().optional(),
 });
 
+const StaticRoutingSchema = z.object({
+	version: z.number(),
```
GregBrimble, line 10:
```suggestion
	version: z.literal(1),
```

packages/workers-shared/router-worker/src/index.ts
GregBrimble: Nope, fine to not bother for now! We can always add in the future if we need to dig deeper. Just wanted to flag for consideration.

packages/workers-shared/router-worker/src/index.ts
```diff
@@ -98,7 +163,9 @@ export default {
 			}
 
 			// If we have a user-Worker, but no assets, dispatch to Worker script
-			const assetsExist = await env.ASSET_WORKER.unstable_canFetch(request);
+			const assetsExist = await env.ASSET_WORKER.unstable_canFetch(request, {
+				hasStaticRouting,
```
GregBrimble, line 167: hmm. the _intention_ of this API was that it would be exposed to users, but we never quite landed it (hence `unstable_`). I don't love having clients having to know anything about "hasStaticRouting". I wonder if we can obfuscate that at all? Either asset worker also gets this prop in its config, or we skirt something into the request? Idk.

Not blocking, because this isn't a public API. But if there is a tidier way to do it, I'd like to do that if we can.

GregBrimble: 😬 How about stripping the `Sec-Fetch-Mode` header off?

GregBrimble (commented):
Yeah, everything else LGTM if we can obfuscate away the `hasStaticRouting` param arg from `unstable_canFetch` 👍

packages/workers-shared/router-worker/src/index.ts
matthewdavidrodgers: Yeah, of these options I think I like including `has_static_routing` in the asset configuration. Will update here

GregBrimble (approved):
ship it

PR #9306, opened 2025-05-19
chore: run ci for external forks by commit sha

edmundhung:
Fixes n/a.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: ci change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: ci change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: ci change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: ci change

### DIFF
--- a/.github/workflows/run-ci-for-external-forks.yml
+++ b/.github/workflows/run-ci-for-external-forks.yml
@@ -5,6 +5,9 @@ on:
       pr-number:
         description: "The PR number to run CI on behalf of"
         required: true
+      commit-sha:
+        description: "The specific commit SHA from the PR branch to run CI on"
+        required: true
       reviewed:
         description: "Confirm that the PR has been reviewed for use/leakage of secrets"
         type: boolean
@@ -32,13 +35,15 @@ jobs:
           team: wrangler
 
       - name: Stop workflow if user is not a team member
-        if: ${{ steps.teamAffiliation.outputs.isTeamMember == false }}
+        if: ${{ steps.teamAffiliation.outputs.isTeamMember == 'false' }}
         run: |
-          echo "You must be on the "wrangler" team to trigger this job."
+          echo "You must be on the \"wrangler\" team to trigger this job."
           exit 1
 
       - name: "Checkout PR"
-        run: gh pr checkout ${{ inputs.pr-number }} -b run-ci-on-behalf-of-${{ inputs.pr-number }} -f
+        run: |
+          gh pr checkout ${{ inputs.pr-number }} --branch run-ci-on-behalf-of-${{ inputs.pr-number }}
+          git reset --hard ${{ inputs.commit-sha }}
         env:
           GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -32,13 +35,16 @@ jobs:
           team: wrangler
 
       - name: Stop workflow if user is not a team member
-        if: ${{ steps.teamAffiliation.outputs.isTeamMember == false }}
+        if: ${{ steps.teamAffiliation.outputs.isTeamMember == 'false' }}
```
edmundhung, line 38: Looks like this never worked before :grimacing:

[Reference from README on tspascoal/get-user-teams-membership](https://github.com/tspascoal/get-user-teams-membership/blob/ba78054988f58bea69b7c6136d563236f8ed2fc0/README.md?plain=1#L67)

dario-piotrowicz (approved):

.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -32,13 +35,15 @@ jobs:
...
```
petebacondarwin, line 39: Would it be safer to do:

```suggestion
        if: ${{ steps.teamAffiliation.outputs.isTeamMember != 'true' }}
```

.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -32,13 +35,15 @@ jobs:
...
         run: |
-          echo "You must be on the "wrangler" team to trigger this job."
+          echo "You must be on the \"wrangler\" team to trigger this job."
           exit 1
 
       - name: "Checkout PR"
-        run: gh pr checkout ${{ inputs.pr-number }} -b run-ci-on-behalf-of-${{ inputs.pr-number }} -f
+        run: |
+          gh pr checkout ${{ inputs.pr-number }} --branch run-ci-on-behalf-of-${{ inputs.pr-number }}
+          git reset --hard ${{ inputs.commit-sha }}
```
petebacondarwin: Can we check what happens if this fails? What if we get the SHA wrong?
Is there a chance it will quietly fail and not reset the HEAD, leaving us at the potentially problematic HEAD commit?

.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -24,29 +28,42 @@ jobs:
...
       - name: "Checkout PR"
-        run: gh pr checkout ${{ inputs.pr-number }} -b run-ci-on-behalf-of-${{ inputs.pr-number }} -f
+        run: |
+          gh pr checkout "$PR_NUM" --branch "run-ci-on-behalf-of-$PR_NUM"
+          git reset --hard "$COMMIT_SHA"
         env:
+          # We need a PAT to checkout the fork
           GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+          PR_NUM: ${{ inputs.pr-number }}
+          COMMIT_SHA: ${{ inputs.commit-sha }}
```
edmundhung, line 52: Moving all the inputs to env var as suggested in https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable

petebacondarwin (changes requested):
Requesting changes only till we check that the `git reset` is guaranteed to work.

.github/workflows/run-ci-for-external-forks.yml
petebacondarwin: As long as the workflow is not configured somehow to try to keep going... But I think we are good.
We can always test it quite easily once we have landed this.

petebacondarwin (approved):

.github/workflows/run-ci-for-external-forks.yml
dario-piotrowicz: @petebacondarwin I think we're all good 🙂👍

PR #9322, opened 2025-05-21
fix(vite-plugin): Support `__` in additional module IDs

justinvdm:
## Problem

There's a regex used to detect if certain files (wasm, .bin, .txt, .html) are being imported at runtime. The current regex looks like this:

```ts
const ADDITIONAL_MODULE_PATTERN = `__CLOUDFLARE_MODULE__(${ADDITIONAL_MODULE_TYPES.join("|")})__(.*?)__`;
```

This means if there are files with `__`s in them, those `__`s will get matched first, rather than the final `__` in the module pattern being matched on.

 ## Solution
Add a `$` to the end of the regex so, to anchor to the end of the string.

 ## Background context
`pnpm` stores files for packages in locations like this (note the `__`s), which uncovered this.

```
/path/to/project/node_modules/.pnpm/@prisma+client@6.8.2_prisma@6.8.2_typescript@5.8.3__typescript@5.8.3/node_modules/@prisma/client/runtime/query_engine_bg.sqlite.wasm
```

### Update
Adding the `$` turned out to break the build step - where we need to now match on the entire code string for a chunk, rather than the virtual module path itself as a string: https://github.com/cloudflare/workers-sdk/pull/9322#issuecomment-2897792616

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: No API change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/neat-jokes-prove.md
+++ b/.changeset/neat-jokes-prove.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+Fix regex to correctly detect additional module imports with \_\_ in path
--- a/packages/vite-plugin-cloudflare/playground/additional-modules/__tests__/additional-modules.spec.ts
+++ b/packages/vite-plugin-cloudflare/playground/additional-modules/__tests__/additional-modules.spec.ts
@@ -22,6 +22,11 @@ test("supports Text modules with a '.txt' extension", async () => {
 	expect(result).toBe("Example text content.\n");
 });
 
+test("supports modules with `__`s in the filename", async () => {
+	const result = await getTextResponse("/text2");
+	expect(result).toBe("Example text content 2");
+});
+
 test("supports CompiledWasm modules with a '.wasm' extension", async () => {
 	const result = await getJsonResponse("/wasm");
 	expect(result).toEqual({ result: 7 });
--- a/packages/vite-plugin-cloudflare/playground/additional-modules/src/index.ts
+++ b/packages/vite-plugin-cloudflare/playground/additional-modules/src/index.ts
@@ -1,5 +1,6 @@
 import bin from "./modules/bin-example.bin";
 import html from "./modules/html-example.html";
+import text2 from "./modules/text__example__2.txt";
 import text from "./modules/text-example.txt";
 import wasm from "./modules/wasm-example.wasm";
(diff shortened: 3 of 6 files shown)


justinvdm:
Hi @jamesopstad, ran into this when trying out Prisma 6.8.2 + its new `prisma-client` generator with RedwoodSDK. Can you take a look when you have a moment?

petebacondarwin (commented):
Thanks for this fix @justinvdm - would you be able to tweak the import in the additional-modules playground to demonstrate this bug and that this change fixes it?

Perhaps change the file name of this import to contain `__`?

https://github.com/cloudflare/workers-sdk/blob/06410ae066bb28ffa161ecfa6670788bff9c2f6b/packages/vite-plugin-cloudflare/playground/additional-modules/src/index.ts#L4

justinvdm:
@jamesopstad Ah right! Will make that regex more resilient and update the playground to test the `__` case. Thanks!

justinvdm:
@jamesopstad Alright I _think_ we are now good to go.
* https://github.com/cloudflare/workers-sdk/pull/9322/commits/50093d47a636daef151e3f8b482e30fe9c27c999 reverts the changes (but keeps the new test) to demonstrate the failure case
* https://github.com/cloudflare/workers-sdk/pull/9322/commits/28265bc60d422392738646f9a5b8f22215f45bc8 reverts the revert so we have both the fix and the new test, to demonstrate we're going from failure to passing test

I checked this locally, but I'll need your help to run CI for these commits to check this on CI - If I looked correctly it looks like running of the CI checks requires the maintainer's approval. Can you help me with that?

jamesopstad (approved):
Looks good! Please could you just add a changeset (`pnpm changeset`) describing what the fix is and then I think we'll be good to go. Thanks.

justinvdm:
> Looks good! Please could you just add a changeset (`pnpm changeset`) describing what the fix is and then I think we'll be good to go. Thanks.

Thanks @jamesopstad, done 👍

PR #9330, opened 2025-05-21
feat: introduce defaultPersistRoot option

edmundhung:
Fixes [DEVX-1828](https://jira.cfdata.org/browse/DEVX-1828)

This introduces a `defaultPersistRoot` setting that, when provided, is used as the base directory for any plugin (i.e. `path.join(defaultPersistRoot, pluginName)`) whose persist option is `undefined` or `true`. Explicit string values for a plugin’s persist path are still honored.

This removes the need for both Wrangler and the Vite plugin to set individual persist options as long as all the miniflare plugins continue to derive the persist path using the `getPersistPath` helper. This will not break existing users as well as the result paths match current settings on both Wrangler / Vite plugin.

Example:
```tsx
// Without `defaultPersistRoot` / Before this change:
new Miniflare({
  kvPersist: undefined,       // → "/(tmp)/kv"
  d1Persist: true,            // → "/.mf/d1"
  r2Persist: false,           // → "/(tmp)/d1"
  cachePersist: "/my-cache",  // → "/my-cache"
});

// With `defaultPersistRoot`
new Miniflare({
  defaultPersistRoot: "/storage",
  kvPersist: undefined,       // → "/storage/kv"
  d1Persist: true,            // → "/storage/d1"
  r2Persist: false,           // → "/(tmp)/d1"
  cachePersist: "/my-cache",  // → "/my-cache"
});
```

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not patch

### DIFF
--- a/packages/miniflare/src/plugins/secret-store/index.ts
+++ b/packages/miniflare/src/plugins/secret-store/index.ts
@@ -67,7 +67,13 @@ export const SECRET_STORE_PLUGIN: Plugin<
 			])
 		);
 	},
-	async getServices({ options, sharedOptions, tmpPath, unsafeStickyBlobs }) {
+	async getServices({
+		options,
+		sharedOptions,
+		tmpPath,
+		defaultPersistRoot,
+		unsafeStickyBlobs,
+	}) {
 		const configs = options.secretsStoreSecrets
 			? Object.values(options.secretsStoreSecrets)
 			: [];
@@ -79,6 +85,7 @@ export const SECRET_STORE_PLUGIN: Plugin<
 		const persistPath = getPersistPath(
 			SECRET_STORE_PLUGIN_NAME,
 			tmpPath,
+			defaultPersistRoot,
 			sharedOptions.secretsStorePersist
 		);
 
--- a/packages/miniflare/src/plugins/secret-store/index.ts
+++ b/packages/miniflare/src/plugins/secret-store/index.ts
@@ -169,4 +169,12 @@ export const SECRET_STORE_PLUGIN: Plugin<
 
 		return [...services, storageService, objectService];
 	},
+	getPersistPath({ secretsStorePersist }, tmpPath) {
+		return getPersistPath(
+			SECRET_STORE_PLUGIN_NAME,
+			tmpPath,
+			undefined,
+			secretsStorePersist
+		);
+	},
 };
--- a/packages/miniflare/src/plugins/shared/index.ts
+++ b/packages/miniflare/src/plugins/shared/index.ts
@@ -71,6 +71,7 @@ export interface PluginServicesOptions<
 	workerIndex: number;
 	additionalModules: Worker_Module[];
 	tmpPath: string;
+	defaultPersistRoot: string | undefined;
 	workerNames: string[];
 	loopbackPort: number;
 	unsafeStickyBlobs: boolean;
@@ -182,6 +183,7 @@ export function maybeParseURL(url: Persistence): URL | undefined {
 export function getPersistPath(
 	pluginName: string,
 	tmpPath: string,
+	defaultPersistRoot: string | undefined,
 	persist: Persistence
 ): string {
 	// If persistence is disabled, use "memory" storage. Note we're still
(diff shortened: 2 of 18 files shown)


packages/miniflare/src/plugins/shared/index.ts
```diff
@@ -191,10 +193,15 @@ export function getPersistPath(
...
 	const memoryishPath = path.join(tmpPath, pluginName);
-	if (persist === undefined || persist === false) {
+	if (persist === false) {
 		return memoryishPath;
 	}
 
+	// If `persist` is undefined, use either the default path or fallback to the tmpPath
+	if (persist === undefined) {
+		return path.join(defaultPersistRoot ?? tmpPath, pluginName);
+	}
```
petebacondarwin, line 203: So I think you are changing the default for Miniflare, right?
Now if you don't provide a persist path then instead of it being in-memory it writes to the default persist path.

edmundhung: Before this change, when `persist` is undefined, it will always be the in-memory path which is based on the `tmpPath`. So this changes the behavior only if `defaultPersistRoot` is set.

packages/miniflare/src/plugins/secret-store/index.ts
```diff
@@ -67,7 +67,13 @@ export const SECRET_STORE_PLUGIN: Plugin<
 			])
 		);
 	},
-	async getServices({ options, sharedOptions, tmpPath, unsafeStickyBlobs }) {
+	async getServices({
```
petebacondarwin, line 70: Out of interest, why does secret-store not implement `getPersistPath()`

packages/miniflare/src/plugins/shared/index.ts
```diff
@@ -191,10 +193,15 @@ export function getPersistPath(
...
 	// destroy on `dispose()`.
 	const memoryishPath = path.join(tmpPath, pluginName);
-	if (persist === undefined || persist === false) {
+	if (persist === false) {
 		return memoryishPath;
 	}
 
+	// If `persist` is undefined, use either the default path or fallback to the tmpPath
+	if (persist === undefined) {
+		return path.join(defaultPersistRoot ?? tmpPath, pluginName);
```
petebacondarwin, line 202: This would be easier for me to follow if it was:

```suggestion
		return defaultPersistRoot === undefined ? memoryishPath : path.join(defaultPersistRoot, pluginName);
```

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -949,25 +949,12 @@ export function buildMiniflareBindingOptions(
...
-		return {
-			cachePersist: path.join(v3Path, "cache"),
-			durableObjectsPersist: path.join(v3Path, "do"),
-			kvPersist: path.join(v3Path, "kv"),
-			r2Persist: path.join(v3Path, "r2"),
-			d1Persist: path.join(v3Path, "d1"),
-			workflowsPersist: path.join(v3Path, "workflows"),
-			secretsStorePersist: path.join(v3Path, "secrets-store"),
-			analyticsEngineDatasetsPersist: path.join(v3Path, "analytics-engine"),
-		};
```
petebacondarwin, line 970: Is the idea that while one could still provide each of these options individually, we can now just provide the base path and these all get inferred naturally?

petebacondarwin (approved):

packages/miniflare/src/plugins/secret-store/index.ts
edmundhung: Good question. I think the one who implemented the secret store binding didn't know that exists until now

cc: @edmundhung

(I will fix that before merging the PR)

packages/wrangler/src/dev/miniflare.ts
edmundhung: Exactly!

packages/miniflare/src/plugins/secret-store/index.ts
edmundhung: Fixed in https://github.com/cloudflare/workers-sdk/pull/9330/commits/2d196006ad7849862c62e9334119e29fb81dc563

PR #9335, opened 2025-05-22
Redesign `wrangler dev`

penalosa:
Fixes https://jira.cfdata.org/browse/DEVX-1911

A fairly big redesign of `wrangler dev`, mostly around bindings display. Here's an example of what a dev session will look like:
Screenshot 2025-05-22 at 01 11 43

Additionally, the hotkeys have been shifted to be pinned to the top (I think this looks better, but it's definitely potentially controversial), and I've removed the local/remote toggle (also potentially controversial, but mixed mode provides a much better solution here).

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: I don't think we have any screenshots of this flow in docs?
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: part of the not-being-backported mixed mode feature

### DIFF
--- a/packages/miniflare/src/shared/log.ts
+++ b/packages/miniflare/src/shared/log.ts
@@ -7,11 +7,11 @@ const cwdNodeModules = path.join(cwd, "node_modules");
 
 const LEVEL_PREFIX: { [key in LogLevel]: string } = {
 	[LogLevel.NONE]: "",
-	[LogLevel.ERROR]: "err",
-	[LogLevel.WARN]: "wrn",
-	[LogLevel.INFO]: "inf",
-	[LogLevel.DEBUG]: "dbg",
-	[LogLevel.VERBOSE]: "vrb",
+	[LogLevel.ERROR]: "error",
+	[LogLevel.WARN]: "warn",
+	[LogLevel.INFO]: "info",
+	[LogLevel.DEBUG]: "debug",
+	[LogLevel.VERBOSE]: "verbose",
 };
 
 const LEVEL_COLOUR: { [key in LogLevel]: Colorize } = {
@@ -83,7 +83,7 @@ export class Log {
 	logWithLevel(level: LogLevel, message: string): void {
 		if (level <= this.level) {
 			const prefix = `[${this.#prefix}${LEVEL_PREFIX[level]}${this.#suffix}]`;
-			this.log(LEVEL_COLOUR[level](`${prefix} ${message}`));
+			this.log(`${LEVEL_COLOUR[level](prefix)} ${message}`);
 		}
 	}
 
--- a/packages/wrangler/e2e/deployments.test.ts
+++ b/packages/wrangler/e2e/deployments.test.ts
@@ -285,7 +285,6 @@ Uploaded 2 of 3 assets
 Uploaded 3 of 3 assets
 ✨ Success! Uploaded 3 files (TIMINGS)
 Total Upload: xx KiB / gzip: xx KiB
-No bindings found.
 Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
 Deployed tmp-e2e-worker-00000000-0000-0000-0000-000000000000 triggers (TIMINGS)
   https://tmp-e2e-worker-00000000-0000-0000-0000-000000000000.SUBDOMAIN.workers.dev
@@ -374,8 +373,8 @@ Uploaded 3 of 3 assets
 ✨ Success! Uploaded 3 files (TIMINGS)
 Total Upload: xx KiB / gzip: xx KiB
 Your Worker has access to the following bindings:
-- Assets:
-  - Binding: ASSETS
+Binding            Resource
+env.ASSETS         Assets
 Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
 Deployed tmp-e2e-worker-00000000-0000-0000-0000-000000000000 triggers (TIMINGS)
   https://tmp-e2e-worker-00000000-0000-0000-0000-000000000000.SUBDOMAIN.workers.dev
@@ -541,8 +540,8 @@ Uploaded 3 of 3 assets
 ✨ Success! Uploaded 3 files (TIMINGS)
 Total Upload: xx KiB / gzip: xx KiB
 Your Worker has access to the following bindings:
-- Assets:
-  - Binding: ASSETS
+Binding            Resource
+env.ASSETS         Assets
 Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
 Deployed tmp-e2e-worker-00000000-0000-0000-0000-000000000000 triggers (TIMINGS)
   https://tmp-e2e-worker-00000000-0000-0000-0000-000000000000.SUBDOMAIN.workers.dev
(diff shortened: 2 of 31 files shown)


packages/miniflare/src/shared/log.ts
```diff
@@ -7,11 +7,11 @@ const cwdNodeModules = path.join(cwd, "node_modules");
...
-	[LogLevel.ERROR]: "err",
-	[LogLevel.WARN]: "wrn",
-	[LogLevel.INFO]: "inf",
-	[LogLevel.DEBUG]: "dbg",
-	[LogLevel.VERBOSE]: "vrb",
+	[LogLevel.ERROR]: "error",
+	[LogLevel.WARN]: "warn",
+	[LogLevel.INFO]: "info",
+	[LogLevel.DEBUG]: "debug",
+	[LogLevel.VERBOSE]: "verbose",
```
penalosa, line 14: This shortening wasn't necessary, and adding a few extra characters makes these clearer

packages/wrangler/e2e/deployments.test.ts
```diff
@@ -285,7 +285,6 @@ Uploaded 2 of 3 assets
 Uploaded 3 of 3 assets
 ✨ Success! Uploaded 3 files (TIMINGS)
 Total Upload: xx KiB / gzip: xx KiB
-No bindings found.
```
penalosa, line 288: This is an unrelated-to-mixed-mode change that @threepointone suggested, that I thought I'd do while in the same area of the codebase. It removes the `No bindings found.` message unless a dry-run is happening

packages/wrangler/e2e/pages-dev.test.ts
```diff
@@ -11,300 +11,297 @@ import { normalizeOutput } from "./helpers/normalize";
...
-		`
-			);
 		});
-
-		it("should support wrangler.toml", async () => {
-			const helper = new WranglerE2ETestHelper();
-			await helper.seed({
-				"public/_worker.js": dedent`
+		const worker = helper.runLongLived(
+			`${cmd} --inspector-port ${inspectorPort} . --port ${port} --service TEST_SERVICE=test-worker --kv TEST_KV --do TEST_DO=TestDurableObject@a --d1 TEST_D1 --r2 TEST_R2 --compatibility-date=2025-05-21`
```
penalosa, line 116: The compat date just removes a previously logged warning—there are no functional changes

packages/wrangler/e2e/provision.test.ts
```diff
@@ -133,12 +131,10 @@ describe("provisioning", { timeout: TIMEOUT }, () => {
...
-			- KV Namespaces:
-			  - KV
-			- D1 Databases:
-			  - D1
-			- R2 Buckets:
-			  - R2
+			Binding                 Resource
+			env.KV (inherited)      KV Namespace
+			env.D1 (inherited)      D1 Database
+			env.R2 (inherited)      R2 Bucket
```
penalosa, line 137: This was not the main point of this PR, but I thought I might as well clean up some of the provisioning logging while I was at it

packages/wrangler/src/__tests__/cli-hotkeys.test.ts
```diff
@@ -266,11 +260,11 @@ describe("Hot Keys", () => {
 				const unregisterHotKeys = registerHotKeys(options);
 
 				expect(std.out).toMatchInlineSnapshot(`
-					"╭─────────────────────╮
+					"╭───────────────────────────────────────────────────────╮
```
petebacondarwin, line 263: This doesn't look right...

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -407,7 +406,6 @@ export class ConfigController extends Controller<ConfigControllerEventMap> {
 				persistent: true,
 				ignoreInitial: true,
 			}).on("change", async (_event) => {
-				logger.log(`${path.basename(configPath)} changed...`);
```
petebacondarwin, line 410: Perhaps change this to `logger.debug`?

packages/wrangler/src/cli-hotkeys.ts
```diff
@@ -55,9 +54,9 @@ export default function (
 			.join("\n");
 
 		return (
-			`╭──${"─".repeat(maxLineLength)}──╮\n` +
+			`╭──${"─".repeat(length)}──╮\n` +
```
petebacondarwin, line 57: Not sure if this change is right

packages/wrangler/src/dev/hotkeys.ts
```diff
@@ -30,20 +27,6 @@ export default function registerDevHotKeys(
...
-				`turn ${devEnv.config.latestConfig?.dev?.remote ? "on" : "off"} local mode`,
-			handler: async () => {
-				await devEnv.config.patch({
-					dev: {
-						...devEnv.config.latestConfig?.dev,
-						remote: !devEnv.config.latestConfig?.dev?.remote,
-					},
-				});
-			},
-		},
```
petebacondarwin, line 46: Is this not a breaking change?

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -8,31 +9,31 @@ export const friendlyBindingNames: Record<
...
 	version_metadata: "Worker Version Metadata",
 	unsafe: "Unsafe Metadata",
-	vars: "Vars",
-	wasm_modules: "Wasm Modules",
-	dispatch_namespaces: "Dispatch Namespaces",
-	mtls_certificates: "mTLS Certificates",
-	workflows: "Workflows",
-	pipelines: "Pipelines",
-	secrets_store_secrets: "Secrets Store Secrets",
+	vars: "Environment Variable",
```
petebacondarwin, line 30: Hmm not sure about this wording? Why not just `Variable`?
Environment variable suggests `process.env` to me.

petebacondarwin (changes requested):

packages/wrangler/src/utils/print-bindings.ts
penalosa: I was going off what we call them in the docs: https://developers.cloudflare.com/workers/configuration/environment-variables/

petebacondarwin: 🤢  - okey dokey!

packages/miniflare/src/shared/log.ts
penalosa: They could be, but they really shouldn't be. We also haven't really considered logs to be part of the semver contract for Wrangler/Miniflare in the past, so I think this should be fine?

packages/wrangler/src/dev/hotkeys.ts
penalosa: Ah yeah, I suppose it is. I'll re-add it

petebacondarwin: I would be up for hiding it from display but still supporting the key-press.

petebacondarwin (approved):

irvinebroque:
nice one, this is a huge improvement

PR #9341, opened 2025-05-22
Integrate assets binding with Vite's middleware

jamesopstad:

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/playground/assets/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+export default {
+	async fetch(request, env) {
+		const { origin, pathname } = new URL(request.url);
+
+		switch (pathname) {
+			case "/public-directory-asset": {
+				const response = await env.ASSETS.fetch(
+					new URL("/public-image.svg", origin)
```
jamesopstad, line 22: You can't just use any random `origin` here because the URL will pass through Vite's `hostCheckMiddleware`, which checks against the [`server.allowedHosts`](https://vite.dev/config/server-options.html#server-allowedhosts) option. Best thing is to use the `origin` from the incoming request URL. I think this is what we should do in examples.

jamesopstad: Also, the advantage of using `new URL()` is that the `base` (second parameter) will be ignored if the `url` includes an `origin`. This means that if importing an asset returns a data URI, it can be used without the user having to know it's not a regular URL.

packages/wrangler/src/api/startDevWorker/utils.ts
```diff
@@ -56,14 +56,11 @@ export function urlFromParts(
...
+	H extends AsyncHook<T, Args>,
 	T extends HookValues = UnwrapHook<H>,
 	Args extends unknown[] = [],
->(
-	hook: H,
-	...args: Args
-): H extends undefined ? UnwrapHook<H> | undefined : UnwrapHook<H> {
-	return typeof hook === "function" ? hook(...args) : hook;
+>(hook: H, ...args: Args) {
+	return (typeof hook === "function" ? hook(...args) : hook) as UnwrapHook<H>;
```
jamesopstad, line 63: Somewhere along the way the TypeScript version got updated and that meant the types in this function started failing. Have updated so they pass but needs confirming that these are correct.

packages/vite-plugin-cloudflare/playground/assets/src/imported-image.svg
petebacondarwin, line 1: Love that you have got on board with using these images 😄

packages/vite-plugin-cloudflare/playground/assets/src/index.ts
petebacondarwin, line 22: NIT: I think you can just do:

```suggestion
					new URL("/public-image.svg", request.url)
```

petebacondarwin: Although I guess if you are extracting the pathname then you might as well use the `origin` too.

packages/vite-plugin-cloudflare/playground/assets/vite.config.ts
```diff
@@ -0,0 +1,21 @@
+import * as path from "node:path";
+import { cloudflare } from "@cloudflare/vite-plugin";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+	environments: {
+		client: {
+			build: {
+				rollupOptions: {
+					input: path.resolve(__dirname, "html-page.html"),
```
petebacondarwin, line 10: Out of curiosity, why is this needed?

petebacondarwin (approved):

packages/vite-plugin-cloudflare/playground/assets/src/index.ts
jamesopstad: Nice! We should think about what pattern we want to encourage here and use it consistently in the docs.

packages/vite-plugin-cloudflare/playground/assets/vite.config.ts
jamesopstad: The default entry point for the `client` build is `index.html` so you need to set it explicitly if you want it to be something else. This is also how you create a build with multiple `client` entry points (https://vite.dev/guide/build.html#multi-page-app).

PR #9344, opened 2025-05-22
add warning about env not specified to potentially risky wrangler commands

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1785.

add warning about env not specified to potentially risky wrangler commands (for more details check the changeset)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: DX improvement
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: this change can't easily be backported, so I think that not doing so here is valid, I'm happy to reconsider if someone disagrees[^1]

[^1]: The main issue is that this change is built upon the changes made in https://github.com/cloudflare/workers-sdk/pull/8966 (where `createCommand` is being introduced for the various commands) which is not backported, so potentially a different solution would need to be implemented for this in order to do a v3 backport

### DIFF
--- a/.changeset/some-cloths-move.md
+++ b/.changeset/some-cloths-move.md
@@ -19,5 +19,5 @@ of the following commands:
 - wrangler secret delete
 - wrangler triggers deploy (TODO: manually test)
 
-this is a measure we're putting in place to try to prevent developers from accidentally apply
+this is a measure we're putting in place to try to prevent developers from accidentally applying
 changes to an incorrect (potentially even production) environment
--- a/.changeset/some-cloths-move.md
+++ b/.changeset/some-cloths-move.md
@@ -1,0 +1,23 @@
+---
+"wrangler": patch
+---
+
+add warning about env not specified to potentially risky wrangler commands
+
+add a warning suggesting users to specify their target environment (via `-e` or `--env`)
+when their wrangler config file contains some environments and they are calling one
+of the following commands:
+
+- wrangler deploy
+- wrangler versions upload
+- wrangler versions deploy
+- wrangler versions secret bulk
+- wrangler versions secret put
+- wrangler versions secret delete
+- wrangler secret bulk
+- wrangler secret put
+- wrangler secret delete
+- wrangler triggers deploy (TODO: manually test)
+
+this is a measure we're putting in place to try to prevent developers from accidentally apply
+changes to an incorrect (potentially even production) environment
--- a/packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
+++ b/packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
@@ -223,7 +223,7 @@ describe("config findWranglerConfig()", () => {
 			expect(std).toEqual(NO_LOGS);
 		});
 
-		it("should error if deploy config file and user config file do not have the same base path", async () => {
+		it.skip("should error if deploy config file and user config file do not have the same base path", async () => {
 			await seed({
 				[`foo/wrangler.toml`]: "DUMMY",
 				["foo/bar/.wrangler/deploy/config.json"]: `{ "configPath": "../../dist/wrangler.json" }`,
--- a/packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
+++ b/packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
@@ -223,7 +223,7 @@ describe("config findWranglerConfig()", () => {
 			expect(std).toEqual(NO_LOGS);
 		});
 
-		it.skip("should error if deploy config file and user config file do not have the same base path", async () => {
+		it("should error if deploy config file and user config file do not have the same base path", async () => {
 			await seed({
 				[`foo/wrangler.toml`]: "DUMMY",
 				["foo/bar/.wrangler/deploy/config.json"]: `{ "configPath": "../../dist/wrangler.json" }`,
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -13021,7 +13021,7 @@ export default{
 				  To avoid unintentional changes to the wrong environment, it is recommended to explicitly specify
 				  the target environment using the \`-e|--env\` flag.
 				  If your intention is to use the top-level environment of your configuration simply pass an empty
-				  string to the flag to target such environment. For example `--env=""`.
+				  string to the flag to target such environment. For example \`--env=\\"\\"\`.
 
 				"
 			`);
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -13000,6 +13000,63 @@ export default{
 			await runWrangler("deploy ./index.js");
 		});
 	});
+
+	describe("multi-env warning", () => {
+		it("should warn if the wrangler config contains environments but none was specified in the command", async () => {
+			writeWorkerSource();
+			writeWranglerConfig({
+				main: "./index.js",
+				env: {
+					test: {},
+				},
+			});
+			mockSubDomainRequest();
+			mockUploadWorkerRequest();
+
+			await runWrangler("deploy");
+
+			expect(std.warn).toMatchInlineSnapshot(`
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mMultiple environments are defined in the Wrangler configuration file, but no target environment was specified for the deploy command.[0m
+
+				  To avoid unintentional changes to the wrong environment, it is recommended to explicitly specify
+				  the target environment using the \`-e|--env\` flag.
+
+				"
+			`);
+		});
+
+		it("should not warn if the wrangler config contains environments and one was specified in the command", async () => {
+			writeWorkerSource();
+			writeWranglerConfig({
+				main: "./index.js",
+				env: {
+					test: {},
+				},
+			});
+			mockSubDomainRequest();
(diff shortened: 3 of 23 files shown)


.changeset/some-cloths-move.md
```diff
@@ -0,0 +1,23 @@
...
+- wrangler versions deploy
+- wrangler versions secret bulk
+- wrangler versions secret put
+- wrangler versions secret delete
+- wrangler secret bulk
+- wrangler secret put
+- wrangler secret delete
+- wrangler triggers deploy (TODO: manually test)
+
+this is a measure we're putting in place to try to prevent developers from accidentally apply
```
petebacondarwin:
```suggestion
this is a measure we're putting in place to try to prevent developers from accidentally applying
```

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
```diff
@@ -223,7 +223,7 @@ describe("config findWranglerConfig()", () => {
 			expect(std).toEqual(NO_LOGS);
 		});
 
-		it("should error if deploy config file and user config file do not have the same base path", async () => {
+		it.skip("should error if deploy config file and user config file do not have the same base path", async () => {
```
petebacondarwin: ??

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -13000,6 +13000,63 @@ export default{
...
+			mockSubDomainRequest();
+			mockUploadWorkerRequest();
+
+			await runWrangler("deploy");
+
+			expect(std.warn).toMatchInlineSnapshot(`
+				"[33m▲ [43;33m[[43;30mWARNING[43;33m][0m [1mMultiple environments are defined in the Wrangler configuration file, but no target environment was specified for the deploy command.[0m
+
+				  To avoid unintentional changes to the wrong environment, it is recommended to explicitly specify
+				  the target environment using the \`-e|--env\` flag.
```
petebacondarwin, line 13022: Should we explain how to target the top level environment so that the warning is not shown in this case?

packages/wrangler/src/config/config-helpers.ts
```diff
@@ -123,6 +123,10 @@ function findRedirectedWranglerConfig(
 					Found both a user configuration file at "${path.relative(".", userConfigPath)}"
 					and a deploy configuration file at "${path.relative(".", deployConfigPath)}".
 					But these do not share the same base path so it is not clear which should be used.
+
+
+					===a= ${path.join(path.dirname(userConfigPath), PATH_TO_DEPLOY_CONFIG)}
+					===b= ${deployConfigPath}
```
petebacondarwin: ???

packages/wrangler/src/core/register-yargs-command.ts
```diff
@@ -157,21 +159,47 @@ function createHandler(def: CommandDefinition) {
...
+								useRedirectIfAvailable:
+									def.behaviour?.useConfigRedirectIfAvailable,
+							})
+						: defaultWranglerConfig;
+
+				if (def.behaviour?.warnIfMultipleEnvsConfiguredButNoneSpecified) {
+					if (!args.env) {
+						const { configPath } = findWranglerConfig(process.cwd(), {
+							useRedirectIfAvailable: true,
+						});
```
petebacondarwin: Can you use `userConfigPath` available on the `config` object?

packages/wrangler/src/core/register-yargs-command.ts
```diff
@@ -157,21 +159,47 @@ function createHandler(def: CommandDefinition) {
...
+							{
+								config: configPath,
+							},
+							{ hideWarnings: true }
+						);
+						const availableEnvs = Object.keys(rawConfig.env ?? {});
+						if (availableEnvs.length > 0) {
+							logger.warn(
+								dedent`
+										Multiple environments are defined in the Wrangler configuration file, but no target environment was specified for the${def.metadata.displayName ? ` ${def.metadata.displayName}` : ""} command.
```
petebacondarwin: Can we not compute the display name of the current command?
The data is definitely in the system, but from looking at the code it is a bit awkward to pass it down to here.  Perhaps as a follow up?

packages/wrangler/src/__tests__/config/findWranglerConfig.test.ts
dario-piotrowicz: oops, sorry I didn't meant to commit this! thanks 🙏

packages/wrangler/src/config/config-helpers.ts
dario-piotrowicz: again, committed by mistake! thanks! 🙇

packages/wrangler/src/core/register-yargs-command.ts
dario-piotrowicz: > Can we not compute the display name of the current command?

We definitely can, I simply didn't want to force every command to have to define this value (even when it would not be actually used at all)

> but from looking at the code it is a bit awkward to pass it down to here

Mostly my issue was as mentioned above, if displayName is not necessary for a command it might be annoying to force people to have to define it always... but it might be good especially for potential future use case?

> Perhaps as a follow up?

Please advise 🙂

I am happy to update this here in this PR so that all commands do have to always specify it.

Or add a todo comment and do it as a followup. 👍 

(Or leave it as if if we indeed don't think that forcing every command to define their displayName to be necessary)

packages/wrangler/src/__tests__/deploy.test.ts
dario-piotrowicz: can the top level environment be targeted? 🤔

packages/wrangler/src/core/register-yargs-command.ts
dario-piotrowicz: mh... I don't think so 🤔 

`userConfigPath` is the non-redirected config path right? so a redirected configuration would not be taken into account, right?

What I'm thinking of is a scenario in which a user has different environments in their config file, then they use some tool (e.g. nitro, vite) to build their worker targeting a specific environment (which builds the appropriate environment-less redirected config), afterwords they run `wrangler deploy` for that command.
In such case we don't want `wrangler deploy` to be called with an environment (since that's not supported in redirected configs) so not taking into consideration the (environment-less) redirected environment and present the warning would be incorrect, wouldn't it?

packages/wrangler/src/core/register-yargs-command.ts
petebacondarwin: I think you have misunderstood me. My goal would be not to have to provide this field at all. Instead compute it from where the command is registered since it is already provided there.

packages/wrangler/src/core/register-yargs-command.ts
petebacondarwin: Good point. I didn't think of that scenario. Indeed we should not warn if the actual config being used is a redirected one with only a top level environment. In which case I alter my suggestion, which is to use the `config.configPath` property from the read config, rather than trying to find the config a second time.

dario-piotrowicz: Done 👍 

Sorry my bad I wasn't sure if the config would be resolved at this point, but I can see that it is, so yeah no reason to try to find it again 👍

packages/wrangler/src/core/register-yargs-command.ts
dario-piotrowicz: yeah I misunderstood your comment 😓 

Yes, I've updated this, sorry I didn't notice the `command` property before 🤦

petebacondarwin (approved):

packages/wrangler/src/__tests__/deploy.test.ts
dario-piotrowicz: https://github.com/cloudflare/workers-sdk/pull/9344/commits/02d23f3c61e743e7aa0cadc8a07ac73383ec5968 🙂

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -13020,6 +13020,8 @@ export default{
 
 				  To avoid unintentional changes to the wrong environment, it is recommended to explicitly specify
 				  the target environment using the \`-e|--env\` flag.
+				  If your intention is to use the top-level environment of your configuration simply pass an empty
+				  string to the flag to target such environment.
```
petebacondarwin:
```suggestion
				  If your intention is to use the top-level environment of your configuration simply pass an empty
				  string to the flag to target such environment. For example `--env=""`.
```

packages/wrangler/src/config/validation.ts
```diff
@@ -178,11 +178,13 @@ export function normalizeAndValidateConfig(
 
 	//TODO: find a better way to define the type of Args that can be passed to the normalizeAndValidateConfig()
 	const envName = args.env;
-	assert(envName === undefined || typeof envName === "string");
+	assert(
+		envName === undefined || envName === "" || typeof envName === "string"
+	);
```
petebacondarwin: No real difference ?

dario-piotrowicz: oh right! 😓 👍  🤦

dario-piotrowicz:
> 2 nits

Both addressed, thank you very much! 🙏

PR #9347, opened 2025-05-23
Improve wrapped bindings UI

penalosa:
Followup to #9335 to improve the display when the terminal is small enough to require wrapping. Additionally, correctly label Vectorive bindings as `not supported`/`remote` depending on the flags provided.

Screenshot 2025-05-23 at 09 21 11

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: should be tested manually
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: small ui change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: change to a v4-only feature

### DIFF
--- a/packages/wrangler/src/utils/print-bindings.ts
+++ b/packages/wrangler/src/utils/print-bindings.ts
@@ -51,6 +51,7 @@ export function printBindings(
 		name?: string;
 		provisioning?: boolean;
 		warnIfNoBindings?: boolean;
+		vectorizeBindToProd?: boolean;
 	} = {}
 ) {
 	let hasConnectionStatus = false;
@@ -58,9 +59,8 @@ export function printBindings(
 		isProvisioning: context.provisioning,
 		isLocalDev: context.local,
 	});
-	const truncate = (item: string | Record<string, unknown>) => {
+	const truncate = (item: string | Record<string, unknown>, maxLength = 40) => {
 		const s = typeof item === "string" ? item : JSON.stringify(item);
-		const maxLength = 40;
 		if (s.length < maxLength) {
 			return s;
 		}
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -565,20 +572,43 @@ export function printBindings(
 			" (".length +
 			maxValueLength +
 			")".length;
+
+		const shouldWrap =
+			bindingLength + 6 + maxTypeLength + 6 + maxModeLength >=
```
petebacondarwin, line 577: Where do these 6s come from?
Should they be computed from something to avoid them getting out of sync in the future?

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -565,20 +572,43 @@ export function printBindings(
...
+				`${padEndAnsi(
+					dim("Binding"),
+					bindingPrefix.length + maxNameLength
+				)}    ${padEndAnsi(dim("Resource"), maxTypeLength)}    ${hasMode ? dim("Mode") : ""}`
 			);
-
+		} else {
 			logger.log(
-				`${bindingString}      ${brandColor(binding.type.padEnd(maxTypeLength))}      ${hasMode ? binding.mode : ""}`
+				`${padEndAnsi(dim("Binding"), bindingLength)}      ${padEndAnsi(dim("Resource"), maxTypeLength)}      ${hasMode ? dim("Mode") : ""}`
```
petebacondarwin, line 590: As far as I can tell, the only difference between these two blocks is `bindingPrefix.length + maxNameLength` vs `bindingLength`. Is that right?

If so then perhaps simplify by creating a new variable for this value and only having one log statement?

petebacondarwin (approved):

emily-shen:
feel free to ignore these unsolicited design opinions, but what about adding a line under the header? and an extra line at the top of the table to mirror the bottom.

like this:

```
Your Worker has access to the following bindings:

Binding                            Resource      
-----------------------------------------------
env.DEFAULT_EXPORT (worker-b)      Worker  
env.DEFAULT_ENTRYPOINT (worker-c)  Worker  

Uploaded worker-a (14.74 sec)
```

PR #9360, opened 2025-05-23
Router worker performs free tier limiting

matthewdavidrodgers:
WC-3584

The Router worker now limits requests when free users are over their daily limit. This decision is made via the new optional param binding (EYEBALL_ROUTER_CONFIG) that is passed in. The Router worker then denies all requests to the user worker when that param indicates limiting should occur.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: handled by others
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


matthewdavidrodgers:
This is rebased over https://github.com/cloudflare/workers-sdk/pull/9416 and thus the diff includes those commits as well until it is merged

packages/workers-shared/router-worker/src/limited-response.html
GregBrimble, line 1: Product approved this messaging, yeah? Looks good to me — just checking.

Do we want to snazz it up a bit with some CSS/illustration at all?

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -4,8 +4,11 @@ import { setupSentry } from "../../utils/sentry";
 import { mockJaegerBinding } from "../../utils/tracing";
 import { Analytics, DISPATCH_TYPE, STATIC_ROUTING_DECISION } from "./analytics";
 import { applyConfigurationDefaults } from "./configuration";
-import type AssetWorker from "../../asset-worker";
+/* @ts-expect-error importing an html file bundles this via esbuild */
```
GregBrimble: We can instead add a `types.d.ts` somewhere and:

```typescript
declare module "*.html" {
  const content: string;
  export default content;
}
```

packages/workers-shared/router-worker/src/limited-response.html
matthewdavidrodgers: yeah i expected this to have plenty of product feedback here and to iterate based on that. i did no styling (or even markup really) and just brought over the copy of what's currently in FL. happy to do a basic pass first

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -77,6 +84,48 @@ export default {
...
 
+			let asset: "static_routing" | "ignored" | true | false = false;
+
+			const routeToUserWorker = async () => {
+				if (!config.has_user_worker) {
+					throw new Error(
+						"Fetch for user worker without having a user worker binding"
+					);
+				}
+				if (eyeballConfig.limitedAssetsOnly) {
```
GregBrimble, line 118: I thought we had discussed the presence of the static routing file being the opt-in for this fail-closed behavior. Maybe I got the wrong end of the stick?

This, as written, will mean that existing projects which are relying on fail open that Workers with static assets does today will suddently stop working. I think we can only do this limiting when (a) we add some explicit opt-in for it (TBD if we ever do) or (b) only for new projects/projects which opt-in to some other new feature which we couple with fail-closed behavior.

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -77,6 +84,48 @@ export default {
 
 			const maybeSecondRequest = request.clone();
 
+			let asset: "static_routing" | "ignored" | true | false = false;
```
GregBrimble: I don't _love_ mutable and non-linear variables like this (though absolutely have used them in a few places myself!)

Is it at all cleaner if this is becomes an arg to the `routeToUserWorker` and `routeToAssets` functions?

GregBrimble (commented):
I think the fail-closed coming hard on even for existing projects is a blocker to this. But the rest of the code looks totally fine to me.

packages/workers-shared/router-worker/src/worker.ts
matthewdavidrodgers: I see what you mean. We definitely have the ability to only fail-closed when static_routing is present, that shouldn't be that big of a change here. The only nuance there is what should happen to fallback worker invocations, that is, when there's no static routing directing a request (or run_worker_first), and an asset is not found for that path. the fallback is to hand the request off to the user worker, but with fail-open that would 404 (because it's continuing to the asset worker instead of the user worker) and with fail-closed that would 429 (because we limit the request).

to keep things simple i would prefer to fail-closed as the default here, but if that's considered a breaking change i think failing-open can stay the default behavior.

packages/workers-shared/router-worker/src/worker.ts
matthewdavidrodgers: cleaner is relative here IMO. the typing becomes cleaner but the call sites is more verbose. no strong opinion here

packages/workers-shared/router-worker/src/worker.ts
GregBrimble: Sent a DM!

petebacondarwin: I agree that we should default to failing closed; it is safer than accidentally exposing assets that should have been secured behind a Worker.

The only way a user could have locked down such assets is if they have `run_worker_first` set to a truthy value, right? There are two ways for this to happen:

- `run_worker_first: true`
- `run_worker_first: [ ... ]`

In the second case, this is a new feature so it would not be a breaking change.
In the first case, it would potentially be a breaking change but I think this is actually a security flaw to fail open here and so we are actually fixing a bug rather than adding a breaking change feature.

Also, non-free tier customers would never hit this. right?

packages/workers-shared/router-worker/src/worker.ts
petebacondarwin: Not a big fan of this variable being mutated and used in different places. Can we make this a parameter instead?

```diff
diff --git a/packages/workers-shared/router-worker/src/worker.ts b/packages/workers-shared/router-worker/src/worker.ts
index d88375ffe..9c188c36b 100644
--- a/packages/workers-shared/router-worker/src/worker.ts
+++ b/packages/workers-shared/router-worker/src/worker.ts
@@ -104,9 +104,7 @@ export default {
 
 			const maybeSecondRequest = request.clone();
 
-			let asset: "static_routing" | "ignored" | true | false = false;
-
-			const routeToUserWorker = async () => {
+			const routeToUserWorker = async ({ asset }: { asset: AssetTag }) => {
 				if (!config.has_user_worker) {
 					throw new Error(
 						"Fetch for user worker without having a user worker binding"
@@ -126,19 +124,19 @@ export default {
 				return env.JAEGER.enterSpan("dispatch_worker", async (span) => {
 					span.setTags({
 						hasUserWorker: true,
-						asset: asset,
+						asset,
 						dispatchType: DISPATCH_TYPE.WORKER,
 					});
 					return env.USER_WORKER.fetch(maybeSecondRequest);
 				});
 			};
 
-			const routeToAssets = async () => {
+			const routeToAssets = async ({ asset }: { asset: AssetTag }) => {
 				analytics.setData({ dispatchtype: DISPATCH_TYPE.ASSETS });
 				return await env.JAEGER.enterSpan("dispatch_assets", async (span) => {
 					span.setTags({
 						hasUserWorker: config.has_user_worker,
-						asset: asset,
+						asset,
 						dispatchType: DISPATCH_TYPE.ASSETS,
 					});
 
@@ -160,8 +158,7 @@ export default {
 					analytics.setData({
 						staticRoutingDecision: STATIC_ROUTING_DECISION.ROUTED,
 					});
-					asset = "static_routing";
-					return await routeToAssets();
+					return await routeToAssets({ asset: "static_routing" });
 				}
 				// evaluate "include" rules
 				const includeRulesMatcher = generateStaticRoutingRuleMatcher(
@@ -181,8 +178,7 @@ export default {
 					analytics.setData({
 						staticRoutingDecision: STATIC_ROUTING_DECISION.ROUTED,
 					});
-					asset = "static_routing";
-					return await routeToUserWorker();
+					return await routeToUserWorker({ asset: "static_routing" });
 				}
 
 				analytics.setData({
@@ -195,19 +191,17 @@ export default {
 			// User's configuration indicates they want user-Worker to run ahead of any
 			// assets. Do not provide any fallback logic.
 			if (config.invoke_user_worker_ahead_of_assets) {
-				asset = "ignored";
-				return await routeToUserWorker();
+				return await routeToUserWorker({ asset: "ignored" });
 			}
 
 			// If we have a user-Worker, but no assets, dispatch to Worker script
 			const assetsExist = await env.ASSET_WORKER.unstable_canFetch(request);
-			asset = assetsExist;
 			if (config.has_user_worker && !assetsExist) {
-				return await routeToUserWorker();
+				return await routeToUserWorker({ asset: assetsExist });
 			}
 
 			// Otherwise, we either don't have a user worker, OR we have matching assets and should fetch from the assets binding
-			return await routeToAssets();
+			return await routeToAssets({ asset: assetsExist });
 		} catch (err) {
 			if (userWorkerInvocation) {
 				// Don't send user Worker errors to sentry; we have no way to distinguish between
@@ -228,3 +222,5 @@ export default {
 		}
 	},
 };
+
+type AssetTag = "static_routing" | "ignored" | true | false;
```

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -55,7 +79,10 @@ export default {
 			);
 
 			const hasStaticRouting = env.CONFIG.static_routing !== undefined;
-			const config = applyConfigurationDefaults(env.CONFIG);
+			const [config, eyeballConfig] = applyConfigurationDefaults(
+				env.CONFIG,
+				env.EYEBALL_CONFIG
+			);
```
petebacondarwin: Personally I'd rather just have two functions here since they are completely independent of each other:

```suggestion
			const config = applyConfigurationDefaults(env.CONFIG);
			const eyeballConfig = applyEyeballConfigurationDefaults(env.EYEBALL_CONFIG);
```

packages/workers-shared/utils/types.ts
```diff
@@ -20,6 +20,13 @@ export const RouterConfigSchema = z.object({
 	...InternalConfigSchema.shape,
 });
 
+export const EyeballRouterConfigSchema = z.union([
+	z.object({
+		limitedAssetsOnly: z.boolean().optional(),
+	}),
+	z.null(),
+]);
```
petebacondarwin, line 28: Why support both `undefined` and `null`. I don't believe there is any semantic difference here? Can we just drop the `null` and simplify this and the `RequiredEyeballRouterConfig` type?

matthewdavidrodgers:
@petebacondarwin yes, this will be observed when there is a `run_worker_first` value that dictates running the worker. however, this change also will affect requests that hit the user worker without a `run_worker_first` value, particularly when limited and falling back to the user worker when an asset is not found. e.g.
GET /missing-asset.html -> asset worker reports not found, fall back to the user worker, fail-open, 404
vs
GET /missing-asset.html -> asset worker reports not found, fall back to the user worker, fail-closed, 429

This is still the right choice in my opinion, as implicitly disabling your compute when you trip a limit is unexpected and certainly a footgun (even if that compute is usually just returning a 404 for the fallback). And of course, setting not_found_handling to single-page-application or 404-page means we won't even hit those fallbacks, no limited behavior will be encountered. and yes, this all only applies to free tier customers.

packages/workers-shared/utils/types.ts
matthewdavidrodgers: the optional param binding which provides this param (as the EYEBALL_CONFIG binding) provides the json when present, otherwise the value is presented in the environment as null, so that union reflects that implementation

the field limitedAssetsOnly within the object is optional to be defensive, but currently the code providing the object always sets the field when providing the object. this can be changed, but i prefer being defensive, as this config may be extended and not all fields may be set in the future

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -3,19 +3,46 @@ import { PerformanceTimer } from "../../utils/performance";
...
 import type AssetWorker from "../../asset-worker";
 import type {
+	EyeballRouterConfig,
 	JaegerTracing,
 	RouterConfig,
 	UnsafePerformanceTimer,
 } from "../../utils/types";
 import type { ColoMetadata, Environment, ReadyAnalytics } from "./types";
 
+const limitedResponse = `<html>
```
GregBrimble, line 19: styling and wording approved now? It's a bit of a departure from our usual styling.

GregBrimble (approved):
style! 💅

petebacondarwin (approved):
~~Since this is already in production and working. Let's land this and then we can discuss any tidy-up refactorings that I would like to see afterwards.~~ Wrong PR sorry.

matthewdavidrodgers:
this PR has been superseded by https://github.com/cloudflare/cloudflare-docs/pull/23008

PR #9368, opened 2025-05-26
hook client req abort to worker request in vite plugin

marbemac:
Related issue: https://github.com/cloudflare/workerd/issues/4185#issuecomment-2907972284

workerd now supports a `enable_request_signal` compat flag, which allows one to hook into `req.signal` in your worker entrypoint code. This is useful to do things like listen to client disconnects in order to cleanup SSE connection maps, etc.

This PR hooks the client request up to the request sent to the CF worker.

I am not sure if:

1. is any kind of documentation update needed?
2. are tests needed? and if so, what kind? at the moment I have simply manually tested and verified these changes are working

Note that the workerd that comes with wrangler is apparently behind what is in CF production, so in order to test the `enable_request_signal` flag you have to download the latest workerd binary and link it with an env variable - explained by @npaun here -> https://github.com/cloudflare/workerd/issues/3033#issuecomment-2902496431.

---

- Tests
  - [x] TODO (before merge)
  - [ ] Tests included
  - [x] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -370,11 +370,19 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 							assert(miniflare, `Miniflare not defined`);
 							const routerWorker = await getRouterWorker(miniflare);
 
+							const abortController = new AbortController();
+							req.on("aborted", () => {
+								abortController.abort();
+							});
+
 							const request = createRequest(req, res);
 							const response = await routerWorker.fetch(
 								toMiniflareRequest(request),
 								{
 									redirect: "manual",
+									// connect the client http request abort hook to the worker fetch request
+									// so that the worker implementation can hook into client abort behaviors
+									signal: abortController.signal,
 								}
 							);
 
@@ -386,6 +394,11 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 
 							await sendResponse(res, response as any);
 						} catch (error) {
+							// Do not bubble client abort errors up to the Vite server
+							if (error instanceof Error && error.name === "AbortError") {
+								return;
+							}
+
 							next(error);
 						}
 					});
@@ -416,10 +429,15 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 				// In preview mode we put our middleware at the front of the chain so that all assets are handled in Miniflare
 				vitePreviewServer.middlewares.use(async (req, res, next) => {
 					try {
+						const abortController = new AbortController();
+						req.on("aborted", () => {
+							abortController.abort();
+						});
+
 						const request = createRequest(req, res);
 						const response = await miniflare.dispatchFetch(
 							toMiniflareRequest(request),
-							{ redirect: "manual" }
+							{ redirect: "manual", signal: abortController.signal }
 						);
 
 						// Vite uses HTTP/2 when `preview.https` is enabled
@@ -430,6 +448,11 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 
 						await sendResponse(res, response as any);
 					} catch (error) {
+						// Do not bubble client abort errors up to the Vite server
+						if (error instanceof Error && error.name === "AbortError") {
+							return;
+						}
+
 						next(error);
 					}
 				});


npaun:
Thanks for this, @marbemac! Tagging @dario-piotrowicz to review it.

jamesopstad (commented):
Thanks for this @marbemac! It's looking good. I believe the version of `workerd` used by Miniflare in the Vite plugin now supports `enable_request_signal`.

Please could you add tests for this and a changeset. I think you could adapt the `streaming` playground for testing this feature.

marbemac:
Are you ok with adding some test cases to the existing streaming playground, or are you suggesting I should copy streaming playground to a new one?

jamesopstad:
> Are you ok with adding some test cases to the existing streaming playground, or are you suggesting I should copy streaming playground to a new one?

Adding to the existing playground is good. You can add a separate route for the Worker to handle if necessary. Thanks.

packages/vite-plugin-cloudflare/playground/request-signal/wrangler.jsonc
```diff
@@ -0,0 +1,20 @@
+{
+	"name": "worker",
+	"main": "./src/index.ts",
+	"compatibility_date": "2024-09-09",
```
petebacondarwin, line 4: This is quite an old compat date. Perhaps make it more recent?

packages/vite-plugin-cloudflare/playground/request-signal/__tests__/signal.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { expect, test, vi } from "vitest";
+import { page } from "../../__test-utils__";
+
+test("renders HTML stream", async () => {
+	const heading = await page.textContent("h1");
+	expect(heading).toBe("Streaming example");
```
petebacondarwin, line 6: Should this be:

```suggestion
	expect(heading).toBe("Request signal playground");
```

packages/vite-plugin-cloudflare/playground/request-signal/__tests__/signal.spec.ts
```diff
@@ -0,0 +1,12 @@
...
+	const finalChunk = await vi.waitUntil(() => page.textContent("#two"), {
+		timeout: 5000,
+		interval: 500,
+	});
+	expect(finalChunk).toBe("Chunk after 2 seconds");
```
petebacondarwin, line 11: Don't we now need to check that the request was cancelled after we received the second chunk?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -369,19 +369,27 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 							const request = createRequest(req, res);
 							let response: MiniflareResponse;
 
+							// connect the client http request abort hook to the worker fetch request
+							// so that the worker implementation can hook into client abort behaviors
+							const abortController = new AbortController();
+							req.on("close", () => {
+								console.log("Request closed", req.url);
```
petebacondarwin, line 376: Can we use the vite logger rather than console here?

petebacondarwin: Or was this just for debugging and should be removed?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -393,6 +401,11 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 
 							await sendResponse(res, response as any);
 						} catch (error) {
+							// Do not bubble client abort errors up to the Vite server
+							if (error instanceof Error && error.name === "AbortError") {
+								return;
+							}
+
```
petebacondarwin, line 408: NIT: since we do the same in another catch block how about we move this to a helper function:

```ts
} catch (error) {
  if (!isAbortError(error)) {
    next(error);
  }
}
```

petebacondarwin (changes requested):
Thanks for adding the playground. I'm afraid the PR needs a rebase and conflicts resolving.

petebacondarwin:
> Hmm ok I've hit a bit of a wall. I'm not able to get the `request.signal.addEventListener("abort"` handler to trigger within the worker, in the playground 🤔.
> 
> I had a simpler test case that used the existing streaming example, but wasn't sure if my approach there was causing this issue, so I I went ahead an pushed a more complete request-signal playground to experiment.

I'll pull it down and have a play later today or tomorrow.

vicb:
@petebacondarwin or @jamesopstad could you please take a look as of what's needed to push this forward?

vicb:
It looks like this might be implemented in https://github.com/cloudflare/workers-sdk/pull/10707

jamesopstad:
Closing as implemented in https://github.com/cloudflare/workers-sdk/pull/10707.

PR #9395, opened 2025-05-28
Add WRANGLER_CI_NETWORK_MODE_HOST for Workers CI

Maximo-Guk:
Fixes #[CC-5328].

In Workers Builds environment we'd like to pass in `--network=host` argument to the docker build command used by `wrangler containers push`. 

This introduces a new `WRANGLER_CI_NETWORK_MODE_HOST` env var which will be set to `true` in Workers Builds, which will in turn add this `--network=host` argument

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


petebacondarwin (commented):
The PR itself looks fine but let's get into some serious bikeshedding...

I see in a few places the word override appears, e.g. in a comment there is `WRANGLER_CI_OVERRIDE_NETWORK_MODE_HOST` rather than `WRANGLER_CI_NETWORK_MODE_HOST`. Also the function to get the env var is called `getCIOverrideNetworkMode()`.

We can sort out that consistency but the thing that is confusing me is why we have CI in there at all. OK it is being consumed by the Workers Builds team in CI but it feels like this should be more like: `WRANGLER_CONTAINERS_NETWORK_MODE_HOST`?

Maximo-Guk:
Oops thanks for catching those inconsistencies @petebacondarwin !

Regarding the name, I think I prefer `WRANGLER_CI_OVERRIDE_NETWORK_MODE_HOST` mostly because this is an override that will only be used in Workers Builds environment, and should not be relied on by other users. If users need the option to use host networking, the CC team should introduce an option to do that at the `wrangler.jsonc` / config level. We're only using the environment variable here, so that we can override this on behalf of users when they're inside of Workers Builds, and so we don't need to mutate their config file

Issue #9065, opened 2025-04-25
Options not working on wrangler login


frankichiro:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What versions & operating system are you using?

wrangler 4.13.2, node v23.11.0, mac 14.6.1

### Describe the Bug

When trying to log in with host and port options, the verification page opens with an incorrect link.

Command:
`npx wrangler login --callback-host 127.0.0.1 --callback-port 3000`

Result:
`Temporary login server listening on 127.0.0.1:3000` <- correct
`Opening a link in your default browser: (opens @ http://localhost:8976/)` <- incorrect

`Opening a link in your default browser: (opens @ http://127.0.0.1:3000/)` <- expected

The browser opens an incorrect link. If I edit the link with the correct host and port, the link works and the login resolves.

Conclusion:
`--callback-host` and `--callback-port` are correctly applied for the listening server but not for the browser link.

petebacondarwin:
I can confirm this bug. Thanks for reporting.

PR #9396, opened 2025-05-29
Fix/wrangler oauth2 url callback host

gnekich:
Fixes regression in both #5937 #9065 .

fix the login oauth2 url when callback host param is provided with 0.0.0.0 or 127.0.0.1, defaults value for callback host param updated from localhost to 0.0.0.0 thus in addition in fixing regressions improving DX.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [x] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [x] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/curly-peas-hunt.md
+++ b/.changeset/curly-peas-hunt.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix the login oauth2 url when callback host param is provided with 0.0.0.0 or 127.0.0.1, defaults value for callback host param updated from localhost to 0.0.0.0
--- a/packages/wrangler/src/user/user.ts
+++ b/packages/wrangler/src/user/user.ts
@@ -392,6 +392,15 @@ export function validateScopeKeys(
 }
 
 function getCallbackUrl(host = "localhost", port = 8976) {
+	if (
+		host === "localhost" ||
+		host === "0.0.0.0" ||
+		host === "::" ||
+		host === "::1" ||
+		host === "127.0.0.1"
+	) {
+		return `http://localhost:${port}/oauth/callback`;
+	}
 	return `http://${host}:${port}/oauth/callback`;
 }
 
@@ -1071,7 +1080,7 @@ export async function login(
 	complianceConfig: ComplianceConfig,
 	props: LoginProps = {
 		browser: true,
-		callbackHost: "localhost",
+		callbackHost: "localhost", // "0.0.0.0",
 		callbackPort: 8976,
 	}
 ): Promise<boolean> {
--- a/packages/wrangler/src/user/user.ts
+++ b/packages/wrangler/src/user/user.ts
@@ -1059,11 +1059,9 @@ export async function getOauthToken(options: {
 			}
 		});
 
-		if (options.callbackHost !== "localhost" || options.callbackPort !== 8976) {
-			logger.log(
-				`Temporary login server listening on ${options.callbackHost}:${options.callbackPort}`
-			);
-		}
+		logger.log(
+			`Temporary login server listening on ${options.callbackHost}:${options.callbackPort}`
+		);
 		server.listen(options.callbackPort, options.callbackHost);
 	});
 	if (options.browser) {
@@ -1080,7 +1078,7 @@ export async function login(
 	complianceConfig: ComplianceConfig,
 	props: LoginProps = {
 		browser: true,
-		callbackHost: "localhost", // "0.0.0.0",
+		callbackHost: "0.0.0.0",
 		callbackPort: 8976,
 	}
 ): Promise<boolean> {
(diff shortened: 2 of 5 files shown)


penalosa:
@gnekich could you clarify what issue this is fixing? Both the linked issues have been closed. If possible, could you open a new issue with a reproduction of what's going wrong?

michaelalhilly:
Just to be clear. The issue we are currently having is with this setting:

--callback-host=0.0.0.0

Currently when working inside Docker we need to listen for localhost on 0.0.0.0. However, the login and callback urls still need to specify localhost as the host.

Currently, this is not working as setting callback-host=0.0.0.0 also changes the host in login and callback urls to 0.0.0.0 which results in an error and prevents the authorization page from loading.

The current workaround is to manually replace 0.0.0.0 in the login url with localhost and everything works.

penalosa:
@gnekich could you create a new issue with a reproduction of the problem you're seeing?

packages/wrangler/src/user/user.ts
```diff
@@ -392,6 +392,15 @@ export function validateScopeKeys(
...
+	if (
+		host === "localhost" ||
+		host === "0.0.0.0" ||
+		host === "::" ||
+		host === "::1" ||
+		host === "127.0.0.1"
+	) {
+		return `http://localhost:${port}/oauth/callback`;
+	}
 	return `http://${host}:${port}/oauth/callback`;
```
petebacondarwin: I have a feeling that we should always return `localhost` for this URL.

The only URIs that are allowed are configured in the OAuth provider, and I can't find where we (Cloudflare) control that configuration.

If we were to do a check here it should be based on an allow-list rather than a block-list as is defined here.

packages/wrangler/src/user/user.ts
```diff
@@ -1071,7 +1078,7 @@ export async function login(
 	complianceConfig: ComplianceConfig,
 	props: LoginProps = {
 		browser: true,
-		callbackHost: "localhost",
+		callbackHost: "0.0.0.0",
```
petebacondarwin: I am not convinced that this is an appropriate default.
This IP address has special meaning and can result in the server listening to requests on all hosts to the machine.

I think it should be opt-in rather than the default.

.changeset/curly-peas-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix the login oauth2 url when callback host param is provided with 0.0.0.0 or 127.0.0.1, defaults value for callback host param updated from localhost to 0.0.0.0
```
petebacondarwin: From a Wrangler user's point of view, I think the changelog should be more along the lines of:


```suggestion
Always pass a valid redirect_uri callback parameter to Cloudflare OAuth API, even when the `--callback-host` param would not be accepted - e.g. 0.0.0.0 or 127.0.0.1.
```

petebacondarwin (changes requested):
Thanks for putting this PR together.
It took me a while to get my head around what the actual problem was.
In general I think this is a good idea, but I have left some comments for changes before we can really land this.

petebacondarwin:
Just rebased and pushed a handful of changes that fit with my suggestions above.
We should also update the docs to explain this scenario to future users.

.changeset/curly-peas-hunt.md
```diff
@@ -0,0 +1,28 @@
...
+wrangler login --calback-host=127.0.0.1 --callback-port=8989
+```
+
+results in Wrangler listening on 127.0.0.1:8989 and a login URL that looks like:
+
+```
+https://dash.cloudflare.com/oauth2/auth?...&redirect_uri=http%3A%2F%2Flocalhost%3A8976%2Foauth%2Fcallback&...
+```
+
+Note that the `redirect_uri` is always `localhost:8976` whatever the callback host and port are.
```
emily-shen:
```suggestion
Fix `wrangler login` with custom `callback-host`/`callback-port`

The Cloudflare OAuth API always requires the `redirect_uri` to be `localhost:8976`. However, sometimes the Wrangler OAuth server needed to listen on a different host/port, for example when running from inside a container. We were previously incorrectly setting the `redirect_uri` to the configured callback host/port, but it needs to be up to the user to map `localhost:8976` to the Wrangler OAuth server in the container.

**Example:**

You might run Wrangler inside a docker container like this: `docker run -p 8989:8976 <image>`, which forwards port 8976 on your host to 8989 inside the container.

Then inside the container, run `wrangler login --callback-host=0.0.0.0 --callback-port=8989`

The OAuth link still has a `redirect_uri` set to`localhost:8976`. For example `https://dash.cloudflare.com/oauth2/auth?...&redirect_uri=http%3A%2F%2Flocalhost%3A8976%2Foauth%2Fcallback&...`

However the redirect to` localhost:8976` is then forwarded to the Wrangler OAuth server inside your container, allowing the login to complete.

```

PR #9401, opened 2025-05-29
Teach Wrangler how to upload Preview Aliases

WillTaylorDev:
Preview aliases give users the ability to define a stable name for their preview during version upload. These aliases are ephemeral, and operate on a sliding window (eg. if you specify many aliases, the least used ones will be deactivated).

Preview aliases may be configured only during version upload. An alias may be provided explicitly on the command line, or by use of an envvar (which is how we will power WCI stable branch aliases in the near future). Workers backend is responsible for alias name validation ( as "valid" is slightly obtuse-- not just in characters allowed, but specifically with length where a valid dns label can only be 63 characters and so also is dependent on worker name size).

These changes are currently behind a gate.

- Tests
  - [ ] TODO (before merge)
  - [X] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: behind a gate, not yet public
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [X] Not necessary because: not a patch change


packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -184,6 +184,16 @@ export const getCIOverrideName = getEnvironmentVariableFactory({
 	variableName: "WRANGLER_CI_OVERRIDE_NAME",
 });
 
+/**
+ * `WRANGLER_CI_OVERRIDE_VERSION_ALIAS` allows specifying an alias for a version during upload
+ *
+ * This is intended to be a way for users to set the version alias dynamically in CI environments.
+ * The CLI arg should always take precedence over the envvar.
+ */
+export const getCIOverridePreviewAlias = getEnvironmentVariableFactory({
```
GregBrimble, line 193: I would _suggest_ rather than using this an a string input where CI provides the value, this is instead a boolean input which tells Wrangler to pick up an alias name from the current git workspace branch name. There's Pages code in Wrangler already somewhere which does that if you want to copy it.

That'll make it more generically useful in other CI systems, I imagine (incl. our own internal ones!).

WillTaylorDev: Hmm, you know I think that's very fair. We keep the CLI arg for user specific configuration, but the env override to autogenerate. Let me re-draft this and get that going.

WillTaylorDev: Updated! Best effort for naming autogeneration. Probably could use some thinking on how we autotune this, but I don't think we should let it block. Once we set this envvar in WCI, users will have a non-blocking best effort name generation, and if the branch alias is too long, we'll simply warn and omit it from the output file (and thus the preview comment in scm).
https://github.com/cloudflare/workers-sdk/pull/9401/files#diff-8fedc91837d24f320f7ae813f02bc7e3d738cf1663102db16893776c53ab4460R929

.changeset/eight-donkeys-play.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
+---
+
+Provide ability for Wrangler to upload version aliases during upload.
```
petebacondarwin: Should this be:

```suggestion
Provide ability for Wrangler to upload preview aliases during version upload.
```

fixtures/workers-with-assets-run-worker-first/wrangler.jsonc
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "worker-with-assets-run-worker-first",
+	"name": "alias-env-demo",
```
petebacondarwin: This change seems like a mistake from testing?

packages/wrangler/src/__tests__/versions/versions.upload.test.ts
```diff
@@ -138,6 +138,41 @@ describe("versions upload", () => {
 		`);
 	});
 
+	test("should allow specifying --preview-alias", async () => {
+		mockGetScript();
+		mockUploadVersion(true);
```
petebacondarwin: Can you add functionality to this mock to ensure that the preview alias is included in the request?

packages/wrangler/src/environment-variables/factory.ts
```diff
@@ -28,6 +28,8 @@ type VariableNames =
 	| "WRANGLER_CI_MATCH_TAG"
 	| "WRANGLER_CI_OVERRIDE_NAME"
 	| "WRANGLER_CI_OVERRIDE_NETWORK_MODE_HOST"
+	| "WRANGLER_CI_OVERRIDE_PREVIEW_ALIAS"
```
petebacondarwin: Redundant?

packages/wrangler/src/versions/upload.ts
```diff
@@ -241,6 +246,11 @@ export const versionsUploadCommand = createCommand({
 			type: "string",
 			requiresArg: true,
 		},
+		"preview-alias": {
+			describe: "Name of an alias for this worker version",
```
petebacondarwin: To be more clear this will result in this upload being assigned to / associated with this preview alias?

packages/wrangler/src/versions/upload.ts
```diff
@@ -323,6 +333,17 @@ export const versionsUploadCommand = createCommand({
 			);
 		}
 
+		let previewAlias: string | undefined = args.previewAlias;
+		const shouldGeneratePreviewAlias = getCIGeneratePreviewAlias();
+		if (shouldGeneratePreviewAlias === "true") {
```
petebacondarwin: Should it generate a preview even if a preview has been provided via the args?
I would expect that the command line will trump the auto-generation...

packages/wrangler/src/versions/upload.ts
```diff
@@ -323,6 +333,17 @@ export const versionsUploadCommand = createCommand({
...
+			previewAlias = generatePreviewAlias(name);
+			if (previewAlias === undefined) {
+				logger.warn(
+					`Preview alias generation requested, but could not be autogenerated.`
+				);
+			}
+		}
```
petebacondarwin: How about moving the warning inside `generatePreviewAlias()` and then you can simplify this to:

```suggestion
		const previewAlias = getCIGeneratePreviewAlias() ? generatePreviewAlias(name) : args.previewAlias;
```

or alternatively if the command line arg gets precedence:

```suggestion
		const previewAlias = args.previewAlias ?? getCIGeneratePreviewAlias() ? generatePreviewAlias(name) : undefined;
```

packages/wrangler/src/versions/upload.ts
```diff
@@ -888,3 +920,43 @@ async function noBundleWorker(
...
+	let isGitDir = true;
+	try {
+		execSync(`git rev-parse --is-inside-work-tree`, { stdio: "ignore" });
+	} catch (err) {
+		isGitDir = false;
+	}
+
+	if (!isGitDir) {
+		return undefined;
+	}
```
petebacondarwin: Does this code appear elsewhere in Wrangler? Could we share it?

petebacondarwin: Could be simplified to:

```suggestion
	try {
		execSync(`git rev-parse --is-inside-work-tree`, { stdio: "ignore" });
	} catch (err) {
	  // not in a git directory so exit
		return undefined;
	}
```

fixtures/workers-with-assets-run-worker-first/wrangler.jsonc
WillTaylorDev: Ack bleh, yes git add all got me. Should have checked, thank you!

packages/wrangler/src/environment-variables/factory.ts
WillTaylorDev: Yup, missed deletion in the refactor to auto-generate.

packages/wrangler/src/versions/upload.ts
WillTaylorDev: Good shout, yes

packages/wrangler/src/versions/upload.ts
WillTaylorDev: Yes-- the preview alias is a stable name for a preview url that is effectively a pointer to the underlying version preview. It's ephemeral with a rolling window (the last N aliases being kept) so that unlimited aliases cannot be created. Aliases can be re-pointed to another preview version at any time.

We don't have docs for this yet since it's behind a gate and not publicly available, which would probably have helped with communicating this functionality 😅

packages/wrangler/src/output.ts
```diff
@@ -142,6 +142,8 @@ interface OutputEntryVersionUpload extends OutputEntryBase<"version-upload"> {
 	version_id: string | null;
 	/** The preview URL associated with this version upload */
 	preview_url: string | undefined;
+	/** The ephemiral aliased preview URL associated with this version upload */
```
CarmenPopoviciu:
```suggestion
	/** The ephemeral aliased preview URL associated with this version upload */
```

packages/wrangler/src/versions/upload.ts
CarmenPopoviciu:
```suggestion
			describe: "Name of an alias for this Worker version",
```

packages/wrangler/src/__tests__/versions/versions.upload.test.ts
```diff
@@ -138,6 +138,41 @@ describe("versions upload", () => {
 		`);
 	});
 
+	test("should allow specifying --preview-alias", async () => {
```
CarmenPopoviciu, line 157: should we test with an invalid preview alias name?

WillTaylorDev: Not in my opinion (or at least not for these tests)-- EWC will provide the validation of the long alias names, returning back a 400 with error details if invalid.

WillTaylorDev: (I did add some additional tests for the generation of the branch alias though!)

packages/wrangler/src/versions/upload.ts
```diff
@@ -323,6 +333,13 @@ export const versionsUploadCommand = createCommand({
 			);
 		}
 
+		const previewAlias =
+			args.previewAlias !== undefined
+				? args.previewAlias
+				: getCIGeneratePreviewAlias()
+					? generatePreviewAlias(name)
+					: undefined;
```
petebacondarwin: NIT this is the same as:

```suggestion
			args.previewAlias ??
				getCIGeneratePreviewAlias()
					? generatePreviewAlias(name)
					: undefined;
```

petebacondarwin (approved):
Nice!

packages/wrangler/src/versions/upload.ts
WillTaylorDev: Ack, took me a hot second to figure out, I think this was running into operator precedence weirdness. Wrapping the parens makes this work 👍

PR #9407, opened 2025-05-30
chore: refactor the `workers-shared` package and Wrangler builds

petebacondarwin:
Fixes DEVX-1922

- The `workers-shared` package has been refactored to allow simpler imports from other packages in the monorepo. It is no longer deployed to npm, as it is not intended for public use.
- The Wrangler builds have been updated to replace the custom esbuild and `api-extractor` with `tsup`, which simplifies the build setup and avoids issues with `api-extractor`.
- The package-lock file in import-npm fixture has been updated to use latest versions.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  such refactors are not backported


petebacondarwin:
If this lands well I might try to remove api-extractor altogether by refactoring Miniflare's builds to use tsup too.

packages/miniflare/src/workers/assets/router.worker.ts
```diff
@@ -1,5 +1,2 @@
-import worker from "@cloudflare/workers-shared/router-worker/src/index";
-
-// Simply re-export the whole of the router-worker so that it gets compiled into the Miniflare code base.
-// This allows us to have it as a devDependency only.
-export default worker;
+// Simply re-export the router-worker so that its contents gets compiled into the Miniflare code base as a Worker template
+export { default } from "@cloudflare/workers-shared/router-worker";
```
penalosa, line 2: TIL!

penalosa (approved):
This looks generally good to me

packages/vite-plugin-cloudflare/src/asset-workers/asset-worker.ts
```diff
@@ -1,46 +1,44 @@
...
 import { UNKNOWN_HOST } from "../shared";
+import type { Env as _Env } from "@cloudflare/workers-shared/asset-worker";
 
-interface Env {
+interface Env extends _Env {
 	__VITE_ASSET_EXISTS__: Fetcher;
 	__VITE_FETCH_ASSET__: Fetcher;
 }
 
 export default class CustomAssetWorker extends AssetWorker {
```
jamesopstad, line 10: We should make the `AssetWorker` generic so that we can pass in the `Env` type. I can do that in a follow up PR if you like so we keep this one focused.

packages/wrangler/package.json
```diff
@@ -51,13 +51,11 @@
 	],
 	"scripts": {
 		"assert-git-version": "node -r esbuild-register scripts/assert-git-version.ts",
-		"build": "pnpm run clean && pnpm run bundle && pnpm run emit-types && pnpm run generate-json-schema",
-		"bundle": "node -r esbuild-register scripts/bundle.ts",
+		"build": "pnpm run clean && pnpm tsup && pnpm run generate-json-schema",
```
jamesopstad, line 54: This seems quite a significant change. Are there any implications of not using the custom build script?

jamesopstad (approved):
Looks great to me!

PR #9411, opened 2025-05-30
test: re-use build cache in Wrangler e2e test runs

petebacondarwin:
Since we were running each test file individually by passing a command line argument to the turbo task, we were not benefitting from the cached builds of the dependencies.

This change moves the selection of a test file into an environment variable that is ignored by build tasks, allowing reuse of their cached output.

Since rebuilding adds 30-60 secs to each task, and there being 20 or so tests, this change should reduce the CI e2e test times by 10-20 mins!


jamesopstad (approved):

packages/wrangler/package.json
```diff
@@ -63,7 +63,7 @@
 		"test": "dotenv -- pnpm run assert-git-version && dotenv -- vitest",
 		"test:ci": "pnpm run test run",
 		"test:debug": "pnpm run test --silent=false --verbose=true",
-		"test:e2e": "dotenv -- vitest run -c ./e2e/vitest.config.mts",
+		"test:e2e": "dotenv -- vitest run -c ./e2e/vitest.config.mts $WRANGLER_E2E_TEST_FILE",
```
penalosa, line 66: I didn't know you could do this—cool!

packages/wrangler/turbo.json
```diff
@@ -65,7 +65,8 @@
 				"WRANGLER_IMPORT",
 				"MINIFLARE_IMPORT",
 				"CLOUDFLARE_ACCOUNT_ID",
-				"CLOUDFLARE_API_TOKEN"
+				"CLOUDFLARE_API_TOKEN",
+				"WRANGLER_E2E_TEST_FILE"
```
penalosa, line 69: Does this not remove the benefit, and make turbo consider that env var as part of the test context?

penalosa (approved):
Approving, but I've got one question about the addition to the turbo.json file

packages/wrangler/turbo.json
petebacondarwin: This is what we want. The tests are affected by this env var - e.g. if you change which test file to run it shouldn't just hit the cache of the last test file. Without this we would never even pass this env var to the task.

Where we want the cache hits is in the `build` tasks, which don't have this env var listed as an input, so ignore it when working out whether it is a cache hit or not.

PR #9304, opened 2025-05-19
Add _routes.json to wrangler dev and deploy

emily-shen:
This also includes #9280 

The routes.json validation logic has been copied from EWC.
This PR also plumbs routes.json through for wrangler dev.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] Tests not necessary because:
- Wrangler / Vite E2E Tests CI Job required? (Use "e2e" label or ask maintainer to run separately)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: handled separately.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: new feature

PR #9416, opened 2025-05-30
Add support for static routing to Workers Assets

matthewdavidrodgers:
For WC-3583

Implements the proposal noted here https://github.com/cloudflare/workers-sdk/discussions/9143

In brief: when static routing is present for a Worker with assets, routing via those static rules takes precedence. When a request is evaluated in the Router Worker, the request path is first compared to the "asset" rules (which are to be specific via "negative" rules, e.g. "!/api/assets"). If any match, the request is forwarded directly to the Asset Worker. If instead any "user\_worker" rules match, the request is forwarded directly to the User Worker. If neither match (or static routing was not provided), the existing behavior takes over.

As part of this explicit routing, when static routing is present, the check against "Sec-Fetch-Mode: navigate" (to guess if this should serve an asset or go to the User Worker for not\_found\_handling) is disabled. Routing can be controlled by setting routing rules via "run\_worker\_first", and asset serving (including when an index.html or 404.html page is served) will be more simple.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: others are providing this
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/workers-shared/utils/types.ts
```diff
@@ -6,8 +6,14 @@ const InternalConfigSchema = z.object({
 	debug: z.boolean().optional(),
 });
 
+const StaticRoutingSchema = z.object({
+	user_worker: z.array(z.string()),
+	asset: z.array(z.string()).optional(),
```
jamesopstad: Where will the logic live for splitting the rules into two separate arrays?

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -75,6 +77,69 @@ export default {
 
 			const maybeSecondRequest = request.clone();
 
+			if (config.static_routing) {
```
jamesopstad, line 80: Do we also need to check `config.has_user_worker` here or is `config.static_routing` only included if there is a user Worker?

packages/workers-shared/utils/types.ts
emily-shen: i'm guessing we'll need to add something to vaguely akin to `parseHeaders.ts` and `parseRedirects.ts` in workers-shared and use that in wrangler/vite.
@matthewdavidrodgers skimming the EWC PR, it looks like currently EWC does the splitting, would it make sense to consolidate this and get wrangler deploy to split it up, or would you prefer the api interface to be as close to wrangler config as possible?

packages/workers-shared/router-worker/src/worker.ts
matthewdavidrodgers: static_routing can only be provided when there is a user Worker, yes. if for some reason there are static_routing rules provided without a user Worker, and a rule resolves to the user Worker, we'll throw before attempting to dispatch there

packages/workers-shared/utils/types.ts
```diff
@@ -6,8 +6,14 @@ const InternalConfigSchema = z.object({
 	debug: z.boolean().optional(),
 });
 
+const StaticRoutingSchema = z.object({
+	worker: z.array(z.string()),
+	asset: z.array(z.string()).optional(),
```
GregBrimble:
```suggestion
	user_worker: z.array(z.string()),
	asset_worker: z.array(z.string()).optional(),
```

Just a pet peeve that the whole thing is a "Worker" (not just the user-provided script!)

.changeset/tasty-hoops-own.md
```diff
@@ -0,0 +1,11 @@
+---
+"@cloudflare/workers-shared": minor
+---
+
+Adds support for static routing to Workers Assets
+
+Implements the proposal noted here https://github.com/cloudflare/workers-sdk/discussions/9143
+
+In brief: when static routing is present for a Worker with assets, routing via those static rules takes precedence. When a request is evaluated in the Router Worker, the request path is first compared to the "asset" rules (which are to be specific via "negative" rules, e.g. "!/api/assets"). If any match, the request is forwarded directly to the Asset Worker. If instead any "user_worker" rules match, the request is forwarded directly to the User Worker. If neither match (or static routing was not provided), the existing behavior takes over.
```
GregBrimble:
```suggestion
In brief: when static routing is present for a Worker with assets, routing via those static rules takes precedence. When a request is evaluated in the Router Worker, the request path is first compared to the `"asset_worker"` rules (which are to be specific via "negative" rules, e.g. `"!/api/assets"`). If any match, the request is forwarded directly to the Asset Worker. If instead any `"user_worker"` rules match, the request is forwarded directly to the User Worker. If neither match (or static routing was not provided), the existing behavior takes over.
```

.changeset/tasty-hoops-own.md
```diff
@@ -0,0 +1,11 @@
...
+
+As part of this explicit routing, when static routing is present, the check against "Sec-Fetch-Mode: navigate" (to guess if this should serve an asset or go to the User Worker for not_found_handling) is disabled. Routing can be controlled by setting routing rules via "run_worker_first", and asset serving (including when an index.html or 404.html page is served) will be more simple.
```
GregBrimble:
```suggestion
As part of this explicit routing, when static routing is present, the check against `Sec-Fetch-Mode: navigate` (to decide if this should serve an asset or go to the User Worker for not_found_handling) is disabled. Routing can be controlled by setting routing rules via `assets.run_worker_first` in your Wrangler configuration file.
```

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1325,41 +1325,56 @@ describe("[Asset Worker] `canFetch`", () => {
 			[{ "Sec-Fetch-Mode": "cors" }, false],
 		] as const;
 
+		const staticRoutingModes = [
```
GregBrimble:
```suggestion
		const hasStaticRoutingModes = [
```

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1325,41 +1325,56 @@ describe("[Asset Worker] `canFetch`", () => {
...
 				for (const headers of headersModes) {
-					matrix.push({
-						compatibilityDate: compatibilityOptions[0].compatibilityDate,
-						compatibilityFlags: compatibilityOptions[0].compatibilityFlags,
-						notFoundHandling: notFoundHandling[0],
-						headers: headers[0],
-						expected:
-							compatibilityOptions[1] && notFoundHandling[1] && headers[1],
-					});
+					for (const hasStaticRouting of staticRoutingModes) {
```
GregBrimble:
```suggestion
					for (const hasStaticRouting of hasStaticRoutingModes) {
```

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1325,41 +1325,56 @@ describe("[Asset Worker] `canFetch`", () => {
...
-							not_found_handling: notFoundHandling,
-						}),
+						{
+							...normalizeConfiguration({
+								compatibility_date: compatibilityDate,
+								compatibility_flags: compatibilityFlags,
+								not_found_handling: notFoundHandling,
+								has_static_routing: hasStaticRouting,
+							}),
+						},
```
GregBrimble:
```suggestion
						normalizeConfiguration({
							compatibility_date: compatibilityDate,
							compatibility_flags: compatibilityFlags,
							not_found_handling: notFoundHandling,
							has_static_routing: hasStaticRouting,
						}),
```

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1369,11 +1384,14 @@ describe("[Asset Worker] `canFetch`", () => {
...
```
GregBrimble:
```suggestion
						normalizeConfiguration({
							compatibility_date: compatibilityDate,
							compatibility_flags: compatibilityFlags,
							not_found_handling: notFoundHandling,
							has_static_routing: hasStaticRouting,
						}),
```

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1383,11 +1401,14 @@ describe("[Asset Worker] `canFetch`", () => {
...
```
GregBrimble:
```suggestion
						normalizeConfiguration({
							compatibility_date: compatibilityDate,
							compatibility_flags: compatibilityFlags,
							not_found_handling: notFoundHandling,
							has_static_routing: hasStaticRouting,
						}),
```

packages/workers-shared/asset-worker/tests/handler.test.ts
```diff
@@ -1397,11 +1418,14 @@ describe("[Asset Worker] `canFetch`", () => {
...
```
GregBrimble:
```suggestion
						normalizeConfiguration({
							compatibility_date: compatibilityDate,
							compatibility_flags: compatibilityFlags,
							not_found_handling: notFoundHandling,
							has_static_routing: hasStaticRouting,
						}),
```

packages/workers-shared/asset-worker/tests/rules-engine.test.ts
```diff
@@ -115,3 +119,253 @@ describe("replacer", () => {
...
+				matcher({ request: new Request("https://example.com/baz") })
+			).toEqual(false);
+			expect(
+				matcher({ request: new Request("https://example.com/baz/foo") })
+			).toEqual(false);
+			expect(
+				matcher({ request: new Request("https://example.com/foobar") })
+			).toEqual(false);
+		}
+
```
GregBrimble, line 214:
```suggestion

		{
			const matcher = generateStaticRoutingRuleMatcher(["/:placeholder"]);
			expect(
				matcher({ request: new Request("http://example.com/foo") })
			).toEqual(false);
			expect(
				matcher({ request: new Request("https://example.com/:placeholder") })
			).toEqual(true);
		}

```

packages/workers-shared/router-worker/src/configuration.ts
```diff
@@ -10,5 +10,8 @@ export const applyConfigurationDefaults = (
 		account_id: configuration?.account_id ?? -1,
 		script_id: configuration?.script_id ?? -1,
 		debug: configuration?.debug ?? false,
+		static_routing: configuration?.static_routing ?? {
+			worker: [],
```
GregBrimble:
```suggestion
			user_worker: [],
```

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -75,6 +77,69 @@ export default {
...
+				// evaluate "exclude" rules
+				const excludeRulesMatcher = generateStaticRoutingRuleMatcher(
+					config.static_routing.asset ?? []
```
GregBrimble:
```suggestion
					config.static_routing.asset_worker ?? []
```

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -75,6 +77,69 @@ export default {
...
+							asset: "static_routing",
+							dispatchType: DISPATCH_TYPE.ASSETS,
+						});
+
+						return env.ASSET_WORKER.fetch(maybeSecondRequest);
+					});
+				}
+				// evaluate "include" rules
+				const includeRulesMatcher = generateStaticRoutingRuleMatcher(
+					config.static_routing.worker
```
GregBrimble:
```suggestion
					config.static_routing.user_worker
```

packages/workers-shared/router-worker/tests/index.test.ts
```diff
@@ -94,4 +94,156 @@ describe("unit tests", async () => {
...
+
+	it("it returns fetch from user worker when static_routing worker rule matches", async () => {
+		const request = new Request("https://example.com/api/includeme");
+		const ctx = createExecutionContext();
+
+		const env = {
+			CONFIG: {
+				has_user_worker: true,
+				static_routing: {
+					worker: ["/api/*"],
```
GregBrimble:
```suggestion
					user_worker: ["/api/*"],
```

packages/workers-shared/router-worker/tests/index.test.ts
```diff
@@ -94,4 +94,156 @@ describe("unit tests", async () => {
...
+	it("it returns fetch from asset worker when static_routing asset rule matches", async () => {
+		const request = new Request("https://example.com/api/excludeme");
+		const ctx = createExecutionContext();
+
+		const env = {
+			CONFIG: {
+				has_user_worker: true,
+				static_routing: {
+					worker: ["/api/includeme"],
+					asset: ["/api/excludeme"],
```
GregBrimble:
```suggestion
					user_worker: ["/api/includeme"],
					asset_worker: ["/api/excludeme"],
```

packages/workers-shared/router-worker/tests/index.test.ts
```diff
@@ -94,4 +94,156 @@ describe("unit tests", async () => {
...
+	it("it returns fetch from asset worker when static_routing asset and worker rule matches", async () => {
+		const request = new Request("https://example.com/api/excludeme");
+		const ctx = createExecutionContext();
+
+		const env = {
+			CONFIG: {
+				has_user_worker: true,
+				static_routing: {
+					worker: ["/api/*"],
+					asset: ["/api/excludeme"],
```
GregBrimble:
```suggestion
					user_worker: ["/api/*"],
					asset_worker: ["/api/excludeme"],
```

packages/workers-shared/router-worker/tests/index.test.ts
```diff
@@ -94,4 +94,156 @@ describe("unit tests", async () => {
...
+
+	it("it returns fetch from asset worker when no static_routing rule matches but asset exists", async () => {
+		const request = new Request("https://example.com/someasset");
+		const ctx = createExecutionContext();
+
+		const env = {
+			CONFIG: {
+				has_user_worker: true,
+				static_routing: {
+					worker: ["/api/*"],
```
GregBrimble:
```suggestion
					user_worker: ["/api/*"],
```

packages/workers-shared/router-worker/tests/index.test.ts
```diff
@@ -94,4 +94,156 @@ describe("unit tests", async () => {
...
+
+	it("it returns fetch from user worker when no static_routing rule matches and no asset exists", async () => {
+		const request = new Request("https://example.com/somemissingasset");
+		const ctx = createExecutionContext();
+
+		const env = {
+			CONFIG: {
+				has_user_worker: true,
+				static_routing: {
+					worker: ["/api/*"],
```
GregBrimble:
```suggestion
					user_worker: ["/api/*"],
```

packages/workers-shared/utils/types.ts
matthewdavidrodgers: updated to handle the splitting in EWC

packages/workers-shared/asset-worker/src/handler.ts
```diff
@@ -203,7 +203,9 @@ export const canFetch = async (
 			flagIsEnabled(
 				configuration,
 				SEC_FETCH_MODE_NAVIGATE_HEADER_PREFERS_ASSET_SERVING
-			) && request.headers.get("Sec-Fetch-Mode") === "navigate"
+			) &&
+			request.headers.get("Sec-Fetch-Mode") === "navigate" &&
+			!configuration.has_static_routing
```
GregBrimble: Can we also test this? It would be great to see this as:

1. a test making a `Sec-Fetch-Mode: navigate` request who's outcome changes based off of the presence of unmatched static routing, and
2. a test which uses the `env.ASSETS.unstable_canFetch(request)` API, again showing different results based off of the presence/absence of static routing.

GregBrimble (approved):

packages/workers-shared/utils/types.ts
matthewdavidrodgers: i'll push back on this only from the point of view that this is currently user-facing configuration (the public api contains this config, and EWC passes it through after validating). so i don't think that it makes more sense to present a user with "user_worker" and "asset_worker" - they have no notion of an asset worker, just their assets.

however, we have the ability to transform this in EWC: the public api would remain  "worker" and "asset" but EWC transforms it into this proposed schema. i think that's not necessary but can quickly do so to keep things clear if that helps.

packages/workers-shared/asset-worker/src/handler.ts
matthewdavidrodgers: there are tests for 1. currently, but located in https://github.com/cloudflare/workers-sdk/pull/9304 (they're fixtures, so they depended on the wrangler implementation). i can add a unit test here

for 2., that seems like the same test as 1. - there's no additional behavior we'd be testing. wouldn't complain if you're just looking to get a more unit-style test here as compared to 1.

packages/workers-shared/utils/types.ts
jamesopstad: Isn't that where it was already handled? We need to use this splitting logic in Miniflare and the Vite plugin for dev so ideally there would be a single implementation that can be shared.

packages/workers-shared/utils/types.ts
GregBrimble: > this is currently user-facing configuration

Wait, is it? Okay, I need to go review your backend PR :) I thought this was all being lumped into `assets.run_worker_first` as an unlabelled array of `["positive", "!negative"]` matches. I figured that was how it was being passed to the API too, but I'll take a look.

Not blocking. We're inconsistent in a few places anyway. But I'll double-check this backend and get back to you :)

packages/workers-shared/asset-worker/src/handler.ts
GregBrimble: Ah yeah, I asked for it there too: https://github.com/cloudflare/workers-sdk/pull/9304#discussion_r2103427738. Sorry, I've lost track with all these PRs you all have in and out and reverted. Fine that they're there if adding a unit test here is even slightly annoying.

Yea, I think I am asking for a unit-style test explicitly covering use of that binding. I just want to ensure we're not ever regressing that API either (even though I know that API is used _today_ to power regular request flows too).

Non-blocking though, so again, if it's at all a hassle, I'm fine to skip.

packages/workers-shared/utils/types.ts
matthewdavidrodgers: whoops, that was not what i meant. updated to _accept_ already-split routing in EWC

packages/workers-shared/utils/types.ts
matthewdavidrodgers: yeah, the schema in wrangler.jsonc is indeed an unlabelled array. but the router worker will receive it as split and labelled, so we could either have EWC do so, Wrangler do so. Given that Wrangler already has to do so to support this in dev, @emily-shen thought it might be better to just let Wrangler do the splitting and only implement it in one place.

if the divergence between the Wrangler configuration and the public API feels like the wrong decision (and we're certainly guilty of unnecessarily doing so in the past) we can make the call to undo that and keep things those APIs in sync (even if that means duplicated logic)

GregBrimble: I vote dupe. I think Wrangler config should look the same as our API here, personally. But I don't put my foot down on it — happy to be overruled if I'm out-voted.

packages/workers-shared/asset-worker/src/handler.ts
matthewdavidrodgers: yeah, it's a bit of a mess between the 3 PRs, sorry about that. I'll add a unit-style test here in the pursuit of completeness

packages/workers-shared/utils/types.ts
matthewdavidrodgers: i do think i agree, the more i think about it. ok, will fix this up, and that will make user_worker and asset_worker sensible keys here

packages/workers-shared/utils/types.ts
jamesopstad: I see you've switched to keeping the splitting in EWC to make things consistent with the API. That makes sense. Out of curiosity, what's the reasoning for not having the splitting executed in the Worker e.g. a `generateStaticRoutingRuleMatchers` function that returns `{ excludeRulesMatcher, includeRulesMatcher }`?

GregBrimble (approved):

packages/workers-shared/utils/types.ts
matthewdavidrodgers: as mentioned async; the split is just to save a few cycles in the router worker. having the lists pre-split means the router worker has less cpu work required to split and/or traverse the entire set of paths just to find the exclude paths

hard to imagine that work mattering, but that was the intent

jamesopstad (approved):

emily-shen:
as its on a fork, the tests were run here: https://github.com/cloudflare/workers-sdk/pull/9491. however, if i used actually the same sha (so github could associate tests passing on that PR with this PR), it was still running the test on the fork 🫠 

so i updated the commit message so the sha is different, it ran on the fork successfully, but now the passing tests don't get associated with the original PR. 

so we're going to bypass checks on this PR, but the tests _did_ pass, _somewhere_ 🫠

CarmenPopoviciu:
I went the less elegant way and checked out this PR locally, then pushed to the remote repo and [PRed](https://github.com/cloudflare/workers-sdk/pull/9505) from my branch. This seems to have done the trick and running the tests on my PR ran the tests on this PR as well. All is 🍏 so this is ready to merge 🎉

PR #9443, opened 2025-06-02
add mTLS mixed mode support

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1900

This PR is adding mTLS support to mixed mode.

As part of this the PR is also adding a `workerName` option to the `startMixedModeSession` API since when starting
the remote preview with an mTLS binding we would normally get the following error:
```
Workers with mTLS bindings need to be deployed before doing edge preview
``` 

having the option to pass the worker's name to `startMixedModeSession` allows us to make the server proxy worker behave like the actual user's worker thus enabling mTLS usage for it (in case the user has already deployed the worker)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: experimental feature
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not related to a v3 feature

### DIFF
--- a/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
+++ b/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
@@ -1,9 +1,16 @@
+import assert from "node:assert";
 import path from "node:path";
 import dedent from "ts-dedent";
 import { beforeEach, describe, expect, it, onTestFinished } from "vitest";
+import {
+	generateLeafCertificate,
+	generateMtlsCertName,
+	generateRootCertificate,
+} from "./helpers/cert";
 import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
 import { generateResourceName } from "./helpers/generate-resource-name";
-import type { Binding } from "../src/api";
+import type { startMixedModeSession } from "../src/api";
+import type { RawConfig } from "../src/config";
 import type { MixedModeConnectionString, WorkerOptions } from "miniflare";
 import type { ExpectStatic } from "vitest";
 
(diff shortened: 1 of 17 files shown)


packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
```diff
@@ -283,67 +316,153 @@ const testCases: TestCase<string>[] = [
...
+		},
+	}),
+	matches: [
+		// Note: in this test we are making sure that TLS negotiation does work by checking that we get an SSL certificate error
+		expect.stringMatching(/The SSL certificate error/),
+		expect.not.stringMatching(/No required SSL certificate was sent/),
+	],
+};
+
+describe.each([...testCases, mtlsTest])("Mixed Mode for $name", (testCase) => {
```
penalosa, line 406: Why this pattern? Why not include mtls in the testCases array?

penalosa (approved):

packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts
dario-piotrowicz: The reason for this pattern are the types, specifically see:
 - https://github.com/cloudflare/workers-sdk/blob/676e490666c20c58a457a46fdf01530b7a098852/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts#L30
 - https://github.com/cloudflare/workers-sdk/blob/676e490666c20c58a457a46fdf01530b7a098852/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts#L323
 
These generic type specifications make the objects within the arrays behave like they should, combining the two arrays would require something like:
```ts
const testCases: TestCase<string | { certificateId: string; workerName: string }>[] = [
```
the problem with this is that then the callbacks ([this](https://github.com/cloudflare/workers-sdk/blob/676e490666c20c58a457a46fdf01530b7a098852/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts#L22) and [this](https://github.com/cloudflare/workers-sdk/blob/676e490666c20c58a457a46fdf01530b7a098852/packages/wrangler/e2e/miniflare-mixed-mode-resources.test.ts#L23-L26)) of the TestCase instances would assume that the generic type is the union, so basically in every single entry we would need to do add checks to discern whether arguments are of type string or object, adding a lot of ugly and unnecessary verbosity.

I hope that this makes sense 🙂 

Please let me know what you think and if I've missed something 🙂

penalosa: Ah, I see. Yeah, that seems fair enough then 👍

PR #9447, opened 2025-06-03
VP fixups

emily-shen:
fix container fixture on linux being weird about formatting

fix changeset

### DIFF
--- a/.changeset/five-ladybugs-sell.md
+++ b/.changeset/five-ladybugs-sell.md
@@ -1,6 +1,6 @@
 ---
-"miniflare": patch
-"wrangler": patch
+"miniflare": minor
+"wrangler": minor
 ---
 
 Support additional Mixed Mode resources in Wrangler:
--- a/.changeset/five-ladybugs-sell.md
+++ b/.changeset/five-ladybugs-sell.md
@@ -1,6 +1,6 @@
 ---
-"miniflare": minor
-"wrangler": minor
+"miniflare": patch
+"wrangler": patch
 ---
 
 Support additional Mixed Mode resources in Wrangler:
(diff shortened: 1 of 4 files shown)


.changeset/five-ladybugs-sell.md
```diff
@@ -1,6 +1,6 @@
 ---
-"miniflare": patch
-"wrangler": patch
+"miniflare": minor
+"wrangler": minor
```
dario-piotrowicz: Nit: these should not be minors since these are additions to experimental features and I think we all agreed that these are generally marked as patches 🤔

jamesopstad (approved):

fixtures/container-app/tests/index.test.ts
```diff
@@ -21,11 +21,6 @@ describe("Containers local dev", () => {
 		expect,
 	}) => {
 		expect(getOutput()).toContain(dedent`
-			Your Worker has access to the following bindings:
-			Binding                        Resource            Mode
-			env.CONTAINER (Container)      Durable Object      local
-
-			⎔ Starting local server...
```
dario-piotrowicz, line 28: is this indicating that we've got some issue with containers?

.changeset/five-ladybugs-sell.md
emily-shen: oh i totally misread the gchat thread, i'll revert this

fixtures/container-app/tests/index.test.ts
emily-shen: no, we were just getting an extra empty line somewhere in there on linux so the matching wasn't working. the lines below are the bits we really care about.

PR #9451, opened 2025-06-03
chore: update node versions in various places

emily-shen:
- update `engines`, `volta.node` field in package.jsons
- update c3 tests to use a slightly more recent release of node 20, since the angular cli dependabot bump needed > 20.19

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: should just make sure existing tests don't fail
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: should just make sure existing tests don't fail
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal chore
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: v3 doesn't follow same node version policy

### DIFF
--- a/.github/workflows/c3-e2e.yml
+++ b/.github/workflows/c3-e2e.yml
@@ -39,7 +39,7 @@ jobs:
         if: steps.changes.outputs.everything_but_markdown == 'true'
         uses: ./.github/actions/install-dependencies
         with:
-          node-version: 20.19.2
+          node-version: 20
           turbo-api: ${{ secrets.TURBO_API }}
           turbo-team: ${{ secrets.TURBO_TEAM }}
           turbo-token: ${{ secrets.TURBO_TOKEN }}
@@ -49,7 +49,7 @@ jobs:
         if: steps.changes.outputs.everything_but_markdown == 'true'
         uses: ./.github/actions/run-c3-e2e
         with:
-          node-version: 20.19.2
+          node-version: 20
           packageManager: ${{ matrix.pm.name }}
           packageManagerVersion: ${{ matrix.pm.version }}
           quarantine: false
@@ -61,7 +61,7 @@ jobs:
         if: steps.changes.outputs.everything_but_markdown == 'true'
         uses: ./.github/actions/run-c3-e2e
         with:
-          node-version: 20.19.2
+          node-version: 20
           packageManager: ${{ matrix.pm.name }}
           packageManagerVersion: ${{ matrix.pm.version }}
           quarantine: false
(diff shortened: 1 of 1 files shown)


.github/workflows/c3-e2e.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         if: steps.changes.outputs.everything_but_markdown == 'true'
         uses: ./.github/actions/install-dependencies
         with:
-          node-version: 20.11.1
```
edmundhung, line 42: Should we make this `20` so that it picks up the latest version all the time?

packages/wrangler/package.json
```diff
@@ -169,7 +169,7 @@
 		"fsevents": "~2.3.2"
 	},
 	"engines": {
-		"node": ">=18.0.0"
```
edmundhung, line 172: I _think_ this will make users running Wrangler on node 18 a hard failure?

emily-shen: we already hard fail on node 18 right? i'm not sure if i've misunderstood

edmundhung: True. I think the difference might be the error message that show up as node (or npm?) will fails it before Wrangler runs, so users might find the error less helpful, which might be fine?

edmundhung: I was wrong. Just give it a test with node 18 and it looks good 👍🏼 

```sh
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'wrangler@0.0.0-ve114e40fd',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
npm warn EBADENGINE   current: { node: 'v18.20.4', npm: '10.7.0' }
npm warn EBADENGINE }
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: '@cloudflare/kv-asset-handler@0.0.0-ve114e40fd',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
npm warn EBADENGINE   current: { node: 'v18.20.4', npm: '10.7.0' }
npm warn EBADENGINE }
npm warn EBADENGINE Unsupported engine {
npm warn EBADENGINE   package: 'miniflare@0.0.0-ve114e40fd',
npm warn EBADENGINE   required: { node: '>=20.0.0' },
npm warn EBADENGINE   current: { node: 'v18.20.4', npm: '10.7.0' }
npm warn EBADENGINE }
Wrangler requires at least Node.js v20.0.0. You are using v18.20.4. Please update your version of Node.js.

Consider using a Node.js version manager such as https://volta.sh/ or https://github.com/nvm-sh/nvm.
```

edmundhung (approved):

PR #9461, opened 2025-06-04
Consider disk limits when building containers

skepticfx:
Fixes #CC-5270

When building container images during docker build, we'd want to ensure the current docker image can be run under the current account. This enforces an useful check that improves DX without having to be denied during application creation with the specific image.

There is also a fix for stuck container builds in some cases where we were not inheriting the right stdio flags for the node spawn.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: this path is already checked in unit tests and the default paths are tested in E2E Tests.
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s):  
  - [x] Documentation not necessary because: The docs are private and already talks about limits.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: these are new features  Not needed because these are new product features in beta.

### DIFF
--- a/.changeset/empty-shirts-judge.md
+++ b/.changeset/empty-shirts-judge.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Enforce disk limits on container builds
--- a/packages/wrangler/src/cloudchamber/build.ts
+++ b/packages/wrangler/src/cloudchamber/build.ts
@@ -2,13 +2,20 @@ import { spawn } from "child_process";
 import { stat } from "fs/promises";
 import { getCIOverrideNetworkModeHost } from "../environment-variables/misc-variables";
 import { UserError } from "../errors";
+import { logger } from "../logger";
 import { ImageRegistriesService } from "./client";
+import { resolveAppDiskSize } from "./common";
+import { loadAccount } from "./locations";
 import type { Config } from "../config";
+import type { ContainerApp } from "../config/environment";
 import type {
 	CommonYargsArgvJSON,
 	StrictYargsOptionsToInterfaceJSON,
 } from "../yargs-types";
-import type { ImageRegistryPermissions } from "./client";
+import type {
+	CompleteAccountCustomer,
+	ImageRegistryPermissions,
+} from "./client";
 
 // default cloudflare managed registry
 const domain = "registry.cloudchamber.cfdata.org";
@@ -101,7 +108,7 @@ export function dockerBuild(options: {
 		const buildExec = options.buildCmd.split(" ").shift();
 		const child = spawn(String(buildExec), buildCmd, {
 			stdio: [
-				options.dockerfile !== undefined ? "pipe" : undefined,
+				options.dockerfile !== undefined ? "pipe" : "inherit",
 				"inherit",
 				"inherit",
 			],
@@ -212,6 +219,8 @@ export type BuildArgs = {
 	push: boolean;
 	// specify the contents of the dockerfile if not wanting to use the dockerfile directory
 	dockerfileContents?: string;
+	// sometimes the container config is not set, and we can use this to run other subcommands in the current context
+	container?: ContainerApp;
 	args?: Record<string, string>;
 };
 
@@ -239,6 +248,22 @@ export async function build(args: BuildArgs): Promise<string> {
 		});
 		await dockerBuild({ buildCmd: bc, dockerfile: args.dockerfileContents });
 
+		// ensure the account is not allowed to build anything that exceeds the current
+		// account's disk size limits
+		const account = await loadAccount();
+		const { size, layers } = await runDockerInspect(
+			getDefaultRegistry() + "/" + args.tag,
+			args.pathToDocker ?? "docker"
+		);
+		// 16MiB is the layer size adjustments we use in devmapper
+		const MiB = 1024 * 1024;
+		const requiredSize = Math.ceil(size * 1.1 + layers * 16 * MiB);
+		await ensureDiskLimits({
+			requiredSize,
+			account: account,
+			containerApp: args.container,
+		});
+
 		if (args.push) {
 			await dockerLoginManagedRegistry({
 				pathToDocker: args.pathToDocker,
@@ -259,14 +284,19 @@ export async function build(args: BuildArgs): Promise<string> {
 
 export async function buildCommand(
 	args: StrictYargsOptionsToInterfaceJSON<typeof buildYargs>,
-	_: Config
+	config: Config
 ) {
-	await build({
-		tag: args.tag,
-		pathToDockerfileDirectory: args.PATH,
-		pathToDocker: args.pathToDocker,
-		push: args.push,
-	});
+	// if containers are not defined, the build should still work.
+	const containers = config.containers ?? [undefined];
+	for (const container of containers.length > 0 ? containers : [undefined]) {
+		await build({
+			container,
+			tag: args.tag,
+			pathToDockerfileDirectory: args.PATH,
+			pathToDocker: args.pathToDocker,
+			push: args.push,
+		});
+	}
 }
 
 export async function pushCommand(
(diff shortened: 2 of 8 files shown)


packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -287,3 +300,56 @@ export async function pushCommand(
...
+	const maxAllowedImageSizeBytes =
+		(options.account.limits.disk_mb_per_deployment ?? 2000) * 1000 * 1000;
+
+	const { size, layers } = await runDockerInspect(imageTag, dockerPath);
+	const adjustedSize = Math.ceil(size * 1.1 + layers * 16 * 1000 * 1000);
+
+	if (maxAllowedImageSizeBytes < adjustedSize) {
+		throw new Error(
+			`Image too large: needs ${Math.ceil(adjustedSize / 1000 / 1000)} MB, but your account is limited to images with size ${maxAllowedImageSizeBytes / 1000 / 1000} MB. Your account needs more disk size per instance to run this container.`
+		);
```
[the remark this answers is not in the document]
skepticfx: Thanks good suggestion.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -287,3 +300,56 @@ export async function pushCommand(
...
+	pathToDocker: string;
+	account: CompleteAccountCustomer;
+}): Promise<string> {
+	const imageTag = domain + "/" + options.imageTag;
+	const dockerPath = options.pathToDocker ?? "docker";
+	const maxAllowedImageSizeBytes =
+		(options.account.limits.disk_mb_per_deployment ?? 2000) * 1000 * 1000;
+
+	const { size, layers } = await runDockerInspect(imageTag, dockerPath);
+	const adjustedSize = Math.ceil(size * 1.1 + layers * 16 * 1000 * 1000);
```
gpanders: The adjustment in devmapper is 16 MiB, not 16 MB. Should keep it the same here.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -287,3 +300,56 @@ export async function pushCommand(
...
+
+	if (maxAllowedImageSizeBytes < adjustedSize) {
+		throw new UserError(
+			`Image too large: needs ${Math.ceil(adjustedSize / 1000 / 1000)} MB, but your account is limited to images with size ${maxAllowedImageSizeBytes / 1000 / 1000} MB. Your account needs more disk size per instance to run this container.`
```
gpanders: Shouldn’t we be checking against the disk size of the application? For example, a user’s account level disk limit could be 20 GB but if the application is configured with a 2 GB disk, then a 2.5 GB image should fail.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -287,3 +300,56 @@ export async function pushCommand(
...
+		proc.stderr.on("data", (chunk) => (stderr += chunk));
+
+		proc.on("close", (code) => {
+			if (code !== 0) {
+				return reject(
+					new Error(`failed inspecting image locally: ${stderr.trim()}`)
+				);
+			}
+			const [sizeStr, layerStr] = stdout.trim().split(" ");
+			resolve({ size: parseInt(sizeStr, 10), layers: parseInt(layerStr, 10) });
```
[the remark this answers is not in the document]
skepticfx: Can you suggest some validation here?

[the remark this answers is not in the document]
0xbad0c0d3: > Can you suggest some validation here?
```suggestion
			const [sizeStr, layerStr] = stdout.trim().split(" ");
			const size = parseInt(sizeStr, 10);
			const layers = parseInt(layerStr, 10);
                        if (Number.isNaN(size) || Number.isNaN(layers)){
                            reject(new Error('size and/or layers are wrongly formatted'))
                        } else{
                            resolve({ size, layers });
                        }
```
Something like this, I guess, must be OK. Or another way is regex
```suggestion
                        const output = stdout.trim()
			if (/^\d+\s+\d+/.test(output)){
                            const [sizeStr, layerStr] = output.split(" ");
			    resolve({ size: parseInt(sizeStr, 10), layers: parseInt(layerStr, 10) });
			} else {
			    reject(new Error('Something went wrong with the inspect output'))
			}
```

.changeset/empty-shirts-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
emily-shen:
```suggestion
"wrangler": patch
```

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -101,7 +105,7 @@ export function dockerBuild(options: {
 		const buildExec = options.buildCmd.split(" ").shift();
 		const child = spawn(String(buildExec), buildCmd, {
 			stdio: [
-				options.dockerfile !== undefined ? "pipe" : undefined,
+				options.dockerfile !== undefined ? "pipe" : "inherit",
```
emily-shen, line 111: not really related to this PR, but i've been doing a bit of refactoring here to pull out the build stuff so we can use that in miniflare for local dev too, and i was wondering why pipe the dockerfile in rather than just passing the path to the cli?

(i'll get this PR in first so you don't have any merge conflicts!)

packages/wrangler/src/cloudchamber/build.ts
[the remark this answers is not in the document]
emily-shen: what about setting `--format json` and using that output?

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -287,3 +300,56 @@ export async function pushCommand(
 		throw new UserError("An unknown error occurred");
 	}
 }
+
+export async function ensureDiskLimits(options: {
```
emily-shen, line 321: could you add a unit test for this?

gabivlj: Agreed, there is some examples in the unit tests too on how to mock the docker cli!

packages/wrangler/src/cloudchamber/build.ts
skepticfx: @IRCody Do we know why we pipe here instead of passing the path?

skepticfx: @gabivlj added this, might know the why?

gabivlj: No strong opinion other than its path independent and less prone to fs races (once we load the dockerfile into memory, we know we have it until the command exits)

skepticfx: @emily-shen Is it possible to keep the "pipe" here? We'd like to keep this open to add our own transformations to the Dockerfile from what the user provides. Using "pipe" would help us do those without a temporary file.

emily-shen: are fs races something that happens often with docker, or is this more of a general concern? 

Re transforming the dockerfile, creating a separate file might actually be useful, so the user could inspect it for debugging purposes.

But again totally don't mind if you guys prefer to pipe the file in, not a big deal :)

skepticfx: The FS races are more of a general concern, we'd like to allow multiple container builds in parallel if the user does so in Makefile one day. This just keeps options open for us. Either way, this is not a requirement today - but might make it easier to support transformations and parallel builds in the future.

packages/wrangler/src/cloudchamber/build.ts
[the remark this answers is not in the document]
skepticfx: Thanks, I was just curious what the windsurf AI bot had to say.

packages/wrangler/src/cloudchamber/build.ts
skepticfx: Now we check the app config and also check account limits during build if the app configuration already exceeds this limit.

packages/wrangler/src/cloudchamber/build.ts
skepticfx: Added unit tests.

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -19,7 +19,7 @@ export async function buildContainers(config: Config, imageTag: string) {
 	for (const container of config.containers) {
 		const options = getBuildArguments(container, imageTag, true);
 		if (options.isDockerImage) {
-			await build(options);
+			await build({ ...options, container: container });
```
0xbad0c0d3: Why not just to do it inside getBuildArguments? Will be more simple and more consistent

packages/wrangler/src/cloudchamber/build.ts
emily-shen: Ah gotcha that makes sense! I'll leave it in :)

packages/wrangler/src/cloudchamber/deploy.ts
skepticfx: Yeah, it shouldn't be. Moved it.

packages/wrangler/src/cloudchamber/build.ts
skepticfx: Also changed the deploy.test.ts a bit to accommodate ordered events of the docker spawns, because with this change we add a docker inspect after the docker build.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -287,3 +313,70 @@ export async function pushCommand(
...
+		);
+	}
+	const maxAllowedImageSizeBytes = appDiskSize ?? accountDiskSize;
+
+	logger.debug(
+		`Disk size limits when building the container: appDiskSize:${appDiskSize}, accountDiskSize:${accountDiskSize}, maxAllowedImageSizeBytes=${maxAllowedImageSizeBytes}(${maxAllowedImageSizeBytes / MB} MB), requiredSized=${options.requiredSize}(${Math.ceil(options.requiredSize / MiB)}MiB)`
+	);
+	if (maxAllowedImageSizeBytes < options.requiredSize) {
+		throw new UserError(
+			`Image too large: needs ${Math.ceil(options.requiredSize / MiB)} MiB, but your app is limited to images with size ${maxAllowedImageSizeBytes / MB} MB. Your account needs more disk size per instance to run this container. The default disk size is 2GB.`
```
emily-shen: is there anything users can do to increase their account's disk size limit?

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -259,9 +284,10 @@ export async function build(args: BuildArgs): Promise<string> {
 
 export async function buildCommand(
 	args: StrictYargsOptionsToInterfaceJSON<typeof buildYargs>,
-	_: Config
+	config: Config
 ) {
 	await build({
+		container: config.containers?.[0],
```
emily-shen: why `[0]`?

packages/wrangler/src/cloudchamber/build.ts
skepticfx: Yes, they'd need to talk to us. But the limits(2GB) are already mentioned in the docs that will be made public when we go open beta.

packages/wrangler/src/cloudchamber/build.ts
skepticfx: Good question.This specific build command is the one used on our app when running the "build" subcommand from a repo. Since the "containers" type is an array today, am accommodating this by selecting the default(ie first container). In the future, we'd allow selecting a specific container by a given name when executing these specific commands. That is a multi-container world with DO bindings within the same worker. We don't fully support that yet. When we do add support, this would instead select the given container instead of the first one.

skepticfx: I'm iterating through all containers now since we do this for deploy as well.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -259,14 +284,19 @@ export async function build(args: BuildArgs): Promise<string> {
...
 ) {
-	await build({
-		tag: args.tag,
-		pathToDockerfileDirectory: args.PATH,
-		pathToDocker: args.pathToDocker,
-		push: args.push,
-	});
+	// if containers are not defined, the build should still work.
+	const containers = config.containers ?? [undefined];
+	for (const container of containers.length > 0 ? containers : [undefined]) {
```
emily-shen, line 291:
```suggestion
	const containers = config.containers ?? [undefined];
	for (const container of containers) {
```
nit

emily-shen (approved):

packages/wrangler/src/cloudchamber/build.ts
skepticfx: I'm going to revisit this section sooner for more changes next week. Will add this suggestion! Thanks!

PR #9473, opened 2025-06-04
add experimental support for mixed mode in vitest-pool-workers

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1856

This PR adds support for mixed mode in vitest-pool-workers it also introduces a public utility to start or update mixed mode sessions.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: experimental feature change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/fixtures/vitest-pool-workers-mixed-mode/run-tests.mjs
+++ b/fixtures/vitest-pool-workers-mixed-mode/run-tests.mjs
@@ -1,3 +1,12 @@
+/**
+ * This fixture is particular since it needs to communicate with remote resources, namely
+ * a remote worker.
+ *
+ * This script is used to deploy a remote worker and run the fixture using said worker.
+ *
+ * Alternatively you can simply deploy, using your account, the `./remote-worker.js` file as
+ * a worker named `my-worker-test` and directly run the fixture using vitest.
+ */
 import { execSync } from "child_process";
 import { randomUUID } from "crypto";
 import { cpSync, readFileSync, rmSync, writeFileSync } from "fs";


fixtures/vitest-pool-workers-mixed-mode/.gitignore
penalosa, line 1: I think this should probably be in the top-level gitignore?

fixtures/vitest-pool-workers-mixed-mode/run-tests.mjs
penalosa, line 1: This seems fairly complex—why does this need to be a custom script? What about using Vitest's setup file?

penalosa: Additionally, is there a reason it's not in the `vitest-pool-workers-examples` fixture?

packages/wrangler/src/api/mixedMode/index.ts
```diff
@@ -97,3 +102,85 @@ export function pickRemoteBindings(
 		})
 	);
 }
+
+/** Map containing all the potential worker mixed mode existing sessions, it maps a worker name to its mixed mode session */
+const mixedModeSessionsMap = new Map<string, MixedModeSession>();
```
penalosa: This seems a bit off to me—why are we storing these in a module level map?

It feels like clients of Miniflare should have a better idea of how to store sessions (what if you want multiple parallel sessions for a single worker name, for instance)—what about just requiring the previous session to be passed in to the `maybeStartOrUpdate` helper?

fixtures/vitest-pool-workers-mixed-mode/run-tests.mjs
dario-piotrowicz: > Additionally, is there a reason it's not in the vitest-pool-workers-examples fixture?

That directory gets run by the vitest ecosystem ci, so we can't have there fixtures that require actual auth requirements

dario-piotrowicz: > This seems fairly complex

Yeah 😓 , if you have a better alternative in mind I'm totally happy to try a different approach 🙂 

> —why does this need to be a custom script? What about using Vitest's setup file?

I put this logic in the custom script so that the fixture could also be used by itself without the extra deployment logic.

Basically if you set up an appropriate remote worker called `my-worker-test` in your account can just easily run vitest here without needing any of the logic in this script. This decoupling seems nicer to me and I think it makes it easier/faster to make changes or troubleshoot issues.

If you disagree I am happy to move this logic in a vitest setup file instead 🙂

fixtures/vitest-pool-workers-mixed-mode/.gitignore
dario-piotrowicz: I don't have a very strong opinion on this... is you prefer it at the top level that's fine by me 🙂

dario-piotrowicz: I noticed that in the top level gitignore we're already ignoring `.tmp` directories, so I renamed the directory to be used here to that 🙂👍

packages/wrangler/src/api/mixedMode/index.ts
dario-piotrowicz: I was hoping to offload as much logic as possible from consumers of the api and have wrangler actually own the mixed mode sessions

(with always the optional possibility of consumers to pass and manage the mixed mode sessions on their own)

However I understand your concerns 👍 

Hopefully this looks ok to you? https://github.com/cloudflare/workers-sdk/pull/9473/commits/cf478021d9dc263c8629b5c3351ec3d6d5e1ac82

fixtures/vitest-pool-workers-mixed-mode/run-tests.mjs
penalosa: Gotcha! That makes sense. Could you explain that in a comment, because that's quite specific to this test

penalosa (approved):

fixtures/vitest-pool-workers-mixed-mode/run-tests.mjs
dario-piotrowicz: Good call! 😄 

Done: https://github.com/cloudflare/workers-sdk/pull/9473/commits/756ed2e5baac6f1c75495f7e0fdff959c876696d

TavoElgueaLopez:
📱

PR #9480, opened 2025-06-04
add `experimentalMixedMode` dev option to `unstable_startWorker`

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1958

This PR adds an `experimentalMixedMode` dev option to `unstable_startWorker` so that developers
can programmatically start a mixed mode session using startWorker, for more details please see the
changeset

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: experimental feature
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/e2e/start-worker-mixed-mode.test.ts
+++ b/packages/wrangler/e2e/start-worker-mixed-mode.test.ts
@@ -1,0 +1,147 @@
+import { readFile, writeFile } from "node:fs/promises";
+import { resolve } from "node:path";
+import { setTimeout } from "node:timers/promises";
+import dedent from "ts-dedent";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
+import { generateResourceName } from "./helpers/generate-resource-name";
+
+describe("startWorker - mixed mode", () => {
+	const remoteWorkerName = generateResourceName();
+	const helper = new WranglerE2ETestHelper();
+
+	beforeAll(async () => {
+		await helper.seed(resolve(__dirname, "./seed-files/mixed-mode-workers"));
+		await helper.seed({
+			"remote-worker.js": dedent/* javascript */ `
+					export default {
+						fetch() {
+							return new Response('Hello from a remote worker (startWorker mixed-mode)');
+						}
+					};
+			`,
+		});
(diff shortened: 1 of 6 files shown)


packages/wrangler/e2e/start-worker-mixed-mode.test.ts
```diff
@@ -0,0 +1,147 @@
...
+		await helper.run(
+			`wrangler deploy remote-worker.js --name ${remoteWorkerName} --compatibility-date 2025-01-01`
+		);
+	}, 35_000);
+
+	afterAll(async () => {
+		await helper.run(`wrangler delete --name ${remoteWorkerName}`);
+	});
+
+	[true, false].forEach((experimentalMixedMode) => {
```
penalosa: Ideally this should be a `describe.each()` block

penalosa (approved):

packages/wrangler/e2e/start-worker-mixed-mode.test.ts
dario-piotrowicz: Screenshot 2025-06-05 at 12 29 08

dario-piotrowicz: would you prefer keeping this as is? or using `describe.each()` and avoid inline snapshots?

(PS: please keep in mind that this will anyways just become the true case once mixed mode is stable 🙂 (unless we'll want an opt-in/out mechanism in SDW for it))

penalosa: I'd prefer using `describe.each()` and avoiding inline snapshots, if that's alright? The reason inline snapshots can't be used in a `describe.each()` is because the snapshot can't be updated independently for each case, which is also true here

dario-piotrowicz: sure I'll make the change 🙂👍

dario-piotrowicz: updated 🙂👍

PR #9495, opened 2025-06-05
[vitest-pool-workers] bump capnp-es to 0.0.10

vicb:


packages/vitest-pool-workers/scripts/rtti/rtti.js
emily-shen, line 1: it looks like a lot of deletions here, can you explain why that's happened?
i thought it was just formatting changes mostly.

vicb: As mentioned in the description:

> It directly output js so there is no need to build ts using esbuild.

You can also see from the diff that we use to generate a ts file and compile/bundle with ESBuild.
The deleted code is ESBuild helpers added during bundling.

emily-shen: yeah i got that we were no longer generating and compiling the ts files, but it was not immediately clear why rtti.js had so many deletions. for example, all the error codes have been deleted. is this all code that is not needed in js?

.changeset/violet-donkeys-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vitest-pool-workers": patch
+---
+
+bump capnp-es to 0.0.10
```
CarmenPopoviciu, line 5:
```suggestion
bump capnp-es to 0.0.11
```

packages/vitest-pool-workers/scripts/rtti/rtti.js
vicb: ESBuild bundles the file while capnp runtime is imported in the new file, see `import * as $ from "capnp-es";` on line 2

emily-shen (approved):

PR #9510, opened 2025-06-06
Full build support for assets

### LINKED ISSUE #9384: Allow Vite plugin to build server first, client last
### Describe the solution

I'm currently building a RSC framework/Vite plugin which I want to use in combination with the Cloudflare Vite plugin. For the RSC plugin to work, it is important that the server/ssr bundle is built first, and the client bundle is built last (because client components get discovered during SSR build and then included in the client build).

The Cloudflare Vite plugin currently takes over the build process via builder.buildApp, yielding to a user config in case there is one (see https://github.com/cloudflare/workers-sdk/blob/d9d937ab6f2868271dde5a8da625773085eaec85/packages/vite-plugin-cloudflare/src/index.ts#L142). I can provide my own builder.buildApp hook to make sure the build happens in the desired order, so this part works.

The one thing that doesn't work is that the assets config in the generated wrangler.json is only added if "hasClientBuild" is set (see https://github.com/cloudflare/workers-sdk/blob/d9d937ab6f2868271dde5a8da625773085eaec85/packages/vite-plugin-cloudflare/src/index.ts#L224), which is only set when there was a previous build for the "client" environment.

Do you think it would be possible to change the hasClientBuild check towards something that checks if there is a client environment configured and doesn't rely on the client environment being built previously? I'm happy to create a PR for this if you'd be open to changing this.

Maybe it would even make sense to offer a "buildClientLast" option for the plugin, so that this could be set in the Cloudflare Vite plugin config? Then I wouldn't even need to take over the build process myself.

jamesopstad:
Fixes #9384.

Assets that are imported in the entry Worker are now automatically moved to the client build output. Additionally, this PR adds support for a broader range of build scenarios. These are:

- Assets only build with client entry/entries
- Assets only build with no client entry/entries that includes `public` directory assets
- Worker(s) + assets build with client entry/entries
- Worker(s) + assets build with no client entry/entries that includes imported and/or `public` directory assets
- Worker(s) build with no assets

This PR switches the build order so that Workers are built first. It also changes the strategy so that the `assets` field is always populated in the output `wrangler.json` and is then removed if there are no assets.

Note: for Vite 7 we will reuse some of the logic introduced here in the new `buildApp` hook.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: to be updated once static routing is added
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/src/build.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+function loadViteManifest(directory: string) {
+	const contents = fs.readFileSync(
+		path.resolve(directory, ".vite", "manifest.json"),
+		"utf-8"
+	);
+
+	return JSON.parse(contents) as vite.Manifest;
+}
```
jamesopstad, line 164: There's been some discussion about accessing the build manifests without reading and writing files and I've opened an issue in the Vite repo for this - https://github.com/vitejs/vite/issues/20052.

packages/vite-plugin-cloudflare/src/build.ts
```diff
@@ -0,0 +1,176 @@
...
+		} else {
+			const entryWorkerConfigPath = path.join(
+				entryWorkerBuildDirectory,
+				"wrangler.json"
+			);
+			const workerConfig = JSON.parse(
+				fs.readFileSync(entryWorkerConfigPath, "utf-8")
+			) as Unstable_Config;
+			// Remove `assets` field as there are no assets
+			workerConfig.assets = undefined;
```
petebacondarwin, line 71: If I understand correctly this is only a problem because the wrangler.jsonc in the `assets` playground has the following:

```
"assets": { "binding": "ASSETS" }
```

and we need to remove that because there is no point in having a binding if there are no assets to binding to.

Is that correct?

In the real world, if there are no assets then the user would not add this field in the first place, right? If so, then should this actually be a user error?

packages/vite-plugin-cloudflare/src/build.ts
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			if (fs.existsSync(dest)) {
+				fs.unlinkSync(src);
+			} else {
+				const destDir = path.dirname(dest);
+
+				if (!fs.existsSync(destDir)) {
+					fs.mkdirSync(destDir, { recursive: true });
+				}
```
petebacondarwin: It might well just be quicker to do?

```suggestion
				fs.mkdirSync(destDir, { recursive: true });
```

packages/vite-plugin-cloudflare/src/build.ts
```diff
@@ -0,0 +1,176 @@
...
+
+				fs.renameSync(src, dest);
```
petebacondarwin, line 98: Would it be safer to just copy rather than move?
Is it possible that the file needs to be in the bundle too?

jamesopstad:
> Cool. Are there any security concerns about the imported server-side asset being moved to client-side assets which could then potentially be accessed from the browser?

Only if this doesn't meet user expectations. I did wonder if we should be blocking direct access to assets that are only imported on the server. We could potentially do this by adding the paths to the `run_worker_first` array once support for that is added. Might be a nice feature. One issue is that there is a limit on the number of rules and we would be adding to the user's rules so this might be unexpected. If we added each file to the array then it's quite possible we could hit this limit.

packages/vite-plugin-cloudflare/src/build.ts
jamesopstad: That's not quite it. I've changed the implementation in this PR so that the `assets` field is always included in the output `wrangler.json` at build time, regardless of whether there are assets or not. It is then removed here if there are no assets. This makes things a lot simpler and removes the need for the `client` environment to be built first. In Vite 7 we will also be able to add this check in the `buildApp` hook and determine whether another plugin has built the `client` environment.

packages/vite-plugin-cloudflare/src/build.ts
jamesopstad: There shouldn't be any need for these files to be in the Worker bundle and I think it's nicer to inspect a clean output directory. We also don't want these files to be inadvertently matched by `rules` and uploaded as additional modules.

packages/vite-plugin-cloudflare/src/build.ts
jamesopstad: [365c4da](https://github.com/cloudflare/workers-sdk/pull/9510/commits/365c4daecfebe68125e645c7fde229cc4b4129e9)

jamesopstad:
> > Cool. Are there any security concerns about the imported server-side asset being moved to client-side assets which could then potentially be accessed from the browser?
> 
> Only if this doesn't meet user expectations. I did wonder if we should be blocking direct access to assets that are only imported on the server. We could potentially do this by adding the paths to the `run_worker_first` array once support for that is added. Might be a nice feature. One issue is that there is a limit on the number of rules and we would be adding to the user's rules so this might be unexpected. If we added each file to the array then it's quite possible we could hit this limit.

Discussed this further with @petebacondarwin. The conclusion was to merge this as is and clearly document how to make assets imported in a Worker private. If, in the future, there is a need for specifying that individual asset imports should be private then we could add a new import attribute for this e.g. `with { cloudflare: 'asset' }`.

PR #9517, opened 2025-06-06
Add validation to containerAppConfiguration that disk.size is not set

IRCody:
disk.size field is deprecated. Add an error pointing the user to the correct field if it is set.

Before this users could get into a weird state when mixing the disk config options.

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: path is not currently tested.
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: validation change only
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: validation change only for undocumented field
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  new feature

### DIFF
--- a/.changeset/angry-bears-attack.md
+++ b/.changeset/angry-bears-attack.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add validation that the deprecated field containers.disk.size is not set.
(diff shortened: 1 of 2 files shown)


emily-shen (commented):
would you be able to mark the various deprecated config options in `config.environment.ts` with `@deprecated`?

.changeset/angry-bears-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add validation that the deprecated field containers.disk.size is not set.
```
petebacondarwin, line 5:
```suggestion
Add validation that the deprecated field containers.disk.size must not be set.
```

PR #9530, opened 2025-06-09
[wrangler] Add --json flag to r2 bucket info command

### LINKED ISSUE #9433: Add --json option to R2 commands in Wrangler
### Describe the solution

When working with D1 via Wrangler, this option is available:
`--json  Return output as clean JSON  [boolean] [default: false]`
but not when working with R2 buckets.

Currently, this command:
`wrangler r2 bucket info <bucket-name>`
will give me the desired `object_count` and `bucket_size` parameters, but not in JSON format.

And this API call:
`/accounts/{account_id}/r2/buckets/{bucket_name}`
will give me JSON format but not the desired parameters.

This is mysteriously incoherent and should be easy to fix.

I request this feature because it would greatly simplify building local systems that interact with Wrangler, and only using Wrangler as an API, rather than having to build around Wrangler and figure out when to use the Cloudflare API or the GraphQL API or the R2 AWS Api.

Cloudflare offers a lot of functionality through Wrangler, but it would be nice to be allowed to make my own custom workflow for handling many projects in different folders, which is what I'm trying to do, and why this feature would be very helpful.

For my specific use case, I want to make a simple GUI that shows me a list of R2 buckets, and then have the option to select a bucket to see its object count and size. This should be simple, but will now require complicated workarounds with either regex or GraphQL requests, while a simple `--json` option had solved it for me.

Seeing as this is already a feature for D1 commands, surely it should also be available for R2 commands?

Also, adding the `--json` parameter should make the output skip this part, and only return the JSON result:

 ⛅️ wrangler 4.18.0
───────────────────
Listing buckets...

Akshit222:
Fixes #9433

## What

Adds a `--json` flag to the `wrangler r2 bucket info` command, returning the bucket info in clean JSON format.

## Why

The current output is not machine-readable.  
Adding this flag makes the output usable in scripts and consistent with D1 commands that already support `--json`.  

## How

- Adds a `--json` flag to `r2 bucket info`
- Omits verbose output when `--json` is passed
- Suppresses the Wrangler banner when `--json` is used
- Adds a unit test verifying valid JSON output
- Linting fixed with `pnpm fix`
- Added a changeset as per contribution guidelines

## Test Plan

- ✅ `pnpm exec vitest run packages/wrangler/src/__tests__/r2/bucket.test.ts` passed
- ✅ Manual output verified
- ✅ Verified that banner is suppressed when `--json` is passed

### With `--json`:
```bash
$ wrangler r2 bucket info my-bucket-name --json

{
  "name": "my-bucket-name",
  "created": "2025-06-07T15:55:22.222Z",
  "location": "APAC",
  "default_storage_class": "Standard",
  "object_count": "0",
  "bucket_size": "0 B"
}
```

### Without `--json`:
```bash
$ wrangler r2 bucket info my-bucket-name

(⛅️ wrangler 4.18.0)
Bucket Name: my-bucket-name  
Created At: 2025-06-07T15:55:22.222Z  
Location: APAC  
Storage Class: Standard  
Object Count: 0  
Size: 0 B
```

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24120
- Wrangler V3 Backport
  - [ ] Wrangler PR:
  - [x] Not necessary because: New features are not backported

### DIFF
--- a/.changeset/happy-streets-leave.md
+++ b/.changeset/happy-streets-leave.md
@@ -1,5 +1,5 @@
 ---
-"wrangler": patch
+"wrangler": minor
 ---
 
 Add --json flag to r2 bucket info command for machine-readable output.
--- a/.changeset/happy-streets-leave.md
+++ b/.changeset/happy-streets-leave.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add --json flag to r2 bucket info command for machine-readable output.
--- a/packages/wrangler/src/__tests__/r2/bucket.test.ts
+++ b/packages/wrangler/src/__tests__/r2/bucket.test.ts
@@ -1,0 +1,54 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { getR2Bucket, getR2BucketMetrics } from "../../r2/helpers";
+import { requireAuth } from "../../user";
+import { mockConsoleMethods } from "../helpers/mock-console";
+import { runWrangler } from "../helpers/run-wrangler";
+
+vi.unmock("../../wrangler-banner");
+
+vi.mock("../../r2/helpers");
+vi.mock("../../user");
+
+const logs = mockConsoleMethods();
+
+const mockRequireAuth = vi.mocked(requireAuth);
+const mockGetR2Bucket = vi.mocked(getR2Bucket);
+const mockGetR2BucketMetrics = vi.mocked(getR2BucketMetrics);
+
+describe("r2 bucket info", () => {
+	beforeEach(() => {
+		vi.resetAllMocks();
+
+		mockRequireAuth.mockResolvedValue("test-account-id");
+		mockGetR2Bucket.mockResolvedValue({
+			name: "my-bucket-name",
+			creation_date: "2025-06-07T15:55:22.222Z",
+			location: "APAC",
+			storage_class: "Standard",
+		});
+		mockGetR2BucketMetrics.mockResolvedValue({
+			objectCount: 0,
+			totalSize: "0 B",
+		});
+	});
+
+	it("should output valid JSON format when --json flag is used", async () => {
+		await runWrangler("r2 bucket info my-bucket-name --json");
+
+		const output = logs.out;
+
+		const jsonMatch = output.match(/\{[\s\S]*\}/);
+		if (!jsonMatch) {
+			throw new Error("No JSON found in output");
+		}
+
+		try {
+			const json = JSON.parse(jsonMatch[0]);
+			expect(json.name).toBe("my-bucket-name");
+		} catch (err) {
+			throw new Error("Output contained invalid JSON");
+		}
+
+		expect(output).not.toMatch(/⛅ wrangler/);
+	});
+});
--- a/packages/wrangler/src/__tests__/r2/bucket.test.ts
+++ b/packages/wrangler/src/__tests__/r2/bucket.test.ts
@@ -16,39 +16,25 @@ const mockGetR2Bucket = vi.mocked(getR2Bucket);
 const mockGetR2BucketMetrics = vi.mocked(getR2BucketMetrics);
 
 describe("r2 bucket info", () => {
-	beforeEach(() => {
-		vi.resetAllMocks();
+  beforeEach(() => {
+    vi.resetAllMocks();
 
-		mockRequireAuth.mockResolvedValue("test-account-id");
-		mockGetR2Bucket.mockResolvedValue({
-			name: "my-bucket-name",
-			creation_date: "2025-06-07T15:55:22.222Z",
-			location: "APAC",
-			storage_class: "Standard",
-		});
-		mockGetR2BucketMetrics.mockResolvedValue({
-			objectCount: 0,
-			totalSize: "0 B",
-		});
-	});
+    mockRequireAuth.mockResolvedValue("test-account-id");
+    mockGetR2Bucket.mockResolvedValue({
+      name: "my-bucket-name",
+      creation_date: "2025-06-07T15:55:22.222Z",
+      location: "APAC",
+      storage_class: "Standard",
+    });
+    mockGetR2BucketMetrics.mockResolvedValue({
+      objectCount: 0,
+      totalSize: "0 B",
+    });
+  });
 
-	it("should output valid JSON format when --json flag is used", async () => {
-		await runWrangler("r2 bucket info my-bucket-name --json");
-
-		const output = logs.out;
-
-		const jsonMatch = output.match(/\{[\s\S]*\}/);
-		if (!jsonMatch) {
-			throw new Error("No JSON found in output");
-		}
-
-		try {
-			const json = JSON.parse(jsonMatch[0]);
-			expect(json.name).toBe("my-bucket-name");
-		} catch (err) {
-			throw new Error("Output contained invalid JSON");
-		}
-
-		expect(output).not.toMatch(/⛅ wrangler/);
-	});
+  it("should output valid JSON format when --json flag is used", async () => {
+    await runWrangler("r2 bucket info my-bucket-name --json");
+    const json = JSON.parse(logs.out);
+    expect(json.name).toBe("my-bucket-name");
+  });
 });
--- a/packages/wrangler/src/__tests__/r2/bucket.test.ts
+++ b/packages/wrangler/src/__tests__/r2/bucket.test.ts
@@ -16,25 +16,25 @@ const mockGetR2Bucket = vi.mocked(getR2Bucket);
 const mockGetR2BucketMetrics = vi.mocked(getR2BucketMetrics);
 
 describe("r2 bucket info", () => {
-  beforeEach(() => {
-    vi.resetAllMocks();
+	beforeEach(() => {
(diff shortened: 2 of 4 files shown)


frankichiro:
Hello. Since I'm who issued [#9433](https://github.com/cloudflare/workers-sdk/issues/9433), I'm very happy about this. Although it would be very preferable if it was possible to remove the banner line when using `--json`, otherwise the result needs to be filtered, which was essentially the original problem. This would also make it consistent with how `--json` works with D1.

Akshit222:
Thanks so much for the feedback, @frankichiro — totally agree with you. 
You're right that having the banner in --json mode defeats the purpose of clean, machine-readable output and makes it inconsistent with other commands like D1.

I'll go ahead and update the PR to suppress the banner when --json is used, so the output is clean and parseable.

Akshit222:
Made the requested changes to suppress the banner when --json is used. Let me know if there's anything else I should address — happy to iterate further. Thanks again for the guidance!

Akshit222:
Just wanted to follow up on this! Let me know if there’s anything else needed from my side to move this forward — happy to make any changes if required. Thanks again for your time!

packages/wrangler/src/r2/bucket.ts
```diff
@@ -190,56 +190,69 @@ export const r2BucketListCommand = createCommand({
...
+      alias: "J",
+      requiresArg: true,
+      type: "string",
+    },
+    json: {
+      describe: "Return the bucket information as JSON",
+      type: "boolean",
+      default: false,
+    },
+  },
```
emily-shen: you can hide the banner like this: https://github.com/cloudflare/workers-sdk/blob/395f36de127c6ee5fbc0ceadbfb508f7f32f5388/packages/wrangler/src/d1/list.ts#L14 instead of directly modifying wrangler-banner.ts

packages/wrangler/src/__tests__/r2/bucket.test.ts
```diff
@@ -0,0 +1,56 @@
+import { vi, describe, it, expect, beforeEach } from "vitest";
+import { runWrangler } from "../helpers/run-wrangler";
+import { requireAuth } from "../../user";
+import { getR2Bucket, getR2BucketMetrics } from "../../r2/helpers";
+
+// Mock the dependencies
```
emily-shen: can you add `vi.unmock("../wrangler-banner");` to test the banner isn't included? we mock it out by default in tests

packages/wrangler/src/__tests__/r2/bucket.test.ts
```diff
@@ -0,0 +1,56 @@
...
+    mockGetR2BucketMetrics.mockResolvedValue({
+      objectCount: 0,
+      totalSize: "0 B"
+    });
+  });
+
+  it("should output valid JSON format when --json flag is used", async () => {
+    let capturedOutput = "";
+
+    const originalConsoleLog = console.log;
```
emily-shen: you should use the `mockConsoleMethods()` helper here

emily-shen:
also, it looks like you have some linting issues. you can run `pnpm fix` to fix them up

packages/wrangler/src/r2/bucket.ts
Akshit222: Thanks for the feedback , I’ve now reverted teh changes to wrangler-banner.ts. Instead, I’m using the behaviour: { showBanner: false } option inside the relevant CLI command for the --json path, similar to how it's done in d1/list.ts.

packages/wrangler/src/__tests__/r2/bucket.test.ts
Akshit222: I've added vi.unmock("../wrangler-banner") to ensure we’re explicitly testing that the banner isn’t printed when using --json.

packages/wrangler/src/__tests__/r2/bucket.test.ts
Akshit222: Updated! I’m now using mockConsoleMethods() in the test instead of manually mocking console.log

Akshit222:
Thanks a lot for the detailed review,
I’ve addressed all feedback:

-Used behaviour: { showBanner: false } instead of editing the banner file
-Added vi.unmock("../wrangler-banner") to ensure the banner isn’t shown during --json output
-Switched to mockConsoleMethods() in tests
-Ran pnpm fix
-Added a changeset

Let me know if i made a mistake anywhere.

Akshit222:
Just a quick note: I ended up cleaning up the commit history because the original one had some unrelated files accidentally committed. To keep things clean and focused, I created a fresh branch from main, reapplied only the relevant changes, and force-pushed — which unfortunately removed previous commit-level context. But I’ve addressed all the feedback again in this clean version.

Once you’ve had a chance to review this updated commit, I’ll go ahead and create a follow-up PR to update the Wrangler documentation for the r2 bucket info command as well.

Let me know if anything else is needed!

penalosa:
@Akshit222 thanks for this, this is looking pretty good! Would you be able to restore the PR template and fill out the relevant checkboxes, so we can get this merged?

Akshit222:
Thanks for the feedback! I'll update the PR template and checkboxes shortly.

Akshit222:
I've updated the PR description with the template and filled out the checkboxes as requested.
Please let me know if there's anything else you'd like me to address!

Also, please let me know if any of the failing checks need action on my side — happy to fix if required.

packages/wrangler/src/__tests__/r2/bucket.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+const logs = mockConsoleMethods();
+
+const mockRequireAuth = vi.mocked(requireAuth);
+const mockGetR2Bucket = vi.mocked(getR2Bucket);
+const mockGetR2BucketMetrics = vi.mocked(getR2BucketMetrics);
+
+describe("r2 bucket info", () => {
+	beforeEach(() => {
+		vi.resetAllMocks();
```
petebacondarwin, line 20: I believe we do this by default already?

packages/wrangler/src/__tests__/r2/bucket.test.ts
```diff
@@ -0,0 +1,54 @@
...
+		if (!jsonMatch) {
+			throw new Error("No JSON found in output");
+		}
+
+		try {
+			const json = JSON.parse(jsonMatch[0]);
+			expect(json.name).toBe("my-bucket-name");
+		} catch (err) {
+			throw new Error("Output contained invalid JSON");
+		}
```
petebacondarwin: This will catch the exception from the expect statement if it failed, which is probably not what we want.

IMO I would just simplify this test:

```
it("should output valid JSON format when --json flag is used", async () => {
	await runWrangler("r2 bucket info my-bucket-name --json");
	const json = JSON.parse(logs.out);
	expect(json.name).toBe("my-bucket-name");
});

**Comment by petebacondarwin** [comment]
there are a bunch of linting errors which would mostly be solved by simplifying the test as suggested.

File: packages/wrangler/src/r2/bucket.ts [hunk 6]
```diff
@@ -236,7 +248,11 @@ export const r2BucketInfoCommand = createCommand({
 			bucket_size: bucketMetrics.totalSize,
 		};
 
-		logger.log(formatLabelledValues(output));
+		if (args.json) {
+			console.log(JSON.stringify(output, null, 2));
```
petebacondarwin:
```suggestion
			logger.json(output);
```

petebacondarwin:
Please open a draft docs-PR before merging and reference it in the PR description. It can then be reviewed and merged after this had been released.

frankichiro:
Wow, this really required much more work than I imagined it would. Thanks @Akshit222 for seeing it through, I really appreciate your efforts for making this work.

Akshit222:
Thanks a lot for the thoughtful feedback and guidance 
I will make the rest of mentioned changes as asked very soon

Akshit222:
Thanks for the detailed review! I've made the following changes based on your feedback:

Snapshot concerns: I'm not using Jest snapshots in these tests, so there's no need to update any. The existing test setup avoids snapshot-based output assertions — but let me know if I misunderstood and you were referring to something else.

Mock reset: You were right — we already call vi.resetAllMocks() by default, so that part was redundant. I've cleaned it up.

JSON parse block simplification: Simplified the --json test as suggested. It now directly asserts on the parsed output rather than wrapping the parse in a try-catch.

Linting issues: These were mostly resolved by the above simplification and cleanup — no lint errors now. I checked after running pnpm check:lint.

logger.json() integration: Updated the bucket.ts implementation to use logger.json() for consistency with other commands when the --json flag is used.

I'll send a follow-up PR to add the relevant docs to docs/content/en/r2 as discussed soon.

Let me know if there’s anything else to tweak!

.changeset/happy-streets-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
```
petebacondarwin:
```suggestion
"wrangler": minor
```

This is a new minor feature not a bug fix.

PR #9532, opened 2025-06-09
chore: add vitest-ecosystem-ci script

edmundhung:
Fixes n/a.

We renamed the fixtures in https://github.com/cloudflare/workers-sdk/pull/9378 recently which broke our vitest-ecosystem-ci setup: https://github.com/vitest-dev/vitest-ecosystem-ci/actions/runs/15384430303/job/43280381332#step:7:886

This adds a dedicate command in the root package json and I will submit a PR to update it in the vitest-ecosystem-ci repository.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: CI change 
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: CI change 
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: CI change 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: CI change

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,8 @@
 		"test:e2e": "dotenv -- turbo test:e2e",
 		"test:e2e:wrangler": "node -r esbuild-register tools/e2e/runIndividualE2EFiles.ts",
 		"test:watch": "turbo test:watch",
-		"type:tests": "dotenv -- turbo type:tests"
+		"type:tests": "dotenv -- turbo type:tests",
+		"vitest-ecosystem-ci": "pnpm test:ci -F @fixture/vitest-pool-workers"
 	},
 	"dependencies": {
 		"@actions/artifact": "^2.2.1",


dario-piotrowicz:
why aren't we just making this update in the vitest-ecosystem-ci? 🤔 are you expecting the directory name to change again at some point?

(you do need to open a new PR in the vitest ecosystem repo, so you might as well just update the directory there instead, no?)

edmundhung:
> why aren't we just making this update in the vitest-ecosystem-ci? 🤔 are you expecting the directory name to change again at some point?
> 
> (you do need to open a new PR in the vitest ecosystem repo, so you might as well just update the directory there instead, no?)

We can absolutely do that. Just thinking having a dedicated script will give us some flexibility in the future in case we update our tests setup or there are additional setup required.

dario-piotrowicz:
> We can absolutely do that. Just thinking having a dedicated script will give us some flexibility in the future in case we update our tests setup or there are additional setup required.

Personally I feel like it's a bit unnecessary, it feels weird to me to include a script that its specifically included for a separate repository to consume 🤔 

One could make the argument that the script can be removed or renamed in the future (by someone not fully knowing what it is there for) 😅 

Given the low probability of things changing on our side, the fact that we'll see failures in the vitest ecosystem right away and that upstreaming the fixes would also be quite fast again I think I would just upstream this change in the vitest repo and not add a script here.

If you or other folks disagree I am totally ok having the script here though 👍

edmundhung:
That's fair. Let's re-consider this only if it happens again 👍🏼 Thanks for the input @dario-piotrowicz

PR #9533, opened 2025-06-09
refactor: run C3 e2e tests more easily locally and individually

petebacondarwin:
- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: c3
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: test refactor
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  c3

### DIFF
--- a/.github/workflows/c3-e2e.yml
+++ b/.github/workflows/c3-e2e.yml
@@ -14,7 +14,7 @@ jobs:
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os.name }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}${{ matrix.experimental && '-experimental' || '' }}
       cancel-in-progress: true
-    name: ${{ format('C3 ({0}, {1}{2})', matrix.os.description, matrix.pm.name, matrix.experimental && ', experimental' || '') }}
+    name: ${{ format('C3 ({0}, {1}{2})', matrix.pm.name, matrix.os.description, matrix.experimental && ', experimental' || '') }}
     strategy:
       fail-fast: false
       matrix:


.github/workflows/c3-e2e.yml
```diff
@@ -9,16 +9,16 @@ env:
 
 jobs:
   e2e:
-    # Note: please keep this job in sync with the e2e-only-dependabot-bumped-framework one
-    #       in .github/workflows/c3-e2e-dependabot.yml
+    # Runs the non-frameworks C3 E2E tests on all supported operating systems and package managers.
     timeout-minutes: 45
     concurrency:
-      group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}
+      group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}${{ matrix.experimental && '-experimental' || '' }}
```
penalosa: Do we have any experimental frameworks left?

.github/workflows/c3-e2e.yml
```diff
@@ -9,16 +9,16 @@ env:
...
       cancel-in-progress: true
-    name: ${{ format('C3 ({0})', matrix.description) }}
+    name: ${{ format('C3 ({0}{1})', matrix.description, matrix.experimental && ', experimental' || '') }}
     strategy:
       fail-fast: false
       matrix:
+        experimental: [true, false]
```
penalosa, line 21: I don't think this will work given the include list below which lists out all the combinations?

.github/workflows/c3-e2e.yml
```diff
@@ -60,26 +61,41 @@ jobs:
...
+        uses: oven-sh/setup-bun@v1
         with:
-          node-version: ${{ env.NODE_VERSION }}
-          packageManager: ${{ matrix.pm.name }}
-          packageManagerVersion: ${{ matrix.pm.version }}
-          quarantine: false
-          experimental: true
-          accountId: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
-          apiToken: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
+          bun-version: ${{ matrix.pm.nameVersion }}
```
penalosa: We no longer test on bun, so we should remove this

.github/workflows/c3-e2e.yml
```diff
@@ -60,26 +61,41 @@ jobs:
...
+
+      # Needed because gatsby requires git
+      - name: Configure Git
+        shell: bash
+        run: |
+          git config --global user.email wrangler@cloudflare.com
+          git config --global user.name 'Wrangler automated PR updater'
```
penalosa: nit, but maybe a different name here?

.github/workflows/c3-e2e.yml
```diff
@@ -60,26 +61,41 @@ jobs:
...
-          accountId: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
-          apiToken: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
+          name: e2e-logs${{matrix.experimental == 'true' && '-experimental' || ''}}-${{runner.os}}-${{matrix.pm.name}}
+          path: packages/create-cloudflare/.e2e-logs${{matrix.experimental == 'true' && '-experimental' || ''}}
+          include-hidden-files: true
+
+      - name: Fail if errors detected
+        shell: bash
+        if: ${{ steps.run-e2e.outcome == 'failure' }}
+        run: exit 1
```
penalosa: What's this for?

.github/workflows/c3-e2e.yml
```diff
@@ -60,26 +61,41 @@ jobs:
...
-      - name: E2E Tests (non-experimental)
+      - id: run-e2e
         if: steps.changes.outputs.everything_but_markdown == 'true'
-        uses: ./.github/actions/run-c3-e2e
+        shell: bash
+        run: pnpm run test:e2e:c3
+        env:
+          NODE_VERSION: ${{ env.NODE_VERSION }}
+          CLOUDFLARE_API_TOKEN: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
```
penalosa, line 81: Are these actually needed anymore, now we just run dev in E2Es?

packages/create-cloudflare/e2e/helpers/to-exist.ts
penalosa, line 1: It looks like this has been changed from a `setupFile` to something that needs manually imported in each test file? Will that not be easier to forget?

packages/create-cloudflare/e2e/tests/workers/test-config.ts
```diff
@@ -0,0 +1,198 @@
...
+				: [{ ...testConfig, name: testConfig.name ?? testConfig.template }],
+		)
+		.filter((testConfig) => {
+			if (!workerToTestFilter) {
+				return true;
+			}
+			if (workerToTestFilter.includes(":")) {
+				return testConfig.name === workerToTestFilter;
+			}
+			return testConfig.name.split(":")[0] === workerToTestFilter;
```
penalosa: Given this logic, workerToTest*Filter* seems like a bit of a misnomer to me. From the name and README I expected some sort of prefix matching, rather than exact name matching

penalosa (approved):

.github/workflows/c3-e2e.yml
petebacondarwin: I think you are right...

.github/workflows/c3-e2e.yml
petebacondarwin: sure - this was copied over from the original action

.github/workflows/c3-e2e.yml
petebacondarwin: This was copied over from the original action. As I understand it, since we have the "Upload Logs" step set to `if: always()` and that always passes, then the job itself will always pass.

Having this extra step overrides that.

.github/workflows/c3-e2e.yml
petebacondarwin: Possibly! I'll try removing

petebacondarwin: Actually, the CLI tests include an `--existing-script` use case that does need these.
They get skipped if the env var is missing.
But I think we should not skip this in CI.

packages/create-cloudflare/e2e/helpers/to-exist.ts
petebacondarwin: Actually I did this on purpose, but I'm willing to discuss.
My view is that stuff in setup files is hard to discover and magical.
My aim was that a simple import at the top of the file ensures that the runtime (and typing!) of a helper is available.

But I realise that once the import is made anywhere in the code, then the typings are available everywhere, even if the runtime is not 😢 

So I'll change this back to being loaded in a setup config.

packages/create-cloudflare/e2e/tests/workers/test-config.ts
petebacondarwin: It is slightly nuanced... if the filter contains `:` then it is an exact match (including the variant).
If not then it will match all variants.

But I see your point. I'll rename.

.github/workflows/c3-e2e.yml
penalosa: We include an `always()` step in the Wrangler E2E tests, and that doesn't seem to make the job always pass, but I'm not super familiar with the setup here

.github/workflows/c3-e2e.yml
penalosa: Sounds good

packages/create-cloudflare/e2e/helpers/to-exist.ts
petebacondarwin: 604795537f5cf28f8615d7335ab4f2ecb780fd62

.github/workflows/c3-e2e.yml
petebacondarwin: 20071d816970a037e3f7972b743084c556723b61

.github/workflows/c3-e2e.yml
petebacondarwin: 20071d816970a037e3f7972b743084c556723b61

packages/create-cloudflare/e2e/tests/workers/test-config.ts
petebacondarwin: b90255e9cccd18f4ced598ff332bcb9e2194a816

.github/workflows/c3-e2e.yml
petebacondarwin: Reading the docs, it looks like you are right. Removing.

.github/workflows/c3-e2e.yml
petebacondarwin: took a different approach using excludes

.github/workflows/c3-e2e.yml
petebacondarwin: 4b62f18e8550066b654ce2a9515395b9795332ae

.github/workflows/c3-e2e.yml
petebacondarwin: Not now that I have moved solid over. But we should still keep these tests for future experimental stuff.

Issue #7870, opened 2025-01-22
How to use breakpoint debugging when testing with Miniflare?


irvinebroque:
### Which Cloudflare product(s) does this pertain to?

Miniflare

### What versions are you using?

latest

### What operating system and version are you using?

latest

### Please provide a link to a minimal reproduction

https://github.com/mjackson/miniflare-test

### Describe the Bug

I'm trying to get breakpoint debugging working with [Miniflare](https://developers.cloudflare.com/workers/testing/miniflare/). The [Cloudflare docs](https://developers.cloudflare.com/workers/testing/miniflare/developing/debugger/) seem to suggest it's possible.

Ideally, what I'd like to be able to do is:

1. Open `worker.js` and add a breakpoint inside the `fetch()` handler
2. Open a "JavaScript Debug Terminal" in VSCode
3. Run `node miniflare.js`
4. Pause at the breakpoint in the worker code

However, this doesn't seem to work right now :/

A nice bonus would be to be able to write everything in TypeScript, but I was trying to keep this example as minimal as possible.

penalosa:
> A nice bonus would be to be able to write everything in TypeScript

Miniflare (intentionally) tries to expose an API layer that's as close to the platform as possible—essentially the local equivalent to the CF API. Because the CF API doesn't support uploading TypeScript workers, Miniflare doesn't. The `startWorker()` API from Wrangler is a wrapper on top of Miniflare that should support Typescript, NPM packages, and all the bundling niceties that Wrangler provides.

In terms of the ask here—breakpoint debugging with Miniflare—that's totally possible. You should be able to follow the examples in the docs to connect to the inspector server that Miniflare starts up automatically.

However, it looks like you're trying to get this working with JavaScript Debug terminals in VSCode? That specifically doesn't work, because https://github.com/microsoft/vscode-js-debug (the library that VSCode uses for this functionality) is pretty Node specific, and e.g. hooks `child_process` and injects loader hooks. I _think_ there's a way we can collaborate with them to make this work for Workers (@mrbbot had some ideas around this, and there's been some internal discussion in the Vite chat with @dario-piotrowicz and @petebacondarwin)—they've already made some Deno-specific changes, so I imagine they'd be open to it. I think we probably need to spend some time investigating what the specific (technical) requirements are here, and then start a dialog with the VSCode team to collaborate on a solution.

PR #9535, opened 2025-06-09
Initial support for VSCode Javascript Debug Terminals

penalosa:
Fixes https://github.com/cloudflare/workers-sdk/issues/7870

Support for [Javascript Debug Terminals](https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_javascript-debug-terminal) in VSCode! This is an improvment to debugging DX when using VSCode—especially with multiple Workers. All you need to do is open a debug terminal, run `wrangler dev`, and the inspector will be connected—even when using using Wrangler's multiworker mode (which means you can set breakpoints across service bindings).

Caveats: this still doesn't work in the Vite plugin, and relies heavily on VSCode internals

See also https://github.com/microsoft/vscode-js-debug/issues/2233

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: this should be manually tested in VSCode
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: this should be manually tested in VSCode
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/issues/23352
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a patch change


packages/miniflare/src/runtime/index.ts
```diff
@@ -156,7 +180,44 @@ export class Runtime {
...
+				const p = spawn(process.execPath, [watchdogPath], {
+					env: {
+						NODE_INSPECTOR_INFO: JSON.stringify({
+							ipcAddress: info.inspectorIpc || "",
+							pid: String(this.#process.pid),
+							scriptName: worker,
+							inspectorURL: `ws://127.0.0.1:${ports?.get(kInspectorSocket)}/core:user:${worker}`,
+							waitForDebugger: true,
+							ownId: randomBytes(12).toString("hex"),
+							openerId: info.openerId,
```
connor4312, line 216: If cloudflare workers support spawning subprocesses and you want debugging to work there, you would want to update the inspector options passed to subprocesses (i.e. in the current process' environment) such that the `openerId` in the options that children receive is the `ownId` of the process you just spawned.

penalosa: Cloudflare workers don't support spawning additional subprocesses (just the list of Workers that are passed into this function). In the context of this function, my read was that `openerId` would be the ID of the Miniflare process which orchestrates these workers, and so all the workers would show up as children of that Node process (which seems to be what happened when testing)

fixtures/worker-app/src/index.js
```diff
@@ -20,7 +20,7 @@ function hexEncode(array) {
 
 export default {
 	async fetch(request, env) {
-		console.log("request log");
+		console.log("request log 5");
```
vicb: revert?

fixtures/worker-ts/src/index.ts
```diff
@@ -28,7 +28,9 @@ export default {
 		ctx: ExecutionContext
 	): Promise<Response> {
 		const url = new URL(request.url);
+		const a = 29;
+		console.log(url.hash);
```
vicb: revert?

packages/miniflare/src/runtime/index.ts
```diff
@@ -119,13 +121,35 @@ function getRuntimeArgs(options: RuntimeOptions) {
 	return args;
 }
 
+function getInspectorOptions() {
```
vicb, line 129: Add JSDoc?

packages/miniflare/src/runtime/index.ts
```diff
@@ -119,13 +121,35 @@ function getRuntimeArgs(options: RuntimeOptions) {
...
+	} catch {
+		return undefined;
+	}
+}
+
 export class Runtime {
 	#process?: childProcess.ChildProcess;
 	#processExitPromise?: Promise<void>;
 
 	async updateConfig(
```
vicb, line 154: Add JSDoc?

packages/miniflare/src/runtime/index.ts
```diff
@@ -156,7 +180,44 @@ export class Runtime {
...
+
+			if (!bootloaderPath) {
+				return ports;
+			}
+			const watchdogPath = path.resolve(bootloaderPath, "../watchdog.js");
+
+			const info = getInspectorOptions();
+
+			for (const worker of workers ?? []) {
+				const p = spawn(process.execPath, [watchdogPath], {
```
vicb, line 207: Add a comment on what the code is doing here?

packages/wrangler/src/api/startDevWorker/ProxyController.ts
```diff
@@ -74,86 +79,89 @@ export class ProxyController extends Controller<ProxyControllerEventMap> {
...
-							return new Response(null, { status: 204 });
-						},
-					},
-					bindings: {
-						PROXY_CONTROLLER_AUTH_SECRET: this.secret,
-					},
+		const inspectorProxyWorkerOptions: WorkerOptions = {
+			name: "InspectorProxyWorker",
+			compatibilityDate: "2023-12-18",
+			compatibilityFlags: ["nodejs_compat", "increase_websocket_message_size"],
```
vicb: Can't seem to find `increase_websocket_message_size` at https://developers.cloudflare.com/workers/configuration/compatibility-flags/

Also should we use a more recent compat data?

packages/wrangler/src/api/startDevWorker/ProxyController.ts
```diff
@@ -74,86 +79,89 @@ export class ProxyController extends Controller<ProxyControllerEventMap> {
...
+					PROXY_CONTROLLER_AUTH_SECRET: this.secret,
+				},
+
+				// no need to use file-system, so don't
+				cache: false,
+				unsafeEphemeralDurableObjects: true,
+			},
+		];
+		if (!inVscodeJsDebugTerminal) {
+			workers.push(inspectorProxyWorkerOptions);
```
vicb: Do you need to init `inspectorProxyWorkerOptions` outside of this if?

packages/wrangler/src/api/startDevWorker/ProxyController.ts
vicb: Looks like it's a dev flag only, maybe add a link to https://github.com/cloudflare/workerd/blob/79ee1f0/src/workerd/io/compatibility-date.capnp#L539

packages/miniflare/src/index.ts
```diff
@@ -1989,7 +1989,10 @@ export class Miniflare {
 		};
 		const maybeSocketPorts = await this.#runtime.updateConfig(
 			configBuffer,
-			runtimeOpts
+			runtimeOpts,
+			this.#workerOpts
+				.filter((w) => w.core.name)
+				.map((w) => w.core.name) as string[]
```
petebacondarwin: NIT:

```suggestion
			this.#workerOpts.flatMap((w) => w.core.name ?? [])
```

packages/miniflare/src/runtime/index.ts
```diff
@@ -119,13 +121,40 @@ function getRuntimeArgs(options: RuntimeOptions) {
...
+
 export class Runtime {
 	#process?: childProcess.ChildProcess;
 	#processExitPromise?: Promise<void>;
 
 	async updateConfig(
 		configBuffer: Buffer,
-		options: Abortable & RuntimeOptions
+		options: Abortable & RuntimeOptions,
+		workers: string[] | undefined
```
petebacondarwin: Can this just be:

```suggestion
		workers: string[]
```

petebacondarwin: What exactly do we mean by `workers` here?
I think it is "the names of all the Workers being hosted in this Miniflare instance"?
Right?

packages/miniflare/src/runtime/index.ts
```diff
@@ -156,7 +185,46 @@ export class Runtime {
...
+			if (!bootloaderPath) {
+				return ports;
+			}
+			const watchdogPath = path.resolve(bootloaderPath, "../watchdog.js");
+
+			const info = getInspectorOptions();
+
+			for (const worker of workers ?? []) {
+				// This is copied from https://github.com/microsoft/vscode-js-debug/blob/0b5e0dade997b3c702a98e1f58989afcb30612d6/src/targets/node/bootloader.ts#L284
+				// It spawns a detached "watchdog" process for each corresponding (user) Worker in workerd which will maintain the VSCode debug connection
```
petebacondarwin, line 206: If this is detached, what kills it? Will it die when the MIniflare process dies? I don't think so.
Do we need to add something to the miniflare dispose code?

penalosa:
> This is awesome and very exciting! What happens if the user hits `d` in Wrangler to open the dev tools when we have not created an inspector proxy?

They get a logged message saying the inspector is not available. Maybe that message should be clearer about why though

packages/miniflare/src/runtime/index.ts
penalosa: I don't think so, I _believe_ this is killed by VSCode? I'll do some poking around to clarify

packages/miniflare/src/runtime/index.ts
penalosa: It's the names of all user Workers in this Miniflare instance—I'll clarify

packages/miniflare/src/runtime/index.ts
penalosa: It kills itself after the inspector connection dies: https://github.com/microsoft/vscode-js-debug/blob/0b5e0dade997b3c702a98e1f58989afcb30612d6/src/targets/node/watchdog.ts#L35

packages/miniflare/src/index.ts
penalosa: c655618d4

packages/miniflare/src/runtime/index.ts
penalosa: c655618d4

petebacondarwin (approved):

PR #9541, opened 2025-06-10
make workers created with `startWorker` await the `ready` promise on `dispose`

dario-piotrowicz:
The changes in this PR make sure that when calling `worker.dispose();` dispose automatically awaits for the `worker.ready` promise instead of requiring the caller to do that.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included (the updated tests would fail without the code changes)
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: the [`startWorker` documentation](https://developers.cloudflare.com/workers/wrangler/api/#unstable_startworker) doesn't go into details about this sort of information
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: changes to an experimental feature

### DIFF
--- a/packages/wrangler/src/api/startDevWorker/DevEnv.ts
+++ b/packages/wrangler/src/api/startDevWorker/DevEnv.ts
@@ -23,6 +23,8 @@ export class DevEnv extends EventEmitter {
 		try {
 			await this.config.set(options, true);
 		} catch (e) {
+			// Note: we need to resolve the ready promise since worker.dispose awaits it
+			this.proxy.ready.resolve(null);
 			await worker.dispose();
 			throw e;
 		}
@@ -146,10 +148,16 @@ function createWorkerObject(devEnv: DevEnv): Worker {
 			return devEnv.proxy.ready.promise.then(() => undefined);
 		},
 		get url() {
-			return devEnv.proxy.ready.promise.then((ev) => ev.url);
+			return devEnv.proxy.ready.promise.then((ev) => {
+				assert(ev);
+				return ev.url;
+			});
 		},
 		get inspectorUrl() {
-			return devEnv.proxy.ready.promise.then((ev) => ev.inspectorUrl);
+			return devEnv.proxy.ready.promise.then((ev) => {
+				assert(ev);
+				return ev.inspectorUrl;
+			});
 		},
 		get config() {
 			assert(devEnv.config.latestConfig);
@@ -162,7 +170,9 @@ function createWorkerObject(devEnv: DevEnv): Worker {
 			return devEnv.config.patch(config);
 		},
 		async fetch(...args) {
-			const { proxyWorker } = await devEnv.proxy.ready.promise;
+			const ev = await devEnv.proxy.ready.promise;
+			assert(ev);
+			const { proxyWorker } = ev;
 			await devEnv.proxy.runtimeMessageMutex.drained();
 
 			return proxyWorker.dispatchFetch(...args);
@@ -172,7 +182,9 @@ function createWorkerObject(devEnv: DevEnv): Worker {
 				this.config.name,
 				"Worker name must be defined to use `Worker.queue()`"
 			);
-			const { proxyWorker } = await devEnv.proxy.ready.promise;
+			const ev = await devEnv.proxy.ready.promise;
+			assert(ev);
+			const { proxyWorker } = ev;
 			const w = await proxyWorker.getWorker(this.config.name);
 			return w.queue(...args);
 		},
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -146,10 +148,16 @@ function createWorkerObject(devEnv: DevEnv): Worker {
 			return devEnv.proxy.ready.promise.then(() => undefined);
 		},
 		get url() {
-			return devEnv.proxy.ready.promise.then((ev) => ev.url);
+			return devEnv.proxy.ready.promise.then((ev) => {
+				assert(ev);
```
petebacondarwin: It is good practice to add a description parameter to `assert()` calls so that we get a helpful message if the assertion fails.

petebacondarwin:
```suggestion
				assert(ev, "Expected the ready promise to resolve to a defined value but got " + ev);
```

packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -162,7 +170,9 @@ function createWorkerObject(devEnv: DevEnv): Worker {
 			return devEnv.config.patch(config);
 		},
 		async fetch(...args) {
-			const { proxyWorker } = await devEnv.proxy.ready.promise;
+			const ev = await devEnv.proxy.ready.promise;
+			assert(ev);
+			const { proxyWorker } = ev;
 			await devEnv.proxy.runtimeMessageMutex.drained();
 
 			return proxyWorker.dispatchFetch(...args);
```
petebacondarwin: NIT:
```suggestion
			await devEnv.proxy.runtimeMessageMutex.drained();
			return ev.proxyWorker.dispatchFetch(...args);
```

packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -181,11 +193,14 @@ function createWorkerObject(devEnv: DevEnv): Worker {
...
-			const { proxyWorker } = await devEnv.proxy.ready.promise;
+			const ev = await devEnv.proxy.ready.promise;
+			assert(ev);
+			const { proxyWorker } = ev;
 			const w = await proxyWorker.getWorker(this.config.name);
 			return w.scheduled(...args);
 		},
 		async dispose() {
+			await devEnv.proxy.ready.promise;
 			await devEnv.teardown();
```
petebacondarwin: What about just waiting for the `ready` promise to be settled (either resolved or rejected)?
As this stands, if the `ready` promise rejects this will throw an error and the `tearDown()` below will never get called.

```suggestion
			await devEnv.proxy.ready.promise.finally(devEnv.teardown);
```

If we do this then I think you don't need to change the type of this promise to include `null`.
Instead you can just reject the promise instead of resolving to `null`.

petebacondarwin (changes requested):

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: Good call 🙂

I've applied the change, please have another look 🙏

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: Code removed 🙂

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: I agree, but I was following the convention in these files where I can see all `assert` being called without messages (and I am generally a big fan of consistency) 🤔 

Anyways the code here has been removed so I'll just resolve the conversation 🙂

packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -23,6 +23,15 @@ export class DevEnv extends EventEmitter {
...
 			await this.config.set(options, true);
 		} catch (e) {
+			const error =
+				e instanceof Error
+					? e
+					: new Error(
+							typeof e === "string"
+								? e
+								: "An unknown error occurred when starting the worker"
+						);
```
petebacondarwin: Perhaps this could just be:

```suggestion
const error = new Error("An error occurred when starting the server", { cause: e });
```

petebacondarwin (approved):

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: sure 🙂👍

PR #9558, opened 2025-06-11
`wrangler containers apply` uses `observability` configuration

ichernetsky-cf:
Enable Observability logging for containers if it is enabled using the top-level field in a Wrangler configuration file.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: done independently and not yet publicly.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: containers are not to be supported with V3, it is still a non-public feature.


packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -802,4 +843,368 @@ describe("cloudchamber apply", () => {
...
+			"
+		`);
+		expect(std.stderr).toMatchInlineSnapshot(`""`);
+		const app = await applicationReqBodyPromise;
+		expect(app.constraints?.tier).toEqual(1);
+		expect(app.instances).toEqual(1);
+		/* eslint-enable */
+	});
+
+	test("can disable observability logging (top-level field)", async () => {
```
mikenomitch: Is this test showing the right behavior? - We set it to false in wrangler config and then it shows up as true in `deploy` output?

I may be misreading

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -802,4 +843,368 @@ describe("cloudchamber apply", () => {
...
+			"
+		`);
+		expect(std.stderr).toMatchInlineSnapshot(`""`);
+		const app = await applicationReqBodyPromise;
+		expect(app.constraints?.tier).toEqual(1);
+		expect(app.instances).toEqual(1);
+		/* eslint-enable */
+	});
+
+	test("can disable observability logging (logs field)", async () => {
```
mikenomitch: ditto

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
ichernetsky-cf: The lines in question are prefixed with a minus, meaning that the fields are getting removed, and thus "observability" should get disabled. However my assumption here might be wrong. I assumed that since we calculated diffs here, we sent new full app definitions. Setting this field explicitly to false is safer in case of modifications

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -108,12 +109,46 @@ function applicationToCreateApplication(
 	return app;
 }
 
+function observabilityToConfiguration(
```
IRCody, line 126: Should this be done in validation of the config here: https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/config/validation.ts#L2377?

The only part I'm unsure about is using the previous apps definition. Should we default to false in the config and then overwrite it with the current config values or is there some reason we need to know the previous apps config here?

packages/containers-shared/src/client/models/AccountDefaults.ts
```diff
@@ -11,7 +11,7 @@ export type AccountDefaults = {
 	vcpus: number;
 	memory_mib?: number;
 	/**
-	 * DEPRECATED in favor of memory_mib
+	 * DEPRECATED
```
vicb: Is this intended?
If it is the line should probably be remove entirely as it has no added value over l15

packages/containers-shared/src/client/models/AddressAssignment.ts
```diff
@@ -0,0 +1,16 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
```
vicb: Can this be removed?

packages/containers-shared/src/client/models/ApplicationConstraintPop.ts
```diff
@@ -0,0 +1,8 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
```
vicb, line 3: remove?

packages/containers-shared/src/client/models/ApplicationHealthInstances.ts
```diff
@@ -3,8 +3,30 @@
 /* eslint-disable */
 
 export type ApplicationHealthInstances = {
+	/**
+	 * Number of active durable object containers in this application.
+	 *
```
vicb, line 8: nit: remove this empty line (same below)

packages/containers-shared/src/client/models/EventName.ts
```diff
@@ -16,8 +16,9 @@
  * - RuntimeStartFailed: It's sent when the runtime had an internal error.
  * - SSHStarted: It's sent when the container has gained network connectivity and has opened the SSH port. This event is only sent when SSH keys are configured.
  * - CheckUpdate: Sent when the status of a health or readiness check changes. This may also affect the health status of the placement.
- * - DurableObjectConnected: Sent when a durable object instance connects and gains control of the deployment. THis event is only sent for durable object deployments.
- * It is sent after VMStarted.
+ * - DurableObjectConnected: Sent when a durable object instance connects and gains control of the deployment.
```
vicb, line 19: suggest: move the documentation to each enum member

packages/containers-shared/src/client/models/InstanceType.ts
```diff
@@ -0,0 +1,19 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
```
vicb, line 3: remove?

packages/containers-shared/src/client/models/InstanceType.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * The instance type will be used to configure vcpu, memory, and disk.
+ *
+ * - "dev": This will use a configuration of 1/16 vCPU, 256 MiB memory, and 2 GB disk
```
vicb, line 8: suggest: move the doc to each enum number instead

packages/containers-shared/src/client/models/PortRange.ts
```diff
@@ -0,0 +1,14 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
```
vicb, line 3: remove? (last time I comment this in this review but I think it should be removed everywhere)

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -108,12 +109,46 @@ function applicationToCreateApplication(
...
+		(observability?.enabled === false && observability?.logs?.enabled !== true);
+	if (observabilityLogsDisabled) {
+		return { logs: { enabled: false } };
+	}
+
+	if (existingObservabilityConfig !== undefined) {
+		return { logs: { enabled: false } };
+	}
+
+	return undefined;
```
emily-shen:
```suggestion
    const observabilityEnabled = (observability?.enabled || (observability?.logs?.enabled))
    if (observabilityEnabled !== undefined){
       return { logs: { enabled: observabilityEnabled } };
    }
	if (existingObservabilityConfig !== undefined) {
		return { logs: { enabled: false } };
	}
	return undefined;
```

but also I think I'm missing a bit of context on `existingObservabilityConfig` - could existingObservabilityConfig be true? why are we forcing it to false? 
is the idea if observability was set previously, but its not set in the current deployment, you want to default to false?

packages/wrangler/src/cloudchamber/apply.ts
emily-shen: i _think_ this is probably the right place to do it, because this isn't part of the container app config, I think it's just passing the overall observability config to the container api?

ichernetsky-cf: It needs to know the current state of affairs because it should not specify, for instance, any observability settings if the observability settings are not set for an application and no observability settings are set in the Wrangler configuration file, however, it should disable them explicitly for an application, if the application has them enabled but they are absent from the configuration file.

packages/containers-shared/src/client/models/AccountDefaults.ts
ichernetsky-cf: It is autogenerated. Nothing under 'packages/containers-shared/src/client` is written by hand, I believe. I had to update the client to make the change in question.

ichernetsky-cf:
> The description of this PR is currently
> 
> > Enable Observability logging for containers if it is enabled using the top-level field in a Wrangler configuration file.
> 
> Could you please expand? There are many changes in this PR and it's not clear how they relate to the description

I apologize for it. Most of the changes is not relevant to the PR description. I had to update the autogenerated client code. The only relevant changes are in `packages/wrangler/src/cloudchamber/apply.ts` and `packages/wrangler/src/__tests__/cloudchamber/apply.test.ts`

packages/wrangler/src/cloudchamber/apply.ts
ichernetsky-cf: Because no diff should be shown unnecessarily.
| App config | Wrangler config | What should be specified |
|-|-|-|
| Undefined | Undefined | Nothing |
| True | True | Nothing |
| False | False | Nothing |
| True | False | False |
| False | True | True |
| False | Undefined | Nothing |
| True | Undefined | False |
| Undefined | True | True |
| Undefined | False | Nothing |

ichernetsky-cf: Let me add this as comment and change the code to mirror the table as close as possible.

packages/wrangler/src/cloudchamber/apply.ts
IRCody: > however, it should disable them explicitly for an application, if the application has them enabled but they are absent from the configuration file.

This is saying if the container configuration has it set but the overall wrangler configuration has it unset we need to explicitly set it?

ichernetsky-cf: Yes, this is true as well, I think

emily-shen: sorry just to clarify, where else can container configuration come from, other than wrangler configuration? And is the point that your API treats false vs unset differently?

ichernetsky-cf: Yes, false and undefined are treated differently. If it is undefined, nothing is done, it is not undefined, it is updated in the DB to a given one.

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -13,23 +13,40 @@ import { msw } from "../helpers/msw";
...
 } from "@cloudflare/containers-shared";
 
-function writeAppConfiguration(...app: ContainerApp[]) {
-	fs.writeFileSync(
-		"./wrangler.toml",
-		TOML.stringify({
-			name: "my-container",
-			containers: app,
-		}),
+function writeWranglerConfiguration(config: {
```
emily-shen: can you use the existing `writeWranglerConfig` test helper? 

I am not sure why all of this logic is necessary, it seems equivalent to just directly writing 
```
{	
    name: "my-container",
	containers: config.apps,
    observability
}
```
if i've missed something, can you add a comment explaining why this is needed?

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -904,4 +949,470 @@ describe("cloudchamber apply", () => {
...
+			"╭ Deploy a container application deploy changes to your application
+			│
+			│ Container application changes
+			│
+			├ EDIT my-container-app
+			│
+			│   [containers.configuration]
+			│   image = \\"./Dockerfile\\"
+			│
+			│ + [containers.configuration.observability.logs]
```
emily-shen, line 1068: - does this mean your api have a different config shape to wrangler config? because you can't set this in wrangler config.

packages/wrangler/src/__tests__/cloudchamber/curl.test.ts
```diff
@@ -240,6 +243,7 @@ describe("cloudchamber curl", () => {
 			            \\"enabled\\": true
 			        },
 			        \\"network\\": {
+			            \\"mode\\": \\"public\\",
```
emily-shen, line 246: why is this change here? is this a separate update to your api client?

packages/wrangler/src/cloudchamber/cli/util.ts
```diff
@@ -15,7 +15,6 @@ export function statusToColored(status?: PlacementStatusHealth): string {
 	const mappings: Record<PlacementStatusHealth, (_: string) => string> = {
 		placed: bgYellow,
 		running: bgGreen,
-		do_connected: bgGreen,
```
emily-shen, line 18: what is this change from?

packages/wrangler/src/cloudchamber/create.ts
```diff
@@ -298,7 +298,9 @@ async function handleCreateCommand(
 	const memoryMib =
 		resolveMemory(args, config.cloudchamber) ??
 		account.defaults.memory_mib ??
-		Math.round(parseByteSize(account.defaults.memory, 1024) / (1024 * 1024));
+		Math.round(
+			parseByteSize(account.defaults.memory ?? "0MiB", 1024) / (1024 * 1024)
```
emily-shen: why this change? is it because memory has been deprecated?

emily-shen (changes requested):
sorry meant to leave that as a comment not approval 😅  - think it should be ok soon, just had a few questions

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
ichernetsky-cf: I wasn't aware of this one. Thanks! Updated

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
ichernetsky-cf: Fixed

packages/wrangler/src/__tests__/cloudchamber/curl.test.ts
ichernetsky-cf: Caused the client update

packages/wrangler/src/cloudchamber/cli/util.ts
ichernetsky-cf: Ditto, due to the client update

packages/wrangler/src/cloudchamber/create.ts
ichernetsky-cf: Yes, the client type has been changed too. I've updated it to 2GiB now

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -108,12 +109,46 @@ function applicationToCreateApplication(
...
+): ObservabilityConfiguration | undefined {
+	const logsAlreadyEnabled =
+		existingObservabilityConfig?.logs?.enabled === true;
+	const observabilityLogsEnabled =
+		observability?.logs?.enabled === true ||
+		(observability?.enabled === true && observability?.logs?.enabled !== false);
+
+	if (logsAlreadyEnabled) {
+		if (observabilityLogsEnabled) {
+			return undefined;
```
IRCody, line 168: Should this be 
{ logs: { enabled: true } } instead of undefined?

vicb:
> > The description of this PR is currently
> > > Enable Observability logging for containers if it is enabled using the top-level field in a Wrangler configuration file.
> > 
> > 
> > Could you please expand? There are many changes in this PR and it's not clear how they relate to the description
> 
> I apologize for it. Most of the changes is not relevant to the PR description. I had to update the autogenerated client code. The only relevant changes are in `packages/wrangler/src/cloudchamber/apply.ts` and `packages/wrangler/src/__tests__/cloudchamber/apply.test.ts`

No problem, only saying the PR description is what I read first before reviewing the code and it would be helpful to add more context there.

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
vicb: > Fixed

Could you please resolve the conversations when they are fixed?
Thanks!

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -108,12 +109,82 @@ function applicationToCreateApplication(
 	return app;
 }
 
+function cleanupObservability(
```
emily-shen, line 112: i'm guessing this observability.logging is deprecated so you're cleaning this up over time? can you add a comment about this?

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -108,12 +109,82 @@ function applicationToCreateApplication(
...
+
 function containerAppToCreateApplication(
 	containerApp: ContainerApp,
+	observability: Observability | undefined,
+	existingApp: Application | undefined,
 	skipDefaults = false
 ): CreateApplicationRequest {
-	const configuration =
-		containerApp.configuration as UserDeploymentConfiguration;
+	const observabilityConfiguration = observabilityToConfiguration(
```
emily-shen, line 181: it doesn't look like this takes into containerApp.configuration.observability from wrangler config? (the field I think you just added)

packages/wrangler/src/cloudchamber/apply.ts
ichernetsky-cf: You are right. Sure thing, adding a comment now.

packages/wrangler/src/cloudchamber/apply.ts
ichernetsky-cf: I will drop the field for now, as the intent of this PR is not about supporting per-container logging configurations, and about using the top-level observability configuration instead.

emily-shen (approved):

PR #9565, opened 2025-06-12
Do not update container apps image if the image is unchanged

IRCody:
Previously we would build the image and retag even if the only change was the build-related metadata. This changes that to detect if the build has resulted in a new image, and if it has not remove the added tag and ensures the applications image is not updated.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Fixing a ux annoyance more than a real behavior change.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Feature doesn't exist in v3.


skepticfx (approved):

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -377,12 +377,20 @@ export async function apply(
 	log(dim("Container application changes\n"));
 
 	for (const appConfigNoDefaults of config.containers) {
+		const application = applicationByNames[appConfigNoDefaults.name];
+		if (
+			(!appConfigNoDefaults.configuration.image ||
+				appConfigNoDefaults.configuration.image === "") &&
+			application !== undefined &&
+			application !== null
```
emily-shen:
```suggestion
			(!appConfigNoDefaults.configuration.image && application  
```
these are all falsy

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -99,10 +100,40 @@ export async function buildAndMaybePush(
...
+		}
+
+		// default changed to true since it is safer to default to updating
+		// the image than not updating
+		let changed = true;
+		if (
+			idAfterBuild &&
+			idAfterBuild.length == 1 &&
+			idsBeforeBuild &&
+			idsBeforeBuild.includes(idAfterBuild[0])
```
emily-shen:
```suggestion
			idAfterBuild?.length &&
			idsBeforeBuild?.includes(idAfterBuild[0])
```
nit

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -99,10 +100,40 @@ export async function buildAndMaybePush(
 			},
 			logger
 		);
+		// Lets grab the current image:ids that match our repository so
+		// we can compare to the result after the build to know if
+		// anything changed.
+		let idsBeforeBuild;
+		if (containerConfig) {
+			idsBeforeBuild = await getDockerImageIds(
```
emily-shen: is this assuming that all images that exist locally in this repo will also exist in the actual registry?

I *think* this isn't valid - the user can run just the build command (which uses this function), so the image would exist locally, but not remotely in the registry. Then we would incorrectly conclude the image doesn't need to be pushed.

IRCody: I think the (implicit) assumption I had made was that if the user was building outside of deploy that they would also push themselves but maybe that is a bad assumption. 

I'll think about how this could work better today. Thanks for pointing this out!

IRCody: I changed it so it grabs the digest locally and uses docker manifest to ensure that it exists at the remote which I think should address this.

emily-shen (approved):

PR #9575, opened 2025-06-12
Support `run_worker_first` in Vite plugin

### LINKED ISSUE #8430: 🐛 BUG: @cloudflare/vite-plugin ignore `run_worker_first`

### LINKED ISSUE #8879: Routing Conflict with `@cloudflare/vite-plugin` in React SPA

Vite Plugin

wrangler: ^4.7.0, vite: ^6.1.0, @cloudflare/vite-plugin: ^1.0.2, hono: ^4.7.5, react: ^19.0.0 (Node.js: unknown, C3: unknown)

_No response_

## Routing Conflict with `@cloudflare/vite-plugin` in React SPA

When using the `@cloudflare/vite-plugin` in a React SPA setup—such as the `vite-react-template` or any custom project following a similar structure—a routing conflict occurs when attempting to directly navigate to server-side API routes via the browser.

### Problem Description

This happens because the SPA fallback behavior (enabled by setting `"not_found_handling": "single-page-application"` in `wrangler.json`) intercepts requests with `Sec-Fetch-Mode: navigate` **before** the Worker script executes. As a result, the SPA’s `index.html` is served even for intended API routes.

This breaks scenarios where API routes need to be reachable directly from the browser, such as:

- OAuth callbacks (e.g., `/api/auth/callback/...`)
- Server-initiated redirects to API endpoints

> **Note:** Programmatic `fetch` calls from the SPA to `/api/*` routes still work as expected, since they use `cors` mode and do reach the Worker.

---

### Use Case / Impact

This breaks critical server-side functionality that relies on **direct browser interaction** with API routes, including:

- OAuth callback handling from providers like Google or GitHub
- Redirect-based flows where the browser is expected to land on a server-side route

---

### Steps to Reproduce (Conceptual)

1. Set up a project using either the `vite-react-template` or a custom project using `@cloudflare/vite-plugin` with Hono.
2. Define a simple API route in Hono:
   ```ts
   app.get('/api/test', (c) => c.text('API OK'))
   ```
3. Deploy using default or SPA-configured `wrangler.json`.
4. Attempt to visit `https://<your-worker-url>/api/test` directly in the browser.

#### Expected Behavior

The API handler should respond with `API OK`.

#### Actual Behavior

The SPA’s `index.html` is served instead, and the Hono route is bypassed.

---

### Troubleshooting / Workarounds Attempted

1. **Changing `not_found_handling`**
   - Switching to `"404"` allows API routes to work
   - But breaks SPA routing
   - Requires manually serving `index.html` from Hono, which adds complexity

2. **Using `run_worker_first`**
   - Setting `run_worker_first: true` along with SPA fallback and a valid `assets` config **does** allow the Worker to run first
   - But for the SPA to still work, a catch-all fallback route is needed in Hono like:
     ```ts
     app.get('*', (c) => c.env.ASSETS.fetch(c.req.raw))
     ```
   - This pattern is not included in the default template and isn’t obvious unless you read deeper into the docs

3. **Header Workarounds**
   - Attempted to handle requests based on `Sec-Fetch-Mode`
   - But routing decisions are made **before** reaching the Worker

---

### Question / Suggestion

What’s the recommended approach for using `@cloudflare/vite-plugin` with React (whether using the `vite-react-template` or a custom setup) to support **both** SPA behavior **and** direct browser access to API routes?

If the correct pattern is:

- `run_worker_first: true`
- `not_found_handling: "single-page-application"`
- A proper `assets` config
- A final catch-all Hono route to serve assets

...could this be documented more clearly or baked into the template directly?

Many full-stack use cases will hit this routing issue, and the workaround isn’t obvious without digging.

---

### Environment

- `wrangler`: ^4.7.0  
- `@cloudflare/vite-plugin`: ^1.0.2  
- `vite`: ^6.1.0  
- `hono`: ^4.7.5  
- `react`: ^19.0.0  
- `node`: [e.g., 20.x.x]

---

Thanks for maintaining these tools and templates!

_No response_

### LINKED ISSUE #9109: SPA fallback broken on LAN IP in dev with @cloudflare/vite-plugin enabled

Vite Plugin

- OS: Arch Linux,
- Node.js: 23.11.0
 - Bun: 1.2.10,
- Vite: 6.3.4,
- @cloudflare/vite-plugin: 1.0.13
- @tanstack/router-plugin: 1.117.2

https://github.com/kernpinoy/issue

### Describe the Bug

When developing a React SPA with Vite and the `@cloudflare/vite-plugin`, direct navigation to non-root routes (e.g., `/home`) via the local network IP (e.g., `http://10.0.0.25:5173/home`) results in a 404 error. This only occurs when the Cloudflare Vite plugin is enabled. The issue does **not** occur when accessing the same route via `localhost`, nor does it occur when the Cloudflare plugin is disabled.

---

### Steps to Reproduce

1. Create a Vite React project with the following plugins:
    - `@vitejs/plugin-react-swc`
    - `@cloudflare/vite-plugin`
    - `@tanstack/router-plugin/vite`
2. Start the dev server with:
    ```bash
    bun run dev -- --host
    # or
    vite --host
    ```
3. On another device on the same network, navigate to `http://<your-ip>:5173/home` (replace `<your-ip>` with your machine’s LAN IP).
4. Observe that a 404 error is returned.
5. Comment out or remove the `@cloudflare/vite-plugin` from `vite.config.js`.
6. Restart the dev server and repeat step 3.
7. The route now loads as expected.

---

### Expected Behavior

- Navigating directly to any client-side route (e.g., `/home`) via the LAN IP should load the SPA and render the correct page, just as it does when using `localhost`.

---

### Actual Behavior

- Navigating directly to a client-side route via the LAN IP returns a 404 error when the Cloudflare Vite plugin is enabled.

---

### Additional Notes

- The issue does **not** occur when using `localhost`.
- The issue does **not** occur when the Cloudflare plugin is disabled.
- Clicking links within the app (client-side navigation) works as expected.
- This behavior breaks mobile and cross-device testing in development.

### Screen recordings
- [Mobile screenrec](https://github.com/user-attachments/assets/84c569de-0d3c-411d-b939-848ccf0d0859)
- [PC screenrec](https://github.com/user-attachments/assets/0313efb7-a389-4dec-876c-8a3ef2c88c35)

I was unable to get error logs as there was none. Sorry.

jamesopstad:
Fixes #8430.
Fixes #8879.
Fixes #9109.

Support `run_worker_first` and static routing.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: to follow
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/tsconfig.plugin.json
```diff
@@ -1,5 +1,5 @@
 {
-	"extends": "@cloudflare/workers-tsconfig/base.json",
+	"extends": "@cloudflare/workers-tsconfig/worker.json",
```
jamesopstad, line 2: This change was necessary to prevent type errors when importing from `workers-shared` as `workers-shared` uses `@cloudflare/worker-types`.

packages/vite-plugin-cloudflare/playground/spa-with-api/api/index.ts
```diff
@@ -1,17 +1,30 @@
+import asset from "./asset.txt?no-inline";
```
petebacondarwin:
```suggestion
import assetURL from "./asset.txt?no-inline";
```

petebacondarwin: or

```suggestion
import assetPath from "./asset.txt?no-inline";
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -330,50 +331,102 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+							const entryWorker = await miniflare.getWorker(entryWorkerName);
+
+							return entryWorker.fetch;
+						});
+					}
+
+					const staticRouting: StaticRouting | undefined =
+						entryWorkerConfig.assets?.run_worker_first === true
+							? { user_worker: ["/*"] }
+							: resolvedPluginConfig.staticRouting;
```
petebacondarwin, line 356: How about moving this bit of logic into the plugin-config or asset-config files?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -330,50 +331,102 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+						const userWorkerHandler = createRequestHandler(async (request) => {
+							assert(miniflare, `Miniflare not defined`);
+							const userWorker = await miniflare.getWorker(entryWorkerName);
+
+							return userWorker.fetch(request, { redirect: "manual" });
+						});
+
+						preMiddleware = async (req, res, next) => {
+							// Only the URL pathname is used to match rules
+							const request = new Request(new URL(req.url!, UNKNOWN_HOST));
```
petebacondarwin: Is this needed? Can't you just pas `req` through to each matcher?

petebacondarwin: Ideally avoid ! assertions and use a runtime assertion to narrow the type:

```suggestion
							assert(req.url, "expected request URL to be defined");
							const request = new Request(new URL(req.url, UNKNOWN_HOST));
```

packages/vite-plugin-cloudflare/tsconfig.plugin.json
petebacondarwin: Hmm. OK.
I think we should be able to avoid that... I'll have a think about it. But not blocking this PR.

petebacondarwin (approved):

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: `req` is a Node.js `http.IncomingMessage`. The matchers require a `Request` object, even though they only use the URL pathname.

https://github.com/cloudflare/workers-sdk/blob/70ba9fbf905a9ba5fe158d0bc8d48f6bf31712a2/packages/workers-shared/asset-worker/src/utils/rules-engine.ts#L151

Would be nice to change this but I guess it's for consistency with the other utils (also odd that `generateStaticRoutingRuleMatcher` is in the Asset Worker utils when it's only used by the Router Worker).

PR #9576, opened 2025-06-12
sync local containers with latest workerd

vicb:
The first commit populate the `container : ContainerOptions` in `DurableObjectNamespace`. Encoding the capnp struct fails for now as the the version of `workerd` used by miniflare does not support containers. Will need to update the generate code after the `workerd` update.

The second commit adds a `container_engine` configuration to wrangler. It should be either a unix socket `unix:/path/to/socket` or a full configuration. Note that as @danlapid mentioned, win32 is not supported ATM.

Edit:

I added 2 more fixup! commits.

- `fixup! do not add container to DurableObjectNamespaceDesignator` fixes the first commit.
- `fixup! socket address default to WRANGLER_DOCKER_SOCKET` reads the Socket address from WRANGLER_DOCKER_SOCKET when no value is specified in the wrangler config file. Open question: do we want to keep the config file option or always use the env var? Also note that `Miniflare` still uses the platform default when no value is provided (for people using `Miniflare` directly)

Edit:

- `WRANGLER_DOCKER_SOCKET` renamed to `WRANGLER_DOCKER_HOST` in the review feedback fixup

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: will be covered by e2e
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: separate effort
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: New feature


packages/miniflare/src/plugins/core/index.ts
```diff
@@ -1023,6 +1040,29 @@ function getWorkerScript(
...
+ * @param engineOrSocketPath Either a full engine config or a unix socket
+ * @returns The container engine, default to local Docker at `unix:/var/run/docker.sock`
+ */
+function getContainerEngine(
+	engineOrSocketPath: Worker_ContainerEngine | string | undefined
+): Worker_ContainerEngine {
+	if (!engineOrSocketPath) {
+		// TODO: workerd does not support win named pipes
+		const socketPath =
+			platform() === "win32" ? "TODO" : "unix:/var/run/docker.sock";
```
emily-shen: can we add an env var to allow users to customise this?

packages/wrangler/src/config/environment.ts
emily-shen, line 1: imo this should be set as an env var rather than in wrangler config, as it is dev only. i think this is more aligned with wrangler's usual patterns, if you look at the other stuff in evnrionment-variables/factory.ts

(i also don't think this is hooked up to anything fwiw)

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -702,25 +703,32 @@ export function buildMiniflareBindingOptions(
...
-					const useSQLite = classNameToUseSQLite.get(class_name);
+				internalObjects.map(({ class_name: className }) => {
 					return [
-						class_name,
-						{ className: class_name, scriptName: getName(config), useSQLite },
+						className,
+						{
+							className,
+							scriptName: getName(config),
+							useSQLite: classNameToUseSQLite.get(className),
```
emily-shen, line 732: i think this should have the container configured here.

thinking aloud here:
this is my understanding of this situation:
Screenshot 2025-06-12 at 14 54 01

so i think if you had a DO in another session that had a container, provided you called the fetch handler only you could access that container. 

so this config here should be the same config as the internal object in line 1015. 


whereas the external object should not have containers configured. (which is what you have already)/

does that sound right to you?

packages/miniflare/src/plugins/core/index.ts
vicb: That is only the default value when the user does not provide a value.

They can add `containerEngine: "unix:/my/own/path"` to their wrangler config to add an explicit path

I don't think we need an env var on top of that

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -168,6 +170,16 @@ const CoreOptionsSchemaInput = z.intersection(
 		// There is an issue with the connect() API and the globalOutbound workerd setting that impacts TCP ingress
 		// We should default it to true once https://github.com/cloudflare/workerd/pull/4145 is resolved
 		stripCfConnectingIp: z.boolean().default(false),
+
+		/** Configuration used to connect to the container engine */
+		containerEngine: z
```
vicb, line 173: @emily-shen my question was about that.

We allow both `/socket/path` and `{localDocker: {socketPath: "socketPath"}`

I think allowing only `/socket/path` would make sense - we could switch to the object form later when there are more options to configure.

emily-shen: i don't feel strongly, but if we changed it later we will probably want to support both as string and object. which might be fine, but might also be messy. but no strong opinions here

packages/miniflare/src/plugins/core/index.ts
emily-shen: maybe we should just error right now for windows? and suggest WSL in the error message

emily-shen: regarding the env var, i meant instead of the wrangler config sorry. 

to mirror this option: https://github.com/cloudflare/workers-sdk/blob/e16fcc747aa7701405eb4f49a73e622425f67527/packages/wrangler/src/environment-variables/misc-variables.ts#L261

if you do prefer to have it in wrangler config, maybe call it `localContainerEngine` instead?

(as a side note, it bugs me that we've ended up using both CLI and API. I think afterwards it would be good to have another go at using the api, but that's too much refactoring to try now while the CC time are trying to land other PRs.)

vicb: I'll take a look later

packages/wrangler/src/dev/miniflare.ts
vicb: Wouldn't that start 2 containers?

emily-shen: i dont know. this config should be referring to the same DO, so i don't think so. anyway its not an important thing to focus on here, i'll stick a followup ticket to think about cross-process wrangler dev + containers.

packages/miniflare/src/plugins/core/index.ts
vicb: We do already have [string | object in DO](https://github.com/cloudflare/workers-sdk/blob/1d3293f0cbf88a45d7b86bae0fc886e08aa6e841/packages/miniflare/src/plugins/do/index.ts#L18)

I think it makes sense to support both in Miniflare, I should have commented on `wrangler` instead.

Maybe keep both but document the string version in wrangler for now?

vicb:
@emily-shen what about merging this PR as is?

We could discuss some points and create follow-up PRs if need be:
- This PR supports both `WRANGLER_DOCKER_SOCKET` and a wrangler dev config to set the socket
- Make sure we do the right thing for external DOs

emily-shen:
> @emily-shen what about merging this PR as is?
> 
> We could discuss some points and create follow-up PRs if need be:
> 
> * This PR supports both `WRANGLER_DOCKER_SOCKET` and a wrangler dev config to set the socket
> * Make sure we do the right thing for external DOs

looks like `DOCKER_HOST` is a standard: https://podman-desktop.io/docs/migrating-from-docker/using-the-docker_host-environment-variable.

totally fine to punt on external DOs, don't think its a big deal.

packages/miniflare/src/plugins/core/index.ts
emily-shen: yeah string | object is all over the place in miniflare. documenting the string version in wrangler sounds good. my only preference in wrangler is to nest it under `dev` like `enableContainers`.

packages/wrangler/src/config/environment.ts
```diff
@@ -546,11 +546,27 @@ export interface EnvironmentNonInheritable {
...
+	 * Either the Docker unix socket i.e. `unix:/var/run/docker.sock` or a full configuration.
+	 * Note that windows is only supported via WSL at the moment
+	 *
+	 * NOTE: This field is not automatically inherited from the top level environment,
+	 * and so must be specified in every named environment.
+	 *
+	 * @default {}
+	 * @nonInheritable
+	 */
+	containerEngine?: ContainerEngine;
```
emily-shen: see above for comment for sticking it in `dev` alognside `enableContainers`

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -1023,6 +1040,29 @@ function getWorkerScript(
...
+/**
+ * Returns the Container engine configuration
+ * @param engineOrSocketPath Either a full engine config or a unix socket
+ * @returns The container engine, default to local Docker at `unix:/var/run/docker.sock`
+ */
+function getContainerEngine(
+	engineOrSocketPath: Worker_ContainerEngine | string | undefined
+): Worker_ContainerEngine {
+	if (!engineOrSocketPath) {
+		// TODO: workerd does not support win named pipes
```
emily-shen, line 1051: we should hard fail on windows for now (DEVX-1996)

packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -264,3 +265,15 @@ export const getDockerPath = getEnvironmentVariableFactory({
 		return "docker";
 	},
 });
+
+/**
+ * `WRANGLER_DOCKER_SOCKET` specifies the Docker socket to connect to.
+ */
+export const getDockerSocket = getEnvironmentVariableFactory({
+	variableName: "WRANGLER_DOCKER_SOCKET",
```
emily-shen:
```suggestion
	variableName: "WRANGLER_DOCKER_HOST",
```

cf https://podman-desktop.io/docs/migrating-from-docker/using-the-docker_host-environment-variable etc.

vicb: Nice!

vicb:
@emily-shen I renamed for the env var to `WRANGLER_DOCKER_HOST`.

I have tried to address the config thing but \<rant\>there is nothing that bothers me more that trying to understand code w/o comments - we have 10 different config types, most have no comments, function have no comments. There is probably a good reason to have all those different objects and functions but I will not spend more time trying to reverse engineering the code - I think we should do a better job commenting our codebase!\</rant\>

emily-shen (approved):

emily-shen:
blocked pending the next workerd release which has a bugfix.

packages/miniflare/src/plugins/containers/service/index.ts
```diff
@@ -26,6 +27,11 @@ export class ContainerController {
 		this.#containerOptions = containerOptions;
 		this.#sharedOptions = sharedOptions;
 		this.#logger = logger;
+		if (platform() === "win32") {
```
vicb: Maybe move this to the DO plugin when there are containers (wrangler do not pass any when they are disabled)

vicb: nvm good to have this here now while Miniflare still builds the image

anonrig: If we use navigator.platform or process.platform you don't have to pull in node:os module, which will impact the boot time of this node.js process

anonrig: (this also applies to all other usages)

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -1,6 +1,7 @@
 import assert from "assert";
 import { readFileSync } from "fs";
 import fs from "fs/promises";
+import { platform } from "node:os";
```
anonrig, line 4: This is not needed

PR #9577, opened 2025-06-12
feat(containers-shared): Move `isDockerfile` utility function to `containers-shared`

CarmenPopoviciu:
As part of DEVX-1966, we are moving the `isDockerfile` utility function and its corresponding tests to `containers-shared`. To support running tests for this package, this PR also adds some of the necessary vitest config.

Fixes DEVX-1966

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [ ] Documentation not necessary because: just a refactor
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because: this is an experimental feature still

### DIFF
--- a/packages/containers-shared/src/utils.ts
+++ b/packages/containers-shared/src/utils.ts
@@ -1,4 +1,5 @@
 import { spawn, StdioOptions } from "child_process";
+import { existsSync, statSync } from "fs";
 
 /** helper for simple docker command call that don't require any io handling */
 export const runDockerCmd = async (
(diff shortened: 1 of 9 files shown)


packages/containers-shared/src/utils.ts
```diff
@@ -40,3 +41,48 @@ export const verifyDockerInstalled = async (dockerPath: string) => {
...
+export function isDir(path: string) {
+	const stats = statSync(path);
+	return stats.isDirectory();
+}
+
+export const isDockerfile = (image: string): boolean => {
+	// TODO: move this into config validation
+	if (existsSync(image)) {
+		if (isDir(image)) {
+			throw new Error(
```
CarmenPopoviciu, line 54: all these were `UserError`s, but I replaced them with `Error` instead, since I believe it makes more sense to uncouple this utility function from any wrangler specific error type. Nothing has changed in the code that calls this fn however, so no `try...catch`. If we need to throw a `UserError` specifically, I'll adapt the code in wrangler, but wanted to get thoughts first 🙏

packages/containers-shared/tests/utils.test.ts
```diff
@@ -0,0 +1,55 @@
...
+	const dockerfile =
+		'FROM node:18\nWORKDIR /app\nCOPY . .\nRUN npm install\nCMD ["node", "index.js"]';
+
+	beforeEach(() => {
+		mkdirSync("./container-context");
+		writeFileSync("./container-context/Dockerfile", dockerfile);
+	});
+
+	afterEach(() => {
+		rmSync("./container-context", { recursive: true });
```
CarmenPopoviciu: this was not in the original code, but the tests kept failing while attempting to write the dockerfile, due to the file already existing. I couldn't make out how the original tests did not run into this, but removing the directory recursively will solve the issue. Also, this node API was introduced in `v14.14.0` so we should be good in terms of node support. No other changes were made to these tests

emily-shen: what about running this in a temporary directory? think thats what the wrangler tests were doing

CarmenPopoviciu: oh right...yeah, let me try that

CarmenPopoviciu:
do we need a changeset here?

CarmenPopoviciu:
> do we need a changeset here?

answering my own question, but no, because no user facing changes, just refactoring

packages/containers-shared/tests/utils.test.ts
CarmenPopoviciu: I ended up duplicating-ish the fn in wrangler, but only the necessary bits. In a future world, we'd want to extract such helpers into a separate package that we can share across our workers-sdk packages

PR #9582, opened 2025-06-12
Update capnp generated code from the workerd.capnp

vicb:


vicb:
Thanks for the review Emily

PR #9584, opened 2025-06-12
Revert "Revert "chore: update node versions in various places""

penalosa:
Reverts cloudflare/workers-sdk#9570


penalosa:
This is blocked on https://github.com/nodejs/node/issues/57869

PR #9596, opened 2025-06-13
feat(containers): add support for handling images that link to the CF registry

CarmenPopoviciu:
Fixes DEVX-1966.

the first commit adds the pulling functionality, the second just makes manual testing a bit easier. e2e testing will happen in a followup.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: to do in followup. manually tested.
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: outside the scope of this PR
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: containers is an experimental feature

### DIFF
--- a/fixtures/container-app/src/index.ts
+++ b/fixtures/container-app/src/index.ts
@@ -7,33 +7,64 @@ export class Container extends DurableObject<Env> {
 	constructor(ctx: DurableObjectState, env: Env) {
 		super(ctx, env);
 		this.container = ctx.container!;
-		void this.ctx.blockConcurrencyWhile(async () => {
-			if (!this.container.running) this.container.start();
-		});
 	}
 
 	async fetch(req: Request) {
-		try {
-			return await this.container
-				.getTcpPort(8080)
-				.fetch(req.url.replace("https:", "http:"), req);
-		} catch (err) {
-			return new Response(`${this.ctx.id.toString()}: ${err.message}`, {
-				status: 500,
-			});
+		const path = new URL(req.url).pathname;
+		switch (path) {
+			case "/status":
+				return new Response(JSON.stringify(this.container.running));
+
+			case "/destroy":
+				if (!this.container.running) {
+					throw new Error("Container is not running.");
+				}
+				await this.container.destroy();
+				return new Response(JSON.stringify(this.container.running));
+
+			case "/start":
+				this.container.start({
+					entrypoint: ["node", "app.js"],
+					env: { A: "B", C: "D", L: "F" },
+					enableInternet: false,
+				});
+				// this doesn't instantly start, so we will need to poll /fetch
+				return new Response("Container create request sent...");
+
+			case "/fetch":
+				const res = await this.container
+					.getTcpPort(8080)
+					// actual request doesn't matter
+					.fetch("http://foo/bar/baz", { method: "POST", body: "hello" });
+				return new Response(await res.text());
+
+			case "/destroy-with-monitor":
+				// if (!this.container.running) {
+				// 	throw new Error("Container is not running.");
+				// }
+				const monitor = this.container.monitor();
+				await this.container.destroy();
+				await monitor;
+				return new Response("Container destroyed with monitor.");
+
+			default:
+				return new Response("Hi from Container DO");
 		}
 	}
 }
 
 export default {
 	async fetch(request, env): Promise<Response> {
-		try {
-			return await env.CONTAINER.get(env.CONTAINER.idFromName("fetcher")).fetch(
-				request
-			);
-		} catch (err) {
-			console.error("Error fetch:", err.message);
-			return new Response(err.message, { status: 500 });
+		const url = new URL(request.url);
+		if (url.pathname === "/second") {
+			// This is a second Durable Object that can be used to test multiple DOs
+			const id = env.CONTAINER.idFromName("second-container");
+			const stub = env.CONTAINER.get(id);
+			const query = url.searchParams.get("req");
+			return stub.fetch("http://example.com/" + query);
 		}
+		const id = env.CONTAINER.idFromName("container");
+		const stub = env.CONTAINER.get(id);
+		return stub.fetch(request);
 	},
 } satisfies ExportedHandler<Env>;
--- a/packages/containers-shared/src/build.ts
+++ b/packages/containers-shared/src/build.ts
@@ -1,10 +1,7 @@
 import { spawn } from "child_process";
 import { readFileSync } from "fs";
 import path from "path";
-import { dockerImageInspect } from "./inspect";
-import { MF_DEV_CONTAINER_PREFIX } from "./registry";
 import { BuildArgs, ContainerDevOptions, Logger } from "./types";
-import { verifyDockerInstalled } from "./utils";
 
 export async function constructBuildCommand(
 	options: BuildArgs,
@@ -70,33 +67,7 @@ export function dockerBuild(
 	});
 }
 
-/**
- *
- * Builds (or pulls - TODO) the container images for local development. This
- * will be called before starting the local development server, and by a rebuild
- * hotkey during development.
- *
- * Because this runs when local dev starts, we also do some validation here,
- * such as checking if the Docker CLI is installed, and if the container images
- * expose any ports.
- */
-export async function prepareContainerImagesForDev(
-	dockerPath: string,
-	containerOptions: ContainerDevOptions[]
-) {
-	if (process.platform === "win32") {
-		throw new Error(
-			"Local development with containers is currently not supported on Windows. You should use WSL instead. You can also set `enable_containers` to false if you do not need to develop the container as part of your application."
-		);
-	}
-	await verifyDockerInstalled(dockerPath);
-	for (const options of containerOptions) {
-		await buildContainer(dockerPath, options);
-		await checkExposedPorts(dockerPath, options.imageTag);
-	}
-}
-
-async function buildContainer(
+export async function buildImage(
 	dockerPath: string,
 	options: ContainerDevOptions
 ) {
@@ -110,17 +81,4 @@ async function buildContainer(
 	});
 
 	await dockerBuild(dockerPath, { buildCmd, dockerfile });
-}
-
-async function checkExposedPorts(dockerPath: string, imageTag: string) {
-	const output = await dockerImageInspect(dockerPath, {
-		imageTag,
-		formatString: "{{ len .Config.ExposedPorts }}",
-	});
-	if (output === "0" && process.platform !== "linux") {
-		throw new Error(
-			`The container "${imageTag.replace(MF_DEV_CONTAINER_PREFIX + "/", "")}" does not expose any ports.\n` +
-				"To develop containers locally on non-Linux platforms, you must expose any ports that you call with `getTCPPort()` in your Dockerfile."
-		);
-	}
 }
--- a/packages/containers-shared/src/images.ts
+++ b/packages/containers-shared/src/images.ts
@@ -1,11 +1,8 @@
 import { execFile } from "child_process";
 import { buildImage } from "./build";
-import {
-	getCloudflareContainerRegistry,
-	isCloudflareRegistryLink,
-} from "./knobs";
+import { isCloudflareRegistryLink } from "./knobs";
 import { dockerLoginManagedRegistry } from "./login";
-import { ContainerDevOptions, Logger } from "./types";
+import { ContainerDevOptions } from "./types";
 import {
 	checkExposedPorts,
 	isDockerfile,
@@ -40,32 +37,21 @@ export async function pullImage(
 	dockerPath: string,
 	options: ContainerDevOptions
 ) {
-	try {
-		await dockerLoginManagedRegistry(dockerPath);
-		await runDockerCmd(dockerPath, [
-			"pull",
-			options.image,
-			"--platform",
-			"linux/amd64",
-		]);
-		// const tag = `${MF_CONTAINER_PREFIX}/${options.name}`;
-		// this needs to be tested cause we might have to extract the actual image tag
-		await runDockerCmd(dockerPath, ["tag", options.image, options.imageTag]);
-		// this.#logger.info(`Successfully pulled image: ${image}`);
-	} catch (error) {
-		if (error instanceof Error) {
-			throw new Error(error.message);
-		}
-
-		throw new Error(
-			`An unknown error occurred while attempting to pull ${options.image} from the Cloudflare container registry`
-		);
-	}
+	await dockerLoginManagedRegistry(dockerPath);
+	await runDockerCmd(dockerPath, [
+		"pull",
+		options.image,
+		// All containers running on our platform need to be built for amd64 architecture, but by default docker pull seems to look for an image matching the host system, so we need to specify this here
+		"--platform",
+		"linux/amd64",
+	]);
+	// re-tag image with the expected dev-formatted image tag for consistency
+	await runDockerCmd(dockerPath, ["tag", options.image, options.imageTag]);
 }
 
 /**
  *
- * Builds (or pulls - TODO) the container images for local development. This
+ * Builds or pulls the container images for local development. This
  * will be called before starting the local development server, and by a rebuild
  * hotkey during development.
  *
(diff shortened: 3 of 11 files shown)


packages/miniflare/src/plugins/containers/service/index.ts
```diff
@@ -46,10 +51,21 @@ export class ContainerController {
...
 		this.#logger.info("Building container(s)...");
 		for (const options of Object.values(this.#containerOptions)) {
-			await this.buildContainer(options);
+			if (isDockerfile(options.image)) {
+				await this.buildContainer(options);
+			} else {
+				if (!isCloudflareRegistryLink(options.image)) {
+					throw new Error(
+						`Image ${options.image} is a registry link but does not point to the Cloudflare container registry.\n` +
+							`All images must use ${getCloudflareContainerRegistry()}, which is the default registry for Wrangler. To use an existing image from another repository, see https://developers.cloudflare.com/containers/image-management/#using-existing-images`
```
CarmenPopoviciu: @mikenomitch is this the correct url I'm using here?

packages/containers-shared/src/knobs.ts
```diff
@@ -2,3 +2,13 @@
 export const getCloudflareContainerRegistry = () => {
 	return process.env.CLOUDFLARE_CONTAINER_REGISTRY ?? "registry.cloudflare.com";
 };
+
+/**
+ * Given a container image that is a registry link, this function
+ * returns true if the link points the Cloudflare container registry
+ * (defined as per `getCloudflareContainerRegistry` above)
+ */
+export function isCloudflareRegistryLink(image: string) {
```
CarmenPopoviciu, line 14: is this function actually doing what we want? `process.env.CLOUDFLARE_CONTAINER_REGISTRY` could technically be set to anything, so I'm wondering if the check here shouldn't be `image.includes("registry.cloudflare.com")` instead 🤔

packages/containers-shared/src/knobs.ts
```diff
@@ -2,3 +2,13 @@
...
+	const cfRegistry = getCloudflareContainerRegistry();
```
anonrig, line 15: This will create a new string everytime it gets executed. Using something like this and definining cfRegistry in global scope will remove the unnecessary string allocation.

```suggestion
	cfRegistry ??= getCloudflareContainerRegistry();
```

packages/miniflare/src/plugins/containers/service/index.ts
```diff
@@ -79,4 +95,24 @@ export class ContainerController {
...
+			await dockerLoginManagedRegistry(this.#sharedOptions.dockerPath);
+			await runDockerCmd(this.#sharedOptions.dockerPath, ["pull", image]);
+			// const tag = `${MF_CONTAINER_PREFIX}/${options.name}`;
+			// this needs to be tested cause we might have to extract the actual image tag
+			// await runDockerCmd(this.#sharedOptions.dockerPath, ["tag", image, tag]);
+			this.#logger.info(`Successfully pulled image: ${image}`);
+		} catch (error) {
+			if (error instanceof Error) {
+				throw new Error(error.message);
+			}
```
anonrig: without the following change, the stack trace will be gone.

```suggestion
			if (Error.isError(error)) {
				throw new Error(error.message, { cause: error );
			}
```

CarmenPopoviciu: TIL 🤯 . But also, this is only partially supported starting node `v24`, which was released just a few weeks ago. wrangler needs to support lower node versions `>=18.20.0`

packages/containers-shared/src/build.ts
```diff
@@ -70,33 +67,7 @@ export function dockerBuild(
 	});
 }
 
-/**
```
emily-shen, line 73: just moved this to `images.ts`

packages/containers-shared/src/build.ts
```diff
@@ -111,16 +82,3 @@ async function buildContainer(
 
 	await dockerBuild(dockerPath, { buildCmd, dockerfile });
 }
-
```
emily-shen, line 114: moved this to utils

packages/containers-shared/src/knobs.ts
emily-shen: i think this is fine, it should work with both

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -400,6 +403,15 @@ async function resolveConfig(
 		);
 	}
 
+	// for pulling containers, we need to make sure the OpenAPI config for the
+	// container API client is properly set
```
penalosa: Why do we need open api config for container image pulling?

packages/containers-shared/src/utils.ts
```diff
@@ -153,3 +155,16 @@ const getContainerIdsFromImage = async (
 	]);
 	return output.split("\n").filter((line) => line.trim());
 };
+
+export async function checkExposedPorts(dockerPath: string, imageTag: string) {
```
penalosa: doc comment?

packages/containers-shared/src/utils.ts
```diff
@@ -153,3 +155,16 @@ const getContainerIdsFromImage = async (
...
+	const output = await dockerImageInspect(dockerPath, {
+		imageTag,
+		formatString: "{{ len .Config.ExposedPorts }}",
+	});
+	if (output === "0" && process.platform !== "linux") {
+		throw new Error(
+			`The container "${imageTag.replace(MF_DEV_CONTAINER_PREFIX + "/", "")}" does not expose any ports.\n` +
```
penalosa: What if the image tag doesn't have this prefix?

packages/containers-shared/src/utils.ts
```diff
@@ -153,3 +155,16 @@ const getContainerIdsFromImage = async (
...
+				"To develop containers locally on non-Linux platforms, you must expose any ports that you call with `getTCPPort()` in your Dockerfile."
```
penalosa, line 178: Could we link to the dockerfile `EXPOSE` docs here?

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -400,6 +403,15 @@ async function resolveConfig(
...
+	const needsPulling = resolved.containers?.some(
+		(c) => !isDockerfile(c.image ?? c.configuration.image)
+	);
+	if (needsPulling && !resolved.dev.remote) {
+		await fillOpenAPIConfiguration(config, isNonInteractiveOrCI());
```
penalosa, line 413: Does this pull in auth at all? `ConfigController` has an auth hook that we should use

packages/containers-shared/src/knobs.ts
```diff
@@ -2,9 +2,19 @@ import { MF_DEV_CONTAINER_PREFIX } from "./registry";
 
 // default cloudflare managed registry, can be overriden with the env var - CLOUDFLARE_CONTAINER_REGISTRY
 export const getCloudflareContainerRegistry = () => {
-	return process.env.CLOUDFLARE_CONTAINER_REGISTRY ?? "registry.cloudflare.com";
+	return process.env.CLOUDFLARE_CONTAINER_REGISTRY ?? "registry.cloudflare.com"; // registry.cloudchamber.cfdata.org
```
penalosa: What's this comment for?

penalosa: This should be using the env var factory functions we have set up in `misc-variables.ts`

fixtures/container-app/src/index.ts
penalosa, line 1: Is this for manual testing? I can't see any tests of this logic in the fixture

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+			"linux/amd64",
+		]);
+		// const tag = `${MF_CONTAINER_PREFIX}/${options.name}`;
+		// this needs to be tested cause we might have to extract the actual image tag
+		await runDockerCmd(dockerPath, ["tag", options.image, options.imageTag]);
+		// this.#logger.info(`Successfully pulled image: ${image}`);
+	} catch (error) {
+		if (error instanceof Error) {
+			throw new Error(error.message);
+		}
```
penalosa: What is this doing? Why do we reconstruct the Error?

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+
+		throw new Error(
+			`An unknown error occurred while attempting to pull ${options.image} from the Cloudflare container registry`
+		);
```
penalosa: Could we add a debug log, at least? What kinds of errors do we think this will swallow?

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+		await runDockerCmd(dockerPath, [
+			"pull",
+			options.image,
+			"--platform",
+			"linux/amd64",
+		]);
+		// const tag = `${MF_CONTAINER_PREFIX}/${options.name}`;
+		// this needs to be tested cause we might have to extract the actual image tag
+		await runDockerCmd(dockerPath, ["tag", options.image, options.imageTag]);
+		// this.#logger.info(`Successfully pulled image: ${image}`);
```
penalosa: Should this be removed?

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+	dockerPath: string,
+	options: ContainerDevOptions
+) {
+	try {
+		await dockerLoginManagedRegistry(dockerPath);
+		await runDockerCmd(dockerPath, [
+			"pull",
+			options.image,
+			"--platform",
+			"linux/amd64",
```
penalosa: Could you add a comment explaining why we override the platform?

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+		]);
+		// const tag = `${MF_CONTAINER_PREFIX}/${options.name}`;
```
penalosa: Should this be removed?

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+		// this needs to be tested cause we might have to extract the actual image tag
```
penalosa: Is this a TODO? Have we tested it?

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+	}
+	await verifyDockerInstalled(dockerPath);
+	for (const options of containerOptions) {
+		if (isDockerfile(options.image)) {
+			await buildImage(dockerPath, options);
+		} else {
+			if (!isCloudflareRegistryLink(options.image)) {
+				throw new Error(
+					`Image "${options.image}" is a registry link but does not point to the Cloudflare container registry.\n` +
+						`All images must use ${getCloudflareContainerRegistry()}, which is the default registry for Wrangler. To use an existing image from another repository, see https://developers.cloudflare.com/containers/image-management/#using-existing-images`
```
penalosa: > which is the default registry for Wrangler

If _all_ images need to use this registry why is it customisable? Why do we use the "default" language?

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +35,66 @@ export async function getDockerImageDigest(
...
+		await dockerLoginManagedRegistry(dockerPath);
+		await runDockerCmd(dockerPath, [
+			"pull",
+			options.image,
+			"--platform",
+			"linux/amd64",
+		]);
+		// const tag = `${MF_CONTAINER_PREFIX}/${options.name}`;
+		// this needs to be tested cause we might have to extract the actual image tag
+		await runDockerCmd(dockerPath, ["tag", options.image, options.imageTag]);
```
penalosa: I'm a bit confused here—are we pulling a remote image and then re-tagging it locally?

fixtures/container-app/src/index.ts
emily-shen: yep, all of the second commit is just cleaning up the fixture for manual testing. e2e tests to follow shortly

packages/containers-shared/src/images.ts
emily-shen: i've tested this manually, just forgot to remove these commented out bits 😅

packages/containers-shared/src/images.ts
emily-shen: yep we're retagging to be in a consistent format for dev. i suppose we could use the original tag, but this keeps things consistent between built images and pulled images.

packages/containers-shared/src/images.ts
emily-shen: i'm not actually sure why we've wrapped this in a try catch in the first place, @CarmenPopoviciu is there any context here i'm missing?

packages/containers-shared/src/images.ts
emily-shen: it has to be either a cloudchamber registry or the containers+ one which is registry.cloudflare.com. 
i've removed that sentence there, the docs link should provide sufficient context.

packages/containers-shared/src/knobs.ts
emily-shen: this is in containers-shared, not sure how we can pull that in here from wrangler?

packages/containers-shared/src/utils.ts
emily-shen: i don't think that's possible. either way, the container's DO class name is probably a better identifier for the user so i've switched to using that

packages/containers-shared/src/utils.ts
emily-shen: linking to external docs feels a bit weird, and they come up if you just search `expose ports dockerfile`. but happy to if you feel strongly about it.

penalosa: Can we add a section in our docs about this?

packages/containers-shared/src/knobs.ts
penalosa: Ah, right... never mind then

packages/containers-shared/src/utils.ts
emily-shen: i think @mikenomitch  was going to do that already :) - i'll add a ticket to link to those docs when they're up

packages/wrangler/src/api/startDevWorker/ConfigController.ts
emily-shen: needs permissions to pull from registry - added to comment

packages/containers-shared/src/knobs.ts
emily-shen: updated comment

packages/containers-shared/src/images.ts
emily-shen: (i removed it hopefuly thats ok)

packages/wrangler/src/api/startDevWorker/ConfigController.ts
emily-shen: talked about this offline, will leave sorting out containers auth out of this PR.

penalosa (approved):

emily-shen:
tracking testing followup in DEVX-1980

packages/containers-shared/src/images.ts
```diff
@@ -22,3 +32,55 @@ export async function getDockerImageDigest(
...
+	}
+	await verifyDockerInstalled(dockerPath);
+	for (const options of containerOptions) {
+		if (isDockerfile(options.image)) {
+			await buildImage(dockerPath, options);
+		} else {
+			if (!isCloudflareRegistryLink(options.image)) {
+				throw new Error(
+					`Image "${options.image}" is a registry link but does not point to the Cloudflare container registry.\n` +
+						`To use an existing image from another repository, see https://developers.cloudflare.com/containers/image-management/#using-existing-images`
```
alycda, line 79: this url is currently 404

alycda: resolved: https://github.com/cloudflare/workers-sdk/pull/10764/commits/ece4a434da2563809253a3ce3edda30744f12c9a

PR #9605, opened 2025-06-13
clean up containers at the end of a dev session

emily-shen:
wip

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/containers-shared/src/utils.ts
```diff
@@ -29,6 +29,78 @@ export const runDockerCmd = async (
...
+				);
+			}
+		});
+		child.on("error", (err) => {
+			if (!errorHandled) {
+				errorHandled = true;
+				reject(new Error(`Docker command failed: ${err.message}`));
+			}
+		});
+	});
```
anonrig:
```suggestion
  const { promise, resolve, reject } = Promise.withResolvers();
	  child.on("close", (code) => {
	  if (code === 0) {
		  resolve();
	  } else if (!errorHandled) {
		  errorHandled = true;
		  reject(
			  new Error(
				  `Docker command exited with code: ${code}. stderr: ${stderr}`
			  )
		  );
	  }
  });
  child.on("error", (err) => {
	  if (!errorHandled) {
		  errorHandled = true;
		  reject(new Error(`Docker command failed: ${err.message}`));
	  }
  });
  await promise;
```

packages/miniflare/src/plugins/containers/service/index.ts
```diff
@@ -1,10 +1,14 @@
+import { platform } from "os";
```
anonrig: process.platform or navigator.platform will give you a similar result without the need to import node:os module (which might increase the boot time of this process)

packages/miniflare/src/plugins/containers/service/index.ts
```diff
@@ -79,4 +93,45 @@ export class ContainerController {
...
+
+			// Workerd should have stopped all containers, but clean up any in case. This might take a while.
+			await runDockerCmd(
+				this.#sharedOptions.dockerPath,
+				["stop", ...allContainerIds],
+				["inherit", "pipe", "pipe"]
+			);
+			await runDockerCmd(
+				this.#sharedOptions.dockerPath,
+				["rm", ...allContainerIds],
```
anonrig: If you run rm with --force, you don't need to stop the container first.

packages/containers-shared/src/utils.ts
emily-shen: i dom't thimk this is available in node 20 unfortunately

packages/containers-shared/src/build.ts
```diff
@@ -65,3 +69,49 @@ export function dockerBuild(
 		});
 	});
 }
+
+// TODO: this should also pull
+export async function buildAllContainers(
```
vicb: Add comments on the function in this file :)

Interesting things to comment:
- `verifyLocalDevSupported` why this does (does it needs to be in this function to start with)
- can the fn throw, when
- `dockerPath` -> `dockerBinPath`?

packages/containers-shared/src/build.ts
```diff
@@ -65,3 +69,49 @@ export function dockerBuild(
...
+	dockerPath: string,
+	logger: Logger,
+	containerOptions: ContainerDevOptions[]
+) {
+	await verifyLocalDevSupported(dockerPath);
+	logger.info("Loading container image(s)...");
```
vicb: Are those log for wrangler?
If yes they should probably be moved in the wrangler package.

If they are general purpose, I would argue that logging in `buildContainer` (with what's being built) would add more value.

packages/containers-shared/src/build.ts
```diff
@@ -65,3 +69,49 @@ export function dockerBuild(
...
+	const { buildCmd, dockerfile } = await constructBuildCommand({
+		tag: options.imageTag,
+		pathToDockerfile: options.image,
+		buildContext: options.imageBuildContext ?? path.dirname(options.image),
+		args: options.args,
+		platform: "linux/amd64",
+	});
+
+	await dockerBuild(dockerPath, { buildCmd, dockerfile });
+	await checkExposedPorts(dockerPath, options.imageTag);
```
vicb: Add a comment on why we are doing that (in the JSDoc of this fn)

packages/containers-shared/src/knobs.ts
```diff
@@ -1,4 +1,10 @@
+import { MF_DEV_CONTAINER_PREFIX } from "./registry";
+
 // default cloudflare managed registry, can be overriden with the env var - CLOUDFLARE_CONTAINER_REGISTRY
 export const getCloudflareContainerRegistry = () => {
 	return process.env.CLOUDFLARE_CONTAINER_REGISTRY ?? "registry.cloudflare.com";
 };
+
+export const getDevContainerImageName = (name: string, tag: string) => {
```
vicb, line 9: change `name` to answer name of what?

packages/containers-shared/src/types.ts
```diff
@@ -18,3 +18,13 @@ export type BuildArgs = {
...
 };
+
+/** build/pull agnostic container options */
+export type ContainerDevOptions = {
+	/** may be dockerfile or registry link */
+	image: string;
+	/** formatted as cloudflare-dev/workername-DOclassname:build-id */
+	imageTag: string;
+	imageBuildContext?: string;
+	args?: Record<string, string>;
```
vicb: What are those?

packages/containers-shared/src/utils.ts
```diff
@@ -29,7 +29,51 @@ export const runDockerCmd = async (
 	});
 };
 
-export const verifyDockerInstalled = async (dockerPath: string) => {
+export const runDockerCmdWithOutput = async (
```
vicb, line 32: Might be usefull to add when the returned promise resolves with what, same for reject

packages/containers-shared/src/build.ts
```diff
@@ -65,3 +69,49 @@ export function dockerBuild(
...
 	});
 }
+
+// TODO: this should also pull
+export async function buildAllContainers(
+	dockerPath: string,
+	logger: Logger,
+	containerOptions: ContainerDevOptions[]
+) {
+	await verifyLocalDevSupported(dockerPath);
```
vicb: Now that I have seen what this function does, I don't think it belong here?
I would rather move the call into wrangler.
(and it's not related to building, more a limitation of workerd IIRC)

packages/containers-shared/src/utils.ts
```diff
@@ -86,3 +130,49 @@ export const isDockerfile = (image: string): boolean => {
...
+	if (imageTags.size === 0) {
+		return true;
+	}
+
+	try {
+		// Find all containers (stopped and running) for each built image in parallel
+		const containerPromises = Array.from(imageTags).map(async (imageTag) => {
+			return await getContainers(dockerPath, imageTag);
+		});
+		const containerResults = await Promise.all(containerPromises);
```
vicb:
```suggestion
		const containerPromises = Array.from(imageTags).map(async (tag) => await getContainers(dockerPath, tag);
		const containerResults = await Promise.all(containerPromises);
```

vicb: But might be on a single line Promises are actually results? (`await`ed)

packages/containers-shared/src/utils.ts
```diff
@@ -86,3 +130,49 @@ export const isDockerfile = (image: string): boolean => {
...
+			["rm", "--force", ...allContainerIds],
+			["inherit", "pipe", "pipe"]
+		);
+		return true;
+	} catch (error) {
+		return false;
+	}
+};
+
+const getContainers = async (dockerPath: string, ancestorImage: string) => {
```
vicb: Add comment, what does this return?

packages/containers-shared/src/knobs.ts
```diff
@@ -1,4 +1,10 @@
+import { MF_DEV_CONTAINER_PREFIX } from "./registry";
+
 // default cloudflare managed registry, can be overriden with the env var - CLOUDFLARE_CONTAINER_REGISTRY
 export const getCloudflareContainerRegistry = () => {
```
vicb, line 4: not a bug deal but this PR mixes `function` and `export const = () =>...`

packages/containers-shared/src/utils.ts
```diff
@@ -86,3 +130,49 @@ export const isDockerfile = (image: string): boolean => {
...
+/**
+ * Kills and removes any containers which come from the given image tag
+ */
+export const cleanupContainers = async (
+	dockerPath: string,
+	imageTags: Set<string>
+) => {
+	if (imageTags.size === 0) {
+		return true;
+	}
```
vicb: not really necessary (but doesn't harm)

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -164,6 +173,15 @@ export class LocalRuntimeController extends RuntimeController {
 		remoteBindings: Record<string, Binding>;
 	} | null = null;
 
+	// Set of container images that have been seen in the current dev session.
+	// This is used to clean up containers at the end of the dev session.
+	#containerImagesSeen: Set<string> = new Set();
```
vicb: Looks like those are tag, add that to the name?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -310,6 +351,15 @@ export class LocalRuntimeController extends RuntimeController {
 		await this.#mf?.dispose();
 		this.#mf = undefined;
 
+		if (this.#containerImagesSeen.size > 0) {
```
vicb: If this test required?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -334,3 +384,34 @@ export class LocalRuntimeController extends RuntimeController {
 		this.emit("reloadComplete", data);
 	}
 }
+
+/**
+ * @returns Container options suitable for building or pulling images,
+ * with image tag set to well-known dev format.
```
vicb, line 393: "well-known dev format` -> @see <I think there is a fn computing that>

is what we call image tag throughout this PR  `name:tag`? is this the right Docker terminology?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -334,3 +384,34 @@ export class LocalRuntimeController extends RuntimeController {
...
+ * Undefined if containers are not enabled or configured.
```
vicb:
```suggestion
 * Undefined if containers are not enabled or not configured.
```

packages/wrangler/src/dev/hotkeys.ts
```diff
@@ -59,6 +61,31 @@ export default function registerDevHotKeys(
...
 			},
 		},
+		{
+			keys: ["r"],
+			// omitting the label means it won't be printed but is still enabled
+			// label: "rebuild container",
+			handler: async () => {
+				if (
+					!devEnv.config.latestConfig?.dev?.enableContainers ||
+					!devEnv.config.latestConfig?.containers
```
vicb: Should it test the length instead - maybe factor that in a function and re-use at the multiple places where you check?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -208,7 +209,8 @@ export interface ConfigBundle {
 	testScheduled: boolean;
 	enableContainers: boolean | undefined;
```
vicb: Do we still need `enableContainers` in Miniflare?
Passing an empty/undefined `containers` should work (and is what wrangler is doing I think)

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -208,7 +209,8 @@ export interface ConfigBundle {
...
 	dockerPath: string | undefined;
```
vicb: Do we still need dockerPath now that building the container has been removed?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -1372,26 +1372,31 @@ export async function buildMiniflareOptions(
...
+		"Build ID should be set if containers are defined and enabled"
+	);
 
-	const container = containers[className];
+	const container = config.containers.find((c) => c.class_name === DOClassName);
 
+	if (!container) {
+		return undefined;
+	}
 	return {
```
vicb, line 1415: nit: use ternary?

packages/containers-shared/src/build.ts
emily-shen: comments added 🫡 
leaving dockerPath as is for now, since its called that in all the docker helper functions and this PR touches enough things as it is 😅

packages/containers-shared/src/build.ts
emily-shen: i've reverted it just check docker is installed, since its useful to have a non-dev specific check for the other wrangler container commands. 

also renamed this function to better describe its purpose `prepareContainerImagesForDev`, since in a followup PR it will also pull images. This is very specific to dev.
Following on from that, I'd rather the validation happen here, since i imagine other clients like vite will call this.

packages/containers-shared/src/build.ts
emily-shen: good point, moved it out to wrangler. 
docker already logs what its building during build, so i'm just going to rely on that for now.

packages/containers-shared/src/knobs.ts
emily-shen: afaict, this is a docker convention to call that particular part the 'name'. anyway added a comment with more detail about what these values should be.

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
emily-shen: probably not, but i would prefer to gate anything that calls docker just in case.

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
emily-shen: unfortunately i think the right docker terminology is that tag refers to the bit after `:` _and_ the whole `repository/name:tag` string :/

packages/containers-shared/src/utils.ts
emily-shen: just reconsidered and there's really no point doing this in parallel since its basically instaneous, so i've simplified all of this out.

packages/wrangler/src/dev/miniflare.ts
emily-shen: nope, good catch

packages/containers-shared/src/build.ts
```diff
@@ -65,3 +69,58 @@ export function dockerBuild(
...
+ * such as checking if the Docker CLI is installed, and if the container images
+ * expose any ports.
+ */
+export async function prepareContainerImagesForDev(
+	dockerPath: string,
+	containerOptions: ContainerDevOptions[]
+) {
+	if (process.platform === "win32") {
+		throw new Error(
+			"Local development with containers is currently not supported on Windows. You should use WSL instead. You can also set `enable_containers` to false if you do not need to develop the container part of your application."
```
CarmenPopoviciu:
```suggestion
			"Local development with containers is currently not supported on Windows. You should use WSL instead. You can also set `enable_containers` to false if you do not need to develop the container as part of your application."
```

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
CarmenPopoviciu: why was this resolved without applying the suggestion?

packages/containers-shared/src/build.ts
vicb: > Following on from that, I'd rather the validation happen here, since i imagine other clients like vite will call this.

Typically you could add that in a "pre-requisite" in the JSDoc - but you have more context than I do and know what's best

PR #9620, opened 2025-06-16
CC-5462: Simplify containers images list output

gpanders:
Fixes CC-5462.

Use a simple table (one row for each entry) instead of a fancy table with box characters (that is hard to read and breaks when text wraps). Print one row for each name:tag combination.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: UI only change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: UI only change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: the cloudchamber images feature does not exist in v3


packages/wrangler/src/cloudchamber/images/list.ts
```diff
@@ -144,15 +152,19 @@ async function ListTags(
...
-			})
-			.flat();
-		logger.table(rows);
+		const rows = responses.map((r) => r.tags.map((t) => [r.name, t])).flat();
+		const headers = ["REPOSITORY", "TAG"];
+		const widths = new Array(headers.length).fill(0);
+
+		// Find the maximum length of each column (except for the last)
+		for (let i = 0; i < widths.length - 1; i++) {
+			widths[i] = rows.map((r) => r[i].length).reduce((a, b) => Math.max(a, b), headers[i].length);
```
vicb: Is this equivalent to:
```suggestion
			widths[i] = Math.max(headers[i].length, ...rows.map((r) => r[i].length));
```

packages/wrangler/src/cloudchamber/images/list.ts
```diff
@@ -144,15 +152,19 @@ async function ListTags(
...
-		const rows = responses
-			.map((resp) => {
-				return {
-					REPOSITORY: resp.name,
-					TAG: resp.tags.join(" "),
-				};
-			})
-			.flat();
-		logger.table(rows);
+		const rows = responses.map((r) => r.tags.map((t) => [r.name, t])).flat();
```
vicb: nit:
```suggestion
		const rows = responses.flatmap((r) => r.tags.map((t) => [r.name, t]));
```

packages/wrangler/src/cloudchamber/images/list.ts
gpanders: Almost, but using the spread operator can fail or return incorrect results if the array is too large (according to the [MDN reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max#getting_the_maximum_element_of_an_array)). `reduce` doesn’t have that problem.

It’s unlikely that this array will ever be so large that that’s an issue in practice, but given that the array is technically unbounded in size I went with the safer option.

packages/wrangler/src/cloudchamber/images/list.ts
gpanders: Nice suggestion, thanks! Applied.

CarmenPopoviciu (approved):
LGTM

PR #9626, opened 2025-06-17
Support version upload for Python Workers

penalosa:
Previously, Python Workers couldn't be used with `version upload` because there was a difference in module type detection between `deploy` and `version upload`. This PR fixes that bug and unifies the two code paths.

After banging my head against the wall trying to figure out how to nicely test this, this PR _also_ introduces a test helper for making assertions against the content of HTTP requests that Wrangler makes. This will work in tandem with MSW (which provides the mock APIs). The issue with our current setup (MSW interceptors making assertions on the content of requests) is that 1) MSW strongly [recommends against this style](https://mswjs.io/docs/best-practices/avoid-request-assertions/), and 2) probably because MSW recommends against it, it works pretty badly (in particular, stack traces and mismatches in assertions show up strangely)

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [x] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/helpers/assert-request.ts
+++ b/packages/wrangler/src/__tests__/helpers/assert-request.ts
@@ -1,0 +1,38 @@
+import type { mockConsoleMethods } from "./mock-console";
+
+/**
+ * Assert that Wrangler has made a request matching a certain pattern. Unlike MSW (which mocks the return value of the request),
+ * this helper asserts that a request has been made. It should be used in combination with MSW—MSW providing the mock API, and this helper being used
+ * to make sure the right data is sent to the mock API.
+ */
+export function makeRequestAsserter(
+	console: ReturnType<typeof mockConsoleMethods>
+) {
+	beforeEach(() => {
+		vi.stubEnv("WRANGLER_LOG", "debug");
+		vi.stubEnv("WRANGLER_LOG_SANITIZE", "false");
+	});
+	return function assertRequest(
+		url: RegExp,
+		{ method, body }: { method?: string; body?: RegExp }
+	) {
+		const startLine = console.debug.match(
+			new RegExp(`-- START CF API REQUEST: ${method} ` + url.source)
+		);
+		assert(startLine !== null, "Request not made by Wrangler");
+		const requestDetails = console.debug
+			.slice(startLine.index)
+			.match(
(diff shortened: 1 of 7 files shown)


packages/wrangler/src/__tests__/helpers/write-worker-source.ts
```diff
@@ -22,10 +22,15 @@ export function writeWorkerSource({
           return new Response('Hello' + foo);
         },
       };`
-			: `import { foo } from "./another";
+			: type === "sw"
+				? `import { foo } from "./another";
       addEventListener('fetch', event => {
         event.respondWith(new Response('Hello' + foo));
       })`
+				: `from js import Response
```
dario-piotrowicz: The ternary here is a bit unreadable in my opinion 😕 

The original one was already ugly but adding the new nested ternary makes it worse

Not a huge deal and not something that necessarily needs amending here, I just figured I'd bring it up

packages/wrangler/src/__tests__/helpers/assert-request.ts
```diff
@@ -0,0 +1,38 @@
+import type { mockConsoleMethods } from "./mock-console";
+
+/**
+ * Assert that Wrangler has made a request matching a certain pattern. Unlike MSW (which mocks the return value of the request),
+ * this helper asserts that a request has been made. It should be used in combination with MSW—MSW providing the mock API, and this helper being used
+ * to make sure the right data is sent to the mock API.
+ */
```
dario-piotrowicz: > asserts that a request has been made

If I understand correctly the code below this is done by checking the `console.debug` and making sure that it contains lines that indicate the request right?

I think that this detail should be included in the comment?

dario-piotrowicz: Also I think I would rename `makeRequestAsserter` and `assertRequest` to something like `makeCfApiRequestAsserter` and `assertCfApiRequest` or something like that to clarify that this is specific to cf api requests 🤔

PR #9638, opened 2025-06-17
Add test for cloudchamber buildAndMaybePush

IRCody:
Fixes n/a

Tests the code path where push flag is set but the image exists already in the remote so we do not want to push and instead want to untag the image.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Non e2e test change only
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: test change only
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  test doesn't exist in v3


CarmenPopoviciu (approved):
thank you so much for adding this test ♥️

PR #9640, opened 2025-06-17
add debug logs for workerd

emily-shen:
Might be useful again, just dumps the workerd config into a file for debugging. Does overwrite the file each time we restart workerd, so when the worker or wrangler config changes. 

@danlapid does this help? any requests while i'm here?

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: trivial miniflare change change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: updated miniflare readme but mostly for internal usage 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not wrangler


packages/miniflare/src/runtime/config/index.ts
```diff
@@ -47,6 +48,9 @@ function encodeCapnpStruct(obj: any, struct: Struct) {
 }
 
 export function serializeConfig(config: Config): Buffer {
+	if (process.env.MINIFLARE_WORKERD_CONFIG_DEBUG === "true") {
+		writeFileSync("workerd-config.json", JSON.stringify(config, null, 2));
+	}
```
petebacondarwin: Can I propose that instead of a boolean you can provide a path to the output file. E.g.

```suggestion
	if (process.env.MINIFLARE_WORKERD_CONFIG_DEBUG_PATH) {
		writeFileSync(process.env.MINIFLARE_WORKERD_CONFIG_DEBUG_PATH, JSON.stringify(config, null, 2));
	}
```

packages/miniflare/src/runtime/config/index.ts
```diff
@@ -47,6 +48,10 @@ function encodeCapnpStruct(obj: any, struct: Struct) {
 }
 
 export function serializeConfig(config: Config): Buffer {
+	const debugPath = process.env.MINIFLARE_WORKERD_CONFIG_DEBUG;
+	if (debugPath) {
+		writeFileSync(debugPath, JSON.stringify(config, null, 2));
```
petebacondarwin, line 53: Just to be clear, if the file path is relative then it will be local to the current working directory.
Is that OK?

petebacondarwin (approved):

packages/miniflare/src/runtime/config/index.ts
emily-shen: would you expect it to be relative to the wrangler config? 

as its just for debugging, and i'm not sure we have the wrangler config path here in miniflare, i'm inclined to just let it be 😅

petebacondarwin: yeah its fine. just wanted to make sure this was expected

PR #9641, opened 2025-06-17
Update buildAndMaybePush to use more robust method to find local digests

IRCody:
Older versions of docker sometimes will report the digest as "<none>" which would break with the previous implementation. This implementation should correctly grab the values.

_Describe your change..._

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Modifying existing behavior that is not e2e tested afaik
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Makes existing method more robust but doesn't change user ux
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  feature doesn't exist in v3


packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -126,18 +126,38 @@ export async function buildAndMaybePush(
...
 			try {
+				// we don't try to parse repoDigests until here because we
+				// don't want to fail on parse errors if we aren't pushing anyways
+				// as it's only relevant when pushing.
+				const parsedDigests = JSON.parse(repoDigests);
+				if (!Array.isArray(parsedDigests)) {
+					// If it's not the format we expect, fall back to pushing
+					// since it's annoying but safe.
+					throw new Error(
+						"Expected RepoDigests from docker inspect to be an array but it wasn't"
```
CarmenPopoviciu:
```suggestion
						`Expected "RepoDigests" from docker inspect to be an array but got JSON.stringify(${parsedDigests})`
```

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -146,14 +166,20 @@ export async function buildAndMaybePush(
 					`Untagging built image: ${imageTag} since there was no change.`
 				);
 				await runDockerCmd(pathToDocker, ["image", "rm", imageTag]);
-				return "";
+				return { image: digests[0], pushed: false };
 			} catch (error) {
 				logger.log(`Image does not exist remotely, pushing: ${imageTag}`);
+				if (error instanceof Error) {
+					logger.debug(
```
CarmenPopoviciu, line 181: how important are these errors for users? are we sure `debug` is the log level we want here? This is not the default log level in wrangler, so it might be a lot of folks will miss out on this info, unless they specifically set the log level to debug. Totally your decision, just thinking out loud

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -146,14 +166,20 @@ export async function buildAndMaybePush(
...
+						`Checking for local image ${imageTag} failed with error: ${error.message}`
+					);
+				}
 
 				await runDockerCmd(pathToDocker, ["push", imageTag]);
+				pushed = true;
 			}
 		}
-		return imageTag;
+		return { image: imageTag, pushed: pushed };
```
CarmenPopoviciu, line 190:
```suggestion
		return { image: imageTag, pushed };
```

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -314,7 +314,12 @@ function sortObjectRecursive<T = Record<string | number, unknown>>(
 }
 
 export async function apply(
-	args: { skipDefaults: boolean | undefined; json: boolean; env?: string },
+	args: {
+		skipDefaults: boolean | undefined;
+		json: boolean;
+		env?: string;
+		pushed?: boolean;
```
CarmenPopoviciu: without having any context, `pushed` doesn't really convey much meaning to me. What state of the world does this arg reflect? Is it that the image already exists? Is there a more meaningful naming we can use here? Or is this some docker-specific nomenclature I'm not aware of?

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -126,18 +126,38 @@ export async function buildAndMaybePush(
...
 		});
-
+		let pushed = false;
 		if (push) {
 			await dockerLoginManagedRegistry(pathToDocker);
 			try {
+				// we don't try to parse repoDigests until here because we
+				// don't want to fail on parse errors if we aren't pushing anyways
+				// as it's only relevant when pushing.
+				const parsedDigests = JSON.parse(repoDigests);
```
CarmenPopoviciu, line 143: are `repoDigests` the same as Docker image digests? If so, can we reference the Docker docs, or add a one liner from the docs that explains what these are? I know I'm being very specific, but I think it can help contributors (external and internal) get context of this code more easily

```
A Docker image digest is a unique, cryptographic identifier (SHA-256 hash) 
representing the content of a Docker image. Unlike tags, which can be reused 
or changed, a digest is immutable and ensures that the exact same image is 
pulled every time. This guarantees consistency across different environments 
and deployments.
```
https://docs.docker.com/dhi/core-concepts/digests/

packages/wrangler/src/cloudchamber/apply.ts
IRCody: It's meant to be whether the image was pushed to the registry before this apply call (which in this case really means by deploy).

It's not specific to docker so if the poor naming is all me, but fortunately that means it can be changed easily!

The significance here is that if it WAS pushed we need to update the image in the apply call and if it wasn't we don't want to as it might create an unnecessary deploy on our side.

So maybe we should change it to something like imageUpdateRequired?

packages/wrangler/src/cloudchamber/build.ts
IRCody: That's a good idea. Will paste the snipper into this comment and link to the docs.

packages/wrangler/src/cloudchamber/build.ts
IRCody: What log level would people be using if they wanted to actively debug something?

I didn't want this to print on every run since it can be part of normal operation but wanted a way for the user to access it if they were debugging something.

CarmenPopoviciu: > What log level would people be using if they wanted to actively debug something?

`debug`

> I didn't want this to print on every run since it can be part of normal operation but wanted a way for the user to access it if they were debugging something.

then you did precisely the right API. I was merely checking that the log level was intentional, and it sounds like it was. Carry on ♥️

packages/wrangler/src/cloudchamber/apply.ts
CarmenPopoviciu: thank you so much for taking the time to explain things. I think `imageUpdateRequired` is a good alternative. Totes happy to go with that if you are 🙇

CarmenPopoviciu:
left a few nit comments, but LGTM otherwise.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -146,14 +174,20 @@ export async function buildAndMaybePush(
 					`Untagging built image: ${imageTag} since there was no change.`
 				);
 				await runDockerCmd(pathToDocker, ["image", "rm", imageTag]);
-				return "";
+				return { image: digests[0], pushed: false };
 			} catch (error) {
 				logger.log(`Image does not exist remotely, pushing: ${imageTag}`);
+				if (error instanceof Error) {
```
dario-piotrowicz, line 180: should the `logger.log` call be removed? (or made conditional based on the error?)

dario-piotrowicz (approved):
I'm not too familiar with this part of the codebase, but the code looks good to me 🙂 

(I am also trusting Carmen's comment https://github.com/cloudflare/workers-sdk/pull/9641#issuecomment-2999822386)

CarmenPopoviciu (approved):
thx for addressing all my comment @IRCody ♥️

packages/wrangler/src/cloudchamber/build.ts
IRCody: I don't think so but here is my reasoning: this is informing the user of something that is happening. 

If we fail to inspect the remote and verify the image exists there for any reason (all of which throw an error) we fall back to pushing the image since that is always a safe thing to try. we added checking the remote for ux reasons because re-pushing constantly for non-changing images is annoying but functionally it should always be fine.

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -18,14 +18,17 @@ export async function maybeBuildContainer(
 	imageTag: string,
 	dryRun: boolean,
 	pathToDocker: string
-) {
+): Promise<{ image: string; pushed: boolean }> {
 	try {
 		if (
 			!isDockerfile(
 				containerConfig.image ?? containerConfig.configuration?.image
```
gpanders, line 25: The type checker is failing to catch this, but the type of `containerConfig.image ?? containerConfig.configuration?.image` is `string | undefined`, but `isDockerfile` only accepts a `string`. This is causing errors at runtime.

gpanders, line 25: `containerConfig.image` has type `string` so the `??` isn't necessary

```suggestion
				containerConfig.image
```

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -18,14 +18,17 @@ export async function maybeBuildContainer(
...
 			)
 		) {
-			return containerConfig.image ?? containerConfig.configuration?.image;
+			return {
+				image: containerConfig.image ?? containerConfig.configuration?.image,
```
gpanders, line 29: The type of `containerConfig.image` is `string`, not `string | undefined`, so should always be defined.

If it's possible for this value to be undefined we should update its type so that we get proper type checking.

```suggestion
				image: containerConfig.image,
```

packages/wrangler/src/cloudchamber/deploy.ts
IRCody: For this should it be:
 `containerconfig.image === "" ? containerConfig.configuration?.image : containerConfig.image;

packages/wrangler/src/cloudchamber/deploy.ts
gpanders: See https://github.com/cloudflare/workers-sdk/pull/9641#discussion_r2164716002

packages/wrangler/src/cloudchamber/deploy.ts
emily-shen: looks like containerConfig.image === "" throws a validation error, so that shouldn't be possible by the time we get here.

 also yeah the nullish coalescing thingy should be unnecessary anymore, that was me earlier today 😅 
we should have a diff type for wrangler config and the api (since one expects containers.image and one containers.configuration.image 🫠 )  to make this sort of thing clearer, but we'll follow up on that later.

PR #9647, opened 2025-06-18
Refactor preview mode and ensure compatibility with Vite 7

jamesopstad:

In Vite 7, the `applyToEnvironment` hook is called in preview mode. This is now accounted for to ensure compatibility. I have refactored the resolved plugin config so that it now returns a config for preview mode.

Additionally, the `root` property of the module runner options has been removed as it is not needed and is dropped in Vite 7.

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: already covered by tests
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/playground/node-compat/__tests__/worker-postgres/postgres.spec.ts
```diff
@@ -12,6 +12,6 @@ test.runIf(!process.env.CI)(
 	"should be able to use pg library to send a query",
 	async () => {
 		const result = await getJsonResponse("/send-query");
-		expect(result!.id).toEqual("1");
+		expect(result!.id).toEqual("21");
```
jamesopstad, line 15: This test was consistently failing with this result so I assume the data in the database has changed.

petebacondarwin: Really we should make a fake pg instance somewhere that we host ourselves for this (maybe even locally!) to avoid this sort of problem. This test also often fails because we get rate limited 😢

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -81,7 +84,21 @@ export interface WorkerPluginConfig extends BasePluginConfig {
...
+export type ResolvedPluginConfig<
+	T extends "assets-only" | "workers" | "preview" =
+		| "assets-only"
+		| "workers"
+		| "preview",
+> = { type: T } & (
+	| AssetsOnlyResolvedConfig
+	| WorkersResolvedConfig
+	| PreviewResolvedConfig
+);
```
petebacondarwin: Is this parameterized type really necessary?

Can't you just use

```suggestion
export type ResolvedPluginConfig = AssetsOnlyResolvedConfig | WorkersResolvedConfig | PreviewResolvedConfig;
```

and then where you narrow elsewhere in the code, instead of:

```ts
resolvedPluginConfig: ResolvedPluginConfig<"assets-only" | "workers">,
```

you can just do:

```ts
resolvedPluginConfig: AssetsONlyResolvedConfig | WorkersResolvedConfig;
```

packages/vite-plugin-cloudflare/src/dev-vars.ts
```diff
@@ -0,0 +1,55 @@
+import * as fs from "node:fs";
+import * as path from "node:path";
+import type { ResolvedPluginConfig } from "./plugin-config";
+
+/**
+ * Gets the content of the `.dev.vars` target file
+ *
+ * Note: This resolves the .dev.vars file path following the same logic
+ *       as `loadDotEnv` in `/packages/wrangler/src/config/index.ts`
```
petebacondarwin, line 12: Can you update the comment in the referenced file https://github.com/cloudflare/workers-sdk/blob/ac0f1ad4c65afdeb1ad6cfc94be7833b08e2d9d5/packages/wrangler/src/config/index.ts#L262 to point back to this one.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -95,17 +96,16 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
 				resolvedPluginConfig = resolvePluginConfig(
 					pluginConfig,
 					userConfig,
 					env
 				);
 
+				if (resolvedPluginConfig.type === "preview") {
+					return { appType: "custom" };
+				}
+
```
petebacondarwin, line 108: Good call to move this down.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -720,11 +722,7 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 		// Plugin that handles Node.js Async Local Storage (ALS) compatibility support for Vite Environments that are hosted in Cloudflare Workers.
 		{
 			name: "vite-plugin-cloudflare:nodejs-als",
-			apply(_config, env) {
-				// Skip this whole plugin if we are in preview mode
-				return !env.isPreview;
-			},
-			configEnvironment(name, config) {
+			configEnvironment(name) {
```
petebacondarwin, line 725: Can we config eslint to pick up on these non-used params and imports?

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -190,17 +218,36 @@ export function resolvePluginConfig(
...
-		cloudflareEnv,
-		experimental,
 	};
 }
+
+export function assertIsNotPreview(
+	resolvedPluginConfig: ResolvedPluginConfig
+): asserts resolvedPluginConfig is ResolvedPluginConfig<
+	"assets-only" | "workers"
+> {
```
petebacondarwin: If you used the type suggested above this declaration simplifies to:

```suggestion
export function assertIsNotPreview(
	resolvedPluginConfig: ResolvedPluginConfig
): asserts Exclude<ResolvedPluginConfig, PreviewResolvedConfig> {
```

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -190,17 +218,36 @@ export function resolvePluginConfig(
...
+	assert(
+		resolvedPluginConfig.type !== "preview",
+		`Expected "assets-only" or "workers" plugin config`
+	);
+}
+
+export function assertIsPreview(
+	resolvedPluginConfig: ResolvedPluginConfig
+): asserts resolvedPluginConfig is ResolvedPluginConfig<"preview"> {
```
petebacondarwin:
```suggestion
): asserts resolvedPluginConfig is PreviewResolvedConfig {
```

petebacondarwin (approved):
Nice bit of refactor. Some non-blocking suggestions about the types.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: Yeah, I did actually have exactly what you've suggested originally. It started to feel quite laborious to always import and specify multiple types so it felt quite nice to use the single `ResolvedPluginConfig` type. The benefit is probably quite marginal though so I can change it back if you prefer.

petebacondarwin: Could have a `type NonPreviewResolvedConfig = Exclude<ResolvedPluginConfig, PreviewResolvedConfig>` type?
It seems that preview and non-preview are the only common ones?

packages/vite-plugin-cloudflare/src/dev-vars.ts
jamesopstad: https://github.com/cloudflare/workers-sdk/pull/9647/commits/3ec5db0903653cbd3ece1551ddd866ddde9ff048

petebacondarwin:
Ugh! Conflicts

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: I've added `"noUnusedLocals": true` and `"noUnusedParameters": true` to the TSConfig.

packages/vite-plugin-cloudflare/src/plugin-config.ts
jamesopstad: https://github.com/cloudflare/workers-sdk/pull/9647/commits/07daee354ff605abf387ec3682f7c7d01b781377

PR #9652, opened 2025-06-18
re-enable vite-plugin remote bindings e2e tests for dev mode

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1953

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: tests-only change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: update related to a non-v3 feature

### DIFF
--- a/packages/vite-plugin-cloudflare/e2e/basic.test.ts
+++ b/packages/vite-plugin-cloudflare/e2e/basic.test.ts
@@ -1,8 +1,13 @@
 import { rm, writeFile } from "node:fs/promises";
 import { describe, test } from "vitest";
-import { fetchJson, runLongLived, seed, waitForReady } from "./helpers.js";
+import {
+	fetchJson,
+	isBuildAndPreviewOnWindows,
+	runLongLived,
+	seed,
+	waitForReady,
+} from "./helpers.js";
 
-const isWindows = process.platform === "win32";
 const packageManagers = ["pnpm", "npm", "yarn"] as const;
 const commands = ["dev", "buildAndPreview"] as const;
 
@@ -11,7 +16,7 @@ describe("basic e2e tests", () => {
 		const projectPath = seed("basic", pm);
 
 		describe.each(commands)('with "%s" command', (command) => {
-			test.skipIf(isWindows && command === "buildAndPreview")(
+			test.skipIf(isBuildAndPreviewOnWindows(command))(
 				"can serve a Worker that uses a Node.js API (crypto)",
 				async ({ expect }) => {
 					const proc = await runLongLived(pm, command, projectPath);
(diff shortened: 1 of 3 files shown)


petebacondarwin (commented):
I think we should also skip remote bindings tests if there is no CLOUDFLARE_API_TOKEN. That way we won't block forked PRs.

dario-piotrowicz:
> I think we should also skip remote bindings tests if there is no CLOUDFLARE_API_TOKEN. That way we won't block forked PRs.

We are: https://github.com/cloudflare/workers-sdk/pull/9652/files#diff-d9b32f69a202cbf69dc77c2777a78758abe37967f4015456c65d005d6117c33dR21

Are we not? 🤔

packages/vite-plugin-cloudflare/e2e/basic.test.ts
```diff
@@ -171,7 +176,7 @@ describe("basic e2e tests", () => {
 			describe.skipIf(
 				!process.env.CLOUDFLARE_ACCOUNT_ID || !process.env.CLOUDFLARE_API_TOKEN
 			)("Workers AI", () => {
-				test.skipIf(command === "buildAndPreview")(
+				test.skipIf(isBuildAndPreviewOnWindows(command))(
```
petebacondarwin, line 179: Why skip on Windows?

petebacondarwin (approved):
Approving with the question about skipping the AI tests on Windows, where we were previously.
Also I am worried that this will make our CI even more flakey, since remote mode has been problematic.

packages/vite-plugin-cloudflare/e2e/basic.test.ts
dario-piotrowicz: Currently this skips the test if the command is `buildAndPreview` on all OSes, but this is only problematic on Windows machines (the CI is happy on linux and mac), so this reduces to skipped tests to only the problematic ones

do you see an issue with this?

dario-piotrowicz:
> Also I am worried that this will make our CI even more flakey, since remote mode has been problematic.

That's totally fair... if you prefer we can hold off on this change

petebacondarwin:
> > Also I am worried that this will make our CI even more flakey, since remote mode has been problematic.
> 
> That's totally fair... if you prefer we can hold off on this change

Nah! Let's do it.

packages/vite-plugin-cloudflare/e2e/basic.test.ts
petebacondarwin: Sorry - my mistake

dario-piotrowicz: ah ok thanks for confirming 😄🙏 (I just wanted to make sure you didn't know about some other issues/flakiness that this change could cause)

PR #9661, opened 2025-06-18
Router worker performs free tier limiting

matthewdavidrodgers:
WC-3584

(supersedes https://github.com/cloudflare/workers-sdk/pull/9360 as this is not from a fork)

The Router worker now limits requests when free users are over their daily limit. This decision is made via the new optional param binding (EYEBALL_ROUTER_CONFIG) that is passed in. The Router worker then denies all requests to the user worker when that param indicates limiting should occur.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23008
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  not a wrangler change


packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -95,6 +123,66 @@ export default {
...
+				analytics.setData({ dispatchtype: DISPATCH_TYPE.WORKER });
+				userWorkerInvocation = true;
+				return env.JAEGER.enterSpan("dispatch_worker", async (span) => {
+					span.setTags({
+						hasUserWorker: true,
+						asset: asset,
+						dispatchType: DISPATCH_TYPE.WORKER,
+					});
+
+					if (shouldBlockNonImageResponse) {
```
petebacondarwin, line 133: We can actually move the computation of this boolean in here too I think, which would save us having to compute it if we never need to hit the user worker.

matthewdavidrodgers: yes, although the only difference that would change is that the analytics for `abuseMitigationURLHost` would not get set for requests that don't go to the asset worker. i think for the most part that's fine, however, since that's all we're interested in. making that change

PR #9664, opened 2025-06-18
Remove confirmation dialog for cloudchamber apply when not interactive

IRCody:
Prior to this the dialog was getting printed but auto-confirmed to yes which is confusing to users.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: ux change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: ux change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  ux change for newer versions only

### DIFF
--- a/packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
+++ b/packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
@@ -135,9 +135,6 @@ describe("cloudchamber apply", () => {
 			│   [containers.constraints]
 			│   tier = 2
 			│
-			├ Do you want to apply these changes?
-			│ yes
-			│
 			│
 			│  SUCCESS  Created application my-container-app (Application ID: abc)
 			│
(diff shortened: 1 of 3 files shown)


gpanders (approved):

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -135,9 +135,6 @@ describe("cloudchamber apply", () => {
 			│   [containers.constraints]
 			│   tier = 2
 			│
-			├ Do you want to apply these changes?
```
CarmenPopoviciu, line 138: it seems that all these tests test the non-interactive mode. Is there a reason why interactivity isn't tested for this cmd?

IRCody: Just because they weren't added when apply was originally added. Good point they should be added (but maybe not as part of this PR?).

CarmenPopoviciu: 👍 ...for posterity... Cody added an internal ticket to track adding tests for interactiviy. Will therefore not block this PR on it

PR #9666, opened 2025-06-19
Add default value for container apps if no name is present

IRCody:
Default container app names to:
	WorkerName-ClassName[-EnvName]

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: no e2e changes, config validation only
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: will be documented as part of container beta docs.
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  config option not in v3.


skepticfx (approved):

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,126 @@ const validateBindingArray =
...
+			return true;
+		}
 
-	if (!Array.isArray(value)) {
-		diagnostics.errors.push(
-			`"containers" should be an array, but got ${JSON.stringify(value)}`
-		);
-		return false;
-	}
+		if (typeof value !== "object") {
```
CarmenPopoviciu: I'm a bit confused. we seem to be validating "containers" to be both an object or an array. what is the type of this config key?

IRCody: An array is a special type of object I think but you're right we don't need both checks.

CarmenPopoviciu: right...I was more thinking from a user perspective. Ultimately the error should make it clear for users what the expected type is, and how they can fix their error. So if "containers" is an Array, I think we should check for that and surface the err as such

IRCody: Removed this duplicate check and rebased onto main.

Thanks for catching this.

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
-		if (
-			"rollout_step_percentage" in containerAppOptional &&
-			containerAppOptional.rollout_step_percentage !== undefined
-		) {
+		for (const containerApp of value) {
+			const containerAppOptional =
+				containerApp as Partial<CreateApplicationRequest> & {
+					image?: string | undefined;
+					class_name?: string | undefined;
+				};
```
emily-shen: why recast this type? i think we want to leave the type unknown during validation because can't assume it actually adheres to this type until we have validated.

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
+
+			if (!isRequiredProperty(containerAppOptional, "name", "string")) {
```
emily-shen: since this isn't required anymore, probably just separately check the type like this:
```
	if (!isOptionalProperty(value, "name", "string")) {
		diagnostics.errors.push(
			`the field "name", when present, should be a string.`
		);
	}
```
and then 
```
if(!containerAppOptional.name){
// same as now
}
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
+					diagnostics.errors.push(
+						`"Must have either a top level name or containers.name defined"`
+					);
+				} else {
+					// if there is worker name defined but no name for this container app default to:
+					// worker_name-class_name[-envName].
+					let name = `${topLevelName}-${containerAppOptional.class_name}`;
+					// config is undefined when we are at the top level instead of in a named env
+					name += config === undefined ? "" : `-${envName}`;
+					containerAppOptional.name = name.toLowerCase().replace(/ /g, "");
```
emily-shen: what about replacing spaces with `-` instead, might be a bit more legible? up to you

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
+					containerAppOptional.rollout_step_percentage < 25
+				) {
+					diagnostics.errors.push(
+						`"containers.rollout_step_percentage" should be a number between 25 and 100, but got ${containerAppOptional.rollout_step_percentage}`
+					);
+				}
 			}
-		}
 
-		if (
```
emily-shen, line 2426: something like this should do that check for you
```
validateOptionalProperty(
			diagnostics,
			field,
			"rollout_kind",
			containerApp.rollout_kind,
			"string",
		["full_auto", "full_manual", "none"]
		) 
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
+				}
+
+				containerAppOptional.configuration ??= {
+					image: containerAppOptional.image,
+				};
+				containerAppOptional.configuration.image = containerAppOptional.image;
+				delete containerAppOptional["image"];
+			}
+
+			if (!("configuration" in containerAppOptional)) {
```
emily-shen, line 2477: think it would make sense to move the check for containers.configuration above the one for containers.configuration.image/ containers.image

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
+				diagnostics.errors.push(`"containers.configuration" should be defined`);
+			} else if (Array.isArray(containerAppOptional.configuration)) {
```
emily-shen, line 2479: should check its not an object
```suggestion
			} else if (typeof containerAppOptional.configuration !== "object" || typeof containerAppOptional.configuration === null || array check etc.) {
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
+					// if there is worker name defined but no name for this container app default to:
+					// worker_name-class_name[-envName].
+					let name = `${topLevelName}-${containerAppOptional.class_name}`;
+					// config is undefined when we are at the top level instead of in a named env
+					name += config === undefined ? "" : `-${envName}`;
+					containerAppOptional.name = name.toLowerCase().replace(/ /g, "");
+				}
+			}
+
 			if (
```
emily-shen: github has rendered the diff horribly here so i can't add a suggestion, but i think the conditional can be condensed into
`!isOptionalProperty(containerApp, "rollout_step_percentage", "number") && (containerApp.rollout_step_percentage > 100 || containerApp.rollout_step_percentage < 25)`

emily-shen (commented):
sorry the validation "helpers" are so unhelpful 🫠 

we should probably do way more validation here (e.g. that the image is a valid dockerfile path or registry link), check all the other config fields, but we shouldn't block on that so i made a ticket
[DEVX-2029](https://jira.cfdata.org/browse/DEVX-2029)

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,114 +2366,119 @@ const validateBindingArray =
...
+
+			if (!isRequiredProperty(containerAppOptional, "name", "string")) {
+				if (!topLevelName) {
+					diagnostics.errors.push(
+						`"Must have either a top level name or containers.name defined"`
+					);
+				} else {
+					// if there is worker name defined but no name for this container app default to:
+					// worker_name-class_name[-envName].
+					let name = `${topLevelName}-${containerAppOptional.class_name}`;
```
emily-shen, line 2413: we also need a check that container.class_name is defined

packages/wrangler/src/config/validation.ts
IRCody: Good call. Switching to dashes in the next update.

packages/wrangler/src/config/validation.ts
IRCody: it should be an object right?

packages/wrangler/src/config/validation.ts
IRCody: None of this logic was changed by this PR. The diff is annoying because everything got indented as we are now returning a validatorFn instead of being one so it looks crazy. I may try to reorganize it/comment it as part of this PR though since it's throwing a crazy diff anyways.

That said I don't think we can put validating configuration before the image logic as the image logic is, in most cases, creating the configuration 😞 .

packages/wrangler/src/config/validation.ts
IRCody: Removed this.

packages/wrangler/src/config/validation.ts
IRCody: Added this check. as well so not it errors if either topLevel name or class_name are not defined.

packages/wrangler/src/config/validation.ts
emily-shen: ah sorry about that, yeah I was confused by the diff! feel free to punt on this here and do any cleanup in a followup

packages/wrangler/src/config/validation.ts
```diff
@@ -2366,61 +2366,95 @@ const validateBindingArray =
...
-	for (const containerApp of value) {
-		const containerAppOptional =
-			containerApp as Partial<CreateApplicationRequest> & {
-				image?: string | undefined;
-			};
-		if (!isRequiredProperty(containerAppOptional, "name", "string")) {
+		if (!Array.isArray(value)) {
 			diagnostics.errors.push(
-				`"containers.name" should be defined and a string`
+				`"containers" should be an array, but got ${JSON.stringify(value)}`
```
CarmenPopoviciu:
```suggestion
				`"containers" field should be an array, but got ${JSON.stringify(value)}`
```

packages/wrangler/src/config/validation.ts
emily-shen: sorry my comment text was confusing - i meant we want to hit this error if it _isn't_ an object, not just if it is an array - ie if its a string or something. not a huge deal for this PR at all!

packages/wrangler/src/config/validation.ts
```diff
@@ -2376,59 +2376,95 @@ const validateBindingArray =
...
 
-		if (
-			"rollout_step_percentage" in containerAppOptional &&
-			containerAppOptional.rollout_step_percentage !== undefined
-		) {
+		for (const containerAppOptional of value) {
+			// validate that either a name is set and is a string
+			if (!isOptionalProperty(value, "name", "string")) {
+				diagnostics.errors.push(
+					`the field "name", when present, should be a string.`
```
CarmenPopoviciu:
```suggestion
					`Field "name", when present, should be a string, but got ${JSON.stringify(value)}`
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2376,59 +2376,95 @@ const validateBindingArray =
...
+
+			// or if unset that we can infer a reasonable name
+			if (!containerAppOptional.name) {
+				// we need at a minimum a topLevelName and a containers.class_name or containers.name
+				if (
+					!topLevelName ||
+					!isOptionalProperty(containerAppOptional, "class_name", "string")
+				) {
+					diagnostics.errors.push(
+						`"Must have either a top level name and a containers.class_name defined or have containers.name defined"`
```
CarmenPopoviciu:
```suggestion
						`Must have either a top level "name" and "containers.class_name" field defined, or have field "containers.name" defined.`
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2376,59 +2376,95 @@ const validateBindingArray =
...
+			//
+			// At the moment logic in other places downstream of this rely on containerAppOptional.configuration.image be set
+			// so we set it here regardless of which place it is set by the user.
+			if (
+				"image" in containerAppOptional &&
+				containerAppOptional.image !== undefined
+			) {
+				if (containerAppOptional.configuration?.image !== undefined) {
+					diagnostics.errors.push(
+						`"containers.image" and "containers.configuration.image" can't be defined at the same time`
```
CarmenPopoviciu:
```suggestion
						`"containers.image" and "containers.configuration.image" fields can't be defined at the same time.`
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2376,59 +2376,95 @@ const validateBindingArray =
...
+				!isOptionalProperty(
+					containerAppOptional,
+					"rollout_step_percentage",
+					"number"
+				) &&
+				(containerAppOptional.rollout_step_percentage > 100 ||
+					containerAppOptional.rollout_step_percentage < 25)
 			) {
 				diagnostics.errors.push(
 					`"containers.rollout_step_percentage" should be a number between 25 and 100, but got ${containerAppOptional.rollout_step_percentage}`
```
CarmenPopoviciu:
```suggestion
					`"containers.rollout_step_percentage" field should be a number between 25 and 100, but got ${containerAppOptional.rollout_step_percentage}`
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2437,62 +2473,46 @@ const validateContainerAppConfig: ValidatorFn = (diagnostics, field, value) => {
 					`"containers.rollout_kind" should be either 'full_auto', 'full_manual' or 'none', but got ${containerAppOptional.rollout_kind}`
```
CarmenPopoviciu:
```suggestion
					`"containers.rollout_kind" field should be either 'full_auto', 'full_manual' or 'none', but got ${containerAppOptional.rollout_kind}`
```

CarmenPopoviciu (approved):
Thank you so much for working on this @IRCody ♥️ 

approving wit the kind request that workers-sdk team does some cleanup of the validation code, tracked in DEVX-2029

PR #9667, opened 2025-06-19
Verify docker installation earlier when deploying container

IRCody:
When containers are configured we want the container and the worker to be updated together (ideally). Failing earlier when we know docker is not installed keeps us from getting into a disjointed state between the worker and container.

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: not e2e test
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: improved ux
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  not a v3 feature.

### DIFF
--- a/packages/containers-shared/src/utils.ts
+++ b/packages/containers-shared/src/utils.ts
@@ -50,15 +50,18 @@ export const runDockerCmdWithOutput = async (
 };
 
 /** throws when docker is not installed */
-export const verifyDockerInstalled = async (dockerPath: string) => {
+export const verifyDockerInstalled = async (
+	dockerPath: string,
+	isDev = true
+) => {
 	try {
 		await runDockerCmd(dockerPath, ["info"], ["inherit", "pipe", "pipe"]);
 	} catch {
 		// We assume this command is unlikely to fail for reasons other than the Docker daemon not running, or the Docker CLI not being installed or in the PATH.
 		throw new Error(
-			`The Docker CLI could not be launched. Please ensure that Docker is installed and running. You can specify an executable with the environment variable WRANGLER_DOCKER_BIN.\n` +
-				`Other container tooling that is compatible with the Docker CLI may work, but is not yet guaranteed to do so.\n` +
-				`To suppress this error if you do not intend on triggering any container instances, set dev.enable_containers to false in your Wrangler config or passing in --enable-containers=false.`
+			`The Docker CLI could not be launched. Please ensure that the Docker CLI is installed and the daemon is running.\n` +
+				`Other container tooling that is compatible with the Docker CLI and engine may work, but is not yet guaranteed to do so. You can specify an executable with the environment variable WRANGLER_DOCKER_BIN and a socket with WRANGLER_DOCKER_HOST.` +
+				`${isDev ? "\nTo suppress this error if you do not intend on triggering any container instances, set dev.enable_containers to false in your Wrangler config or passing in --enable-containers=false." : ""}`
 		);
 	}
 };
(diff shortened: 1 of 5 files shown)


IRCody:
> i think you can deploy without docker if you have an image registry link instead of a docker file? bit of an edge case to have containers and _not_ docker, just flagging in case

That is a good point. I can add a check that you're supplying a Dockerfile and not a uri as well. Let me see how difficult that is.

emily-shen:
> > i think you can deploy without docker if you have an image registry link instead of a docker file? bit of an edge case to have containers and _not_ docker, just flagging in case
> 
> That is a good point. I can add a check that you're supplying a Dockerfile and not a uri as well. Let me see how difficult that is.

isDockerfile() === true should do it (helper function in containers-shared)

packages/containers-shared/src/utils.ts
```diff
@@ -50,15 +50,18 @@ export const runDockerCmdWithOutput = async (
 };
 
 /** throws when docker is not installed */
-export const verifyDockerInstalled = async (dockerPath: string) => {
+export const verifyDockerInstalled = async (
+	dockerPath: string,
+	isDev = true
```
CarmenPopoviciu, line 55: just sanity check...so the `isDev` flag is meant to be used for the error copy only, correct?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -8667,6 +8667,84 @@ addEventListener('fetch', event => {});`
 				expect(std.warn).toMatchInlineSnapshot(`""`);
 			});
 
+			it("should fail early if no docker is detected when deploying a container from a dockerfile", async () => {
```
penalosa, line 8670: I don't understand this test. It's:
1. overriding `WRANGLER_CONTAINERS_DOCKER_PATH` (which is no longer the correct path) to a "bad" binary
2. Adding a mock binary that asserts the path is different.

I _think_ this test should be adding a mock binary that _fails_, but it doesn't seem to be?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -9059,6 +9143,26 @@ addEventListener('fetch', event => {});`
 
 			it("should support durable object bindings to SQLite classes with containers", async () => {
 				// note no docker commands have been mocked here!
```
penalosa, line 9122: What is this note for?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -764,9 +768,23 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			config.containers === undefined;
 
 		let workerBundle: FormData;
+		const dockerPath = getDockerPath();
```
penalosa, line 771: This should be in the `if` block below

packages/containers-shared/src/utils.ts
emily-shen: yep

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: removed this mock which is unnecessary now

emily-shen: i think the point is that the docker cli is not used in this flow

packages/wrangler/src/deploy/deploy.ts
emily-shen: its also used a bit further down

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: fixed up the test so we just have a fake binary that will fail when we actually call docker , triggering the error.

PR #9675, opened 2025-06-19
fix changeset issue with containers-shared

emily-shen:
allows changesets to mention containers-shared. note tools tests only run on linux.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: tooling change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal tooling
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: package doesn't exist in v3

### DIFF
--- a/packages/containers-shared/package.json
+++ b/packages/containers-shared/package.json
@@ -17,6 +17,7 @@
 	"scripts": {
 		"check:lint": "eslint src --ext ts",
 		"check:type": "tsc -p ./tsconfig.json && pnpm run type:tests",
+		"deploy": "echo 'no deploy'",
 		"test": "vitest",
 		"test:ci": "pnpm run test run",
 		"test:watch": "pnpm run test --testTimeout=50000 --watch",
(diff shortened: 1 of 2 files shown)


.changeset/deep-moments-stay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/containers-shared": minor
```
CarmenPopoviciu, line 2: for posterity...`containers-shared` was added to workers-sdk without a changeset, and we kept adding /building on top of it without tracking the work with changesets. This brings us back to square one...sort of...so that we can add changesets accordingly for this package from here on <3

petebacondarwin (approved):

PR #9688, opened 2025-06-19
add remote bindings support to `getPlatformProxy`

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2028

add remote bindings support to `getPlatformProxy`

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: skipping the documentation while this is experimental
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -103,27 +111,37 @@ export async function getPlatformProxy<
...
-		{
-			MULTIWORKER: false,
-			RESOURCES_PROVISION: false,
-			// TODO: when possible remote bindings should be made available for getPlatformProxy
-			REMOTE_BINDINGS: false,
-		},
-		() => getMiniflareOptionsFromConfig(rawConfig, env, options)
+	let remoteProxySession: RemoteProxySession | undefined = undefined;
+	if (experimentalRemoteBindings && rawConfig.configPath) {
+		const a = await maybeStartOrUpdateRemoteProxySession(
```
penalosa: Does this need to use `maybeStartOrUpdate()` given it always passes null as the previous session?

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -103,27 +111,37 @@ export async function getPlatformProxy<
...
+			rawConfig.configPath,
+			null
+		);
+		remoteProxySession = a?.session;
+	}
+
+	const miniflareOptions = await getMiniflareOptionsFromConfig(
```
penalosa: We should modify `getMiniflareOptionsFromConfig()` to pass through `remoteBindingsEnabled` to `buildMiniflareBindingOptions()`

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -103,27 +111,37 @@ export async function getPlatformProxy<
...
+		rawConfig,
+		env,
+		options,
+		remoteProxySession?.remoteProxyConnectionString
 	);
 
 	const mf = new Miniflare({
 		script: "",
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
```
penalosa: Can you make this a `@ts-expect-error` with an explanatory comment?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -1339,11 +1340,7 @@ export async function buildMiniflareOptions(
 
 	const { sourceOptions, entrypointNames } = await buildSourceOptions(config);
 	const { bindingOptions, internalObjects, externalWorkers } =
-		buildMiniflareBindingOptions(
-			config,
-			remoteProxyConnectionString,
-			remoteBindingsEnabled
-		);
+		buildMiniflareBindingOptions(config, remoteProxyConnectionString);
```
penalosa: This isn't quite right. "enabled" and "has a connection string" are subtly different—even without a connection string, "enabled" will affect things like Images local mode and the logged warnings

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: No it does not, but I find it to be much more convenient than having to manually parse the config, get the bindings, convert them and pass them onto `startRemoteProxySession` 🤔 

I can do it that way if you prefer but `maybeStartOrUpdateRemoteProxySession` with a `null` second argument basically does exactly the above (which actually makes me wonder if maybe the function should be renamed and suggested more broadly? I think that most of the time this might just be what users might need? 🤔)

dario-piotrowicz: PS: I didn't mean to use `a` as a variable name, I need to change that 🤦

penalosa: Ah, fair enough. In that case, sounds good (although maybe make the second argument optional?)

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: mh.... sorry I'm not following, can you clarify what you mean?

Basically I removed all the `remoteBindingsEnabled` arguments since all of them actually seems very redundant to me, if a function's caller is passing a remoteProxySessionConnectionString, then the fact that remote bindings is enabled should be implicit (otherwise no connection string would be passed), doesn't that sound ok to you?

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: good call 🙂👍

packages/wrangler/src/dev/miniflare.ts
dario-piotrowicz: no.... I don't think I quite follow... 😕 

well since this change is quite irrelevant to `getPlatformProxy` I am also happy to revert this aspect of the PR and revisit this later 🤔

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: done 🙂

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: done 🙂

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: Second argument made optional 🙂👍

fixtures/get-platform-proxy-remote-bindings-node-test/index.test.js
```diff
@@ -0,0 +1,92 @@
...
+	...process.env,
+	CLOUDFLARE_API_TOKEN: process.env.TEST_CLOUDFLARE_API_TOKEN,
+	CLOUDFLARE_ACCOUNT_ID: process.env.TEST_CLOUDFLARE_ACCOUNT_ID,
+};
+
+describe("getPlatformProxy remote-bindings", () => {
+	const remoteWorkerName = `get-platform-proxy-remote-worker-test-${randomUUID().split("-")[0]}`;
+
+	before(async () => {
+		// Note: ideally we pass the auth data to `getPlatformProxy`, that currently is not
```
vicb, line 20: Does that mean that l 16-20 and 37 are not needed?

.changeset/better-sheep-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+add remote bindings support to `getPlatformProxy`
```
vicb, line 5: "remote" (here and in other files in this PR) -> do we still use the terminolgy or switched to "hybrid"?

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -136,15 +152,20 @@ export async function getPlatformProxy<
...
-		dispose: () => mf.dispose(),
+		dispose: async () => {
+			await remoteProxySession?.dispose();
+			await mf.dispose();
+		},
 	};
 }
 
 // this is only used by getPlatformProxy
 async function getMiniflareOptionsFromConfig(
```
vicb: Could you please add comments to this function? (and added/touched function in general).

One question I have look at this code is what is `env` vs `config.environment`.

Also maybe (as discussed in the weekly) do not add too many parameters but add an object - would be nice to refactor `buildMiniflareBindingOptions` to do the same (in a follow up PR).

i.e. Whats does that mean?

```ts
	const { bindingOptions, externalWorkers } = buildMiniflareBindingOptions(
		{
      // ...
		},
		undefined,
		false
	);
```

.changeset/better-sheep-reply.md
dario-piotrowicz: yes I think that remote bindings is the correct terminology: https://developers.cloudflare.com/workers/development-testing/#remote-bindings 🙂

fixtures/get-platform-proxy-remote-bindings-node-test/index.test.js
dario-piotrowicz: yeah that's a good point, they aren't really needed...

I can remove them, although keeping those line could be good, as later (after DEVX-1857 is solved) we could actually use some auth options in `getPlaformProxy` and having those lines would minimize the changes needed here...

I'm not sure I am 50/50 for keeping them 🤔 🤷

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: > Could you please add comments to this function? (and added/touched function in general).

done

> One question I have look at this code is what is env vs config.environment.

`env` is the environment that is being targeted (it's basically the equivalent to wrangler's `--env`), I've renamed it to `targetEnvironment` I hope that helps?

I'm not sure what you mean by `config.environment` can you clarify?

> Also maybe (as discussed in the weekly) do not add too many parameters but add an object

done

vicb: > I'm not sure what you mean by `config.environment` can you clarify?

Sorry, I meant `options.environment` (`options: GetPlatformProxyOptions,`)

fixtures/get-platform-proxy-remote-bindings-node-test/index.test.js
penalosa: I would vite for removing them. As it stands, this is kinda confusing to read

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -103,29 +111,33 @@ export async function getPlatformProxy<
 >(
 	options: GetPlatformProxyOptions = {}
 ): Promise<PlatformProxy<Env, CfProperties>> {
-	const env = options.environment;
+	const experimentalRemoteBindings = !!options.experimental?.remoteBindings;
+
+	const targetEnvironment = options.environment;
```
penalosa: Why are we renaming targetEnvironment like this? It's passed into `readConfig()` as `env`

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -103,29 +111,33 @@ export async function getPlatformProxy<
...
-			MULTIWORKER: false,
-			RESOURCES_PROVISION: false,
-			// TODO: when possible remote bindings should be made available for getPlatformProxy
-			REMOTE_BINDINGS: false,
-		},
-		() => getMiniflareOptionsFromConfig(rawConfig, env, options)
-	);
+	let remoteProxySession: RemoteProxySession | undefined = undefined;
+	if (experimentalRemoteBindings && rawConfig.configPath) {
+		const maybeRemoteProxySessionWrap =
```
penalosa: I dont think we need this intermediate variable

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -136,17 +148,40 @@ export async function getPlatformProxy<
...
+	remoteProxyConnectionString?: RemoteProxyConnectionString;
+	remoteBindingsEnabled: boolean;
+}): Promise<MiniflareOptions> {
+	const {
+		rawConfig,
+		targetEnvironment,
+		options,
+		remoteProxyConnectionString,
+		remoteBindingsEnabled,
+	} = args;
```
penalosa: nit, but this destructing seems a bit messy to me.

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: I was trying to address @vicb confusion regarding `env`.... I'm happy to revert the change

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: I totally agree

I was simply passing the values as an object as requested, I am happy to either revert the change or proceed any way you see fit (like always passing `args` around?)

fixtures/get-platform-proxy-remote-bindings-node-test/index.test.js
vicb: +1 to remove them and add details to DEVX-1857 (or a tracking issue on GH) with wht needs to be done

dario-piotrowicz: removed 👍

packages/wrangler/src/api/integrations/platform/index.ts
vicb: I prefer the current version FWIW

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: I don't think you can destructure a value into an already existing variable? can you? 🤔

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -136,17 +148,40 @@ export async function getPlatformProxy<
...
-	rawConfig: Config,
-	env: string | undefined,
-	options: GetPlatformProxyOptions
-): Promise<Partial<MiniflareOptions>> {
-	const bindings = getBindings(rawConfig, env, true, {});
+/**
+ * Builds an options configuration object for the `getPlatformProxy` functionality that
+ * can be then passed to the Miniflare constructor
+ *
+ * @param args.rawConfig The raw configuration to base the options from
```
vicb: IMO what would help in this comment is to know what "raw" means

packages/wrangler/src/api/integrations/platform/index.ts
vicb: We pass `options`, do we need to also pass `targetEnvironment = options.environment` ? l130

dario-piotrowicz: good call 👍

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: ah ok, as you commented below that's the same and unnecessary 🙂👍

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -122,9 +122,9 @@ export async function getPlatformProxy<
 
 	let remoteProxySession: RemoteProxySession | undefined = undefined;
 	if (experimentalRemoteBindings && rawConfig.configPath) {
-		const maybeRemoteProxySessionWrap =
-			await maybeStartOrUpdateRemoteProxySession(rawConfig.configPath);
-		remoteProxySession = maybeRemoteProxySessionWrap?.session;
+		remoteProxySession = (
```
vicb, line 125: That look messy to me. IMO the former version looked better!

packages/wrangler/src/api/integrations/platform/index.ts
dario-piotrowicz: good call, it actually is not "raw", I've updated the variable name and comment 🙂👍

packages/wrangler/src/api/integrations/platform/index.ts
vicb: and by that I also mean that you do not have to implement all comments if they don't make sense to you - even my comments :)

packages/wrangler/src/api/integrations/platform/index.ts
penalosa: For future context, Wrangler has both a `RawConfig` type (pre validation) and a `Config` type (post validation & defaults)

packages/wrangler/src/api/integrations/platform/index.ts
penalosa: It's pretty much entirely a style choice, so I'll leave it up to your judgement @dario-piotrowicz, but fwiw I don't think it necessarily makes sense to change this function to accept an object of options. `remoteBindingsEnabled` is temporary while we're in the experimental phase, and IMO 4 parameters is a perfectly reasonable numbr of parameters to a function. Making it an options object adds this destructuring, which in my mind just means that there's an extra 7 lines of boilerplate obscuring what this function does. If we want to keep it as an options object, I would find either 1) using `args.remoteProxyConnectionString` etc... consistently throughout the function body or 2) destructuring in the function definition much clearer to read.

packages/wrangler/src/api/integrations/platform/index.ts
penalosa: Yeah, this wasn't quite what I was suggestion. I was thinking of `(await maybeStartOrUpdateRemoteProxySession(config.configPath))?. session`.

That said, this was mostly because the intermediate variable didn't add much clarity. What about:

`({session: remoteProxySession } = await maybeStartOrUpdateRemoteProxySession(config.configPath)`

PR #9705, opened 2025-06-23
feat(type-generation): improve type generation for workflow

### LINKED ISSUE #9436: Stricter generated types for Workflows in worker-configuration.d.ts
### Describe the solution

When you have durable objects defined and run `wrangler types` then the following strict types will be generated in `worker-configuration.d.ts`

```ts
declare namespace Cloudflare {
	interface Env {
		MY_DURABLE_OBJECT: DurableObjectNamespace<import("./workers/my-script").MyDurableObject>;
		...
	}
}
```

When a workflow is defined the generated type is:

```ts
declare namespace Cloudflare {
	interface Env {
		MY_WORKFLOW: Workflow;
		...
	}
}
```

It would be really helpful to have something akin to this instead:

```ts
declare namespace Cloudflare {
	interface Env {
		MY_WORKFLOW: Workflow<MyWorkflowParams>;
		...
	}
}
```

hiendv:
Fixes #9436

Make Workflow type generation stricter with `generateImportSpecifier` (thanks to @ottomated https://github.com/cloudflare/workers-sdk/pull/6259).

Before:
```ts
interface Env {
	MY_WORKFLOW: Workflow;
}
```

After:
```ts
interface Env {
	MY_WORKFLOW: Workflow<Parameters<import("./src/index").MyWorkflow['run']>[0]['payload']>;
}
```

This type-hint looks a bit messy because params is only reflected with the run function. I think we could improve that by introducing a utility type (for now, I keep it simple with the "verbose" version).
```ts

type AnyWorkflowClass = new (...args: any) => WorkflowEntrypoint<any, any>;

export type GetWorkflowPayload<T extends AnyWorkflowClass> =
  InstanceType<T> extends WorkflowEntrypoint<any, infer P> ? P : never;
```

And then:
```ts
interface Env {
	  MY_WORKFLOW: Workflow<GetWorkflowPayload<typeof MyWorkflow>>;
}
```

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [] I don't know
  - [ ] Required
  - [x] Not required because: It's not a breaking change and all we did is adding a type to unknown generic
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: It doesn't directly serve users
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a patch change

### DIFF
--- a/packages/wrangler/src/__tests__/type-generation.test.ts
+++ b/packages/wrangler/src/__tests__/type-generation.test.ts
@@ -570,7 +570,7 @@ describe("generate types", () => {
 	it("should include stringified process.env types for vars, secrets, and json", async () => {
 		fs.writeFileSync(
 			"./index.ts",
-			`import { DurableObject } from 'cloudflare:workers';
+			`import { DurableObject, WorkflowEntrypoint, type WorkflowEvent, type WorkflowStep } from 'cloudflare:workers';
 				export default { async fetch () {} };
 				export class DurableDirect extends DurableObject {}
 				export { DurableReexport } from './durable-2.js';
@@ -689,7 +689,7 @@ describe("generate types", () => {
 
 		fs.writeFileSync(
 			"./a/index.ts",
-			`import { DurableObject } from 'cloudflare:workers';
+			`import { DurableObject, WorkflowEntrypoint, type WorkflowEvent, type WorkflowStep } from 'cloudflare:workers';
 				export default { async fetch () {} };
 				export class DurableDirect extends DurableObject {}
 


penalosa:
@sidharthachatterjee could you take a look at this?

sidharthachatterjee:
Thanks @penalosa 

@LuisDuarte1 from the Workflows team will be reviewing this, cheers

LuisDuarte1 (changes requested):
While the current approach technically works, I would prefer your latter approach with `GetWorkflowPayload` that uses the `infer` param instead of introspecting the type of `payload` field, wdyt?

hiendv:
@LuisDuarte1 Got it. Just one more thing, since we have no access to `workerd/worker.mjs` afaik, is it acceptable for me to add the utility type to the `type-generation`?

amcclosky:
@LuisDuarte1 @penalosa just recently started using workflows and this PR would be a nice improvement.

hiendv:
Well, if `workerd/worker.mjs` is not really meant to be touched, I'm happy to move the utility type to the generated output. Just think `workerd/worker.mjs` suits better though.

vicb:
@penalosa can we get this in (after rebase)?

penalosa:
@LuisDuarte1 are there concerns with the approach here?

LuisDuarte1:
> @LuisDuarte1 are there concerns with the approach here?

Would prefer to somehow hide the `.['run']>[0]['payload']>` part because that seems a bit verbose. But it's non-blocking and can be improved later.

penalosa:
@hiendv in order to add the type helper, you'd have to make a PR to workerd modifying the workflows type definitions: https://github.com/cloudflare/workerd/blob/main/types/defines/workflows.d.ts. I'd be happy to review and get that released if you open one!

hiendv:
OK so I took a look at workerd and it seems more complicated than anticipated.
This is my proposal
1. workerd/types/defines/rpd.d.ts (yes, not workflow.d.ts because WorkflowEntrypoint belongs here)
```ts
// After export abstract class WorkflowEntrypoint
type AnyWorkflowEntrypoint = new (...args: any) => WorkflowEntrypoint;
export type WorkflowPayload<T extends AnyWorkflowEntrypoint> = InstanceType<T> extends WorkflowEntrypoint<unknown, infer P> ? P : never;
```

2. type-generation
```ts
interface Env {
		MY_WORKFLOW: Workflow<import('cloudflare:workers').WorkflowPayload<typeof import("./src/index").MyWorkflow>>;
	}
```

Doesn't seem better because of the `import('cloudflare:workers')`, to be honest. What do you think?

hiendv:
> 2. type-generation
> 
> ```ts
> interface Env {
> 		MY_WORKFLOW: Workflow<import('cloudflare:workers').WorkflowPayload<typeof import("./src/index").MyWorkflow>>;
> 	}
> ```
> 
> Doesn't seem better because of the `import('cloudflare:workers')`, to be honest. What do you think?

Lol, pipeline looks messy as well :laughing: 

```ts
import("cloudflare:pipelines").Pipeline<import("cloudflare:pipelines").PipelineRecord>;
```

PR #9711, opened 2025-06-23
Include tag and message in assets-only version uploads

### LINKED ISSUE #9611: `wrangler versions upload` ignores `--tag` and `--message` arguments for Static Assets deployments

Wrangler v4.20.3, Node v22.16.0, Ubuntu 24.04.2 LTS

`wrangler.jsonc` defined as follows
```
{
	"name": "example",
	"compatibility_date": "2025-06-14",
	"assets": {
		"directory": "./public",
        "not_found_handling": "404-page"
	}
}
``` 

And running
```
wrangler versions upload --tag "abc123" --message "feat: Added xyz"`
```

### Describe the Bug

When using `wrangler versions upload` with --tag and --message arguments for a project that is solely a static assets deployment (i.e. no workers script added) the arguments are ignored. The resulting deployment in workers is created without the specified tag or message.

> [!NOTE]
>  After some further testing, this is only happening on deployments without a workers script provided (i.e. static only).

**Expected Behavior:** The uploaded version should have the specified tag and message applied, and show in the workers dashboard under `Version Tag & Message`

**Actual Behavior:** The version is uploaded successfully, but the tag and message are not set. No error is reported during the upload process.

This is with a basic static site (using Hugo), following [the latest pages migration guide](https://developers.cloudflare.com/workers/static-assets/migration-guides/migrate-from-pages/) to create a wrangler.toml (or jsonc in my case)

Running wrangler completes, with the version being uploaded successfully - just the metadata doesn't make it.

I suspect this isn't intended behaviour, given it works for worker deployments otherwise - just static ones it breaks on - suspect something needs tweaking in wrangler/the backend.

N/A

emily-shen:
Fixes #9611 

TODO: refactor so we don't have that short circuit there for asset only uploads to prevent this problem in the future. 

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: went GA after v4

### DIFF
--- a/.changeset/quiet-stars-chew.md
+++ b/.changeset/quiet-stars-chew.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+fix: include tag and message in assets-only version uploads
--- a/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
+++ b/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
@@ -211,6 +211,23 @@ describe("versions upload", () => {
 		expect(std.info).toContain("Retrying API call after error...");
 	});
 
+	it("should include annotations in the upload request even if there are only assets", async () => {
+		mockGetScript();
+		mockUploadVersion(true, 1, {
+			"workers/message": "test message",
+			"workers/tag": "test",
+		});
+		mockGetWorkerSubdomain({ enabled: true, previews_enabled: true });
+		mockSubDomainRequest();
+		writeWranglerConfig({
+			name: "test-name",
+			main: "./index.js",
+		});
+		writeWorkerSource();
+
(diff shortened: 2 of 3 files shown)


emily-shen:
i am not sure if the changes in the last commit are better - on one hand, we're less likely to accidentally not include something that should be uploaded with assets only cases. on the other hand, we're more likely to accidentally include something that _shouldn't_ be uploaded with assets only cases, but i think/hope we should be catching that earlier in validation.

PR #9715, opened 2025-06-23
R2-3164: ETags are now MD5

helloimalastair:
Fixes R2-3164.

ETags have been Hex-encoded MD5 hashes of the object since [June 21, 2023](https://developers.cloudflare.com/r2/objects/multipart-objects/#etags). This change makes Miniflare use the MD5 hash it already computes as the ETag, rather than a random value it generates.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: Not a Wrangler/Vite change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: This change is to match documented behaviour
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR:

---

<a href="https://app.devin.ai/review/cloudflare/workers-sdk/pull/9715" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/miniflare/test/plugins/r2/index.spec.ts
+++ b/packages/miniflare/test/plugins/r2/index.spec.ts
@@ -1237,6 +1237,15 @@ test("completeMultipartUpload", async (t) => {
 		message:
 			"completeMultipartUpload: One or more of the specified parts could not be found. (10025)",
 	};
+	// Check completing with multiple parts of same part number
+	const uploady = await r2.createMultipartUpload("key");
+	const part1a = await uploady.uploadPart(1, "1".repeat(PART_SIZE));
+	const part1b = await uploady.uploadPart(1, "2".repeat(PART_SIZE));
+	const part1c = await uploady.uploadPart(1, "3".repeat(PART_SIZE));
+	await t.throwsAsync(
+		uploady.complete([part1a, part1b, part1c]),
+		internalErrorExpectations("completeMultipartUpload")
+	);
 	// Check completing with out-of-order parts
 	const upload5a = await r2.createMultipartUpload("key");
 	part1 = await upload5a.uploadPart(1, "1".repeat(PART_SIZE));
--- a/packages/miniflare/test/plugins/r2/index.spec.ts
+++ b/packages/miniflare/test/plugins/r2/index.spec.ts
@@ -556,29 +556,29 @@ test("put: can copy values", async (t) => {
 		script: `export default {
       async fetch(request, env, ctx) {
         await env.BUCKET.put("key", "0123456789");
-        
+
         let object = await env.BUCKET.get("key");
         await env.BUCKET.put("key-copy", object.body);
         const copy = await (await env.BUCKET.get("key-copy"))?.text();
-        
+
         object = await env.BUCKET.get("key", { range: { offset: 1, length: 4 } });
         await env.BUCKET.put("key-copy-range-1", object.body);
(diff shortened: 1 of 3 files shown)


packages/miniflare/test/plugins/r2/index.spec.ts
```diff
@@ -1237,31 +1227,16 @@ test("completeMultipartUpload", async (t) => {
...
-	// Check completing with multiple parts of same part number
-	const upload4 = await r2.createMultipartUpload("key");
-	part1a = await upload4.uploadPart(1, "1".repeat(PART_SIZE));
-	part1b = await upload4.uploadPart(1, "2".repeat(PART_SIZE));
-	const part1c = await upload4.uploadPart(1, "3".repeat(PART_SIZE));
-	await t.throwsAsync(
-		upload4.complete([part1a, part1b, part1c]),
-		internalErrorExpectations("completeMultipartUpload")
-	);
-
```
penalosa: Why are these tests being removed?

penalosa:
@helloimalastair will this break behaviour with existing Miniflare databases?

packages/miniflare/test/plugins/r2/index.spec.ts
helloimalastair: Just re-added one test that on a second pass, would still pass.

The first one in your selection here was removed because it relies on the test above to attempt to upload the same data multiple times and get different ETags. Since ETags are now the same across multiple uploadParts of the same content, it doesn't make sense to test that case.

helloimalastair:
> @helloimalastair will this break behaviour with existing Miniflare databases?

Yeah, it would. Is there some way you would recommend of handling larger changes to the DB across versions?

penalosa:
> Yeah, it would. Is there some way you would recommend of handling larger changes to the DB across versions?

I'd first question if this change is absolutely necessary in local mode? If you think it is, then have a look at this migration code path, which will need extended for this case: https://github.com/cloudflare/workers-sdk/tree/main/packages/miniflare/src/plugins/r2/index.ts#L151-L153

helloimalastair:
I guess it isn't _necessarily_ a required change? This change was specifically to bring local and prod into sync, as before June 21 2023 ETags were opaque blobs, but they are now MD5. I haven't seen any reports of people who expect it to be an MD5, I just wanted to be proactive about it. If it is too much fuss though, I can close this PR.

penalosa:
Since this will break behaviour with existing DBs and has differed with prod with no complaints for ~2 years I'm going to close this for now.

gabivlj:
Could we consider bringing this back? @penalosa 
I've been building something that uploads parts to R2 and it'd be nice if the md5 I calculate on the client matches on R2. I think I can't test locally this in specific because miniflare is not 1:1 to production. Now I need an if statement on my worker code that checks if it has to calculate the md5 on the stream...

workers-devprod:
- @cloudflare/wrangler

- packages/miniflare/src/workers/r2/bucket.worker.ts: [@cloudflare/wrangler]
- packages/miniflare/src/workers/r2/schemas.worker.ts: [@cloudflare/wrangler]
- packages/miniflare/test/plugins/r2/index.spec.ts: [@cloudflare/wrangler]

</details>

PR #9728, opened 2025-06-24
add e2e container tests

emily-shen:
adds tests for basic container interactions to containers that are built and pulled.

TODO:
rebuild hotkey tests

would like to test the container class too but maybe best to set something up on that repo directly?

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [ ] I don't know
  - [ ] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/fixtures/container-app/tests/index.test.ts
+++ b/fixtures/container-app/tests/index.test.ts
@@ -1,0 +1,95 @@
+import { ChildProcess, execSync } from "node:child_process";
+import { resolve } from "node:path";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import { runWranglerDev } from "../../shared/src/run-wrangler-long-lived";
+
+describe
+	.skipIf(process.env.platform !== "linux" && process.env.CI === "true")
+	.sequential.each([
+		{ source: "dockerfile (built)", config: "wrangler.jsonc" },
+		{
+			source: "registry uri (pulled)",
+			config: "wrangler.registry.jsonc",
+		},
+	])("container local dev with image from $source", (testCase) => {
+	let ip: string,
+		port: number,
+		stop: (() => Promise<unknown>) | undefined,
+		wranglerProcess: ChildProcess;
+
+	beforeAll(async () => {
+		({ ip, port, stop, wranglerProcess } = await runWranglerDev(
+			resolve(__dirname, ".."),
+			["--port=0", "--inspector-port=0", `-c=${testCase.config}`]
+		));
+		// cleanup any running containers
+		const ids = getContainerIds();
+		if (ids.length > 0) {
+			execSync("docker rm -f " + ids.join(" "), {
+				encoding: "utf8",
+			});
+		}
+	});
+
+	afterAll(async () => {
+		await stop?.();
+		// cleanup any containers that were started during the tests
+		const ids = getContainerIds();
+		if (ids.length > 0) {
+			execSync("docker rm -f " + ids.join(" "), {
+				encoding: "utf8",
+			});
+		}
+	});
+
+	it("should check initial container status (not running)", async () => {
+		const response = await fetch(`http://${ip}:${port}/status`);
+		const status = await response.json();
+		expect(response.status).toBe(200);
+		expect(status).toBe(false);
+	});
+
+	it("should start a container and be able to make requests to it", async () => {
+		let response = await fetch(`http://${ip}:${port}/start`);
+		let text = await response.text();
+		expect(response.status).toBe(200);
+		expect(text).toBe("Container create request sent...");
+
+		// Wait a bit for container to start
+		await new Promise((resolve) => setTimeout(resolve, 2000));
+
+		response = await fetch(`http://${ip}:${port}/status`);
+		const status = await response.json();
+		expect(response.status).toBe(200);
+		expect(status).toBe(true);
+
+		response = await fetch(`http://${ip}:${port}/fetch`);
+		expect(response.status).toBe(200);
+		text = await response.text();
+		expect(text).toBe("Hello World! Have an env var! I'm an env var!");
+		// Check that a container is running using `docker ps`
+		const ids = getContainerIds();
+		expect(ids.length).toBe(1);
+	});
+	it("should clean up any containers that were started", async () => {
+		// we want to give wrangler a chance to clean up, rather than using treekill
+		wranglerProcess.kill("SIGINT");
+		await new Promise<void>((resolve) => {
+			wranglerProcess.once("exit", () => resolve());
+		});
+		vi.waitFor(() => {
+			const remainingIds = getContainerIds();
+			expect(remainingIds.length).toBe(0);
+		});
+	});
+});
+
+const getContainerIds = () => {
+	// note the -a to include stopped containers
+	const ids = execSync("docker ps -a -q");
+	return ids
+		.toString()
+		.trim()
+		.split("\n")
+		.filter((id) => id.trim() !== "");
+};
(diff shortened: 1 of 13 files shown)


.github/workflows/test-and-check.yml
```diff
@@ -217,8 +217,8 @@ jobs:
         if: steps.changes.outputs.everything_but_markdown == 'true'
         run: pnpm run test:ci --concurrency 1 ${{ matrix.filter }}
         env:
-          TEST_CLOUDFLARE_API_TOKEN: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
-          TEST_CLOUDFLARE_ACCOUNT_ID: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
+          CLOUDFLARE_API_TOKEN: ${{ secrets.TEST_CLOUDFLARE_API_TOKEN }}
+          CLOUDFLARE_ACCOUNT_ID: ${{ secrets.TEST_CLOUDFLARE_ACCOUNT_ID }}
```
penalosa: Why are we adding auth unconditionally to the fixtures? The reason it's `TEST_` is so that fixtures don't automatically pick up auth and have specifically opt-in if they need it. In general, we should try and keep the fixtures runnable without auth, because the run on forks and locally as well

.github/workflows/test-and-check.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-24.04-arm, windows-latest]
+        os: [ubuntu-latest, windows-latest]
```
penalosa: Why this change? The arm runners are generally faster

fixtures/container-app/tests/index.test.ts
```diff
@@ -0,0 +1,102 @@
+import { ChildProcess, execSync } from "node:child_process";
+import { mkdtemp } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path, { resolve } from "node:path";
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import { runWranglerDev } from "../../shared/src/run-wrangler-long-lived";
+
+describe
+	.skipIf(process.platform !== "linux" && process.env.CI === "true")
```
penalosa: Why is this only running on Linux?

.github/workflows/test-and-check.yml
emily-shen: sorry i was just flipping random switches trying to get these to run, will figure out how to do this properly once i've got the tests runing properly _somewhere_ 😅

.github/workflows/test-and-check.yml
emily-shen: our containers are built for amd64 so we can't run them on the arm runners without another VM layer. im thinking we might need to set up a separate containers workflow...

fixtures/interactive-dev-tests/tests/index.test.ts
```diff
@@ -292,3 +288,158 @@ it.each(exitKeys)("multiworker cleanly exits with $name", async ({ key }) => {
...
+		return [];
+	}
+	const jsonOutput = allContainers.map((line) => JSON.parse(line));
+
+	const matches = jsonOutput.map((container) => {
+		if (container.Image.includes("cloudflare-dev/fixturetestcontainer")) {
+			return container.ID;
+		}
+	});
+	return matches.filter((id) => id);
```
dario-piotrowicz: Nit 🙂 
```suggestion
	return matches.filter(Boolean);
```

fixtures/interactive-dev-tests/tests/index.test.ts
```diff
@@ -292,3 +288,158 @@ it.each(exitKeys)("multiworker cleanly exits with $name", async ({ key }) => {
...
+
+baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
+	"container dev",
+	{ retry: 0, timeout: 90000 },
+	() => {
+		let tmpDir: string;
+		beforeAll(async () => {
+			tmpDir = fs.mkdtempSync(path.join(tmpdir(), "wrangler-container-"));
+			fs.cpSync(
+				path.resolve(__dirname, "../..", "container-app"),
```
dario-piotrowicz: could you copy the relevant files from the `container-app` here instead of adding a dependency from one fixture to another? this doesn't feel right to me 🤔

fixtures/interactive-dev-tests/tests/index.test.ts
```diff
@@ -292,3 +288,158 @@ it.each(exitKeys)("multiworker cleanly exits with $name", async ({ key }) => {
...
+
+		it("should rebuild a container when the hotkey is pressed", async () => {
+			const wrangler = await startWranglerDev([
+				"dev",
+				"-c",
+				path.join(tmpDir, "wrangler.jsonc"),
+			]);
+			await fetch(wrangler.url + "/start");
+
+			await new Promise((resolve) => setTimeout(resolve, 2000));
```
dario-piotrowicz: instead of an hardcoded `setTimeout` could you instead use [`vi.waitFor`](https://vitest.dev/api/vi.html#vi-waitfor)?

fixtures/interactive-dev-tests/tests/index.test.ts
```diff
@@ -292,3 +288,158 @@ it.each(exitKeys)("multiworker cleanly exits with $name", async ({ key }) => {
...
+
+			const server = createServer(function (req, res) {
+				res.writeHead(200, { "Content-Type": "text/plain" });
+				res.write("Blah! " + process.env.MESSAGE);
+				res.end();
+			});
+
+			server.listen(8080, function () {
+				console.log("Server listening on port 8080");
+			});`,
```
dario-piotrowicz:
```suggestion
				const server = createServer(function (req, res) {
					res.writeHead(200, { "Content-Type": "text/plain" });
					res.write("Blah! " + process.env.MESSAGE);
					res.end();
				});

				server.listen(8080, function () {
					console.log("Server listening on port 8080");
				});`,
					"utf-8"
```

fixtures/interactive-dev-tests/tests/index.test.ts
dario-piotrowicz: PS: this comment applies to all the other hardcoded `setTimeout`s in this file 🙂

dario-piotrowicz (commented):
Generally looks good to me, but I don't think that we should extend `interactive-dev-tests` to also include containers, I think that having a separate `interactive-containers-dev` fixture would be much cleaner, however the extension of the fixture has already been done in https://github.com/cloudflare/workers-sdk/pull/9695 so I won't really push hard against it at this point

fixtures/interactive-dev-tests/tests/index.test.ts
emily-shen: oh nice TIL

fixtures/interactive-dev-tests/tests/index.test.ts
emily-shen: that was a useful diversion - turns out the default waitfor interval (100ms) results in the container freezing up and crashing. notably, so is 300ms, which is the containers packages' retry interval... 

anyway using waitfor at a 500ms interval now

dario-piotrowicz: oof... I hope that that wasn't too painful 😓

PR #9729, opened 2025-06-24
Spawn docker build relative to the Dockerfile

404Wolf:
Fixes [this beta container issue](https://discord.com/channels/595317990191398933/1354872988616364213/1383130425890967624).

Set build context relative to the Dockerfile so that `COPY`s do not break.
Also just cleaning up some duplicated helpers tests

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/clever-regions-build.md
+++ b/.changeset/clever-regions-build.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/containers-shared": patch
+---
+
+Spawn docker from the correct path
--- a/.changeset/clever-regions-build.md
+++ b/.changeset/clever-regions-build.md
@@ -1,5 +1,5 @@
 ---
-"@cloudflare/containers-shared": patch
+"wrangler": patch
 ---
 
-Spawn docker from the correct path
+Correctly set docker build context during deploy when `image_build_context` is not provided
--- a/.changeset/clever-regions-build.md
+++ b/.changeset/clever-regions-build.md
@@ -2,4 +2,4 @@
 "wrangler": patch
 ---
 
-Correctly set docker build context during deploy when `image_build_context` is not provided
+Set docker build context to the Dockerfile directory when `image_build_context` is not explicitly provided
--- a/packages/containers-shared/src/build.ts
+++ b/packages/containers-shared/src/build.ts
@@ -28,7 +28,7 @@ export async function constructBuildCommand(
 	const dockerfile = readFileSync(options.pathToDockerfile, "utf-8");
 	// pipe in the dockerfile
 	buildCmd.push("-f", "-");
-	buildCmd.push(options.buildContext);
+	buildCmd.push(options.buildContext ?? path.dirname(options.pathToDockerfile));
 	logger?.debug(`Building image with command: ${buildCmd.join(" ")}`);
 	return { buildCmd, dockerfile };
 }
@@ -75,7 +75,7 @@ export async function buildImage(
 	const { buildCmd, dockerfile } = await constructBuildCommand({
 		tag: options.imageTag,
 		pathToDockerfile: options.image,
-		buildContext: options.imageBuildContext ?? path.dirname(options.image),
+		buildContext: options.imageBuildContext,
 		args: options.args,
 		platform: "linux/amd64",
 	});
--- a/packages/containers-shared/src/build.ts
+++ b/packages/containers-shared/src/build.ts
@@ -44,6 +44,7 @@ export function dockerBuild(
 	return new Promise((resolve, reject) => {
 		const child = spawn(dockerPath, options.buildCmd, {
 			stdio: ["pipe", "inherit", "inherit"],
+			cwd: options.dockerfile,
 		});
 		if (child.stdin !== null) {
 			child.stdin.write(options.dockerfile);
--- a/packages/containers-shared/src/build.ts
+++ b/packages/containers-shared/src/build.ts
@@ -44,7 +44,7 @@ export function dockerBuild(
 	return new Promise((resolve, reject) => {
 		const child = spawn(dockerPath, options.buildCmd, {
 			stdio: ["pipe", "inherit", "inherit"],
-			cwd: options.dockerfile,
+			cwd: path.dirname(options.dockerfile),
 		});
 		if (child.stdin !== null) {
 			child.stdin.write(options.dockerfile);
--- a/packages/containers-shared/src/build.ts
+++ b/packages/containers-shared/src/build.ts
@@ -44,7 +44,6 @@ export function dockerBuild(
 	return new Promise((resolve, reject) => {
 		const child = spawn(dockerPath, options.buildCmd, {
 			stdio: ["pipe", "inherit", "inherit"],
-			cwd: path.dirname(options.dockerfile),
 		});
 		if (child.stdin !== null) {
 			child.stdin.write(options.dockerfile);
--- a/packages/containers-shared/src/types.ts
+++ b/packages/containers-shared/src/types.ts
@@ -10,7 +10,8 @@ export type BuildArgs = {
 	/** image tag in the format `name:tag`, where tag is optional */
 	tag: string;
 	pathToDockerfile: string;
-	buildContext: string;
+	/** image_build_context or args.PATH. if not provided, defaults to the dockerfile directory */
+	buildContext?: string;
 	/** any env vars that should be passed in at build time */
 	args?: Record<string, string>;
 	/** platform to build for. defaults to linux/amd64 */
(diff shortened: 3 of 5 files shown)


skepticfx (approved):

skepticfx:
Thank you 🚀

404Wolf:
Realized that we want the dirname of the dockerfile path, not the dockerfile path :/

emily-shen:
hey, thanks for the contribution! would you be able to add a regression test for this? probably in `e2e/containers.dev.test.ts`
let me know if you need any help with this :)

emily-shen:
also i'm not sure if misunderstanding the problem, but _do_ we want to spawn docker relative to the dockerfile? the dockerfile and image_build_context is defined in the wrangler config relative to the wrangler config, so it would make sense to spawn docker relative to the wrangler config too, and pass in the paths defined there. there might be a bug with how we resolve paths, but i'm struggling to reproduce. 

can you give an example of something that fails without this fix and works with? (i guess this comment is also just still asking for the regression test :p)

404Wolf:
> also i'm not sure if misunderstanding the problem, but _do_ we want to spawn docker relative to the dockerfile? the dockerfile and image_build_context is defined in the wrangler config relative to the wrangler config, so it would make sense to spawn docker relative to the wrangler config too, and pass in the paths defined there. there might be a bug with how we resolve paths, but i'm struggling to reproduce.
> 
> can you give an example of something that fails without this fix and works with? (i guess this comment is also just still asking for the regression test :p)

Hey Emily, I should have more time to look into this tomorrow, but the general issue is for when you have a setup that looks like:

cloudflare/wrangler.jsonc
cloudflare/main.ts
server/main.ts
Dockerfile

And you need the Dockerfile in the scope of the docker build, so you specify ".." for the image, but the build fails / COPYs fail because docker is run from CWD=./cloudflare

According to the docs:

image_build_context: "The build context of the application, by default it is the directory of image."
And if I have image set to ".." I feel it should spawn docker from ".."

emily-shen:
ah i see if i deploy with that file structure, I was confused because it doesn't seem to happen in dev. I think the actual bug is on this line: https://github.com/cloudflare/workers-sdk/blob/869b7551d719ccfe3843c25e9907b74024458561/packages/wrangler/src/cloudchamber/deploy.ts#L152

it should be falling back to the directory of the dockerfile rather than the cwd - buildContext is what gets passed into the docker build command, and the build command itself can spawn from wherever. 

also, does setting `image_build_context` fix the current issue, or is it also broken somehow? here is a quick repro i've been working off, where setting image_build_context to ".." seems to fix the issue - https://github.com/emily-shen/ancient-firefly-1260/tree/main

sorry for all the back and forth, just making sure we've not messed this up in multiple places! thanks so much for the contribution 🧡

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -8727,49 +8727,9 @@ addEventListener('fetch', event => {});`
 
 			it("should support durable object bindings to SQLite classes with containers (dockerfile flow)", async () => {
 				vi.stubEnv("WRANGLER_DOCKER_BIN", "/usr/bin/docker");
-				function mockGetVersion(versionId: string) {
```
vicb, line 8730: The change here does not seem to be directly related to the PR, could you please add a note in the PR description?

.changeset/clever-regions-build.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Correctly set docker build context during deploy when `image_build_context` is not provided
```
vicb: - Clarify "Correctly" 
- Looks to me that it should be "build" rather than "deploy"?

```suggestion
Set docker build context to the Dockerfile directory when `image_build_context` is not explicitly provided
```

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -9220,6 +9096,250 @@ addEventListener('fetch', event => {});`
...
+							"-t",
+							getCloudflareContainerRegistry() +
+								"/some-account-id/my-container:Galaxy",
+							"--platform",
+							"linux/amd64",
+							"--provenance=false",
+							"-f",
+							"-",
+							// this is the key! the build context should be based on the dockerfile location
+							"..",
```
vicb, line 9125: Question: this is a relative path so depends on cwd, should cwd be tested as well (or `..` be changed to an absolute path?)

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -9220,6 +9096,250 @@ addEventListener('fetch', event => {});`
...
+							output: [null],
+							on: (reason: string, cbPassed: (code: number) => unknown) => {
+								if (reason === "exit") {
+									expect(dockerfile).toEqual("FROM scratch");
+									cbPassed(0);
+								}
+							},
+						} as unknown as ChildProcess;
+					})
+					// 2. docker image inspect
```
vicb, line 9154: Does this test need 2, 3, 4, 5 ? 

Sounds like what we want to test is the build path ("resolve the docker build context path based on the dockerfile location, if image_build_context is not provided")

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -143,7 +144,7 @@ export function getBuildArguments(
 	return {
 		tag: imageTag,
 		pathToDockerfile: imageRef,
-		buildContext: container.image_build_context ?? ".",
+		buildContext: container.image_build_context ?? dirname(imageRef),
 		args: container.image_vars,
 	};
```
vicb, line 148: nit for readability:

```ts
	const pathToDockerfile = container.image ?? container.configuration?.image;
	const imageTag = container.name + ":" + idForImageTag.split("-")[0];
	return {
		tag: imageTag,
		pathToDockerfile,
		buildContext: container.image_build_context ?? dirname(pathToDockerfile),
		args: container.image_vars,
	};
	```

File: .changeset/clever-regions-build.md (hunk 2 above)
Comment (emily-shen) [reply .changeset/clever-regions-build.md]: this error was only happening when building at deploy time, not during dev

Comment (vicb) [reply .changeset/clever-regions-build.md]: Oh interesting...
Then maybe the fix should be

```diff
- buildContext: container.image_build_context ?? ".",
+ buildContext: container.image_build_context,
```

So that both deploy and [dev](https://github.com/cloudflare/workers-sdk/blob/0bb619a92911415957d8788923302c15364638c9/packages/containers-shared/src/images.ts#L39-L63) have the same behavior - they both use  [`buildImage`](https://github.com/cloudflare/workers-sdk/blob/0bb619a92911415957d8788923302c15364638c9/packages/containers-shared/src/build.ts#L70-L84) where the default `dirname(dockerFile)` is applied

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: i wanted to test all the config resolution up until the build, but I don't think there's a good way to test just that part of the deploy path?

.changeset/clever-regions-build.md
emily-shen: set it one layer down in constructBuildCommand 👍

packages/containers-shared/src/build.ts
```diff
@@ -28,7 +28,7 @@ export async function constructBuildCommand(
 	const dockerfile = readFileSync(options.pathToDockerfile, "utf-8");
 	// pipe in the dockerfile
 	buildCmd.push("-f", "-");
-	buildCmd.push(options.buildContext);
+	buildCmd.push(options.buildContext ?? path.dirname(options.pathToDockerfile));
```
vicb, line 31: Could you add some JSDoc to the function to describe this behavior?

packages/containers-shared/src/types.ts
```diff
@@ -10,7 +10,8 @@ export type BuildArgs = {
 	/** image tag in the format `name:tag`, where tag is optional */
 	tag: string;
 	pathToDockerfile: string;
-	buildContext: string;
+	/** image_build_context or args.PATH. if not provided, defaults to the dockerfile directory */
```
vicb, line 13: +1 for updating the doc

packages/wrangler/src/__tests__/deploy.test.ts
vicb: I agree that ideally we should test that `pathToDockerFile` is correctly set by the build command. And then add another test testing that the build command reflects the passed value (explicit path or default to Dockerfile)

With the current, there are ~100 LOC that are not related to what we are trying to test and that might need to be updated when something unrelated to what we want to test changes.

If there is no easy/quick way to fix this, I don't think it should block merging the PR but we should revisit later. It could benefit test speed, coverage and maintainability.

PR #9734, opened 2025-06-24
Make Wrangler warn more loudly if you're missing auth scopes

penalosa:
Previously, when an authentication error was thrown, Wrangler would print out your Oauth token scopes with the passive message "If scopes are missing, you may need to logout and re-login." However, it was hard for users to know if scopes were missing, because users don't necessarily know what the right set of scopes are. 

As it turns out, Wrangler _does_ know what the set of scopes should be, so now we print out a much more actionable warning:
```
Wrangler is missing some expected Oauth scopes. To fix this, run \`wrangler login\` to refresh your token. The missing scopes are:
  - containers:write
```

Additionally, this fixes an issue where auth errors from Containers & Cloudchamber were not handled the same as other auth errors, because of the custom API client that those code paths use.

---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [ ] Required
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: messaging improvement
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: in progress, but I don't want to block this PR

### DIFF
--- a/packages/wrangler/src/index.ts
+++ b/packages/wrangler/src/index.ts
@@ -1,5 +1,6 @@
 import os from "node:os";
 import { setTimeout } from "node:timers/promises";
+import { ApiError } from "@cloudflare/containers-shared";
 import chalk from "chalk";
 import { ProxyAgent, setGlobalDispatcher } from "undici";
 import makeCLI from "yargs";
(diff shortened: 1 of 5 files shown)


irvinebroque:
YES 🐐 🐐 🐐 🐐

irvinebroque:
how does wrangler know what is missing? asking because IIRC no way to get this back directly from the Cloudflare API?

packages/wrangler/src/index.ts
```diff
@@ -1509,10 +1510,21 @@ export async function main(argv: string[]): Promise<void> {
 			// The workaround is to re-run the parsing with an additional `--help` flag, which will result in the correct help message being displayed.
 			// The `wrangler` object is "frozen"; we cannot reuse that with different args, so we must create a new CLI parser to generate the help message.
 			await createCLIParser([...argv, "--help"]).parse();
-		} else if (isAuthenticationError(e)) {
+		} else if (
+			isAuthenticationError(e) ||
+			// Is this a Containers/Cloudchamber-based auth error?
+			// This is different because
```
CarmenPopoviciu: i think this sentence is unfinished

packages/wrangler/src/user/whoami.ts
```diff
@@ -69,13 +76,30 @@ function printTokenPermissions(user: UserInfo) {
 			`🔓 To see token permissions visit https://dash.cloudflare.com/profile/api-tokens.`
 		);
 	}
-	logger.log(
-		`🔓 Token Permissions: If scopes are missing, you may need to logout and re-login.`
-	);
+	logger.log(`🔓 Token Permissions:`);
```
CarmenPopoviciu, line 84: should the order of these two logs be reversed, now that we're listing the missing token scopes?

penalosa:
> how does wrangler know what is missing? asking because IIRC no way to get this back directly from the Cloudflare API?

It decides based on what scopes it would request if it performed a `wrangler login` (which is a static list in the codebase)

packages/wrangler/src/user/whoami.ts
petebacondarwin: I "think" it is OK?

PR #9737, opened 2025-06-25
Run more tests, more of the time

penalosa:
One of the most painful category of release issues is _Something is wrong in VP that could have been caught earlier._ e.g.
- A PR was merged that didn’t run E2E tests but that broke them
- A PR was merged which combined badly with something else already on main

This PR aims to address that pain by running all tests (including E2E tests) in the Merge Queue, which we already enforce as a merge strategy. That will mean that forks can run E2E tests (since at the point a fork PR is in the Merge Queue it’s been approved, so can be trusted to run E2E tests with tokens). It will also mean that nothing lands in main without fully passing E2E tests, having been rebased on top of the tip of main. This also means we no longer have a split between PRs and VP in terms of what OS tests are run on.

There are a couple of things this PR has done to make this work well:
- Some hacking around of GH Actions to get things properly marked as required even if they’re skipped (basically, how can we make E2E tests a required check without completely breaking forks?). To do this, I've changed the stage at which runs that will be skipped are skipped—in exactly the same way as we do for markdown-only PRs, where CI tests don’t properly “run” but they do “run” from the perspective of GH Actions. 
- Apply pending changeset version bumps to packages before running tests. Every so often we get test failures on VP that didn’t show up on PRs because VP has bumped the version numbers in package.json files according to the pending changesets publish. This PR also applies this pending version update to the package.json files in packages before running tests on all workflows. This also has the benefit of allowing test runs on VP to hit cached values.

This PR _also_ simplifies the pull request template by:
- Removing the TODO checkboxes
- Removing the E2E checkboxes (since now they always run)
- Changing the tests checkbox to only have the option to mark "Tests included". This can be bypassed with the `no-tests` label, but hopefully removing the visible option for not including tests will encourage people to include tests
---

- Tests
  - [ ] TODO (before merge)
  - [x] Tests included
  - [ ] I don't know
  - [x] Required
  - [ ] Not required because:
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 


fixtures/interactive-dev-tests/package.json
```diff
@@ -12,7 +12,7 @@
 		"vitest": "catalog:default"
 	},
 	"optionalDependencies": {
-		"@cdktf/node-pty-prebuilt-multiarch": "0.10.1-pre.11"
+		"@cdktf/node-pty-prebuilt-multiarch": "0.10.2"
```
emily-shen, line 15: why this change?

.github/workflows/test-and-check.yml
```diff
@@ -44,42 +39,16 @@ jobs:
           turbo-token: ${{ secrets.TURBO_TOKEN }}
           turbo-signature: ${{ secrets.TURBO_REMOTE_CACHE_SIGNATURE_KEY }}
 
-      - name: Check for errors
-        run: pnpm run check --summarize
+      - name: Bump package versions
+        run: node .github/changeset-version.js
```
emily-shen, line 42: is there anything to worry about here re: order of merges? is this on every workflow just in case, or could they conflict somehow? i guess i don't really understand why different versions has caused failures in VP/what sort of failures.

.github/pull_request_template.md
emily-shen, line 1: 🎉

.github/workflows/test-and-check.yml
penalosa: This won't affect the actual code that is merged from a PR, it just ensures that the tests run against the PR as if it was on VP (with the version bumps applied). This is both to prevent problems (e.g. C3 inadvertently works because the published version of Wrangler is the same as the version on disk, but fails for some reason if they're different, which has happened in the past), but also to fix turbo caches so that VP can be cached. Currently VP is completely uncached because all the package.json files change, which busts the cache

fixtures/interactive-dev-tests/package.json
penalosa: Just a version bump while trying to get this fixture working in later versions of Node. Turns out it doesn't support them though (see https://github.com/cdktf/node-pty-prebuilt-multiarch/pull/119), but there's no harm in using a slightly newer version

.github/pull_request_template.md
```diff
@@ -10,19 +10,11 @@ The following selections do not need to be completed if this PR only contains ch
 -->
 
 - Tests
-  - [ ] TODO (before merge)
   - [ ] Tests included
-  - [ ] Tests not necessary because:
```
dario-piotrowicz, line 15: did you mean to remove this checkbox? this change doesn't seem intentional to me? 🤔 

(test being the only checkbox without the not-necessary checkbox)
Screenshot 2025-06-25 at 17 12 50

.github/workflows/c3-e2e.yml
```diff
@@ -8,23 +9,29 @@ env:
...
+            label: pnpm, Windows
+            pm: { name: pnpm, version: "9.12.0" }
+          - os: ubuntu-latest
+            label: npm, Linux
+            pm: { name: npm, version: "0.0.0" }
+          - os: ubuntu-latest
+            label: yarn, Linux
+            pm: { name: yarn, version: "1.0.0" }
+          - os: ubuntu-latest
+            label: pnpm, Linux
```
dario-piotrowicz: why this explicit list?

.github/workflows/c3-e2e.yml
```diff
@@ -1,5 +1,6 @@
-name: C3 E2E Tests
+name: E2E
 on:
+  merge_group:
   pull_request:
     types: [synchronize, opened, reopened, labeled, unlabeled]
```
dario-piotrowicz: if we're not checking for labels anymore shouldn't we remove `labeled`/`unlabeled` from here? (we don't need to re-run the workflow when labels change, right?)

dario-piotrowicz: (this also applies to the other workflows)

.github/workflows/c3-e2e.yml
```diff
@@ -8,23 +9,29 @@ env:
...
+  e2e:
     # Note: please keep this job in sync with the e2e-only-dependabot-bumped-framework one
     #       in .github/workflows/c3-e2e-dependabot.yml
-    if: github.head_ref == 'changeset-release/main' || (contains(github.event.*.labels.*.name, 'c3-e2e' ) && contains(github.event.*.labels.*.name, 'every-os' ))
     timeout-minutes: 45
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.pm.name }}-${{ matrix.pm.version }}
       cancel-in-progress: true
-    name: ${{ format('Run tests for {0}@{1} on {2}', matrix.pm.name, matrix.pm.version, matrix.os) }}
+    name: ${{ format('C3 ({0})', matrix.label) }}
```
dario-piotrowicz: I would really avoid using the word `label` used for something different from what GitHub refers as labels, sounds to me like something bound to cause confusion at some point

dario-piotrowicz: (this also applies to the other workflows)

.github/pull_request_template.md
edmundhung: Samuel mentioned in the description that `Changing the tests checkbox to only have the option to mark "Tests included". This can be bypassed with the no-tests label`.

But I wonder if it's better to keep the no test checkbox as we lost the context on why if we rely on the label.

.github/workflows/c3-e2e.yml
```diff
@@ -1,5 +1,6 @@
-name: C3 E2E Tests
+name: E2E
```
dario-piotrowicz, line 1: This is what shows up in the github UI is it not? I think including C3 in the name would make it much more clear 🤔

.github/workflows/e2e.yml
```diff
@@ -1,31 +1,41 @@
-name: E2E tests
+name: E2E
```
dario-piotrowicz, line 1: same as with the other workflow, I would include wrangler in the name here

dario-piotrowicz (approved):
I've left a few comments, but nothing blocking, looks good to me 🙂

.github/workflows/e2e.yml
penalosa: This will show up in github as `E2E / Wrangler`

.github/pull_request_template.md
dario-piotrowicz: I see... yeah but the way I would expect it, similarly to how V3 backporting works, is that we do have the check and also the label 🤔

.github/workflows/c3-e2e.yml
penalosa: This will show up `E2E / C3`

.github/workflows/c3-e2e.yml
penalosa: This doesn't change the things we're running on, just makes it a bit more explicit. It read clearer to me to do it like this rather than a mix of combinations + includes, but happy to revert if you prefer the previous way!

.github/pull_request_template.md
penalosa: The description had my motivation here: "hopefully removing the visible option for not including tests will encourage people to include tests", but happy to revert this. My thinking was that basically all PRs should have tests, and it should be fairly hidden option to not add tests. It'll also be much more obvious for us when reviewing that a PR has the `no-tests` label, which should make it easier to spot if something has been inadvertently marked as not requiring tests.

.github/workflows/e2e.yml
dario-piotrowicz: ah ok 🙂

.github/workflows/c3-e2e.yml
dario-piotrowicz: ah ok 🙂

.github/workflows/c3-e2e.yml
dario-piotrowicz: ah nono I was just wondering why specifically this set and why not having an os x pm matrix and run every pm (we want to support) on every os (we want to support)

but including all might be a bit overkill and I'm ok not changing the current behavior 👍 

thanks 🙂

.github/pull_request_template.md
penalosa: I've reverted this change

dario-piotrowicz: > It'll also be much more obvious for us when reviewing that a PR has the `no-tests` label, which should make it easier to spot if something has been inadvertently marked as not requiring tests.

External contributors can't add the label though 😕

dario-piotrowicz: >  My thinking was that basically all PRs should have tests

I agree if we're talking about actual code changes, but often there other types of changes that don't require tests, such as code comments, changeset changes, templates changes, infra changes, md changes (which we do have a comment about in the pr template anyways) 🤔

PR #9746, opened 2025-06-25
add permissions to github workflows

dario-piotrowicz:
Fixes https://github.com/cloudflare/workers-sdk/security/code-scanning/173

update all the github workflows to specify the set of permissions they require, ensuring that each workflow includes the least privileges it required to complete the task

> [!Note]
> Hopefully I got everything right but it's tough to test the changes and be 100% sure (since the changes also only apply after merge), so there is a non zero chance that this might break some workflows, I'll monitor things for a few days after the PR is merged and if something does break I'll fix it right away

---

- Tests
  - [ ] TODO (before merge)
  - [ ] Tests included
  - [x] Tests not necessary because: ci change
  - [ ] I don't know
  - [ ] Required
  - [x] Not required because: ci change
- Public documentation
  - [ ] TODO (before merge)
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: ci change
- Wrangler V3 Backport
  - [ ] TODO (before merge)
  - [ ] Wrangler PR: 
  - [x] Not necessary because: ci change

### DIFF
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -6,6 +6,7 @@ on:
       - main
 
 permissions:
+  # content:write and pull-request:write permissions needed to create changeset PRs
   contents: write
   pull-requests: write
 
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -6,9 +6,8 @@ on:
       - main
 
 permissions:
-  # content:write and pull-request:write permissions needed to create changeset PRs
-  contents: write
-  pull-requests: write
+  contents: read
+  # note: no write permissions are needed since the workflow uses GH_ACCESS_TOKEN instead of GITHUB_TOKEN
 
 jobs:
   release:
--- a/.github/workflows/changesets.yml
+++ b/.github/workflows/changesets.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
+permissions:
+  contents: write
+  pull-requests: write
+
 jobs:
   release:
     if: ${{ github.repository_owner == 'cloudflare' }}
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -5,6 +5,7 @@ on:
     types: [opened, transferred]
 
 permissions:
+  # issues:write permission needed so that the workflow can comment on issues
   issues: write
 
 jobs:
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -4,6 +4,9 @@ on:
   issues:
     types: [opened, transferred]
 
+permissions:
+  issues: write
+
 jobs:
   add-to-project:
     name: Add issue to project
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -5,7 +5,7 @@ on:
     types: [opened, transferred]
 
 permissions:
-  # issues:write permission needed so that the workflow can comment on issues
+  # issues:write permission needed so that the workflow can manage issues
   issues: write
 
 jobs:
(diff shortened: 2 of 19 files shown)


CarmenPopoviciu:
for workflows that require write access, would be nice to have a one line comment explainer why that level of access is needed. WDYT?

dario-piotrowicz:
> for workflows that require write access, would be nice to have a one line comment explainer why that level of access is needed. WDYT?

sure 🙂👍 

done 😄

.github/workflows/changesets.yml
```diff
@@ -5,6 +5,11 @@ on:
     branches:
       - main
 
+permissions:
+  # content:write and pull-request:write permissions needed to create changeset PRs
+  contents: write
+  pull-requests: write
```
penalosa: This shouldn't be needed. We use `GH_ACCESS_TOKEN` to create the PR, not `GITHUB_TOKEN`

.github/workflows/issues.yml
```diff
@@ -4,6 +4,10 @@ on:
   issues:
     types: [opened, transferred]
 
+permissions:
+  # issues:write permission needed so that the workflow can comment on issues
+  issues: write
```
penalosa: This workflow doesn't write any comments to issues, I don't believe?

.github/workflows/open-v3-maintenance-prs.yml
```diff
@@ -7,6 +7,11 @@ on:
     paths:
       - ".changeset/**.md"
 
+permissions:
+  # content:write and pull-request:write permissions needed to create PRs
+  contents: write
```
penalosa: This isn't needed, since we use `GH_ACCESS_TOKEN` for the PR creation

.github/workflows/pr-labeler.yml
```diff
@@ -2,11 +2,13 @@ name: "Pull Request Labeler"
 on:
   - pull_request_target
 
+permissions:
+  contents: read
+  # pull-request:write permission needed so that the workflow can comment on PRs
+  pull-requests: write
```
penalosa: I don't think this workflow comments on PRs?

penalosa (commented):
Most of this is adding `contents: read`. Did you consider changing the default permissions for our repo to be more restrictive instead? It's the Workflow permissions on https://github.com/cloudflare/workers-sdk/settings/actions

dario-piotrowicz:
> Most of this is adding `contents: read`. Did you consider changing the default permissions for our repo to be more restrictive instead? It's the Workflow permissions on https://github.com/cloudflare/workers-sdk/settings/actions

No I didn't, I feel like having it explicitly stated on each workflow file clearly states/documents the permissions that the workflow requires instead of having such information implicit/hidden (to only admins)

Do you feel that it's too verbose?

.github/workflows/changesets.yml
dario-piotrowicz: oh! right! good call! 👍

.github/workflows/issues.yml
dario-piotrowicz: sorry, copy-pasta 😓👍

.github/workflows/pr-labeler.yml
dario-piotrowicz: yes my bad 😓, comment updated, thanks 🙏

dario-piotrowicz:
@penalosa all (silly 😓) mistakes addressed, thanks for spotting them! 🙏

petebacondarwin:
Can you resolve the conflicts here @dario-piotrowicz ?

dario-piotrowicz:
> Can you resolve the conflicts here @dario-piotrowicz ?

of course 🙂 

Sorry I didn't notice that there were conflicts here 👍

petebacondarwin (approved):

PR #9762, opened 2025-06-26
clarify and ensure that for local e2e any account can be used

dario-piotrowicz:
in the CONTRIBUTING.md document clarify the fact that any account can be used to create an API token for the tests (instead of suggesting to use DevProd Testing account alongise saying that other accounts can be used as well)

as part of this remove the unnecessary validation check that ensures that the account being used for the tests is the DevProd Testing one (since this is not really necessary)

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: md change
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: md change
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: md change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -353,7 +353,8 @@ WRANGLER="node ~/path/to/workers-sdk/packages/wrangler/wrangler-dist/cli.js" CLO
 1. Go to ["My Profile" > "User API Tokens"](https://dash.cloudflare.com/profile/api-tokens)
 1. Click "Create Token"
 1. Use the "Edit Cloudflare Workers" template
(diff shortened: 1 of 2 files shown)


dario-piotrowicz:
PS: I'm not sure.... we might as well also backport this? 🤔

penalosa (approved):

penalosa:
> PS: I'm not sure.... we might as well also backport this? 🤔

I don't think it's worth backporting this, personally

dario-piotrowicz:
> > PS: I'm not sure.... we might as well also backport this? 🤔
> 
> I don't think it's worth backporting this, personally

ok, sounds good to me 🙂👍

PR #1037, opened 2022-05-17
fix: don't attempt to login during a --dryRun

rozenmd:
Closes #1023 

//TODO:

- [x] Update tests to ensure dry-run works even without an account

Issue #1023, opened 2022-05-16
🐛 BUG: Running wrangler with --dry-run shouldn't try log me in


rozenmd:
### What version of `Wrangler` are you using?

2.0.5

### What operating system are you using?

Mac

### Describe the Bug

Running wrangler with the `--dry-run` option shouldn't try to log me in

PR #9765, opened 2025-06-26
Build containers without account ID

hasip-timurtas:
This allows us to build containers in dry run mode. When we push the container to the managed registry, we re-tag it to include the account ID. We only ever push when not in dry run mode though, so when pushing we can safely grab the account information.

Related: https://github.com/cloudflare/workers-sdk/pull/9813, https://github.com/cloudflare/workers-sdk/pull/9811

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: do not backport containers changes

### DIFF
--- a/packages/wrangler/src/deploy/deploy.ts
+++ b/packages/wrangler/src/deploy/deploy.ts
@@ -770,6 +770,23 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 		let workerBundle: FormData;
 		const dockerPath = getDockerPath();
 
+		// lets fail earlier in the case where docker isn't installed
+		// and we have containers so that we don't get into a
+		// disjointed state where the worker updates but the container
+		// fails.
+		if (config.containers) {
+			// if you have a registry url specified, you don't need docker
+			const hasDockerfiles = config.containers.some((container) =>
+				isDockerfile(
+					container.image ?? container.configuration?.image,
+					config.configPath
+				)
+			);
+			if (hasDockerfiles) {
+				await verifyDockerInstalled(dockerPath, false);
+			}
+		}
+
 		if (props.dryRun) {
 			if (config.containers) {
 				for (const container of config.containers) {
@@ -791,23 +808,6 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
 			);
 		} else {
 			assert(accountId, "Missing accountId");
-
-			// lets fail earlier in the case where docker isn't installed
-			// and we have containers so that we don't get into a
-			// disjointed state where the worker updates but the container
-			// fails.
-			if (config.containers) {
-				// if you have a registry url specified, you don't need docker
-				const hasDockerfiles = config.containers?.some((container) =>
-					isDockerfile(
-						container.image ?? container.configuration?.image,
-						config.configPath
-					)
-				);
-				if (hasDockerfiles) {
-					await verifyDockerInstalled(dockerPath, false);
-				}
-			}
 
 			if (getFlag("RESOURCES_PROVISION")) {
 				await provisionBindings(


hasip-timurtas:
@emily-shen  I tested manually, it fixes the issue I had 😅 
But yep sure, I will add the changeset and unit test as well. Will push soon, thanks for checking.

hasip-timurtas:
@emily-shen check it again please

petebacondarwin (approved):

hasip-timurtas:
## CI Failures - Unrelated to Container Fix

The CI failures appear to be flaky test issues unrelated to the container dry-run changes:

### Main Fix Status ✅
The core wrangler test suite passed (325/325 tests), including the new test for the container URL parsing fix.

### Failed Tests (Infrastructure Issues)
1. **vitest-pool-workers** - Network error (`ECONNRESET`)
2. **create-cloudflare e2e** - Authentication timeout (tries to open browser in CI)
3. **dev-registry** - Tail event timing issue (failed 3x, likely race condition)

These test different components (test infrastructure, CLI scaffolding, dev session communication) than the container deployment code I modified.

**Success rate:** 91/95 test suites passed (95.8%)

Could someone re-run the failed jobs? The container dry-run URL parsing fix is working correctly.

petebacondarwin:
The C3 tests will always fail right now due to this PR being on a fork. The other failures are probably just flakes.

petebacondarwin:
I'll rebase and rerun once we fix the C3 tests.

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -65,13 +65,14 @@ export async function deployContainers(
...
 
-	if (!dryRun) {
-		await fillOpenAPIConfiguration(
-			config,
-			isNonInteractiveOrCI(),
-			containersScope
-		);
-	}
+	// Always fill OpenAPI configuration when containers are defined, even in dry-run mode
+	// because maybeBuildContainer needs to call loadAccount() which requires API access
```
gpanders: I don't think we need this comment. It only makes sense in the context of this diff/PR, but doesn't provide insight to anyone reading the code directly (it will also immediately become stale if `maybeBuildContainer` changes).

gpanders (changes requested):
I'd like to suggest an alternative approach.

Right now, we don't fill out the OpenAPI configuration when we use dry run mode. This is actually a nice property because it gives us a kind of soft-guarantee that we won't interact with the API in dry run mode.

As you found, this breaks because `buildAndMaybePush` loads the user account limits using the API, even in dry run mode. However, it only does this to validate disk limits on the account before pushing.

I suggest changing `buildAndMaybePush` to only load the account and call `ensureDiskLimits` when dry run mode is disabled. That lets us maintain the property that we do not fill out the OpenAPI configuration in dry run mode and should only require changing `buildAndMaybePush`, which will be a much smaller and tighter scoped change.

packages/wrangler/src/cloudchamber/deploy.ts
hasip-timurtas: Removed

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -77,13 +77,22 @@ export async function buildAndMaybePush(
...
+		let cloudflareAccountID: string;
+
+		if (!dryRun) {
+			// account is also used to check limits below, so it is better to just pull the entire
+			// account information here
+			account = await loadAccount();
+			cloudflareAccountID = account.external_account_id;
+		} else {
+			cloudflareAccountID = "dry-run-account-id";
+		}
```
gpanders: Make `accountId` an argument to `buildAndMaybePush` and pass it in from the caller.

```suggestion
```

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -105,27 +114,30 @@ export async function buildAndMaybePush(
...
 		});
 
-		// ensure the account is not allowed to build anything that exceeds the current
-		// account's disk size limits
-		const inspectOutput = await dockerImageInspect(pathToDocker, {
-			imageTag,
-			formatString:
-				"{{ .Size }} {{ len .RootFS.Layers }} {{json .RepoDigests}}",
-		});
+		if (!dryRun && account) {
```
gpanders:
```suggestion
		if (!dryRun) {
			const account = await loadAccount();
```

packages/wrangler/src/cloudchamber/build.ts
gpanders: `buildAndMaybePush` is called from `buildCommand` and from `maybeBuildContainer`.

In `buildCommand`, get the account ID with

```typescript
const accountId = config.account_id || (await getAccountId(config));
```

then pass it into `buildAndMaybePush`.

In `maybeBuildContainer`, you'll need to add `accountId` as an argument to that function too and then pass it into `buildAndMaybePush`, and then update the call sites of `maybeBuildContainer` (there are only two, and both already have access to the account ID).

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -78,20 +79,11 @@ export async function buildAndMaybePush(
 	pathToDocker: string,
 	push: boolean,
 	containerConfig?: ContainerApp,
+	accountId?: string,
```
gpanders: This shouldn't be optional

```suggestion
	accountId: string,
```

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -78,20 +79,11 @@ export async function buildAndMaybePush(
...
-
-		if (!dryRun) {
-			// account is also used to check limits below, so it is better to just pull the entire
-			// account information here
-			account = await loadAccount();
-			cloudflareAccountID = account.external_account_id;
-		} else {
-			cloudflareAccountID = "dry-run-account-id";
-		}
+		const cloudflareAccountID = accountId || "dry-run-account-id";
```
gpanders: `accountId` shouldn't be optional, so this line isn't needed.

```suggestion
```

packages/wrangler/src/__tests__/cloudchamber/build.test.ts
```diff
@@ -39,7 +39,7 @@ const dockerfile =
 describe("buildAndMaybePush", () => {
 	runInTempDir();
 	mockApiToken();
-	mockAccountId();
+	mockAccountId({ accountId: "test_account_id" });
```
gpanders: Any reason to change this from the default?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -9335,6 +9335,7 @@ addEventListener('fetch', event => {});`
 						class_name: "ExampleDurableObject",
 						image: "./Dockerfile",
 					}),
+					"dry-run-account-id", // accountId
```
gpanders: We definitely do not want a special fake account id for dry runs. We should still use the user's real account ID (or in the case of a test, the mocked account ID).

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -770,26 +770,13 @@ See https://developers.cloudflare.com/workers/platform/compatibility-dates for m
...
-			}
-		}
-
 		if (props.dryRun) {
 			if (config.containers) {
 				for (const container of config.containers) {
 					await maybeBuildContainer(
 						container,
 						workerTag ?? "worker-tag",
+						accountId || "dry-run-account-id",
```
gpanders: Ok I see the problem here, when `props.dryRun` is set `accountId` can be (will be?) `undefined`.

That was changed in https://github.com/cloudflare/workers-sdk/pull/1037 (see https://github.com/cloudflare/workers-sdk/issues/1023). That was 3 years ago, and now that some of our containers commands rely on the account ID even for dry runs (e.g. for the container image name) I'm not sure it's correct to call that a bug anymore.

Maybe we only require login on dry run when certain commands (like building a container) are used. cc @rozenmd would like to hear your thoughts.

gpanders: One other possibility is to not use the account ID in the image name when building the image, and only re-tagging the image to include the account ID when we push (which we never do on dry runs).

gpanders: I implemented the above and pushed it to your fork @hasip-timurtas. See 3ed586a0.

gpanders:
I'll rebase this on https://github.com/cloudflare/workers-sdk/pull/9813 after that is merged, since we'll want to re-use some of the work in that PR.

emily-shen:
> I'd like to suggest an alternative approach.
> 
> Right now, we don't fill out the OpenAPI configuration when we use dry run mode. This is actually a nice property because it gives us a kind of soft-guarantee that we won't interact with the API in dry run mode.
> 
> As you found, this breaks because `buildAndMaybePush` loads the user account limits using the API, even in dry run mode. However, it only does this to validate disk limits on the account before pushing.
> 
> I suggest changing `buildAndMaybePush` to only load the account and call `ensureDiskLimits` when dry run mode is disabled. That lets us maintain the property that we do not fill out the OpenAPI configuration in dry run mode and should only require changing `buildAndMaybePush`, which will be a much smaller and tighter scoped change.

@gpanders  to play devils advocate, isn't making sure your image is within limits and will actually deploy something you really _do_ want to find out in a dry run?

gpanders:
> @gpanders to play devils advocate, isn't making sure your image is within limits and will actually deploy something you really do want to find out in a dry run?

Definitely, but we currently have no way to do that without making an API call, which we cannot do in dry run mode.

emily-shen:
looks like we don't validate the worker bundle size either during dry run, so we've got precedent at least. I don't mind either way, just thinking it through :)

gpanders (approved):
I've rebased on main and implemented the changes I suggested.

Thanks @hasip-timurtas for the issue report and initial work on this. Apologies for hijacking your PR, but in this case it was much faster to do it myself rather than go back and forth via GitHub. Thanks very much for your contribution :)

Issue #9759, opened 2025-06-26
@cloudflare/vite-plugin - support for vite v7


MarkSFrancis:
### What versions & operating system are you using?

@cloudflare/vite-plugin@1.7.5, wrangler@4.22.0, vite@7.0.0

### Describe the Bug

Currently, the `vite-plugin` only supports [`vite@^6.1.0`](https://github.com/cloudflare/workers-sdk/blob/main/pnpm-workspace.yaml#L35). Could this be changed to the specifier `^6.1.0 || ^7.0.0`? 

I have no idea what would be involved to ensure that v7 works with the plugin - I've not investigated the changes in vite@7

### Please provide any relevant error logs

```txt
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: ai@undefined
npm error Found: vite@7.0.0
npm error node_modules/vite
npm error   dev vite@"^7.0.0" from the root project
npm error
npm error Could not resolve dependency:
npm error peer vite@"^6.1.0" from @cloudflare/vite-plugin@1.7.5
npm error node_modules/@cloudflare/vite-plugin
npm error   dev @cloudflare/vite-plugin@"^1.7.5" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

jamesopstad:
Thanks for the issue. We've already made the changes to support Vite 7 and we'll update the peer dependency in the next release.

Issue #9733, opened 2025-06-24
Internal server error: __vite_ssr_exportName__ is not defined


contrebande-labs:
### What versions & operating system are you using?

Wrangler 4.21.1, Vite 7, React Router 7.6.2, React 19.1.0

### Describe the Bug

Updating a working Vite 6 project to Vite 7 fails to run in dev mode.

### Please provide any relevant error logs

Starting React Router in dev mode with `react-router dev` gives `4:07:48 PM [vite] Internal server error: __vite_ssr_exportName__ is not defined
      at async ProxyServer.fetch (file:///home/user/Documents/git/project/node_modules/.pnpm/miniflare@4.20250617.4/node_modules/miniflare/src/workers/core/proxy.worker.ts:173:11)`

chhch:
Maybe it has something to do with this https://github.com/vitejs/vite/pull/19996 change.

Destreyf:
This is related to Vite 7, I think @chhch is correct on the vite PR.

I had to rollback to vite 6.3.5 for now.

alicercedigital:
We got the same, any solutions?

Destreyf:
@alicercedigital The only solution currently is to roll back to 6.3.5.

As far as I can see, there's no pull requests that address this issue.  There is a version bump pull request but all it does is updates internal dependencies for the vite plugin, and does not address this issue. 

I don't think it'll be a difficult fix, but I haven't spent any time looking into it myself.  I don't think there's a huge need/rush for upgrading to vite 7, at least in my project.  My testing shows no differences in the build/dev process, but obviously it would be nice to be able to keep dependencies up to date.

jfsiii:
I'm curious if this solves the issue for @contrebande-labs , because they don't mention using the Cloudflare Vite plugin, but that's what changed in the linked PR.

I have the same issue as the OP and am also not using the Cloudflare Vite plugin.

Destreyf:
@jfsiii They are using vite with react/react-router, which means they need to add they vite-plugin to use workers in conjunction with it (in my experience anyways).

I had the exact same error message with the cloudflare plugin on vite 7, rolling back to 6.3.5 "fixed" it.

The updated plugin fixes the issue I had with cloudflare, although I have a couple other deps that aren't updated for vite 7 yet, although an override worked fine for them.

ducan-ne:
still getting this issue these days. Downgraded to 6.3.5 fixed it but I don't think it's long term solution

[...]

PR #9773, opened 2025-06-26
Support Vite 7

jamesopstad:
Fixes https://github.com/cloudflare/workers-sdk/issues/9759
Fixes https://github.com/cloudflare/workers-sdk/issues/9733

Adds support for Vite 7. I've also pinned the versions of Vite we use in the repo so that we are more explicit about what we're using and when we update. Vite 7.0.0 is used to build the plugin and run the playgrounds. Vite 6.1.0 is used in the E2E tests. This gives us some coverage for the oldest and most recent versions we support.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/e2e/fixtures/basic/package.json
+++ b/packages/vite-plugin-cloudflare/e2e/fixtures/basic/package.json
@@ -27,6 +27,6 @@
 		"globals": "^15.14.0",
 		"typescript": "~5.7.2",
 		"typescript-eslint": "^8.22.0",
-		"vite": "^6.1.0"
+		"vite": "6.1.0"
 	}
 }
--- a/packages/vite-plugin-cloudflare/e2e/fixtures/basic/package.json
+++ b/packages/vite-plugin-cloudflare/e2e/fixtures/basic/package.json
@@ -27,6 +27,6 @@
 		"globals": "^15.14.0",
 		"typescript": "~5.7.2",
 		"typescript-eslint": "^8.22.0",
-		"vite": "6.1.0"
+		"vite": "^6.1.0"
 	}
 }
--- a/.changeset/olive-rabbits-tell.md
+++ b/.changeset/olive-rabbits-tell.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": minor
+---
+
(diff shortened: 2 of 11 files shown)


packages/vite-plugin-cloudflare/e2e/fixtures/basic/package.json
```diff
@@ -27,6 +27,6 @@
 		"globals": "^15.14.0",
 		"typescript": "~5.7.2",
 		"typescript-eslint": "^8.22.0",
-		"vite": "^6.1.0"
+		"vite": "6.1.0"
```
dario-piotrowicz: why this change? 😮  (also for the other fixtures, don't we want to support the latest vite@6 versions?)

jamesopstad: I thought it would be better to be explicit about the version we are testing and have some coverage for the oldest version we support. Happy to change it to `6.3.5` (the final Vite 6 version) if you prefer.

dario-piotrowicz: mh... I see what you mean... but I would imagine that the most important part is supporting the latest v6 version, for example, if a user wanted to use 6.1.0 and say for some reason our plugin didn't properly work for that version, all they'd need to do would be bump their vite to 6.3.5 which should in theory be trivial right? if on the other hand we supported 6.1.0 but our plugin didn't work for 6.3.5, someone coming with 6.3.5 would need to downgrade to 6.1.0 right? which seems like a worse solution for them no? (as vite could have for example introduced features or bugfixed between 6.1.0 and 6.3.5) 🤔 

basically what I am trying to say I guess is, why would we want to be more confident that the plugin works on an older version or v6 or or a newer one? and I'd argue for the latter (especially if vite backports bugfixes etc... (which I have no idea if they do))

jamesopstad: Yeah, that makes perfect sense. Thanks! I'll make that change.

PR #9774, opened 2025-06-26
Validate instance type in wrangler

nikitassharma:
Fixes CC-5513

Validate instance type against account limits to give early feedback to the user.

---

- Tests
  - [X] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: improved error handling
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [X] Not necessary because: not a v3 feature


packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+function isValidKey(
+	instanceType: string
+): instanceType is keyof typeof instanceTypes {
+	return instanceType in instanceTypes;
+}
+
+// checks a given InstanceType against account limits
+export async function checkInstanceTypeAgainstLimits(
```
petebacondarwin: I don't believe this function needs to be async.

petebacondarwin (approved):
One small nit. But this looks good.

emily-shen (commented):
i noticed that the instance type is not being used in `ensureDiskLimits` when building and validating a container, is that intentional? i notice this is mostly targeting cloudchamber-only commands

packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
nikitassharma: I was using the same structure as `ensureDiskLimits`. The reason both are async is for the tests to be able to check what error is thrown

petebacondarwin: Do you mean tests like this?

```ts
	it("should throw error if instance type disk exceeds account limit", async () => {
		await expect(() =>
			checkInstanceTypeAgainstLimits(InstanceType.BASIC, {
				limits: {
					...commonLimits,
					disk_mb_per_deployment: 2000,
				},
			} as CompleteAccountCustomer)
		).rejects.toThrow(
			"Exceeded account limits: Your configured instance type uses 4000 MB of disk. However, that exceeds the account limit of 2000"
		);
	});
```

If you make this function not-async then these tests also become simpler, right?

```
	it("should throw error if instance type disk exceeds account limit", () => {
		expect(() =>
			checkInstanceTypeAgainstLimits(InstanceType.BASIC, {
				limits: {
					...commonLimits,
					disk_mb_per_deployment: 2000,
				},
			} as CompleteAccountCustomer)
		).toThrow(
			"Exceeded account limits: Your configured instance type uses 4000 MB of disk. However, that exceeds the account limit of 2000"
		);
	});
```

nikitassharma: Ah, I see! Thanks for the catch, fixed it

nikitassharma:
> i noticed that the instance type is not being used in `ensureDiskLimits` when building and validating a container, is that intentional?

@emily-shen thanks for the catch! I updated it so that `ensureDiskLimits` goes to the instance type disk size first before falling back on defaults

>  i notice this is mostly targeting cloudchamber-only commands

the changes to `apply` will be used by `wrangler deploy`

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -91,7 +104,17 @@ describe("cloudchamber apply", () => {
 
 	mockAccountId();
 	mockApiToken();
-	beforeEach(mockAccount);
+	beforeEach(() => {
+		mockAccount({
+			external_account_id: "test_account_id",
```
gpanders: Use `some-account-id` which is the standard mocked account ID used elsewhere (for some reason a lot of our tests were using `test_account_id` but I've been trying to phase that out).

```suggestion
			external_account_id: "some-account-id",
```

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -533,12 +538,17 @@ export async function apply(
 			appConfigNoDefaults.configuration.image = application.configuration.image;
 		}
 
+		const account = await loadAccount();
```
gpanders: We need to only do this when `dryRun` is not set. When using `--dry-run` mode we can't make any API calls. See https://github.com/cloudflare/workers-sdk/pull/9765 (which I'm working on now).

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -51,6 +52,18 @@ function mockCreateApplication(
...
+		http.get(
+			"*/me",
+			async () => {
+				return HttpResponse.json(account);
+			},
+			{ once: true }
+		)
+	);
+}
+
```
gpanders: Can we re-use the `mockAccount` function from `cloudchamber/utils.ts` instead of adding a new function?

packages/wrangler/src/cloudchamber/apply.ts
nikitassharma: updated to only load account when dry run is not set

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
nikitassharma: updated `mockAccount` in `utils.ts` so that it can be used

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -449,6 +454,7 @@ export async function apply(
 		skipDefaults: boolean | undefined;
 		env?: string;
 		imageUpdateRequired?: boolean;
+		dryRun: boolean;
```
emily-shen: we don't hit this code path at all in a dry run - wrangler deploy bails before `deployContainers` (which calls apply) is called

packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
emily-shen, line 1: nit but can you rename to `instanceType` 😅

packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
```diff
@@ -0,0 +1,185 @@
...
+
+function isValidKey(
+	instanceType: string
+): instanceType is keyof typeof instanceTypes {
+	return instanceType in instanceTypes;
+}
+
+// checks a given InstanceType against account limits
+export function checkInstanceTypeAgainstLimits(
+	instanceType: InstanceType | string | undefined,
```
emily-shen:
```suggestion
	instanceType: InstanceType | undefined,
```
i don't think it can be a string at this point.

packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
```diff
@@ -0,0 +1,185 @@
...
+		throw new UserError(
+			`Exceeded account limits: Your configured instance type uses ${usage.memory_mib} MiB of memory. However, that exceeds the account limit of ${memoryLimit}`
+		);
+	}
+
+	if (usage.disk_mb > diskLimit) {
+		throw new UserError(
+			`Exceeded account limits: Your configured instance type uses ${usage.disk_mb} MB of disk. However, that exceeds the account limit of ${diskLimit}`
+		);
+	}
```
emily-shen: presumably a user could hit multiple of these at the same time? would you be able to collect all limits they hit and error once with all them?

packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
```diff
@@ -0,0 +1,185 @@
...
+): instanceType is keyof typeof instanceTypes {
+	return instanceType in instanceTypes;
+}
+
+// checks a given InstanceType against account limits
+export function checkInstanceTypeAgainstLimits(
+	instanceType: InstanceType | string | undefined,
+	account: CompleteAccountCustomer
+) {
+	if (instanceType === undefined || !isValidKey(instanceType)) {
```
emily-shen: if instanceType is undefined, what does that mean? it looks like it can only be undefined for cloudchamber configs, where configuration.disk/vcpu/memory/memory_mib can be set individually. should those be validated as well?

nikitassharma: `instanceType` could be undefined if the user doesn't set it in the config-- in this case the API falls back on the default values for memory/cpu/disk which match a `dev` instance type

nikitassharma: I think it's fine to leave that validation to the API instead of having a faster failure in wrangler. The default user doesn't have permission to individually set those values and it wouldn't be a common path

packages/wrangler/src/cloudchamber/apply.ts
nikitassharma: reverted that change

packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
nikitassharma: renamed to `instance-type` to match what I was seeing in other files, lmk if `instanceType` matches the style better

packages/wrangler/src/cloudchamber/instancetype/instancetype.ts
nikitassharma: changed to collect the errors

packages/wrangler/src/cloudchamber/apply.ts
nikitassharma: reverted this change in response to https://github.com/cloudflare/workers-sdk/pull/9774#discussion_r2192437637

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -91,7 +92,17 @@ describe("cloudchamber apply", () => {
 
 	mockAccountId();
 	mockApiToken();
-	beforeEach(mockAccount);
+	beforeEach(() => {
+		mockAccount({
+			external_account_id: "some-account-id",
```
gpanders: Use the default, which uses the `CLOUDFLARE_ACCOUNT_ID` env var (I know I asked you to change it to this but I forgot that we had updated `mockAccount` to use the env var, so we should do that instead of hard coding here).

```suggestion
```

packages/wrangler/src/__tests__/cloudchamber/create.test.ts
```diff
@@ -89,7 +92,17 @@ describe("cloudchamber create", () => {
 
 	mockAccountId();
 	mockApiToken();
-	beforeEach(mockAccount);
+	beforeEach(() => {
+		mockAccount({
+			external_account_id: "some-account-id",
```
gpanders:
```suggestion
```

packages/wrangler/src/__tests__/cloudchamber/utils.ts
```diff
@@ -17,22 +18,20 @@ export function setWranglerConfig(cloudchamber: CloudchamberConfig) {
 	);
 }
 
-export function mockAccount() {
+export function mockAccount(
+	account: CompleteAccountCustomer = {
+		external_account_id: process.env.CLOUDFLARE_ACCOUNT_ID,
+		limits: { disk_mb_per_deployment: 2000 },
+	} as CompleteAccountCustomer
```
gpanders: Is the cast here necessary? Casting in TS silences all type errors/warnings and can hide actual bugs if not used carefully. Generally we should avoid using casts like this unless they're strictly necessary.

gpanders: There are quite a few `as CompleteAccountCustomer` casts in this PR, so same comment for all of those.

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -617,7 +584,8 @@ export function resolveMemory(
 // Return the amount of disk size in (MB) for an application, falls back to the account limits if the app config doesn't exist
 // sometimes the user wants to just build a container here, we should allow checking those based on the account limits if
 // app.configuration is not set
-// ordering: app.configuration.disk.size -> account.limits.disk_mb_per_deployment -> default fallback to 2GB in bytes
+// ordering:
+// app.instance_type[disk] -> app.configuration.disk.size -> account.limits.disk_mb_per_deployment -> default fallback to 2GB in bytes
```
gpanders: It looks like `account` isn't used at all in this function, so this comment is not correct and we can remove that argument completely.

packages/wrangler/src/cloudchamber/instance-type/instance-type.ts
```diff
@@ -0,0 +1,188 @@
...
+
+// infers the instance type from a given configuration
+function inferInstanceType(
+	config: UserDeploymentConfiguration
+): InstanceType | undefined {
+	for (const [instanceType, configuration] of Object.entries(instanceTypes)) {
+		if (
+			config?.vcpu === configuration.vcpu &&
+			config?.memory_mib === configuration.memory_mib &&
+			config?.disk?.size_mb === configuration.disk_mb
```
gpanders: `config` cannot be `undefined` according to the type definition

```suggestion
			config.vcpu === configuration.vcpu &&
			config.memory_mib === configuration.memory_mib &&
			config.disk?.size_mb === configuration.disk_mb
```

packages/wrangler/src/__tests__/cloudchamber/utils.ts
nikitassharma: I removed some of the casts by marking `CompleteAccountCustomer` as `Partial` in `mockAccount`. However, I left the casts in the tests for `checkInstanceTypeAgainstLimits`. Adding all of the unrelated fields needed to remove the cast would add a lot of noise to the tests

.changeset/twenty-seas-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Validate instance type against account limits in wrangler to give early feedback to the user
```
emily-shen:
```suggestion
Validate container instance type against account limits in wrangler to give early feedback to the user
```

emily-shen (commented):
can we move all of this to

nikitassharma:
Refactored the PR to update functionality. Here's a summary of the goal of these changes:
1. Maintains changes from the original PR to refactor instance type processing into `instance-type/instance-type.ts`
2. As a change from the original PR, updates validation against account limits to work for both instance types and custom limits. This is done in `ensureContainerLimits`
3. As a change from the original PR, we are no longer validating the configuration against account limits for cloudchamber only commands
4. Consolidates `ensureDiskLimits` with the other limit checks. A large amount of the functionality here was moved into `ensureContainerLimits` and the remainder was moved to a new function `ensureImageFitsLimits`

ichernetsky-cf (approved):

gpanders (approved):

packages/wrangler/src/__tests__/cloudchamber/build.test.ts
```diff
@@ -36,7 +32,9 @@ describe("buildAndMaybePush", () => {
 
 	beforeEach(() => {
 		vi.clearAllMocks();
-		vi.mocked(dockerImageInspect).mockResolvedValue("53387881 2 []");
+		vi.mocked(dockerImageInspect)
+			.mockResolvedValueOnce("[]")
+			.mockResolvedValueOnce("53387881 2");
```
emily-shen:
```suggestion
		vi.mocked(dockerImageInspect)
			// return empty array of repo digests (i.e. image does not exist remotely)
			.mockResolvedValueOnce("[]")
			// return image size and number of layers
			.mockResolvedValueOnce("53387881 2");
```

packages/wrangler/src/__tests__/cloudchamber/limits.test.ts
```diff
@@ -0,0 +1,241 @@
...
+	return Object.assign({}, actual, {
+		dockerImageInspect: vi.fn(),
+	});
+});
+
+describe("ensureContainerLimits", () => {
+	beforeEach(() => {
+		vi.clearAllMocks();
+
+		// required size = 53387881 * 1.1 + 2 * 16 * MiB = 93MB
```
emily-shen, line 33:
```suggestion
		// Returns image size and number of layers - used to calculate required size.
		// required size = 53387881 * 1.1 + 2 * 16 * MiB = 93MB
```

packages/wrangler/src/__tests__/cloudchamber/utils.ts
```diff
@@ -17,22 +18,24 @@ export function setWranglerConfig(cloudchamber: CloudchamberConfig) {
...
 }
 
-export function mockAccount() {
+type DeepPartial<T> = {
+	[P in keyof T]?: DeepPartial<T[P]>;
+};
+
+export function mockAccount(
+	account: DeepPartial<CompleteAccountCustomer> = {
+		external_account_id: process.env.CLOUDFLARE_ACCOUNT_ID,
```
emily-shen, line 27: can we use a random placeholder like `test-account-id` instead? (i know this code's just been moved in this PR, but while we're in the area...)

packages/wrangler/src/cloudchamber/build.ts
emily-shen, line 1: i left this comment previously somewhere in here (around line 240):
```
		// this means we won't be able to read the disk size from the config, but that option is deprecated anyway at least for containers.
		// and this never actually worked for cloudchamber as this command was previously reading it from config.containers not config.cloudchamber
```
think it needs to be update to reflect that you can set custom limits now, but we're only validating image size and not the other limits in the `containers build` command.

packages/wrangler/src/cloudchamber/limits.ts
```diff
@@ -0,0 +1,115 @@
...
+	pathToDocker: string;
+	imageTag: string;
+	account: CompleteAccountCustomer;
+	containerConfig?: ContainerNormalizedConfig;
+}): Promise<void> {
+	const limits = accountToLimits(options.account);
+	if (!options.containerConfig) {
+		// in this case we are only building an image-- we still want to check that the image fits
+		// within the account limits on disk size
+		await ensureImageFitsLimits({
```
emily-shen, line 51: can we pass in the errors array to ensureImageFitsLimits too so we can error once if the image is too big and the other limits are exceeded?

also then we would be able to simplify slightly to:

```
await ensureImageFitsLimits() 
if (options.containerConfig) {check others}
if(errors.length) throw
```

packages/wrangler/src/cloudchamber/limits.ts
```diff
@@ -0,0 +1,115 @@
...
+		return;
+	}
+
+	const usage = configToUsage(options.containerConfig);
+
+	// test configuration against account limits
+	const errors = [];
+	if (usage.vcpu > limits.vcpu) {
+		errors.push(
+			`Your container configuration uses ${usage.vcpu} vCPU which exceeds the account limit of ${limits.vcpu}.`
```
emily-shen:
```suggestion
			`Your container configuration uses ${usage.vcpu} vCPU which exceeds the account limit of ${limits.vcpu} vCPU.`
```

emily-shen (approved):

packages/wrangler/src/__tests__/cloudchamber/utils.ts
nikitassharma: I think there's been an effort to consolidate to using `some-account-id` which is the reason I left it this way here (https://github.com/cloudflare/workers-sdk/pull/9774#discussion_r2190907293 and https://github.com/cloudflare/workers-sdk/pull/9774#discussion_r2192887410)

packages/wrangler/src/cloudchamber/limits.ts
nikitassharma: I left it separate is because when the containerConfig is undefined, the image is being checked against account limits, but when the containerConfig is defined, the image size is being checked against the configured limits. because of this, the `Exceeded account limits` error message isn't accurate for all cases

packages/wrangler/src/cloudchamber/build.ts
nikitassharma: updated the comment

packages/wrangler/src/cloudchamber/limits.ts
emily-shen: Oh I totally missed that, fair enough!

PR #9775, opened 2025-06-27
Cap the number od errors and warnings for bulk KV put

### LINKED ISSUE #9498: Bulk putting big JSON file with malformed records burns the CPU

Wrangler 4.19.1, macOS Sequoia 15.5, Node 20.16.0, aarch64

_No response_

### Describe the Bug

Putting a big JSON file with malformed records using `bulk put` will freeze the Wrangler when it's supposed to crash or print out the error log (same symptoms for both local and remote), I have to `SIGTERM` to stop the process:

Image

If I keep the process running, it will burns the CPU up, makes my MacBook Air get super hot shortly after (90-100 DegC), which doesn't make any sense.

Image

To generate a malformed record, please use this script:

```python3
import json
import uuid
import random
record_count = 500000
json_data = [{
    # malformed key
    "key": random.randint(1, record_count),
    # random bogus value
    "value": uuid.uuid4().hex + uuid.uuid4().hex + uuid.uuid4().hex + uuid.uuid4().hex
} for _ in range(record_count)]

with open('test.json', 'w') as f:
    json.dump(json_data, f)
```

then use the `test.json` generated above with bulk put command:
```sh
npx wrangler kv bulk put --namespace-id SOME_NAMESPACE ./test.json
```

_No response_

vicb:
Fixes #9498

Queue a lot of errors/warnings can fill up all the memory and slow down the process as described in the linked issue. 

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not needed
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


.changeset/shaggy-frogs-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Cap the number of errors and warnings for bulk KV put
```
dario-piotrowicz: Maybe add some information as to why we want to cup the number of errors?

packages/wrangler/src/kv/helpers.ts
```diff
@@ -17,6 +17,9 @@ const API_MAX = 10000;
 // The const below are lowered from the API's true capacity to help avoid
 // hammering it with large requests.
 export const BATCH_KEY_MAX = API_MAX / 10;
+// Limit the number of errors to logs during a bulk put.
```
dario-piotrowicz:
```suggestion
// Limit the number of errors or warnings to logs during a bulk put.
```

packages/wrangler/src/kv/index.ts
```diff
@@ -816,20 +817,32 @@ export const kvBulkPutCommand = createCommand({
...
 		const errors: string[] = [];
+		let maxNumberOfWarningsReached = false;
 		const warnings: string[] = [];
 		for (let i = 0; i < content.length; i++) {
 			const keyValue = content[i];
-			if (!isKVKeyValue(keyValue)) {
-				errors.push(`The item at index ${i} is ${JSON.stringify(keyValue)}`);
+			if (!isKVKeyValue(keyValue) && !maxNumberOfErrorsReached) {
+				if (errors.length === BATCH_MAX_ERRORS_WARNINGS) {
+					maxNumberOfErrorsReached = true;
```
dario-piotrowicz, line 828: instead of using this `maxNumberOfErrorsReached` flag, have you considered just `break`ing and exiting the loop? I feel like it might make sense to just quite after the few initial errors? is there any benefit in keeping looping through all the remaining entries?

.changeset/shaggy-frogs-tap.md
vicb:
```suggestion
Cap the number of errors and warnings for bulk KV put to avoid consuming too much memory
```

packages/wrangler/src/kv/index.ts
vicb: That's true, we could break if both `maxNumberOfErrorsReached` and `maxNumberOfWarningsReached` are reached. But it's probably not worth adding more code to optimize an error case?

PR #9781, opened 2025-06-27
chore: Force fail E2Es from fork PRs (attempt #2)

CarmenPopoviciu:
Fixes n/a

This is attempt no.2, since https://github.com/cloudflare/workers-sdk/pull/9726 didn't solve the problem :/

E2E tests cannot run on pull requests created from forks due to security concerns. In such cases we require a member of 
the `workers-sdk` maintainers group to manually run these tests on behalf of the PR.

In such cases we want CI to fail with a descriptive message, instead of skipping the tests, which gives the false optics that 
CI is green. Having CI intentionally fail for such use cases will act as a reminder that tests need to be manually run, and 
will prevent us from accidentally merging fork PRs that are green, but never ran the e2e tests.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: gh workflow change
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: gh workflow change
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: gh workflow change

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   e2e-wrangler-test:
-    name: ${{ format('Wrangler ({0})', matrix.description) }}
+    name: ${{ format('Wrangler E2E ({0})', matrix.description) }}
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.node }}-wrangler
       cancel-in-progress: true
@@ -27,6 +27,21 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
+      # E2E tests cannot run on pull requests created from forks due to security concerns.
+      # In such cases we require a member of the `workers-sdk` maintainers group to manually
+      # run these tests on behalf of the PR.
+      #
+      # In such cases we want CI to fail with a descriptive message, instead of skipping the
+      # tests, which gives the false optics that CI is green. Having CI intentionally fail
+      # for such use cases will act as a reminder that tests need to be manually run, and will
+      # prevent us from accidentally merging fork PRs that are green, but never ran the e2e
+      # tests.
+      - name: Check if fork PR
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
+        run: |
+          echo: "E2E tests cannot run on pull requests created from forks due to security reasons. Please reach out to a workers-sdk maintainer to run the E2E tests on behalf of this PR (via https://github.com/cloudflare/workers-sdk/actions/workflows/run-ci-for-external-forks.yml)."
+          exit 1
+
       - name: Checkout Repo
         if: github.event.pull_request.head.repo.owner.login == 'cloudflare'
         uses: actions/checkout@v4
@@ -72,7 +87,7 @@ jobs:
           path: .turbo/runs
 
   e2e-vite-test:
-    name: ${{ format('Vite ({0})', matrix.description) }}
+    name: ${{ format('Vite E2E ({0})', matrix.description) }}
     concurrency:
       group: ${{ github.workflow }}-${{ github.ref }}-${{ matrix.os }}-${{ matrix.node }}-vite
       cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -91,6 +106,21 @@ jobs:
             node: 20.19.1
     runs-on: ${{ matrix.os }}
     steps:
+      # E2E tests cannot run on pull requests created from forks due to security concerns.
```
CarmenPopoviciu, line 109: I'm assuming this applies for vite e2es tests as well?

.github/workflows/e2e.yml
```diff
@@ -6,7 +6,7 @@ on:
 
 jobs:
   e2e-wrangler-test:
-    name: ${{ format('Wrangler ({0})', matrix.description) }}
+    name: ${{ format('Wrangler E2E ({0})', matrix.description) }}
```
penalosa, line 9:
```suggestion
    name: ${{ format('Wrangler ({0})', matrix.description) }}
```

This test shows up in CI as `E2E / Wrangler` already

.github/workflows/e2e.yml
```diff
@@ -27,6 +27,21 @@ jobs:
...
+      # E2E tests cannot run on pull requests created from forks due to security concerns.
+      # In such cases we require a member of the `workers-sdk` maintainers group to manually
+      # run these tests on behalf of the PR.
+      #
+      # In such cases we want CI to fail with a descriptive message, instead of skipping the
+      # tests, which gives the false optics that CI is green. Having CI intentionally fail
+      # for such use cases will act as a reminder that tests need to be manually run, and will
+      # prevent us from accidentally merging fork PRs that are green, but never ran the e2e
+      # tests.
+      - name: Check if fork PR
```
CarmenPopoviciu, line 39: instead of making this a separate job, I instead opted for converting what I had into a step for each of the wrangler/vite e2e jobs. This means that before running the actual e2e tests, we first check if this is a PR made from a fork. If so, we fail early, and the e2e job will appear in CI as 🔴 (at least that is what I'm hoping will happen)

.github/workflows/e2e.yml
```diff
@@ -27,6 +27,21 @@ jobs:
...
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository
+        run: |
+          echo: "E2E tests cannot run on pull requests created from forks due to security reasons. Please reach out to a workers-sdk maintainer to run the E2E tests on behalf of this PR (via https://github.com/cloudflare/workers-sdk/actions/workflows/run-ci-for-external-forks.yml)."
+          exit 1
+
```
penalosa, line 44: Do we definitely need this anymore? Now we run E2E on every PR this will fail _all_ fork PRs. We should be running E2E in the merge queue, so fork PRs won't get past the merge queue with failing E2E

.github/workflows/e2e.yml
```diff
@@ -72,7 +87,7 @@ jobs:
           path: .turbo/runs
 
   e2e-vite-test:
-    name: ${{ format('Vite ({0})', matrix.description) }}
+    name: ${{ format('Vite E2E ({0})', matrix.description) }}
```
penalosa, line 90: As above

.github/workflows/e2e.yml
CarmenPopoviciu: It actually shows up like this 👇 (see https://github.com/cloudflare/workers-sdk/actions/runs/15922705635/job/44912932635?pr=9761)

Screenshot 2025-06-27 at 17 06 23

For me personally it is not immediately clear what job I am looking at. If I have multiple tabs open with CI detail views, which happens very often, there's just added extra readability friction

I won't die on this hill tho

.github/workflows/e2e.yml
CarmenPopoviciu: hmmm...looking at https://github.com/cloudflare/workers-sdk/actions/runs/15922705635/job/44912932635?pr=9761 there are no e2e failures. In fact CI is entirely green. Am I doing smth funky with that PR?

CarmenPopoviciu: oh...so is your argument that even though CI is green for the PR itself, adding it to the merge queue will prompt the e2e failures?

CarmenPopoviciu: personally I'd like to have this feedback immediately, but if it's just me, I won't die on this hill either. In this case this PR is irrelevant

petebacondarwin: I see on that linked job above:

```
Error when evaluating 'runs-on' for job 'e2e-test-forks'. .github/workflows/e2e.yml (Line: 150, Col: 14): Unexpected value ''
```

The line in question:

```
runs-on: ${{ matrix.os }}
```

Probably because that step wasn't part of the matrix.

This suggests that the job was skipped because the yml file was syntactically invalid.

But I would have hoped it would have failed rather than passed...
I wonder if it passed because the "upload turbo logs" step still ran due to:

```
     if: always()
```

and because that was valid it passed??

CarmenPopoviciu: that is indeed why that job I added previously failed. If your assumption is correct, then the current setup won't fail the job either. If we agree we actually want to keep the code I have here, I'll check the gh docs again to see what's needed.

CarmenPopoviciu:
will go ahead and close this as we'll be relying on e2es being run when in the merge queue

PR #9785, opened 2025-06-27
Also handle next apps hosted on a subpath

matthewdavidrodgers:
Next apps can be hosted on a subpath, i.e. not at the root, so we need to be a little more permissive with the path we check on.

_Describe your change..._

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: small patch change
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  not a wrangler change

### DIFF
--- a/packages/workers-shared/router-worker/tests/index.test.ts
+++ b/packages/workers-shared/router-worker/tests/index.test.ts
@@ -247,10 +247,19 @@ describe("unit tests", async () => {
 		expect(await response.text()).toEqual("hello from user worker");
 	});
 
-	it("blocks /_next/image requests with remote URLs when not fetched as image", async () => {
-		const request = new Request(
-			"https://example.com/_next/image?url=https://evil.com/ssrf"
-		);
+	it.each([
+		{
+			description:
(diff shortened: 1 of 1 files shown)


packages/workers-shared/router-worker/tests/index.test.ts
```diff
@@ -272,6 +272,31 @@ describe("unit tests", async () => {
 		expect(await response.text()).toBe("Blocked");
 	});
 
+	it("blocks /_next/image requests with remote URLs when not fetched as image for next apps host not at the root", async () => {
```
vicb: Maybe:

```suggestion
	it("blocks /base/path/_next/image requests with remote URLs when not fetched as image", async () => {
```

I think using [`it.each`](https://vitest.dev/api/#test-each) to parametrize the test by path would be nice here to remove duplication.

vicb (approved):

petebacondarwin (approved):

PR #9791, opened 2025-06-29
Add support for Bun WebSockets to miniflare

kylecarbs:
Bun does not support the upgrade event, but open can be used since most of this function is just piping responses.

This does not support the use-case of a client reading headers from the WebSocket response, because Bun does not expose that.

That seems like an edge-case though, so I think merging partial support is better than none.

See: https://github.com/oven-sh/bun/issues/5951

### DIFF
--- a/packages/miniflare/src/http/fetch.ts
+++ b/packages/miniflare/src/http/fetch.ts
@@ -56,6 +56,14 @@ export async function fetch(
 			rejectUnauthorized = { rejectUnauthorized: false };
 		}
 
+		// @ts-expect-error bun is not typed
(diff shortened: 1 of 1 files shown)


penalosa:
Thanks for opening this! However, Wrangler & Miniflare do not support alternative runtimes like Bun & Deno (and we don't intend to add support), and so I'm going to close this for now.

In particular, the maintenance and testing burden would be too high (this is also why we don't support Deno). While Bun has relatively good Node compat, Wrangler (and Miniflare) are fairly complex applications which exercise a lot of the Node.js API surface, and it's unclear that the additional testing overhead to make sure things work in Bun would be worthwile.

PR #9813, opened 2025-07-01
Support image name without account ID

gpanders:
Fixes CC-5603

If a user uses an image in the Cloudflare managed registry they can specify the image name without including their full account ID. For example:

  image = "registry.cloudflare.com/hello/world:1.0"

will now "just work", as Wrangler will automatically include their account ID in the appropriate location when creating the application. This matches the new behavior in the Dash, where we also omit the account ID in the image name.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23427
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: experimental command


packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -417,6 +420,33 @@ function sortObjectRecursive<T = Record<string | number, unknown>>(
...
 
+// Resolve an image name to the full unambiguous name.
+//
+// For now, this only converts images stored in the managed registry to contain
+// the user's account ID in the path.
+async function resolveImageName(
+	config: Config,
+	image: string
+): Promise<string> {
+	// If the image is a relative path return it immediately
```
emily-shen: is a check for whether its a dockerfile? there is an `isDockerfile` function for that. since i guess you might provide an absolute path..?

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -177,7 +177,8 @@ export async function promiseSpinner<T>(
 		message: "Loading",
 	}
 ): Promise<T> {
-	if (json) {
+	const interactive = !json && !isNonInteractiveOrCI();
+	if (!interactive) {
```
emily-shen:
```suggestion
	if (json || isNonInteractiveOrCI()) {
```
to avoid the double negative :)

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -102,9 +116,8 @@ describe("cloudchamber apply", () => {
 			],
 		});
 		mockGetApplications([]);
-		mockCreateApplication({ id: "abc" } as Application);
-		await runWrangler("cloudchamber apply --json");
-		/* eslint-disable */
+		mockCreateApplication({ id: "abc" });
+		await runWrangler("cloudchamber apply");
```
emily-shen, line 119: why remove the json flag? i guess it looks like it wasn't working anyway...? should it?

emily-shen: oh i see the follow up PR

gpanders: Use of `--json` here was a hack to force non-interactive mode, which is the wrong way to do this (the `cloudchamber apply` command should be checking for interactivity based on a connected TTY, not just using the `--json` flag).

packages/wrangler/src/cloudchamber/apply.ts
gpanders: Hm `isDockerfile` does a lot more than I need and also throws errors, which I'd need to add handling for.

I don't think I need this check. The `if (url.hostname !== registry)` check will fail for a relative/absolute path anyway.

packages/wrangler/src/cloudchamber/common.ts
gpanders: 5624941d

packages/wrangler/src/cloudchamber/apply.ts
gpanders: 6b0f13e9

gpanders:
Resolved conflicts and rebased on main

emily-shen (approved):

PR #9811, opened 2025-07-01
wrangler containers images delete should prefix the account id by default

gpanders:
Fixes CC-5572.
Fixes https://github.com/cloudflare/workers-sdk/issues/9837

When listing container images with `wrangler container images list`, we omit the account ID prefix that each container image has in the managed registry. When deleting images we also need to include the account ID.

Also fix an issue with the "containers images list" command where the final line of output was being truncated.

This PR contains two other minor "housekeeping" commits which don't merit their own PR, but were broken out into separate commits to make reviewing easier and to keep the log cleaner:

* The source files for the `containers images` command were incorrectly named. Rename them so that `registries.ts` contains the implementations for the `registries` subcommand and `images.ts` contains the implementations for the `images` subcommand.
* Tidy up casing and phrasing for some strings in help text output

---

- Tests
  - [x] Tests included
  - [ ] Tests not necessary because:
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Bug fix
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not in v3

PR #9815, opened 2025-07-01
Remove --json arg from cloudchamber and containers commands

gpanders:
Fixes CC-5611.

This is a relic from an earlier time. The only command for which this is useful is the "cloudchamber curl" subcommand, in which case it should just be the default and non-configurable anyway.

~~It might make sense in the future to have a `--json` flag for commands like `containers images list`, but if we do add that it should be well-scoped and have proper user requirements, rather than spreading over all of the cloudchamber/containers commands. But for now the output of this command is well-formatted and is easy to parse (same as `docker images`).~~ We will keep it for `images list` too. It's trivial to support.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: minor change CLI (removing non-public option)
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: experimental commands


packages/wrangler/src/__tests__/cloudchamber/images.test.ts
emily-shen, line 1: can you unmock the print banner function (`vi.unmock("../wrangler-banner");`) and add a test to see the banner doesn't print with --json? (historically, the --json flag has been really easy to accidentally break 🥲 )

probably to curl as well.

.changeset/tender-cups-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Remove --json flag from containers and cloudchamber commands
```
emily-shen: can you note that image and curl still have --json

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -420,9 +421,9 @@ function sortObjectRecursive<T = Record<string | number, unknown>>(
 export async function apply(
 	args: {
 		skipDefaults: boolean | undefined;
-		json: boolean;
 		env?: string;
 		imageUpdateRequired?: boolean;
+		skipPrompts?: boolean;
```
emily-shen: can this be configured anywhere and if not, is it necessary? looks like its hardcoded to true in one place

packages/wrangler/src/containers/containers.ts
```diff
@@ -62,10 +54,13 @@ export async function deleteCommand(
...
 	}
 
-	const [, err] = await wrap(
-		ApplicationsService.deleteApplication(deleteArgs.ID)
-	);
-	if (err) {
+	try {
+		await ApplicationsService.deleteApplication(deleteArgs.ID);
+	} catch (err) {
+		if (!(err instanceof Error)) {
```
emily-shen, line 60: is this backwards? shouldn't we throw if it _is_an error?

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -110,47 +108,21 @@ export function handleFailure<
...
+				: "open-beta";
+			logger.warn(constructStatusMessage(command, commandStatus));
 		}
+		const config = readConfig(args);
+		await fillOpenAPIConfiguration(config, scope);
+		await cb(args, config);
 	};
 }
 
-export async function loadAccountSpinner({ json }: { json?: boolean }) {
```
emily-shen, line 150: why don't we need this anymore?

packages/wrangler/src/cloudchamber/apply.ts
gpanders: It's necessary to maintain current behavior with how `wrangler deploy` works, which is always non-interactive.

`wrangler cloudchamber apply` determines interactivity based on a connected TTY, but `wrangler deploy` always skips prompts (this is current behavior).

packages/wrangler/src/cloudchamber/common.ts
gpanders: The only places it was used were not needed -- after removing those call sites there were none left, so I removed the function completely.

packages/wrangler/src/containers/containers.ts
gpanders: This is just here to satisfy the type checker, since the type of `err` in the `catch` block is `any`. This condition lets us assume that the type of `err` is `Error` in the rest of the block.

Regardless of the type though, we re-throw the error here no matter what. We just reformat the message first.

.changeset/tender-cups-grin.md
gpanders: 230626c9

packages/wrangler/src/__tests__/cloudchamber/images.test.ts
gpanders: Ok done in c87a86e9, let me know if I did it wrong.

packages/wrangler/src/cloudchamber/apply.ts
emily-shen: ah i see now - could you add a comment to skipPrompts to explain this is for deploy?

packages/wrangler/src/containers/containers.ts
emily-shen: ah i see, missed the bit below outside the diff 😅

emily-shen (approved):

packages/wrangler/src/cloudchamber/apply.ts
gpanders: Discussed in chat. We'll instead change `apply` to always be non-interactive, so we don't need `skipPrompts` after all.  0cf6456d

gpanders:
I will merge this PR before https://github.com/cloudflare/workers-sdk/pull/9813 and https://github.com/cloudflare/workers-sdk/pull/9811 to make resolving conflicts slightly easier.

.changeset/tender-cups-grin.md
gpanders: Updated again to remove reference to `curl`, since `curl` doesn't have `--json` (only `images list` does).

Issue #9730, opened 2025-06-24
Feature Request: Automatically update the .jsonc Wrangler configuration file when a resource is created


thomasgauvin:
### Describe the solution

If I create a D1 database, KV namespace, Hyperdrive, Queues, R2, etc. resource within a project, Wrangler should provide me the option to automatically update the .jsonc file with the new binding 

We would need to support naming conflicts, ie multiple bindings with the same binding "HYPERDRIVE", "HYPERDRIVE_2"

dario-piotrowicz:
Sounds like a sensible idea to me 🙂 :+1:

GregBrimble:
@shaunpersad wrote a utility for replacing TOML values (preserving structure and comments): https://github.com/shaunpersad/toml-json-path-replacer. Perhaps it could be extended to support adding new values?

dario-piotrowicz:
> [@shaunpersad](https://github.com/shaunpersad) wrote a utility for replacing TOML values (preserving structure and comments): https://github.com/shaunpersad/toml-json-path-replacer. Perhaps it could be extended to support adding new values?

I actually think that the updating TOML/json(c) part to potentially be the simplest part here 🤔  (for example, if we were to scope this to just json(c)s, we could very easily parse, update and serialize json5 text I think, right?)

The trickier bit to me sounds deciding on the best/more sensible behavior and DX (how it should work in interactive mode, non-interactive mode, if there should be CLI flags for it, how to potentially handle naming conflicts, etc...)

Nothing really difficult but not super quick to implement and this indeed might need some design decisions? 🤔

emily-shen:
> [@shaunpersad](https://github.com/shaunpersad) wrote a utility for replacing TOML values (preserving structure and comments): https://github.com/shaunpersad/toml-json-path-replacer. Perhaps it could be extended to support adding new values?

👀 we actually already have a working implementation of this feature for json + jsonc already ([here](https://github.com/cloudflare/workers-sdk/blob/f6c8930ca544f3acc2d2bbab0fb7c1ea0b31967d/packages/wrangler/src/config/patch-config.ts#L9)), if we can integrate toml + comments then we're pretty much most of the way there then!

we just haven't had the capacity to add this in yet (this was originally made for the VSCode plugin, may it rest in peace). maybe if hyperdrive wants to be the guinea pig, it shouldn't be a huge amount of effort to get in :)

thomasgauvin:
Makes sense, I think:
[...]

PR #9816, opened 2025-07-01
thomasgauvin: support automatic updates of wrangler.jsonc when creating a new resource in a worker project

thomasgauvin:
addresses https://github.com/cloudflare/workers-sdk/issues/9730

1. Only support wrangler.jsonc. Supports KV, R2, D1, Hyperdrive, Vectorize
3. Make updating wrangler.jsonc optional, interactive, logs to console otherwise
4. Provide a flag to permit updating wrangler.jsonc from initial command line (--update-config)
5. Try to add binding and add _2, _3... to binding variable in event of conflict
6. Don't support auto-adding preview KV namespaces
7. Preserves comments and structure of wrangler.jsonc

### DIFF
--- a/packages/wrangler/src/config/auto-update-helpers.ts
+++ b/packages/wrangler/src/config/auto-update-helpers.ts
@@ -1,0 +1,243 @@
+import { confirm, prompt } from "../dialogs";
+import { logger } from "../logger";
+import { isValidIdentifier } from "../type-generation";
+import { formatConfigSnippet } from "./index";
+import type { ResourceBinding } from "./auto-update";
+import type { RawConfig } from "./config";
+
+// Registry of resource configuration patterns for better maintainability
+export const RESOURCE_CONFIG_REGISTRY = {
+	d1_databases: {
+		displayName: "D1 Database",
+		genericBindingName: "DB",
+		identifierField: "database_id" as const,
+		createConfig: (resource: ResourceBinding, bindingName: string) => ({
+			binding: bindingName,
+			database_name: resource.name,
+			database_id: resource.id,
+		}),
+	},
+	r2_buckets: {
+		displayName: "R2 Bucket",
+		genericBindingName: "BUCKET",
+		identifierField: "bucket_name" as const,
+		createConfig: (resource: ResourceBinding, bindingName: string) => ({
+			binding: bindingName,
+			bucket_name: resource.name,
+		}),
+	},
+	kv_namespaces: {
+		displayName: "KV Namespace",
+		genericBindingName: "KV",
+		identifierField: "id" as const,
+		createConfig: (resource: ResourceBinding, bindingName: string) => ({
+			binding: bindingName,
+			id: resource.id,
+		}),
+	},
+	vectorize: {
+		displayName: "Vectorize Index",
+		genericBindingName: "VECTORIZE",
+		identifierField: "index_name" as const,
+		createConfig: (resource: ResourceBinding, bindingName: string) => ({
+			binding: bindingName,
+			index_name: resource.name,
+		}),
+	},
+	hyperdrive: {
+		displayName: "Hyperdrive Configuration",
+		genericBindingName: "HYPERDRIVE",
+		identifierField: "id" as const,
+		createConfig: (resource: ResourceBinding, bindingName: string) => ({
+			binding: bindingName,
+			id: resource.id,
+		}),
+	},
+} as const;
+
+export function createBindingConfig(
+	resource: ResourceBinding,
+	bindingName: string
+): Record<string, unknown> {
+	const configTemplate = RESOURCE_CONFIG_REGISTRY[resource.type];
+	if (!configTemplate) {
+		throw new Error(`Unsupported resource type: ${resource.type}`);
+	}
+	return configTemplate.createConfig(resource, bindingName);
+}
+
+export function getBindingIdentifier(
+	binding: Record<string, unknown>,
+	type: ResourceBinding["type"]
+): string | undefined {
+	const configTemplate = RESOURCE_CONFIG_REGISTRY[type];
+	if (!configTemplate) {
+		return undefined;
+	}
+
+	const field = configTemplate.identifierField;
+	return binding[field] as string;
+}
+
+/**
+ * Checks if a binding name conflicts with existing bindings across all resource types.
+ * Binding names are case-insensitive in JavaScript.
+ */
+export function hasBindingNameConflict(
+	config: RawConfig,
+	bindingName: string
+): boolean {
+	const normalizedName = bindingName.toUpperCase();
+
+	// Check all resource types for existing bindings
+	for (const resourceType of Object.keys(RESOURCE_CONFIG_REGISTRY)) {
+		const bindings = config[
+			resourceType as keyof typeof RESOURCE_CONFIG_REGISTRY
+		] as Array<Record<string, unknown>> | undefined;
+		if (bindings) {
+			for (const binding of bindings) {
+				if (binding.binding && typeof binding.binding === "string") {
+					if (binding.binding.toUpperCase() === normalizedName) {
+						return true;
+					}
+				}
+			}
+		}
+	}
+
+	return false;
+}
+
+/**
+ * Gets the display name for a resource type from the registry.
+ */
+export function getResourceDisplayName(
+	resourceType: ResourceBinding["type"]
+): string {
+	const config = RESOURCE_CONFIG_REGISTRY[resourceType];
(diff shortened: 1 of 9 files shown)


packages/wrangler/src/config/auto-update.ts
```diff
@@ -0,0 +1,316 @@
...
+	configPath: string,
+	resource: ResourceBinding,
+	config: RawConfig
+) {
+	// Read the original file content to preserve formatting and comments
+	const originalContent = readFileSync(configPath);
+	let edits: any[] = [];
+
+	// Generate a unique binding name that doesn't conflict
+	const bindingName = generateUniqueBindingName(config, resource);
```
dario-piotrowicz: Why do we want to generate the binding name in this way?

Couldn't we just add another interactive question asking the user for what binding name they'd like to use? 🤔 

(then check the config file to make sure there isn't a conflict)

Basically what I was thinking was something along the lines of for example:
```
Would you like to update the wrangler.jsonc file with the new KV Namespace binding? … yes
What binding name would you like to use? MY_KV
Updated wrangler.jsonc with new KV Namespace binding: MY_KV (my-kv, ID: 6d2dd6ae297441688c8376c43e4360d6)
```

Potentially with an error/or other questions in case there are conflicts

I don't think that just creating the binding with generic names is very polished 🤔

thomasgauvin: using generic names is what we often have in docs/dashboard, and it's faster, so i wanted to match that. how about I default to generic and allow writing a custom binding?

packages/wrangler/src/kv/index.ts
```diff
@@ -80,6 +80,11 @@ export const kvNamespaceCreateCommand = createCommand({
 			type: "boolean",
 			describe: "Interact with a preview namespace",
 		},
+		"update-config": {
```
dario-piotrowicz: Again I don't think this feels very polished and I would suggest to have something different, for example a flag `config-binding-name` that once set to a string it adds a binding with such name to the wrangler config (+ naming conflict to either error or create a new unique name)

For example:
```
npx wrangler kv namespace create my-kvvvv3 --config-binding-name MY_KV

Creating namespace with title "my-kv"
Success!
Updated wrangler.jsonc with new KV Namespace binding: MY_KV (my-kv, ID: 3e22ff5cbc63489984343bca4b825fe4)
```

dario-piotrowicz (commented):
Thanks for the PR 🙂 

Code wise I think some things need to be polished also more testing (fixtures) will be needed anyways for now I just left some comments regarding the DX, I don't think it's very polished to just create the bindings names for the user and that we should instead ask the user what names they do want to use instead

packages/wrangler/src/config/auto-update.ts
dario-piotrowicz: > using generic names is what we often have in docs/dashboard

Yeah, but there (especially in the docs) aren't really dealing with the user specific configs so it does make sense to use a generic naming 🤔 

> how about I default to generic and allow writing a custom binding?

Sorry what do you mean by allowing writing a custom binding?

dario-piotrowicz: PS: to be clear I don't think that using a generic binding name is terrible, I just think that not doing so would provide a better DX for our users

thomasgauvin:
Thanks @dario-piotrowicz for the feedback & insights, I've started to address the DX ideas (interactive + config-binding-name) so am converting this to a draft PR for now

packages/wrangler/src/config/auto-update.ts
```diff
@@ -0,0 +1,132 @@
...
+			resource.binding = await promptForValidBindingName(config, resourceType);
+		}
+	}
+
+	// We now have a valid binding name. We can use it to update the wrangler.jsonc config.
+	if (resource.binding) {
+		updateJsonConfig(config.configPath!, resource, config);
+	} else {
+		// Show snippet only
+		const envString = args.env ? ` under [env.${args.env}]` : "";
```
vicb, line 86: This looks unused?

packages/wrangler/src/config/auto-update.ts
```diff
@@ -0,0 +1,132 @@
...
+		logger.log(`Add the following to your configuration file:`);
+
+		displayConfigSnippet(resource, config.configPath);
+	}
+}
+
+export function updateJsonConfig(
```
vicb, line 93: Could you please add JSDoc to this function?

packages/wrangler/src/config/auto-update-helpers.ts
```diff
@@ -0,0 +1,243 @@
...
+	for (const resourceType of Object.keys(RESOURCE_CONFIG_REGISTRY)) {
+		const bindings = config[
+			resourceType as keyof typeof RESOURCE_CONFIG_REGISTRY
+		] as Array<Record<string, unknown>> | undefined;
+		if (bindings) {
+			for (const binding of bindings) {
+				if (binding.binding && typeof binding.binding === "string") {
+					if (binding.binding.toUpperCase() === normalizedName) {
+						return true;
+					}
```
vicb, line 102: Collapse the 2 conditions?

```ts
				if (binding.binding && typeof binding.binding === "string" && binding.binding.toUpperCase() === normalizedName) {
						return true;
					}
```					

I don't think `binding.binding` is needed with the typeof?

```ts
				if (typeof binding.binding === "string" && binding.binding.toUpperCase() === normalizedName) {
						return true;
					}
```

packages/wrangler/src/config/auto-update-helpers.ts
```diff
@@ -0,0 +1,243 @@
...
+				bindingName = currentDefault;
+			} else {
+				// If no default and user pressed enter, continue the loop
+				continue;
+			}
+		}
+
+		const validation = validateBindingName(config, bindingName);
+		if (validation.valid) {
+			isValid = true;
```
vicb, line 234: Do you need `isValid` here or can you `while(true)` and `return bindingName;` here?

packages/wrangler/src/config/auto-update.ts
```diff
@@ -0,0 +1,132 @@
...
+ * Unified function to handle binding name validation, prompting, and config updates
+ * for all resource creation commands. This eliminates code duplication across
+ * KV, D1, R2, Vectorize, and Hyperdrive commands.
+ *
+ * This function handles early binding name validation when provided,
+ * prompts for binding names when needed, and manages config updates.
+ */
+export async function handleResourceBindingAndConfigUpdate(
+	args: { configBindingName?: string; env?: string },
+	config: RawConfig & { configPath?: string },
```
vicb, line 40: Looking at the calls outside of tests, I see 

```ts
{ ...config, configPath: config.configPath },
```
Would it make sense for `configPath` to be a distinct parameter?

packages/wrangler/src/config/auto-update.ts
```diff
@@ -0,0 +1,132 @@
...
+
+		// If the user does want to update wrangler.jsonc, get a valid binding name
+		if (shouldUpdate) {
+			resource.binding = await promptForValidBindingName(config, resourceType);
+		}
+	}
+
+	// We now have a valid binding name. We can use it to update the wrangler.jsonc config.
+	if (resource.binding) {
+		updateJsonConfig(config.configPath!, resource, config);
```
vicb, line 83: `config.configPath!` -> should the value be required instead of optional ?

packages/wrangler/src/config/auto-update.ts
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+export function updateJsonConfig(
+	configPath: string,
+	resource: ResourceBinding,
+	config: RawConfig
+) {
+	// Read the original file content to preserve formatting and comments
+	const originalContent = readFileSync(configPath);
```
vicb, line 99: nit: use async version? (also for write)

packages/wrangler/src/d1/create.ts
```diff
@@ -82,15 +88,16 @@ export const d1CreateCommand = createCommand({
...
-			formatConfigSnippet(
-				{
-					d1_databases: [
-						{ binding: "DB", database_name: db.name, database_id: db.uuid },
-					],
-				},
-				config.configPath
-			)
+
+		// Handle binding name and config update using unified utility
```
vicb, line 92: I love to see code comments and I must say you have done a great job in this PR!
But (nit) this one does not look to add value over the fn name?

```suggestion
```

penalosa:
Superceded by https://github.com/cloudflare/workers-sdk/pull/10414

PR #9819, opened 2025-07-02
feat(vite-plugin): Add containers support in vite-plugin

### LINKED ISSUE #9793: Containers don't work with @cloudflare/vite-plugin

Wrangler 4.22.0, Node 22.16.0, MacOS 15.5 

https://github.com/craigsdennis/sandboxing-day-containers

### Describe the Bug

Containers don't work in local development with the @cloudflare/vite-plugin.  When a container endpoint is attempted to be fetched one gets:

> Error: Container is not enabled for this durable object class. Have you correctly setup your wrangler.toml?

I filed a [separate ticket](https://github.com/cloudflare/containers/issues/34) in the containers project, but I imagine the root cause is in the vite-plugin.

The latest compatibility date supported by the installed Cloudflare Workers Runtime is "2025-06-17",
but you've requested "2025-06-22". Falling back to "2025-06-17"...

  VITE v6.3.5  ready in 524 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  Debug:   http://localhost:5173/__debug
  ➜  press h + enter to show help
Request received

Getting container demo
Container started

Error: Container is not enabled for this durable object class. Have you correctly setup your wrangler.toml?
    at Array.<anonymous> (/Users/vanpelt/Development/sandboxing-day-containers/worker/index.ts:39:18) {
  remote: true
}

CarmenPopoviciu:
Fixes #9793 & DEVX-1944
---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -467,6 +516,13 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					})
 				);
 			},
+			async closeBundle() {
```
CarmenPopoviciu: I would really appreciate some sanity check here. According to [vite docs](https://vite.dev/guide/api-plugin.html#universal-hooks) this is the right hook where we'd want to cleanup process resources, 

Screenshot 2025-07-08 at 18 25 45

but I'm new to vite so I might be overlooking smth

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -905,4 +961,33 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 			? resolvedPluginConfig.workers[environmentName]
 			: undefined;
 	}
+
+	/**
+	 * @returns Container options suitable for building or pulling images,
+	 * with image tag set to well-known dev format, or undefined if
+	 * containers are not enabled or not configured.
+	 */
+	async function getContainerOptions(
```
CarmenPopoviciu, line 1045: this aaalmost duplicates the wrangler counterpart, with the exception that `containerBuildId` needs to be passed into this function, as the vite-plugin does not have access to it via the config, as wrangler does

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -392,6 +418,27 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
 					}
+
+					if (
+						entryWorkerConfig.containers?.length &&
+						entryWorkerConfig.dev.enable_containers
+					) {
+						// Assemble container options and build if necessary
+						const containerOptions = await getContainerOptions(
+							entryWorkerConfig,
+							containerBuildId as string
```
emily-shen: maybe instead of casting it to a string, assert this is defined?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -392,6 +418,27 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+						);
+						const dockerPath = unstable_getDockerPath();
```
emily-shen: instead of exporting this function in wrangler, i think we can just get it off process.env. According to pete all the env var factories exist because we wanted to consolidate and warn on certain env vars. but we don't need that for this env var, so we can just get it off process.env to reduce complexity.

emily-shen (commented):
lgtm on an initial pass 🧡 

@petebacondarwin can you have a look at this from the vite side?

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: fantastic!

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -905,4 +964,40 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+		}
+		return containers;
+	}
+}
+
+function getDockerPath(): string {
+	const defaultDockerPath = "docker";
+	const dockerPathEnvVar = "WRANGLER_DOCKER_BIN";
+
+	return process.env[dockerPathEnvVar] || defaultDockerPath;
```
CarmenPopoviciu: @emily-shen is `||` appropriate here? My thinking was that if the env var is set to an empty string that's prolly not intended and we should default to `"docker"` anyway. But ofc one could argue that we're, in that case, ignoring user set env vars 🤔 . I can go either way here

emily-shen: falling back on the default if there's an empty string makes sense to me yeah!

vanpelt:
Impressive @CarmenPopoviciu, thanks for putting this together 😍.  I'm really excited for this to Just Work™

CarmenPopoviciu:
> Impressive @CarmenPopoviciu, thanks for putting this together 😍. I'm really excited for this to Just Work™

we're aaaalmost there ⏳  ...

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: went with `buildEnd` eventually, since that's more appropriate for our use case. Left a comment in the code explaining all the things

Hebilicious:
Thanks for the work @CarmenPopoviciu , would this also fix `@cloudflare/vitest-pool-workers` compatibility with containers?

.changeset/wet-rockets-stare.md
```diff
@@ -0,0 +1,9 @@
+---
+"@cloudflare/vite-plugin": minor
+"@cloudflare/containers-shared": minor
+"wrangler": minor
+---
+
+feat(vite-plugin): Add containers support in `vite dev`
+
+Adds support for Cloudflare Containers in `vite dev`. Please note that at the time of this PR a container image can only specify the path to a `Dockerfile`. Support for registry links will be added in a later version.
```
emily-shen: Also note that vite preview will be in a follow up too?

.changeset/wet-rockets-stare.md
```diff
@@ -0,0 +1,9 @@
+---
+"@cloudflare/vite-plugin": minor
+"@cloudflare/containers-shared": minor
+"wrangler": minor
```
emily-shen:
```suggestion
"@cloudflare/vite-plugin": patch
"@cloudflare/containers-shared": patch
"wrangler": patch
```
still in beta

packages/containers-shared/src/client/core/request.ts
```diff
@@ -263,6 +263,10 @@ export const sendRequest = async (
 	};
 
 	if (config.WITH_CREDENTIALS) {
+		// :(
```
emily-shen, line 266: can you also note in this comment that this path is only hit by the vite plugin when pulling images? (which is not in this PR)

emily-shen (approved):

CarmenPopoviciu:
> Thanks for the work @CarmenPopoviciu , would this also fix `@cloudflare/vitest-pool-workers` compatibility with containers?

as @emily-shen mentioned in https://github.com/cloudflare/workers-sdk/issues/9793#issuecomment-3088714902 unfortunately not. This strictly adds support in `vite dev`. A follow-up PR will add support in `vite-preview`. But this work is all strictly related to the vite plugin

.changeset/wet-rockets-stare.md
CarmenPopoviciu: oh right. my apologies

.changeset/wet-rockets-stare.md
CarmenPopoviciu: sure, I'll add that

jamesopstad: Should this be a minor for the Vite plugin as it's a new feature?

packages/containers-shared/src/client/core/request.ts
```diff
@@ -263,6 +263,10 @@ export const sendRequest = async (
...
+		// The vite-plugin is attempting to typecheck everything with worker types, which does not support request.credentials
+		// Also note this is always set to "omit".
+		// @ts-ignore
```
jamesopstad, line 269: I added the `workers-types` to the plugin types to fix type errors when importing from `workers-shared`. It would be good to find a different solution though as it pollutes the types and leads to other issues. @petebacondarwin, you mentioned [here](https://github.com/cloudflare/workers-sdk/pull/9575#discussion_r2149390940) that you would have a think about how to solve this. Any ideas?

packages/vite-plugin-cloudflare/playground/containers/__tests__/containers.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { expect, test } from "vitest";
+import { getTextResponse } from "../../__test-utils__";
+
+// skip until containers support is implemented in `vite preview`
+test.skip("starts container", async () => {
```
jamesopstad:
```suggestion
// skip build test until containers support is implemented in `vite preview`
test.skipIf(isBuild)("starts container", async () => {
```

We should still run the test for dev (thought this will require having Docker running to run tests locally).

packages/vite-plugin-cloudflare/playground/containers/__tests__/containers.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { expect, test } from "vitest";
+import { getTextResponse } from "../../__test-utils__";
```
jamesopstad:
```suggestion
import { getTextResponse, isBuild } from "../../__test-utils__";
```

packages/vite-plugin-cloudflare/playground/containers/__tests__/containers.spec.ts
```diff
@@ -0,0 +1,11 @@
...
+
+// skip until containers support is implemented in `vite preview`
+test.skip("starts container", async () => {
+	let response = await getTextResponse("/start");
+	expect(response).toBe("Container create request sent...");
+
+	response = await getTextResponse("/status");
+	expect(response).toBe("true");
```
jamesopstad:
```suggestion
	const startResponse = await getTextResponse("/start");
	expect(startResponse).toBe("Container create request sent...");

	const statusResponse = await getTextResponse("/status");
	expect(statusResponse).toBe("true");
```

packages/vite-plugin-cloudflare/playground/containers/package.json
```diff
@@ -0,0 +1,20 @@
...
+		"build": "vite build",
+		"check:types": "tsc --build",
+		"dev": "vite dev",
+		"preview": "vite preview"
+	},
+	"devDependencies": {
+		"@cloudflare/vite-plugin": "workspace:*",
+		"@cloudflare/workers-tsconfig": "workspace:*",
+		"@cloudflare/workers-types": "^4.20250617.0",
+		"ts-dedent": "^2.2.0",
```
jamesopstad:
```suggestion
```

This doesn't appear to be used.

packages/vite-plugin-cloudflare/playground/containers/container/simple-node-app.js
```diff
@@ -0,0 +1,49 @@
+const { createServer } = require("http");
```
jamesopstad, line 1: Should this file be kept in sync with `/fixtures/container-app/container/simple-node-app.js`? If so, it would be good to add a comment to both files.

packages/vite-plugin-cloudflare/playground/containers/src/index.ts
```diff
@@ -0,0 +1,70 @@
+import { DurableObject } from "cloudflare:workers";
```
jamesopstad, line 1: Same question about keeping this file in sync with `fixtures/container-app/src/index.ts`

packages/vite-plugin-cloudflare/playground/containers/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export class Container extends DurableObject<Env> {
+	container: globalThis.Container;
```
jamesopstad, line 4: I see this is the same as in the fixture but out of curiosity, why does the type come from `globalThis`?

packages/vite-plugin-cloudflare/playground/containers/wrangler.jsonc
```diff
@@ -0,0 +1,27 @@
+{
+	"name": "container-app",
+	"main": "src/index.ts",
```
jamesopstad:
```suggestion
	"main": "./src/index.ts",
```

Just for consistency with the other playgrounds.

packages/vite-plugin-cloudflare/playground/containers/wrangler.registry.jsonc
jamesopstad, line 1: Is this file included for use in the future?

packages/vite-plugin-cloudflare/src/containers.ts
```diff
@@ -0,0 +1,47 @@
+import { runDockerCmd } from "@cloudflare/containers-shared/src/utils";
+
+/**
+ * Returns the path to the Docker executable as defined by the
+ * `WRANGLER_DOCKER_BIN` environment variable, or the default value
```
jamesopstad, line 5: Not to be resolved here but I'm not sure we should be prefixing these environment variables with `WRANGLER_` given that we now support multiple tools. I'll add this to the team meeting agenda.

packages/vite-plugin-cloudflare/src/containers.ts
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Performs the forced removal of running Docker containers, given their ids.
+ *
+ * Please note that this function is almost identical to `cleanupContainers`
+ * defined in `containers-shared`, with the exception that the current fn
+ * expects the list of ids of all containers that are to be removed, as a
+ * parameter. This is because extracting these ids is an async operation, and
+ * we are currently restricted from performing async cleanup work upon
+ * closing/exiting the vite dev process (see vite-plugin-cloudflare/src/index.ts
+ * for more details).
```
jamesopstad, line 24: Can we add a corresponding comment in `containers-shared`.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -1,6 +1,13 @@
 import assert from "node:assert";
 import * as fsp from "node:fs/promises";
 import * as path from "node:path";
+import { prepareContainerImagesForDev } from "@cloudflare/containers-shared/src/images";
+import { getDevContainerImageName } from "@cloudflare/containers-shared/src/knobs";
+import { type ContainerDevOptions } from "@cloudflare/containers-shared/src/types";
```
```suggestion
import type { ContainerDevOptions } from "@cloudflare/containers-shared/src/types";
```

We always import types separately (I thought this was enforced by prettier but it looks like it's not).

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -89,12 +98,16 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 
 	const additionalModulePaths = new Set<string>();
 	const nodeJsCompatWarningsMap = new Map<WorkerConfig, NodeJsCompatWarnings>();
+	const containerImageTagsSeen = new Set<string>();
+	let runningContainersIds: Array<string>;
```
```suggestion
	let runningContainerIds: Array<string>;
```

(name will need changing elsewhere)

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -323,11 +340,21 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					miniflare
 				);
 
-				const miniflareDevOptions = await getDevMiniflareOptions(
+				let containerBuildId: string | undefined;
+				const workerConfig = getEntryWorkerConfig(resolvedPluginConfig);
```
```suggestion
				const entryWorkerConfig = getEntryWorkerConfig(resolvedPluginConfig);
```

Are we only allowing containers in the entry Worker?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -323,11 +340,21 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+				if (
+					workerConfig?.containers?.length &&
+					workerConfig.dev.enable_containers
+				) {
```
```suggestion
        const hasDevContainers = entryWorkerConfig?.containers?.length && entryWorkerConfig.dev.enable_containers
				if (hasDevContainers) {
```

Let's make this a variable as it's also used later on.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -392,6 +419,76 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 							}
 						};
 					}
+
+					if (
+						entryWorkerConfig.containers?.length &&
+						entryWorkerConfig.dev.enable_containers
+					) {
```
```suggestion
					if (hasDevContainers) {
```

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Also, `entryWorkerConfig` is already defined on 370 so that could now be removed.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -392,6 +419,76 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+
+						if (containerOptions) {
+							await prepareContainerImagesForDev({
+								dockerPath,
+								configPath: entryWorkerConfig.configPath,
+								containerOptions,
+								onContainerImagePreparationStart: () => {},
+								onContainerImagePreparationEnd: () => {},
+							});
+						}
```
```suggestion
						if (containerOptions) {
						  // keep track of them so we can clean up later
						  for (const container of containerOptions) {
							  containerImageTagsSeen.add(container.imageTag);
						  }
						  
							await prepareContainerImagesForDev({
								dockerPath,
								configPath: entryWorkerConfig.configPath,
								containerOptions,
								onContainerImagePreparationStart: () => {},
								onContainerImagePreparationEnd: () => {},
							});
						}
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -392,6 +419,76 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+						 * not to be emitted).
+						 *
+						 * Furthermore, since the `exit` event handler cannot perform async
+						 * ops as per spec (https://nodejs.org/api/process.html#event-exit),
+						 * we also need a mechanism to ensure that list of containers to be
+						 * cleaned up on exit is up to date. This is what the interval with id
+						 * `dockerPollIntervalId` is for.
+						 *
+						 * It is possible, though very unlikely, that in some rare cases,
+						 * we might be left with some orphaned conatiners, due to the fact
```
```suggestion
						 * we might be left with some orphaned containers, due to the fact
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -392,6 +419,76 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+						 * that at the point of exiting the dev process, our internal list
+						 * of container ids is out of date. We accept this caveat for now.
+						 *
+						 */
+						process.on("exit", () => {
+							clearInterval(dockerPollIntervalId);
+							removeContainersByIds(dockerPath, runningContainersIds);
+						});
+					}
```
jamesopstad: Are we still hoping to add an API to Vite to better support this?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -467,6 +566,17 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 					})
 				);
 			},
+			async buildEnd() {
```
jamesopstad, line 588: This hook will also run at build time. I think that should be OK because everything inside will be a noop but wanted to check?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -905,4 +1015,33 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+				imageTag: getDevContainerImageName(
+					container.class_name,
+					containerBuildId
+				),
+				args: container.image_vars,
+				imageBuildContext: container.image_build_context,
+				class_name: container.class_name,
+			});
+		}
+		return containers;
```
jamesopstad: `map` seems more natural here:

```suggestion
		return config.containers.map((container) => {
			return {
				image: container.image ?? container.configuration?.image,
				imageTag: getDevContainerImageName(
					container.class_name,
					containerBuildId
				),
				args: container.image_vars,
				imageBuildContext: container.image_build_context,
				class_name: container.class_name,
			};
		});
```

.changeset/wet-rockets-stare.md
CarmenPopoviciu: even if containers is still in beta?

packages/vite-plugin-cloudflare/playground/containers/__tests__/containers.spec.ts
CarmenPopoviciu: oh, I didn't realise the tests run for both dev and preview. TY for highlighting that 🙏

packages/vite-plugin-cloudflare/playground/containers/wrangler.registry.jsonc
CarmenPopoviciu: that is correct. A container image can be declared as the path to a Dockerfile or a registry link. Unfortunately registry links don't work out of the box with just the changes in this PR. That will require some more work on my part to make it all work. I'm happy to remove the file though and re-add when that work lands. In fact, let me just do that so that this PR is more clean

packages/vite-plugin-cloudflare/src/containers.ts
CarmenPopoviciu: that is a very fair point and a great callout. Is there a naming convention for env vars in vite-plugin? I'll have a look in wrangler and see if all our internal env vars are prefixed with `WRANGLER`. This particular one might have been just an oversight 🤔 . Otherwise...I think renaming to `DOCKER_BIN` should be ok? I'd have to check though that we don't overlap with other tools

packages/vite-plugin-cloudflare/src/containers.ts
CarmenPopoviciu: will do!

packages/containers-shared/src/client/core/request.ts
emily-shen: as discussed with @jamesopstad elsewhere, but we should probably move the config utils out of `workers-shared` into its own thing, and then any config util imports (in the vite plugin or elsewhere) will stop dragging the asset/router worker code along with it and messing up typechecking. 

since this PR doesn't actually add anything that actually hits this part of containers-shared, let's leave this for a followup?

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: yes! very much hoping for that! I'll bring it up with the team on Discord

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: done

.changeset/wet-rockets-stare.md
jamesopstad: Not sure. Maybe leave then and it can be fixed up in VP if necessary.

packages/vite-plugin-cloudflare/playground/containers/container/simple-node-app.js
CarmenPopoviciu: not necessarily. Technically the containers playground and fixture "apps" could have been two entirely diff apps, with diff setups. I just copied the fixture app over to make things easier for myself. Personally I wouldn't worry abt keeping them in sync. As long as they both test for the right thing, I think we're good

packages/vite-plugin-cloudflare/playground/containers/src/index.ts
CarmenPopoviciu: same reply https://github.com/cloudflare/workers-sdk/pull/9819#discussion_r2218766976

packages/vite-plugin-cloudflare/playground/containers/src/index.ts
CarmenPopoviciu: I have no idea tbh. The fixture code was copied over in its entirety, and I don't have all the context around that code

packages/vite-plugin-cloudflare/playground/containers/wrangler.registry.jsonc
CarmenPopoviciu: removed ✅

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: added a code comment highlighting this. resolving....

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: I was a little bit confused here, as I originally wanted to write the code just as you are suggesting now, but then realised we have two diff APIs (`getWorkerConfig` and `getEntryWorkerConfig`) that seemingly seem to do the same, with very slight diff nuances. The setting of `entryWorkerConfig` on line `370` is conditioned by `if (resolvedPluginConfig.type === "workers")`, and I assumed there's a reason for that. I also assumed there's a reason why, outside of that conditional, `getDevMiniflareOptions` calls `getEntryWorkerConfig` and not `getWorkerConfig`, as we do on line `370`. But if you are saying that it's alright to call `getEntryWorkerConfig` outside of that conditional, I'd be very happy to apply your suggestion. It makes the code cleaner.

As a side node it is rather confusing to have both `getWorkerConfig` and `getEntryWorkerConfig`. We should either unify them, or add some code comments as to what the diff between them are. Happy to do that myself, if I get a better understanding of these APIs <3

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: we basically had two options here: `buildEnd` or `closeBundle`. Pete suggested `buildEnd` is more appropriate for our use case, so here we are ¯\_(ツ)_/¯

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Yep, good point. The main differences are that with `getWorkerConfig` you pass in the environment name and it gets the `resolvedPluginConfig` via closure. It is used a lot in the other hooks where you can pass in the environment name using `this.environment.name`. Previously `getEntryWorkerConfig` was only used in `miniflare-options.ts` where you need to pass in `resolvedPluginConfig`.

I would use `getEntryWorkerConfig` further up to get `entryWorkerConfig` and remove the variable at 370. This is fine as long as the assertion at 373 stays where it is.

If containers aren't supported in auxiliary Workers then we'll need to document this.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Yep, `buildEnd` is more appropriate. I think it currently works fine as is but let's guard the logic to only apply in dev. It will be safer for future changes that way. I'll add a suggestion.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -467,6 +562,17 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+				const dockerPath = getDockerPath();
+				runningContainerIds = await getContainerIdsByImageTags(
+					dockerPath,
+					containerImageTagsSeen
+				);
+
+				await removeContainersByIds(dockerPath, runningContainerIds);
+				containerImageTagsSeen.clear();
+				runningContainerIds = [];
+			},
```
```suggestion
				if (resolvedViteConfig.command === "serve") {
					const dockerPath = getDockerPath();
					runningContainersIds = await getContainerIdsByImageTags(
						dockerPath,
						containerImageTagsSeen
					);

					await removeContainersByIds(dockerPath, runningContainersIds);
					containerImageTagsSeen.clear();
					runningContainersIds = [];
				}
```

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: what about `vite preview`. We'll want a similar cleanup mechanism there as well no?

jamesopstad: Yep, we can also check for preview in that PR but we still want to make sure the logic doesn't run during the build.

CarmenPopoviciu: gotcha! I'll add that check when I implement the `preview` support, and restrict it just to `dev` for now

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: done!

brettimus:
ooooh came to check in on this PR just as CI kicked off, now i'm intently watching github progress checks  👀  🍿

CarmenPopoviciu:
> ooooh came to check in on this PR just as CI kicked off, now i'm intently watching github progress checks 👀 🍿

whoohooo! it's 🍏 🎉 ✨

PR #9820, opened 2025-07-02
containers local dev e2e tests 

emily-shen:
These tests will also only run in linux in CI, because github runners don't support any container tools (without janky workarounds).

Tests for rebuild hotkey and removing containers on dev session will happen in a followup PR but are currently blocked on a path resolution bug, and i'd like to get _some_ tests in first.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: tests
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: feature doesn't exist in v3

### DIFF
--- a/packages/wrangler/e2e/containers.dev.test.ts
+++ b/packages/wrangler/e2e/containers.dev.test.ts
@@ -8,6 +8,7 @@ import {
 	it,
 	vi,
 } from "vitest";
+import { getDockerPath } from "../src/environment-variables/misc-variables";
 import { dedent } from "../src/utils/dedent";
 import { CLOUDFLARE_ACCOUNT_ID } from "./helpers/account-id";
 import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
@@ -38,14 +39,14 @@ describe
 				containers: [
 					{
 						image: "./Dockerfile",
-						class_name: "Container",
+						class_name: `E2EContainer`,
 						name: `${workerName}-container`,
 					},
 				],
 				durable_objects: {
 					bindings: [
 						{
-							class_name: "Container",
+							class_name: `E2EContainer`,
 							name: "CONTAINER",
 						},
 					],
@@ -53,7 +54,7 @@ describe
 				migrations: [
 					{
 						tag: "v1",
-						new_classes: ["Container"],
+						new_classes: [`E2EContainer`],
 					},
 				],
 			};
@@ -62,7 +63,7 @@ describe
 				"src/index.ts": dedent`
 					import { DurableObject } from "cloudflare:workers";
 
-					export class Container extends DurableObject<Env> {
+					export class E2EContainer extends DurableObject<Env> {
 						container: globalThis.Container;
 
 						constructor(ctx: DurableObjectState, env: Env) {
@@ -115,7 +116,7 @@ describe
 
 				const server = createServer(function (req, res) {
 					res.writeHead(200, { "Content-Type": "text/plain" });
-					res.write("Hello World!" + " Have an env var! " + process.env.MESSAGE);
+					res.write("Hello World! Have an env var! " + process.env.MESSAGE);
 					res.end();
 				});
 
@@ -136,7 +137,7 @@ describe
 					containers: [
 						{
 							image: `registry.cloudflare.com/${CLOUDFLARE_ACCOUNT_ID}/${workerName}:tmp-e2e`,
-							class_name: "Container",
+							class_name: `E2EContainer`,
 							name: `${workerName}-container`,
 						},
 					],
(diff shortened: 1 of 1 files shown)


packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -1,80 +1,222 @@
...
-	],
-};
+const imageSource = ["pull", "build"];
 
-// TODO: docker is not installed by default on macOS runners in github actions.
-// And windows is being difficult.
-// So we skip these tests in CI, and test this locally for now :/
-describe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
-	"containers local dev tests",
+// We can only really run these tests on Linux, because we build our images for linux/amd64,
```
vicb, line 19: Do changing the os in e2e.yml has side effects?

Could another be to pass the platform to the bulid command? (I think there is an option, not sure if settable here)

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -1,80 +1,222 @@
...
+
+					COPY ./container/app.js app.js
 					EXPOSE 8080
 					`,
+				"container/app.js": dedent`
+				const { createServer } = require("http");
+
+				const server = createServer(function (req, res) {
+					res.writeHead(200, { "Content-Type": "text/plain" });
+					res.write("Hello World!" + " Have an env var! " + process.env.MESSAGE);
```
vicb: nit/question
Should `Have an env var!` be displayed only when true?
If not the first two strings can be merged

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -1,80 +1,222 @@
...
+				server.listen(8080, function () {
+					console.log("Server listening on port 8080");
+				});
+				`,
 			});
+			// if we are pulling we need to push the image first
+			if (source === "pull") {
+				// pull a container image from the registry
+				await helper.run(
+					`wrangler containers build . -t ${workerName}:tmp-e2e -p`
```
vicb, line 132: I think `--platform` is accepted here

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -1,80 +1,222 @@
...
+				});
+				await new Promise((resolve) => setTimeout(resolve, 5000));
+			}
+		}, 30000);
+		beforeEach(async () => {
+			await helper.seed({
+				"wrangler.json": JSON.stringify(wranglerConfig),
+			});
+			// cleanup any running containers
+			const ids = getContainerIds();
```
vicb: Would this be cleaning all my local containers when I run the tests?

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -1,80 +1,222 @@
...
+			if (ids.length > 0) {
+				execSync("docker rm -f " + ids.join(" "), {
```
vicb: Is `docker` ok when running locally or should we provide a way to override the location (not sure what we do with the wrangler commands)

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -1,80 +1,222 @@
...
 		});
 
+		it(`will be able to interact with the container`, async () => {
+			const worker = helper.runLongLived("wrangler dev");
+			const ready = await worker.waitForReady();
+			await worker.readUntil(/Container image\(s\) ready/);
+
+			let response = await fetch(`${ready.url}/status`);
+			let status = await response.json();
+			expect(response.status).toBe(200);
```
vicb, line 197: nit: check `response.status` before aceessing the response `reponse.json()`

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -1,80 +1,222 @@
...
+							image: `registry.cloudflare.com/${CLOUDFLARE_ACCOUNT_ID}/${workerName}:tmp-e2e`,
+							class_name: "Container",
+							name: `${workerName}-container`,
+						},
+					],
+				};
+				await helper.seed({
+					"wrangler.json": JSON.stringify(wranglerConfig),
+				});
+				await new Promise((resolve) => setTimeout(resolve, 5000));
```
vicb: nits: 
- add a comment on why we need to wait here
- use 5_000 and 30_000 here and below for readability and consistency with the rest of the file

vicb (approved):
LGTM.

2 main things:
- Does changing e2e os has side effects - IMO we should try to avoid if setting `--platform` work
- IIUC the PR would rm all the containers on my local machine when running the tests locally, not nice

A few minor comments.

packages/wrangler/e2e/containers.dev.test.ts
emily-shen: we always build for linux/amd64 at dev and deploy time, so i don't think we can/should change that. changing the CI os shouldn't make a difference, apparently it might be marginally slower but we don't really have any other options. we've used the amd runners in the past.

packages/wrangler/e2e/containers.dev.test.ts
emily-shen: ah yeah good point, will filter

packages/wrangler/e2e/containers.dev.test.ts
emily-shen: good point i'll pull in the WRANGLER_DOCKER_BIN env var

packages/wrangler/e2e/containers.dev.test.ts
emily-shen: it is, but tbh not sure why - we don't support anything other than linux/amd64. i'm guessing maybe a hold over from cloudchamber? i'll check with the containers team.

PR #9829, opened 2025-07-02
Use the right token when checking out external forks

penalosa:


dario-piotrowicz (commented):
Can you explain why the change? 🤔 

The workflow already has `contents: read` under `permissions`, is that not enough to checkout the repo? 🫤

petebacondarwin (approved):

penalosa:
> Can you explain why the change? 🤔
> 
> The workflow already has `contents: read` under `permissions`, is that not enough to checkout the repo? 🫤

It's enough to checkout the repo, yeah, but it's not enough to push to the repo. Checking out with the token means that the later push has permission

petebacondarwin:
@dario-piotrowicz - see https://github.com/cloudflare/workers-sdk/commit/28ecf853f815005f7c3f2f57557c57d167c14e53#diff-900be2a000a3c8e9022554fe17d14efc73ea31b1e7ee16052bcc166b9411ed42R19

dario-piotrowicz:
> > Can you explain why the change? 🤔
> > The workflow already has `contents: read` under `permissions`, is that not enough to checkout the repo? 🫤
> 
> It's enough to checkout the repo, yeah, but it's not enough to push to the repo. Checking out with the token means that the later push has permission

Sorry I totally missed your reply :sweat: 

What I was saying is that you were adding the token for the checkout step... that's why I am not getting why it is needed :confused:

PR #9833, opened 2025-07-02
ensure that container builds don't disrupt dev hotkey handling

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2012

_Describe your change..._

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/containers-shared/src/build.ts
```diff
@@ -43,32 +43,51 @@ export function dockerBuild(
...
+		child.stdin.write(options.dockerfile);
+		child.stdin.end();
+	}
+
+	child.on("exit", (code) => {
+		if (code === 0) {
+			resolve();
+		} else if (!errorHandled) {
+			errorHandled = true;
+			reject(new Error(`Build exited with code: ${code}`));
```
emily-shen, line 72: not related to this pr really but can we update the error message to 'docker build exited etc.' for clarity

fixtures/interactive-dev-tests/container-app/src/tsconfig.json
emily-shen, line 1: up to you, but what about deleting everything from this sub-fixture(?) except for the totally necessary bits so its a bit more minimal? ie deleting worker-configuration.d.ts, tsconfig, etc. that's what the other test cases in this fixture does.

fixtures/interactive-dev-tests/tests/index.test.ts
```diff
@@ -435,6 +436,125 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
...
 
+baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
+	"container dev where image build takes a long time",
+	{ retry: 0, timeout: 90000 },
+	() => {
+		let tmpDir: string;
+		beforeAll(async () => {
+			tmpDir = fs.mkdtempSync(path.join(tmpdir(), "wrangler-container-sleep-"));
+			fs.cpSync(
+				path.resolve(__dirname, "../", "container-sleep-noop-app"),
```
emily-shen: or what about just copying over the existing container-app folder in this fixture and overwriting the dockerfile? to prevent minimise duplication

packages/containers-shared/src/images.ts
```diff
@@ -48,16 +67,40 @@ export async function prepareContainerImagesForDev(
 	await verifyDockerInstalled(dockerPath);
 	for (const options of containerOptions) {
```
emily-shen, line 68: if we have multiple containers, will aborting one abort subsequent builds too, or do we need to add a check?

packages/containers-shared/src/utils.ts
```diff
@@ -5,31 +5,58 @@ import { dockerImageInspect } from "./inspect";
...
+export const runDockerCmd = (
 	dockerPath: string,
 	args: string[],
 	stdio?: StdioOptions
-) => {
+): {
+	abort: () => void;
+	ready: Promise<{ aborted: boolean }>;
+	// Note: we make the return type a thenable just for convenience so that callers can directly await it
+	then: (resolve: () => void, reject: () => void) => void;
```
emily-shen, line 16: this is cool  🤯

emily-shen (approved):
🧡

packages/containers-shared/src/build.ts
dario-piotrowicz: sorry could you clarify what error message you have in mind? 🤔

packages/containers-shared/src/images.ts
dario-piotrowicz: mh... that's a good point... I didn't actually consider multiple containers 😓 (mainly just because our fixture only has one 😅)

I think the changes in this PR won't work super well when multiple containers are in place (but not worse than the current wrangler behavior anyways)

If you'd be ok with it I would just ignore multiple containers in this PR and I'll create a new PR specifically for that 🙂

fixtures/interactive-dev-tests/tests/index.test.ts
dario-piotrowicz: good call! 😄 

done 🙂

fixtures/interactive-dev-tests/container-app/src/tsconfig.json
dario-piotrowicz: I've addressed this in the other way you proposed: https://github.com/cloudflare/workers-sdk/pull/9833#discussion_r2197642532 🙂

packages/containers-shared/src/build.ts
emily-shen: `Docker build exited with code: ${code}`

emily-shen: was a bit confusing when this error showed up in sentry

packages/containers-shared/src/images.ts
emily-shen: do you know what will happen when you've got multiple containers? will it just proceed on to the next one?
not an urgent one to fix, can you create a ticket to keep track?

packages/containers-shared/src/build.ts
dario-piotrowicz: Ah right, you meant just adding `Docker` in front of it? sorry missed that part 😅👍

packages/containers-shared/src/images.ts
dario-piotrowicz: I'm pretty sure only the last container built can be aborted, while the ones previously being built would keep running  and the user wouldn't be able to abort them

of course, I'll create a ticket 🙂👍

dario-piotrowicz: ah no sorry 🤦 

I didn't forgot about how the build is done (just below), we're building/pulling things sequentially, so this should just work regardless on whether you have one or multiple containers

dario-piotrowicz: yep, just manually checked that, it works perfectly fine 😅

PR #9843, opened 2025-07-03
fix(miniflare): improve log formatting with error cause

edmundhung:

_Describe your change..._

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/miniflare/src/shared/log.ts
+++ b/packages/miniflare/src/shared/log.ts
@@ -45,6 +45,24 @@ function dimInternalStackLine(line: string): string {
 	return line;
 }
 
+function formatError(error: unknown): string {
+	if (error instanceof Error && error.stack) {
+		let message = error.stack
+			.split("\n")
+			// Dim internal stack trace lines to highlight user code
+			.map(dimInternalStackLine)
+			.join("\n");
+
+		if (error.cause) {
+			message += `\nCaused by: ${formatError(error.cause)}`;
+		}
+
+		return message;
+	}
+
+	return error?.toString() ?? "";
+}
+
 export interface LogOptions {
 	prefix?: string;
 	suffix?: string;
@@ -94,15 +112,8 @@ export class Log {
 	error(message: Error): void {
 		if (this.level < LogLevel.ERROR) {
 			// Ignore message if it won't get logged
-		} else if (message.stack) {
-			// Dim internal stack trace lines to highlight user code
-			const lines = message.stack.split("\n").map(dimInternalStackLine);
-			this.logWithLevel(LogLevel.ERROR, lines.join("\n"));
 		} else {
-			this.logWithLevel(LogLevel.ERROR, message.toString());
-		}
-		if ((message as any).cause) {
-			this.error(prefixError("Cause", (message as any).cause));
+			this.logWithLevel(LogLevel.ERROR, formatError(message));
 		}
 	}
 


emily-shen:
what if we hid the middleware stack trace entirely? its not really relevant to the user.

packages/miniflare/src/plugins/core/errors/index.ts
```diff
@@ -278,19 +285,6 @@ export async function handlePrettyErrorRequest(
 	const youch = new Youch();
 
 	youch.useTransformer((error) => {
-		error.frames = error.frames
-			.filter(
-				(frame) =>
-					!frame.fileName?.includes(".wrangler/tmp") &&
-					!frame.fileName?.includes("wrangler/templates/middleware")
-			)
```
edmundhung, line 286: We are now filtering out internal stack frames when we revive the error.

packages/miniflare/src/plugins/core/errors/index.ts
```diff
@@ -278,19 +285,6 @@ export async function handlePrettyErrorRequest(
...
-			.map((frame) => {
-				// To avoid Youch throwing an error if the frame has no fileName
-				// This happens in tests which hides some parts of the stack trace
-				frame.fileName ??= "";
-
-				return frame;
-			});
```
edmundhung, line 293: This was added to workaround an issue in tests in which youch is too strict when parsing the location in the stack frame. It is no longer needed once I change the script path format from `script:${workerIndex}` to `script-${workerIndex}` in packages/miniflare/src/plugins/core/modules.ts

packages/miniflare/src/plugins/core/modules.ts
```diff
@@ -30,9 +30,9 @@ const builtinModulesWithPrefix = builtinModules.concat(
 
 // Module identifier used if script came from `script` option
 export function buildStringScriptPath(workerIndex: number) {
-	return `script:${workerIndex}`;
+	return `script-${workerIndex}`;
```
edmundhung, line 33: The colon here cause issue with how stack trace is parsed by both `youch` and `@cspotcode/source-map-support` as `:` is used to indicate column and row.

packages/miniflare/src/plugins/core/errors/index.ts
```diff
@@ -235,7 +235,14 @@ export function reviveError(
...
 	// Try to apply source-mapping to the stack trace
-	error.stack = getSourceMappedStack(workerSrcOpts, error);
+	error.stack = processStackTrace(
+		getSourceMappedStack(workerSrcOpts, error),
+		(line, location) =>
+			!location.includes(".wrangler/tmp") &&
+			!location.includes("wrangler/templates/middleware")
+				? line
+				: null
+	);
```
edmundhung, line 245: We are now filtering out internal stack frame when we revive the error, so that Youch and error logs are consistent.

edmundhung:
@emily-shen @petebacondarwin May I get a re-review? I have updated it to hide all internal stack frames now and make sure it is consistent on both error log and the error screen. I have included some screenshots of the final version.

penalosa (approved):
Looks good to me, but worth getting another review as well

PR #9844, opened 2025-07-03
update wrangler's `pnpm dev` to also watch the containers-shared src directory

dario-piotrowicz:
wrangler imports files from the containers-shared package (which does not have a build/dev command itself), so the wrangler dev command needs to also watch the containers-shared directory to make sure that changes to files in such directory trigger appropriate wrangler rebuilds

___

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: infra change
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: infra change
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: unrelated to wrangler v3


emily-shen (approved):
thanks for looking into this!

PR #9847, opened 2025-07-03
Bump undici & TS

penalosa:
Bump Undici & TS to the latest versions. I had a quick read of the Undici v6 & v7 breaking changes, and nothing seems like it'll affect us, but I'm mostly relying on E2E test coverage here, and the fact that we use the `fetch()` API which shouldn't have changed.

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: this has no user-facing changes in wrangler


packages/miniflare/src/http/fetch.ts
```diff
@@ -40,13 +33,13 @@ export async function fetch(
...
 		// Sec-WebSocket-Protocol header as ws treats this differently
-		const headers: Record<string, string> = {};
+		const headers = new undici.Headers();
 		let protocols: string[] | undefined;
 		for (const [key, value] of request.headers.entries()) {
 			if (key.toLowerCase() === "sec-websocket-protocol") {
 				protocols = value.split(",").map((protocol) => protocol.trim());
 			} else {
-				headers[key] = value;
+				headers.set(key, value);
```
edmundhung: Nit: Would it be better to make it `headers.append(key, value)`? As there could be multiple entries with the same key.

packages/miniflare/src/http/fetch.ts
```diff
@@ -59,13 +52,13 @@ export async function fetch(
 		// Establish web socket connection
 		const ws = new NodeWebSocket(url, protocols, {
 			followRedirects: request.redirect === "follow",
-			headers,
+			headers: Object.fromEntries(headers.entries()),
```
edmundhung, line 55: Nit: `Object.fromEntries` will ignores repeated keys, how about `new undici.Headers(headers)`?

packages/vitest-pool-workers/src/worker/fetch-mock.ts
```diff
@@ -21,17 +21,19 @@ interface BufferedRequest {
 	request: Request;
 	body: Uint8Array | null;
 }
-const requests = new WeakMap<Dispatcher.DispatchOptions, BufferedRequest>();
-const responses = new WeakMap<Dispatcher.DispatchOptions, Response>();
+const requests = new Map<string, BufferedRequest>();
+const responses = new Map<string, Response>();
```
penalosa: Previously, we relied on stability of the `DispatchOptions` variable, but that is no longer guaranteed in Undici v7 (see https://github.com/nodejs/undici/blob/c6a8b23ee55b34fedd7b90d3c7f4b51a35d3ee25/lib/mock/mock-agent.js#L74). As such, we serialise the options for lookup, and make sure we remove them manually from the `Map` after access so as not to cause a memory leak.

packages/vitest-pool-workers/src/worker/fetch-mock.ts
```diff
@@ -115,12 +118,7 @@ globalThis.fetch = async (input, init) => {
 	});
 
 	// Dispatch the request through the mock agent
-	const dispatchHandlers: Dispatcher.DispatchHandlers = {
-		onConnect(abort) {
-			if (abortSignalAborted) {
-				abort();
-			}
-		},
```
penalosa, line 123: This doesn't appear to work in Undici v7. The `abort()` parameter isn't documented so I'm not 100% sure, but in manual testing calling `abort()` here (rather than in `onComplete`) causes `fetch()` calls to abort with `undefined`

packages/miniflare/src/http/fetch.ts
penalosa: It will, but this is being passed in to `WebSocket` from `ws`, which only accepts an object here.

packages/miniflare/src/http/fetch.ts
```diff
@@ -40,13 +33,13 @@ export async function fetch(
 
 		// Normalise request headers to a format ws understands, extracting the
 		// Sec-WebSocket-Protocol header as ws treats this differently
-		const headers: Record<string, string> = {};
+		const headers = new undici.Headers();
```
vicb, line 36: Should the imported `UndiciHeaders` be used instead?

penalosa: Confusingly `UndiciHeaders` is type-only, and also different from `undici.Headers`

edmundhung (approved):

packages/miniflare/src/http/fetch.ts
```diff
@@ -99,9 +92,85 @@ export type DispatchFetch = (
...
-function addHeader(/* mut */ headers: AnyHeaders, key: string, value: string) {
-	if (Array.isArray(headers)) headers.push(key, value);
-	else headers[key] = value;
+function addHeader(
+	/* mut */ headers: undici.Headers,
+	key: string,
+	value: string
+) {
+	headers.set(key, value);
+}
```
petebacondarwin: Is there much point in this function now?

packages/miniflare/src/http/fetch.ts
```diff
@@ -99,9 +92,85 @@ export type DispatchFetch = (
...
+
+function isArray(headers: UndiciHeaders): headers is string[] {
+	return Array.isArray(headers);
+}
```
petebacondarwin: This is probably not really needed either.

packages/vitest-pool-workers/src/worker/fetch-mock.ts
```diff
@@ -153,7 +151,7 @@ globalThis.fetch = async (input, init) => {
 			responseChunks.push(chunk);
 			return true;
 		},
-		onComplete(_trailers) {
+		onComplete(_trailers: unknown) {
```
petebacondarwin:
```suggestion
		onComplete() {
```

??

packages/vitest-pool-workers/src/worker/fetch-mock.ts
```diff
@@ -184,7 +183,7 @@ globalThis.fetch = async (input, init) => {
 				responseResolve(maybeResponse);
 			}
 		},
-		onBodySent(_chunk) {}, // (ignored)
+		onBodySent(_chunk: unknown) {}, // (ignored)
```
petebacondarwin:
```suggestion
		onBodySent() {}, // (ignored)
```

???

packages/vitest-pool-workers/src/worker/fetch-mock.ts
petebacondarwin: In an ideal world we would wrap these in a class that made sure items were removed when we get them to avoid having to remember doing so at each callsite.

packages/miniflare/src/http/fetch.ts
```diff
@@ -99,9 +92,85 @@ export type DispatchFetch = (
...
+function isIterable(
+	headers: UndiciHeaders
+): headers is Iterable<[string, string | string[] | undefined]> {
+	return Symbol.iterator in Object(headers);
+}
+
+// See https://github.com/nodejs/undici/blob/main/docs/docs/api/Dispatcher.md?plain=1#L1151 for documentation
+function convertUndiciHeadersToStandard(
+	headers: NonNullable<UndiciHeaders>
+): undici.Headers {
```
petebacondarwin, line 105: Further down I see code that looks like:

```ts
new Headers(init.headers)
```

Why can't we use that here?

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -193,7 +192,10 @@ export async function fillOpenAPIConfiguration(
 		OpenAPI.BASE = base;
 	}
 
-	OpenAPI.HEADERS = headers;
+	OpenAPI.HEADERS = {
+		...(OpenAPI.HEADERS !== undefined ? OpenAPI.HEADERS : {}),
```
petebacondarwin:
```suggestion
		...(OpenAPI.HEADERS ?? {}),
```

??

petebacondarwin (approved):
Bunch of nits but LGTM

packages/miniflare/src/http/fetch.ts
penalosa: The allowed values for `UndiciHeaders` are wider than the `Header` constructor can take (in particular, `string[]`)

penalosa: The allowed values for UndiciHeaders are wider than the Header constructor can take (in particular, string[])

PR #9852, opened 2025-07-04
perf: experiment with an alternate binary search implementation

vicb:
Experiment with an alternate binary search implementation for asset.

The new implementation is ~10% faster in test.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: perf test
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/workers-shared/asset-worker/src/worker.ts
+++ b/packages/workers-shared/asset-worker/src/worker.ts
@@ -4,6 +4,7 @@ import { setupSentry } from "../../utils/sentry";
 import { mockJaegerBinding } from "../../utils/tracing";
 import { Analytics } from "./analytics";
 import { AssetsManifest } from "./assets-manifest";
+import { AssetsManifest as AssetsManifest2 } from "./assets-manifest.2";
 import { normalizeConfiguration } from "./configuration";
 import { ExperimentAnalytics } from "./experiment-analytics";
 import { canFetch, handleRequest } from "./handler";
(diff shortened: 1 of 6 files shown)


packages/workers-shared/asset-worker/src/worker.ts
```diff
@@ -231,6 +232,12 @@ export default class<TEnv extends Env = Env> extends WorkerEntrypoint<TEnv> {
 		pathname: string,
 		_request?: Request
 	): Promise<string | null> {
+		const BINARY_SEARCH_EXPERIMENT_SAMPLE_RATE = 0.5;
```
GregBrimble: I'd consider starting off with a lower threshold. 5%? Just make sure no errors are thrown — then we can jump up to more accurately compare performance.

vicb: > I'd consider starting off with a lower threshold. 5%? Just make sure no errors are thrown — then we can jump up to more accurately compare performance.

That is something I actually wanted to discuss with you and I think makes complete sense.

vicb:
@GregBrimble I have rebased to fix an issue with the CI (the main branch was failing when I branched off).

I have reduced the experiment to 5% of the request for now and added a fallback to the current impl.

Let me know if it looks good.

Thanks!

PR #9854, opened 2025-07-04
fix incorrect bindings remote deduplication logic

### LINKED ISSUE #9842: multi-worker wrangler dev incorrect remote bindings deduplication behavior

Wrangler@4.23

_No response_

### Describe the Bug

Let's say that I have 2 workers, `worker-a` and `worker-b` with the following wrangler config files:

```json
// wrangler-a.config.jsonc
{
  "name": "worker-a",
  "compatibility_date": "2025-04-03",
  "main": "./index-a.js",
  "services": [
    {
      "binding": "WORKER_B",
      "service": "worker-b",
      "experimental_remote": true
    }
  ],
  "kv_namespaces": [
    { "binding": "MY_KV", "id": "a4f1b5c514214a02b521ad60419b88db", "experimental_remote": true }
  ]
}
```
```json
// wrangler-b.config.jsonc
{
  "name": "worker-b",
  "compatibility_date": "2025-04-03",
  "main": "./index-b.js",
  "kv_namespaces": [
    { "binding": "MY_KV", "id": "a4f1b5c514214a02b521ad60419b88db" }
  ]
}
```

If I then run wrangler dev in the following way:
```sh
$ npx wrangler dev -c wrangler-a.config.jsonc -c wrangler-b.config.jsonc --x-remote-bindings
```

and try to interact with the KV binding in both workers I'll see that both bindings are remote, although the second one should actually be local

___

This is caused by some deduplication logic we have in Miniflare that causes only the first binding definition to be used

_No response_

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2062
Fixes #9842

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: bugfix to a non-wrangler-v3 feature


packages/miniflare/src/plugins/shared/index.ts
```diff
@@ -288,5 +288,60 @@ export async function migrateDatabase(
...
+ * @returns suffix to use in the service name
+ */
+function getRemoteServiceNameSuffix(
+	remoteProxyConnectionString: RemoteProxyConnectionString
+) {
+	const remoteSha = createHash("sha256")
+		.update(remoteProxyConnectionString.href)
+		.digest("hex");
+	const remoteShortSha = remoteSha.slice(0, 6);
+	return `remote-${remoteShortSha}`;
```
edmundhung, line 309: Is 6 characters safe enough to make sure it is unique? What will happen if there are two services end up with the same `remoteShortSha`?

What about extracting parts of the `remoteProxyConnectionString` instead? Assuming there would be some common bits in all connection strings.

dario-piotrowicz: > Is 6 characters safe enough to make sure it is unique?

I think it's very safe, only a handful of remote proxy connection strings can be present at a time and having collision on such a small amount seems quite impossible to me (I use 6 digit SHAs in git and I virtually never see conflicts even in huge repositories with thousand of commits).

> What will happen if there are two services end up with the same remoteShortSha?

The wrong remote resource will be used by one of the two services, but this will only happen if:
 - the resources are practically identical (for example an R2 binding with the exact same bucket name)
 - the remote resources below to different users, meaning that the remote proxies are started with different auth values (which is unlikely something that we'll provide out of the box anyways)

So based on the above I think that the risk of this being problematic is extremely extremely small, restarting the dev session would fix it, and most likely the impact would also be quite minimal.

dario-piotrowicz: > What about extracting parts of the remoteProxyConnectionString instead? Assuming there would be some common bits in all connection strings.

Yes, I could extract the port from the remoteProxyConnectionString and use that, that would guarantee that we do get different values here

I did consider that but that would require us to basically add coupling/rely on what the remoteProxyConnectionString is (a localhost URL). That's why I preferred going with a SHA approach where we don't care/need to know that the remoteProxyConnectionString contains, all we care about is that it is a string, and as I mentioned above I think that the risk of collisions is so small that it's practically non-existent.

But if you prefer I'm ok using the URL port instead 🙂

edmundhung: I don't have a strong preference here. We can change it only when users report any issue :)

packages/wrangler/e2e/wrangler-mixed-remote-resources.test.ts
```diff
@@ -0,0 +1,62 @@
+import path from "node:path";
```
penalosa: Shouldn't this test be in `wrangler-remote-resources.test.ts`?

packages/miniflare/src/plugins/shared/index.ts
```diff
@@ -288,5 +288,60 @@ export async function migrateDatabase(
...
+	) & { remoteProxyConnectionString?: RemoteProxyConnectionString }
+): string {
+	const identifier =
+		"binding" in bindingSpecifier
+			? bindingSpecifier.binding
+			: "id" in bindingSpecifier
+				? bindingSpecifier.id
+				: "name" in bindingSpecifier
+					? bindingSpecifier.name
+					: bindingSpecifier.namespace;
```
penalosa: nit, but I would find this clearer to read if it took a `string` id parameter directly, along with a separate (optional) `remoteProxyConnectionString`

penalosa (approved):
One readability comment, but otherwise ✅

packages/wrangler/e2e/wrangler-mixed-remote-resources.test.ts
dario-piotrowicz: Yes, I considered doing that, but all the tests in that file follow the same exact structure, so it felt more natural to me to have this test in a separate file instead, if you'd prefer it to be in `wrangler-remote-resources.test.ts` I'm totally ok to move it 🙂

packages/miniflare/src/plugins/shared/index.ts
dario-piotrowicz: this is definitely not super readable (but readable enough IMO), but it makes the code of callers very nice and very readable....
it is a tradeoff I think can I went with the one I found nicer, I don't have a huge preference so if you do prefer for callers to specify the id directly that's fine by me 🙂

packages/wrangler/e2e/wrangler-mixed-remote-resources.test.ts
dario-piotrowicz: moved 🙂

packages/miniflare/src/plugins/shared/index.ts
dario-piotrowicz: Updated 🙂

PR #9861, opened 2025-07-04
update eslint versions and ban direct console usage in Wrangler, Miniflare and C3

petebacondarwin:
Fixes DEVX-2035

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  eslint based refactoring


packages/eslint-config-worker/package.json
```diff
@@ -4,21 +4,19 @@
 	"private": true,
 	"description": "ESLint config for workers-sdk",
 	"main": "index.js",
-	"dependencies": {
-		"@typescript-eslint/parser": "^6.9.0",
+	"devDependencies": {
+		"@typescript-eslint/eslint-plugin": "catalog:default",
+		"@typescript-eslint/parser": "catalog:default",
 		"eslint": "^8.57.1",
```
alsuren, line 10: (optional): this is a private package, so it doesn't really matter, but if this was public then I think eslint and friends might need to be proper runtime dependencies of this package.

packages/wrangler/src/__tests__/pages/pages-deployment-tail.test.ts
```diff
@@ -849,14 +849,13 @@ function isRequest(event: TailEventMessageType): event is RequestEvent {
...
- * re-stringifies with some spacing, the same way wrangler tail does.
+ * of just JSON.parsing a raw string.
  *
  * @param message a buffer of data received from the websocket
- * @returns a string ready to be printed to the terminal or compared against
+ * @returns a JSON object ready to be compared against
  */
-function deserializeToJson(message: WebSocket.RawData): string {
-	return JSON.stringify(JSON.parse(message.toString()), null, 2);
+function deserializeToJson(message: WebSocket.RawData) {
```
alsuren: Since we're no longer re-serializing it **to** json:

```suggestion
function deserializeJson(message: WebSocket.RawData) {
```

or

```suggestion
function deserializeJsonMessage(message: WebSocket.RawData) {
```

alsuren: Same in the file below.

packages/wrangler/src/worker.d.ts
```diff
@@ -1,4 +1,4 @@
 declare module "worker:*" {
-	const module: { path: string; contents: string };
+	// const module: { path: string; contents: string };
```
alsuren: Commented out code. Is this on purpose?

alsuren (approved):
looks good to me

packages/wrangler/src/worker.d.ts
petebacondarwin: Not sure. I can't remember why I commented this out.
I think because it is not actually needed?

packages/eslint-config-worker/package.json
petebacondarwin: discussed offline

packages/wrangler/src/__tests__/pages/pages-deployment-tail.test.ts
petebacondarwin: 0971a8877

packages/wrangler/src/worker.d.ts
petebacondarwin: Deleting.

petebacondarwin: 6ef01b6ab

packages/wrangler/src/cloudchamber/delete.ts
```diff
@@ -33,7 +34,8 @@ export async function deleteCommand(
 		const deployment = await DeploymentsService.deleteDeploymentV2(
 			deleteArgs.deploymentId
 		);
-		console.log(JSON.stringify(deployment), null, 4);
+		// eslint-disable-next-line no-console
```
penalosa: Is this needed?

packages/wrangler/src/cloudchamber/images/registries.ts
```diff
@@ -42,23 +42,6 @@ function configureImageRegistryOptionalYargs(yargs: CommonYargsArgv) {
...
-		.option("push", {
-			type: "boolean",
-			description: "If you want these credentials to be able to push",
-		})
-		.option("pull", {
-			type: "boolean",
-			description: "If you want these credentials to be able to pull",
-		});
-}
-
```
penalosa, line 61: Is this intentional?

packages/wrangler/src/logger.ts
```diff
@@ -102,6 +102,11 @@ export class Logger {
 
 	columns = process.stdout.columns;
 
+	json = (data: unknown) => {
+		// eslint-disable-next-line no-console
+		console.log(JSON.stringify(data, null, 4));
```
penalosa, line 107: Shouldn't this use `logger.log()`? Otherwise it won't respect the configured log level

penalosa (approved):
Some questions, but overall ✅

packages/wrangler/src/cloudchamber/delete.ts
petebacondarwin: No good spot. Thanks

petebacondarwin: f441cb19d

packages/wrangler/src/cloudchamber/images/registries.ts
petebacondarwin: Yes - these functions are not actually used at runtime, they were only created to make a type.
I could have put eslint exclusions around them but rather chose to get the types elsewhere.

packages/wrangler/src/logger.ts
petebacondarwin: We can't use `logger.log()` because it may add extra "stuff" that would stop the output from being valid JSON.

petebacondarwin: And this helper is kind of outside of the scope of loglevels. It is a way to dump out a programmatic output.

PR #9869, opened 2025-07-07
chore: update workflow permissions

edmundhung:
Fixes n/a.

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/.github/workflows/open-v3-maintenance-prs.yml
+++ b/.github/workflows/open-v3-maintenance-prs.yml
@@ -45,7 +45,7 @@ jobs:
           FILES: ${{ steps.files.outputs.all }}
           PR_NUMBER: ${{ github.event.number }}
           PR_TITLE: ${{ toJson(github.event.pull_request.title) }}
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }} # We cannot create a PR with the default GITHUB_TOKEN with the `pull_request` event.
 
       - name: "Comment on PR with error details"
         if: steps.open_pr.outputs.has_wrangler_patch == 'true' && failure()


.github/workflows/open-v3-maintenance-prs.yml
```diff
@@ -9,8 +9,8 @@ on:
       - ".changeset/**.md"
 
 permissions:
-  contents: read
-  # note: no write permissions are needed since the workflow uses GH_ACCESS_TOKEN instead of GITHUB_TOKEN
+  contents: write # Required to push the v3 backport branch up
```
edmundhung, line 12: This should fix `fatal: unable to access 'https://github.com/cloudflare/workers-sdk/': The requested URL returned error: 403`

.github/workflows/open-v3-maintenance-prs.yml
```diff
@@ -9,8 +9,8 @@ on:
...
+  pull-requests: write # Required for opening the backport PR and commenting on the original PR
```
edmundhung, line 13: This should fix `Error: Resource not accessible by integration`

.github/workflows/open-v3-maintenance-prs.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           FILES: ${{ steps.files.outputs.all }}
           PR_NUMBER: ${{ github.event.number }}
           PR_TITLE: ${{ toJson(github.event.pull_request.title) }}
-          GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
```
edmundhung: I think we were using this GH_TOKEN only for creating the PR but I don't see why we can't use the default github token

.github/workflows/run-ci-for-external-forks.yml
```diff
@@ -15,8 +15,8 @@ on:
         required: true
 
 permissions:
-  contents: read
-  # note: no write permissions are needed since the workflow uses GH_ACCESS_TOKEN instead of GITHUB_TOKEN
+  contents: read # no write permissions are needed since the workflow uses GH_ACCESS_TOKEN instead of GITHUB_TOKEN
+  pull-requests: write # Required for creating the draft PR on behalf of the user
```
edmundhung, line 19: This should fix https://github.com/cloudflare/workers-sdk/actions/runs/16076425204/job/45372623437#step:7:15

dario-piotrowicz (approved):

petebacondarwin (approved):
The reason for using a different token is that you cannot trigger workflows from other workflows - they just don't run. I think this is to prevent infinite recursion taking down teh system?

So if you want to cause another workflow to run based off a change this one is making, it needs to use a different token to the one provided by GH.

Is that relevant here?

edmundhung:
> Is that relevant here?

~I am not sure. But I can't think of another workflow based off this. Let's give this a try :)~

It looks like we can create a PR using the default GitHub token with the `pull_request_target` event but not with `pull_request` event. So I reverted it to use a different token.

PR #9871, opened 2025-07-07
V3 Backport [#9843]: fix(miniflare): improve error logging to include error causes in stack traces

workers-devprod:
This is an automatically opened PR to backport patch changes from #9843 to Wrangler v3


edmundhung:
This is actually more like a new feature / improvement than a bugfix as it involves changes with the upgraded youch version which is only available in v4.

PR #9872, opened 2025-07-07
fix dockerfile path resolution

emily-shen:
Fixes DEVX-2036
- resolves the dockerfile path relative to the wrangler config path, falling back on the cwd. 
- also extracts containers deploy tests out of the really really big deploy.test.ts file because working with a 14,000 line file is not fun. also consolidates mocks that are repeated in the container deploy tests.
- i would like to add a test for dev as well, but we already have this PR: https://github.com/cloudflare/workers-sdk/pull/9728 which was actually dependent on this path resolution being fixed. this bug is blocking quite a few things, so i'll leave the dev test for that PR. I have manually verified this behaviour is fixed in dev - you can easily test it by running `wrangler dev -c ../container-app/wrangler.jsonc` from a random fixture. 

please review by commit

The long term solution should be to normalise the config so that it passes around the absolute path for the dockerfile. This would require some containers config refactoring, which we are doing in a separate follow up PR.

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


dario-piotrowicz (approved):
LGTM 😄

PR #9875, opened 2025-07-07
Make error message for containers images delete more helpful.

gpanders:
Fixes CC-5181.

Print the expected format to make it clear(er) to users that the argument must be of the form IMAGE:TAG.

Also update the error handling for some of the images subcommands so that they properly print errors and exit with a non-zero status code in error cases.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: UX change
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: do not backport containers changes


packages/wrangler/src/cloudchamber/images/images.ts
```diff
@@ -60,7 +60,7 @@ export const imagesCommand = (
 function deleteImageYargs(yargs: CommonYargsArgv) {
 	return yargs.positional("image", {
 		type: "string",
-		description: "image to delete",
+		description: "image and tag to delete",
```
petebacondarwin: is it worth putting the format here too?

petebacondarwin:
```suggestion
		description: "image and tag to delete, of the form IMAGE:TAG",
```

packages/wrangler/src/cloudchamber/images/images.ts
```diff
@@ -82,76 +82,68 @@ async function handleDeleteImageCommand(
...
-			{ message: "Listing" }
-		);
+			return responses_;
+		}),
+		{ message: "Listing" }
+	);
 
-		await listImages(responses, false, args.json);
-	} catch (error) {
-		logger.log(`Error listing images: ${error}`);
```
petebacondarwin, line 153: This seems like a different change not associated with the main point of this PR?

petebacondarwin:
I added the `skip-v3-pr` label because containers stuff does not get backported to v3

packages/wrangler/src/cloudchamber/images/images.ts
gpanders: We changed the error handling for `images delete` as part of this PR, and since the error handling for `images list` had the same issues as `images delete` I changed it there too.

This is mentioned in the commit message and PR description so it's not meant to be sneaky:

>Also update the error handling for some of the images subcommands so that they properly print errors and exit with a non-zero status code in error cases.

packages/wrangler/src/cloudchamber/images/images.ts
gpanders: Added in d6108519

PR #9879, opened 2025-07-07
update containers error message presented when no port is exported by the image

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2024

---

- Tests
  - [ ] Tests included
  - [ ] Tests not necessary because: TODO
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23495
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a wrangler v3 change


packages/containers-shared/src/utils.ts
```diff
@@ -182,8 +182,8 @@ export async function checkExposedPorts(
 	});
 	if (output === "0" && process.platform !== "linux") {
 		throw new Error(
-			`The container "${options.class_name}" does not expose any ports.\n` +
-				"To develop containers locally on non-Linux platforms, you must expose any ports that you call with `getTCPPort()` in your Dockerfile."
+			`The container "${options.class_name}" does not expose any ports, please expose any port that you want to use with \`getTCPPort()\`.\n` +
```
emily-shen: hmm that reads like you should use `getTCPPort` to expose the ports somehow. what about:

The container "${options.class_name}" does not expose any ports. In your Dockerfile, please expose any ports you intend to connect to. For additional etc...

i'd avoid mentioning getTCPPort directly because the containers package abstracts that call away from the user.

dario-piotrowicz: > hmm that reads like you should use getTCPPort to expose the ports somehow

oh.... yeah I think you're right! 😓 


> i'd avoid mentioning getTCPPort directly because the containers package abstracts that call away from the user.

good to know, ok thanks 🙂👍

dario-piotrowicz: Updated 🙂

emily-shen (approved):

PR #9893, opened 2025-07-09
fix zod-format prototype-polluting assignment

odaysec:
https://github.com/cloudflare/workers-sdk/blob/6a0d99066ef0633fb28d8ab3989ce943a02ec514/packages/miniflare/src/zod-format.ts#L406-L406

Most JavaScript objects inherit the properties of the built-in `Object.prototype` object. Prototype pollution is a type of vulnerability in which an attacker is able to modify `Object.prototype`. Since most objects inherit from the compromised `Object.prototype` object, the attacker can use this to tamper with the application logic, and often escalate to remote code execution or cross-site scripting. one way to cause prototype pollution is by modifying an object obtained via a user-controlled property name. Most objects have a special `__proto__` property that refers to `Object.prototype`. An attacker can abuse this special property to trick the application into performing unintended modifications of `Object.prototype`.

fix the `zod-format` need to ensure that the `head` value is validated before being used as a property name in the `annotated` object. This can be achieved by rejecting keys that could lead to prototype pollution, such as `__proto__`, `constructor`, or `prototype`. 

The best approach is to add a validation step for `head` before the assignment on line 406. If `head` contains a disallowed key, the function should throw an error or skip the assignment.

#### References
[Object.prototype.proto](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto)
[Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/miniflare/src/zod-format.ts
+++ b/packages/miniflare/src/zod-format.ts
@@ -403,6 +403,10 @@ function annotate(
 	}
 	assert(isRecord(annotated), "Expected object/array for nested issue");
 	// Recursively annotate
+	assert(
+		head !== "__proto__" && head !== "constructor" && head !== "prototype",
+		`Invalid key "${head}" detected, which could lead to prototype pollution.`
+	);
 	annotated[head] = annotate(
 		groupCounts,
 		annotated[head],


edmundhung:
Hi @odaysec!

Thanks for bringing this up to our attention. We have reviewed the relevant codebase and believe that our usage of the formatZodError helper is **not** vulnerable to prototype pollution. This helper only walks the schema-defined paths reported by Zod during validation of the minifalre option, so it never processes arbitrary input keys like `__proto__` or `constructor`, and those names can never appear in the zod issue path.

If you believe there's a specific scenario where this could still lead to prototype pollution, please provide a minimal reproducible example and report it to https://github.com/cloudflare/workers-sdk/security.

We will close this issue for now, but feel free to reach out if you have any further questions or concerns.

PR #9905, opened 2025-07-09
Implements support for python_modules for packages in Python Workers

dom96:
Fixes EW-9326

We don't want the vendored packages to live in src/vendor anymore, instead we want them to be similar to `node_modules`. This PR implements this but retains support for src/vendor for now, with an error being thrown if both directories are present.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not a public facing feature
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: beta product

### Test Plan

```
$ pnpm run test --filter wrangler -- -t "should print vendor modules correctly in table" src/__tests__/deploy.test.ts
```

### DIFF
--- a/packages/wrangler/src/deployment-bundle/find-additional-modules.ts
+++ b/packages/wrangler/src/deployment-bundle/find-additional-modules.ts
@@ -1,3 +1,4 @@
+import { existsSync } from "node:fs";
 import { mkdir, readdir, readFile, writeFile } from "node:fs/promises";
 import path from "node:path";
 import chalk from "chalk";
@@ -56,11 +57,13 @@ function filterPythonVendorModules(
 	if (!isPythonEntrypoint) {
 		return modules;
 	}
-	return modules.filter((m) => !m.name.startsWith("vendor/"));
+	return modules.filter((m) => !m.name.startsWith("python_modules" + path.sep));
 }
 
 function getPythonVendorModulesSize(modules: CfModule[]): number {
-	const vendorModules = modules.filter((m) => m.name.startsWith("vendor/"));
+	const vendorModules = modules.filter((m) =>
+		m.name.startsWith("python_modules" + path.sep)
+	);
 	return vendorModules.reduce((total, m) => total + m.content.length, 0);
 }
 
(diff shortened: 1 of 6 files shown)


packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -128,6 +129,51 @@ export async function findAdditionalModules(
 				filePath: undefined,
 			});
 		}
+
+		// Look for a `python_modules` directory in the root of the project and add all the .py and .so files in it
+		const pythonModulesDir = path.resolve(entry.projectRoot, "python_modules");
+		const vendorDir = path.resolve(entry.moduleRoot, "vendor");
```
danlapid: why is vendorDir still relevant?

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -128,6 +129,51 @@ export async function findAdditionalModules(
...
+				{ type: "Data", globs: ["**/*.so", "**/*.py"] },
+			];
+			const vendoredModules = (
+				await matchFiles(
+					pythonModulesFiles,
+					pythonModulesDir,
+					parseRules(vendoredRules)
+				)
+			).map((m) => {
+				const prefixedPath = path.join("vendor", m.name);
```
danlapid: vendor?

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
dom96: Because that's where EW expects the vendored modules to be. I guess you're gonna say that should be changed to `python_modules` too?

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -128,6 +129,50 @@ export async function findAdditionalModules(
 				filePath: undefined,
 			});
 		}
+
+		// Look for a `python_modules` directory in the root of the project and add all the .py and .so files in it
```
penalosa, line 135: This seems relatively complex—is there any reason you can't just inject a default rule for python workers when parsing the config? e.g. 

```
[
  { type: "Data", globs: ["python_modules/**/*.so", "python_modules/**/*.py"] },
]
```

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
```diff
@@ -128,6 +129,50 @@ export async function findAdditionalModules(
...
+		const pythonModulesDir = path.resolve(entry.projectRoot, "python_modules");
+		const pythonModulesDirInModuleRoot = path.resolve(entry.moduleRoot, "python_modules");
+
+		// Check for conflict between a `python_modules` directory in the module root and the project root.
+		const pythonModulesExistsInModuleRoot = existsSync(pythonModulesDirInModuleRoot);
+		if (pythonModulesExistsInModuleRoot) {
+			throw new UserError(
+				"The 'python_modules' directory is used during bundling and so cannot exist in your module root. Delete it to continue."
```
penalosa: What does this mean? Python workers aren't bundled

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
danlapid: AFAIU we can't because those rules are relative to `src/`

packages/wrangler/src/deployment-bundle/find-additional-modules.ts
danlapid: `python_modules` directory is auto-generated by pywrangler and bundled into the user bundle along with the user's `src` directory.

penalosa: That seems like a pretty internal detail though—what does this mean from a users perspective?

dom96: It means the user cannot have a `python_modules` directory in their `src`

penalosa: Could you clarify this in the error message?

penalosa:
For posterity, docs tracked at https://jira.cfdata.org/browse/EW-8916

PR #9910, opened 2025-07-09
[wrangler] make sure that the ready-on message is printed after the local runtime is ready

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1978

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: changes related to a non-v3 feature


packages/wrangler/src/dev.ts
```diff
@@ -670,7 +676,17 @@ export async function startDev(args: StartDevOptions) {
 				unregisterHotKeys = registerDevHotKeys(primaryDevEnv, args);
 			}
 		} else {
-			devEnv = new DevEnv();
+			devEnv = new DevEnv({
```
emily-shen: why is this change necessary? is it so we don't try and wait for both the remote and local runtime controllers?

dario-piotrowicz: >  is it so we don't try and wait for both the remote and local runtime controllers?

Yes, by default we get both a local and remote controller:
https://github.com/cloudflare/workers-sdk/blob/3bb69fae168a7254c0eb396ea90cc274d0d9ce92/packages/wrangler/src/api/startDevWorker/DevEnv.ts#L38-L44

But it looks like only one will be actually used (depending on whether we're in local or remote mode), so only one would emit the `reloadComplete` meaning that waiting for both when present would cause the `ready on` message to never appear 😓 

I _think_ that this is the right solution 🤔 I don't really know why we were creating a extra runtime if it's just then left unused (that doesn't seem right to me anyways), I will double check again to make sure that the extra runtime is actually unused

emily-shen: does the multiworker set up have the same problem?

emily-shen: well, it looks like multiworker certainly has some problems 🫠 
https://github.com/cloudflare/workers-sdk/issues/9933

dario-piotrowicz: Yes I did notice that no containers code is present in the multi-worker controller... so I am not fully surprised that it doesn't work 😓 (I didn't look too deeply into it, I was hoping that maybe things worked by inheritance 😅)

dario-piotrowicz: Mh... we are indeed just tearing down the local/remote runtime without using it... so yeah I think that it just makes sense not to start the runtimes unnecessarily if we're just going to then tearing them down 🤔 

https://github.com/cloudflare/workers-sdk/blob/6c4a65258bafaae520c31196593931a7da2e389f/packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts#L357-L360

https://github.com/cloudflare/workers-sdk/blob/6c4a65258bafaae520c31196593931a7da2e389f/packages/wrangler/src/api/startDevWorker/RemoteRuntimeController.ts#L318-L321

emily-shen: sounds good to me. let's solve the multiworker issues in a followup, and merge this one as is?

packages/wrangler/src/dev.ts
```diff
@@ -670,7 +676,16 @@ export async function startDev(args: StartDevOptions) {
...
+				config: new ConfigController(),
+				bundler: new BundlerController(),
+				runtimes: [
+					...(args.remote
+						? [new RemoteRuntimeController()]
+						: [new LocalRuntimeController()]),
+				] as RuntimeController[],
```
penalosa: This isn't correct—it'll break switching between remote and local via hotkeys. I _thought_ we had a test for that, but maybe not?

dario-piotrowicz: Thanks for pointing that out 🙂 

I'll address this, I'll first wait to see if the dust settles on whether we do want to apply this change or not 👍

dario-piotrowicz: I've reverted this change 🙂👍

fixtures/pages-workerjs-app/tests/index.test.ts
```diff
@@ -72,6 +72,15 @@ describe("Pages _worker.js", () => {
 				"--port=0",
 				"--inspector-port=0",
 			]);
+		vi.waitFor(
```
emily-shen, line 75: why are there changes to a pages fixture?

packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -93,6 +95,34 @@ export class DevEnv extends EventEmitter {
 		});
 	}
 
+	/**
+	 * Awaits for all the local runtime controllers to be ready, and when they are
+	 * it resolves a promise on the proxy controller to let it know that the
+	 * local server is now ready to handle requests
+	 */
+	private async resolveProxyLocalServerReady(): Promise<void> {
```
emily-shen: isn't this also waiting for a remote controllers to be ready? (well, torn down)
should it be resolveAllRuntimeControllersReady (or something similar?)

dario-piotrowicz:
> so if i understand correctly, now we wait for all runtime controllers to be ready or torn down before printing ready - this might be a proxy server, the one with the actual user worker, or remote. is that right?

Yes 🙂

fixtures/pages-workerjs-app/tests/index.test.ts
dario-piotrowicz: The test below does the following:
 - clears the output (`clearOutput()`)
 - renames a file
 - makes sure that the output is (still) empty (`expect(getOutput()).toBe("")`)
 
Before `ready on` was being printed basically at the earlier time (as soon as the proxy controller is initialized), before the output gets cleared, but now the log is being printed asynchronously when all runtimes are ready, which actually happens after the output gets cleared (making the `toBe("")` assertion fail).
That's why now we need to wait for the `ready on` message to be there first (so that we can be sure that that won't appear in the output)

If you'd prefer a different change here please let me know 🙂

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: yeah I was thinking of that too... local here stands for the `localServer` not the local runtime 😅

but yeah I agree that it's confusing, I'm happy to go with `resolveAllRuntimeControllersReady` 🙂👍

dario-piotrowicz: updated 🙂👍

packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -93,6 +95,34 @@ export class DevEnv extends EventEmitter {
...
+	 * local server is now ready to handle requests
+	 */
+	private async resolveAllRuntimeControllersReady(): Promise<void> {
+		await Promise.all(
+			this.runtimes
+				.filter(
+					(runtime) =>
+						runtime instanceof RemoteRuntimeController ||
+						runtime instanceof LocalRuntimeController
+				)
```
penalosa: Why are we filtering on runtime type here?

packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -93,6 +95,34 @@ export class DevEnv extends EventEmitter {
...
+				.map((runtime) => {
+					return new Promise<void>((resolve) => {
+						// A runtime controller is ready when it either emits the
+						// reloadComplete event (meaning that the runtime is in use and
+						// ready) or when it emits the teardown event (meaning that the
+						// runtime is not currently in use)
+						runtime.once("reloadComplete", () => resolve());
+						runtime.once("teardown", () => resolve());
+					});
```
penalosa: Should this be a promise race?

packages/wrangler/src/api/startDevWorker/DevEnv.ts
```diff
@@ -93,6 +95,34 @@ export class DevEnv extends EventEmitter {
 		});
 	}
 
+	/**
+	 * Awaits for all the local runtime controllers to be ready, and when they are
+	 * it resolves a promise on the proxy controller to let it know that the
+	 * local server is now ready to handle requests
+	 */
+	private async resolveAllRuntimeControllersReady(): Promise<void> {
```
penalosa: Why is this needed? the proxy controller already receives `reloadComplete` events from the runtime controllers: https://github.com/cloudflare/workers-sdk/pull/9910/files#diff-49321de92dbc2c86de8aea8dc98fb93bc26a4ff1a2e491844f5f6b6ad0bb6b55R87

packages/wrangler/src/api/startDevWorker/BaseController.ts
```diff
@@ -75,6 +75,7 @@ export abstract class Controller<
 type RuntimeControllerEventMap = ControllerEventMap & {
 	reloadStart: [ReloadStartEvent];
 	reloadComplete: [ReloadCompleteEvent];
+	teardown: [];
```
penalosa: What's this for?

packages/wrangler/src/dev.ts
```diff
@@ -684,7 +689,12 @@ export async function startDev(args: StartDevOptions) {
 				unregisterHotKeys = registerDevHotKeys(primaryDevEnv, args);
 			}
 		} else {
-			devEnv = new DevEnv();
+			devEnv = new DevEnv({
+				config: new ConfigController(),
+				bundler: new BundlerController(),
+				runtimes: [new RemoteRuntimeController(), new LocalRuntimeController()],
+				proxy: new ProxyController(),
+			});
```
penalosa: This doesn't look right—why are we passing all these in? We should be able to rely on defaults here (if not, the defaults need changed)

packages/wrangler/src/api/startDevWorker/DevEnv.ts
penalosa: What about resolving the proxy controller promise based on the first reload complete event received in the proxy controller? Then you don't need the new teardown event or any of this logic, I don't think

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: good callout, this is a leftover from my previous iteration, I'll remove it 👍

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: no, we want all the runtime controllers to be ready

for example the runtime controller's teardown even will fire much earlier than the local runtime controller that could be building a docker image, and the whole point of this change is to show the ready-on log after the docker build is completed

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: > What about resolving the proxy controller promise based on the first reload complete event received in the proxy controller? Then you don't need the new teardown event or any of this logic, I don't think

`DevEnv` has in theory an arbitrary number of runtime controllers, do you think that it is always guaranteed that only one will work at a time? If that's the case then yeah we could just listen for the first `reloadComplete` event I'd reckon. Otherwise from a generic perspective checking that all the runtime controllers are done seems the more correct approach to me.

packages/wrangler/src/api/startDevWorker/BaseController.ts
dario-piotrowicz: To satisfy the types, without it you'd get errors such as:
Screenshot 2025-07-18 at 13 22 43

This is just declaring the event, but the event does not carry any data

packages/wrangler/src/dev.ts
dario-piotrowicz: Yes, sorry this is a leftover from the previous iteration of this, thanks for noticing it 🙏 

I've removed it 👍

packages/wrangler/src/api/startDevWorker/DevEnv.ts
penalosa: > do you think that it is always guaranteed that only one will work at a time?

Yes, it is—that's the intention with runtime controllers. Only one should be active at any time, and the rest of DevEnv will only work with a single active runtime controller (for instance, the proxy controller will only proxy a single one at a time)

penalosa: What we actually care about here is "When a user goes to the proxy worker URL are we confident that something will be returned quickly". The proxy worker URL reflects the _first_ runtime controller to emit a `reloadComplete` event, and so waiting for the first `reloadComplete` event should be sufficient

dario-piotrowicz: ok, sounds good, thanks for the clarification, I'll update the code 🙂👍

dario-piotrowicz: Updated :slightly_smiling_face:

packages/wrangler/src/api/startDevWorker/BaseController.ts
dario-piotrowicz: removed :slightly_smiling_face:

packages/wrangler/src/api/startDevWorker/DevEnv.ts
dario-piotrowicz: Updated to `race` :slightly_smiling_face:

packages/wrangler/src/api/startDevWorker/DevEnv.ts
penalosa, line 1: Why does this file have/need any changed? The proxy controller already receives `reloadComplete` events from the runtime—could that not be used to resolve it's promise?

penalosa: As in, could you just call `this.localServerReady.resolve()` here? https://github.com/cloudflare/workers-sdk/pull/9910/files#diff-67a7816fe7acc447751a04c6ae8632f9d8be04bab7041e541b07737860c0bb42L421

dario-piotrowicz: yes, good call 🙂👍

penalosa (approved):
Looks great! Thanks for putting up with all my nitpicking here

dario-piotrowicz:
> Looks great! Thanks for putting up with all my nitpicking here

No problem, thanks a lot for correcting all my silly mistakes! 🙏

PR #9914, opened 2025-07-10
Add support for reading `.env` files as runtime vars in local development

petebacondarwin:
Fixes https://github.com/cloudflare/workers-sdk/discussions/9394

TODO
 - docs changes:
   - https://developers.cloudflare.com/workers/configuration/environment-variables/
   - https://developers.cloudflare.com/workers/development-testing/environment-variables/
   - https://developers.cloudflare.com/workers/wrangler/environments/
   - https://opennext.js.org/cloudflare/howtos/env-vars

- [x] escape hatch for i.e. Next (which load `.env` file using a different mechanism)
        added `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV` opt-out env var

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23837
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  New feature

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -740,9 +740,6 @@ if (import.meta.hot) {
 					});
 					miniflare = undefined;
 				}
-				// Reset the flag so that if a `buildEnd` hook is called again before the next
-				// configureServer hook then we do dispose of miniflare correctly.
-				restartingServer = false;
 			},
 		},
 		// Plugin to provide a fallback entry file


packages/wrangler/e2e/__snapshots__/dev.test.ts.snap
petebacondarwin, line 1: These snapshots were out of date and redundant.

packages/wrangler/e2e/helpers/wrangler.ts
```diff
@@ -25,13 +25,6 @@ export async function runWrangler(
 	return runCommand(getWranglerCommand(wranglerCommand), { cwd, env, timeout });
 }
 
-export function runWranglerLongLived(
-	wranglerCommand: string,
-	options: WranglerCommandOptions = {}
-) {
-	return new WranglerLongLivedCommand(wranglerCommand, options);
-}
-
```
petebacondarwin: This function is not used anywhere.

packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -132,19 +130,14 @@ describe.sequential("wrangler dev", () => {
...
 	runInTempDir();
 	mockAccountId();
 	mockApiToken();
 	const std = mockConsoleMethods();
 	afterEach(() => {
-		patchConsole(() => {});
 		msw.resetHandlers();
-		spy.mockClear();
-		setSpy.mockClear();
-		logger.resetLoggerLevel();
```
petebacondarwin: These helpers are redundant

packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -156,6 +149,11 @@ describe.sequential("wrangler dev", () => {
 		} catch (e) {
 			console.error(e);
 		}
+		if (spy.mock.calls.length === 0) {
+			throw new Error(
+				"Config was never reached:\n" + JSON.stringify(std, null, 2)
+			);
+		}
```
petebacondarwin: Previously if reading the config was never reached in the Wrangler run, we would just get a useless error on the return statement below.
Now we dump out the stdout etc so we can diagnose the problem.

packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -948,45 +946,113 @@ describe.sequential("wrangler dev", () => {
 		});
 
 		describe(".env", () => {
+			const processEnv = process.env;
+			beforeEach(() => (process.env = { ...processEnv }));
+			afterEach(() => (process.env = processEnv));
```
petebacondarwin: The `.env` handling may mutate process.env so we make a copy for the duration of each test.

packages/wrangler/src/__tests__/helpers/run-wrangler.ts
petebacondarwin, line 1: Changes in this file just make use of the Vitest env var mocking rather than us handcrafting it.

packages/wrangler/src/dev/dev-vars.ts
```diff
@@ -18,6 +21,11 @@ import type { Config } from "../config";
  *
  * Any values in this file, formatted like a `dotenv` file, will add to or override `vars`
  * bindings provided in the Wrangler configuration file.
+ *
+ * Note: The `getDotDevDotVarsContent` function in the `packages/vite-plugin-cloudflare/src/index.ts` file
+ *       follows the same logic implemented here, the two need to be kept in sync, so if you modify some logic
+ *       here make sure that, if applicable, the same change is reflected there
+ *
```
petebacondarwin: remove

petebacondarwin: and update the vite-plugin to use this stuff

fixtures/nodejs-hybrid-app/tests/index.test.ts
```diff
@@ -88,4 +88,26 @@ describe("nodejs compat", () => {
 		const response = await fetch(`http://${ip}:${port}/test-http`);
 		await expect(response.text()).resolves.toBe("OK");
 	});
+
+	test("process.env contains vars", async ({ expect }) => {
+		const { ip, port } = wrangler;
+		const response = await fetch(`http://${ip}:${port}/process-env`);
+		await expect(response.json()).resolves.toEqual(
+			expect.objectContaining({
```
vicb: nit: sound like what you want is [`toMatchObject`](https://vitest.dev/api/expect.html#tomatchobject) instead?
(same below)

packages/wrangler/src/api/integrations/platform/index.ts
```diff
@@ -53,6 +54,17 @@ export type GetPlatformProxyOptions = {
...
+	/**
+	 * Paths to `.env` file to load environment variables from.
+	 *
+	 * If not specified, the default behavior:
+	 *  - look for a `.env` file in the same directory as the Wrangler configuration file, or in the current working
+	 *    directory if no Wrangler configuration file is specified.
+	 *  - also look for a `.env.local` file in the same directory as the `.env` file.
+	 *  - if the `environment` option is specified, also look for `.env.<environment>` and `.env.<environment>.local`
+	 *    files in the same directory as the `.env` file.
+	 */
```
vicb: Few comments here:
- maybe disambiguate "not specified" - i.e. "undefined" or empty list
- for `.dev.vars`, a single file is loaded and the priority is the .dev.vars.<env>. I would be nice to clarify here if one or multiple files are loaded and what the priority is.

packages/wrangler/src/config/dot-env.ts
```diff
@@ -0,0 +1,71 @@
+import path from "path";
+import dotenv from "dotenv";
+import dotenvExpand from "dotenv-expand";
+import { logger } from "../logger";
+
+/**
+ * Generates the default set of paths for .env file loading.
```
vicb: So maybe this is the priority?

Could you please clarify in the comment (i.e. "set" -> "ordered list", "order" of what)

penalosa (approved):

packages/wrangler/src/config/dot-env.ts
vicb: Maybe also add a comment for `basePath` to say that it includes `.env`

packages/wrangler/src/dev/dev-vars.ts
```diff
@@ -16,29 +19,88 @@ import type { Config } from "../config";
...
+ * If there are no `.dev.vars*` file, we will look for `.env*` files in the same directory.
+ * If the `--env-file <file-path>` option is set, we'll look for the `.env*` files at that path.
+ *
+ * Any values in these files (all formatted like `.env` files) will add to or override `vars`
  * bindings provided in the Wrangler configuration file.
+ *
  */
 export function getVarsForDev(
-	config: Pick<Config, "userConfigPath" | "vars">,
+	configPath: string | undefined,
```
vicb, line 41: Could you please add
```ts
/** @params @returns */
```

packages/wrangler/src/dev/dev-vars.ts
```diff
@@ -16,29 +19,88 @@ import type { Config } from "../config";
...
+	try {
+		const contents = maybeGetFile(basePath);
+		if (contents === undefined) {
+			logger.debug(
+				`local dev variables file not found at "${path.relative(".", basePath)}". Continuing... For more details, refer to https://developers.cloudflare.com/workers/wrangler/system-environment-variables/`
+			);
+			return;
+		}
+
+		const parsed = dotenv.parse(contents);
```
vicb, line 89: Could you please add a comment whether this can throw?
(If it can then we might fallback to another file instead of error on invalid syntax)

packages/wrangler/src/dev/dev-vars.ts
```diff
@@ -16,29 +19,88 @@ import type { Config } from "../config";
...
+	} catch (e) {
+		logger.debug(
+			`Failed to load local dev variables file "${path.relative(".", basePath)}":`,
+			e
+		);
+	}
+}
+
+/**
+ * Loads a .dev.vars (or .env style) file from `envPath`, preferring to read `${envPath}.${env}` if
```
vicb: `.dev.vars (or .env style)` => maybe change the name "tryLoadDotDevDotVars"

packages/wrangler/src/index.ts
```diff
@@ -400,6 +402,13 @@ export function createCLIParser(argv: string[]) {
 			type: "string",
 			requiresArg: true,
 		})
+		.option("env-file", {
+			describe:
+				"Path to an .env file to load - can be specified multiple times - earlier ones are overridden by later ones",
```
vicb: "earlier ones are overridden by later ones"

Maybe clarify that the values are overridden, not the files

packages/wrangler/src/index.ts
```diff
@@ -419,13 +428,15 @@ export function createCLIParser(argv: string[]) {
...
 		})
 		.check((args) => {
-			// Grab locally specified env params from `.env` file
-			const loaded = loadDotEnv(".env", args.env);
-			for (const [key, value] of Object.entries(loaded?.parsed ?? {})) {
-				if (!(key in process.env)) {
-					process.env[key] = value;
-				}
-			}
+			// Set process environment params from `.env` files if available.
```
vicb, line 431: Question: Is "Set" a Merge? (is that what `includeProcessEnv` means ?).

Maybe add a comment to explain that

vicb: If it is a merge, is `?? process.env` needed?

fixtures/nodejs-hybrid-app/tests/index.test.ts
petebacondarwin: nice!

packages/wrangler/src/dev/dev-vars.ts
petebacondarwin: Great spot! If it can't throw then there is no point in the catch.
I'll return an empty object in that case.

packages/wrangler/src/index.ts
petebacondarwin: Ah yes, it never returns `undefined`.

.changeset/moody-breads-pump.md
```diff
@@ -0,0 +1,20 @@
...
+they will now try to load additional local dev vars from `.env`, `.env.local`, `.env.<environment>` and `.env.<environment>.local` files.
+
+These loaded vars are only for local development and have no effect in production to the vars in a deployed Worker.
+Wrangler and Vite will continue to load `.env` files in order to configure themselves as a tool.
+
+Further details:
+
+- In `vite build` the local vars will be computed and stored in a `.dev.vars` file next to the compiled Worker code, so that `vite preview` can use them.
+- The `wrangler types` command will similarly read the `.env` files (if no `.dev.vars` files) in order to generate the `Env` interface.
+- If the `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV` environment variable is `false` then local dev variables will not be loaded from `.env` files.
```
vicb:
```suggestion
- If the `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV` environment variable is `"false"` then local dev variables will not be loaded from `.env` files.
```

packages/wrangler/src/dev/dev-vars.ts
vicb: Do we really want to really "swallow" the error?

Maybe use `logger.error` and throw an error (`UserError`) ?

packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -285,3 +285,21 @@ export const getDockerHost = getEnvironmentVariableFactory({
...
 	},
 });
+
+/**
+ * `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV` specifies whether to load vars for local dev from `.env` files.
+ */
+export const getCloudflareLoadDevVarsFromDotEnv =
+	getBooleanEnvironmentVariableFactory({
+		variableName: "CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV",
+		defaultValue: () => true,
```
vicb:
```suggestion
		defaultValue: true,
```

packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -285,3 +285,21 @@ export const getDockerHost = getEnvironmentVariableFactory({
...
+	});
+
+/**
+ * `CLOUDFLARE_INCLUDE_PROCESS_ENV` specifies whether to include the `process.env` in vars loaded from `.env` for local development.
+ */
+export const getCloudflareIncludeProcessEnvFromEnv =
+	getBooleanEnvironmentVariableFactory({
+		variableName: "CLOUDFLARE_INCLUDE_PROCESS_ENV",
+		defaultValue: () => false,
```
vicb:
```suggestion
		defaultValue: false,
```

packages/wrangler/src/dev/dev-vars.ts
petebacondarwin: Yeah good call. The `maybeGetFile()` will not throw if the file is missing. So if there is an error then it is bad.

.changeset/moody-breads-pump.md
```diff
@@ -0,0 +1,20 @@
...
+
+These loaded vars are only for local development and have no effect in production to the vars in a deployed Worker.
+Wrangler and Vite will continue to load `.env` files in order to configure themselves as a tool.
+
+Further details:
+
+- In `vite build` the local vars will be computed and stored in a `.dev.vars` file next to the compiled Worker code, so that `vite preview` can use them.
+- The `wrangler types` command will similarly read the `.env` files (if no `.dev.vars` files) in order to generate the `Env` interface.
+- If the `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV` environment variable is `"false"` then local dev variables will not be loaded from `.env` files.
+- If the `CLOUDFLARE_INCLUDE_PROCESS_ENV` environment variable is `"true"` then all the environment variables found on `process.env` will be included as local dev vars.
```
vicb, line 19:
```suggestion
- If the `CLOUDFLARE_LOAD_DEV_VARS_FROM_DOT_ENV` environment variable is `"false"` then local dev variables will not be loaded from `.env` files. Defaults to `"true"`.
- If the `CLOUDFLARE_INCLUDE_PROCESS_ENV` environment variable is `"true"` then all the environment variables found on `process.env` will be included as local dev vars. . Defaults to `"false"`.
```

packages/vite-plugin-cloudflare/playground/dev-vars/__tests__/vars-changes.spec.ts
```diff
@@ -1,34 +1,10 @@
...
-					expect(await getJsonResponse()).toEqual({
-						"variables present in .dev.vars": {
-							MY_DEV_VAR_A: "my .dev.vars variable A",
-							MY_DEV_VAR_B: "my .dev.vars variable B",
-							MY_DEV_VAR_C: "my .dev.vars variable C",
-						},
-					});
-				},
-				{ timeout: 5000 }
-			);
```
jamesopstad, line 29: I notice that your `mockFileChange` helper doesn't wait for the original content to be restored in this way. I'm pretty sure this was necessary before to avoid flakes (agree that `expect` inside `onTestFinished` is a bit strange but couldn't find a good alternative).

petebacondarwin: Ah... Maybe that is it...

petebacondarwin: I found that using `vi.waitFor()` at the start of the tests made them resilient enough.
Previously we had checks at the start that would check that things were in a pristine state but they would only make one attempt.

I think we probably do need to look into this race condition further though. If only I could recreate it locally...

jamesopstad (approved):

packages/vite-plugin-cloudflare/playground/dev-vars/__tests__/vars-changes.spec.ts
jamesopstad: I have a suspicion about what's going on here. You match `.env` files in `hasLocalDevVarsFileChanged` and call `server.restart` when they change. Vite also restarts the server when `.env` files change - https://github.com/vitejs/vite/blob/a1aff2279398c48064e261941cf90cc60a031dee/packages/vite/src/node/server/hmr.ts#L390-L410. This might explain the race condition.

petebacondarwin: Oh! So shall I just remove that? Will the Vite restart be enough to trigger a reload of Miniflare config?

jamesopstad: It will only restart on `.env` files related to the current mode - https://github.com/vitejs/vite/blob/a1aff2279398c48064e261941cf90cc60a031dee/packages/vite/src/node/env.ts#L10. Is that a problem?

jamesopstad: But yes, a restart will reload the Miniflare config.

petebacondarwin: > It will only restart on `.env` files related to the current mode - https://github.com/vitejs/vite/blob/a1aff2279398c48064e261941cf90cc60a031dee/packages/vite/src/node/env.ts#L10. Is that a problem?

Hmm. I think that could be problematic since the Cloudflare environment can be different to the Vite mode. (We even suggest that the `.env.<mode>` file could contain a `CLOUDFLARE_ENV=<cloudflare-env>` property, which would then mean that in local dev we add the fields from `.env.<cloudflare-env>` to the `vars`.

petebacondarwin: When I say it could be a problem, I mean that it would not be enough to rely upon Vite to trigger updates to Miniflare config.

petebacondarwin:
Debugging "AssertionError: Failed to initialize module runner, error: Runner already initialized":

- awaiting a promise from the previous `restartServer()` before running next restart
- removing watcher listeners before restarting doesn't help
-

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
```diff
@@ -160,7 +161,7 @@ beforeAll(async (s) => {
 					viteTestUrl = mod.viteTestUrl ?? viteTestUrl;
 				}
 			} else {
-				await startDefaultServe();
+				server = await startDefaultServe();
```
jamesopstad, line 164: Why set this here when it's set in `startDefaultServe()`?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -336,37 +341,68 @@ if (import.meta.hot) {
...
+						restartingServer = false;
+					}
+				};
+
 				assertIsNotPreview(resolvedPluginConfig);
 
+				// It is possible to get into a situation where the dev server is restarted by a config file change
+				// right in the middle of the Vite server and the supporting Workers being initialized.
+				// We use an abort controller to signal to the initialization code that it should stop if the config has changed.
+				const restartAbortController = new AbortController();
```
jamesopstad, line 365: It would be good to establish if this is still necessary with all the other changes as it makes the code quite a bit more complex.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -336,37 +341,68 @@ if (import.meta.hot) {
...
+						)
+					) {
+						debuglog(configId, "Config changed: " + changedFilePath);
+						if (!restartAbortController.signal.aborted) {
+							debuglog(
+								configId,
+								"Restarting dev server and aborting previous setup"
+							);
+							restartAbortController.abort();
+							await viteDevServer.watcher.close();
```
jamesopstad, line 396: Is this needed when using `watcher.once()`?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -387,18 +423,58 @@ if (import.meta.hot) {
...
+					debuglog(configId, "Creating new Miniflare instance");
 					miniflare = new Miniflare(miniflareDevOptions);
 				} else {
+					debuglog(configId, "Waiting for Miniflare to be ready before update");
+					await miniflare.ready;
+					debuglog(configId, "Updating the Miniflare instance");
 					await miniflare.setOptions(miniflareDevOptions);
+					debuglog(configId, "Waiting for Miniflare to be ready after update");
+					await miniflare.ready;
+					debuglog(configId, "Miniflare is ready");
```
jamesopstad, line 455: Can you explain why `await miniflare.ready` is needed in these two places?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -653,6 +729,18 @@ if (import.meta.hot) {
...
+				if (!restartingServer) {
+					debuglog("buildEnd: disposing Miniflare instance");
+					await miniflare?.dispose().catch((error) => {
+						debuglog("buildEnd: failed to dispose Miniflare instance:", error);
+					});
+					miniflare = undefined;
+				}
+				// Reset the flag so that if a `buildEnd` hook is called again before the next
+				// configureServer hook then we do dispose of miniflare correctly.
+				restartingServer = false;
```
jamesopstad: Can this definitely happen?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -336,37 +341,68 @@ if (import.meta.hot) {
...
+
 				assertIsNotPreview(resolvedPluginConfig);
 
+				// It is possible to get into a situation where the dev server is restarted by a config file change
+				// right in the middle of the Vite server and the supporting Workers being initialized.
+				// We use an abort controller to signal to the initialization code that it should stop if the config has changed.
+				const restartAbortController = new AbortController();
+
+				// We use a `configId` to help debug how the config changes are triggering the restarts.
+				const configId = randomUUID();
```
jamesopstad, line 368: Creating this value and passing it around has added a lot of code. Do we still need it now that you've resolved the issues?

jamesopstad (commented):
Great work persevering Pete!

A couple of general comments in addition to those inline:
- A lot more of the tests now wrap `expect` in `waitUntil`. As I understand it each test file should now get its own `Miniflare` instance so this shouldn't be necessary unless the tests cause the dev server to restart. Is that incorrect?
- There is now a requirement for tests that use `page` to call `page.goto(viteTestUrl)` manually at the start of each test. I think that's OK but just something we'll have to watch out for as tests in the same file now continue from the previous test's state.
- If we're going to start including debug logs then I think we could do this more generally  and use specifiers like Vite does. Not for this PR but might be nice to follow up using the approach [here](https://github.com/vitejs/vite/blob/946831f986cb797009b8178659d2b31f570c44ff/packages/vite/src/node/utils.ts#L200).
- Generally, I'm most concerned about how much code has been added and that it makes reading the source more challenging. If there's any room for now stripping back some of the changes that have been made then that would be good.
- For posterity, it would be really helpful if you could add a comment to this PR summarising what the issues were and how they've been resolved.

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: Agh. That was leftover from the other approach I used to check for restarting.

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: It's probably not. I was just chucking stuff in there to see if waiting for things helps.

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: I just realised that `watcher.once()` is incorrect, since it will be triggered by lots of other file changes.
I meant to revert that and just remove the watcher if we are reverting.
It happens to work in the tests because we don't touch any other files.

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: Yes, we should rip this out if it is not needed

packages/vite-plugin-cloudflare/playground/vitest-setup.ts
petebacondarwin: Good point. I think at this stage I was being overly conservative since I wasn't sure whether the `startDefaultServe()` was mutating things correctly. Arguably it shouldn't mutate `server` at all.

petebacondarwin:
Fixes attempted that should be checked and reverted if not necessary:

 - [ ] remove `beforeEach(() => page.goto())` from vitest-setup.ts and apply it where needed at the start of tests
   - this was because this hook often crashed with a `ERRCONNREFUSED` error that I couldn't use `vi.waitFor()` nor simply catch in the hook as that then caused other tests to fail
   - https://github.com/cloudflare/workers-sdk/pull/10140
 - [ ] run `await miniflare.ready` all over the place
   - probably not needed any more
   - https://github.com/cloudflare/workers-sdk/pull/10135
 - [ ] add abort signal code to avoid running `initRunners()`
   - this was the original fix attempt but doesn't necessarily help if the problem is across tests
   - https://github.com/cloudflare/workers-sdk/pull/10137
 - [ ] replace `poolOptions: `{ forks: { singleFork: true } }` with `fileParallelism: false`
   - this is actually probably a solid part of the fix
 - [ ] use of `debuglog()` all over the place for diagnosing the failures
   - could try switching to the Vite equivalent (e.g. `createDebugger()`)
 - [ ] add `await viteDevServer.watcher.close();` to the watch handler
   - almost definitely not needed now that it has a `watcher.off()`
   - https://github.com/cloudflare/workers-sdk/pull/10138

packages/vite-plugin-cloudflare/src/index.ts
petebacondarwin: Removed.

PR #9922, opened 2025-07-10
Revert "Build containers without account ID"

edmundhung:
Reverts cloudflare/workers-sdk#9765


gpanders:
We have a fix in https://github.com/cloudflare/workers-sdk/pull/9923

PR #9923, opened 2025-07-10
Resolve container image name when modifying

gpanders:
The container image name should be resolved into the full (namespaced) image name both when creating an application and when modifying an application.

Move the image name resolution out of the create path and into the shared code path taken for both creation and modification.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: containers change


packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -1904,7 +1904,7 @@ describe("cloudchamber apply", () => {
 			│   tier = 2
 			│
 			│   [containers.configuration]
-			│   image = \\"${registry}/hello:1.0\\"
+			│   image = \\"${registry}/some-account-id/hello:1.0\\"
```
gpanders, line 1907: Before, we were also omitting the namespace (account ID) in the diff that we showed to the user. This ends up being more trouble than it's worth, so we include the namespace (account ID) in the diff now. Updated this test to reflect that.

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -1915,4 +1915,79 @@ describe("cloudchamber apply", () => {
...
+		setIsTTY(false);
+		const registry = getCloudflareContainerRegistry();
+		writeWranglerConfig({
+			name: "my-container",
+			containers: [
+				{
+					name: "my-container-app",
+					instances: 3,
+					class_name: "DurableObjectClass",
+					image: `${registry}/hello:1.0`,
```
gpanders, line 1929: Note: image name _does not_ include namespace (account ID) in Wrangler config here

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -1915,4 +1915,79 @@ describe("cloudchamber apply", () => {
...
+			{
+				id: "abc",
+				name: "my-container-app",
+				instances: 3,
+				created_at: new Date().toString(),
+				version: 1,
+				account_id: "1",
+				scheduling_policy: SchedulingPolicy.REGIONAL,
+				configuration: {
+					image: `${registry}/some-account-id/hello:1.0`,
```
gpanders, line 1948: But in the API, we do include namespace (account ID) in the image name

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -1915,4 +1915,79 @@ describe("cloudchamber apply", () => {
...
+		await runWrangler("cloudchamber apply");
+		expect(std.stdout).toMatchInlineSnapshot(`
+			"╭ Deploy a container application deploy changes to your application
+			│
+			│ Container application changes
+			│
+			├ EDIT my-container-app
+			│
+			│   [containers.configuration]
+			│   image = \\"${registry}/some-account-id/hello:1.0\\"
```
gpanders, line 1973: And the namespace (account ID) is included in the diff here as well (unmodified).

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -216,10 +221,17 @@ function containerAppToCreateApplication(
 			telemetryMessage: true,
 		});
 	}
+
+	console.log(`configuration.image=${configuration.image}`);
```
gpanders: oops

emily-shen (approved):

PR #9927, opened 2025-07-10
refactor containers/cloudchamber config - part 1

emily-shen:
This PR simply copies the content from cloudchamber/apply.ts to containers/deploy.ts, and does the same for the corresponding tests. (also moves some shared diffing code).

Currently, the same code is being reused for both containers and cloudchamber commands, and there is some confusion resulting from this as the wrangler config shapes for each are quite different. Furthermore, the `cloudchamber apply` command has actually been reading config.containers rather than config.cloudchamber (unintentionally according to @gpanders).

Much more substantial changes to containers/deploy.ts will be happening in follow up PRs, but i'm separating these out first asap in order to prevent more unintentional changes to `wrangler cloudchamber apply` (and because i can't keep up with the merge conflicts 😅).

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/src/__tests__/containers/apply.test.ts
+++ b/packages/wrangler/src/__tests__/containers/apply.test.ts
@@ -1,0 +1,1959 @@
+import {
+	getCloudflareContainerRegistry,
+	OpenAPI,
+	SchedulingPolicy,
+	SecretAccessType,
+} from "@cloudflare/containers-shared";
+import { http, HttpResponse } from "msw";
+import patchConsole from "patch-console";
+import { apply } from "../../containers/deploy";
+import { mockAccount } from "../cloudchamber/utils";
+import { mockAccountId, mockApiToken } from "../helpers/mock-account-id";
+import { mockCLIOutput } from "../helpers/mock-console";
+import { useMockIsTTY } from "../helpers/mock-istty";
+import { msw } from "../helpers/msw";
+import { runInTempDir } from "../helpers/run-in-tmp";
+import type { Config } from "../../config";
+import type {
+	Application,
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/__tests__/containers/apply.test.ts
```diff
@@ -0,0 +1,1959 @@
...
+					memory: "256MB",
+					memory_mib: 256,
+				},
+				constraints: {
+					tier: 3,
+				},
+			},
+		]);
+		const applicationReqBodyPromise = mockModifyApplication();
+		await apply(
```
emily-shen, line 193: the only real difference between this and cloudchamber/apply.test.ts is that this calls apply directly instead of `wrangler cloudchamber apply`

packages/wrangler/src/containers/deploy.ts
emily-shen, line 1: this is also pretty much the same but i removed the more egregiously `cloudchamber apply` only bits. notably the command itself, and the messaging if you don't have anything in config.containers because that is unreachable from wrangler deploy. 

more substantive changes to come

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -258,175 +268,6 @@ function containerAppToCreateApplication(
 	return app;
 }
 
-function isNumber(c: string | number) {
```
emily-shen, line 261: this is moved into diff.ts with no changes.

PR #9908, opened 2025-07-09
perf: graduate asset-server binary search experiment to 100%

GregBrimble:
Graduates the asset-server binary search experiment from 50% sampling to 100% usage of the improved implementation.

The iterative binary search algorithm is currently being tested at 50% sample rate to validate its performance compared to the original recursive implementation. This PR prepares the graduation to 100% usage once the experiment proves successful.

## Changes

- Replace experimental logic in `worker.ts` with direct usage of improved implementation
- Update `assets-manifest.ts` to use iterative binary search algorithm  
- Remove experimental files (`assets-manifest.2.ts` and its test file)
- Update tests to work with new implementation signature
- Add changeset documenting the performance improvement

## Performance Impact

The improved implementation is expected to provide better performance characteristics for asset manifest lookups, which is currently being validated through the 50% experiment rollout.

@vicb This should be merged if the 50% rollout looks good and shows improvement over the original implementation.

---

- Tests
  - [x] Tests included
  - [ ] Tests not necessary because:
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Internal performance optimization
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Not a Wrangler change, workers-shared package change

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #9931, opened 2025-07-11
refactor: move compare to the tests

vicb:
Moving `compare` to the tests where it is used.

We do not need `compare` in production code as it is only used in the tests.

@GregBrimble I had that change in my initial PR but didn't pay attention that it was re-introduced in https://github.com/cloudflare/workers-sdk/pull/9908. I don't think it belongs there and we need to embed that code in the worker but feel free to close the PR if you did that on purpose.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: refactor
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not wrangler


vicb:
rebased and formatted

PR #9934, opened 2025-07-11
add more validation to containers config

emily-shen:
Fixes CC-5480

- validates types for all fields
- warns for unexpected fields
- warns on deprecated fields

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: config already documented
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -2421,13 +2421,17 @@ function validateContainerApp(
 				name += config === undefined ? "" : `-${envName}`;
 				containerAppOptional.name = name.toLowerCase().replace(/ /g, "-");
 			}
-
 			if (
 				!containerAppOptional.configuration?.image &&
 				!containerAppOptional.image
 			) {
 				diagnostics.errors.push(
 					`"containers.image" field must be defined for each container app. This should be the path to your Dockerfile or a image URI pointing to the Cloudflare registry.`
+				);
+			}
+			if ("configuration" in containerAppOptional) {
+				diagnostics.warnings.push(
+					`"containers.configuration" is deprecated. Use top level "containers" fields instead. "configuration.image" should be "image", "configuration.disk" should be set via "instance_type".`
 				);
 			}
 
@@ -2522,6 +2526,19 @@ function validateContainerApp(
 				"string",
 				["regional", "moon", "default"]
 			);
+
+			// Add deprecation warnings for legacy fields
+			if ("instances" in containerAppOptional) {
+				diagnostics.warnings.push(
+					`"containers.instances" is deprecated. Use "containers.max_instances" instead.`
+				);
+			}
+			if ("durable_objects" in containerAppOptional) {
+				diagnostics.warnings.push(
+					`"containers.durable_objects" is deprecated. Use the "class_name" field instead.`
+				);
+			}
+
 			validateAdditionalProperties(
 				diagnostics,
 				field,
@@ -2540,6 +2557,7 @@ function validateContainerApp(
 					"constraints",
 					"rollout_step_percentage",
 					"rollout_kind",
+					"durable_objects",
 				]
 			);
 		}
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/config/validation.ts
```diff
@@ -2500,16 +2491,85 @@ function validateContainerApp(
...
+				field,
+				"instance_type",
+				containerAppOptional.instance_type,
+				"string",
+				["dev", "basic", "standard"]
+			);
+			validateOptionalProperty(
+				diagnostics,
+				field,
+				"max_instances",
```
emily-shen, line 2503: is there a range this should be within? also should it warn if its unset as it defaults to 0, iirc?

dario-piotrowicz (approved):
LGTM 🙂

packages/wrangler/src/config/validation.ts
```diff
@@ -2500,16 +2491,91 @@ function validateContainerApp(
...
+					"rollout_kind",
+					"durable_objects",
+				]
+			);
+			validateAdditionalProperties(
+				diagnostics,
+				`${field}.configuration`,
+				Object.keys(containerAppOptional.configuration),
+				["image", "secrets", "labels", "disk", "memory", "vcpu", "memory_mib"]
+			);
```
emily-shen: are there any other fields that are being set, but aren't in the schema in `environment.ts`? since i noticed a couple of these configuration fields are being set by some customers but not actually in the schema (e.g. memory, vcpu)

packages/wrangler/src/config/validation.ts
```diff
@@ -2500,14 +2491,91 @@ function validateContainerApp(
...
+					"rollout_kind",
+					"durable_objects",
+				]
+			);
+			if ("configuration" in containerAppOptional) {
+				validateAdditionalProperties(
+					diagnostics,
+					`${field}.configuration`,
+					Object.keys(containerAppOptional.configuration),
+					["image", "secrets", "labels", "disk", "memory", "vcpu", "memory_mib"]
```
```suggestion
					["image", "secrets", "labels", "disk", "vcpu", "memory_mib"]
```
require the use of `memory_mib` for custom instance types

packages/wrangler/src/config/validation.ts
IRCody: Our API at allows 0 so the default should be ok.

The API will error if it's <0 so should maybe validate that it's at least 0?

packages/wrangler/src/config/validation.ts
```diff
@@ -2490,14 +2481,99 @@ function validateContainerApp(
...
+					"rollout_kind",
+					"durable_objects",
+				]
+			);
+			if ("configuration" in containerAppOptional) {
+				validateAdditionalProperties(
+					diagnostics,
+					`${field}.configuration`,
+					Object.keys(containerAppOptional.configuration),
+					["image", "secrets", "labels", "disk", "vcpu", "memory_mib"]
```
gpanders, line 2576: I don't think `vcpu` and `memory_mib` are included in `configuration`? https://github.com/cloudflare/workers-sdk/blob/64af5da7270bcd73db74cf5742bcadb76e5d1f8b/packages/wrangler/src/config/environment.ts#L109-L114

Users shouldn't be able to set these directly in the Wrangler config anyway, they are set by the instance type.

nikitassharma: these fields are being used right now for customers that need custom instance types, see https://github.com/cloudflare/workers-sdk/pull/9962 for better handling for custom instance types

gpanders: Do they need to be added to the type definition then?

emily-shen: i'd strongly prefer to have them in the type definition so that we're not trying to read values that typescript doesn't expect to be there. they're hidden from the wrangler config schema, so we don't need to worry about showing this to the general user.

nikitassharma: I agree, they should be in the type definition

PR #9951, opened 2025-07-14
Recommend remote bindings to remote-mode users

penalosa:
Recommend [remote bindings](https://developers.cloudflare.com/changelog/2025-06-18-remote-bindings-beta/) when `wrangler dev --remote` is used

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: simple messaging change
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no change in behaviour
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: advertising a v4-only feature


packages/wrangler/src/dev.ts
```diff
@@ -633,6 +635,18 @@ export async function startDev(args: StartDevOptions) {
 			};
 		};
 
+		if (args.remote) {
+			logger.log(
+				bold(
+					dedent`
+						Support for remote bindings in ${green("`wrangler dev`")} is now available in public beta as a replacement for ${green("`wrangler dev --remote`")}. Try it out now with ${green("`wrangler dev --x-remote-bindings`")} and let us know how it goes!
+						This gives you access to remote data in development while retaining all the usual benefits of local dev: fast iteration speed, breakpoint debugging, and more.
```
petebacondarwin: Perhaps?

```suggestion
						This gives you access to remote resources in development while retaining all the usual benefits of local dev: fast iteration speed, breakpoint debugging, and more.
```

packages/wrangler/src/dev.ts
```diff
@@ -633,6 +635,18 @@ export async function startDev(args: StartDevOptions) {
 			};
 		};
 
+		if (args.remote) {
+			logger.log(
+				bold(
+					dedent`
+						Support for remote bindings in ${green("`wrangler dev`")} is now available in public beta as a replacement for ${green("`wrangler dev --remote`")}. Try it out now with ${green("`wrangler dev --x-remote-bindings`")} and let us know how it goes!
```
edmundhung: Should we clarify the need to enable the `experimental_remote` option as well? Something like this?

```suggestion
						Support for remote bindings in ${green("`wrangler dev`")} is now available in public beta as a replacement for ${green("`wrangler dev --remote`")}. Try it out now by running ${green("`wrangler dev --x-remote-bindings`")} with the `experimental_remote` option enabled on your resources and let us know how it goes!
```

PR #9953, opened 2025-07-14
V3 Backport [#9951]: Recommend remote bindings to remote-mode users

workers-devprod:
This is an automatically opened PR to backport patch changes from #9951 to Wrangler v3


lrapoport-cf:
backport for https://github.com/cloudflare/workers-sdk/pull/9951 is not needed, as it's advertising for a feature introduced in v4

PR #9954, opened 2025-07-14
Support Images binding in `getPlatformProxy()`

penalosa:
Adds support for the Images binding in `getPlatformProxy()`. Images bindings are some of the most complex bindings from an API perspective, other than RPC. In particular, they expose a _synchronous_ API that accepts ReadableStream arguments. Multiple synchronous APIs are chained together in a builder pattern (e.g. `await env.IMAGES.input(stream).transform(...).output(...)`) before the final `.output()` call is awaited. This breaks our assumptions around functions that accept ReadableStream arguments always being async, and so doesn't work without some special casing. 

In future, this could be replaced with a client JSRPC implementation (using promise pipelining), but for now special casing the entire binding seems to work well.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/9970
  - [ ] Not necessary because:


fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
```diff
@@ -199,6 +202,43 @@ describe("getPlatformProxy - env", () => {
 		}
 	});
 
+	it.only("correctly obtains functioning Image bindings", async () => {
```
petebacondarwin: only!

packages/miniflare/src/plugins/core/proxy/client.ts
```diff
@@ -282,11 +290,69 @@ class ProxyStubHandler<T extends object>
...
+		const binding = {
+			info: (stream: ReadableStream<Uint8Array>) => {
+				// @ts-expect-error The stream types are mismatched
+				return (this.bridge.getProxy(target) as ImagesBinding)["info"](stream);
+			},
+			input: (stream: ReadableStream<Uint8Array>) => {
+				return transformer(this.bridge.getProxy(target), stream, []);
+			},
+		};
+		console.log(binding);
```
petebacondarwin: remove?

fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
```diff
@@ -199,6 +202,43 @@ describe("getPlatformProxy - env", () => {
...
+		expect.extend({ toMatchImageSnapshot });
+
+		const { env, dispose } = await getPlatformProxy<Env>({
+			configPath: wranglerConfigFilePath,
+		});
+		try {
+			const { IMAGES } = env;
+			const streams = (
+				await fetch("https://playground.devprod.cloudflare.dev/flares.png")
+			).body?.tee()!;
```
petebacondarwin:
```suggestion
			const streams = (
				await fetch("https://playground.devprod.cloudflare.dev/flares.png")
			).body!.tee();
```

fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
```diff
@@ -199,6 +202,43 @@ describe("getPlatformProxy - env", () => {
...
+
+			// @ts-expect-error The stream types aren't matching up properly?
```
petebacondarwin, line 217:
```suggestion
			// @ts-expect-error `ReadableStream` between Node.js and Workers doesn't quite along (Property 'readAtLeast' is missing in type 'import("stream/web").ReadableStreamBYOBReader')
```

This method is [non-standard](https://developers.cloudflare.com/workers/runtime-apis/streams/readablestreambyobreader/#:~:text=readAtLeast%20is%20a%20non%2Dstandard%20extension%20to%20the%20Streams%20API%20which%20allows%20users%20to%20specify%20that%20at%20least%20minBytes%20bytes%20must%20be%20read%20into%20the%20buffer%20before%20resolving%20the%20read.)

What happens if the IMAGE binding actually starts to use this?

Should we have some kind of transform/polyfill for `getPlatformProxy()`?

penalosa: I think this will be fine at runtime, because by the time the `ReadableStream` gets to the Images binding it'll be a Workers Readable Stream. I'm not quite sure how we should fix this from a typing perspective though.

penalosa: Maybe a wrapper around the env type? e.g. `getPlatformProxy()` currently returns `Env` as-is, maybe it should be `ConvertToNodeTypes<Env>` or something?

PR #9955, opened 2025-07-14
Use the Cloudflare SDK

penalosa:
Milestone 0 of what will be a relatively long-term project to transition Wrangler to use the [Cloudflare SDK](https://www.npmjs.com/package/cloudflare). This PR:
- Integrates the SDK
- Uses it for 2 low-risk operations: `wrangler kv namespace list`, and `wrangler kv namespace create`

There are still various hacks required to get this working (`@ts-expect-error` and dummy auth credentials), but these are being worked on in the background and will be fixed with a future upgrade to the SDK version.

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: no user-facing impact

### DIFF
--- a/packages/wrangler/src/kv/index.ts
+++ b/packages/wrangler/src/kv/index.ts
@@ -141,7 +141,7 @@ export const kvNamespaceListCommand = createCommand({
 
 		for await (const namespace of sdk.kv.namespaces.list({
 			account_id: accountId,
-			per_page: 500,
+			per_page: 1000,
 			order: "title",
 			direction: "asc",
 		})) {


.changeset/serious-islands-hope.md
dario-piotrowicz, line 1: Why does this internal change require a changeset?

packages/wrangler/src/__tests__/kv.test.ts
```diff
@@ -294,7 +292,7 @@ describe("wrangler", () => {
 				await runWrangler("kv namespace list");
 
 				expect(JSON.parse(std.out)).toEqual(kvNamespaces);
-				expect(requests.count).toEqual(6);
+				expect(requests.count).toBeGreaterThan(1);
```
dario-piotrowicz, line 340: Are we sure this change is correct? With this change we're not checking the pagination anymore, are we?
if you want to be more generic here instead of using `toEqual`, what about adding a `toBeLessThan` check as well?

e.g.
```ts
expect(requests.count).toBeGreaterThan(1);
expect(requests.count).toBeLessThan(100);
```

packages/wrangler/src/cfetch/internal.ts
```diff
@@ -9,7 +11,86 @@ import { loginOrRefreshIfRequired, requireApiToken } from "../user";
...
+		// However, `cloudflare` doesn't like not being given an API token, so we provide a dummy one
+		// Otherwise, errors like "Could not resolve authentication method." are thrown.
+		apiToken: "dummy",
+		baseURL: getCloudflareApiBaseUrl(complianceConfig),
+	});
+}
+
+export function apiErrorCode(error: unknown, code: number): boolean {
+	return error instanceof Cloudflare.APIError && error.errors[0].code === code;
+}
```
dario-piotrowicz:
```suggestion
```

Is this function actually used? 🤔

dario-piotrowicz (commented):
Awesome stuff! 😄 🚀 

I've left a few minor comments

flakey5 (approved):

.changeset/serious-islands-hope.md
penalosa: Because otherwise it won't be released

packages/wrangler/src/__tests__/kv.test.ts
penalosa: > With this change we're not checking the pagination anymore, are we?

We're not, but I think that was checking too much in the way of internal implementation details. Now, we're just checking:
1. That we make more than 1 request (so pagination is happening)
2. And that the output is correct

dario-piotrowicz: aren't those two points mostly covered in the previous test?

https://github.com/cloudflare/workers-sdk/blob/1ca9b8ea2088ecad928f2ed1a3a4dce25253b519/packages/wrangler/src/__tests__/kv.test.ts#L273-L283

as far as I can tell this test is specific for pagination, so I think that we should either keep it (and keep testing the pagination aspect of it) or remove it entirely (and simply add the `request.count` check in the listing test since the basic listing test)

.changeset/serious-islands-hope.md
dario-piotrowicz: it'd be eventually released on the next release no? 🤔

packages/wrangler/src/__tests__/kv.test.ts
dario-piotrowicz: PS: I also misread the test... I thought that it was testing the number of kvs... 🤦 yeah I agree with you then that resting the number of requests is just testing an implementation details 😕

dario-piotrowicz: ok, sorry yeah all good 🙂

packages/wrangler/src/cfetch/internal.ts
```diff
@@ -9,7 +11,86 @@ import { loginOrRefreshIfRequired, requireApiToken } from "../user";
...
+	logger.debugWithSanitization("RESPONSE:", jsonText);
+	logger.debug("-- END CF API RESPONSE");
+}
+/**
+ * This function constructs an instance of the `Cloudflare` SDK client,
+ * with a custom fetcher that uses `fetchInternal`.
+ */
+export function createCloudflareClient(complianceConfig: ComplianceConfig) {
+	return new Cloudflare({
+		// @ts-expect-error Something is messed up in the `cloudflare` fetch typings...
```
1000hz: Fixing this involves pnpm-overriding the version of `undici-types` depended on by `@types/node` to match that of the version of `undici` being used in wrangler. Unfortunately, that does mean overriding the type of global `fetch`, but should be ok since it's not being used.

.changeset/serious-islands-hope.md
penalosa: Yeah? But we tend to use changesets for releasable artifacts, even if they're just internal refactors. e.g. the command registry refactor had a changeset.

Maybe we need a new type of changeset that marks things as to-be-released but not to be included in the public release notes?

packages/wrangler/src/cfetch/internal.ts
penalosa: Yeah, this seems to work. Alternatively we could upgrade @types/node to v24, but that seems potentially problematic since we still support v20. As you say, we don't use global fetch so this shouldn't be a problem.

penalosa: I've added an eslint rule to ban global fetch, which should help enforce this

packages/wrangler/src/cfetch/internal.ts
penalosa: Not any more! I've removed it

.changeset/serious-islands-hope.md
dario-piotrowicz: > Yeah? But we tend to use changesets for releasable artifacts, even if they're just internal refactors. e.g. the command registry refactor had a changeset.

I see, ok I understand, I'm ok with it, especially since this seems to be an established pattern here 🙂👍 

> Maybe we need a new type of changeset that marks things as to-be-released but not to be included in the public release notes?

Nah, don't worry about it I think this is fine 😄👍

dario-piotrowicz (approved):

lrapoport-cf (approved):

lrapoport-cf:
@penalosa  i've left approval, but as discussed let's please get kv team review on this as well before landing 👍

packages/wrangler/src/kv/index.ts
```diff
@@ -135,16 +138,21 @@ export const kvNamespaceListCommand = createCommand({
...
 
-		// TODO: we should show bindings if they exist for given ids
+		const allNamespaces = [];
 
-		logger.log(
-			JSON.stringify(await listKVNamespaces(config, accountId), null, "  ")
-		);
+		for await (const namespace of sdk.kv.namespaces.list({
+			account_id: accountId,
+			per_page: 500,
```
ferhatelmas: max is 100

https://developers.cloudflare.com/api/resources/kv/subresources/namespaces/methods/list/

ferhatelmas: Updating docs for 1000 so fine

PR #9960, opened 2025-07-14
Add some basic smoke tests to check that workers with multiple containers work as expected

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2086

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: test changes
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: changes to a non-v3 feature

### DIFF
--- a/fixtures/interactive-dev-tests/multi-containers-app/src/index.ts
+++ b/fixtures/interactive-dev-tests/multi-containers-app/src/index.ts
@@ -14,6 +14,10 @@ class FixtureTestContainerBase extends DurableObject<Env> {
 				entrypoint: ["node", "app.js"],
 				enableInternet: false,
 			});
+			// On the first request we simply start the container and return,
+			// on the following requests the container can actually be accessed.
+			// Note that we do this this way becase container.start is not awaitable
+			// meaning that we can't simply wait here for the container to be ready
 			return new Response("Container started");
 		}
 		return this.container
--- a/fixtures/interactive-dev-tests/multi-containers-app/src/index.ts
+++ b/fixtures/interactive-dev-tests/multi-containers-app/src/index.ts
@@ -1,0 +1,45 @@
+import { DurableObject } from "cloudflare:workers";
+
+class FixtureTestContainerBase extends DurableObject<Env> {
+	container: globalThis.Container;
+
+	constructor(ctx: DurableObjectState, env: Env) {
+		super(ctx, env);
+		this.container = ctx.container;
+	}
+
+	async fetch(req: Request) {
+		if (!this.container.running) {
+			this.container.start({
+				entrypoint: ["node", "app.js"],
+				enableInternet: false,
+			});
+			return new Response("Container started");
+		}
+		return this.container
+			.getTcpPort(8080)
+			.fetch("http://foo/bar/baz", { method: "POST", body: "hello" });
+	}
+}
+
+export class FixtureTestContainerA extends FixtureTestContainerBase {}
+export class FixtureTestContainerB extends FixtureTestContainerBase {}
+
+export default {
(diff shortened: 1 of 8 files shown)


fixtures/interactive-dev-tests/tests/index.test.ts
```diff
@@ -561,6 +561,170 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
...
+		});
+
+		async function fetchWithTimeout(input: RequestInfo) {
+			const controller = new AbortController();
+			const { signal } = controller;
+			setTimeout(3_000).then(() => {
+				controller.abort();
+			});
+			return await fetch(input, { signal });
+		}
```
dario-piotrowicz, line 622: @emily-shen 🙂 

I call `fetchWithTimeout` inside the `waitFor` callbacks, to prevent `fetch` from actually getting stuck and consuming the whole `waitFor` timeout time 🙂

emily-shen: oh that's way better. yeah the problem is that fetching the container results in a hang if you hit it at the really specific moment where docker says the port is ready but the app inside hasn't started listening yet :/

fixtures/interactive-dev-tests/multi-containers-app/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+		this.container = ctx.container;
+	}
+
+	async fetch(req: Request) {
+		if (!this.container.running) {
+			this.container.start({
+				entrypoint: ["node", "app.js"],
+				enableInternet: false,
+			});
+			return new Response("Container started");
```
emily-shen, line 21:
```suggestion
```
does it work if we don't return and just wait for the fetch to the container to work? since right now the first request will always fail the test assertion because it gets container started, which can be a bit confusing?

dario-piotrowicz: no that wouldn't work because `this.container.start` is not awaitable so we don't know how long that can take 🤔 

> since right now the first request will always fail the test assertion because it gets container started, which can be a bit confusing?

Yes, I've set the `vi.waitFor` to make requests until it gets the expected response, this means that the first response the test will see will be this one, is that confusing?

emily-shen: not a huge deal, but maybe leave a quick comment explaining the expected behaviour?

dario-piotrowicz: I added this comment, is this ok? 🙂🙏 

https://github.com/cloudflare/workers-sdk/pull/9960/commits/b9bb43d6e68c658e0878ba2f2ed244b3eae5162d

fixtures/interactive-dev-tests/multi-containers-app/wrangler.jsonc
```diff
@@ -0,0 +1,41 @@
...
+			{
+				"class_name": "FixtureTestContainerB",
+				"name": "CONTAINER_B",
+			},
+		],
+	},
+	"migrations": [
+		{
+			"tag": "v1",
+			"new_classes": ["FixtureTestContainerA"],
```
emily-shen: nit - this should be new_sqlite_classes. 
also we should validate this at dev :( - yet another ticket to add

dario-piotrowicz: mh... right... 😓 👍

PR #9961, opened 2025-07-14
refactor containers config handling

emily-shen:
The main point of this PR is to add a function that normalises containers config in containers/config.ts, and is more narrowly typed than the wrangler config schema. This should hopefully make the flow from wrangler config to API clearer and safer.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


packages/containers-shared/src/build.ts
emily-shen, line 1: the dockerfile and build context paths get transformed into absolute paths by `getNormalizedContainerOptions()` so we don't need to do that here. Also, the normalised config type is different for dockerfiles and image uris so we can use the types to distinguish between them instead of calling `isDockerfile` each time.

packages/containers-shared/src/types.ts
```diff
@@ -11,24 +11,53 @@ export type BuildArgs = {
...
+	  };
 
-/** build/pull agnostic container options */
-export type ContainerDevOptions = {
-	/** may be dockerfile or registry link */
-	image: string;
-	/** formatted as cloudflare-dev/workername-DOclassname:build-id */
-	imageTag: string;
-	/** container's DO class name */
+export type SharedContainerConfig = {
```
emily-shen, line 54: note that most of these fields cannot be undefined - defaults will be set in `getNormalizedContainerOptions()`.

packages/containers-shared/src/types.ts
```diff
@@ -11,24 +11,53 @@ export type BuildArgs = {
...
+	/** absolute path, resolved relative to the wrangler config file. defaults to the directory of the dockerfile */
+	image_build_context: string;
+	image_vars?: Record<string, string>;
+};
+export type RegistryLinkConfig = SharedContainerConfig & {
+	registry_link: string;
+};
+
+export type ContainerNormalisedConfig = RegistryLinkConfig | DockerfileConfig;
+export type InstanceTypeOrLimits =
```
emily-shen, line 38: i've assumed these are mutually exclusive?

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -80,12 +80,6 @@ function mockModifyApplication(
 }
 
 describe("cloudchamber apply", () => {
-	/* eslint no-irregular-whitespace: ["error", { "skipTemplates": true }]
```
emily-shen, line 83: the stdout mock now normalises the weird whitespaces into regular ones

packages/wrangler/src/__tests__/cloudchamber/deploy.test.ts
emily-shen, line 1: moved into containers/deploy.test.ts

packages/wrangler/src/__tests__/cloudchamber/build.test.ts
```diff
@@ -280,32 +267,10 @@ describe("buildAndMaybePush", () => {
...
 				account: accountBase,
-				containerApp: undefined,
+				configDiskSize: undefined,
 			});
 
 			expect(result).toEqual(undefined);
 		});
 	});
-
-	describe("resolveAppDiskSize", () => {
```
emily-shen, line 320: this function doesn't exist anymore but it was a very thin wrapper around parseByteSize which is well tested

packages/wrangler/src/__tests__/containers/apply.test.ts
emily-shen, line 1: moved into containers/deploy.test.ts. also instead of just calling `apply`, we call `wrangler deploy` which means we test the full config flow

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -219,21 +219,30 @@ export class LocalRuntimeController extends RuntimeController {
...
 				this.#currentContainerBuildId !== data.config.dev.containerBuildId
 			) {
+				this.#dockerPath = data.config.dev?.dockerPath ?? getDockerPath();
+				assert(
+					data.config.dev.containerBuildId,
+					"Build ID should be set if containers are enabled and defined"
+				);
+				for (const container of data.config.containers) {
+					this.#containerImageTagsSeen.add(
+						getDevContainerImageName(
```
emily-shen: so that we didn't end up with another config type, i got rid of the ContainerDevOptions type and just construct the tag name from the class_name and containerBuildId where necessary or pass the tag around. i think that's better than having yet more slightly different config types?

(also really think i should have used container.name here in the tag - not sure if it is too late to change it 🤔 )

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -272,7 +272,6 @@ export async function apply(
 	args: {
 		skipDefaults: boolean | undefined;
 		env?: string;
-		imageUpdateRequired?: boolean;
```
emily-shen, line 274: in the cloudchamber apply command, this is always required, so now that cloudchamber apply and containers deploy are separate, we don't need this anymore

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -347,16 +346,6 @@ export async function apply(
 					`${config.name}-${appConfigNoDefaults.class_name}`
 			];
 
-		// while configuration.image is deprecated to the user, we still resolve to this for now.
```
emily-shen, line 344: config.validation now ensures both containers.image and containers.configuraiton.image are set - containerAppToCreateFunction already deletes containers.image so it doesn't show up in the diff

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -216,33 +212,31 @@ export async function buildAndMaybePush(
...
 	if (existsSync(args.PATH) && !isDir(args.PATH)) {
 		throw new UserError(
 			`${args.PATH} is not a directory. Please specify a valid directory path.`
 		);
 	}
-	// if containers are not defined, the build should still work.
-	const containers = config.containers ?? [undefined];
+
 	const pathToDockerfile = join(args.PATH, "Dockerfile");
-	for (const container of containers) {
```
emily-shen, line 238: This would have resulted in multiple container builds/pushes if you had multiple containers in your config, even though the command only accepts values from args and not config. The only value we get from config here is disk size (hidden config value), and i think it probably wasn't the correct logic to redeploy one container multiple times based on the various disk limits of all your containers.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -272,23 +266,22 @@ export async function pushCommand(
 export async function ensureDiskLimits(options: {
 	requiredSize: number;
 	account: CompleteAccountCustomer;
-	containerApp: ContainerApp | undefined;
+	configDiskSize: number | undefined;
```
emily-shen: ensureDiskLimits called in both the containers build command and wrangler deploy - as noted above, the diskSize is not read from config in the build command. In deploy, we do check the wrangler config for disk limits per container, and `getNormalizedContainerOptions()` will have already normalised it to a number, so we don't need to do that here.

packages/wrangler/src/cloudchamber/common.ts
```diff
@@ -700,9 +686,9 @@ function inferInstanceType(
 // diffs.
 export function cleanForInstanceType(
```
emily-shen, line 691: TODO: note to self - revert these changes to minimise changes to cloudchamber apply.

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -9,50 +8,44 @@ import { logger } from "../logger";
...
+		"disk_size" in containerConfig ? containerConfig.disk_size : undefined
 	);
 
-	return { image: buildResult.image, imageUpdated: buildResult.pushed };
+	if (buildResult.pushed) {
+		return { newImageLink: buildResult.image };
+	}
+	// if the image has not changed, it will not have been pushed
+	// so we don't need to update anything when we apply the container config
+	return { newImageLink: undefined };
```
emily-shen, line 48: combined this into one value as i think it is simpler to reason about - hopefully it is still correct

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -65,27 +58,25 @@ export type DeployContainersArgs = {
...
-	{ versionId, accountId, scriptName, dryRun }: DeployContainersArgs
+	normalisedContainerConfig: ContainerNormalisedConfig[],
+	{ versionId, accountId, scriptName }: DeployContainersArgs
 ) {
-	if (config.containers === undefined) {
-		return;
-	}
+	await fillOpenAPIConfiguration(config, containersScope);
 
-	if (!dryRun) {
```
emily-shen, line 74: dry run bails before we get here

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -65,27 +58,25 @@ export type DeployContainersArgs = {
...
-		await fillOpenAPIConfiguration(config, containersScope);
-	}
 	const pathToDocker = getDockerPath();
-	for (const container of config.containers) {
-		const version = await fetchVersion(
-			config,
-			accountId,
-			scriptName,
-			versionId
+	const version = await fetchVersion(config, accountId, scriptName, versionId);
```
emily-shen, line 67: there can only be one script so i think we don't need to do this per container

packages/wrangler/src/containers/config.ts
```diff
@@ -0,0 +1,112 @@
...
+
+		if (!targetDurableObject) {
+			throw new UserError(
+				`The container class_name ${container.class_name} does not match any durable object class_name defined in your Wrangler config file. Note that the durable object must be defined in the same script as the container.`
+			);
+		}
+		const shared: Omit<SharedContainerConfig, "disk_size" | "instance_type"> = {
+			name: container.name,
+			class_name: container.class_name,
+			max_instances: container.max_instances ?? 0, // :(
```
emily-shen: should this be the default? should we print a warning?

packages/wrangler/src/containers/deploy.ts
```diff
@@ -172,66 +172,43 @@ function observabilityToConfiguration(
 	}
 }
 
-function containerAppToInstanceType(
```
emily-shen, line 174: this is already done in `getNormalisedContainerConfig`

packages/wrangler/src/containers/deploy.ts
```diff
@@ -268,328 +242,245 @@ export async function apply(
...
+	const prevApp = existingApplications.find(
+		(app) => app.name === containerConfig.name
+	);
 
 	// TODO: JSON formatting is a bit bad due to the trimming.
 	// Try to do a conditional on `configFormat`
 
 	log(dim("Container application changes\n"));
 
-	for (const appConfigNoDefaults of config.containers) {
```
emily-shen, line 298: wrangler deploy already loops over the containers to build them and calls apply immediately after.

packages/wrangler/src/containers/deploy.ts
```diff
@@ -268,328 +242,245 @@ export async function apply(
...
 		{ message: "Loading applications" }
 	);
-	applications.forEach((app) =>
+	existingApplications.forEach((app) =>
 		cleanupObservability(app.configuration.observability)
 	);
-	const applicationByNames: Record<ApplicationName, Application> = {};
 	// TODO: this is not correct right now as there can be multiple applications
 	// with the same name.
-	for (const application of applications) {
```
emily-shen, line 280: deployContainers (which calls apply) already loops over config, so this would have caused us to loop n*containers times D:

packages/wrangler/src/containers/deploy.ts
```diff
@@ -268,328 +242,245 @@ export async function apply(
...
-			// lines
-			const prevApp = sortObjectRecursive<CreateApplicationRequest>(
-				stripUndefined(applicationToCreateApplication(accountId, application))
+		if (!normalisedPrevApp.durable_objects?.namespace_id) {
+			throw new FatalError(
+				"The previous deploy of this container application was not associated with a durable object"
 			);
+		}
 
-			// fill up fields that their defaults were changed over-time,
```
emily-shen, line 332: already done in getNormalisedContainersConfig

packages/wrangler/src/containers/deploy.ts
```diff
@@ -268,328 +242,245 @@ export async function apply(
...
 
-				for (let i = index; i < printedLines.length; i++) {
-					log(printedLines[i]);
-					if (printedLines.length - i > 2) {
-						i = printedLines.length - 2;
-						printLine(dim("..."), "  ");
-					}
-				}
+		const prev = JSON.stringify({ containers: [normalisedPrevApp] }, null, 2);
+		const now = JSON.stringify({ containers: [nowContainer] }, null, 2);
```
emily-shen: this was previously calling formatConfigSnippet, but the config shape here is different from the config type, which imo is more confusing if we make it look like like it is a wrangler config snippet. But i'm happy to revert if there are strong opinions

packages/wrangler/src/containers/deploy.ts
```diff
@@ -268,328 +242,245 @@ export async function apply(
...
 
 		// print the header of the app
 		updateStatus(bold.underline(green.underline("NEW")) + ` ${appConfig.name}`);
 
-		const s = formatConfigSnippet(
-			{
-				containers: [
-					{
-						...appConfig,
-						instances:
```
emily-shen, line 417: is this just to hide instances in the diff? (because it is showing up anyway in the EDIT diff along with all sorts of stuff).

packages/wrangler/src/config/environment.ts
```diff
@@ -103,14 +103,27 @@ export type ContainerApp = {
...
 		secrets?: { name: string; type: "env"; secret: string }[];
-		disk?: { size: string };
+		disk?:
+			| {
+					/**
+					 * @deprecated Use `size_mb` instead.
+					 */
+					size: string;
+			  }
+			| { size_mb: number };
```
nikitassharma:
```suggestion
		disk?: { size_mb: number };
```

packages/wrangler/src/config/environment.ts
```diff
@@ -103,14 +103,27 @@ export type ContainerApp = {
...
+		vcpu?: number;
+		memory_mib?: number;
+		/**
+		 * @deprecated Use `size_mb` instead.
+		 */
+		memory?: string;
```
nikitassharma:
```suggestion
```

packages/wrangler/src/containers/config.ts
```diff
@@ -0,0 +1,112 @@
...
+			}
+
+			instanceTypeOrDisk = {
+				disk_size: normalisedDiskSize,
+				vcpu: container.configuration?.vcpu,
+				memory_mib:
+					container.configuration.memory !== undefined
+						? Math.round(parseByteSize(container.configuration?.memory))
+						: container.configuration?.memory_mib,
+			};
```
nikitassharma:
```suggestion
		if (
			container.configuration?.disk?.size_mb !== undefined ||
			container.configuration?.vcpu !== undefined ||
			container.configuration?.memory_mib !== undefined
		) {
			instanceTypeOrDisk = {
				disk_size: container.configuration?.disk?.size_mb,
				vcpu: container.configuration?.vcpu,
				memory_mib: container.configuration?.memory_mib,
			};
```
we should enforce the strongly typed fields here, I don't think there's a need to support the string fields. afaik, they're not being used for custom instance types

packages/containers-shared/src/types.ts
```diff
@@ -11,24 +11,53 @@ export type BuildArgs = {
...
+	| {
+			/** if undefined in config, defaults to instance_type */
+			disk_size?: number;
```
nikitassharma:
```suggestion
			disk_mb?: number;
```
just so this is slightly more consistent with the naming elsewhere

packages/wrangler/src/__tests__/cloudchamber/build.test.ts
```diff
@@ -255,13 +248,7 @@ describe("buildAndMaybePush", () => {
...
 					requiredSize: 333 * MiB, // 333MiB
 					account: accountBase,
-					containerApp: {
-						...defaultConfiguration,
-						configuration: {
-							image: "",
-							disk: { size: "3GB" }, // This exceeds the account limit of 2GB
-						},
-					},
+					configDiskSize: 3000 * MiB, // This exceeds the account limit of 2GB
```
nikitassharma: should this be MB?

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -272,23 +266,22 @@ export async function pushCommand(
...
-			`Exceeded account limits: Your container is configured to use a disk size of ${appDiskSize / MB} MB. However, that exceeds the account limit of ${accountDiskSize / MB}`
+			`Exceeded account limits: Your container is configured to use a disk size of ${options.configDiskSize / MB} MB. However, that exceeds the account limit of ${accountDiskSize / MB}`
 		);
 	}
-	const maxAllowedImageSizeBytes = appDiskSize ?? accountDiskSize;
+	const maxAllowedImageSizeBytes = options.configDiskSize ?? accountDiskSize;
 
 	logger.debug(
-		`Disk size limits when building the container: appDiskSize:${appDiskSize}, accountDiskSize:${accountDiskSize}, maxAllowedImageSizeBytes=${maxAllowedImageSizeBytes}(${maxAllowedImageSizeBytes / MB} MB), requiredSized=${options.requiredSize}(${Math.ceil(options.requiredSize / MiB)}MiB)`
+		`Disk size limits when building the container: appDiskSize:${options.configDiskSize}, accountDiskSize:${accountDiskSize}, maxAllowedImageSizeBytes=${maxAllowedImageSizeBytes}(${maxAllowedImageSizeBytes / MB} MB), requiredSized=${options.requiredSize}(${Math.ceil(options.requiredSize / MiB)}MiB)`
```
nikitassharma:
```suggestion
		`Disk size limits when building the container: appDiskSize:${options.configDiskSize}, accountDiskSize:${accountDiskSize}, maxAllowedImageSizeBytes=${maxAllowedImageSizeBytes}(${maxAllowedImageSizeBytes / MB} MB), requiredSized=${options.requiredSize}(${Math.ceil(options.requiredSize / MB)}MB)`
```

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -272,23 +266,22 @@ export async function pushCommand(
 export async function ensureDiskLimits(options: {
 	requiredSize: number;
 	account: CompleteAccountCustomer;
-	containerApp: ContainerApp | undefined;
+	configDiskSize: number | undefined;
 }): Promise<void> {
 	const MB = 1000 * 1000;
 	const MiB = 1024 * 1024;
```
nikitassharma:
```suggestion
```

packages/wrangler/src/cloudchamber/build.ts
nikitassharma: I might be missing something, but it doesn't seem like there's a reason for the units for this function to be in bytes-- both the user and the API use MB as units for disk, and this seems to be adding unnecessary complexity (same comment for both `ensureDiskLimits` and `resolveAppDiskSize`)

packages/wrangler/src/config/validation.ts
```diff
@@ -2429,34 +2429,38 @@ function validateContainerApp(
...
 				);
 			}
+
 			if ("configuration" in containerAppOptional) {
 				diagnostics.warnings.push(
 					`"containers.configuration" is deprecated. Use top level "containers" fields instead. "configuration.image" should be "image", "configuration.disk" should be set via "instance_type".`
 				);
-			}
+				if (
+					typeof containerAppOptional !== "object" ||
```
```suggestion
					typeof containerAppOptional.configuration !== "object" ||
```

packages/wrangler/src/containers/config.ts
```diff
@@ -0,0 +1,102 @@
...
+		if (
+			container.configuration?.disk !== undefined ||
+			container.configuration?.vcpu !== undefined ||
+			container.configuration?.memory_mib !== undefined
+		) {
+			instanceTypeOrDisk = {
+				disk_mb: container.configuration.disk?.size_mb,
+				vcpu: container.configuration?.vcpu,
+				memory_mib: container.configuration?.memory_mib,
+			};
```
nikitassharma:
```suggestion
			instanceTypeOrDisk = {
				disk_mb: container.configuration.disk?.size_mb ?? 2000,
				vcpu: container.configuration?.vcpu ?? 0.0625,
				memory_mib: container.configuration?.memory_mib ?? 256,
			};
```

emily-shen: so to double check, a user can set just vcpu (for example) as a custom value, and then the other limits will default to the values you've specified there?

nikitassharma: yup, that's the expected behavior. anything not specified will fall back on the corresponding value for a dev instance type.

packages/wrangler/src/__tests__/cloudchamber/build.test.ts
emily-shen: probably yes for consistency, although I don't _think_ it matters as long as its bigger than 2GB? anyway updated :)

packages/wrangler/src/cloudchamber/build.ts
emily-shen: ah i think the problem was originally this was parsing a string into bytes, but now we're expecting a number value with MB as units. also, the required size uses MiB (argh) so i guess that's why we need it in bytes? i will try and fix this logic...

nikitassharma: 😭 yeah i think this function would definitely benefit from being refactored, since the logic seems unaffected with the current changes, it might be worth it to leave a ticket so that the refactoring work can be done in a separate pr to avoid increasing the scope of this one?

packages/wrangler/src/containers/deploy.ts
gpanders: https://github.com/cloudflare/workers-sdk/pull/10009 will conflict with this.

I think I _slightly_ lean toward keeping this as it was (using `formatConfigSnippet`). I think as a user if I'm using a `wrangler.toml` file and then when I run `wrangler deploy` I see some arbitrary JSON, I might find that confusing. If it's TOML at least I can infer that this is showing a difference in my config file.

packages/wrangler/src/cloudchamber/build.ts
emily-shen: fixed the incorrect units here [bcb07d8](https://github.com/cloudflare/workers-sdk/pull/9961/commits/bcb07d85419fbd6521e24b9b08cf5ab614c6c225) but leaving further cleanup to a followup (will add ticket here: <TODO>)

packages/containers-shared/src/client/core/request.ts
```diff
@@ -263,6 +263,10 @@ export const sendRequest = async (
 	};
 
 	if (config.WITH_CREDENTIALS) {
+		// :(
+		// The vite-plugin is attempting to typecheck everything with worker types, which does not support request.credentials
+		// Also note this is always set to "omit".
+		// @ts-ignore
```
dario-piotrowicz: I am not sure I follow, isn't this code run inside workerd? why is it using the node types here? (in other words I think that there's something off with the types and that the vite-plugin is likely doing the right thing here, is that not the case?)

packages/containers-shared/src/types.ts
```diff
@@ -20,15 +22,50 @@ export type BuildArgs = {
...
 
-/** build/pull agnostic container options */
-export type ContainerDevOptions = {
-	/** may be dockerfile or registry link */
-	image: string;
-	/** formatted as cloudflare-dev/workername-DOclassname:build-id */
-	imageTag: string;
+export type ContainerNormalisedConfig = RegistryLinkConfig | DockerfileConfig;
+export type DockerfileConfig = SharedContainerConfig & {
+	/**  absolute path, resolved relative to the wrangler config file */
```
dario-piotrowicz:
```suggestion
	/** absolute path, resolved relative to the wrangler config file */
```

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -399,10 +399,10 @@ async function resolveConfig(
 	// for pulling containers, we need to make sure the OpenAPI config for the
 	// container API client is properly set so that we can get the correct permissions
 	// from the cloudchamber API to pull from the repository.
-	const needsPulling = resolved.containers?.some(
-		(c) => !isDockerfile(c.image ?? c.configuration?.image, config.configPath)
+	const needsPulling = resolved.containers.some(
+		(c) => "registry_link" in c && c.registry_link
 	);
-	if (needsPulling && !resolved.dev.remote) {
+	if (resolved.containers.length && needsPulling && !resolved.dev.remote) {
```
dario-piotrowicz: why this addition?

if `needsPulling` is `true` then there must at least be a container in `resolved.containers` no?

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -166,12 +162,13 @@ export async function buildAndMaybePush(
...
 				// the managed registry.
 				const [image, hash] = digest.split("@");
-				const resolvedImage = await resolveImageName(
+				const resolvedImage = resolveImageName(
 					account.external_account_id,
 					image
 				);
 				const remoteDigest = `${resolvedImage}@${hash}`;
 
+				// if errors, means it doesn't exist remotely
```
dario-piotrowicz: could you explain what this means? 😅 🙏

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -270,29 +265,30 @@ export async function pushCommand(
...
 		(options.account.limits.disk_mb_per_deployment ?? 2000) * MB;
 	// if appDiskSize is defined and configured to be more than the accountDiskSize, error
-	if (appDiskSize && appDiskSize > accountDiskSize) {
+	if (
+		options.configDiskInBytes &&
+		options.configDiskInBytes > accountDiskSize
+	) {
 		throw new UserError(
-			`Exceeded account limits: Your container is configured to use a disk size of ${appDiskSize / MB} MB. However, that exceeds the account limit of ${accountDiskSize / MB}`
+			`Exceeded account limits: Your container is configured to use a disk size of ${Math.ceil(options.configDiskInBytes / MB)} MB. However, that exceeds the account limit of ${accountDiskSize / MB}MB`
```
dario-piotrowicz: nit: in the first size there's a space between the number and MB, in there second there's not, I think we should be consistent in this?

basically:

```suggestion
			`Exceeded account limits: Your container is configured to use a disk size of ${Math.ceil(options.configDiskInBytes / MB)}MB. However, that exceeds the account limit of ${accountDiskSize / MB}MB`
```

or

```suggestion
			`Exceeded account limits: Your container is configured to use a disk size of ${Math.ceil(options.configDiskInBytes / MB)} MB. However, that exceeds the account limit of ${accountDiskSize / MB} MB`
```

dario-piotrowicz: In the `UserError` below I can see spaces, so maybe that's the convention to follow?

packages/wrangler/src/containers/deploy.ts
```diff
@@ -2,6 +2,7 @@
  * Note! Much of this is copied and modified from cloudchamber/apply.ts
  * However this code is only used for containers interactions, not cloudchamber ones!
  */
+import assert from "assert";
```
dario-piotrowicz:
```suggestion
import assert from "node:assert";
```

packages/wrangler/src/containers/deploy.ts
```diff
@@ -79,14 +77,21 @@ function createApplicationToModifyApplication(
 ): ModifyApplicationRequestBody {
 	return {
 		configuration: req.configuration,
-		instances: req.max_instances !== undefined ? 0 : req.instances,
+		// deprecated in favour of max_instances
+		instances: 0,
```
dario-piotrowicz: `instances` seems to be optional, could it just be removed here (and in the other occurrences of this in this file)?

packages/wrangler/src/containers/deploy.ts
```diff
@@ -79,14 +77,21 @@ function createApplicationToModifyApplication(
...
 		max_instances: req.max_instances,
 		constraints: req.constraints,
 		affinities: req.affinities,
 		scheduling_policy: req.scheduling_policy,
 	};
 }
 
+/**
+ *
+ * // QUESTION: why create applicaiton request? this is only used when modifying
```
dario-piotrowicz: does this question need to be answered before the PR is merged?

packages/containers-shared/src/build.ts
```diff
@@ -27,12 +24,11 @@ export async function constructBuildCommand(
...
 	}
-	const baseDir = configPath ? path.dirname(configPath) : process.cwd();
-	const absDockerfilePath = path.resolve(baseDir, options.pathToDockerfile);
-	const dockerfile = readFileSync(absDockerfilePath, "utf-8");
+
+	const dockerfile = readFileSync(options.pathToDockerfile, "utf-8");
 	// pipe in the dockerfile
 	buildCmd.push("-f", "-");
-	buildCmd.push(options.buildContext ?? path.dirname(absDockerfilePath));
+	buildCmd.push(options.buildContext);
```
penalosa, line 36: What does this change do?

packages/containers-shared/src/client/core/request.ts
penalosa: Why is this being type-checked in a workers-types context? I don't think we should be ignoring that, if we can

packages/containers-shared/src/client/core/request.ts
```diff
@@ -233,7 +233,7 @@ const parseResponseSchemaV4 = <T>(
 			result = {};
 		}
 	} else {
-		result = { error: fetchResult.errors?.[0].message };
+		result = { error: fetchResult.errors?.[0]?.message };
```
penalosa: Should this have a fallback error message?

packages/containers-shared/src/client/core/request.ts
```diff
@@ -113,8 +113,8 @@ const getFormData = (options: ApiRequestOptions): FormData | undefined => {
 	if (options.formData) {
 		const formData = new FormData();
 
-		const process = (key: string, value: any) => {
-			if (isString(value) || isBlob(value)) {
+		const process = async (key: string, value: any) => {
```
penalosa: Does this need to be async?

packages/wrangler/src/containers/deploy.ts
```diff
@@ -268,328 +253,237 @@ export async function apply(
...
-		const appConfig = containerAppToCreateApplication(
-			accountId,
-			appConfigNoDefaults,
-			config.observability,
-			application,
-			args.skipDefaults
-		);
+	const accountId = config.account_id || (await getAccountId(config));
+	const imageRef = args.newImageLink ?? prevApp?.configuration.image;
+	assert(imageRef, "No changes detected but no previous image found");
```
dario-piotrowicz, line 234: I am not sure I understand this message 😓

packages/containers-shared/src/types.ts
```diff
@@ -20,15 +22,50 @@ export type BuildArgs = {
...
+	dockerfile: string;
+	/** absolute path, resolved relative to the wrangler config file. defaults to the directory of the dockerfile */
+	image_build_context: string;
+	image_vars?: Record<string, string>;
+};
+export type RegistryLinkConfig = SharedContainerConfig & {
+	registry_link: string;
+};
```
penalosa: Is "registry link" a common term? I think intuitively I'd think the two types were "dockerfile" and "image url"

packages/containers-shared/src/types.ts
```diff
@@ -20,15 +22,50 @@ export type BuildArgs = {
 	setNetworkToHost?: boolean;
 };
 
-/** build/pull agnostic container options */
-export type ContainerDevOptions = {
-	/** may be dockerfile or registry link */
-	image: string;
-	/** formatted as cloudflare-dev/workername-DOclassname:build-id */
-	imageTag: string;
```
penalosa, line 28: Why is `imageTag` no longer included?

packages/wrangler/src/containers/deploy.ts
```diff
@@ -268,328 +253,237 @@ export async function apply(
...
-							continue;
-						}
-
-						printedLines.push(createLine(l, "  "));
-					}
-				}
-			}
+		const rawResults = diffLines(prev, now);
+		const results = filterTrailingCommaChanges(rawResults);
+		const changes = results.find((l) => l.added || l.removed) !== undefined;
```
dario-piotrowicz:
```suggestion
		const changes = results.some((l) => l.added || l.removed);
```

packages/containers-shared/src/types.ts
penalosa, line 1: Why the change from `camelCase` to `snake_case` for options?

packages/containers-shared/src/images.ts
```diff
@@ -1,33 +1,34 @@
...
 	runDockerCmd,
 	verifyDockerInstalled,
 } from "./utils";
+import type { ContainerNormalisedConfig, RegistryLinkConfig } from "./types";
 
 export async function pullImage(
 	dockerPath: string,
-	options: ContainerDevOptions
+	options: RegistryLinkConfig,
+	tag: string
```
penalosa: `imageTag`, maybe?

packages/wrangler/src/containers/deploy.ts
emily-shen: when i was using this initially, i was more confused by something that _mostly_ looked like my config, but also with some fields i didn't expect like `constraints`, or `configuration.image` when i had `image` in my config. in fact the main reason i changed it here was because formatConfigSnippet expects the `Config` type, whereas this is actually the `CreateApplicationRequestType`. Anyway, I don't really mind - i can change this back to printing toml if you have wrangler.toml :)

Totally out of the scope for this PR, but just FYI - regular wrangler deploy getting diffs too soon, and at that point this will probably become slightly redundant. However, we should definitely still keep this around in the debug output - normally that includes all API calls, but the containers api calls are missing from that currently, and the diff is more readable too :)

packages/containers-shared/src/utils.ts
```diff
@@ -55,7 +56,7 @@ export const runDockerCmd = (
 			}
 		},
 		ready,
-		then: async (resolve, reject) => ready.then(resolve).catch(reject),
+		then: async (onResolve, onReject) => ready.then(onResolve).catch(onReject),
```
penalosa, line 60: This is a bit of a weird API—is this to make this a custom thenable?

packages/wrangler/src/containers/config.ts
```diff
@@ -0,0 +1,105 @@
...
+	ContainerNormalisedConfig,
+	InstanceTypeOrLimits,
+	SharedContainerConfig,
+} from "@cloudflare/containers-shared";
+
+// This normalises config into an intermediate shape for building or pulling.
+// We set defaults here too, because we can assume that if the value is undefined,
+// we want to revert to the default rather than inheriting from the prev deployment
+
+// todo add args resolution
```
dario-piotrowicz: Could you clarify this todo? 🙂🙏

packages/wrangler/src/containers/config.ts
```diff
@@ -0,0 +1,105 @@
...
+export const getNormalizedContainerOptions = async (
+	config: Config
+): Promise<ContainerNormalisedConfig[]> => {
+	if (!config.containers || config.containers.length === 0) {
+		return [];
+	}
+
+	const normalizedContainers: ContainerNormalisedConfig[] = [];
```
dario-piotrowicz: nit: I think we should be consistent with the naming, here we have both _normali**z**ed_ 🇺🇸  and _normali**s**ed_ 🇬🇧

packages/wrangler/src/containers/config.ts
```diff
@@ -0,0 +1,105 @@
...
+
+	for (const container of config.containers) {
+		assert(container.name, "container name should have been set by validation");
+		const targetDurableObject = config.durable_objects.bindings.find(
+			(durableObject) =>
+				durableObject.class_name === container.class_name &&
+				// the durable object must be defined in the same script as the container
```
dario-piotrowicz: should/is this validated somewhere? 🙂

packages/wrangler/src/containers/deploy.ts
penalosa: Does this need resolved before merging?

packages/containers-shared/src/build.ts
emily-shen: buildContext is now taking `image_build_context` from the resolved config, which will already have a default value set if not defined in user config -  i've changed the corresponding type not be required, not optional.  it will also already be converted to an absolute path in the config normalisation function

packages/wrangler/src/containers/deploy.ts
gpanders: > regular wrangler deploy getting diffs too soon

oh interesting I didn’t realize that. Is that going to reuse the diff code that Cloudchamber is already using or is that being implemented from scratch?

I would honestly be totally happy removing the cloudchamber-specific diff code. The less stuff like that we have to maintain ourselves the better as far as I’m concerned.

emily-shen: i don't really have much context on this - @penalosa will be working on it i think, but we can definitely remove the cloudchamber diffing code once its added.

packages/wrangler/src/cloudchamber/build.ts
emily-shen: added more details to the comment

packages/wrangler/src/containers/deploy.ts
emily-shen: instances is not actually optional in `CreateApplicationRequest`, so we set it to 0. we can probably remove it from `ModifyApplicationRequest` though

packages/wrangler/src/containers/deploy.ts
emily-shen: ah i forgot about that. so createApplicationRequest is a superset of the modify version, so we convert config -> createApplicationRequest -> modifyApplicationRequestBody for the modify path. added comments to explain

packages/wrangler/src/containers/deploy.ts
emily-shen: Ah this should be unreachable, i've added a comment to elaborate

packages/wrangler/src/containers/deploy.ts
emily-shen: This bit of code has been refactored out of this file in a different PR

packages/containers-shared/src/types.ts
emily-shen: i've changed it to `image_uri`

packages/wrangler/src/containers/config.ts
emily-shen: as much as it hurts, changed to `normalized` 🥲

packages/wrangler/src/containers/deploy.ts
emily-shen: https://github.com/cloudflare/workers-sdk/pull/9961#discussion_r2219544227

packages/containers-shared/src/utils.ts
emily-shen: yep - this pr doesn't actually add any functionality, this is to appease the linter

packages/containers-shared/src/client/core/request.ts
emily-shen: landed in this PR first, and comment here for context: https://github.com/cloudflare/workers-sdk/pull/9819#discussion_r2218681550

emily-shen: landed in this PR instead, comment for context: https://github.com/cloudflare/workers-sdk/pull/9819#discussion_r2218681550

packages/containers-shared/src/types.ts
emily-shen: it's back [75817eb](https://github.com/cloudflare/workers-sdk/pull/9961/commits/75817eba75224c027dcf7efa3e8a109db885eb89)

packages/containers-shared/src/types.ts
```diff
@@ -20,15 +22,60 @@ export type BuildArgs = {
...
+	/** if undefined in config, defaults to 25 */
+	rollout_step_percentage: number;
+	/** if undefined in config, defaults to "full_auto" */
+	rollout_kind: "full_auto" | "full_manual" | "none";
+	constraints: {
+		regions?: string[];
+		cities?: string[];
+		tier: number;
+	};
+	observability: { logsEnabled: boolean };
```
dario-piotrowicz: how come `logsEnabled` is camelCase while all the other configs are snake_case?

packages/containers-shared/src/utils.ts
dario-piotrowicz: @penalosa For context I've added the thenable logic a few days ago: https://github.com/cloudflare/workers-sdk/blob/bfe70964a70d93e27546b53fc95e6f1b8d12ce0d/packages/containers-shared/src/utils.ts#L17-L18

It's simply because most `runDockerCmd` calls would just `await` the operation so I thought it'd be neat not to force all callers to extract the ready promise (`await runDockerCmd();` feels much clearer/natural), if you find it a confusing API I am totally happy reverting that change (in a followup PR of course)

packages/wrangler/src/containers/deploy.ts
```diff
@@ -527,69 +462,63 @@ export async function apply(
...
+	}
+};
 
-			success(`Modified application ${brandColor(action.name)}`, {
-				shape: shapes.bar,
-			});
+/**
+ * clean up application object received from API so that we get a nicer diff when comparing it to the current config.
+ */
+export function cleanPrevious(
```
dario-piotrowicz:
```suggestion
function cleanPrevious(
```

Also what about remaining it to something like: `cleanApplicationFromAPI` or something like that? 🤔 I think that `cleanPrevious` sounds a bit too vague

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -232,24 +231,23 @@ export async function buildCommand(
...
+		{
+			tag: args.tag,
+			pathToDockerfile,
+			buildContext: args.PATH,
+			platform: args.platform,
+			// no option to add env vars at build time...?
+		},
+		getDockerPath() ?? args.pathToDocker,
+		args.push,
+		// this means we wont be able to read the disk size from the config, but that option is deprecated anyway at least for containers.
```
dario-piotrowicz:
```suggestion
		// this means we won't be able to read the disk size from the config, but that option is deprecated anyway at least for containers.
```
🙈

packages/wrangler/src/containers/config.ts
```diff
@@ -0,0 +1,131 @@
...
+import { UserError } from "../errors";
+import { getAccountId } from "../user";
+import type { Config } from "../config";
+import type {
+	ContainerNormalizedConfig,
+	InstanceTypeOrLimits,
+	SharedContainerConfig,
+} from "@cloudflare/containers-shared";
+
+// TODO: allow merging in args from commands with config. e.g. for `wrangler containers build`
```
dario-piotrowicz: Is this for now or for later?

In either case could you add a bit more context? I am not sure what this is referring to, are there CLI arguments that people can pass to wrangler that should be merged into the configs?

packages/wrangler/src/containers/deploy.ts
```diff
@@ -123,7 +98,7 @@ function cleanupObservability(
 }
 
 function observabilityToConfiguration(
-	observability: Observability | undefined,
+	observabilityFromConfig: boolean,
```
dario-piotrowicz, line 105: Is this parameter name correct? I feel like it should be `observabilityLogsEnabled`? 🤔

packages/wrangler/src/__tests__/containers/config.test.ts
```diff
@@ -0,0 +1,455 @@
+import { mkdirSync, writeFileSync } from "fs";
+import path from "path";
```
dario-piotrowicz:
```suggestion
import { mkdirSync, writeFileSync } from "node:fs";
import path from "node:path";
```

packages/wrangler/src/__tests__/containers/config.test.ts
```diff
@@ -0,0 +1,455 @@
...
+						name: "TEST_DO",
+						class_name: "TestContainer",
+					},
+				],
+			},
+		} as Partial<Config> as Config;
+
+		const result = await getNormalizedContainerOptions(config);
+
+		expect(result[0]).toMatchObject({
```
dario-piotrowicz, line 367:
```suggestion
        expect(result).toHaveLength(1);
		expect(result[0]).toMatchObject({
```

packages/wrangler/src/__tests__/containers/config.test.ts
```diff
@@ -0,0 +1,455 @@
...
+						name: "TEST_DO",
+						class_name: "TestContainer",
+					},
+				],
+			},
+		} as Partial<Config> as Config;
+
+		const result = await getNormalizedContainerOptions(config);
+
+		// Check that it has dockerfile properties (not image_uri)
```
dario-piotrowicz, line 450:
```suggestion
         expect(result).toHaveLength(1);

		// Check that it has dockerfile properties (not image_uri)
```

packages/containers-shared/src/utils.ts
penalosa: Yeah, this feels pretty weird to me, but if it's existing logic then 👍

packages/wrangler/src/containers/deploy.ts
```diff
@@ -171,92 +143,60 @@ function observabilityToConfiguration(
...
+						memory_mib: containerApp.memory_mib,
+						vcpu: containerApp.vcpu,
+					}),
+			observability: observabilityToConfiguration(
+				containerApp.observability.logsEnabled,
+				prevApp?.configuration.observability
 			),
 		},
+		// deprecated in favour of max_instances
+		instances: 0,
```
penalosa, line 183: Why include it if it's deprecated?

packages/wrangler/src/containers/deploy.ts
```diff
@@ -171,92 +143,60 @@ function observabilityToConfiguration(
...
-	return app;
 }
 
 export async function apply(
 	args: {
-		skipDefaults: boolean | undefined;
 		imageUpdateRequired?: boolean;
+		/**
+		 * If the image was built and pushed, or is a registry link, we have to update the image ref and this will be defined
+		 * If it is undefined, the image has not change, and we do not need to update the image ref
```
penalosa:
```suggestion
		 * If it is undefined, the image has not changed, and we do not need to update the image ref
```

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -1761,7 +1641,7 @@ describe("cloudchamber apply", () => {
 				account_id: "1",
 				scheduling_policy: SchedulingPolicy.REGIONAL,
 				configuration: {
-					image: "./Dockerfile",
+					image: "docker.io/beep:boop",
```
penalosa, line 1644: Is there a reason we seem to be getting rid of all tests that reference a dockerfile rather than an image?

packages/wrangler/src/__tests__/containers/config.test.ts
```diff
@@ -0,0 +1,455 @@
...
+		writeFileSync("Dockerfile", "FROM scratch");
+
+		const config: Config = {
+			name: "test-worker",
+			configPath: "./wrangler.toml",
+			topLevelName: "test-worker",
+			containers: [
+				{
+					class_name: "TestContainer",
+					image: path.resolve("./Dockerfile"),
```
penalosa, line 122: Does this need `path.resolve()`? What happens if a user just puts `./Dockerfile`?

penalosa (approved):
This is looking pretty good. At a high level though, I'm a bit confused by the purpose of the normalize function:
1. Why is it separate to all our other config normalization in `normalizeAndValidateEnvironment()`? Does it need to be it's own function that's called individually?
2. Why do we need to have an intermediate type with conversion back and forth from config types to API types?

That said, I'm not going to block on that—this is a massive PR that's likely to get stale quickly

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
emily-shen: This shouldn't have ever used a dockerfile, as running cloudchamber apply has no build step - the tests were previously 'incorrect'.

packages/wrangler/src/containers/deploy.ts
emily-shen: The API type expects to be set to 0 if max_instances is present according to @gpanders and the existing code. We can consider cleaning this up and making changes to the containers API, but this preserves existing behaviour at least.

gpanders: Correct, it's a required field as far as the API is concerned, but we don't expose it as a config option in Wrangler (only `max_instances` is exposed), so we have Wrangler just set it to 0.

The "correct" long-term action is probably to make this field non-required in our API.

packages/wrangler/src/containers/config.ts
emily-shen: agh this is probably something we should think about more fully later. for example, currently wrangler containers build requires you to pass everything in via CLI. it probably should read from wrangler config too, especially since not all config options are present as flags. but also you might have multiple containers. anyway, a bigger problem than this one random comment, so removed and will create a ticket.

packages/wrangler/src/containers/deploy.ts
emily-shen: the idea is that one is the current value from config and the other is the value from the prev deployment. added comments to clarify

packages/wrangler/src/__tests__/containers/config.test.ts
emily-shen: this should have been converted to an absolute path at this stage by the main validation function. currently the vite plugin doesn't call the normalisation function (mostly because it doesn't need all the deploy time config), but it does call readConfig. if we normalise the paths there, we don't have to repeat it in wrangler, vite dev and vite build when we write the redirected config. i'm not pleased with it conceptually, but it does simplify things.

emily-shen:
> This is looking pretty good. At a high level though, I'm a bit confused by the purpose of the normalize function:
> 
> 1. Why is it separate to all our other config normalization in `normalizeAndValidateEnvironment()`? Does it need to be it's own function that's called individually?
> 2. Why do we need to have an intermediate type with conversion back and forth from config types to API types?
> 
> That said, I'm not going to block on that—this is a massive PR that's likely to get stale quickly

1. there isn't a really good reason (plenty of now invalid ones!) - i would definitely like to reconsider this, but in a followup... i'd like to get this out asap, for my sanity and because its blocking some other PRs.
2. because the intermediate build/push step requires a bunch of things that the API does not need. i don't think we convert back and forth, just config -> normalised config (for build and push) -> API?

PR #9962, opened 2025-07-14
CC-5692: Add support for custom instance types

nikitassharma:
Custom instance types could previously only be set through setting vcpu/memory/disk through the `configuration` block. As we deprecate `configuration`, move setting vcpu, memory, and disk to the instance type field of wrangler. By nature, this enforces that custom instance types are mutually exclusive with the named instance types and will allow us to remove a lot of the logic that verifies this once `configuration` is removed.

Fixes CC-5692

---

- Tests
  - [X] Tests included
- Public documentation
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23674
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/hungry-turtles-beam.md
+++ b/.changeset/hungry-turtles-beam.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add support for custom instance types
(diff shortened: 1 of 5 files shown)


nikitassharma:
https://github.com/cloudflare/workers-sdk/pull/10051

PR #9969, opened 2025-07-15
Further updates to e2e test running helpers and instructions

petebacondarwin:


packages/wrangler/e2e/README.md
```diff
@@ -4,15 +4,37 @@ This folder contains e2e tests for Wrangler.
...
-Run each of the test files as a separately cached turbo task.
+Run each of the e2e test files as a separately cached turbo task. This is what we run in the CI.
+
+This is the most effective way to run the Wrangler e2e tests. If any of the tests flake, the ones that passed will be cached and will be skipped on re-runs.
 
 ```zsh
 pnpm test:e2e:wrangler
 ```
 
+You can also run the turbo task directly if you want more fine grain control although this is not as resilient to flakes and is less effective at caching:
```
penalosa:
```suggestion
You can also run the turbo task directly if you want more fine grained control although this is not as resilient to flakes and is less effective at caching:
```

petebacondarwin: Strictly I think my wording is still correct but since you suggested it, I'll accept.

PR #9971, opened 2025-07-15
feat(wrangler): improve script source display on pretty error screen

edmundhung:
Fixes [DEVX-1974](https://jira.cfdata.org/browse/DEVX-1974) and [DEVX-1975](https://jira.cfdata.org/browse/DEVX-1975).

Example screenshots:

- Error from native module (Requrie "View All Frames" checked):
 
  Screenshot 2025-07-21 at 13 24 52

- Error from missing source file (e.g. `cookie`)

  Screenshot 2025-07-16 at 12 55 41

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: error screen is not documented
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not bugfix


packages/wrangler/templates/startDevWorker/ProxyWorker.ts
```diff
@@ -129,7 +129,6 @@ export class ProxyWorker implements DurableObject {
 				request.url
 			);
 			headers.set("MF-Original-URL", innerUrl.href);
-			headers.set("MF-Disable-Pretty-Error", "true"); // disables the UserWorker miniflare instance from rendering the pretty error -- instead the ProxyWorker miniflare instance will intercept the json error response and render the pretty error page
```
edmundhung, line 132: We are now relying on the UserWorker miniflare instance to prettify the error so that we can retrieve the script source from its inspector.

packages/miniflare/src/workers/core/entry.worker.ts
```diff
@@ -124,7 +125,9 @@ function getUserRequest(
 
 	request.headers.delete(CoreHeaders.PROXY_SHARED_SECRET);
 	request.headers.delete(CoreHeaders.ORIGINAL_URL);
-	request.headers.delete(CoreHeaders.DISABLE_PRETTY_ERROR);
+	if (env[CoreBindings.STRIP_DISABLE_PRETTY_ERROR]) {
+		request.headers.delete(CoreHeaders.DISABLE_PRETTY_ERROR);
+	}
```
edmundhung, line 130: We will continue stripping this header by default, except on the ProxyWorker miniflare which will pass the header over to the UserWorker miniflare so that user can disable the pretty error screen when needed.

patches/youch@4.1.0-beta.10.patch
```diff
@@ -0,0 +1,66 @@
+diff --git a/build/chunk-4XB2BYKC.js b/build/chunk-4XB2BYKC.js
+index 6e6e490115bc4b107b53834ceac3d7d19cdbe3fe..37373d734d034caf45e28c1b165e5c176f901137 100644
+--- a/build/chunk-4XB2BYKC.js
++++ b/build/chunk-4XB2BYKC.js
+@@ -34,10 +34,10 @@ var ErrorStackSource = class extends BaseComponent {
+    */
+   async toHTML(props) {
+     const frame = props.frame;
+-    if (frame.type === "native" || !frame.source || !frame.fileName) {
++    if (!frame.source || !frame.fileName) {
```
edmundhung: Youch currently skip displaying the script source if it is an internal module. But we do want it to show up if `frame.source` is defined. I will put up a PR to youch  later.

patches/youch@4.1.0-beta.10.patch
```diff
@@ -0,0 +1,66 @@
...
+       return "";
+     }
+-    const language = LANGS_MAP[extname(frame.fileName)] ?? "plain";
++    const language = LANGS_MAP[extname(frame.fileName)] ?? (frame.type === "native" ? "js" : "plain");
```
edmundhung: Native modules have no extension so there is no syntax highlight by default. But it should be safe to assume native modules to be javascript if available?

patches/youch@4.1.0-beta.10.patch
```diff
@@ -0,0 +1,66 @@
...
++import { existsSync } from "node:fs"
+ import { dump, themes } from "@poppinss/dumper/html";
+ import { dump as dumpCli } from "@poppinss/dumper/console";
+ var CHEVIRON = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" width="24" height="24" stroke-width="2">
+@@ -50,7 +51,7 @@ var ErrorStack = class extends BaseComponent {
+    */
+   #getEditorLink(ide, frame) {
+     const editorURL = EDITORS[ide] || ide;
+-    if (!editorURL || frame.type === "native") {
++    if (!editorURL || frame.type === "native" || !existsSync(frame.fileName)) {
```
edmundhung: Not sure if this is a good patch upstream as it assumes youch is running on node environment only. But I will raise it up when I submit a PR to youch.

packages/miniflare/src/plugins/core/errors/index.ts
```diff
@@ -247,19 +273,95 @@ export function reviveError(
...
+	// Find the first available worker
+	const worker = Array.isArray(workers) ? workers[0] : null;
+
+	if (!worker || !worker.webSocketDebuggerUrl) {
+		return;
+	}
+
+	// Connect to the specific worker's WebSocket
+	const promise = new DeferredPromise<string | undefined>();
+	const ws = new NodeWebSocket(worker.webSocketDebuggerUrl);
```
penalosa: Does this connecting to a Workers websocket URL cause issues with multiple clients/already active debugger sessions?

packages/miniflare/src/plugins/core/errors/index.ts
```diff
@@ -281,10 +383,53 @@ export async function handlePrettyErrorRequest(
...
+	youch.defineSourceLoader(async (stackFrame) => {
+		if (!stackFrame.fileName) {
+			return;
+		}
+
+		if (stackFrame.fileName.startsWith("cloudflare:")) {
+			stackFrame.type = "native";
+
+			if (inspectorURL) {
+				const source = await getScriptSource(inspectorURL, stackFrame.fileName);
```
penalosa: Does this start a websocket connection per file in the stack trace? Can we bundle them all up together somehow?

packages/miniflare/src/plugins/core/errors/index.ts
```diff
@@ -281,10 +383,53 @@ export async function handlePrettyErrorRequest(
...
+		const workerIndex = maybeGetStringScriptPathIndex(stackFrame.fileName);
+		if (workerIndex !== undefined) {
+			const srcOpts = workerSrcOpts[workerIndex];
+			if ("script" in srcOpts && srcOpts.script !== undefined) {
+				return { contents: srcOpts.script };
+			}
+		}
+
+		// Check if the source map has the source content for this file so we don't need to read it from disk
+		const sorucesContent = sourcesContentMap.get(stackFrame.fileName);
```
penalosa:
```suggestion
		const sourcesContent = sourcesContentMap.get(stackFrame.fileName);
```

patches/youch@4.1.0-beta.10.patch
penalosa: Can we make this optional so it can run in non-node environments? We run Youch ourselves in a non-node environment (format-errors), so it would be nice to make that possible without patches (that we apply to format-errors)

packages/miniflare/src/plugins/core/errors/index.ts
edmundhung: It does :( It seems like there could only be one connection max and the moment we connect it for the script source, the other debugger is disconnected.

Maybe we can extend the inspector proxy for this purpose but it seems better wait until we migrate wrangler over to the miniflare implementation first...

I will remove this logic for now.

edmundhung: Dropped native module support in [ae2bbd6](https://github.com/cloudflare/workers-sdk/pull/9971/commits/ae2bbd610fb6cd372f8ca539d33d9a8f96be8975)

patches/youch@4.1.0-beta.10.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/build/chunk-JAN2TFI2.js b/build/chunk-JAN2TFI2.js
+index a6321161988abe906e7745bf9d540670dee5b289..28c79363b321cd99809c44fdfeebcf86b05cada6 100644
+--- a/build/chunk-JAN2TFI2.js
++++ b/build/chunk-JAN2TFI2.js
+@@ -50,7 +50,7 @@ var ErrorStack = class extends BaseComponent {
+    */
+   #getEditorLink(ide, frame) {
+     const editorURL = EDITORS[ide] || ide;
+-    if (!editorURL || frame.type === "native") {
++    if (!editorURL || frame.type === "native" || frame.fileType !== 'fs') {
```
edmundhung, line 10: We are now relying on the fileType to decide whether Youch should generate a link.

packages/miniflare/src/plugins/core/errors/index.ts
edmundhung: I have also updated the screenshot in the description to reflect the latest implementation

PR #9976, opened 2025-07-15
add warning for when users run `wrangler dev --remote` with (enabled) containers

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1969

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23702
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a v3 change

### DIFF
--- a/packages/wrangler/src/api/startDevWorker/ConfigController.ts
+++ b/packages/wrangler/src/api/startDevWorker/ConfigController.ts
@@ -418,11 +418,12 @@ async function resolveConfig(
 
 	if (resolved.dev.remote) {
 		// We're in remote mode (`--remote`)
-		const containerClassNames = new Set(
-			resolved.containers?.map((c) => c.class_name)
-		);
 
-		if (containerClassNames.size > 0 && resolved.dev.enableContainers) {
+		if (
+			resolved.dev.enableContainers &&
+			resolved.containers &&
+			resolved.containers.length > 0
+		) {
 			logger.warn(
 				"Containers are only supported in local mode, to suppress this warning set `dev.enable_containers` to `false` or pass `--enable-containers=false` to the `wrangler dev` command"
 			);
@@ -433,16 +434,8 @@ async function resolveConfig(
 			resolved.migrations
 		);
 		if (
-			Array.from(classNamesWhichUseSQLite.entries()).some(
-				([className, usesSQLite]) => {
-					if (containerClassNames.has(className)) {
-						// We want to ignore class names of containers since
-						// we have a separate warning dedicated for those
-						return false;
-					}
-					return usesSQLite;
-				}
-			)
+			resolved.dev.remote &&
+			Array.from(classNamesWhichUseSQLite.values()).some((v) => v)
 		) {
 			logger.warn("SQLite in Durable Objects is only supported in local mode.");
 		}


Mahmoodyazdanpanah:
Dear [Recipient's Name],

I am writing to express my sincerest gratitude for your invaluable help and support. Your assistance with [briefly mention the specific task or situation] was instrumental in [mention the positive outcome or benefit].

I particularly appreciate your willingness to [mention a specific action or characteristic, e.g., "take the time to explain the process," "offer insightful suggestions," or "go the extra mile"]. Your expertise and dedication were truly commendable, and I learned a great deal from your guidance.

The project's success would not have been possible without you. Thank you once again for your collaborative spirit, patience, and commitment. I am deeply grateful for your kindness.

Best regards,

[Your Name]


در ۱۵ ژوئیهٔ ۲۰۲۵ ۲۳:۳۵:۱۰ (GMT+03:30)، "changeset-bot[bot]" ***@***.***> نوشت:

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -416,15 +416,36 @@ async function resolveConfig(
...
-		logger.warn("SQLite in Durable Objects is only supported in local mode.");
+	if (resolved.dev.remote) {
+		// We're in remote mode (`--remote`)
+		const containerClassNames = new Set(
+			resolved.containers?.map((c) => c.class_name)
+		);
+
+		if (containerClassNames.size > 0 && resolved.dev.enableContainers) {
+			logger.warn(
+				"Containers are only supported in local mode, to suppress this warning set `dev.enable_containers` to `false` or pass `--enable-containers=false` to the `wrangler dev` command"
```
dario-piotrowicz, line 428: Quick note: the warning mentioned the flag to pass to `wrangler dev` but not that this code can also be run via `runWorker` by any node program, so mentioning wrangler directly is a bit off...

Alternatively I could:
 - only mention `dev.enable_containers` (which applies for both `wrangler dev` and `startWorker`),
 - or conditionally check if we're running inside `wrangler dev` and only in that case add the flag information (the problem with this solution is that we currently do not have a way to discern whether we're running in `wrangler dev` so I'd have to implement a solution for that)
 
Also note that there are other parts of the `devEnv` code that have the same issue, for example:
https://github.com/cloudflare/workers-sdk/blob/14ce5775c775b32bc1166d4e7a1546a00c049ab0/packages/wrangler/src/api/startDevWorker/ConfigController.ts#L416

So instead of overthinking this, after a quick chat with @penalosa, I decided to just include both config and flag in the warning message. If anyone disagrees I am happy to change this 🙂

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -416,15 +416,36 @@ async function resolveConfig(
...
+
+		// TODO(do) support remote wrangler dev
+		const classNamesWhichUseSQLite = getClassNamesWhichUseSQLite(
+			resolved.migrations
+		);
+		if (
+			Array.from(classNamesWhichUseSQLite.entries()).some(
+				([className, usesSQLite]) => {
+					if (containerClassNames.has(className)) {
+						// We want to ignore class names of containers since
```
emily-shen: i don't think that's true - someone might expect to be able to use the sqlite storage operations on a DO that has a container, and be confused why that isn't working when they were only warned about the container part not working? i mean admittedly its an edge case,  but i think providing more info is better.

packages/wrangler/src/config/config.ts
```diff
@@ -238,7 +238,6 @@ export interface DevConfig {
 	 * When developing, whether to build and connect to containers. This requires a Docker daemon to be running.
 	 * Defaults to `true`.
 	 *
-	 * @inheritable
```
emily-shen, line 241: thanks 😅

packages/wrangler/src/api/startDevWorker/ConfigController.ts
dario-piotrowicz: oh yeah sorry I think you're right

my bad 🙇 

I'll amend this, thanks for spotting my mistake 🫶

dario-piotrowicz: fixed, thanks again 🙏

PR #9978, opened 2025-07-15
Bump `tar-fs` where possible

penalosa:
Bump `tar-fs` to fix audit warnings. This also involved removing the `cloudflare-workers-extension code`, since `vsce` depended on `tar-fs`. I removed it rather than updating `vsce` because it was already on the roadmap to remove this code.

Note: this _doesn't_ bump `tar-fs` in `fixtures/browser-rendering` because it's in a pinned dependency chain from `@cloudflare/puppeteer`.

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal dependency bump
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


petebacondarwin:
So to clarify, we don't use this directly, which is why you just updated the package-lock.json?

petebacondarwin (approved):

penalosa:
> So to clarify, we don't use this directly, which is why you just updated the package-lock.json?

Yes, we have no direct dependency (only indirect through prebuild-install)

PR #9982, opened 2025-07-16
avoid a major bump to vite-plugin when its peer dependency (wrangler) gets a minor bump

petebacondarwin:
The `workspace:*` peer dependency of vite-plugin on wrangler meant that every time wrangler had a minor bump, the vite-plugin would get a major bump.

This PR fixes that by moving from `workspace:*` (which corresponds to the exact current version of the dependency) to `workspace:^` which means that it will work with newer versions (up to the next major bump).

It required a small patch to the changesets tool to support this, which we could upstream via PR to their project.
Looks like there is already a PR: https://github.com/changesets/changesets/pull/1291

You can test this locally by checking out this PR and then running:

```
GITHUB_TOKEN=ghp_xxxx node .github/changeset-version.js
```

replacing the GITHUB_TOKEN value with one you have created.

> It is possible we could achieve the same result by setting the peer dependency to `workspace:^4` without the patch to the tool but I think this is cleaner.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: CI job
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: CI job
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  vite-plugin is not backported

### DIFF
--- a/packages/vite-plugin-cloudflare/package.json
+++ b/packages/vite-plugin-cloudflare/package.json
@@ -69,7 +69,7 @@
 	},
 	"peerDependencies": {
 		"vite": "^6.1.0 || ^7.0.0",
-		"wrangler": "workspace:*"
+		"wrangler": "workspace:^"
 	},
 	"publishConfig": {
 		"access": "public"
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -10,6 +10,9 @@
 	"baseBranch": "main",
 	"updateInternalDependencies": "patch",
 	"bumpVersionsWithWorkspaceProtocolOnly": true,
+	"___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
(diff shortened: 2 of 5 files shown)


packages/vite-plugin-cloudflare/package.json
```diff
@@ -69,7 +69,7 @@
 	},
 	"peerDependencies": {
 		"vite": "^6.1.0 || ^7.0.0",
-		"wrangler": "workspace:*"
+		"wrangler": "workspace:^"
```
penalosa, line 72: Just so I'm clear in my head, this will change the published package.json from:

```
  "wrangler": "4.1.0"
```
to
```
  "wrangler": "^4.1.0"
```
?

penalosa (approved):

PR #9984, opened 2025-07-16
Allow unenv-preset tests a few retries

petebacondarwin:
Fix a DNS error:

```
@cloudflare/unenv-preset:test:ci:    × @cloudflare/unenv-preset darwin /Users/pbacondarwin/dev/cloudflare/workers-sdk/node_modules/.pnpm/@cloudflare+workerd-darwin-arm64@1.20250612.0/node_modules/@cloudflare/workerd-darwin-arm64/bin/workerd > testDns 5045ms
@cloudflare/unenv-preset:test:ci:      → Test timed out in 5000ms.
```

### DIFF
--- a/packages/unenv-preset/tests/index.test.ts
+++ b/packages/unenv-preset/tests/index.test.ts
@@ -2,7 +2,7 @@ import path from "node:path";
 import { platform } from "node:process";
 import { fileURLToPath } from "node:url";
 import { fetch } from "undici";
-import { afterAll, beforeAll, describe, expect, test } from "vitest";
+import { afterAll, beforeAll, describe, expect, test, vi } from "vitest";
 import workerdPath from "workerd";
 import { runWranglerDev } from "../../../fixtures/shared/src/run-wrangler-long-lived";
 import { TESTS } from "./worker/index";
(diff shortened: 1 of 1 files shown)


packages/unenv-preset/tests/index.test.ts
```diff
@@ -41,9 +41,15 @@ describe(`@cloudflare/unenv-preset ${platform} ${workerdPath}`, () => {
 
 	test.for(Object.keys(TESTS))("%s", async (testName) => {
 		expect(wrangler).toBeDefined();
-		const { ip, port } = wrangler!;
-		const response = await fetch(`http://${ip}:${port}/${testName}`);
-		const body = await response.text();
-		expect(body).toMatch("OK!");
+		const { ip, port, getOutput } = wrangler!;
+		try {
```
vicb, line 45: How is that allowing retries?
Do failed tests automatically retry?

packages/unenv-preset/tests/index.test.ts
```diff
@@ -41,9 +41,15 @@ describe(`@cloudflare/unenv-preset ${platform} ${workerdPath}`, () => {
...
+			await vi.waitFor(async () => {
+				const response = await fetch(`http://${ip}:${port}/${testName}`);
+				const body = await response.text();
+				expect(body).toMatch("OK!");
+			});
+		} catch {
+			console.log("OUTPUT", await getOutput());
```
vicb, line 52: Should this be a catch to only log on exceptions?
(and probably should rethrow)

vicb (commented):
How are the tests failing?

packages/unenv-preset/tests/index.test.ts
petebacondarwin: yes I updated it when you weren't looking :D

packages/unenv-preset/tests/index.test.ts
petebacondarwin: The `vi.waitFor()` below will try to make the fetch repeatedly until the `expect()` passes or a timeout happens.

petebacondarwin:
> How are the tests failing?

I can no longer find the failing job. But basically the fetch was returning the wrong response.

And then it passed on a retry of the whole job.

Issue #9945, opened 2025-07-13
browser.close() fails when with browser rendering


threepointone:
### What versions & operating system are you using?

wrangler@4.24.3, @cloudflare/vite-plugin@1.9.4

### Please provide a link to a minimal reproduction

https://github.com/threepointone/browser-close-hangs

### Describe the Bug

Like it says, `browser.close()` hangs in dev, in both vite and wrangler. It works fine in production, but this is a blocker for dev. 

```ts
   const browser = await launch(env.browser);
    const page = await browser.newPage();
    await page.goto("https://www.example.com");
    const content = await page.content();
    // this line hangs
    await browser.close();
    return new Response(content, {
      headers: {
        "Content-Type": "text/html",
      },
    });
```

ruifigueira:
Debugging it, it gets stuck waiting for a `disconnect` event which is never triggered:

https://github.com/cloudflare/playwright/blob/81e76bca0b5657530974033d8c66d8aeb8ec72a7/packages/playwright-core/src/server/browser.ts#L181

I'm investigating why that event is not triggered in local dev (with `wrangler dev --remote` it's properly triggered).

ruifigueira:
In remote dev, websocket `close` event is properly triggered, and in local dev it's not. It may be related with this issue I reported to the workerd team:

https://github.com/cloudflare/workerd/issues/4327

threepointone:
thanks for checking this out! it's blocking our investigations into using it for agents stuff

PR #9989, opened 2025-07-16
fix: polls miniflare to check browser process

ruifigueira:
Playwright relies on a WebSocket close event that is never triggered in local dev. As a workaround, we poll miniflare to check if the browser is still up and running, otherwise we close the websocket explictly.

Fixes: #9945

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s):
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  New feature

### DIFF
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -900,6 +900,7 @@ export class Miniflare {
 	#workerOpts: PluginWorkerOptions[];
 	#log: Log;
 
+	// key is the browser wsEndpoint, value is the browser process
 	#browserProcesses: Map<string, ChildProcess> = new Map();
 
 	readonly #runtime?: Runtime;
@@ -1310,12 +1311,10 @@ export class Miniflare {
 				response = new Response(wsEndpoint);
 			} else if (url.pathname === "/browser/status") {
 				const wsEndpoint = url.searchParams.get("wsEndpoint");
-				assert(wsEndpoint !== null);
+				assert(wsEndpoint !== null, "Missing wsEndpoint query parameter");
 				const process = this.#browserProcesses.get(wsEndpoint);
-				assert(process);
-
 				const status = {
-					stopped: process.exitCode !== null,
+					stopped: !process || process.exitCode !== null,
 				};
 				response = new Response(JSON.stringify(status), {
 					headers: { "Content-Type": "application/json" },
@@ -1970,9 +1969,10 @@ export class Miniflare {
 	}
 
 	async #assembleAndUpdateConfig() {
-		for (const process of this.#browserProcesses.values()) {
+		for (const [wsEndpoint, process] of this.#browserProcesses.entries()) {
 			// .close() isn't enough
 			process.kill("SIGKILL");
+			this.#browserProcesses.delete(wsEndpoint);
 		}
 		// This function must be run with `#runtimeMutex` held
 		const initial = !this.#runtimeEntryURL;
@@ -2674,9 +2674,10 @@ export class Miniflare {
 		try {
 			await this.#waitForReady(/* disposing */ true);
 		} finally {
-			for (const process of this.#browserProcesses.values()) {
+			for (const [wsEndpoint, process] of this.#browserProcesses.entries()) {
 				// .close() isn't enough
 				process.kill("SIGKILL");
+				this.#browserProcesses.delete(wsEndpoint);
 			}
 
 			// Remove exit hook, we're cleaning up what they would've cleaned up now
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -1301,7 +1301,10 @@ export class Miniflare {
 				);
 
 				// @ts-expect-error Puppeteer is dynamically installed, and so doesn't have types available
-				const browser = await puppeteer.launch({ headless: "old" });
+				const browser = await puppeteer.launch({
+					headless: "old",
+					args: process.env.CI ? ["--no-sandbox"] : [],
+				});
 				const wsEndpoint = browser.wsEndpoint();
 				this.#browserProcesses.set(wsEndpoint, browser.process());
 				response = new Response(wsEndpoint);
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -1304,6 +1304,7 @@ export class Miniflare {
 				// @ts-expect-error Puppeteer is dynamically installed, and so doesn't have types available
 				const browser = await puppeteer.launch({
 					headless: "old",
+					// workaround for CI environments, to avoid sandboxing issues
 					args: process.env.CI ? ["--no-sandbox"] : [],
 				});
 				const wsEndpoint = browser.wsEndpoint();
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -901,8 +901,8 @@ export class Miniflare {
 	#log: Log;
 
 	#browsers: Set<{
-		wsEndpoint: () => string;
-		process: () => ChildProcess;
+		wsEndpoint: string;
+		process: ChildProcess;
 	}> = new Set();
 
 	readonly #runtime?: Runtime;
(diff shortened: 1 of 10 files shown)


packages/miniflare/src/workers/browser-rendering/binding.worker.ts
```diff
@@ -36,34 +48,56 @@ export class BrowserSession extends DurableObject {
 		});
 
 		server.addEventListener("message", (m) => {
-			if (m.data === "ping") return;
+			if (m.data === "ping") {
```
penalosa, line 69: Where does this ping come from?

packages/miniflare/src/index.ts
```diff
@@ -1305,9 +1305,27 @@ export class Miniflare {
...
+					wsEndpoint,
+					process: browser.process(),
+				});
+				response = new Response(wsEndpoint);
+			} else if (url.pathname === "/browser/status") {
+				const wsEndpoint = url.searchParams.get("wsEndpoint");
+				const { process } =
+					[...this.#browsers].find(
+						(browser) => browser.wsEndpoint === wsEndpoint
+					) ?? {};
```
penalosa: Given this is being used for lookup now, I think it should probably be changed to a `Map`?

packages/miniflare/src/index.ts
```diff
@@ -1305,9 +1305,27 @@ export class Miniflare {
...
+				if (process) {
```
penalosa: What should the behaviour be if `process` isn't defined? Should this be an `assert(process)`?

packages/miniflare/src/index.ts
```diff
@@ -1305,9 +1305,27 @@ export class Miniflare {
...
+					const status = {
+						wsEndpoint,
+						stopped: process.exitCode !== null,
```
penalosa: I'm not sure I understand what this is doing? The only way a browser can be killed is if Miniflare restarts, which will reset the user code context anyway. Is there a way from user-code to stop a browser early? What's the API for that?

packages/miniflare/src/workers/browser-rendering/binding.worker.ts
```diff
@@ -36,34 +48,56 @@ export class BrowserSession extends DurableObject {
...
+				this.#checkStatus().catch((err) => {
+					console.error("Error checking browser status:", err);
+				});
```
penalosa, line 72: It doesn't looks like you're doing anything with the result of `checkStatus`?

packages/miniflare/src/workers/browser-rendering/binding.worker.ts
ruifigueira: Both [puppeteer](https://github.com/cloudflare/puppeteer/blob/main/packages/puppeteer-core/src/cloudflare/WorkersWebSocketTransport.ts#L35) and [playwright](https://github.com/cloudflare/playwright/blob/main/packages/playwright-cloudflare/src/cloudflare/webSocketTransport.ts#L28) ping the browser.

This was not my initial approach, because those ping calls rely on those libraries. I tried with DurableObject alarm but it was crashing the worker runtime, so I changed to this approach as a workaround.

packages/miniflare/src/index.ts
ruifigueira: There's a [Browser.close](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-close) that kills the browser. That's how playwright and puppeteer close the browser.

Besides, I think there's a requirement to allow a headful browser in local dev, and the user can just close the browser window.

In the particular case of playwright, it waits for the websocket to be closed, but we never get that event and that's why it was getting stuck, so instead of waiting for the event, we just poll the browser process to check if it was killed already and we explicitly close the websocket ourselves.

packages/miniflare/src/workers/browser-rendering/binding.worker.ts
ruifigueira: It checks browser process in miniflare and if was killed, it explicitly closes the browser webocket (see https://github.com/cloudflare/workers-sdk/pull/9989/files/ce07676496baa728b9baa8f8c253405235afbd45#r2213153307).

BTW, I think the websocket not being closed may be a bug in the worker runtime, because there are other situations where it is not triggered: cloudflare/workerd#4327

packages/miniflare/src/index.ts
penalosa: Right, but how does `browser.close()` fill the browser? I can't see any logic for that here? (or does it just close the websocket?)

ruifigueira: It's chrome that kills itself (actually, it performs a graceful shutdown) when it receives a `Browser.close` CDP instruction.

packages/miniflare/src/index.ts
```diff
@@ -900,10 +900,7 @@ export class Miniflare {
 	#workerOpts: PluginWorkerOptions[];
 	#log: Log;
 
-	#browsers: Set<{
-		wsEndpoint: () => string;
-		process: () => ChildProcess;
-	}> = new Set();
+	#browserProcesses: Map<string, ChildProcess> = new Map();
```
petebacondarwin, line 904: NIT: it would be helpful to have a comment here explaining what the key represents (e.g. the websocket endpoint URL?)

packages/miniflare/src/index.ts
```diff
@@ -1304,10 +1301,25 @@ export class Miniflare {
 				);
 
 				// @ts-expect-error Puppeteer is dynamically installed, and so doesn't have types available
-				const browser = await puppeteer.launch({ headless: "old" });
-				this.#browsers.add(browser);
-
-				response = new Response(browser.wsEndpoint());
+				const browser = await puppeteer.launch({
+					headless: "old",
+					args: process.env.CI ? ["--no-sandbox"] : [],
```
petebacondarwin, line 1308: Out of interest why is this needed?

packages/miniflare/src/index.ts
```diff
@@ -1304,10 +1301,25 @@ export class Miniflare {
...
+				});
+				const wsEndpoint = browser.wsEndpoint();
+				this.#browserProcesses.set(wsEndpoint, browser.process());
+				response = new Response(wsEndpoint);
+			} else if (url.pathname === "/browser/status") {
+				const wsEndpoint = url.searchParams.get("wsEndpoint");
+				assert(wsEndpoint !== null);
```
petebacondarwin: NIT: I know we are not consistent in this file, but I prefer to add a second parameter message to `assert()` calls so that there is some context when they throw. E.g.

```suggestion
				assert(wsEndpoint !== null, "missing wsEndpoint search param");
```

petebacondarwin: etc

packages/miniflare/src/index.ts
```diff
@@ -1958,9 +1970,9 @@ export class Miniflare {
 	}
 
 	async #assembleAndUpdateConfig() {
-		for (const browser of this.#browsers) {
+		for (const process of this.#browserProcesses.values()) {
 			// .close() isn't enough
-			await browser.process().kill("SIGKILL");
+			process.kill("SIGKILL");
 		}
```
petebacondarwin: Should we then clear this `#browsers` map for the next round?
Otherwise, each time we update Miniflare we will leak more and more browser process objects into the map, I think?

packages/miniflare/src/workers/browser-rendering/binding.worker.ts
petebacondarwin: Can you add a comment explaining this for future readers of the code.

packages/miniflare/src/index.ts
ruifigueira: I added it to fix this issue:

```
miniflare:test:ci:   - `TypeError: Fetch API cannot load: error: Failed to launch the browser process![9992:9992:0718/153239.744156:FATAL:zygote_host_impl_linux.cc(126)] No usable sandbox! Update your kernel or see https://chromium.googlesource.com/chromium/src/+/main/docs/linux/suid_sandbox_development.md for more information on developing with the SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.[0718/153239.756306:ERROR:file_io_posix.cc(145)] open /sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq: No such file or directory (2)[0718/153239.756559:ERROR:file_io_posix.cc(145)] open /sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq: No such file or directory (2)TROUBLESHOOTING: https://pptr.dev/troubleshooting    at ChildProcess.onClose (/home/runner/.npm/_npx/9c191fc107e155e1/node_mo
```
They state thar:

```
If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox
```

I don't want to live dangerously, but I needed an immediate workaround :D

I'll also add a comment there.

packages/miniflare/src/index.ts
ruifigueira: You're right. In that case, I need to assume `const process = this.#browserProcesses.get(wsEndpoint)` can be undefined (I was asserting `process` existed, but I'll change it accordingly).

packages/miniflare/src/index.ts
ruifigueira: Apparently, it's a recurring issue:

https://github.com/cloudflare/workers-sdk/blob/d785dc33e49cbe3e23f37614b185957c563973a4/fixtures/workers-with-assets-spa/tests/index.test.ts#L17-L22

petebacondarwin:
@ruifigueira - I added the checkboxes back into the PR description as these are needed to pass CI checks. Please keep them there in future PRs.

packages/miniflare/src/index.ts
petebacondarwin: No comment yet...

petebacondarwin:
Just a changeset needed to pass CI

PR #9993, opened 2025-07-16
Enable sharing concurrent requests for the same module across invocations

### LINKED ISSUE #9518: Vite plugin has race condition with HMR - `A hanging Promise was canceled.`

4.19.1 [wrangler], 4.7.7 [hono], 7.5.2 [react-router] (but have tried many versions of each)

_No response_

### Describe the Bug

I reported a bug in the templates repo, but I was told that this might be the better spot for it. I believe there is a race condition in the vite plugin where if two events fire close together you can get an error below. I have also since found several individuals who also run into the issue, which I think helps point at a problem in the plugin.

```
A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.
```

https://github.com/cloudflare/templates/issues/600

Here's the text from that issue since it's the best description of the issue. I think the key is that I can more or less trigger this in the cloudflare template.

----

This is a bit of a weird one, but I have spent maybe 20+ hours trying to get to the bottom of the source and so I'm pretty sure on most of this stuff: <3

In my app, if two saves happen in fast succession, then it will very commonly (like 1 out of 4 times) trigger the following error.

Image

At first I assumed I had a dangling promise somewhere in my app, but peeled back every file until I had none left and was still able to trigger it. So I attempted to reproduce it directly in the hono-react-router template and after a little bit of wiggling, I was able to do it.

However, in the template app, it's *very difficult* to make happen. I am fairly certain that this is because the app builds so quickly, that it can never get into a state where two HMR reloads cross over each other (at least on my M3 macbook pro). So in order to try to trigger it I changed the template app in 1 way, which is to just load any largeish library. I chose `@adobe/spectrum` since it has a lot of components, but also confirmed I could make it happen with other things like radix/dummy files.

My current best assumption is that the following code is not guaranteed to work when two HMR events occur very close together:

```ts
// app.ts from the hono-react-router template
app.get("*", (c) => {
  const requestHandler = createRequestHandler(
    () => import("virtual:react-router/server-build"), // <-- this line is my best guess
    import.meta.env.MODE,
  );

  return requestHandler(c.req.raw, {
    cloudflare: { env: c.env, ctx: c.executionCtx },
  });
});
```

If I had to do some conjecture, in some cases the app boots up, but before the `createRequestHandler` function invokes that dynamic import, a new server-build exists in that spot.

In order to make the error happen in a small app like the demo template, you may have to press and hold `cmd+s` on the `welcome.tsx` file, or hit it really fast in bursts, however in our real life app, I believe many of our developers run into this issue on every few saves (which kills the entire dev server) because their editors probably save once when they hit save, and then save again after formatting is applied.

I'm sorry I don't have a minimum reproduction linked, but I'm happy to get on a call at any time and show the error in practice and in the template project with my changes.

```
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (ssr) page reload app/welcome/welcome.tsx
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (__react_router_css_dev_helper__) page reload app/welcome/welcome.tsx
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (ssr) page reload app/welcome/welcome.tsx
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (__react_router_css_dev_helper__) page reload app/welcome/welcome.tsx
A hanging Promise was canceled. This happens when the worker runtime is waiting for a Promise from JavaScript to resolve, but has detected that the Promise cannot possibly ever resolve because all code and events related to the Promise's I/O context have already finished.

5:35:00 PM [vite] Internal server error: The script will never generate a response.
      at async ProxyServer.fetch (file:///Users/alex/code/test-template/wispy-pond-2aa3/node_modules/.pnpm/miniflare@4.20250416.0/node_modules/miniflare/src/workers/core/proxy.worker.ts:173:11)
Warning: A promise was resolved or rejected from a different request context than the one it was created in. However, the creating request has already been completed or canceled. Continuations for that request are unlikely to run safely and have been canceled. If this behavior breaks your worker, consider setting the `no_handle_cross_request_promise_resolution` compatibility flag for your worker.
    at /Users/alex/code/test-template/wispy-pond-2aa3/app/root.tsx:15:393

5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (client) hmr update /app/welcome/welcome.tsx, /app/app.css
5:35:00 PM [vite] (ssr) page reload app/welcome/welcome.tsx
```

----

Oh! I should add that I did attempt to simply throttle the vite hmr events like so:

```ts
  server: {
    watch: {
      awaitWriteFinish: {
        stabilityThreshold: 200,
        pollInterval: 100,
      },
    },
  },
```

But unfortunately this didn't resolve the issue in my testing. It just somewhat changes the exact set of things that will trigger it. I do think this may be a temporary solution for folks who have editors that save files twice though, but it seems like there's gotta be some way to make the request more stable through the request into react-router.

----

One more thing that isn't necessarily obvious from the single log output is that there is a line number for the error in the logs:

```
 /Users/alex/code/test-template/wispy-pond-2aa3/app/root.tsx:15:393
```

But this is random as far as I can tell. Each time it'll give me a different file and line number. Often on lines that are comments or could never error. Many times it's in dependencies like react, etc.

My best guess is that the asynchronous event that fires to cause this just chooses whatever line happened to be executing at the time.

_No response_

jamesopstad:
Fixes #9518.

Fix issue that resulted in `A hanging Promise was canceled` errors when developing large applications.
We now handle requests for modules in a Durable Object so that they can be shared across invocations.

Additionally, using `import.meta.hot.send` within the context of a request is now supported.

I have also used this as an opportunity to add extensive JSDoc annotations in the`runner-worker` directory.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
+++ b/packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
@@ -1,5 +1,5 @@
 import { DurableObject } from "cloudflare:workers";
-import { ModuleRunner } from "vite/module-runner";
+import { ModuleRunner, ssrModuleExportsKey } from "vite/module-runner";
 import { INIT_PATH, UNKNOWN_HOST } from "../shared";
 import { stripInternalEnv } from "./env";
 import type { WrapperEnv } from "./env";
@@ -184,7 +184,7 @@ async function createModuleRunner(env: WrapperEnv, webSocket: WebSocket) {
 				try {
 					const fn = env.__VITE_UNSAFE_EVAL__.eval(code, module.id);
 					await fn(...Object.values(context));
-					Object.seal(context.__vite_ssr_exports__);
+					Object.seal(context[ssrModuleExportsKey]);
 				} catch (error) {
 					if (error instanceof Error) {
 						error.message = `Error running module "${module.id}".\n${error.message}.`;


packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -293,15 +365,17 @@ export function createDurableObjectWrapper(
 
 					if (
 						typeof key === "symbol" ||
-						IGNORED_KEYS.includes(key) ||
+						(IGNORED_KEYS as readonly string[]).includes(key) ||
```
petebacondarwin, line 369: Can you, instead declare this var as:

```ts
const IGNORED_KEYS = ["self", "tailStream"] as const;
```

(i.e. with the `as const`)

petebacondarwin: But also, why is this cast needed now when it wasn't before?

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -169,65 +218,67 @@ export function createWorkerEntrypointWrapper(
...
+					workerEntryPath = workerEntryPathHeader;
+					const stub = this.env.__VITE_RUNNER_OBJECT__.get("singleton");
+
+					// Forward the request to the Durable Object to initialize the module runner and return the WebSocket
+					return stub.fetch(request);
 				}
 			}
 
-			const entrypointValue = await getWorkerEntryExport(entryPath, entrypoint);
+			const exportValue = await getWorkerEntryExport(
```
petebacondarwin, line 243: Could we move these functions inside the DO and access them via RPC like:

```suggestion
			const stub = this.env.__VITE_RUNNER_OBJECT__.get("singleton");
			const exportValue = await stub.getWorkerEntryExport(
```

petebacondarwin: and so on to allow us to move the moduleRunner object itself into the DO?

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
jamesopstad: > Can you, instead declare this var as:
> 
> ```ts
> const IGNORED_KEYS = ["self", "tailStream"] as const;
> ```
> 
> (i.e. with the `as const`)

This is how it's declared. It needs the cast because I added `as const`. Adding `as const` isn't strictly necessary but makes it consistent with the other key arrays.

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
jamesopstad: I don't think the return value could be sent over RPC. It's the export from the evaluated module and can be any arbitrary JS.

jamesopstad: Even if this was possible, `moduleRunner.import()` isn't the only way that the module runner is accessed. Modules are also fetched via the module runner when running the user code.

jamesopstad:
I have tested this with a large React Router app that previously produced the issue and can confirm that this fixes it.

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
```diff
@@ -1,46 +1,174 @@
...
+	ModuleEvaluator,
+	ModuleRunnerOptions,
+} from "vite/module-runner";
 
-let moduleRunner: ModuleRunner;
+/**
+ * Custom `ModuleRunner`.
+ * The `cachedModule` method is overriden to ensure compatibility with the Workers runtime.
+ */
+// @ts-expect-error: `cachedModule` is private
```
petebacondarwin, line 16: Is this something we want to upstream to Vite?

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
```diff
@@ -1,46 +1,174 @@
...
+		}
+
+		const { 0: client, 1: server } = new WebSocketPair();
+		server.accept();
+		this.#webSocket = server;
+		moduleRunner = await createModuleRunner(this.env, this.#webSocket);
+
+		return new Response(null, { status: 101, webSocket: client });
+	}
+	/**
```
petebacondarwin, line 85: NIT: do you normally separate methosd with blank lines?

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
```diff
@@ -50,25 +178,22 @@ export async function createModuleRunner(
...
-				const code = `${codeDefinition}${transformed}\n}}`;
+				const code = `"use strict";async (${Object.keys(context).join(",")})=>{${transformed}}`;
+
 				try {
 					const fn = env.__VITE_UNSAFE_EVAL__.eval(code, module.id);
 					await fn(...Object.values(context));
-					Object.freeze(context.__vite_ssr_exports__);
-				} catch (e) {
-					console.error("error running", module.id);
-					console.error(e instanceof Error ? e.stack : e);
```
petebacondarwin, line 63: Is this not useful for debugging?
Wthout this we just swallow the error details and only report the failing module id.

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
```diff
@@ -1,46 +1,174 @@
...
+						event: "vite:ws:connect",
+						data: { webSocket },
+					});
+				},
+				disconnect() {
+					webSocket.close();
+				},
+				send(data) {
+					// We send messages via a binding to the Durable Object.
+					// This is because `import.meta.send` may be called within the context of a Worker.
```
petebacondarwin: ... and if that happens it might be in a different context to the DO, so wouldn't have access to the ES module level `moduleRunner` variable?

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
```diff
@@ -1,46 +1,174 @@
...
+		let cached = this.#concurrentModuleNodePromises.get(url);
 
-	moduleRunner = new ModuleRunner(
+		if (!cached) {
+			const cachedModule = moduleRunner.evaluatedModules.getModuleByUrl(url);
+			cached = moduleRunner
+				// @ts-expect-error: `getModuleInformation` is private
+				.getModuleInformation(url, importer, cachedModule)
+				.finally(() => {
+					this.#concurrentModuleNodePromises.delete(url);
```
petebacondarwin, line 120: So to be clear, this cache is only shortlived, long enough to prevent two requests for the same module from messing up if the first is still processing.
But once they it is processed any future request will trigger new processing for that module?

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -45,10 +62,48 @@ const DURABLE_OBJECT_KEYS = [
 	"webSocketMessage",
 ] as const;
 
+/** Available methods for `WorkflowEntrypoint` classes */
 const WORKFLOW_ENTRYPOINT_KEYS = ["run"] as const;
 
-let entryPath = "";
+/** The Worker entry path */
```
petebacondarwin: Bit of a pointless jsdoc IMO. What kind of path? File path? Request path?
How is this used, why are we storing it in the module?

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -45,10 +62,48 @@ const DURABLE_OBJECT_KEYS = [
...
+let workerEntryPath = "";
+
+/**
+ * Creates a callable thenable that can be both awaited and invoked as a function.
+ * This enables RPC properties to be used both as promises and callable functions.
+ * @param key - The property key name used for error messages
```
petebacondarwin, line 75: So this is actually the name of the property on the RPC object being accessed?

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
```diff
@@ -45,10 +62,48 @@ const DURABLE_OBJECT_KEYS = [
...
 
+/** Available methods for `WorkflowEntrypoint` classes */
 const WORKFLOW_ENTRYPOINT_KEYS = ["run"] as const;
 
-let entryPath = "";
+/** The Worker entry path */
+let workerEntryPath = "";
+
+/**
+ * Creates a callable thenable that can be both awaited and invoked as a function.
```
petebacondarwin:
```suggestion
 * Creates a callable thenable that represents accessing the property of an RPC target. It can be both awaited and invoked as a function.
```

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
jamesopstad: Yes. I think we should first get these changes out to users to see if there are any unforeseen issues. Then we can upstream changes to Vite that enable providing a custom `cachedModule` implementation.

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
jamesopstad: I don't. I can though if you prefer.

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
jamesopstad: Yes, you're right. Is there a neat way to throw the original error with this added message? I'm trying to avoid using `console.error` as that won't make it to the overlay.

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
jamesopstad: No, the `moduleRunner` variable will always be available. The issue is that using a WebSocket within the context of a Worker request handler that was created when handling another request is forbidden (https://developers.cloudflare.com/workers/observability/errors/#cannot-perform-io-on-behalf-of-a-different-request).

I'll update the comment to make this clearer.

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
jamesopstad: Yes, it's only for concurrent requests for the same module.

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
jamesopstad: Yep.

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
petebacondarwin: Perhaps this?

```ts
throw new Error(`Error running module "${module.id}"`, { cause: error });
```

jamesopstad: I tried that but it didn't seem to work (perhaps because of the type of the error). Will see if there's another way.

petebacondarwin: Perhaps just serialize the information from the inner error into the message of the new one?

jamesopstad: [38deb2d](https://github.com/cloudflare/workers-sdk/pull/9993/commits/38deb2da7d8550d628f891258fb4e9b18121bc04)

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
jamesopstad: [38deb2d](https://github.com/cloudflare/workers-sdk/pull/9993/commits/38deb2da7d8550d628f891258fb4e9b18121bc04)

packages/vite-plugin-cloudflare/src/runner-worker/index.ts
jamesopstad: [38deb2d](https://github.com/cloudflare/workers-sdk/pull/9993/commits/38deb2da7d8550d628f891258fb4e9b18121bc04)

packages/vite-plugin-cloudflare/src/runner-worker/module-runner.ts
jamesopstad: [38deb2d](https://github.com/cloudflare/workers-sdk/pull/9993/commits/38deb2da7d8550d628f891258fb4e9b18121bc04)

PR #9994, opened 2025-07-16
Add boolean env var factory

petebacondarwin:
Internal refactor

### DIFF
--- a/packages/wrangler/src/environment-variables/factory.ts
+++ b/packages/wrangler/src/environment-variables/factory.ts
@@ -46,6 +46,27 @@ type DeprecatedNames =
 
 type ElementType<A> = A extends readonly (infer T)[] ? T : never;
 
+export function getBooleanEnvironmentVariableFactory(options: {
+	variableName: VariableNames;
+}): () => boolean | undefined;
+export function getBooleanEnvironmentVariableFactory(options: {
+	variableName: VariableNames;
+	defaultValue: boolean | (() => boolean);
+}): () => boolean;
+export function getBooleanEnvironmentVariableFactory(options: {
+	variableName: VariableNames;
+	defaultValue?: boolean | (() => boolean);
+}): () => boolean | undefined {
+	return () => {
+		if (options.variableName in process.env) {
+			return process.env[options.variableName]?.toLowerCase() === "true";
+		}
+		return typeof options.defaultValue === "function"
+			? options.defaultValue()
+			: options.defaultValue;
+	};
+}
+
 /**
  * Create a function used to access an environment variable. It may return undefined if the variable is not set.
  *
--- a/packages/wrangler/src/environment-variables/factory.ts
+++ b/packages/wrangler/src/environment-variables/factory.ts
@@ -46,6 +46,13 @@ type DeprecatedNames =
 
 type ElementType<A> = A extends readonly (infer T)[] ? T : never;
 
+/**
+ * Create a function used to access a boolean environment variable. It may return undefined if the variable is not set.
+ *
+ * This is not memoized to allow us to change the value at runtime, such as in testing.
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/environment-variables/factory.ts
```diff
@@ -46,6 +46,27 @@ type DeprecatedNames =
 
 type ElementType<A> = A extends readonly (infer T)[] ? T : never;
 
+export function getBooleanEnvironmentVariableFactory(options: {
```
vicb, line 56: Add JSDoc?

Should "1" be considered true too?

packages/wrangler/src/logger.ts
```diff
@@ -109,13 +109,13 @@ export class Logger {
 
 	debug = (...args: unknown[]) => this.doLog("debug", args);
 	debugWithSanitization = (label: string, ...args: unknown[]) => {
-		if (getSanitizeLogs() === "false") {
-			this.doLog("debug", [label, ...args]);
-		} else {
+		if (getSanitizeLogs()) {
```
vicb, line 112: technically that's a breaking change - clearly documenting what value is `true` is important

packages/wrangler/src/metrics/metrics-config.ts
```diff
@@ -66,7 +66,7 @@ export function getMetricsConfig({
 	const sendMetricsEnv = getWranglerSendMetricsFromEnv();
 	if (sendMetricsEnv !== undefined) {
 		return {
-			enabled: sendMetricsEnv.toLowerCase() === "true",
+			enabled: sendMetricsEnv,
```
vicb, line 69: Consider adding a type on `export const getWranglerSendMetricsFromEnv =`?

```suggestion
			enabled: sendMetricsEnv === true,
```

vicb (approved):
few nits

packages/wrangler/src/environment-variables/factory.ts
petebacondarwin: I don't think `1` should be considered true. It should be one of (after lowercaing) `"true"`, `"false"` or `undefined`. I add validation for this and throw an error if anything else is found.

packages/wrangler/src/logger.ts
petebacondarwin: Once I add validation above this should be fine.

packages/wrangler/src/metrics/metrics-config.ts
petebacondarwin: What do you mean?

The type of `getWranglerSendMetricsFromEnv()` is `() => boolean | undefined`; so the type of `sendMetricsEnv` will be `boolean | undefined`.

But by the time we get to this line it has narrowed to `boolean`.

vicb: > The type of `getWranglerSendMetricsFromEnv()` is `() => boolean | undefined`;

That's what I meant:

```ts
export const getWranglerSendMetricsFromEnv =
	getBooleanEnvironmentVariableFactory({
	variableName: "WRANGLER_SEND_METRICS",
});
```

You have to go to level higher to figure the type.

Agreed that narrowing make the `=== true` unneeded

petebacondarwin: If you hover over the `getWranglerSendMetricsFromEnv()` function VS Code will tell you the return type is `boolean | undefined`.

packages/wrangler/src/environment-variables/factory.ts
petebacondarwin: Added JSDOC and validation 5d7438c105bf6114edddb45e09344978eac51c0b

PR #9995, opened 2025-07-16
Some wrangler testing cleanups

petebacondarwin:
Internal refactor


packages/wrangler/src/__tests__/dev.test.ts
```diff
@@ -948,6 +946,10 @@ describe.sequential("wrangler dev", () => {
 		});
 
 		describe(".env", () => {
+			const processEnv = process.env;
```
vicb, line 949: can't vitest mock process.env?

vicb (approved):
LGTM

packages/wrangler/src/__tests__/dev.test.ts
petebacondarwin: I don't know of a way to mock a property in Vitest. Is there?

PR #9996, opened 2025-07-16
Require container images use a supported platform

nikitassharma:
The runtime only runs on amd64 right now. Require that images pushed to the managed registry are configured for a linux/amd64 platform to avoid runtime errors.

Fixes CC-5610
---

- Tests
  - [X] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: better error handling
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/src/cloudchamber/build.ts
+++ b/packages/wrangler/src/cloudchamber/build.ts
@@ -229,7 +229,7 @@ export async function buildCommand(
 	}
 	if (args.platform !== "linux/amd64") {
 		throw new UserError(
-			`Unsupported platform: Platform "${args.platform}" is currently unsupported. Please use "linux/amd64" instead.`
+			`Unsupported platform: Platform "${args.platform}" is unsupported. Please use "linux/amd64" instead.`
 		);
 	}
 	// if containers are not defined, the build should still work.


packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -269,6 +275,23 @@ export async function pushCommand(
 	}
 }
 
+export async function checkImagePlatform(
```
gpanders: Since it's only used in this file (for now) let's keep it private.

```suggestion
async function checkImagePlatform(
```

nikitassharma:
> i am a bit confused why the build command has this optional at all if there is only one option. would it be better to remove or at least hide this flag for now? (if you are planning on supporting other platforms later...?) sorry if i've misunderstood!

I'm not aware of any current plans to support other platforms, my main motivation for keeping the argument instead of removing was to not break anyone that might currently be using it

emily-shen:
> > i am a bit confused why the build command has this optional at all if there is only one option. would it be better to remove or at least hide this flag for now? (if you are planning on supporting other platforms later...?) sorry if i've misunderstood!
> 
> I'm not aware of any current plans to support other platforms, my main motivation for keeping the argument was to not break anyone that might currently be using it

In that case, can you hide the arg to make sure no one uses it going forwards? you can add `hidden: true` (and maybe `deprecated: true` too?) to the yargs options

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -225,6 +227,11 @@ export async function buildCommand(
 			`${args.PATH} is not a directory. Please specify a valid directory path.`
 		);
 	}
+	if (args.platform !== "linux/amd64") {
+		throw new UserError(
+			`Unsupported platform: Platform "${args.platform}" is currently unsupported. Please use "linux/amd64" instead.`
```
emily-shen:
```suggestion
			`Unsupported platform: Platform "${args.platform}" is unsupported. Please use "linux/amd64" instead.`
```

emily-shen (approved):

PR #10001, opened 2025-07-17
Inject HMR code in Worker entry file

### LINKED ISSUE #9518: Vite plugin has race condition with HMR - `A hanging Promise was canceled.`

jamesopstad:
Fixes #000.

This transforms the Worker entry file to inject an HMR boundary. By providing an HMR boundary at the root, we ensure that the full module graph is not invalidated on code changes. Ideally, frameworks should be adding server HMR code at route boundaries but React Router does not currently do this. This follows discussion with the Vite team and they have also merged a PR to update the docs (https://github.com/vitejs/vite/pull/20401).

From feedback gathered from users, it appears this change fixes the issues experienced in https://github.com/cloudflare/workers-sdk/issues/9518. I believe this may be because HMR updates are queued (https://github.com/vitejs/vite/blob/3e81af38a80c7617aba6bf3300d8b4267570f9cf/packages/vite/src/module-runner/hmrHandler.ts#L25) whereas full reloads are not (https://github.com/vitejs/vite/blob/3e81af38a80c7617aba6bf3300d8b4267570f9cf/packages/vite/src/module-runner/hmrHandler.ts#L54). If this turns out not to be the case then we can reopen the issue.

---

- Tests
  - [x] Tests included
  - [ ] Tests not necessary because: not testable
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/orange-donkeys-check.md
+++ b/.changeset/orange-donkeys-check.md
@@ -11,4 +11,3 @@ if (import.meta.hot) {
 ```
 
 This prevents file changes from invalidating the full module graph and improves HMR performance in development.
-It also resolves issues that users were facing with HMR in larger applications.
--- a/.changeset/orange-donkeys-check.md
+++ b/.changeset/orange-donkeys-check.md
@@ -1,0 +1,14 @@
+---
+"@cloudflare/vite-plugin": minor
+---
+
+We now automatically inject the following HMR code into your Worker entry file:
+
+```ts
+if (import.meta.hot) {
+	import.meta.hot.accept();
+}
+```
+
+This prevents file changes from invalidating the full module graph and improves HMR performance in development.
+It also resolves issues that users were facing with HMR in larger applications.
--- a/packages/vite-plugin-cloudflare/playground/worker/__tests__/worker.spec.ts
+++ b/packages/vite-plugin-cloudflare/playground/worker/__tests__/worker.spec.ts
@@ -3,6 +3,7 @@ import * as path from "node:path";
 import { expect, test, vi } from "vitest";
 import {
 	getTextResponse,
+	isBuild,
 	rootDir,
 	serverLogs,
 	viteTestUrl,
@@ -28,12 +29,15 @@ test("does not cause unhandled rejection", async () => {
 	expect(serverLogs.errors.join()).not.toContain("__unhandled rejection__");
 });
 
-test("updates using HMR code in Worker entry file", async () => {
-	const workerEntryPath = path.join(rootDir, "src", "index.ts");
-	const originalContent = fs.readFileSync(workerEntryPath, "utf-8");
-	fs.writeFileSync(workerEntryPath, originalContent);
+test.runIf(!isBuild)(
+	"updates using HMR code in Worker entry file",
+	async () => {
+		const workerEntryPath = path.join(rootDir, "src", "index.ts");
+		const originalContent = fs.readFileSync(workerEntryPath, "utf-8");
+		fs.writeFileSync(workerEntryPath, originalContent);
 
-	await vi.waitFor(() => {
-		expect(serverLogs.info.join()).toContain("[vite] hot updated");
-	});
-});
+		await vi.waitFor(() => {
+			expect(serverLogs.info.join()).toContain("[vite] hot updated");
+		});
+	}
+);
--- a/packages/vite-plugin-cloudflare/playground/worker/__tests__/worker.spec.ts
+++ b/packages/vite-plugin-cloudflare/playground/worker/__tests__/worker.spec.ts
@@ -1,5 +1,12 @@
-import { expect, test } from "vitest";
-import { getTextResponse, serverLogs, viteTestUrl } from "../../__test-utils__";
+import * as fs from "node:fs";
+import * as path from "node:path";
+import { expect, test, vi } from "vitest";
+import {
+	getTextResponse,
+	rootDir,
+	serverLogs,
+	viteTestUrl,
+} from "../../__test-utils__";
 
 test("basic hello-world functionality", async () => {
 	expect(await getTextResponse()).toEqual("Hello World!");
@@ -19,4 +26,14 @@ test("receives the original host as the `X-Forwarded-Host` header", async () => 
 
 test("does not cause unhandled rejection", async () => {
 	expect(serverLogs.errors.join()).not.toContain("__unhandled rejection__");
+});
+
+test("updates using HMR code in Worker entry file", async () => {
+	const workerEntryPath = path.join(rootDir, "src", "index.ts");
+	const originalContent = fs.readFileSync(workerEntryPath, "utf-8");
+	fs.writeFileSync(workerEntryPath, originalContent);
+
+	await vi.waitFor(() => {
+		expect(serverLogs.info.join()).toContain("[vite] hot updated");
+	});
 });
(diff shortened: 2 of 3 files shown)


Phoenixmatrix:
I'm looking at the code but I'm not familiar with the full extent of the HMR setup in Vite. Does this do more than what I'd do if I added the HMR snippet in my entry myself?

Because adding it, while it seems to improve stability somewhat, definitely does NOT fix HMR entirely in large application. Its still very uninstable with just that snippet.

I can't get together a reproduceable sample without shipping my entire app (not open source unfortunately).

But looking at the behavior, the instant I hit save on my file, I see the HMR message in the terminal. I also see my app's page in the browser flicker a few times as the various HMRed impacted file get processed. I see network chatter as the files get loaded, especially the CSS. And then nothing happens.

My theory is because the Cloudflare Vite plugin rebundle the whole app (correct me if that's not true, its what folks in the Discord are saying), it takes several seconds to finish bundling, but the HMR step seem to "complete" long before that, so the changes don't get picked up. This is all uneducated guesses as best, but just adding the HMR code doesn't seem to do much.

If I use a small app like the RR7 template for Cloudflare, then it works 100% of the time, but that app builds almost instantly.

jamesopstad:
> I can see how this could improve the performance on code change. But I wonder if this addresses the issue you mentioned in [#9518 (comment)](https://github.com/cloudflare/workers-sdk/issues/9518#issuecomment-2991401574) as it might just making it less likely to happen?

Yep, I agree we shouldn't close the issue yet. This is an improvement rather than a fix and more investigation is still needed.

Issue #8063, opened 2025-02-08
🐛 BUG: wrangler prints ERROR for console.warn()


mxschumacher:
### Which Cloudflare product(s) does this pertain to?

Wrangler

### What versions are you using?

Wrangler 3.107.3

### What operating system and version are you using?

Fedora Linux 41 ( 6.11.10-300.fc41.x86_64 )

### Describe the Bug

```js
console.warn('Received request without Upgrade header on /websocket path');
```

prints

```console
✘ [ERROR] Received request without Upgrade header on /websocket path
```

when I run **pnpm wrangler dev**

It should print

```console
⚠️  [WARNING] Received request without Upgrade header on /websocket path
```

irvinebroque:
This is really confusing, we should fix it

petebacondarwin:
This has changed a bit. Now warnings are just logged as info and don't have any additional prefix.

PR #10004, opened 2025-07-17
fix wrangler dev logs being logged on the incorrect level in some cases

dario-piotrowicz:
Fixed https://jira.cfdata.org/browse/DEVX-2087
Fixes https://github.com/cloudflare/workers-sdk/issues/8063

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/10015
  - [ ] Not necessary because:


dario-piotrowicz:
@danlapid I've added you as a reviewer just in case you were interested since you added the structured logging feature to workerd 🙌 🫶

danlapid:
> @danlapid I've added you as a reviewer just in case you were interested since you added the structured logging feature to workerd 🙌 🫶

🥳 🥳 🥳 
Did everything work well for you with it? Any issues?

dario-piotrowicz:
> > @danlapid I've added you as a reviewer just in case you were interested since you added the structured logging feature to workerd 🙌 🫶
> 
> 🥳 🥳 🥳 Did everything work well for you with it? Any issues?

Hey @danlapid :slightly_smiling_face: 

All very good generally thanks :heart_hands: 

One thing that I thought could be a bit problematic but I am getting worried that it could be quite problematic is this regex: https://github.com/cloudflare/workers-sdk/blob/fdf3ecbf8b07c5193a0d5215e4c193059719182a/packages/wrangler/src/dev/miniflare.ts#L1192C3-L1192C74

I can be quite brittle.... I will try to make it more robust (I need to add some more testing around this :thinking:)

But however much I can improve it, I feel like it can always be a bit brittle to some extent :thinking: (since user logs can literally have anything in them...)
I was considering.... do you think we could also add a workerd option to have the message in the structured logs be in base64 or something to that extent? (so that we could extract the message and convert it back to text format on our end) that would make this sort of operation very robust I think :thinking:

danlapid:
> > > @danlapid I've added you as a reviewer just in case you were interested since you added the structured logging feature to workerd 🙌 🫶
> > 
> > 
> > 🥳 🥳 🥳 Did everything work well for you with it? Any issues?
> 
> Hey @danlapid 🙂
> 
> All very good generally thanks 🫶
> 
> One thing that I thought could be a bit problematic but I am getting worried that it could be quite problematic is this regex: https://github.com/cloudflare/workers-sdk/blob/fdf3ecbf8b07c5193a0d5215e4c193059719182a/packages/wrangler/src/dev/miniflare.ts#L1192C3-L1192C74
> 
> I can be quite brittle.... I will try to make it more robust (I need to add some more testing around this 🤔)
> 
> But however much I can improve it, I feel like it can always be a bit brittle to some extent 🤔 (since user logs can literally have anything in them...) I was considering.... do you think we could also add a workerd option to have the message in the structured logs be in base64 or something to that extent? (so that we could extract the message and convert it back to text format on our end) that would make this sort of operation very robust I think 🤔

Why not just use a json parser?

dario-piotrowicz:
> Why not just use a json parser?

mh... yeah that's a good point! Thanks I'll try that :slightly_smiling_face::+1:

dario-piotrowicz:
@danlapid, ok it should hopefully be all good 🙂 

I ended up with this logic:
https://github.com/cloudflare/workers-sdk/blob/b23b8440fbb86d062a3b50816f287c7135c9dbb2/packages/wrangler/src/dev/miniflare.ts#L1284-L1348

which seems to be working well 🙂

danlapid:
> @danlapid, ok it should hopefully be all good 🙂
> 
> I ended up with this logic:
> 
> https://github.com/cloudflare/workers-sdk/blob/b23b8440fbb86d062a3b50816f287c7135c9dbb2/packages/wrangler/src/dev/miniflare.ts#L1284-L1348
> 
> which seems to be working well 🙂

Why not just split by newlines though?
Are you worried that the logs can have newlines? 
I am pretty sure these should be correctly escaped by our json encoder.

dario-piotrowicz:
> Why not just split by newlines though? Are you worried that the logs can have newlines? I am pretty sure these should be correctly escaped by our json encoder.

Nice! I actually didn't know that they were consistently divided by newlines 😃 

Thanks for the suggestion, I've applied that and it seems to work very nicely 🫶

.changeset/deep-lizards-move.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+(note that the warning is printed as an error and the debug log is printed even if by default it should not)
+
+the changes here make sure that the logs are instead logged to their correct level, so for the code about the following will be logged instead:
+
+```text
+✘ [ERROR] this is an error
+
+▲ [WARNING] this is a warning
```
emily-shen, line 32: mention that `--log-level=debug` will show the debug?

packages/miniflare/src/index.ts
```diff
@@ -967,6 +969,8 @@ export class Miniflare {
 		}
 
 		this.#log = this.#sharedOpts.core.log ?? new NoOpLog();
+		this.#structuredWorkerdLogs =
+			this.#sharedOpts.core.structuredWorkerdLogs ?? false;
```
emily-shen, line 921: is this false because we would break direct miniflare users otherwise?

packages/miniflare/src/runtime/config/generated.ts
```diff
@@ -952,6 +966,7 @@ export const Worker_Binding_Type_Which = {
 	QUEUE: 11,
 	ANALYTICS_ENGINE: 12,
 	HYPERDRIVE: 13,
+	DURABLE_OBJECT_CLASS: 14,
```
emily-shen, line 969: this is all auto-generated right? but do you why the DO stuff has been added in? is this just because we haven't updated this file recently since the facets stuff was added to the runtime? i wonder if this is something we should automate

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -1195,94 +1192,123 @@ export function handleRuntimeStdio(stdout: Readable, stderr: Readable) {
...
+			return logger.log(getSourceMappedString(message));
+		});
+	});
 
-		// anything not explicitly handled above should be logged as an error (via stderr)
-		else {
-			logger.error(getSourceMappedString(chunk));
-		}
+	stderr.on("data", (chunk: Buffer | string) => {
+		// TODO: Some errors still get reported through stderr, investigate
```
emily-shen: what sort of errors? should we create an issue to track?

packages/miniflare/src/index.ts
dario-piotrowicz: yes, we want this as a miniflare opt-in otherwise this would be a breaking change for the miniflare package, right?

packages/miniflare/src/runtime/config/generated.ts
dario-piotrowicz: yeah this was automatically generated via 
https://github.com/cloudflare/workers-sdk/blob/031c7f60df4a26b52c13a8c0395f12603a4b8b8e/packages/miniflare/package.json#L34

honestly I haven't really looked much at the changes in the file 😕  (as it just works and doesn't seem to break anything)

I guess it can be periodically automated, but I _think_ that there's no much benefit in it since the file only needs to update when miniflare needs to pass new/different capnp configs, and that updating it only when necessary makes sense to me 🤔 
(In other words, I think that the file getting updated when Miniflare doesn't use the new options is not going to have any effect anyways)

I'd be keen to hear what @penalosa thinks of this 🙂

packages/wrangler/src/dev/miniflare.ts
dario-piotrowicz: I can't fully remember the errors, I'll have a look and let you know 👍

We could create a ticket but I was also getting the feeling that this would 99.9% of the time work as expected anyways, so might not be a huge deal 🤔 🤷

.changeset/deep-lizards-move.md
dario-piotrowicz: sure 🙂👍

dario-piotrowicz: done 🙂

packages/wrangler/src/dev/miniflare.ts
dario-piotrowicz: mh... I don't seem to be able to get an errors here.... I wonder if this is actually no longer correct and something that only happened when I implemented the thing 🤔

danlapid: When structuredLogging is turned on you shouldn’t get anything on stderr
Would love to investigate what you got

dario-piotrowicz: Ah ok, found one, if I throw an error at the top level of a worker I indeed get the error printed to `stderr`:
Screenshot 2025-07-23 at 09 58 39

dario-piotrowicz: (I can't find other instances or errors being printed to `stderr`... maybe that's just it 🤔)

danlapid: Can repro, trying to figure out where it's coming from 🙏

danlapid: Also found that logWarning goes directly to stderr which is probably a bug as well though I already had a PR in flight for that, can just bump the priority

danlapid: Ok it's coming from logs on the kj::ProcessContext for configErrors, so errors during startup.
That's rather annoying but it'll be quite annoying to fix, how high prio is this?
Is it actually blocking something on your end or just another code branch to handle?

dario-piotrowicz: Thanks a lot for having had a look 🫶 

> Is it actually blocking something on your end or just another code branch to handle?

No, I think that it's totally fine to keep it like this, it is not a huge deal

The only thing that I'd love to understand here is whether the text we do get from stderr we can just print as errors, or if we should also try to classify it and potentially redirect it to different log levels.

Basically I am wondering if we need to apply this logic:
https://github.com/cloudflare/workers-sdk/blob/98b0c9eed3a20f0fd00c4ac521fa338cb205fcd8/packages/wrangler/src/dev/miniflare.ts#L1163
Also to the stderr stream, or if we can just run `logger.error` on all the test we get from it.

Do you know?

danlapid: It’ll never be user originated error logs

dario-piotrowicz: mh... what I meant is that workerd can produce some (error) logs such as for example:
- `Address already in use; etc...`
- `error: CODE_MOVED for unknown code block?`
- etc...

Currently wrangler listens on `stderr` and performs checks to see if it is receiving such (error) logs, so that it can either ignore or tweak them (or print them with a specific log level).

See:
https://github.com/cloudflare/workers-sdk/blob/1e92dd815cdccfa4d48cc988f45addea356eedb9/packages/wrangler/src/dev/miniflare.ts#L1230-L1290

Do you know if the same checks will be needed on the raw `stderr` logs when structured logs is on? or, with structured logs the errors that `stderr` can receive are of a specific category (that doesn't include for example `CODE_MOVED` type of errors)

danlapid: CODE_MOVED for unknown code block?
for example will definitely be on stdout as part of structured logging.

the etc part, idk.

Any config error can be sent to stderr at the moment.

dario-piotrowicz: ok, thank you very much, I'll keep the existing stderr stream handling then just to be on the safe side 🙂👍

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -1199,34 +1199,128 @@ export function handleRuntimeStdio(stdout: Readable, stderr: Readable) {
...
 
+	// TODO: Even after the introduction of structured logs some errors can make it
+	//       into stderr (likely only startup errors), so to play it safe we handle
+	//       the stderr stream in the exact same way we did before structured logs
+	//       were introduced. So some of the code below can be unnecessary, although
+	//       not harmful if we want we can try to clean up some of it at some point
+	//       in the future (and possibly the workerd could also fix this on their
+	//       side and never send anything to the stderr).
+	//       For more context see:
+	//         https://github.com/cloudflare/workers-sdk/pull/10004#discussion_r2222806461
```
dario-piotrowicz: @emily-shen pinging you to conclude the conversation regarding this 🙂 

Sounds like the issues we get here are startup ones and that we're not 100% which errors can make it here and which don't, thus I reinstated the old stderr handling code and put this comment

I think this can be a todo that stays here for a while... (instead of a TODO I can make it into just an informative note? 🤔)

I'm ok creating a ticket for it if we want 🙂 
(I can do it as soon as we merge the PR?)

dario-piotrowicz:
PR converted to draft as I think it now needs to wait for https://github.com/cloudflare/workerd/pull/4620 to be merged (released and bumped in wrangler).

Also a test for top level errors (like the one I [mentioned above](https://github.com/cloudflare/workers-sdk/pull/10004#discussion_r2224896223)) that would get logged onto stderr, should be added.

packages/wrangler/src/dev/miniflare.ts
emily-shen: looks like you've identified the issue and everything still works, this comment seems like enough :) thanks!

dario-piotrowicz: Comment removed anyways 😅

fixtures/worker-logs/tests/index.test.ts
```diff
@@ -0,0 +1,256 @@
+import { setTimeout } from "node:timers/promises";
+import { resolve } from "path";
+import { describe, test } from "vitest";
+import { runWranglerDev } from "../../shared/src/run-wrangler-long-lived";
+
+async function getWranglerDevOutput(
+	type: "module" | "service",
+	extraArgs: string[] = [],
+	customMessage?: string,
+	twoRequests = false
```
emily-shen: i don't suppose you could make this more generic by making this number of requests or something?

fixtures/worker-logs/tests/index.test.ts
```diff
@@ -0,0 +1,256 @@
...
+				// needed to trigger the log here...
+				true
+			);
+			expect(output).toContain("<<<<<this is a debug message>>>>>");
+		});
+
+		test("with --log-level=none", async ({ expect }) => {
+			const output = await getWranglerDevOutput("module", ["--log-level=none"]);
+			expect(output).toMatchInlineSnapshot(`""`);
+		});
```
emily-shen, line 141: could you add a test for the `WRANGLER_LOG=debug` env var? we've had issues with that not working previously

packages/miniflare/src/plugins/core/index.ts
```diff
@@ -247,6 +247,10 @@ export const CoreSharedOptionsSchema = z.object({
 	defaultPersistRoot: z.string().optional(),
 	// Strip the MF-DISABLE_PRETTY_ERROR header from user request
 	stripDisablePrettyError: z.boolean().default(true),
+
+	// Whether to get structured logs from workerd or not (default to `false`)
+	// This option is useful in combination with a custom handleRuntimeStdio.
+	structuredWorkerdLogs: z.boolean().optional(),
```
emily-shen:
```suggestion
	structuredWorkerdLogs: z.boolean().default(false),
```

packages/miniflare/test/index.spec.ts
```diff
@@ -3386,3 +3386,96 @@ test("Miniflare: MINIFLARE_WORKERD_CONFIG_DEBUG controls workerd config file cre
 	);
 	await mf.dispose();
 });
+
+test.only("Miniflare: logs are treated as standard stdout/stderr chunks be default", async (t) => {
```
emily-shen:
```suggestion
test("Miniflare: logs are treated as standard stdout/stderr chunks be default", async (t) => {
```

packages/miniflare/test/index.spec.ts
```diff
@@ -3386,3 +3386,96 @@ test("Miniflare: MINIFLARE_WORKERD_CONFIG_DEBUG controls workerd config file cre
...
+	t.teardown(() => mf.dispose());
+
+	const response = await mf.dispatchFetch("http://localhost");
+	await response.text();
+
+	t.is(collected.stdout, "__LOG__\n__INFO__\n__DEBUG__\n");
+	t.is(collected.stderr, "__WARN__\n__ERROR__\n");
+});
+
+test.only("Miniflare: logs are structured and all sent to stdout when structuredWorkerdLogs is true", async (t) => {
```
emily-shen:
```suggestion
test("Miniflare: logs are structured and all sent to stdout when structuredWorkerdLogs is true", async (t) => {
```

packages/wrangler/src/dev/miniflare/stdio.ts
```diff
@@ -0,0 +1,232 @@
...
+			// up to that point, the rest (which is the beginning of a
+			// new structured log) will be saved for later
+			currentLogsStr = fullStreamOutput.slice(0, lastNewlineIdx);
+			streamAccumulator = fullStreamOutput.slice(lastNewlineIdx + 1);
+		} else {
+			// If we didn't find a newline we're dealing with a structured
+			// log that has been split, so let's save the whole thing for
+			// later (so that we can process the log once we've seen it
+			// in full)
+			streamAccumulator = fullStreamOutput;
```
emily-shen, line 53: can it be split across only two chunks or could it be spread across more? (should this be a += that gets reset wonceen the newline is found)

packages/miniflare/test/index.spec.ts
dario-piotrowicz: 🤦👍

packages/wrangler/src/dev/miniflare/stdio.ts
dario-piotrowicz: it can spread across an arbitrary number of chunks (for example `console.log("z".repeat(2 ** 20));` spreads across a bunch of chunks)

> should this be a += that gets reset wonceen the newline is found

I don't fully understand... that's basically what this is already, isn't it?

packages/miniflare/test/index.spec.ts
dario-piotrowicz: 🤦 👍

fixtures/worker-logs/tests/index.test.ts
dario-piotrowicz: sure I'll do that, since this is only a testing function I thought it was clearer this way, but I'm happy to make it more generic, the difference in my opinion is very small anyways 👍

fixtures/worker-logs/tests/index.test.ts
dario-piotrowicz: sure 🙂👍

packages/wrangler/src/dev/miniflare/stdio.ts
emily-shen: oh sorry i reread the code and so it does 😶‍🌫️

fixtures/worker-logs/tests/index.test.ts
dario-piotrowicz: does this work for you? 🙂 
https://github.com/cloudflare/workers-sdk/pull/10004/commits/444491e99b1b799397cc2f714b73227d58fe56c2

fixtures/worker-logs/tests/index.test.ts
dario-piotrowicz: done 🙂 
https://github.com/cloudflare/workers-sdk/pull/10004/commits/9bc9e370fb29922c84866f5f78c00819afadfd03

dario-piotrowicz:
I've added the `skip-v3-pr` label just to make the CI happy, I do already have the v3 PR open, once this PR is merged I will fix the v3 PR

PR #10005, opened 2025-07-17
Resolve dockerfile path based on original, non-redirected config path

emily-shen:
userConfigPath is the path of the original wrangler config, while configPath could point to a redirected wrangler config. We should resolve the dockerfile and build context based on the original config's path. This makes sure containers can be deployed if the app has been built by the vite plugin.

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: preemptive bugfix
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -17,6 +17,7 @@ export async function maybeBuildContainer(
 	imageTag: string,
 	dryRun: boolean,
 	pathToDocker: string,
+	/** The original (non-redirected) user config path */
```
dario-piotrowicz, line 20: what about renaming the parameter `userConfigPath` to avoid possible future confusion? 🤔

packages/wrangler/src/__tests__/cloudchamber/deploy.test.ts
```diff
@@ -346,6 +346,113 @@ describe("wrangler deploy with containers", () => {
 		expect(std.warn).toMatchInlineSnapshot(`""`);
 	});
 
+	it("resolve the dockerfile based on the non-redirected userConfigPath rather than the actual config path", async () => {
```
dario-piotrowicz, line 349:
```suggestion
	it("resolve the dockerfile based on the non-redirected userConfigPath rather than the redirected config path", async () => {
```
🤔 ?

dario-piotrowicz (approved):
Looks good to me 😄

(just left a couple of very minor comments 🙂)

flakey5 (approved):

packages/wrangler/src/cloudchamber/deploy.ts
emily-shen: ah that's probably better, but this needs to go out in the next release so i'm just going to kick this one through as is - sorry!

Issue #9999, opened 2025-07-17
Types not properly generated for Durable Objects


astralarya:
### What versions & operating system are you using?

wrangler 4.24.4

### Please provide a link to a minimal reproduction

Specify a durable object in `wrangler.jsonc`
```json
  "durable_objects": {
    "bindings": [
      {
        "name": "DURABLE_OBJECT",
        "class_name": "MyDurableObject",
      },
    ],
  }
```

Run `wrangler types --env-interface Env` to get this output:

```ts
// Generated by Wrangler by running `wrangler types --env-interface Env` (hash: 3be16e8d647f1e872fcb621c185ccd33)
// Runtime types generated with workerd@1.20250712.0 2025-07-14 nodejs_compat
declare namespace Cloudflare {
	interface Env {
		DURABLE_OBJECT: DurableObjectNamespace /* MyDurableObject */;
	}
}
interface Env extends Cloudflare.Env {}
```

### Describe the Bug

Type of `DURABLE_OBJECT` should be `DurableObjectNamespace<MyDurableObject>`.

nehaaprasad:
Hello @astralarya !
I hope this change ensures Durable Object types use proper TypeScript generics instead of comments, providing better type safety

tv42:
Is it triggered by explicitly saying `--env-interface Env`? My worker has `--env-interface FooEnv` and gets this right:

```
declare namespace Cloudflare {
	interface Env {
		Item: DurableObjectNamespace<import("./index").Item>;
	}
}
interface ItemEnv extends Cloudflare.Env {}
```

emily-shen:
Hi, we do generate something like `DurableObjectNamespace<MyDurableObject>` if we can find the corresponding DO export in your script.

i can't seem to reproduce this myself with the snippets you've provided, although I did add a very mininmal script that exports the DO like this:
```
export default {
	async fetch() {
		return new Response("foo");
	},
};

export class FooBar extends DurableObject {}
```

 Could you share a minimal reproduction of this issue so we can debug this? 


The` --env-interface` part shouldn't have an impact here, that just determines what the env type name should be.

petebacondarwin:
Closing as we have no reproduction to diagnose.

implicit-invocation:
@astralarya 
did you use `export * from "./path/to/your/DO";` instead of `export { MyDurableObject } from "./path/to/your/DO";`?  
For me, `export *` caused the problem.

PR #10006, opened 2025-07-17
fix: Durable Object type generation to use proper TypeScript generics

naaa760:
Description:

#9999 

> Fixes issue where wrangler types generated Durable Object types using comments instead of proper TypeScript generics. Now generates DurableObjectNamespace<MyDurableObject> instead of DurableObjectNamespace /* MyDurableObject */ for better type safety.

> Checkboxes:

- Tests: [ ] Tests not necessary because: Simple logic change with clear behavior
- Public documentation: [ ] Documentation not necessary because: No API changes
- Wrangler V3 Backport: [ ] Not necessary because: This is a Wrangler change

### DIFF
--- a/packages/wrangler/src/type-generation/index.ts
+++ b/packages/wrangler/src/type-generation/index.ts
@@ -412,10 +412,8 @@ export async function generateEnvTypes(
 
 			if (importPath && exportExists) {
 				typeName = `DurableObjectNamespace<import("${importPath}").${durableObject.class_name}>`;
-			} else if (durableObject.script_name) {
-				typeName = `DurableObjectNamespace /* ${durableObject.class_name} from ${durableObject.script_name} */`;
 			} else {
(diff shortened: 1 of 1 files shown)


emily-shen:
Hi, first off, thanks for taking the time to make this contribution!

However, I'm not sure we want to do this - we do import the DO type and pass it in to the type _if_  we are able to find the export. However, we can't do this if we can't find the export - the output of the change in this PR will not be correct, as the types file will not have access to the DO type. Compare this to this line above https://github.com/cloudflare/workers-sdk/pull/10006/files#diff-80c0ac2b61f944d3d38c4e4ef4ffce1e85b751b564027163ee41d62acd34736fL414, where we are able to import it and pass it in to the generic param.

Again, thanks for the contribution, and feel free to follow up on the issue with more details if you're still having problems.

PR #10007, opened 2025-07-17
Do not clone request in router-worker

WillTaylorDev:
- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: router-worker change, tested in staging.
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: router-worker change

### DIFF
--- a/packages/workers-shared/router-worker/src/worker.ts
+++ b/packages/workers-shared/router-worker/src/worker.ts
@@ -110,7 +110,7 @@ export default {
 						dispatchType: DISPATCH_TYPE.WORKER,
 					});
 
-					let shouldBlockNonImageResponse = false;
+					let shouldCheckContentType = false;
 					if (url.pathname.endsWith("/_next/image")) {
 						// is a next image
 						const queryURLParam = url.searchParams.get("url");
@@ -121,7 +121,7 @@ export default {
 								request.headers.get("sec-fetch-dest") !== "image"
 							) {
 								// that was not loaded via a browser's <img> tag
-								shouldBlockNonImageResponse = true;
+								shouldCheckContentType = true;
 								analytics.setData({ abuseMitigationURLHost: queryURLParam });
 							}
 							// otherwise, we're good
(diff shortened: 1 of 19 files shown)


petebacondarwin (approved):

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -119,8 +117,8 @@ export default {
 						if (queryURLParam && !queryURLParam.startsWith("/")) {
 							// that's a remote resource
 							if (
-								maybeSecondRequest.method !== "GET" ||
-								maybeSecondRequest.headers.get("sec-fetch-dest") !== "image"
+								request.method !== "GET" ||
+								request.headers.get("sec-fetch-dest") !== "image"
```
petebacondarwin, line 121: Hmm, does reading the headers cause the request to need to be cloned?

petebacondarwin (dismissed):
Actually this appears to be legitimately failing.

vicb:
Looks like the body is consumed at https://github.com/cloudflare/workers-sdk/blob/c7de92e8702da59fd0858147192c2a70abfb35e9/packages/workers-shared/router-worker/src/worker.ts#L231

packages/workers-shared/router-worker/src/worker.ts
WillTaylorDev: Resolved internally, thanks for the unlock Pete/Victor!

vicb (approved):
Thanks!

GregBrimble (approved):

packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -218,7 +216,13 @@ export default {
 			}
 
 			// If we have a user-Worker, but no assets, dispatch to Worker script
-			const assetsExist = await env.ASSET_WORKER.unstable_canFetch(request);
+			// Do not pass the original request as it would consume the body
+			const assetsExist = await env.ASSET_WORKER.unstable_canFetch(
+				new Request(request.url, {
```
matthewdavidrodgers, line 221: interesting - are the url and headers really all that's needed? if so i wonder if we should change the signature of this method to make that explicit so we don't mess this up in the future by trying to read the body

not an easy thing to do i don't think, so don't think that's necessary

WillTaylorDev: Yup, that's all that's needed! We did talk about changing the signature (it's unstable, so we have the ability), but this change is a much quicker unblock until we get to that point.

PR #10009, opened 2025-07-17
Fix container diffs when using JSONC config

gpanders:
Fixes CC-5522.

The diff printing code contains a number of issues:

1. It is replicated between cloudchamber/apply.ts and containers/deploy.ts.

   * Move all of the diff printing code into cloudchamber/helpers/diff.ts. At the same time, we refactor the diff code to remove the singleton static "Diff" object and move methods related to generating the diff into the Diff class. The diff generation code comes from a 3rd party BSD-3 licensed repo, so we also include the license in the file to adhere to the license requirements.

2. It is tailored specifically for TOML files. At the time this was written, wrangler.toml was the primary/default config file format. That is no longer the case, and increasingly users are using wrangler.jsonc. JSONC files were not formatted correctly at all: they had no proper indentation and the diff printing code didn't properly account for the "context" around changed lines.

   * The solution here is to eliminate the TOML specific code completely and have the diffing code just diff plain strings. This means we lose the syntax highlighting for TOML, however I think this is acceptable since we do not also offer syntax highlighting for JSON, and providing the syntax highlighting requires a lot of code (it also only works in specific scenarios and hard codes color codes which may not match the user's terminal colorscheme, which is an accessibility issue). The diff code now just prints 3 lines of context before and after each changed chunk, without attempting to parse anything.

3. The helpers/diff.ts file contains some functions which are unrelated to generating diffs.

   * Create a new file `src/utils/sortObjectRecursive.ts` that contains these functions.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: containers change

### DIFF
--- a/.changeset/mean-schools-see.md
+++ b/.changeset/mean-schools-see.md
@@ -1,0 +1,8 @@
+---
+"miniflare": patch
+---
+
+Fix D1 SQL dump generation: escape identifiers and handle SQLite's dynamic typing
+
+Escape column and table names to prevent SQL syntax errors.
+Escape values based on their runtime type to support SQLite's flexible typing.
--- a/.changeset/old-clubs-swim.md
+++ b/.changeset/old-clubs-swim.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Disallow users from pushing images with unsupported platforms to the container image registry
--- a/.changeset/shaggy-insects-deny.md
+++ b/.changeset/shaggy-insects-deny.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-shared": patch
+---
+
+Bugfix: Removes unnecessary cloning of the request. This is no longer needed. We were also seeing failures in runtime for large files due to this.
--- a/.changeset/stale-houses-matter.md
+++ b/.changeset/stale-houses-matter.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/workers-shared": patch
+---
+
+block responses with multiple Content Type values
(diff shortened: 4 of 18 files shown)


packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -130,12 +130,12 @@ describe("cloudchamber apply", () => {
...
-			│   tier = 2
+			│     [containers.constraints]
+			│     tier = 2
 			│
-			│   [containers.configuration]
-			│   image = \\"./Dockerfile\\"
-			│   instance_type = \\"dev\\"
+			│     [containers.configuration]
+			│     image = \\"./Dockerfile\\"
+			│     instance_type = \\"dev\\"
```
gpanders, line 138: The extra indentation here and throughout is because the [TOML library we use](https://www.npmjs.com/package/@iarna/toml) adds indentation to nested tables. This is, unfortunately, not configurable. We can propose a patch upstream to make this configurable, but it's also not really a big deal.

flakey5 (approved):

packages/wrangler/src/__tests__/containers/apply.test.ts
```diff
@@ -1956,4 +1962,113 @@ describe("containers apply", () => {
 		const app = await applicationReqBodyPromise;
 		expect(app.configuration?.instance_type).toEqual("standard");
 	});
+
+	it("formats JSONC diffs properly", async () => {
```
emily-shen, line 1966: I noticed that the JSON diffs can show a change sometimes due to a trailing comma being added - since the last item doesn’t have a comma, if something gets added below that, the item below will show a diff even if the only change is the comma. This is incredibly trivial tbh, but just flagging it :)

PR #10010, opened 2025-07-18
reduce flakes on interactive-dev fixture tests

petebacondarwin:
I've had this test fail many times locally.
@emily-shen suggested the problem was the fetch can hang.
So hopefully this change will make the test a bit more reliable.


petebacondarwin:
Oops! I forgot to remove the `RequestInfo` type import. Sorry.

PR #10012, opened 2025-07-18
Support dynamic worker loading

penalosa:
Support https://github.com/cloudflare/workerd/pull/4383

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this is an experimental pre-release feature
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not available in v3

### DIFF
--- a/.changeset/tall-emus-sing.md
+++ b/.changeset/tall-emus-sing.md
@@ -1,0 +1,6 @@
+---
+"miniflare": minor
+"wrangler": minor
+---
+
+Support unsafe dynamic worker loading bindings
--- a/fixtures/dynamic-worker-loading/wrangler.jsonc
+++ b/fixtures/dynamic-worker-loading/wrangler.jsonc
@@ -1,0 +1,13 @@
+{
+	"name": "dynamic-worker-loading",
+	"main": "src/index.ts",
+	"compatibility_date": "2023-05-04",
+	"unsafe": {
+		"bindings": [
+			{
+				"name": "LOADER",
+				"type": "worker-loader",
+			},
+		],
+	},
+}
--- a/packages/miniflare/src/plugins/worker-loader/index.ts
+++ b/packages/miniflare/src/plugins/worker-loader/index.ts
@@ -1,0 +1,34 @@
+import { z } from "zod";
+import { Worker_Binding } from "../../runtime";
+import { Plugin } from "../shared";
+
+export const WorkerLoaderConfigSchema = z.object({
+	id: z.string().optional(),
+});
+export const WorkerLoaderOptionsSchema = z.object({
+	workerLoaders: z.record(WorkerLoaderConfigSchema).optional(),
+});
+
+export const WORKER_LOADER_PLUGIN_NAME = "worker-loader";
+
+export const WORKER_LOADER_PLUGIN: Plugin<typeof WorkerLoaderOptionsSchema> = {
+	options: WorkerLoaderOptionsSchema,
+	getBindings(options) {
+		if (!options.workerLoaders) {
+			return [];
+		}
+		const bindings = Object.entries(options.workerLoaders).map<Worker_Binding>(
+			([name]) => ({
+				name,
+				workerLoader: {},
+			})
+		);
+		return bindings;
+	},
+	getNodeBindings() {
+		return {};
+	},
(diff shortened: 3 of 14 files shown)


danlapid (approved):

danlapid:
Maybe @kentonv will have some comments

.changeset/tall-emus-sing.md
```diff
@@ -0,0 +1,6 @@
+---
+"miniflare": patch
+"wrangler": patch
```
petebacondarwin: Surely minors?

fixtures/dynamic-worker-loading/wrangler.jsonc
```diff
@@ -0,0 +1,14 @@
+{
+	"name": "dynamic-worker-loading",
+	"main": "src/index.ts",
+	"compatibility_date": "2023-05-04",
+	"unsafe": {
+		"bindings": [
+			{
+				"name": "LOADER",
+				"type": "worker-loader",
+				"id": "random-id",
```
petebacondarwin: What does `id` mean in this binding.
I guess my question is, "What does it mean to have multiple Worker Loaders?"

packages/miniflare/src/plugins/worker-loader/index.ts
```diff
@@ -0,0 +1,36 @@
...
+				name,
+				workerLoader: {
+					id: config.id,
+				},
+			})
+		);
+		return bindings;
+	},
+	getNodeBindings() {
+		return {};
```
petebacondarwin, line 29: Not accessible from `getPlatformProxy()`?

petebacondarwin (approved):

petebacondarwin:
Approved as this is good as it stands but I do think it would be nice to support `getPlatformProxy()` if it doesn't already.

fixtures/dynamic-worker-loading/wrangler.jsonc
kentonv: Miniflare should randomly-generate this ID instead. It turns out the edge runtime doesn't support multiple bindings to the same worker loader namespace so there's not much reason to expose this.

kentonv: Oh wait, no need to randomly generate. It can just be left null (not empty string, but null/omitted), in which case workerd will do the right thing.

kentonv:
I'd love to see this merged and released soon as I need it for a project. :)

fixtures/dynamic-worker-loading/wrangler.jsonc
petebacondarwin: Will it be needed for production deployments? In other words can we just remove it from wrangler.jsonc altogether? Or just ignore it in local dev?

threepointone:
Let's talk about this before landing it? Just want to make sure we agree on the api and how we're going to roll it out. @kentonv you can use this locally immediately with the prerelease builds `npm i https://pkg.pr.new/wrangler@10012`

penalosa:
@threepointone this isn't a stable release—it's going out as an unsafe binding. We can (and probably will) change the API/binding config before graduating to a stable binding

fixtures/dynamic-worker-loading/wrangler.jsonc
kentonv: Production does not support this option at all, that's why I recommend removing it.

kentonv:
Not sure if this is a problem in this PR or pre-existing, but I noticed that when I run `wrangler dev`, this binding is displayed wrong:

```
Your Worker has access to the following bindings:
Binding                                                      Resource             Mode
env.worker-loader (LOADER)                                   Unsafe Metadata      not supported
```

The binding name and type are swapped in the display -- it should say:

```
env.LOADER (worker-loader)                                   Unsafe Metadata      not supported
```

(The binding seems to work, though. This is just a display problem.)

erifystudio (commented):
✅ Solid work — this PR strengthens the foundation. Keep building forward.

erifystudio (commented):
✅ Solid work — this PR strengthens the foundation. Keep building forward.

PR #10014, opened 2025-07-18
feat(unenv-preset): shim debug package for cloudflare env support

edmundhung:
Fixes [DEVX-1973](https://jira.cfdata.org/browse/DEVX-1973)

The `debug` package is used by a lot of libraries, but in Wrangler it either gets replaced with a no-op or resolves to the browser build in Vite. That makes it pretty hard to see debug output when working on your Worker.

This PR adds a custom shim to the unenv preset by bundling a patched version of `debug` that includes a `workerd` export. The patch lives in the `worker-sdk` repo and uses the importable `env` interface from `cloudflare:workers` instead of relying on `process.env`, which is only available when the `nodejs_compat` flag is enabled.

After this change, you can just set a `DEBUG` var on the worker with the namespaces you want, and you will start seeing logs from any matching `debug` calls.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no docs for unenv
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: unenv only


patches/debug@4.4.1.patch
```diff
@@ -0,0 +1,40 @@
...
+@@ -53,6 +53,12 @@
+   },
+   "main": "./src/index.js",
+   "browser": "./src/browser.js",
++  "exports": {
++    "node": "./src/index.js",
++    "browser": "./src/browser.js",
++    "workerd": "./src/workerd.js",
++    "default": "./src/index.js"
++  },
```
vicb: Some builds use both the "node" and "workerd" conditions.
In that case we should use "workerd" and the order conditions are checked is [the order in this file](https://esbuild.github.io/api/#how-conditions-work)

```suggestion
+  "exports": {
+    "workerd": "./src/workerd.js",
+    "node": "./src/index.js",
+    "browser": "./src/browser.js",
+    "default": "./src/index.js"
+  },
```

package.json
```diff
@@ -79,6 +79,7 @@
 		},
 		"patchedDependencies": {
 			"@cloudflare/component-listbox@1.10.6": "patches/@cloudflare__component-listbox@1.10.6.patch",
+			"debug@4.4.1": "patches/debug@4.4.1.patch",
```
vicb, line 82: I guess there is no way to move that to the preset?

vicb (approved):
LGTM with a few minor comments.

Have you considered submitting a PR upstream so that we can eventually get ride of the patch and shim?

package.json
edmundhung: Unfortunately, no. I did a bit of search on pnpm and it seems to be by design because pnpm uses symlinks and there are only one copy of the package.

patches/debug@4.4.1.patch
edmundhung: Good call! Updated in [b4271d1](https://github.com/cloudflare/workers-sdk/pull/10014/commits/b4271d1bd18e8e89bb931e67ee5e52a3474c2e52)

edmundhung:
> Have you considered submitting a PR upstream so that we can eventually get ride of the patch and shim?

I have only implemented the bare minimum in this patch. There is no color or namespace update support. It is not a bad idea but it will require some more effort to get a full implementation.

PR #10016, opened 2025-07-18
feat: interactively handle wonky asset deploy commands

emily-shen:
Interactively handle `wrangler deploy`s that are probably assets-only, where there is **no config file found** and flags are incorrect or missing.

For example:

`npx wrangler deploy ./public` will now ask if you meant to deploy an folder of assets only, ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.

`npx wrangler deploy --assets=./public` will now ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.

In non-interactive contexts, we will error as we currently do.

Out of scope (for this PR) are worker or worker and asset deploys.

Screenshot 2025-07-18 at 14 31 26

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: self-documenting?
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/deep-onions-move.md
+++ b/.changeset/deep-onions-move.md
@@ -6,8 +6,8 @@ Interactively handle `wrangler deploy`s that are probably assets-only, where the
 
 For example:
 
-`npx wrangler deploy ./public` will now ask if you meant to deploy an folder of assets only, ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.
+`npx wrangler deploy ./public` will now ask if you meant to deploy a folder of assets only, ask for a name, set the compat date and then ask whether to write your choices out to `wrangler.json` for subsequent deployments.
 
-`npx wrangler deploy --assets=./public` will now ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.
+`npx wrangler deploy --assets=./public` will now ask for a name, set the compat date and then ask whether to write your choices out to `wrangler.json` for subsequent deployments.
 
-In non-interactive contexts, we will error as we currently do.
+In non-interactive contexts, Wrangler will error as it currently does.
--- a/.changeset/deep-onions-move.md
+++ b/.changeset/deep-onions-move.md
@@ -1,0 +1,13 @@
+---
+"wrangler": minor
+---
+
+Interactively handle `wrangler deploy`s that are probably assets-only, where there is no config file and flags are incorrect or missing.
+
+For example:
+
+`npx wrangler deploy ./public` will now ask if you meant to deploy an folder of assets only, ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.
+
+`npx wrangler deploy --assets=./public` will now ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.
+
+In non-interactive contexts, we will error as we currently do.
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -2729,6 +2729,7 @@ addEventListener('fetch', event => {});`
 				// Mock the date to ensure consistent compatibility_date
 				vi.setSystemTime(new Date("2024-01-01T00:00:00Z"));
 
+				// so that we can test that the name prompt defaults to the directory name
 				fs.mkdirSync("my-site");
 				process.chdir("my-site");
 				const assets = [
@@ -2747,6 +2748,7 @@ addEventListener('fetch', event => {});`
 			});
 			afterEach(() => {
 				setIsTTY(false);
+				vi.useRealTimers();
 			});
 
 			it("should handle `wrangler deploy <directory>`", async () => {
@@ -2777,7 +2779,7 @@ addEventListener('fetch', event => {});`
 						},
 					},
 				});
-				expect(fs.readFileSync("wrangler.json", "utf-8"))
+				expect(fs.readFileSync("wrangler.jsonc", "utf-8"))
 					.toMatchInlineSnapshot(`
 					"{
 					  \\"name\\": \\"test-name\\",
@@ -2801,8 +2803,10 @@ addEventListener('fetch', event => {});`
 					    \\"directory\\": \\"./assets\\"
 					  }
 					}
-					 to <cwd>/wrangler.json.
-					Please run wrangler deploy instead of wrangler deploy ./assets next time. Wrangler will automatically use the configuration saved to wrangler.json.
+					 to <cwd>/wrangler.jsonc.
+					Please run \`wrangler deploy\` instead of \`wrangler deploy ./assets\` next time. Wrangler will automatically use the configuration saved to wrangler.jsonc.
+
+					Proceeding with deployment...
 
 					Total Upload: xx KiB / gzip: xx KiB
 					Worker Startup Time: 100 ms
@@ -2812,6 +2816,7 @@ addEventListener('fetch', event => {});`
 					Current Version ID: Galaxy-Class"
 				`);
 			});
+
 			it("should handle `wrangler deploy --assets` without name or compat date", async () => {
 				// if the user has used --assets flag and args.script is not set, we just need to prompt for the name and add compat date
 				mockPrompt({
@@ -2837,7 +2842,7 @@ addEventListener('fetch', event => {});`
 						},
 					},
 				});
-				expect(fs.readFileSync("wrangler.json", "utf-8"))
+				expect(fs.readFileSync("wrangler.jsonc", "utf-8"))
 					.toMatchInlineSnapshot(`
 					"{
 					  \\"name\\": \\"test-name\\",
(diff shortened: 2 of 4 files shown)


packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2720,6 +2721,157 @@ addEventListener('fetch', event => {});`
...
+					  \\"assets\\": {
+					    \\"directory\\": \\"./assets\\"
+					  }
+					}"
+				`);
+				expect(std.out).toMatchInlineSnapshot(`
+					"
+
+
+					No compatibility date found Defaulting to today: 2024-01-01
```
vicb, line 2796: nit: Defaulting => defaulting

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2720,6 +2721,157 @@ addEventListener('fetch', event => {});`
...
+
+					Wrote
+					{
+					  \\"name\\": \\"test-name\\",
+					  \\"compatibility_date\\": \\"2024-01-01\\",
+					  \\"assets\\": {
+					    \\"directory\\": \\"./assets\\"
+					  }
+					}
+					 to <cwd>/wrangler.json.
```
vicb: question: is json over jsonc intentional?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2720,6 +2721,157 @@ addEventListener('fetch', event => {});`
...
+					Please run wrangler deploy instead of wrangler deploy ./assets next time. Wrangler will automatically use the configuration saved to wrangler.json.
```
vicb: nit: backquote the commands for easier readability?

packages/wrangler/src/deploy/index.ts
```diff
@@ -362,3 +387,91 @@ export const deployCommand = createCommand({
 });
 
 export type DeployArgs = (typeof deployCommand)["args"];
+
+/**
+ * Handles the case where a user provides a directory as a positional argument,
+ * probably intending to deploy static assets. e.g. wrangler deploy ./public
```
vicb: Because this is called in 2 scenarios, it might be worth listing both

Also maybe add `@param args`

Consider describing the sequence fully instead of "etc."?

packages/wrangler/src/deploy/index.ts
```diff
@@ -244,8 +248,29 @@ export const deployCommand = createCommand({
...
 			config.userConfigPath && path.dirname(config.userConfigPath);
 
-		const entry = await getEntry(args, config, "deploy");
+		if (!config.configPath) {
+			// Attempt to interactively handle `wrangler deploy <directory>`
+			if (args.script) {
+				try {
+					const stats = statSync(args.script);
+					if (stats.isDirectory()) {
+						args = await handleMaybeAssetsDeployment(args);
```
vicb: Consider moving the dir check inside `handleMaybeAssetsDeployment` ?

Maybe comment on `handleMaybeAssetsDeployment` what the req are (i.e. directory)

packages/wrangler/src/deploy/index.ts
```diff
@@ -362,3 +387,91 @@ export const deployCommand = createCommand({
...
+			return args;
+		}
+	}
+
+	// Check if name is provided, if not ask for it
+	if (!args.name) {
+		const projectName = await prompt("What do you want to name your project?", {
+			defaultValue: process.cwd().split(path.sep).pop(),
+		});
+		args.name = projectName;
```
vicb, line 433: validate the name?

packages/wrangler/src/deploy/index.ts
```diff
@@ -362,3 +387,91 @@ export const deployCommand = createCommand({
...
+	}
+
+	// Set compatibility date if not provided
+	if (!args.compatibilityDate) {
+		const today = new Date();
+		const compatibilityDate = [
+			today.getFullYear(),
+			(today.getMonth() + 1).toString().padStart(2, "0"),
+			today.getDate().toString().padStart(2, "0"),
+		].join("-");
```
vicb: nit: should we have an helper for that? (I think we do that at multiple places)

packages/wrangler/src/deploy/index.ts
```diff
@@ -362,3 +387,91 @@ export const deployCommand = createCommand({
...
+			null,
+			2
+		);
+		writeFileSync(configPath, jsonString);
+		logger.log(`Wrote \n${jsonString}\n to ${chalk.bold(configPath)}.`);
+		logger.log(
+			`Please run ${chalk.bold("wrangler deploy")} instead of ${chalk.bold(`wrangler deploy ${args.assets}`)} next time. Wrangler will automatically use the configuration saved to wrangler.json.`
+		);
+	} else {
+		logger.log("Proceeding with deployment...");
```
vicb: Is this function deploying? 
If yes, only in the else branch?
(maybe clarify in the JSDoc)

vicb (approved):

packages/wrangler/src/__tests__/deploy.test.ts
vicb: Actually maybe not, looks like different chunks (colors)

packages/wrangler/src/deploy/index.ts
vicb: FYI https://github.com/cloudflare/workers-sdk/pull/10043
I also needed it for one of my PR

.changeset/deep-onions-move.md
```diff
@@ -0,0 +1,13 @@
+---
+"wrangler": minor
+---
+
+Interactively handle `wrangler deploy`s that are probably assets-only, where there is no config file and flags are incorrect or missing.
+
+For example:
+
+`npx wrangler deploy ./public` will now ask if you meant to deploy an folder of assets only, ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.
```
petebacondarwin:
```suggestion
`npx wrangler deploy ./public` will now ask if you meant to deploy a folder of assets only, ask for a name, set the compat date and then ask whether to write your choices out to `wrangler.json` for subsequent deployments.
```

.changeset/deep-onions-move.md
```diff
@@ -0,0 +1,13 @@
...
+
+`npx wrangler deploy --assets=./public` will now ask for a name, set the compat date and then ask to write your choices out to `wrangler.json` for subsequent deployments.
```
petebacondarwin:
```suggestion
`npx wrangler deploy --assets=./public` will now ask for a name, set the compat date and then ask whether to write your choices out to `wrangler.json` for subsequent deployments.
```

.changeset/deep-onions-move.md
```diff
@@ -0,0 +1,13 @@
...
+
+In non-interactive contexts, we will error as we currently do.
```
petebacondarwin:
```suggestion
In non-interactive contexts, Wrangler will error as its currently does.
```

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2720,6 +2721,157 @@ addEventListener('fetch', event => {});`
...
+
+					Total Upload: xx KiB / gzip: xx KiB
+					Worker Startup Time: 100 ms
+					Uploaded test-name (TIMINGS)
+					Deployed test-name triggers (TIMINGS)
+					  https://test-name.test-sub-domain.workers.dev
+					Current Version ID: Galaxy-Class"
+				`);
+			});
+			it("should handle `wrangler deploy --assets` without name or compat date", async () => {
```
petebacondarwin, line 2820: add blank line before `it()` clauses?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2720,6 +2721,157 @@ addEventListener('fetch', event => {});`
...
+
+		describe("should interactively handle misconfigured asset-only deployments", () => {
+			beforeEach(() => {
+				setIsTTY(true);
+
+				// Mock the date to ensure consistent compatibility_date
+				vi.setSystemTime(new Date("2024-01-01T00:00:00Z"));
+
+				fs.mkdirSync("my-site");
+				process.chdir("my-site");
```
petebacondarwin, line 2734: Can you add a comment here explaining that we are creating and entering this directory so that there is a stable default name for the Worker being deployed.

petebacondarwin: Alternatively, don't do this but just use the basename of the tmp directory that was created wherever the tests currently refer to `my-site`.

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -2720,6 +2721,157 @@ addEventListener('fetch', event => {});`
...
+				});
+				mockPrompt({
+					text: "What do you want to name your project?",
+					options: { defaultValue: "my-site" },
+					result: "test-name",
+				});
+				mockConfirm({
+					text: "Do you want Wrangler to write a wrangler.json config file to store this configuration?\nThis will allow you to simply run `wrangler deploy` on future deployments.",
+					result: true,
+				});
```
petebacondarwin, line 2767: Could we also have a couple of tests that check:
- we bail out if the user answers no to the first prompt
- we do not write out a wrangler config if the user answers no to the third prompt

packages/wrangler/src/deploy/index.ts
petebacondarwin: I think the dir check here is correct but the signature on `handleMaybeAssetsDeployment()` should be something like `(assetDirectory: string, args: DeployArgs)` to make it clear that this function expects a directory of some kind.

packages/wrangler/src/deploy/index.ts
petebacondarwin: Will it not get validated later in the normal deploy validation?

packages/wrangler/src/deploy/index.ts
```diff
@@ -362,3 +387,91 @@ export const deployCommand = createCommand({
...
+		logger.log("");
+	}
+
+	// Ask if user wants to write config file
+	const writeConfig = await confirm(
+		`Do you want Wrangler to write a wrangler.json config file to store this configuration?\n${chalk.dim("This will allow you to simply run `wrangler deploy` on future deployments.")}`
+	);
+
+	if (writeConfig) {
+		const configPath = path.join(process.cwd(), "wrangler.json");
```
petebacondarwin: We definitely encourage JSONC.

Moreover, we could actually create a JSONC file with comments in if we felt inclined.

```suggestion
		const configPath = path.join(process.cwd(), "wrangler.jsonc");
```

petebacondarwin: We use this in C3 - https://www.npmjs.com/package/comment-json

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: switching to jsonc!

packages/wrangler/src/__tests__/deploy.test.ts
emily-shen: added - its also bold fwiw, but that doesn't show up in the test snapshot

packages/wrangler/src/deploy/index.ts
emily-shen: i'll validate that the default we suggest is valid, at least.

packages/wrangler/src/deploy/index.ts
emily-shen: done, thanks!

packages/wrangler/src/deploy/index.ts
emily-shen: yes we deploy if this function completes successfully. updated the comment and moved the log out of the branch so it always prints

packages/wrangler/src/deploy/index.ts
emily-shen: updated to use jsonc. i can't really think of a useful comment to add though, so i might just leave it as is. since its assets only, linking to the full wrangler config docs might just cause more confusion.

petebacondarwin (approved):

PR #10031, opened 2025-07-21
Move the unenv tests from the preset to wrangler e2e

vicb:
Move the preset tests from the preset package to wrangler e2e.

This is done to allow removing the dependency from the preset to wrangler so that wrangler can depend on the workspace version of the preset.

The tests now run both locally and in prod.

/cc @petebacondarwin @edmundhung @jamesopstad 

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: preset change

### DIFF
--- a/packages/wrangler/e2e/tsconfig.json
+++ b/packages/wrangler/e2e/tsconfig.json
@@ -5,5 +5,6 @@
 		"module": "esnext",
 		"jsx": "preserve" // not used, but required for import of types (which transitively import jsx)
 	},
-	"include": ["vitest.config.mts", "**/*.ts", "../src/*.d.ts", "**/*.js"]
+	"include": ["vitest.config.mts", "**/*.ts", "../src/*.d.ts", "**/*.js"],
+	"exclude": []
 }


packages/unenv-preset/README.md
```diff
@@ -18,3 +18,7 @@ const { alias, inject, external, polyfill } = env;
 ```
 
 See the unenv [README](https://github.com/unjs/unenv/blob/main/README.md) for more details.
+
+## Tests
+
+This packages is tested via wrangler e2e tests.
```
petebacondarwin:
```suggestion
This package is tested via wrangler e2e tests.
```

packages/vite-plugin-cloudflare/package.json
```diff
@@ -42,7 +42,7 @@
 		"test:watch": "vitest"
 	},
 	"dependencies": {
-		"@cloudflare/unenv-preset": "2.3.3",
+		"@cloudflare/unenv-preset": "workspace:*",
```
petebacondarwin, line 45: 🎉

packages/wrangler/.eslintrc.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
 		"templates",
 		"emitted-types",
 		"e2e/seed-files/**/*",
+		"e2e/unenv-preset/**/*",
```
petebacondarwin: Why are we ignoring these tests files?
Can we not just fix up any linting problems now?

packages/wrangler/e2e/tsconfig.json
```diff
@@ -6,5 +6,5 @@
 		"jsx": "preserve" // not used, but required for import of types (which transitively import jsx)
 	},
 	"include": ["vitest.config.mts", "**/*.ts", "../src/*.d.ts"],
-	"exclude": []
+	"exclude": ["unenv-preset"]
```
petebacondarwin: Why exclude this?

packages/wrangler/e2e/unenv-preset/tsconfig.json
petebacondarwin, line 1: Is this file needed? Can't it just use the top-level one at `packages/wrangler/e2e/tsconfig.json`.

packages/wrangler/package.json
```diff
@@ -67,7 +67,7 @@
 	},
 	"dependencies": {
 		"@cloudflare/kv-asset-handler": "workspace:*",
-		"@cloudflare/unenv-preset": "2.3.3",
+		"@cloudflare/unenv-preset": "workspace:*",
```
petebacondarwin, line 70: 🎉

.changeset/gold-beans-tie.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cloudflare/vite-plugin": patch
+"@cloudflare/unenv-preset": patch
+"wrangler": patch
+---
+
+Move the unenv tests from the preset to wrangler e2e
```
petebacondarwin: I think the actual user facing change is that wrangler and vite-plugin now depend upon the latest version of unenv-preset.

(The test changes would not require a changeset)

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin, line 1: Move these `unenv-preset` seed files inside the `packages/wrangler/e2e/seed-files` directory.
Then they are automatically excluded from linting and type checking.

petebacondarwin (changes requested):

packages/wrangler/.eslintrc.js
vicb: Because eslint would complain that the files are not covered by the tsconfig.
unenv needs a different config to be able to resolve imports (`module`, `moduleResolution`), which the PR adds a distinct `tsconfig.json`.

packages/wrangler/e2e/tsconfig.json
vicb: See my previous comment, we need a different config for module resolution.

packages/wrangler/e2e/unenv-preset/tsconfig.json
vicb: See my previous comments

packages/wrangler/e2e/unenv-preset/worker/index.ts
vicb: I think having a distinct directory for the test makes this clearer
Maybe we could update the patterns to have `e2e/**/seed-files/**/*` instead but those are not seed files anyways.

Looking at the remote-binding-worker test using `seed-files`:
- it mixes using this dir and static hardcoded text string
- it copies all the files each times
-> I'll do a cleanup pass for that test in a follow up PR

packages/wrangler/.eslintrc.js
petebacondarwin: I checked out the PR and removed this line, and there were no linting errors when running `pnpm check`.

packages/wrangler/e2e/tsconfig.json
petebacondarwin: It seems that this is all caused because the Node.js test code is importing the `TESTS` object from the Worker code, just to be able to grab the keys. I assume this is to avoid duplication of this list in both the Worker and the tests.

The extra complexity around the typings and linting set up to accommodate this doesn't feel worth it.

I think we can come up with a better way that doesn't require Worker code being imported into Node.js code. A few options:

 - similar to the nodejs_compat fixture, have only one fetch request that then runs all the tests
 - add a default `/` fetch response to the Worker that returns the list of tests  that can be used to generate the individual fetch requests
 - duplicate the list of paths in the test and the Worker and just leave a comment to say that they should be kept in sync

petebacondarwin (changes requested):
I'd rather we don't mix Worker code and Node.js test in the same tsconfig set up.
The seed-files directory is already the place for this sort of thing.
I think we should have something here more along the lines of this: https://github.com/cloudflare/workers-sdk/compare/vicb/unenv-tests...pbd/unenv-tests-alt

vicb:
@petebacondarwin I have refactored the code to remove changes to the eslint config.

The changes in `tsconfig.json`s match what's usually done across the repo: exclude nest folders when they store their own `tsconfig.json`.

PTAL

vicb:
> OK let's go with this.
> Did you see, by the way, in my side branch, that you can simplify the remote tests quite a lot by using `wrangler dev --remote`? It saves having to create and delete a Worker in each test.

No I didn't.
I'll take another look and brings in stuff from there in a follow up PR.

Thanks for the reviews and feedback!

PR #10033, opened 2025-07-21
[vitest-pool-workers] Fix Workflow binding's script_name when it matches its own Worker name

pombosilva:
Fixes WOR-760.

A user can define a Workflow binding where the `script_name` is the same as the name of the Worker on which the binding is defined. Although redundant, it should be supported.

`vitest-pool-workers` breaks because it does not use the user's Worker as the test entrypoint. It uses its own runner worker, which has a different name generated before execution. This leads to a mismatch between the `script_name` in the binding and the Worker that will run.

To fix this, the code now checks it a workflow binding's  `script_name` matches its own Worker's name. When it does, `script_name` is removed, making it default to the current worker.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: bug not reproduced in fixtures for some unknown reason
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: only makes vitest-pool-workers not break in a specific case for Workflows
- Wrangler V3 Backport
  - [ ] Wrangler PR:
  - [x] Not necessary because: does not affect wrangler

### DIFF
--- a/packages/vitest-pool-workers/src/pool/index.ts
+++ b/packages/vitest-pool-workers/src/pool/index.ts
@@ -347,7 +347,7 @@ function updateWorkflowsScriptNames(
 	testWorkerName: string
 ): void {
 	const workflows = runnerWorker.workflows;
-	if (!workflows || !wranglerWorkerName) {
+	if (!workflows || wranglerWorkerName === undefined) {
 		return;
 	}
 	for (const workflow of Object.values(workflows)) {
@@ -680,7 +680,7 @@ function buildProjectMiniflareOptions(
 
 		// Set Workflows scriptName to the runner worker name if it matches the Wrangler worker name
 		const wranglerWorkerName = getWranglerWorkerName(
-			"project.options.wrangler?.configPath"
+			project.options.wrangler?.configPath
 		);
 		updateWorkflowsScriptNames(
 			runnerWorker,
--- a/packages/vitest-pool-workers/src/pool/index.ts
+++ b/packages/vitest-pool-workers/src/pool/index.ts
@@ -343,8 +343,7 @@ function getWranglerWorkerName(
 
 function updateWorkflowsScriptNames(
 	runnerWorker: WorkerOptions,
-	wranglerWorkerName: string | undefined,
-	testWorkerName: string
+	wranglerWorkerName: string | undefined
 ): void {
 	const workflows = runnerWorker.workflows;
 	if (!workflows || wranglerWorkerName === undefined) {
@@ -352,7 +351,7 @@ function updateWorkflowsScriptNames(
 	}
 	for (const workflow of Object.values(workflows)) {
 		if (workflow.scriptName === wranglerWorkerName) {
-			workflow.scriptName = testWorkerName;
+			delete workflow.scriptName;
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -328,8 +329,37 @@ function fixupDurableObjectBindingsToSelf(
...
+
+function updateWorkflowsScriptNames(
+	runnerWorker: WorkerOptions,
+	wranglerWorkerName: string | undefined,
+	testWorkerName: string
+): void {
+	const workflows = runnerWorker.workflows;
+	if (!workflows || !wranglerWorkerName) {
+		return;
+	}
```
LuisDuarte1: `!wranglerWorkerName` will be true when it's not defined and also when it's an empty string `""`, is this intended?

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -634,6 +677,17 @@ function buildProjectMiniflareOptions(
 		//  --> single instance with single runner worker
 		// Multiple Workers, Isolated Storage:
 		//  --> multiple instances each with single runner worker
+
+		// Set Workflows scriptName to the runner worker name if it matches the Wrangler worker name
+		const wranglerWorkerName = getWranglerWorkerName(
+			"project.options.wrangler?.configPath"
+		);
```
LuisDuarte1: is this supposed to be stringified?

packages/vitest-pool-workers/src/pool/index.ts
pombosilva: yes, well I guess that the worker name from Wrangler.jsonc will never be an empty string, since it is required, I can change that to only check if it is undefined

packages/vitest-pool-workers/src/pool/index.ts
pombosilva: no, will remove it

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -328,8 +329,37 @@ function fixupDurableObjectBindingsToSelf(
...
+	wranglerWorkerName: string | undefined,
+	testWorkerName: string
+): void {
+	const workflows = runnerWorker.workflows;
+	if (!workflows || wranglerWorkerName === undefined) {
+		return;
+	}
+	for (const workflow of Object.values(workflows)) {
+		if (workflow.scriptName === wranglerWorkerName) {
+			workflow.scriptName = testWorkerName;
```
petebacondarwin: Would it also be enough just to remove the `scriptName` since it would then default to the current worker?

petebacondarwin: Do we have the same problem with normal service bindings?

petebacondarwin (approved):

emily-shen:
I _think_ the name is available under `project.project.name` so you won't have to re-read the wrangler config

emily-shen:
but if Pete's suggestion to strip out the name works, that'd be great as its much simpler

petebacondarwin:
> but if Pete's suggestion to strip out the name works, that'd be great as its much simpler

I think we'd still need to know the Worker name in case it is actually pointing to a different Worker not itself?

emily-shen:
> > but if Pete's suggestion to strip out the name works, that'd be great as its much simpler
> 
> I think we'd still need to know the Worker name in case it is actually pointing to a different Worker not itself?

oh sorry fair enough - i guess check if it matches project.project.name, and strip it out if it does?

packages/vitest-pool-workers/src/pool/index.ts
pombosilva: > Would it also be enough just to remove the `scriptName` since it would then default to the current worker?

Yes, removing it works

pombosilva:
> > > but if Pete's suggestion to strip out the name works, that'd be great as its much simpler
> > 
> > 
> > I think we'd still need to know the Worker name in case it is actually pointing to a different Worker not itself?
> 
> oh sorry fair enough - i guess check if it matches project.project.name, and strip it out if it does?

Yes, I'll do it, however I'm not able to get the worker name through project.project.name. It actually returns nothing

PR #10036, opened 2025-07-22
wrangler e2e cleanup

vicb:
- Move static files to the seed-files folder
- Parallelize long running tasks
- Deflake a test (log order is not deterministic) 

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: tests
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: remote bindings

### DIFF
--- a/packages/wrangler/e2e/dev-remote-bindings.test.ts
+++ b/packages/wrangler/e2e/dev-remote-bindings.test.ts
@@ -22,35 +22,21 @@ describe.skipIf(!CLOUDFLARE_ACCOUNT_ID)(
 			await helper.seed(
 				resolve(__dirname, "./seed-files/remote-binding-workers")
 			);
-
-			await helper.seed({
-				"remote-worker.js": dedent/* javascript */ `
-					export default {
-						fetch() {
-							return new Response('Hello from a remote worker (wrangler dev mixed-mode)');
-						}
-					};
-			`,
-			});
-			await helper.run(
+			const deploy = helper.run(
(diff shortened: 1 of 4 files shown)


packages/wrangler/e2e/dev-remote-bindings.test.ts
```diff
@@ -188,15 +174,30 @@ describe.skipIf(!CLOUDFLARE_ACCOUNT_ID)(
 			);
 
 			// This should only include logs from the user Wrangler session (i.e. a single list of attached bindings, and only one ready message)
-			expect(normalizeOutput(worker.currentOutput)).toMatchInlineSnapshot(`
-			"Your Worker has access to the following bindings:
+			// Logs order are not deterministic, so we match against to possible outputs.
```
petebacondarwin, line 177: This approach feels a bit awkward. How about we check that the output contains the string `[wrangler:info] Ready on http://<HOST>:<PORT>\n` and then do a snapshot test on the log with that string removed?

petebacondarwin (approved):
Nice. Small nit.

packages/wrangler/e2e/dev-remote-bindings.test.ts
vicb: I think explicit is easier to understand and maintain, also see `			// This should only include logs from the user Wrangler session (i.e. a single list of attached bindings, and only one ready message)` l 176

PR #10038, opened 2025-07-22
Resolve dockerfile to absolute path in redirected config in vite plugin

emily-shen:
To the same purpose as https://github.com/cloudflare/workers-sdk/pull/10005, but in a different place  :)

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: cannot be tested right now because preview doesn't work with containers. But this has been tested manually.
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: fixup
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -278,11 +278,26 @@ if (import.meta.hot) {
 										const output = {
 											...container,
 											image: path.resolve(baseDir, image),
+											...(container.image_build_context
+												? {
+														image_build_context: path.resolve(
+															baseDir,
+															container.image_build_context
+														),
+													}
+												: {}),
 										};
 										// however, wrangler deploy will re-resolve the config, so we should
 										// deduplicate the container.configuration.image field in favour of
 										// the non-deprecated one when we write out the deploy config
-										delete container.configuration?.image;
+										delete output.configuration?.image;
+										// if we don't do this, we get a warning that container.configuration is deprecated
+										if (
+											output.configuration &&
+											Object.keys(output.configuration).length === 0
+										) {
+											delete output.configuration;
+										}
 										return output;
 									} else return container;
 								}


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -263,6 +264,30 @@ if (import.meta.hot) {
...
+										container.configuration?.image ?? container.image;
+									if (isDockerfile(image, workerConfig.configPath)) {
+										const output = {
+											...container,
+											image: path.resolve(baseDir, image),
+										};
+										// however, wrangler deploy will re-resolve the config, so we should
+										// deduplicate the container.configuration.image field in favour of
+										// the non-deprecated one when we write out the deploy config
+										delete container.configuration?.image;
```
jamesopstad: Could we set this to `undefined` in the `output` variable above rather than deleting here?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -263,6 +264,30 @@ if (import.meta.hot) {
...
+						if (workerConfig.containers?.length) {
+							workerConfig.containers = workerConfig.containers.map(
+								(container) => {
+									const baseDir = workerConfig.configPath
+										? path.dirname(workerConfig.configPath)
+										: resolvedViteConfig.root;
+									// Wrangler's config validation resolves to container.configuration.image, even though it is deprecated
+									const image =
+										container.configuration?.image ?? container.image;
+									if (isDockerfile(image, workerConfig.configPath)) {
```
jamesopstad, line 277: Does it matter that `workerConfig.configPath` is possible `undefined` here? Not sure if it ever can actually be `undefined` or if that's just a types thing.

packages/vite-plugin-cloudflare/src/index.ts
emily-shen: What are your concerns with deleting it? I'd rather not have two different values for `image` kicking around here, the `ContainerApp` type and validation in wrangler is quite messed up. I have a refactor PR that should clean this up soon, but in the meantime its really hard in wrangler to know whether you should use container.configuration.image or container.image, since the types are just wrong 😭

packages/vite-plugin-cloudflare/src/index.ts
emily-shen: Ah - if its possibly undefined, then we fall back on the cwd. is there a better fallback i should be using here?

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Your version is fine and it doesn't make any difference to the behaviour.

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: I think that should be OK for now. We may be able to assert the `configPath` if it's always defined but would need to investigate more.

PR #10039, opened 2025-07-22
group remote bind tests e2e in a single folder

vicb:
Mostly moving the remote tests into their own folder 

@petebacondarwin IMO grouping related tests and resources in the same folder makes sense, but you might not agree :)

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: tests
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: tests
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: remote tests


packages/wrangler/.eslintrc.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
 		"pages/functions/template-worker.ts",
 		"templates",
 		"emitted-types",
-		"e2e/seed-files/**/*",
+		"e2e/remote-binding/workers/*",
```
petebacondarwin: The downside of this is that for the unenv-preset we would end up with a `e2e/unenv-preset/workers/*` rule here as well.

So either we should group all the workers under a single top level `e2e/workers` directory and have a single rule, or perhaps change the rule to be something like `e2e/*/workers/*`?

vicb: I have updated the other PR to drop any changes to the eslint config.
Maybe a better thing to do would to remove this line altogether by adding `workers/*.js` to the tsconfig of e2e.
Happy to do that here or as a follow up if you like.

petebacondarwin: Let's do it here!

PR #10043, opened 2025-07-22
add a util to format compatibility dates

vicb:
Add a util to format a date as `YYYY-MM-DD`

It is called multiple times in the current code and +2 times in pending PRs.

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: refactoring

### DIFF
--- a/packages/wrangler/src/__tests__/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/deploy.test.ts
@@ -5353,6 +5353,14 @@ addEventListener('fetch', event => {});`
 	});
 
 	describe("workers_dev setting", () => {
+		beforeEach(() => {
+			vi.useFakeTimers();
+		});
+
+		afterEach(() => {
+			vi.useRealTimers();
+		});
+
 		it("should deploy to a workers.dev domain if workers_dev is undefined", async () => {
 			writeWranglerConfig();
 			writeWorkerSource();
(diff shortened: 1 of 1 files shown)


packages/wrangler/src/utils/compatibility-date.ts
```diff
@@ -3,6 +3,16 @@ import { configFileName } from "../config";
 import { logger } from "../logger";
 import type { Config } from "../config";
 
+/**
+ * Returns the compatibility date to use in development.
+ *
+ * When no compatibility date is configured, used the workerd date.
+ * That's the latest valid compatibility date.
```
petebacondarwin: It's not strictly the latest valid date. It is the date of the currently installed Miniflare's workerd. There maybe a new workerd version published which this version of Wrangler/Miniflare doesn't yet depend upon.

packages/wrangler/src/utils/compatibility-date.ts
```diff
@@ -13,15 +23,25 @@ export function getDevCompatibilityDate(
 	const miniflareWorkerd = miniflareRequire("workerd") as {
```
petebacondarwin, line 22: I think we now also have a dependency from Wrangler directly on workerd, so we could probably simplify this code to just get that directly.

petebacondarwin (approved):

packages/wrangler/src/utils/compatibility-date.ts
vicb: Do you have a suggestion on how to word this to be clearer?

packages/wrangler/src/utils/compatibility-date.ts
vicb: This PR only ~moves code around, we can handle that in a follow up PR maybe?

danowar2:
Thanks for the update

packages/wrangler/src/utils/compatibility-date.ts
petebacondarwin: Sure.

packages/wrangler/src/utils/compatibility-date.ts
petebacondarwin:
```suggestion
 * When no compatibility date is configured, uses the installed Workers runtime's latest supported date.
```

PR #10044, opened 2025-07-22
Dario/devx 2130/config entrypoint

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2130

add a new `/config` export that third party tool can import and from which they can get config utilities without actually bringing in wrangler in its entirety.

---

- Tests
  - [x] Tests included
- Public documentation - __TODO__
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: new wrangler feature


.changeset/pretty-dancers-decide.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+Add the following dependencies to wrangler: `@iarna/toml`, `chalk`, `cli-table3z`, `dotenv`, `find-up`, `strip-ansi`, `ts-dedent`
+
+These dependencies were `devDependencies` but they are being moves as proper `dependencies` do that third party tool can have the ability to override them when required.
```
penalosa, line 7: I don't think I understand this reasoning? They're in dev dependencies, but are bundled as part of Wrangler's build—I don't think we'd want these to be overrideable

packages/wrangler/src/r2/helpers.ts
```diff
@@ -4,6 +4,7 @@ import { Miniflare } from "miniflare";
 import prettyBytes from "pretty-bytes";
 import { fetchGraphqlResult, fetchResult } from "../cfetch";
 import { fetchR2Objects } from "../cfetch/internal";
+import { bucketFormatMessage, isValidR2BucketName } from "../config/validation";
```
penalosa, line 7: Why this change in location?

packages/wrangler/package.json
```diff
@@ -35,6 +35,20 @@
...
 	"author": "wrangler@cloudflare.com",
+	"exports": {
+		".": {
+			"types": "./wrangler-dist/cli.d.ts",
+			"require": "./wrangler-dist/cli.js",
+			"import": "./wrangler-dist/cli.js"
+		},
+		"./package.json": {
+			"default": "./package.json"
+		},
```
penalosa, line 46: Why are we exporting the package.json file?

packages/wrangler/src/utils/log-file.ts
```diff
@@ -1,12 +1,71 @@
...
 import { getGlobalWranglerConfigPath } from "../global-wrangler-config-path";
 import { logger } from "../logger";
 import { ensureDirectoryExists } from "./filesystem";
 import type { LoggerLevel } from "../logger";
+import type { Awaitable } from "miniflare";
+
+// Note: this is a copy of the same class from Miniflare as we don't want this file to directly depend
+//       on Miniflare (since this would prevent effective treesheking)
+// TODO: we should have this Mutex class defined in a shared place that both Miniflare
+//       and Wrangle can safely import from
```
penalosa, line 15: Can we no-op this file somehow when building the config utilities instead? I doubt consumers of readConfig really care about writing to a Wrangler log file

packages/wrangler/package.json
```diff
@@ -68,10 +82,17 @@
...
 		"blake3-wasm": "2.1.5",
+		"chalk": "^5.2.0",
+		"cli-table3": "^0.6.3",
+		"dotenv": "^16.3.1",
 		"esbuild": "catalog:default",
+		"find-up": "^6.3.0",
 		"miniflare": "workspace:*",
 		"path-to-regexp": "6.3.0",
+		"strip-ansi": "^7.1.0",
+		"ts-dedent": "^2.2.0",
```
penalosa, line 95: If we're moving these to deps we also need to:
1. pin to specific versions to prevent changes in those deps breaking wrangler
2. add them to the deps.ts script of non-bundled deps

That said, do we _really_ want to do this? Having non-bundled dependencies has caused us pretty bad issues in the past, and we've up until now been going in the other direction (bundling _more_ things, rather than less).

.changeset/pretty-dancers-decide.md
dario-piotrowicz: You can get more extra context here: https://github.com/nitrojs/nitro/pull/3238#issuecomment-3104912805

Basically @pi0 would potentially like to override some of these for simplified versions (or if for example they already have that dependency in nitro)

packages/wrangler/src/r2/helpers.ts
dario-piotrowicz: Because we having the `validation.ts` file importing from `r2/helpers.ts` would cause anything imported by `r2/helpers.ts` to also be bundled, which includes Miniflare and I suspect a lot more extra stuff we don't want present in the `/config` bundle.

I can also move these somewhere else instead of moving them in `config/validation.ts` but that did seem to me like a pretty correct place for these 🤔

dario-piotrowicz: TLDR; without this change this test check would fail: https://github.com/cloudflare/workers-sdk/pull/10044/files/f02f6a8cc77206265c21f96f594d1e55468dd5f3#diff-fdda3ffc6bca101d0a74cf729ddda7c07f57c9f1d51e43b8d534f518a06d0b93R25

packages/wrangler/package.json
dario-piotrowicz: Because C3 imports it in order to get the current Wrangler's version:
https://github.com/cloudflare/workers-sdk/blob/f8a80a807576f7fa6d9eca37d297c50793bca188/packages/create-cloudflare/src/git.ts#L10

Alternatively we could just export the version in the index file (like workerd for example does)

packages/wrangler/src/utils/log-file.ts
dario-piotrowicz: > Can we no-op this file somehow when building the config utilities instead?

I don't fully see how we can do that since this code is used by both wrangler and the `/config` entrypoint 🤔

packages/wrangler/package.json
dario-piotrowicz: thanks for the points 🙂 

I think we kind of need to do this to satisfy Nitro's requirements for a very lean (for more context on this also see: https://github.com/nitrojs/nitro/pull/3238#issuecomment-3104988821)

But I also see a few different potential alternatives:
 - only doing this in the tsup config for the /config entrypoint, so keeping them bundled for wrangler but external for /config
 - changing the idea of a /config entrypoint into having a new public package, in that way that package could be separate from wrangler and avoid influencing it with these external dependencies (this would likely require a bit of extra work)
 - working on making the config code less dependent on these dependencies (this might be combined with the previous point)
 
 🤔

dario-piotrowicz:
Closing since we decided not to proceed with this solution.

PR #10048, opened 2025-07-23
pass the compatibility date and flags to the unenv preset

vicb:
This PR is passing the `compatibilityDate` and `compatibilityFlags` to be cloudflare preset.

Notes:
- The preset currently not depends on the date and flags. It will in follow up PRs
- `compatibilityDate` and `compatibilityFlags` and passed from wrangler to the preset
- Constant date and time are passed from the Vite plugin. We should pass the actual values in a follow up PR. I'll create an issue with the details once this PR is approved/merged
- We have no plan to pass the date and flags for Pages at the moment.

/cc @petebacondarwin @jamesopstad @anonrig @edmundhung 

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: tested manually for now with the different wrangler command (dev, deploy, upload). This sill be e2e tested in follow up PR when the preset actually depends on date/flags
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not a user facing change
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: we do not plan to update the preset there.


packages/unenv-preset/src/index.ts
```diff
@@ -1 +1 @@
-export { cloudflare } from "./preset";
+export { getCloudflarePreset } from "./preset";
```
edmundhung, line 3: Do we expect anyone using our unenv-preset directly? This looks like a breaking change to me. How about this?

```suggestion
import { getCloudflarePreset } from "./preset";

export {
  cloudflare: getCloudflarePreset(),
  getCloudflarePreset,
};
```

.changeset/gentle-mice-create.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cloudflare/vite-plugin": patch
+"@cloudflare/unenv-preset": patch
```
edmundhung: Should this be a `minor` for `@cloudflare/unenv-preset`?

packages/unenv-preset/src/index.ts
vicb: Good catch!

.changeset/gentle-mice-create.md
vicb: Very good question.

I am on the fence here... there is indeed a new API `getCloudflareContext()` so technically it should be minor but not real new functionality. I was thinking about bumping the minor when the returned preset actually changes with the passed date/flags.

I guess we can do both: minor for this PR and minor for the PR changing the presets later.

Thoughts?

vicb: > Should this be a `minor` for `@cloudflare/unenv-preset`?

I have updated the PR to set a minor. I think it's best this way.

packages/unenv-preset/src/preset.ts
```diff
@@ -52,46 +52,63 @@ const hybridNodeCompatModules = [
...
-};
+			// To override the npm shim from unenv
+			debug: "@cloudflare/unenv-preset/npm/debug",
+		},
+		inject: {
+			// Setting symbols implemented by workerd to `false` so that `inject`s defined in base presets are not used.
+			Buffer: false as const,
+			global: false as const,
+			clearImmediate: false as const,
+			setImmediate: false as const,
```
jamesopstad: Why use `as const` here? The return type is already set on the function.

packages/vite-plugin-cloudflare/src/node-js-compat.ts
```diff
@@ -10,8 +10,14 @@ import * as vite from "vite";
 import type { WorkerConfig } from "./plugin-config";
 
 const { env } = defineEnv({
-	nodeCompat: true,
```
jamesopstad, line 13: Is this no longer needed?

vicb: exactly, it is the default for ~5 months
See https://github.com/unjs/unenv/pull/434

packages/unenv-preset/src/preset.ts
vicb: [The type for inject](https://github.com/unjs/unenv/blob/8383922e357d799e32befd9bf08cbe650590e534/src/types.ts#L52) is:

```ts
inject: Readonly<Record<string, string | readonly string[] | false>>;
```

Without `as const`, the type is inferred as `Boolean` and it trigger a TS error.

jamesopstad: Not a big deal either way but that shouldn't be the case. I can't reproduce the error.

vicb: You are right, it works without `as const`, I have added a fixup commit to clean that up.
Not sure what I did wrong earlier.

vicb:
Thanks for the reviews Edmund and James!

PR #10051, opened 2025-07-23
Add support for custom instance types

nikitassharma:
Custom instance types could previously only be set through setting vcpu/memory/disk through the `configuration` block. As we deprecate `configuration`, move setting vcpu, memory, and disk to the instance type field of wrangler. By nature, this enforces that custom instance types are mutually exclusive with the named instance types and will allow us to remove a lot of the logic that verifies this once `configuration` is removed.

Fixes CC-5692

---

- Tests
  - [X] Tests included
- Public documentation
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23674
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


.changeset/hungry-turtles-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add support for custom instance types
```
emily-shen: i vaguely remember seeing that there's a gate or something that enables this for certain accounts only? if that's the case, can you note that here?

emily-shen (changes requested):
and as i mentioned, this'll need changes to `containers/config.ts` and everything downstream of that :)

.changeset/hungry-turtles-beam.md
nikitassharma: updated to document this

packages/wrangler/src/__tests__/containers/deploy.test.ts
```diff
@@ -306,6 +306,91 @@ describe("wrangler deploy with containers", () => {
 		`);
 	});
 
+	it("should be able to deploy a new container with custom instance limits (instance_type)", async () => {
```
emily-shen, line 318: could you mention in the test above that that test is for the deprecated config that's nested under configuration? and check the snapshot of std.warn in both to make sure we warn in the deprecated version but not the new one.

.changeset/hungry-turtles-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add support for custom instance types for users without the `REQUIRE_INSTANCE_TYPE` capability
```
emily-shen: maybe something like:

Add support for custom instance limits for containers. For example, instead of having to use the preconfigured dev/standard/basic instance types, you can now set:
```
instance_type: {
vcpu: 1,
memory_mib: 1024,
disk_mb: 4000
}
```
These values still need to adhere to account limits.


does this work for anyone, or do they still need to be ENT customers? if its ent only maybe just add a line like 'This is only currently available to customers on an enterprise plan.' just so that we don't give everyone who reads the changesets false hope :')

packages/wrangler/src/containers/config.ts
emily-shen, line 1: and could you add a test to containers/config.test.ts to test that setting something like `instance_type:  {vcpu:1}` defaults the other limits to the dev values?

emily-shen (approved):
approved pending the commented changes, but don't want to block you on time zones :)

packages/wrangler/src/config/environment.ts
```diff
@@ -103,7 +103,15 @@ export type ContainerApp = {
 	 * @optional
 	 * @default "dev"
 	 */
-	instance_type?: "dev" | "basic" | "standard";
+	instance_type?:
+		| "dev"
+		| "basic"
+		| "standard"
+		| {
```
emily-shen, line 115: sorry one more thing - can you update the jsdoc comments for instance_type? also if its ENT only can you note that too in the comment.

packages/wrangler/src/config/environment.ts
```diff
@@ -103,7 +103,15 @@ export type ContainerApp = {
...
+				vcpu?: number;
```
emily-shen, line 117:
```suggestion
/** @defaults to 0.0625 (1/16 vCPU)*/
				vcpu?: number;
```
etc.

packages/wrangler/src/__tests__/containers/deploy.test.ts
nikitassharma: updated comments and added checks to `std.warn` for both

.changeset/hungry-turtles-beam.md
nikitassharma: Updated the changeset and added the line specifying it's restricted to enterprise only customers

packages/wrangler/src/containers/config.ts
nikitassharma: added along with a test for the updated path to set custom limits and added comments for the test on the deprecated path

packages/wrangler/src/config/environment.ts
nikitassharma: updated jsdoc comments for instance_type and for default values

PR #10054, opened 2025-07-24
add `worker` to Vite plugin default conditions

eltigerchino:
The Vite plugin's `resolve.conditions` setting is missing the `worker` condition that's present in the esbuild conditions used by Wrangler:

https://github.com/cloudflare/workers-sdk/blob/1e92dd815cdccfa4d48cc988f45addea356eedb9/packages/wrangler/src/deployment-bundle/bundle.ts#L70

This causes Vite to resolve to the [`browser` export in the `svelte` package](https://github.com/cloudflare/workers-sdk/blob/1e92dd815cdccfa4d48cc988f45addea356eedb9/packages/wrangler/src/deployment-bundle/bundle.ts#L70), which is probably unintended as that's meant for CSR rather than SSR.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: not sure if a test is needed since it's really dependent on the package that's being consumed. I can imagine maybe setting up a test project with a test dep that exports the worker condition?
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: I don't think these conditions are documented anywhere currently. Should they be?
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  not a Wrangler change


jamesopstad (approved):
Thanks for the PR @eltigerchino! I'll merge when it's green. CI is being quite problematic today.

By the way, you might also want to consider adding the `workerd` condition to Svelte as that's the preferred condition for Cloudflare Workers.

PR #10057, opened 2025-07-24
unquarantine the C3 next-on-pages e2e

dario-piotrowicz:
Reverting https://github.com/cloudflare/workers-sdk/pull/10056 since the issue causing C3 to create non-functioning next-on-pages applications has been fixed upstream

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: this is already tested
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: no functional changes needed documenting
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/create-cloudflare/e2e/tests/frameworks/test-config.ts
+++ b/packages/create-cloudflare/e2e/tests/frameworks/test-config.ts
@@ -372,7 +372,6 @@ function getFrameworkTestConfig(pm: string): NamedFrameworkTestConfig[] {
 			flags: ["--typescript", "--no-install", "--no-git-init"],
 		},
 		{
-			quarantine: true,
 			name: "next:pages",
 			argv: ["--platform", "pages"],
 			timeout: LONG_TIMEOUT,


packages/create-cloudflare/templates/next/pages/c3.ts
```diff
@@ -136,7 +136,7 @@ export const writeEslintrc = async (ctx: C3Context): Promise<void> => {
 
 const addDevDependencies = async (installEslintPlugin: boolean) => {
 	const packages = [
-		"@cloudflare/next-on-pages@1",
+		"@cloudflare/next-on-pages",
```
petebacondarwin: why is this needed? shouldn't `@1` still pull in the latest (e.g. 1.13.13)?

dario-piotrowicz: > shouldn't `@1` still pull in the latest (e.g. 1.13.13)?

I don't think so, this is what C3 gave me locally when I used it to create a next-on-pages app:
Screenshot 2025-07-24 at 14 53 45

petebacondarwin: And if you do `pnpm i` with that package, don't you get `1.13.13` installed?

dario-piotrowicz: I don't think so... let me give that a try....

dario-piotrowicz: It does work, ok then I've reverted the change 🙂 

Not sure why we need it in place but I am happy with either version 🙂

petebacondarwin: I think the idea was so that it didn't automatically bump up to v2 since that would be a breaking change.

PR #10058, opened 2025-07-24
fix(miniflare): service binding fetch over dev registry should work without host header

### LINKED ISSUE #9569: Calling fetch service binding requires a request object with 'host' header in vite-plugin runtime

vite: 6.3.5, wrangler: 4.19.2, @cloudflare/vite-plugin: ^1.5.1

_No response_

### Describe the Bug

You need to have

1. Two workers with a service binding between them
2. Run the worker using `vite dev` with cloudflare-vite-plugin
3. Try to call fetch using a service binding

Code in the worker 

```typescript
export default {
  async fetch(request, env) {
      await env.WORKER_API.fetch(request); <-- works
      await env.WORKER_API.fetch(new Request("http://internal/api")); <- doesn't work
  }
}
```

this code works when you try to run it using `wrangler dev`, 

I found out that if you don't have the `"host"` header specified in a custom request instance then the request fails with status: `400` and text: `"Bad Request"`.

_No response_

edmundhung:
Fixes #9569

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: v4 feature

### DIFF
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -1534,7 +1534,14 @@ export class Miniflare {
 
 		return new Promise((resolve) => {
 			const server = stoppable(
-				http.createServer(this.#handleLoopback),
+				http.createServer(
+					{
+						// There might be no HOST header when proxying a fetch request made over service binding
+						//  e.g. env.MY_WORKER.fetch("https://example.com")
+						requireHostHeader: false,
+					},
+					this.#handleLoopback
+				),
 				/* grace */ 0
 			);
 			server.on("connect", this.#handleLoopbackConnect);
--- a/packages/miniflare/src/plugins/core/errors/callsite.ts
+++ b/packages/miniflare/src/plugins/core/errors/callsite.ts
@@ -130,8 +130,8 @@ export class CallSite implements NodeJS.CallSite {
 	getMethodName(): string | null {
 		return this.opts.methodName;
 	}
-	getFileName(): string | undefined {
-		return this.opts.fileName ?? undefined;
+	getFileName(): string | null {
+		return this.opts.fileName ?? null;
 	}
 	getScriptNameOrSourceURL(): string {
 		return this.opts.fileName;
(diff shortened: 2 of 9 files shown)


packages/miniflare/src/index.ts
```diff
@@ -1534,7 +1534,14 @@ export class Miniflare {
 
 		return new Promise((resolve) => {
 			const server = stoppable(
-				http.createServer(this.#handleLoopback),
+				http.createServer(
+					{
+						// There might be no HOST header when proxying a fetch request made over service binding
+						//  e.g. env.MY_WORKER.fetch("https://example.com")
+						requireHostHeader: false,
```
edmundhung, line 1541: This option requires Node v20+ which should matches our current node requirement and available only after I updated @types/node version.

See https://github.com/DefinitelyTyped/DefinitelyTyped/commit/1b2a8609ad34d372efa0604acdfce86a8298ab42#diff-50ed3ba474b514ee7443ee72b00abc59b143f94f70bb8a0a83c3293ec21f4540R306-R312

packages/miniflare/src/plugins/core/errors/callsite.ts
```diff
@@ -130,8 +130,8 @@ export class CallSite implements NodeJS.CallSite {
 	getMethodName(): string | null {
 		return this.opts.methodName;
 	}
-	getFileName(): string | undefined {
-		return this.opts.fileName ?? undefined;
+	getFileName(): string | null {
+		return this.opts.fileName ?? null;
```
edmundhung, line 134: The type seems to be slightly changed when I updated the @types/node version

jamesopstad (approved):
Great! Glad there was a simple solution. Out of curiosity, why didn't this occur when using Wrangler?

edmundhung:
> Great! Glad there was a simple solution. Out of curiosity, why didn't this occur when using Wrangler?

It's because Wrangler pre-configure the address of the other runtime when it start up (and restart the runtime when the address changes) whereas Miniflare uses a proxy to resolves the address dynamically.

PR #10061, opened 2025-07-24
Automatically get docker socket path (and fix socket path config in vite plugin)

emily-shen:
Fixes DEVX-2103.

Uses `docker context ls` to get the socket path that the docker daemon is listening on, without having to manually set it via config or an env var. This is he

Also checks the env var DOCKER_HOST (in addition to WRANGLER_DOCKER_HOST), which is a standard.
 
Also sets this properly in the vite plugin - we were previously only picking up the value set in the wrangler config.

This has unit tests but i've also tested this manually with the vite plugin/wrangler dev and colima (which listens at a special colima socket).

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/23963
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/full-pugs-say.md
+++ b/.changeset/full-pugs-say.md
@@ -3,8 +3,8 @@
 "wrangler": patch
 ---
 
-feat: try to automatically get path of docker socket
+feat(containers): try to automatically get the socket path that the container engine is listening on.
 
-Currently, if your container tool isn't set up to listen at `unix:///var/run/docker.sock` (or isn't symlinked to that), then you have to manually set this via the `dev.containerEngine` field in your Wrangler config, or via the env vars `WRANGLER_DOCKER_HOST`. This change means that we will try and get the socket of the current context automatically. This should reduce the occurrence of opaque `internal error`s thrown by the runtime when the daemon is not listening at `unix:///var/run/docker.sock`.
+Currently, if your container engine isn't set up to listen on `unix:///var/run/docker.sock` (or isn't symlinked to that), then you have to manually set this via the `dev.containerEngine` field in your Wrangler config, or via the env vars `WRANGLER_DOCKER_HOST`. This change means that we will try and get the socket of the current context automatically. This should reduce the occurrence of opaque `internal error`s thrown by the runtime when the daemon is not listening on `unix:///var/run/docker.sock`.
 
-You can still override this with `WRANGLER_DOCKER_HOST`, and we also now read `DOCKER_HOST` too.
+In addition to `WRANGLER_DOCKER_HOST`, `DOCKER_HOST` can now also be used to set the container engine socket address.
--- a/packages/containers-shared/src/utils.ts
+++ b/packages/containers-shared/src/utils.ts
@@ -244,6 +244,9 @@ export function generateContainerBuildId() {
 	return randomUUID().slice(0, 8);
 }
 
+/**
+ * Output of docker context ls --format json
+ */
 type DockerContext = {
 	Current: boolean;
 	Description: string;
@@ -253,9 +256,10 @@ type DockerContext = {
 };
 
 /**
- * Run `docker context ls` to get the Docker socket from the currently active Docker context
+ * Run `docker context ls` to get the socket from the currently active Docker context
+ * @returns The socket path or null if we are not able to determine it
  */
-export function getDockerSocketFromContext(dockerPath: string) {
+export function getDockerSocketFromContext(dockerPath: string): string | null {
 	try {
 		const output = runDockerCmdWithOutput(dockerPath, [
 			"context",
@@ -271,15 +275,11 @@ export function getDockerSocketFromContext(dockerPath: string) {
 		// Find the current context
 		const currentContext = contexts.find((context) => context.Current === true);
 
-		if (
-			currentContext &&
-			currentContext.DockerEndpoint &&
-			typeof currentContext.DockerEndpoint === "string"
-		) {
+		if (currentContext && currentContext.DockerEndpoint) {
 			return currentContext.DockerEndpoint;
 		}
 	} catch {
-		// Fall back to null if docker context inspection fails
+		// Fall back to null if docker context inspection fails so that we can use platform defaults
 	}
 	return null;
 }
@@ -290,7 +290,7 @@ export function getDockerSocketFromContext(dockerPath: string) {
  * 3. Try to get socket from active Docker context
  * 4. Fall back to platform-specific defaults
  */
-export function resolveDockerHost(dockerPath: string) {
+export function resolveDockerHost(dockerPath: string): string {
 	if (process.env.WRANGLER_DOCKER_HOST) {
 		return process.env.WRANGLER_DOCKER_HOST;
 	}
@@ -300,12 +300,11 @@ export function resolveDockerHost(dockerPath: string) {
 	}
 
 	// 3. Try to get socket from by running `docker context ls`
-	try {
-		const contextSocket = getDockerSocketFromContext(dockerPath);
-		if (contextSocket) {
-			return contextSocket;
-		}
-	} catch {}
+
+	const contextSocket = getDockerSocketFromContext(dockerPath);
+	if (contextSocket) {
+		return contextSocket;
+	}
 
 	// 4. Fall back to platform-specific defaults
 	// (note windows doesn't work yet due to a runtime limitation)
@@ -319,12 +318,10 @@ export function resolveDockerHost(dockerPath: string) {
  * Get docker host from environment variables or platform defaults.
  * Does not use the docker context ls command, so we
  */
-export const getDockerHostFromEnv = () => {
+export const getDockerHostFromEnv = (): string => {
 	const fromEnv = process.env.WRANGLER_DOCKER_HOST ?? process.env.DOCKER_HOST;
-	if (fromEnv) {
-		return fromEnv;
-	}
-	return process.platform === "win32"
+
+	return fromEnv ?? process.platform === "win32"
 		? "//./pipe/docker_engine"
 		: "unix:///var/run/docker.sock";
 };
--- a/packages/containers-shared/tests/docker-context.test.ts
+++ b/packages/containers-shared/tests/docker-context.test.ts
@@ -1,64 +1,50 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { execFileSync } from "child_process";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { resolveDockerHost } from "../src/utils";
 
-const mockedDockerContextLsOutput = `{"Current":true,"Description":"Current DOCKER_HOST based configuration","DockerEndpoint":"unix:///current/run/docker.sock","Error":"","Name":"default"}
-{"Current":false,"Description":"Docker Desktop","DockerEndpoint":"unix:///other/run/docker.sock","Error":"","Name":"desktop-linux"}`;
-
 vi.mock("node:child_process");
-
+// We can only really run these tests on Linux, because we build our images for linux/amd64,
+// and github runners don't really support container virtualization in any sane way
 describe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
 	"resolveDockerHost",
 	() => {
-		let mockExecFileSync: ReturnType<typeof vi.fn>;
-
-		beforeEach(async () => {
-			vi.clearAllMocks();
-			const childProcess = await import("node:child_process");
-			mockExecFileSync = vi.mocked(childProcess.execFileSync);
-
-			mockExecFileSync.mockReturnValue(mockedDockerContextLsOutput);
-		});
-
 		afterEach(() => {
 			vi.unstubAllEnvs();
 		});
 
 		it("should return WRANGLER_DOCKER_HOST when set", async () => {
-			vi.stubEnv("WRANGLER_DOCKER_HOST", "unix:///foo/wrangler/socket");
-			vi.stubEnv("DOCKER_HOST", "unix:///bar/docker/socket");
+			vi.stubEnv(
+				"WRANGLER_DOCKER_HOST",
+				"unix:///FROM/WRANGLER/DOCKER/HOST/wrangler/socket"
+			);
+			vi.stubEnv("DOCKER_HOST", "unix:///FROM/DOCKER/HOST/docker/socket");
 
-			const result = await resolveDockerHost("/no/op/docker");
-			expect(result).toBe("unix:///foo/wrangler/socket");
-
-			expect(mockExecFileSync).not.toHaveBeenCalled();
+			const result = resolveDockerHost("/no/op/docker");
+			expect(result).toBe("unix:///FROM/WRANGLER/DOCKER/HOST/wrangler/socket");
 		});
 
 		it("should return DOCKER_HOST when WRANGLER_DOCKER_HOST is not set", async () => {
-			expect(process.env.WRANGLER_DOCKER_HOST).toBeUndefined();
-			vi.stubEnv("DOCKER_HOST", "unix:///bar/docker/socket");
+			vi.stubEnv("WRANGLER_DOCKER_HOST", undefined);
+			vi.stubEnv("DOCKER_HOST", "unix:///FROM/DOCKER/HOST/docker/socket");
 
-			const result = await resolveDockerHost("/no/op/docker");
-			expect(result).toBe("unix:///bar/docker/socket");
-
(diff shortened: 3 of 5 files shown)


.changeset/full-pugs-say.md
```diff
@@ -0,0 +1,10 @@
+---
+"@cloudflare/containers-shared": patch
+"wrangler": patch
+---
+
+feat: try to automatically get path of docker socket
```
vicb: A few questions:

Should "feat" imply a minor release?

Should "docker socket" be "container engine socket" (and use "container engine" below and in the other changeset rather than "docker" or "container tool")

.changeset/full-pugs-say.md
```diff
@@ -0,0 +1,10 @@
...
+
+Currently, if your container tool isn't set up to listen at `unix:///var/run/docker.sock` (or isn't symlinked to that), then you have to manually set this via the `dev.containerEngine` field in your Wrangler config, or via the env vars `WRANGLER_DOCKER_HOST`. This change means that we will try and get the socket of the current context automatically. This should reduce the occurrence of opaque `internal error`s thrown by the runtime when the daemon is not listening at `unix:///var/run/docker.sock`.
```
vicb: nit: listen at -> listen on? (multiple occurences)

.changeset/full-pugs-say.md
```diff
@@ -0,0 +1,10 @@
...
+
+You can still override this with `WRANGLER_DOCKER_HOST`, and we also now read `DOCKER_HOST` too.
```
vicb:
```suggestion
In addition to `WRANGLER_DOCKER_HOST`, `DOCKER_HOST` can now also be used to set the container engine socket address.
```

packages/containers-shared/src/utils.ts
```diff
@@ -61,22 +61,15 @@ export const runDockerCmd = (
...
-				return reject(
-					new Error(
-						`Failed running docker command: ${error.message}. Command: ${dockerPath} ${args.join(" ")}`
-					)
-				);
-			}
-			return resolve(stdout.trim());
-		});
-	});
+export const runDockerCmdWithOutput = (dockerPath: string, args: string[]) => {
```
vicb, line 64: Out of curiosity what what this changed from async to sync? (not mentioned in the PR description).

Does it deserves a comment explaining why it should not be async?

packages/containers-shared/src/utils.ts
```diff
@@ -250,3 +243,88 @@ export async function checkExposedPorts(
 export function generateContainerBuildId() {
 	return randomUUID().slice(0, 8);
 }
+
+type DockerContext = {
```
vicb, line 250: Looks like this is parsed from `docker context ls`, maybe add a comment.

Would it be worth adding some validations (zod or whatever) on top of `JSON.parse`

packages/containers-shared/src/utils.ts
```diff
@@ -250,3 +243,88 @@ export async function checkExposedPorts(
...
+	Description: string;
+	DockerEndpoint: string;
+	Error: string;
+	Name: string;
+};
+
+/**
+ * Run `docker context ls` to get the Docker socket from the currently active Docker context
+ */
+export function getDockerSocketFromContext(dockerPath: string) {
```
vicb: Maybe:
- add an explicit return type
- (and/or) add a JSDoc comment about the return value
- specify if it can throw in the JSDoc (i.e. when the socket is not found)

packages/containers-shared/src/utils.ts
```diff
@@ -250,3 +243,88 @@ export async function checkExposedPorts(
...
+		const lines = output.trim().split("\n");
+		const contexts: DockerContext[] = lines.map((line) => JSON.parse(line));
+
+		// Find the current context
+		const currentContext = contexts.find((context) => context.Current === true);
+
+		if (
+			currentContext &&
+			currentContext.DockerEndpoint &&
+			typeof currentContext.DockerEndpoint === "string"
```
vicb: The type says `DockerEndpoint: string;` so either this code is not needed or the type is wrong.

cf my comment about validation earlier

packages/containers-shared/src/utils.ts
```diff
@@ -250,3 +243,88 @@ export async function checkExposedPorts(
...
+	if (process.env.WRANGLER_DOCKER_HOST) {
+		return process.env.WRANGLER_DOCKER_HOST;
+	}
+
+	if (process.env.DOCKER_HOST) {
+		return process.env.DOCKER_HOST;
+	}
+
+	// 3. Try to get socket from by running `docker context ls`
+	try {
```
vicb: cf my earlier comment about commenting if `getDockerSocketFromContext` can throw.

I don't think it can so there should be no need for the try catch here ?

packages/containers-shared/src/utils.ts
```diff
@@ -250,3 +243,88 @@ export async function checkExposedPorts(
...
+ * Does not use the docker context ls command, so we
+ */
+export const getDockerHostFromEnv = () => {
+	const fromEnv = process.env.WRANGLER_DOCKER_HOST ?? process.env.DOCKER_HOST;
+	if (fromEnv) {
+		return fromEnv;
+	}
+	return process.platform === "win32"
+		? "//./pipe/docker_engine"
+		: "unix:///var/run/docker.sock";
```
vicb:
```suggestion
	return fromEnv ?? process.platform === "win32"
		? "//./pipe/docker_engine"
		: "unix:///var/run/docker.sock";
```

packages/containers-shared/src/utils.ts
```diff
@@ -250,3 +243,88 @@ export async function checkExposedPorts(
...
+	return null;
+}
+/**
+ * Resolve Docker host as follows:
+ * 1. Check WRANGLER_DOCKER_HOST environment variable
+ * 2. Check DOCKER_HOST environment variable
+ * 3. Try to get socket from active Docker context
+ * 4. Fall back to platform-specific defaults
+ */
+export function resolveDockerHost(dockerPath: string) {
```
vicb: It would be nice to add an explicit return type / JSDoc here and for the next function

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { resolveDockerHost } from "../src/utils";
+
+const mockedDockerContextLsOutput = `{"Current":true,"Description":"Current DOCKER_HOST based configuration","DockerEndpoint":"unix:///current/run/docker.sock","Error":"","Name":"default"}
+{"Current":false,"Description":"Docker Desktop","DockerEndpoint":"unix:///other/run/docker.sock","Error":"","Name":"desktop-linux"}`;
+
+vi.mock("node:child_process");
```
vicb: It looks like this can be move inside the next describe

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+describe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
```
vicb, line 8: `process.platform !== "linux" && process.env.CI === "true"`

If think I have already seen this condition but I can't remember why it is.

Would factoring the code into an helper function with some comment about the rationale help understanding/maintaining this code?

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
...
+	"resolveDockerHost",
+	() => {
+		let mockExecFileSync: ReturnType<typeof vi.fn>;
+
+		beforeEach(async () => {
+			vi.clearAllMocks();
+			const childProcess = await import("node:child_process");
```
vicb: Shame on me but I am not so familiar with vitest mock.

Is `vi.mock("node:child_process");` further still needed with that?
If yes, they would benefit from being closer in the source code (cf my earlier comment)

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
...
+		afterEach(() => {
+			vi.unstubAllEnvs();
+		});
+
+		it("should return WRANGLER_DOCKER_HOST when set", async () => {
+			vi.stubEnv("WRANGLER_DOCKER_HOST", "unix:///foo/wrangler/socket");
+			vi.stubEnv("DOCKER_HOST", "unix:///bar/docker/socket");
+
+			const result = await resolveDockerHost("/no/op/docker");
+			expect(result).toBe("unix:///foo/wrangler/socket");
```
vicb: 2 nits:
- When I see "foo" l31, I have to go back l27/28 to see what it is vs seeing unix:///WRANGLER_DOCKER_HOST
- using `.resolves`
 
```suggestion
			vi.stubEnv("WRANGLER_DOCKER_HOST", "unix:///WRANGLER_DOCKER_HOST");
			vi.stubEnv("DOCKER_HOST", "unix:///DOCKER_HOST");

			await expect(resolveDockerHost("/no/op/docker")).resolves.toBe("unix:///WRANGLER_DOCKER_HOST");
```

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+			expect(mockExecFileSync).not.toHaveBeenCalled();
+		});
+
+		it("should return DOCKER_HOST when WRANGLER_DOCKER_HOST is not set", async () => {
+			expect(process.env.WRANGLER_DOCKER_HOST).toBeUndefined();
```
vicb:
```suggestion
			vi.stubEnv("WRANGLER_DOCKER_HOST", undefined);
```

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
...
+		});
+
+		it("should return WRANGLER_DOCKER_HOST when set", async () => {
+			vi.stubEnv("WRANGLER_DOCKER_HOST", "unix:///foo/wrangler/socket");
+			vi.stubEnv("DOCKER_HOST", "unix:///bar/docker/socket");
+
+			const result = await resolveDockerHost("/no/op/docker");
+			expect(result).toBe("unix:///foo/wrangler/socket");
+
+			expect(mockExecFileSync).not.toHaveBeenCalled();
```
vicb: I have 0 idea of what this is testing without going back to the implementation of the function. Maybe there should be a comment on the variable or a more descriptive name.

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
...
+			const result = await resolveDockerHost("/no/op/docker");
+			expect(result).toBe("unix:///bar/docker/socket");
+
+			expect(mockExecFileSync).not.toHaveBeenCalled();
+		});
+
+		it("should use Docker context when no env vars are set", async () => {
+			const result = await resolveDockerHost("/no/op/docker");
+			expect(result).toBe("unix:///current/run/docker.sock");
+			expect(mockExecFileSync).toHaveBeenCalledWith(
```
vicb: IIUC this test depends on l4 and l19.

- It makes it hard to understand the test in isolation
- I don't think any any other depend on those values

Could both of the values be moved inside this test insstead ?

--

General comment: For test I feel like it's important to understand them in isolation. i.e. if one test fail you do not want to have to parse the whole file to understand what's going on.

Sometimes this might mean that I duplicate code in tests (i.e. I would repeat l4 for inside every single test that depend on it).

If you're curious about this approach, you can look up "DAMP vs DRY tests".

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
...
+				"/no/op/docker",
+				["context", "ls", "--format", "json"],
+				{ encoding: "utf8" }
+			);
+		});
+
+		it("should fall back to platform default on Unix when context fails", async () => {
```
vicb:
```suggestion
		it("should fall back to platform default when context fails", async () => {
```

packages/containers-shared/tests/docker-context.test.ts
```diff
@@ -0,0 +1,65 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { resolveDockerHost } from "../src/utils";
+
+const mockedDockerContextLsOutput = `{"Current":true,"Description":"Current DOCKER_HOST based configuration","DockerEndpoint":"unix:///current/run/docker.sock","Error":"","Name":"default"}
```
vicb: See my comment below (added this one after)

```suggestion
const mockedDockerContextLsOutput = `{"Current":true,"Description":"Current DOCKER_HOST based configuration","DockerEndpoint":"unix:///FROM/DOCKER/CONTEXT","Error":"","Name":"default"}
```

packages/wrangler/src/api/dev.ts
```diff
@@ -227,8 +226,8 @@ export async function unstable_dev(
 		enableIpc: options?.experimental?.enableIpc,
 		nodeCompat: undefined,
 		enableContainers: options?.experimental?.enableContainers ?? false,
-		dockerPath: options?.experimental?.dockerPath ?? getDockerPath(),
-		containerEngine: options?.experimental?.containerEngine ?? getDockerHost(),
+		dockerPath,
+		containerEngine: options?.experimental?.containerEngine,
```
vicb, line 230: It looks like the diff on the first line is only moving code around.

Is the second line diff expected ?

packages/wrangler/src/environment-variables/misc-variables.ts
```diff
@@ -273,15 +273,3 @@ export const getDockerPath = getEnvironmentVariableFactory({
 		return "docker";
 	},
 });
-
-/**
- * `WRANGLER_DOCKER_HOST` specifies the Docker socket to connect to.
- */
-export const getDockerHost = getEnvironmentVariableFactory({
-	variableName: "WRANGLER_DOCKER_HOST",
```
vicb, line 281: Technically I think you can delete `WRANGLER_DOCKER_HOST` from `type VariableNames` after this change.

But I think it is good to have all used variables there, so maybe:

- you should add a comment that while used in in misc-variables, the variable is still used to retrieve the host
- you should also add `DOCKET_HOST` there for the same reason of documenting the env vars used by wrangler.

.changeset/full-pugs-say.md
emily-shen: our general rule of thumb apparently is features on experimental products get releases as patches, so i'll probably keep this as is.

 good point on the docker specific language, i'll update the changesets

packages/containers-shared/src/utils.ts
emily-shen: the `unstable_getMiniflareWorkerOptions` api needs to use this to resolve the socket path, and that function is sync (and part of wranglers public api surface) so i wanted to avoid changing that.

packages/containers-shared/src/utils.ts
emily-shen: i don't really want to add zod just to validate this one thing, if it fails we have a sensible fallback. i mean, i would _love_ to add zod to validate very many things across wrangler, but this probably isn't the place to start 😅 

but comment added though :)

packages/containers-shared/src/utils.ts
emily-shen: removed

packages/containers-shared/src/utils.ts
emily-shen: good point, removed

packages/containers-shared/tests/docker-context.test.ts
emily-shen: a helper would be great, but the other places this is used is fixtures and wrangler, and i don't know if it would be good to be importing helpers between these three contexts?

i have added a comment here though

packages/containers-shared/tests/docker-context.test.ts
emily-shen: ah i am also not great with vitest mocks, but that makes sense and it works. thanks!

packages/containers-shared/tests/docker-context.test.ts
emily-shen: moved the mocks in, that approach makes sense to me, thanks

packages/wrangler/src/api/dev.ts
emily-shen: yep - i don't want to try and get the socket path unless we are sure there are containers configured and enableContainers is true, but i don't think we have read the config yet here so we can't do that. instead i set it later on in configController.

PR #10071, opened 2025-07-25
feat(vite-plugin): Add Containers-related info logs

CarmenPopoviciu:
Fixes DEVX-2052

Adds logs, when a Worker has Containers configured, providing information about container build status, and how to rebuild containers during local development.

`dev`

Screenshot 2025-07-29 at 14 26 59

`preview`

Screenshot 2025-07-29 at 14 30 56

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: just a log change
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: we're keeping docs lite on this sort of details
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: containers is a beta feat that only exists in v4

### DIFF
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -449,6 +449,9 @@ if (import.meta.hot) {
 					}
 
 					if (hasDevContainers) {
+						viteDevServer.config.logger.info(
+							"Building container images for local development..."
+						);
 						containerImageTagsSeen = await prepareContainerImages({
 							containersConfig: entryWorkerConfig.containers,
 							containerBuildId,
@@ -456,6 +459,9 @@ if (import.meta.hot) {
 							dockerPath,
 							configPath: entryWorkerConfig.configPath,
 						});
+						viteDevServer.config.logger.info(
+							"Containers successfully built. To rebuild your containers during development, restart the Vite dev server (r + enter)."
+						);
 
 						// poll Docker every two seconds and update the list of ids of all
 						// running containers
--- a/packages/vite-plugin-cloudflare/src/index.ts
+++ b/packages/vite-plugin-cloudflare/src/index.ts
@@ -10,6 +10,7 @@ import { generateStaticRoutingRuleMatcher } from "@cloudflare/workers-shared/ass
 import replace from "@rollup/plugin-replace";
 import MagicString from "magic-string";
 import { Miniflare } from "miniflare";
+import colors from "picocolors";
 import * as vite from "vite";
 import {
 	createModuleReference,
@@ -450,7 +451,11 @@ if (import.meta.hot) {
 
 					if (hasDevContainers) {
 						viteDevServer.config.logger.info(
-							"Building container images for local development..."
+							colors.dim(
+								colors.yellow(
+									"∷ Building container images for local development...\n"
+								)
+							)
 						);
 						containerImageTagsSeen = await prepareContainerImages({
 							containersConfig: entryWorkerConfig.containers,
@@ -460,7 +465,11 @@ if (import.meta.hot) {
 							configPath: entryWorkerConfig.configPath,
 						});
 						viteDevServer.config.logger.info(
-							"Containers successfully built. To rebuild your containers during development, restart the Vite dev server (r + enter)."
+							colors.dim(
+								colors.yellow(
+									"\n⚡️ Containers successfully built. To rebuild your containers during development, restart the Vite dev server (r + enter)."
+								)
+							)
 						);
 
 						// poll Docker every two seconds and update the list of ids of all
@@ -581,6 +590,13 @@ if (import.meta.hot) {
 				if (hasDevContainers) {
 					const dockerPath = getDockerPath();
 
+					vitePreviewServer.config.logger.info(
+						colors.dim(
+							colors.yellow(
+								"∷ Building container images for local preview...\n"
+							)
+						)
+					);
 					containerImageTagsSeen = await prepareContainerImages({
 						containersConfig: entryWorkerConfig.containers,
 						containerBuildId,
@@ -588,6 +604,9 @@ if (import.meta.hot) {
 						dockerPath,
 						configPath: entryWorkerConfig.configPath,
 					});
+					vitePreviewServer.config.logger.info(
+						colors.dim(colors.yellow("\n⚡️ Containers successfully built.\n"))
+					);
 
 					const dockerPollIntervalId = setInterval(async () => {
 						if (containerImageTagsSeen?.size) {
--- a/.changeset/afraid-experts-take.md
+++ b/.changeset/afraid-experts-take.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+feat(vite-plugin): Add Containers-related info logs
+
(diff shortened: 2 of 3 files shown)


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -497,6 +498,27 @@ if (import.meta.hot) {
...
+									description: "rebuild container(s)",
+									action: async () => {
+										if (containerImageTagsSeen?.size) {
+											await cleanupContainers(
+												dockerPath,
+												containerImageTagsSeen
+											);
+											containerImageTagsSeen.clear();
+
+											viteDevServer.restart();
```
CarmenPopoviciu: @jamesopstad technically we could also call `miniflare.setOptions()` here, but that would require us to:

1. clean up the containers
2. generate a new `containerBuildId` 
3. re-compute the miniflare options based on that new id
4. `miniflare.setOption(newOptions)`
5. `prepareContainerImages()` based on the new `containerBuildId`

Given that, I'm wondering if there is anything speaking agains restarting the dev server instead, which will do all of the above for us. Technically we're only ever doing this for Workers with Containers, and restarting the server doesn't **seem** to be compute intensive, but perhaps there are best practices in the vite community around this that I am not aware of

CarmenPopoviciu: I also know from Pete that restarting the server might cause some race conditions in our tests...so smth to keep in mind as well. We don't have automated tests for this specific hot key behaviour atm

jamesopstad: ~The advantage of not restarting the server is that `workerd` is not restarted~ (and the browser is not refreshed if the user is building a frontend). If the server restarts then all the user code needs to be reloaded into an empty cache in the module runner, which can be quite slow for large apps.

There is already a keyboard shortcut to restart the server (`r`), which is equivalent to the behaviour here, so I think it's only worth adding this feature if we can keep ~the server~ workerd alive.

(edited this comment because I incorrectly implied that calling `miniflare.setOptions` wouldn't restart workerd).

CarmenPopoviciu: ok got it! thank you for the context 🙏

CarmenPopoviciu: after discussing this internally, we agreed to rely on Vite's server restart hot key (r + enter), since that would basically achieve the same thing as we have in this PR. If there are any reports raised with us regarding consequences of this decision, we can revisit the problem.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -449,13 +449,19 @@ if (import.meta.hot) {
 					}
 
 					if (hasDevContainers) {
+						viteDevServer.config.logger.warn(
+							"Building container images for local development..."
+						);
```
```suggestion
						viteDevServer.config.logger.info(
							"Building container images for local development..."
						);
```

Unless there was a reason you chose `warn`?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -449,13 +449,19 @@ if (import.meta.hot) {
...
 						containerImageTagsSeen = await prepareContainerImages({
 							containersConfig: entryWorkerConfig.containers,
 							containerBuildId,
 							isContainersEnabled: entryWorkerConfig.dev.enable_containers,
 							dockerPath,
 							configPath: entryWorkerConfig.configPath,
 						});
+						viteDevServer.config.logger.warn(
+							"Containers succesfully built. To rebuild your containers during development, restart the Vite dev server (r + enter)."
+						);
```
```suggestion
						viteDevServer.config.logger.info(
							"Containers successfully built. To rebuild your containers during development, restart the Vite dev server (r + enter)."
						);
```

jamesopstad (commented):
I would have thought `info` is more appropriate than `warn` but interested to know your thoughts.

Should we also add similar logs to preview?

It might be nice to add some colour to the logging (we haven't been very consistent about this but it's one area to improve). I'll let you decide what's best, but you can `import colors from "picocolors";` (like [here](https://github.com/cloudflare/workers-sdk/blob/c070c7799b0edbf33610a3da68f738da7bc78405/packages/vite-plugin-cloudflare/src/build.ts#L107-L114)).

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: oh dear...I thought it was `info` 🙃 . No need for `warn`. I must have wrongly autocompleted :/

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: same

CarmenPopoviciu:
> I would have thought `info` is more appropriate than `warn` but interested to know your thoughts.
> 
> Should we also add similar logs to preview?
> 
> It might be nice to add some colour to the logging (we haven't been very consistent about this but it's one area to improve). I'll let you decide what's best, but you can `import colors from "picocolors";` (like [here](https://github.com/cloudflare/workers-sdk/blob/c070c7799b0edbf33610a3da68f738da7bc78405/packages/vite-plugin-cloudflare/src/build.ts#L107-L114)).

My apologies for the unintended `warn` use. I think I did something funky when autocompleting. Log level was meant to be `info` indeed. Thank you for spotting that!

Indeed, some color would be nice, since the message gets lost in the sea of log output. Juts wasn't sure if we have any convention. I'll have a look

I'll add something for preview as well 👍

jamesopstad:
> Indeed, some color would be nice, since the message gets lost in the sea of log output. Juts wasn't sure if we have any convention. I'll have a look

I think the convention is basically, yellow (like [here](https://github.com/cloudflare/workers-sdk/blob/c070c7799b0edbf33610a3da68f738da7bc78405/packages/vite-plugin-cloudflare/src/miniflare-options.ts#L215-L221)) 😅. Definitely room for improvement though.

CarmenPopoviciu:
@jamesopstad I addressed your comments in https://github.com/cloudflare/workers-sdk/pull/10071/commits/9b9abfdf21ef4b610263716283654c8f3e873beb and updated the screenshots in the PR description

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -449,13 +450,27 @@ if (import.meta.hot) {
...
 							containersConfig: entryWorkerConfig.containers,
 							containerBuildId,
 							isContainersEnabled: entryWorkerConfig.dev.enable_containers,
 							dockerPath,
 							configPath: entryWorkerConfig.configPath,
 						});
+						viteDevServer.config.logger.info(
+							colors.dim(
+								colors.yellow(
+									"\n⚡️ Containers successfully built. To rebuild your containers during development, restart the Vite dev server (r + enter)."
```
jamesopstad, line 470:
```suggestion
									"\n⚡️ Containers successfully built. To rebuild your containers during development, restart the Vite dev server (r + enter).\n"
```

jamesopstad (approved):
Thanks @CarmenPopoviciu!

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: for some reason, indep. of our code, `vite dev` already factors this new line in, whereas `vite preview` does not, so adding it here won't be necessary.

PR #10073, opened 2025-07-25
respect environment of services when searching them in local running wranglers

sassanh:
Shouldn't it be like this? When I write:
```json
        {
          "environment": "development",
          "binding": "some_binding",
          "service": "x",
        },
```
it should look for "x-development" not "x", right?
---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/api/integrations/platform/services.ts
+++ b/packages/wrangler/src/api/integrations/platform/services.ts
@@ -30,11 +30,14 @@ export async function getServiceBindings(
 
 	const foundServices: AvailableBindingInfo[] = [];
 	for (const { binding: bindingName, service: serviceName } of services) {
-		const worker = registeredWorkers[serviceName];
+		const serviceNameWithEnvironment = environment == null
+			? serviceName
+			: `${serviceName}-${environment}`;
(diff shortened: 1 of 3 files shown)


emily-shen:
Hiya, thanks for taking the time to put up this PR - unfortunately in this case, the `environments` key is confusingly referring to a separate and deprecated concept of service environments rather than the wrangler environments that almost everyone is using currently. 

The wrangler config schema is definitely contributing to the confusion here, so i've put up a PR to hide this from the suggestion. The correct way to bind to a worker in a specific environment would be to set `service: <worker-name>-<env-name>`.

thanks again for the pr though, if you do spot anything else weird, please feel free to keep contributing  🧡

sassanh:
Thanks for the response!
It's a very old and known anti-pattern to use string composition to encode multiple instances of information in a single string, especially in the human-facing configuration. For example, one might have these combinations:

worker-name: a-b
env-name: c

worker-name: a
env-name: b-c

PR #10099, opened 2025-07-28
Clean up containers started by local dev during process exit hook instead

emily-shen:
previously these methods were async so we could not call them from `process.on('exit')`. by using synchronous apis instead to run docker commands, we can actually do cleanup from that hook, which simplifies a lot of the vite plugin cleanup code and also ensures proper cleanup if you run and stop wrangler programmatically (e.g. using something like `concurrently`). 

Vite manual testing:
I don't think we can test this with the vite plugin because we start and stop the vite server in vitest's global setup hook, which runs after test hooks like `afterAll`.
You can manually test this by:
1. running vite dev in `container-app` in the vite playground
2. hit `/start` 
3. verify a container is running with `docker ps`
4. stop the vite dev session
5. run `docker ps` again to verify no containers are still running

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: these should just pass existing tests in fixtures/interactive-dev-tests that check that containers are cleaned up after a dev session. fine. 
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: new to v4

### DIFF
--- a/.changeset/great-terms-roll.md
+++ b/.changeset/great-terms-roll.md
@@ -3,4 +3,4 @@
 "wrangler": patch
 ---
 
-fix: move local dev container cleanup to process exit hook. this should hopefully clean up containers in more scenarios.
+fix: move local dev container cleanup to process exit hook. This should ensure containers are cleaned up even when Wrangler is shut down programatically.
--- a/.changeset/great-terms-roll.md
+++ b/.changeset/great-terms-roll.md
@@ -1,0 +1,6 @@
+---
+"@cloudflare/vite-plugin": patch
+"wrangler": patch
+---
+
+fix: move local dev container cleanup to process exit hook. this should hopefully clean up containers in more scenarios.
--- a/fixtures/interactive-dev-tests/tests/index.test.ts
+++ b/fixtures/interactive-dev-tests/tests/index.test.ts
@@ -395,8 +395,7 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
 			);
 
 			wrangler.pty.write("r");
-			await setTimeout(2000); // wait for the rebuild to start
-			// wait for build to finish
+
 			await vi.waitFor(async () => {
 				const status = await fetch(wrangler.url + "/status");
 				expect(await status.json()).toBe(false);
--- a/fixtures/interactive-dev-tests/tests/index.test.ts
+++ b/fixtures/interactive-dev-tests/tests/index.test.ts
@@ -395,7 +395,7 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
 			);
 
 			wrangler.pty.write("r");
-
+			await setTimeout(2000); // wait for the rebuild to start
 			// wait for build to finish
 			await vi.waitFor(async () => {
 				const status = await fetch(wrangler.url + "/status");
@@ -435,7 +435,7 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
 			await new Promise<void>((resolve) => {
 				wrangler.pty.onExit(() => resolve());
 			});
-			vi.waitFor(() => {
+			await vi.waitFor(() => {
 				const remainingIds = getContainerIds();
 				expect(remainingIds.length).toBe(0);
 			});
@@ -719,7 +719,7 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
 			await new Promise<void>((resolve) => {
 				wrangler.pty.onExit(() => resolve());
 			});
-			vi.waitFor(() => {
+			await vi.waitFor(() => {
 				const remainingIds = getContainerIds();
 				expect(remainingIds.length).toBe(0);
 			});
@@ -779,7 +779,7 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
 			}
 		});
 
-		it("should be interact with both workers, rebuild the containers with the hotkey and all containers should be cleaned up at the end", async () => {
+		it("should be able to interact with both workers, rebuild the containers with the hotkey and all containers should be cleaned up at the end", async () => {
 			const wrangler = await startWranglerDev([
 				"dev",
 				"-c",
--- a/packages/containers-shared/src/utils.ts
+++ b/packages/containers-shared/src/utils.ts
@@ -147,27 +147,20 @@ export const isDockerfile = (
  * If you make any changes to this fn, please make sure you persist those
  * changes in `removeContainersByIds` if necessary.
  */
-export const cleanupContainers = async (
+export const cleanupContainers = (
 	dockerPath: string,
 	imageTags: Set<string>
 ) => {
 	try {
 		// Find all containers (stopped and running) for each built image
-		const containerIds = await getContainerIdsByImageTags(
-			dockerPath,
-			imageTags
-		);
+		const containerIds = getContainerIdsByImageTags(dockerPath, imageTags);
 
 		if (containerIds.length === 0) {
 			return true;
 		}
 
 		// Workerd should have stopped all containers, but clean up any in case. Sends a sigkill.
-		await runDockerCmd(
-			dockerPath,
-			["rm", "--force", ...containerIds],
-			["inherit", "pipe", "pipe"]
-		);
+		runDockerCmdWithOutput(dockerPath, ["rm", "--force", ...containerIds]);
 		return true;
 	} catch {
 		return false;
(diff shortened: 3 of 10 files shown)


packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -191,7 +191,9 @@ export class LocalRuntimeController extends RuntimeController {
 	};
 
 	onBundleStart(_: BundleStartEvent) {
-		// Ignored in local runtime
+		process.on("exit", () => {
```
emily-shen, line 194: not really sure if this is the right spot to register this

.changeset/great-terms-roll.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cloudflare/vite-plugin": patch
+"wrangler": patch
+---
+
+fix: move local dev container cleanup to process exit hook. this should hopefully clean up containers in more scenarios.
```
jamesopstad:
```suggestion
fix: move local dev container cleanup to process exit hook. This cleans up containers in more scenarios.
```

Let's be more confident about this (even if we turn out to be wrong later) 🙂.

fixtures/interactive-dev-tests/tests/index.test.ts
```diff
@@ -395,7 +395,7 @@ baseDescribe.skipIf(process.platform !== "linux" && process.env.CI === "true")(
 			);
 
 			wrangler.pty.write("r");
-
+			await setTimeout(2000); // wait for the rebuild to start
```
jamesopstad: Why is this needed?

packages/containers-shared/src/utils.ts
```diff
@@ -147,27 +147,20 @@ export const isDockerfile = (
  * If you make any changes to this fn, please make sure you persist those
  * changes in `removeContainersByIds` if necessary.
  */
```
jamesopstad: This comment can be removed now.

```suggestion
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -457,7 +452,7 @@ if (import.meta.hot) {
 								)
 							)
 						);
-						containerImageTagsSeen = await prepareContainerImages({
+						const imageTags = await prepareContainerImages({
```
jamesopstad: Why not assign to `containerImageTagsSeen` here like before rather than iterating over `imageTags` below?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -597,7 +574,7 @@ if (import.meta.hot) {
 							)
 						)
 					);
-					containerImageTagsSeen = await prepareContainerImages({
+					const imageTags = await prepareContainerImages({
```
jamesopstad: Same question as above?

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -158,16 +158,19 @@ for (const source of imageSource) {
 				await helper.seed({
 					"wrangler.json": JSON.stringify(wranglerConfig),
 				});
+				// wait a bit in case any cleanup is already happening
+				await new Promise((resolve) => setTimeout(resolve, 500));
```
jamesopstad: Why would cleanup already be happening here?

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -158,16 +158,19 @@ for (const source of imageSource) {
...
 				if (ids.length > 0) {
-					console.log(ids);
 					execSync(`${getDockerPath()} rm -f ${ids.join(" ")}`, {
 						encoding: "utf8",
 					});
 				}
 			});
 			afterAll(async () => {
+				// wait a bit in case any cleanup is already happening
+				await new Promise((resolve) => setTimeout(resolve, 500));
```
jamesopstad: Same as above. Just curious why this is needed.

packages/wrangler/e2e/containers.dev.test.ts
```diff
@@ -209,21 +212,29 @@ for (const source of imageSource) {
...
+				await vi.waitFor(
+					async () => {
+						response = await fetch(`${ready.url}/fetch`, {
+							signal: AbortSignal.timeout(3_000),
+							headers: { "MF-Disable-Pretty-Error": "true" },
+						});
+						text = await response.text();
+						expect(text).toBe("Hello World! Have an env var! I'm an env var!");
+					},
+					{ timeout: 30_000 }
```
jamesopstad: Why do we need such a long timeout?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -381,12 +383,22 @@ export class LocalRuntimeController extends RuntimeController {
...
 			"Docker path should have been set if containers are enabled"
 		);
-		await cleanupContainers(this.dockerPath, this.containerImageTagsSeen);
+		try {
+			cleanupContainers(this.dockerPath, this.containerImageTagsSeen);
+		} catch {
+			logger.warn(
+				`Failed to clean up containers. You may have to stop and remove them up manually.`
+			);
+		}
```
jamesopstad: Do we need this error handling in the Vite plugin?

jamesopstad (commented):
Various questions, mainly so I understand better. Looks good otherwise.

packages/vite-plugin-cloudflare/src/index.ts
emily-shen: ah i thought i would have to collect the image tags across dev server reloads, but seems like i've misunderstood how things are scoped across dev servers - it looks like if i restart the dev server once, process.exit is called twice when i end the dev session on the image tags created for the respective dev sessions.

so i think it should be ok to assign it as before :)

packages/wrangler/e2e/containers.dev.test.ts
emily-shen: oops that doesn't need to be that long

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
emily-shen: i haven't seen this actually being hit but i suppose it can't hurt - added

jamesopstad: Just looked again and see that `cleanupContainers` will never throw so it would be better to remove this `try/catch` rather than adding it in the Vite plugin (especially as it needs to go in three places).

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Hmm, not sure that's the behaviour we want. On restarts the cleanup should happen in `buildEnd`. We should only cleanup the current containers in `process.on('exit')` but it looks like these listeners will currently stack up because `process.off('exit')` is never called. I would be surprised if this could still refer to old tags though.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -472,17 +467,6 @@ if (import.meta.hot) {
 							)
 						);
 
-						// poll Docker every two seconds and update the list of ids of all
-						// running containers
-						const dockerPollIntervalId = setInterval(async () => {
```
jamesopstad, line 477: Where does this polling now happen?

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Fine not to resolve the `process.on` event listeners accumulating in this PR by the way.

emily-shen: hmm is there a better place to register the `process.on` listener so there's only one?

weirdly it does look like each event listener uses the tags from when the listener is registered, which i'm a bit confused by.

(also just to check i've understood you correctly - is the current code ok now that i'm not accumulating tags across server reloads?)

packages/vite-plugin-cloudflare/src/index.ts
emily-shen: the polling no longer needs to happen since we can just get the container ids during the process exit hook, now that we've made the docker command use `execSync` instead of `exec`

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
emily-shen: oh good spot i forgot `cleanupContainers` is actually wrapped in a try catch already. i'll cleanup the wrangler ones too.

packages/wrangler/e2e/containers.dev.test.ts
emily-shen: for some reason, _locally_ the process.exit container cleanup hook never gets called when we stop wrangler in the tests. 
however, in CI, the process.exit hook does get called and we end up with a race condition where the process.exit hook container cleanup is trying to run at the same time as the test beforeEach hook and we get something like 'container is already being cleaned up' from docker :( so the timeout just tries to prevent that clash.

this is being tested in interactive-dev-fixture, and i've tested this manually, so i'm fairly confident this is just an issue with this test set up (we have apparently had similar issues before with the dev registry).

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Yes, I think it's fine as it is now. We can look at improving the `process.on` part separately.

fixtures/interactive-dev-tests/tests/index.test.ts
emily-shen: it isn't, removed

PR #10107, opened 2025-07-29
Add containerBuildId configuration to vitest-pool-workers

ghostwriternr:
Fixes "Build ID should be set if containers are defined and enabled" assertion error when testing container-enabled Durable Objects with vitest-pool-workers.

## Problem

When using `@cloudflare/vitest-pool-workers` with container-enabled Durable Objects, tests fail with:

```
AssertionError: Build ID should be set if containers are defined and enabled
```

This occurs because the vitest integration doesn't pass the required `containerBuildId` parameter to `unstable_getMiniflareWorkerOptions()`.

## Solution

This PR adds a `containerBuildId` configuration option to vitest-pool-workers, allowing users to specify the build ID that matches their pre-built Docker images.

## Usage

```typescript
// vitest.config.ts
export default defineWorkersConfig({
  test: {
    poolOptions: {
      workers: {
        wrangler: {
          configPath: "./wrangler.jsonc",
          containerBuildId: "my-test-id" // Must match docker build tag
        }
      }
    }
  }
});
```

```bash
# Build container with matching tag
docker build -t cloudflare-dev/myclass:my-test-id .
```

This approach ensures proper synchronization between test configuration and available container images, preventing build ID mismatches.

---

- Tests
  - [x] Tests not necessary because: This is a configuration fix that resolves an assertion error during test execution. The fix enables existing container tests to run rather than adding new functionality that needs testing.
- Public documentation
  - [x] Documentation not necessary because: This fixes a bug in the testing infrastructure. While it adds a new configuration option, it's for fixing broken functionality rather than new features.
- Wrangler V3 Backport
  - [x] Not necessary because: This is a patch to vitest-pool-workers testing infrastructure, not a change to wrangler itself, so no v3 back-port is required.

### DIFF
--- a/.changeset/shy-pots-cheer.md
+++ b/.changeset/shy-pots-cheer.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/vitest-pool-workers": patch
+---
+
+Fix container build ID generation when containers are present
+
+Generate container build ID when containers are defined in wrangler config, resolving "Build ID should be set if containers are defined and enabled" assertion error during testing.
--- a/packages/vitest-pool-workers/src/pool/config.ts
+++ b/packages/vitest-pool-workers/src/pool/config.ts
@@ -65,7 +65,11 @@ const WorkersPoolOptionsSchema = z.object({
 		.passthrough()
 		.optional(),
 	wrangler: z
-		.object({ configPath: z.ostring(), environment: z.ostring() })
+		.object({
+			configPath: z.ostring(),
+			environment: z.ostring(),
+			containerBuildId: z.ostring(),
+		})
 		.optional(),
 });
 export type SourcelessWorkerOptions = Omit<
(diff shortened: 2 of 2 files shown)


emily-shen:
hey thanks for putting this up! was just planning to get vitest working this week. 

I'm not sure this is the approach we want to go for though - i'd rather not expose containerBuildId. the name is a bit of a random implementation detail to id different wrangler instances, and we're probably going to change that to use a stable tag and add the dev session/build id as a label instead.

interested in your opinion of how vitest + containers should work though - would you rather we built/pulled each time on setup (like dev), or should we let users just provide an image tag that we always use?

thinking aloud here but i think we should:
1. use a stable dev image name without a new id each time (e.g. `container-name:cloudflare-dev`)
2. during dev this is rebuilt/pulled each time
3. during testing, we check if an image exists with that well-known tag, and use that if it exists, and build/pull and retag if it doesn't

lmk if you have any ideas about other useful test apis for containers :)

emily-shen:
hopefully the pre-release keeps you unblocked in the meantime though!

ghostwriternr:
> 1. use a stable dev image name without a new id each time (e.g. container-name:cloudflare-dev)
> 2. during dev this is rebuilt/pulled each time
> 3. during testing, we check if an image exists with that well-known tag, and use that if it exists, and build/pull and retag if it doesn't

I 100% agree with this being an ideal solution (or something else too!). Frankly, my exercise with doing this PR was to understand how to write tests when using containers (which I think I kinda get now) - but there's still quite a few issues. I'll try to list them out as usual and share? Doing that would help me get a better idea of useful APIs for testing containers too.

Closing this PR!

PR #10119, opened 2025-07-29
DEVX-2073: Draft of external plugin model

dxh9845:
This implements an 'external plugin' model for extending Miniflare with Workers + Services *outside* of the [Workers-SDK](https://github.com/cloudflare/workers-sdk/tree/main) repo. 

With this change, a user could specify the following configuration in an unsafe binding and have it emulated locally in a Worker simulator as existing plugins do. 

```jsonc
{
        // The name of the binding made available to this Worker
        "name": "MY_INTERNAL_BINDING",
        "service": "some-unsafe-service-binding",
        // We expect the '@cloudflare/example-external-plugin' package
        // to expose a plugin with name 'name-of-plugin-exposed-by-plugin'
        // that will provide the service binding for 'account-services-staging'
        "dev": {
          "package": "@cloudflare/example-external-plugin",
          "plugin": "name-of-plugin-exposed-by-plugin"
        },
        "type": "service",
      }
```

This package would need to expose a module with the following function export:
```
import {pluginimplementation } from 'plugins/my-plugin-implementation'
export registerMiniflarePlugins() {
   return {
       'name-of-plugin-exposed-by-plugin': pluginImplementation
   }
}
```

Notably, this can now happen **outside** of the Workers-SDK repo, which is important for plugins that may have non-public or implementation details that are confidential or restricted for internal Cloudflare usage.

### Context: 
Internally at Cloudflare, there is a desire to have a more functional local development environment when using "unsafe" bindings that aren't available for public consumption. Examples of internal bindings may include "unsafe" service bindings between internal Worker services or accessing a Quicksilver binding that will never be publicly exposed to customers but is still highly relevant to internal developers. 

This MR tries to allow internal developers at Cloudflare to write and publish Miniflare plugins *outside* of the workers-sdk repo. These plugins can then be published internally and dynamically imported and merged into Miniflare's existing plugin model. 

This MR is in rough draft state and definitely needs comments from the Workers ANT team members to help point out inconsistencies or areas for improvement. Particularly, I'd love reviews and help with:
1. Cleaning up the logic of passing unsafe bindings and validating their options in the [miniflare.ts](packages/miniflare/src/index.ts) file. I believe there's a better, more strongly typed approach here but it will require a bit of thinking.
2. ~Tests. I am slightly lost on how to best specify and mock these options in the existing Miniflare test structure.~ ✅
3. ~Opinions on the approach for specifying unsafe bindings via a `dev` section in wrangler.jsonc.~ ✅
4. ~Improving the documentation for the anatomy of a plugin. It has taken me a while to understand what each individual plugin method should be returning, even after reading [CONTRIBUTING.md](packages/miniflare/CONTRIBUTING.md). For users who struggle with the intracacies of Miniflare and how it interacts with Wrangler, better comments could go a long way in helping folks contribute plugins.~ ✅

- Tests
  - [x] Tests included - Need a hand with how to best test this!
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not for public consumption
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  new features are not backported


packages/miniflare/src/plugins/shared/index.ts
```diff
@@ -123,6 +124,102 @@ export type Plugin<
...
+ */
+export async function createPluginLoaderFromUnsafePlugin({
+	resolveWorkingDirectory,
+	packageName,
+	pluginName,
+}: UnsafePluginOption): Promise<PluginLoaderResult> {
+	try {
+		const requireRelativeToWorker = createRequire(resolveWorkingDirectory);
+		const pluginPath = requireRelativeToWorker.resolve(packageName, {
+			paths: [resolveWorkingDirectory],
```
petebacondarwin: Do you need the `paths` property if the `requireRelativeToWorker` is already based off this path?

packages/miniflare/src/plugins/shared/index.ts
```diff
@@ -123,6 +124,102 @@ export type Plugin<
...
+		});
+		const moduleURL = pathToFileURL(pluginPath).href;
+		const { registerMiniflarePlugins } = await import(moduleURL);
+		if (!registerMiniflarePlugins) {
+			return {
```
petebacondarwin: Out of interest, why return these go-style objects with error options and ok discriminators.
Generally we just allow the exception handling to deal with failures, in which case this function could just throw when it errors and return the pluginLoader when it succeeds?

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
petebacondarwin, line 1: Did the whitespace changes in here get triggered by our formatter?

packages/wrangler/src/config/environment.ts
```diff
@@ -918,11 +918,41 @@ export interface EnvironmentNonInheritable {
...
+					 * The name of the binding. provided to the Worker
+					 */
+					name: string;
+					/**
+					 * The 'type' of the unsafe binding.
+					 */
+					type: string;
+					service?: never;
+					[key: string]: unknown;
+			  }[]
```
petebacondarwin: Should this array indicator be wrapped around the union?
As it stands I think this bindings type can only be an array of one sort or the other, not a mixed array.

petebacondarwin: It would be nice to name these two types of binding and then have something like:

```
bindings?: (UnsafeBinding|UnsafeBindingWithDev)[];
```

petebacondarwin: Actually could we reuse the types you created in deployment-bundler/worker.ts?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -1,6 +1,6 @@
 import assert from "node:assert";
 import { randomUUID } from "node:crypto";
-import path from "node:path";
+import path, { resolve } from "node:path";
```
petebacondarwin: Can you just use `path.resolve`?

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -10,6 +10,21 @@ import {
...
+	type CfKvNamespace,
+	type CfPipeline,
+	type CfQueue,
+	type CfR2Bucket,
+	type CfScriptFormat,
+	type CfUnsafeBinding,
+	type CfUnsafeServiceBinding,
+	type CfWorkerInit,
+	type CfWorkflow,
+} from "../deployment-bundle/worker";
```
petebacondarwin: Why this change? Generally we put the `type` outside and then the type imports get grouped together at the bottom by our formatter.

packages/wrangler/src/dev/miniflare.ts
```diff
@@ -1423,3 +1523,26 @@ function getImageNameFromDOClassName(
...
+ * hasUnsafeBindings is a typeguard that checks whether the user has specified unsafe
+ * bindings in their Worker options
+ */
+export function hasUnsafeBindings(
+	bindings: CfWorkerInit["bindings"]
+): bindings is CfWorkerInit["bindings"] & {
+	unsafe: { bindings: CfUnsafeBinding[] };
+} {
+	const { unsafe } = bindings;
+	return !!unsafe && !!unsafe.bindings && unsafe.bindings.length > 0;
```
petebacondarwin: NIT: I think this can just be:

```suggestion
	return unsafe?.bindings?.length > 0;
```

packages/wrangler/src/dev/miniflare.ts
dxh9845: I *believe* Prettier did that but I can undo

packages/miniflare/src/plugins/shared/index.ts
dxh9845: I think I can remove, it seems to work without.

packages/miniflare/src/plugins/shared/index.ts
dxh9845: Personal preference, I can remove it if needed. Generally when I write JS I do like to explicitly handle errors and type them via discriminated unions, rather than throw, so that the user can get stronger typing on error scenarios and handle them without needing to wrap things in `try / catch`

packages/wrangler/src/dev/miniflare.ts
dxh9845: Hmm, ts is unhappy with that: `'unsafe.bindings.length' is possibly 'undefined'.ts(18048)`
image

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
dxh9845: This was after running prettier from the workspace root `pnpm run -w prettify`

packages/wrangler/src/config/environment.ts
dxh9845: I can do that - I wasn't sure if it was kosher to reuse those types

dxh9845: Actually @petebacondarwin I think it can be a mixed array - I could have an unsafe service binding AND another unsafe binding like `internal-capability-grants` that isn't of type `service`.

fixtures/unsafe-external-plugin/README.md
```diff
@@ -0,0 +1,3 @@
+# Unsafe External Plugin
+
+This folder contains an example of a Miniflare Plugin that extends Miniflare's suite of plugins with local development simulators. An example of usage can be found in the [worker-with-unsafe-exteral-plugin fixture](../worker-with-unsafe-external-plugin/).
```
petebacondarwin:
```suggestion
This folder contains an example of a Miniflare Unsafe External Plugin that extends Miniflare's suite of plugins with local development simulators. An example of usage can be found in the [worker-with-unsafe-exteral-plugin fixture](../worker-with-unsafe-external-plugin/).
```

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+	getMiniflareObjectBindings,
+	getPersistPath,
+	Plugin,
+	ProxyNodeBinding,
+	SERVICE_LOOPBACK,
+	SharedBindings,
+} from "miniflare";
+import { z } from "miniflare-shared";
+import BINDING_WORKER from "worker:binding.worker";
+import OBJECT_WORKER from "worker:object.worker";
```
petebacondarwin: Worth adding a comment here to explain that these special imports are handled by the bundler to all us to import the content of a Worker source.

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+import type { Service, Worker_Binding } from "miniflare";
+
+export const UNSAFE_PLUGIN_NAME = "unsafe-plugin";
+
+/**
+ * Options for the unsafe plugin. It takes a map of binding names to options for that specific binding
+ */
+export const UnsafeServiceBindingOption = z.record(
```
petebacondarwin: Maybe rename to
```suggestion
export const UnsafeServiceBindingOptionSchema = z.record(
```
then you don't get the type/instance collision further down.

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+	typeof UnsafeServiceBindingSharedOptions;
+
+export const UNSAFE_SERVICE_PLUGIN: Plugin<
+	typeof UnsafeServiceBindingOption,
+	typeof UnsafeServiceBindingSharedOptions
+> = {
+	options: UnsafeServiceBindingOption,
+	sharedOptions: UnsafeServiceBindingSharedOptions,
+	/**
+	 * getBindings will add bindings to the user's Workers. Specifically, we add a service
```
petebacondarwin:
```suggestion
	 * getBindings will add bindings to the user's Workers. Specifically, we add a binding to a service
```

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+	 * that will expose an `UnsafeBindingServiceEntrypoint`
+	 * @param options
```
petebacondarwin: Comment on where this `options` object comes from? I think it is the object passed along to the unsafe binding itself?

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+			return {
+				name: bindingName,
+				service: {
+					name: `${UNSAFE_PLUGIN_NAME}:${bindingName}`,
+					entrypoint: "UnsafeBindingServiceEntrypoint",
+				},
+			};
+		});
+	},
+	getNodeBindings(options: z.infer<typeof UnsafeServiceBindingOption>) {
```
petebacondarwin: why does this need a type when the `getBindings()` param doesn't?

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+import OBJECT_WORKER from "worker:object.worker";
+import type { Service, Worker_Binding } from "miniflare";
+
+export const UNSAFE_PLUGIN_NAME = "unsafe-plugin";
+
+/**
+ * Options for the unsafe plugin. It takes a map of binding names to options for that specific binding
+ */
+export const UnsafeServiceBindingOption = z.record(
+	z.string(),
```
petebacondarwin: if `accountConverters` is a special field then why not include it here?

packages/wrangler/src/config/environment.ts
dxh9845: @petebacondarwin Unless I'm missing something, I'm going to resolve this comment. Feel free to reopen if you disagree.

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+	 * @returns
+	 */
+	async getBindings(options) {
+		const configOptions = Object.entries(options);
+
+		// If the user hasn't specified pre-determined mappings, we will skip adding any services
+		if (!configOptions.length) {
+			return [];
+		}
+
```
petebacondarwin: Not needed as the `configOptions.map()` will result in an empty array anyway.

```suggestion
```

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
```diff
@@ -0,0 +1,163 @@
...
+				name: bindingName,
+				service: {
+					name: `${UNSAFE_PLUGIN_NAME}:${bindingName}`,
+					entrypoint: "UnsafeBindingServiceEntrypoint",
+				},
+			};
+		});
+	},
+	getNodeBindings(options: z.infer<typeof UnsafeServiceBindingOption>) {
+		if (!options?.accountConverters) {
```
petebacondarwin: What is `accountConverters`? It seems strange that this is here.

petebacondarwin (commented):
Just a few comments on the fixture...

fixtures/unsafe-external-plugin/src/plugins/unsafeService.ts
dxh9845: This was copy paste from a different plugin - will remove

fixtures/unsafe-external-plugin/src/plugins/unsafe-service.ts
```diff
@@ -0,0 +1,159 @@
...
+);
+export type UnsafeServiceBindingOption =
+	typeof UnsafeServiceBindingOptionSchema;
+
+export const UnsafeServiceBindingSharedOptions = z.undefined();
+export type UnsafeServiceBindingSharedOption =
+	typeof UnsafeServiceBindingSharedOptions;
+
+export const UNSAFE_SERVICE_PLUGIN: Plugin<
+	typeof UnsafeServiceBindingOptionSchema,
```
petebacondarwin:
```suggestion
	UnsafeServiceBindingOption,
```

fixtures/unsafe-external-plugin/src/plugins/unsafe-service.ts
```diff
@@ -0,0 +1,159 @@
...
+	options: UnsafeServiceBindingOptionSchema,
+	sharedOptions: UnsafeServiceBindingSharedOptions,
+	/**
+	 * getBindings will add bindings to the user's Workers. Specifically, we add a binding to a service
+	 * that will expose an `UnsafeBindingServiceEntrypoint`
+	 * @param options - A map of bindings names to options provided for that binding.
+	 * @returns
+	 */
+	async getBindings(options: z.infer<typeof UnsafeServiceBindingOptionSchema>) {
+		const configOptions = Object.entries(options);
```
petebacondarwin: Isn't the `options` object that is passed is actually the entire options given to Miniflare?

If so then it can contain lots of other config destined for other plugins.

Shouldn't the plugin extract the bits that are relevant only to this plugin by putting them in a well-defined section of the config. E.g. `unsafe-plugin`.

fixtures/unsafe-external-plugin/src/plugins/worker-shim.ts
petebacondarwin, line 1: I wonder if we could add this to the typings of the miniflare-shared package?
Not for this PR.

fixtures/unsafe-external-plugin/src/plugins/unsafe-service.ts
```diff
@@ -0,0 +1,159 @@
...
+						modules: [
+							{
+								name: "binding.worker.js",
+								esModule: BINDING_WORKER(),
+							},
+						],
+						bindings: [
+							{
+								name: "config",
+								json: JSON.stringify(config),
```
petebacondarwin: So currently the only way to pass in values to the internal services is via the well-known binding types.

fixtures/worker-with-unsafe-external-plugin/package.json
```diff
@@ -0,0 +1,14 @@
+{
+	"name": "@fixture/worker-with-unsafe-external-plugin",
+	"private": true,
+	"scripts": {
+		"dev": "wrangler dev"
+	},
+	"devDependencies": {
+		"@cloudflare/workers-tsconfig": "workspace:*",
+		"@fixture/unsafe-external-plugin": "workspace:*",
+		"miniflare-shared": "workspace:*",
```
petebacondarwin: Is this needed for the Worker?

fixtures/worker-with-unsafe-external-plugin/src/index.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Welcome to Cloudflare Workers! This is your first worker.
+ *
+ * - Run `wrangler dev src/index.ts` in your terminal to start a development server
+ * - Open a browser tab at http://localhost:8787/ to see your worker in action
+ * - Run `wrangler deploy src/index.ts --name my-worker` to deploy your worker
+ *
+ * Learn more at https://developers.cloudflare.com/workers/
+ */
+
```
petebacondarwin, line 10: Delete?

fixtures/worker-with-unsafe-external-plugin/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+	// MY_KV_NAMESPACE: KVNamespace;
+	//
+	// Example binding to Durable Object. Learn more at https://developers.cloudflare.com/workers/runtime-apis/durable-objects/
+	// MY_DURABLE_OBJECT: DurableObjectNamespace;
+	//
+	// Example binding to R2. Learn more at https://developers.cloudflare.com/workers/runtime-apis/r2/
+	// MY_BUCKET: R2Bucket;
+	//
+	// Example binding to a Service. Learn more at https://developers.cloudflare.com/workers/runtime-apis/service-bindings/
+	// MY_SERVICE: Fetcher;
```
petebacondarwin, line 30: Delete?

packages/miniflare/src/index.ts
```diff
@@ -1404,6 +1575,24 @@ export class Miniflare {
...
+			string,
+			PluginBase<z.ZodType<any>, z.ZodType<any>>
+		> = {};
+		for (const [pluginName, loadExternalPlugin] of Object.entries(
+			this.#pluginLoaders
+		)) {
+			this.#log.info(`Loading external plugin ${pluginName}`);
+			const loadedPlugin = await loadExternalPlugin();
+			const plugin = loadedPlugin.registerMiniflarePlugins();
+			plugins = { ...plugin };
```
petebacondarwin:
```suggestion
			plugins = { ...plugins, ...plugin };
```

???

packages/miniflare/src/index.ts
```diff
@@ -1412,6 +1601,12 @@ export class Miniflare {
 		const allWorkerOpts = this.#workerOpts;
 		const sharedOpts = this.#sharedOpts;
 
+		const plugins = await this.#loadExternalPlugins();
+		this.#externalPlugins = Object.entries(plugins).map(([name, plugin]) => [
+			name,
+			plugin,
+		]);
```
petebacondarwin:
```suggestion
		this.#externalPlugins = [...Object.entries(plugins)];
```
???

packages/wrangler/src/config/environment.ts
```diff
@@ -40,6 +40,53 @@ export type CloudchamberConfig = {
...
+type UnsafeServiceBinding = {
+	/**
+	 * The name of the unsafe binding provided to the Worker
+	 */
+	name: string;
+	/**
+	 * The 'type' of the unsafe binding being provided to the Worker. This must be "service" to use
+	 * local development features available exclusively to unsafe service bindings.
+	 */
+	type: "service";
```
petebacondarwin: we need to be sure there are no other unsafe bindings out there that have this type.

packages/wrangler/src/config/environment.ts
```diff
@@ -40,6 +40,53 @@ export type CloudchamberConfig = {
...
 
+type UnsafeBinding = {
+	/**
+	 * The name of the binding provided to the Worker
+	 */
+	name: string;
+	/**
+	 * The 'type' of the unsafe binding.
+	 */
+	type: string;
```
petebacondarwin, line 51:
```suggestion
	type: Exclude<string, "service">;
```

Is this actually possible?

fixtures/worker-with-unsafe-external-plugin/wrangler.jsonc
```diff
@@ -0,0 +1,24 @@
...
+			{
+				"name": "UNSAFE_SERVICE_BINDING",
+				"dev": {
+					"package": "@fixture/unsafe-external-plugin",
+					"plugin": "unsafe-plugin",
+					"pluginOptions": {
+						"UNSAFE_SERVICE_BINDING": {
+							"emitLogs": true,
+						},
+					},
```
petebacondarwin: should this just be:

```suggestion
					"pluginOptions": {
							"emitLogs": true,
					},
```

fixtures/worker-with-unsafe-external-plugin/wrangler.jsonc
```diff
@@ -0,0 +1,24 @@
...
+				},
+				"type": "service",
+				"service": "some-unsafe-service",
+			},
```
petebacondarwin, line 21: would there be additional "production" config here that should be merged with the dev config above?

fixtures/worker-with-unsafe-external-plugin/package.json
```diff
@@ -0,0 +1,14 @@
+{
+	"name": "@fixture/worker-with-unsafe-external-plugin",
+	"private": true,
+	"scripts": {
+		"dev": "wrangler dev"
```
petebacondarwin: We should have tests here?

fixtures/unsafe-external-plugin/src/plugins/unsafe-service.ts
dxh9845: It is not, it is the options as passed to `pluginOptions` in the unsafe options part of the wrangler config. That being said if you have a preferred approach about making the typing / option validating better I'm open for it. However I don't have much more bandwidth to do so in this PR so perhaps we can do that in a followup?
image

packages/wrangler/src/config/environment.ts
dxh9845: I had tried this actually but it didn't seem to help anything so I removed it
```
let a: UnsafeBinding =  {
	type: "service",
	name: "unsafebinding",
}
```

image

fixtures/worker-with-unsafe-external-plugin/wrangler.jsonc
dxh9845: TBH I had trouble with the options parsing part of miniflare (as every option comes in "unknown" and needs to be reparsed out via `validateOptions`). The key thing we need here is to have the options apply *per binding*. If you or @penalosa have an alternative approach that would be appreciated.

packages/wrangler/src/config/environment.ts
dxh9845: I will follow up in chat about this.

fixtures/worker-with-unsafe-external-plugin/wrangler.jsonc
```diff
@@ -0,0 +1,24 @@
...
+	"compatibility_date": "2025-07-30",
+	"unsafe": {
+		"bindings": [
+			{
+				"name": "UNSAFE_SERVICE_BINDING",
+				"dev": {
+					"package": "@fixture/unsafe-external-plugin",
+					"plugin": "unsafe-plugin",
+					"pluginOptions": {
+						"UNSAFE_SERVICE_BINDING": {
```
penalosa: @penalosa remove this duplication

penalosa: Also figuring out Vitest not registering plugins

packages/wrangler/src/deployment-bundle/worker.ts
```diff
@@ -286,64 +286,31 @@ export interface CfPipeline {
...
-	/**
-	 * dev is an optional field specifying options to emulate this service binding
-	 * locally via Miniflare and Wrangler
-	 */
+	type: string;
+
 	dev?: {
+		plugin: {
+			/**
+			 * Package is the bare specifier of the package that exposes plugins to integrate into Miniflare via a `registerMiniflarePlugins` function.
```
dxh9845: Seems like this will need to change

packages/miniflare/src/plugins/shared/index.ts
```diff
@@ -97,31 +96,13 @@ export interface PluginBase<
 	SharedOptions extends z.ZodType | undefined,
 > {
 	options: Options;
-	/**
-	 * getBindings returns a list of bindings that will be made available
-	 * to the "entry" worker.
-	 * @param options
-	 * @param workerIndex
-	 */
```
dxh9845: If possible, can we keep these comments (and correct them if they're not accurate?) I want some form of documentation for plugin developers to know how to write their plugins.

PR #10122, opened 2025-07-29
fix `startWorker` not respecting `auth` options for remote bindings

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-1857

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this is fixing expected behavior
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: related to remote bindings which are not a v3 feature

### DIFF
--- a/packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
+++ b/packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
@@ -11,7 +11,6 @@ import { Miniflare, Mutex } from "miniflare";
 import * as MF from "../../dev/miniflare";
 import { getDockerPath } from "../../environment-variables/misc-variables";
 import { logger } from "../../logger";
-import { pickRemoteBindings } from "../remoteBindings";
 import { RuntimeController } from "./BaseController";
 import { castErrorCause } from "./events";
 import {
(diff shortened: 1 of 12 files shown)


penalosa (approved):

packages/wrangler/src/api/startDevWorker/LocalRuntimeController.ts
```diff
@@ -211,6 +212,8 @@ export class LocalRuntimeController extends RuntimeController {
 					"../remoteBindings"
 				);
 
+				const auth = await unwrapHook(data.config.dev.auth);
```
penalosa: This needs to be wrapped in a check that some bindings are actually remote, otherwise local-only sessions would get an auth prompt

dario-piotrowicz: yeah, nice spot! thanks I'll address this 🙏

dario-piotrowicz: fixed 🙂

PR #10136, opened 2025-07-30
faster runtime for containers images list

nikitassharma:
Fixes CC-4855

Listing images was getting noticeably slower with more images as we make 1 API call to get repos and N API calls to get tags. Refactoring so that we only make 1 API call to get images and tags to minimize the delay on `containers images list`.

Also contains some reorganizing with the test file for container image commands.

---

- Tests
  - [X] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: performance improvement
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


flakey5 (approved):

packages/wrangler/src/cloudchamber/images/images.ts
```diff
@@ -184,25 +183,12 @@ async function listImages(
...
-
-async function listRepos(creds: string): Promise<string[]> {
+async function listReposWithTags(
+	creds: string
+): Promise<Record<string, string[]>> {
 	const url = new URL(`https://${getCloudflareContainerRegistry()}`);
-
-	const catalogUrl = `${url.protocol}//${url.host}/v2/_catalog`;
+	const query = "tags=true";
+	const catalogUrl = `${url.protocol}//${url.host}/v2/_catalog?${query}`;
```
emily-shen:
```suggestion
	const catalogUrl = `${url.protocol}//${url.host}/v2/_catalog?tags=true`;
```

PR #10140, opened 2025-07-30
refactor: add `page.goto()` back into beforeEach

petebacondarwin:
Will remove gotos from individual tests if this doesn't fail CI.


petebacondarwin:
@jamesopstad and I agreed not to land this. It is nice to be explicit in the tests if you need a step to have moved the browser to the starting page.

PR #10150, opened 2025-07-31
refactor http overrides

vicb:
The scope of this PR was changed to only include the refactoring to the preset/tests.

http server APIs (and other node modules) will be submitted as follow up PRs.

--

Former description:

The server APIs have been implemented in https://github.com/cloudflare/workerd/pull/4591 and released in 20250730

The are behing the "enable_nodejs_http_server_modules" and "experimental" compat flag.

Note that the tests will be failing until the APIs reach prod workerd.

/cc @anonrig @petebacondarwin 

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Should be documented by the runtime team
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: unenv is not backported

### DIFF
--- a/packages/unenv-preset/src/preset.ts
+++ b/packages/unenv-preset/src/preset.ts
@@ -9,7 +9,7 @@ import type { Preset } from "unenv";
 // Last checked: 2025-01-24
 //
 // NOTE: Please sync any changes to `testNodeCompatModules`.
-const nodeCompatModules = [
+const nativeModules = [
 	"_stream_duplex",
 	"_stream_passthrough",
 	"_stream_readable",
@@ -42,7 +42,7 @@ const nodeCompatModules = [
 ];
 
 // Modules implemented via a mix of workerd APIs and polyfills.
-const hybridNodeCompatModules = [
+const hybridModules = [
 	"async_hooks",
 	"console",
 	"crypto",
@@ -66,6 +66,23 @@ export function getCloudflarePreset({
 	compatibilityDate?: string;
 	compatibilityFlags?: string[];
 }): Preset {
+	const httpOverrides = getHttpOverrides({
+		compatibilityDate,
+		compatibilityFlags,
+	});
+
+	// "dynamic" as they depend on the compatibility date and flags
+	const dynamicNativeModules = [
+		...nativeModules,
+		...httpOverrides.nativeModules,
+	];
+
+	// "dynamic" as they depend on the compatibility date and flags
+	const dynamicHybridModules = [
+		...hybridModules,
+		...httpOverrides.hybridModules,
+	];
+
 	return {
 		meta: {
 			name: "unenv:cloudflare",
@@ -76,7 +93,7 @@ export function getCloudflarePreset({
 			// `nodeCompatModules` are implemented in workerd.
 			// Create aliases to override polyfills defined in based environments.
 			...Object.fromEntries(
-				nodeCompatModules.flatMap((p) => [
+				dynamicNativeModules.flatMap((p) => [
 					[p, p],
 					[`node:${p}`, `node:${p}`],
 				])
@@ -88,14 +105,11 @@ export function getCloudflarePreset({
 
 			// `hybridNodeCompatModules` are implemented by the cloudflare preset.
 			...Object.fromEntries(
-				hybridNodeCompatModules.flatMap((m) => [
+				dynamicHybridModules.flatMap((m) => [
 					[m, `@cloudflare/unenv-preset/node/${m}`],
 					[`node:${m}`, `@cloudflare/unenv-preset/node/${m}`],
 				])
 			),
-
-			// Use either the unenv or native implementation
-			...getHttpAliases({ compatibilityDate, compatibilityFlags }),
 
 			// To override the npm shim from unenv
 			debug: "@cloudflare/unenv-preset/npm/debug",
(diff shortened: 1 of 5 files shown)


packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -1,23 +1,5 @@
 import assert from "node:assert";
 
-// List all the test functions.
```
vicb, line 3: refactored to not have to duplicate the test names here.

All the test are now in an obj, l. 61

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -74,214 +56,273 @@ function generateTestListResponse(testName: string): Response {
 	);
 }
 
-async function testCryptoGetRandomValues() {
-	const crypto = await import("node:crypto");
+// Test functions executed on worked.
+// The test can be executing by fetching the `/${testName}` url.
+export const WorkerdTests: Record<string, () => void> = {
```
vicb, line 61: Mostly moving the code.

The only 2 new functions are the last 2 `testHttpServer` and `testHttpsServer`.

Other functions have only been moved as properties of `WorkerdTest`

packages/unenv-preset/src/preset.ts
```diff
@@ -110,64 +124,59 @@ export function getCloudflarePreset({
...
+	// Override unenv base aliases with native and hybrid modules
+	// `node:https` is fully implemented by workerd
+	return {
+		nativeModules: [
+			"_http_agent",
+			"_http_client",
+			"_http_common",
+			"_http_incoming",
+			"_http_outgoing",
+			...(httpServerEnabled ? ["_http_server", "https"] : []),
```
anonrig: HTTPS is enabled already. Server doesn't enable https module.

vicb: When `httpServerEnabled === false` then we need to polyfill the server API of https.
That is some of the APIs come from workerd and the rest from uennv.
It becomes and hybrid module as you can see 2 lines down.

When `httpServerEnabled === true` all the APIs of https are implemented natively and we can use the native module without any polyfill.

vicb: To add to that, there are always APIs we polyfill in http (w/o "s") regardless the native http module is available.
That is why http is always an hybrid module.

anonrig (approved):

packages/unenv-preset/src/runtime/node/http.ts
```diff
@@ -41,6 +36,20 @@ export const {
...
 } = workerdHttp;
 
+// Use the workerd implementation of server APIs when the
+// `enable_nodejs_http_server_modules` flag is on.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isWorkerdServerEnabled = (globalThis as any).Cloudflare.compatibilityFlags
+	.enable_nodejs_http_server_modules;
+
+export const createServer = isWorkerdServerEnabled
+	? workerdHttp.createServer
```
petebacondarwin: I assume that this is not `undefined` even if the flag is off?
Otherwise you could do more simply:

```ts
export const createServer = workerdHttp.createServer ?? unenvCreateServer;
```

packages/unenv-preset/src/runtime/node/http.ts
```diff
@@ -1,26 +1,21 @@
...
 	CloseEvent,
-	createServer,
 	maxHeaderSize,
 	MessageEvent,
-	Server,
-	ServerResponse,
 	setMaxIdleHTTPParsers,
+	createServer as unenvCreateServer,
+	Server as unenvServer,
+	ServerResponse as unenvServerResponse,
```
petebacondarwin: Does this mean that we still pull in these unenv polyfills even if the native flag is turned on?
I don't think that our bundlers will be clever enough to tree shake this away since it is a runtime check.
How big are these? Could we have a build time way to do this?

packages/wrangler/e2e/unenv-preset/preset.test.ts
```diff
@@ -54,6 +55,19 @@ const testConfigs: TestConfig[] = [
 				enable_nodejs_http_modules: true,
 			},
 		},
+		{
+			name: "http server enabled by flag",
+			compatibilityDate: "2025-07-26",
+			compatibilityFlags: [
+				"enable_nodejs_http_modules",
+				"enable_nodejs_http_server_modules",
+				"experimental",
```
petebacondarwin: AFAIK you cannot actually send the `experimental` flag to production workers. It is not allowed in edgeworker. So I assume that this will need to change once we get the server_modules flags into production.

petebacondarwin (approved):
A few questions but I don't think they block, given that our tests will have to pass on production before can land this.

petebacondarwin:
Can we check that the tests actually fail on CI - the current failure is an unrelated flake.

packages/unenv-preset/src/runtime/node/http.ts
vicb: Maybe that would work but I find the explicit check of `enable_nodejs_http_server_modules /isWorkerdServerEnabled` to be more self-documenting.

packages/unenv-preset/src/runtime/node/http.ts
vicb: Good question.

Yes that will pull [`notImplemented`](https://github.com/unjs/unenv/blob/f9934544231e360a68f1051debb76608659e5cea/src/runtime/_internal/utils.ts#L35) and [`notImplementedClass`](https://github.com/unjs/unenv/blob/f9934544231e360a68f1051debb76608659e5cea/src/runtime/_internal/utils.ts#L59C17-L59C36) - a total of 10 lines of code that are pulled by other polyfills anyway.

I went the runtime way because I think it's simpler.

The build time solution would require 2 hybrid polyfills `http-no-server.ts` and `http-yes-server.ts`, updating the mechanism to load hybrid modules as the name is currently assumed to be the module name. So more code, more changes for no benefit.

packages/wrangler/e2e/unenv-preset/preset.test.ts
vicb: TIL, then we need to skip experimental remote test.

petebacondarwin: FWIW the `experimental` flag is always passed to Miniflare's workerd instances because it is needed to enable the loopback server and other things. So it might just be fine to remove it here altogether.

vicb:
> Can we check that the tests actually fail on CI - the current failure is an unrelated flake.

See https://github.com/cloudflare/workers-sdk/pull/10174 (I'll rebase on top of that when it's merged)

packages/wrangler/e2e/unenv-preset/preset.test.ts
vicb: > FWIW the `experimental` flag is always passed to Miniflare's workerd instances because it is needed to enable the loopback server and other things. So it might just be fine to remove it here altogether.

If we remove `experimental` then we don't have a way to skip the test on remote. So I would prefer to keep it there.

That way users can test experimental feature locally - they will get an error when tying to deploy because of the `experimental` flag.

WDYT?

petebacondarwin: I don't mind either way, since before we land this PR the experimental flag will be unnecessary.

vicb: We will skip experimental tests on remote.

packages/unenv-preset/src/runtime/node/https.ts
```diff
@@ -1,8 +1,8 @@
+// This hybrid polyfill is used only when the native implementation is not available
```
anonrig: Can you expand this comment here and explaining how and why the native implementation is not available.

PR #10153, opened 2025-07-31
draft: Add support for the new hosted method in the images binding

mglewis:
Fixes [IMAGES-1704](https://jira.cfdata.org/browse/IMAGES-1704).

This PR is the counterpart for the [workerd change](https://github.com/mglewis/workerd/pull/1) that introduces the hosted method to the images binding. The hosted method allows users to fetch images stored in the Cloudflare Images product.

The changes here support two modes of operation:

- *remote*: running `wrangler dev` will connect to the Cloudflare Images API
- *local*: running `wrangler dev --experimental-images-local-mode` will return return placeholder 200px x 200px coloured squares.

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/miniflare/src/workers/images/images.worker.ts
```diff
@@ -0,0 +1,221 @@
+// KV-backed mock for Images binding CRUD operations
+// Image data is stored as KV values, metadata as KV metadata
+// Transforms and info operations are handled via HTTP loopback to Node.js Sharp
+
+import { WorkerEntrypoint } from "cloudflare:workers";
+import { CoreBindings, CoreHeaders } from "../core/constants";
+
+interface ImageMetadata {
```
emily-shen: can you import these from workers-types? (if its not in there yet i'll bump workerd/types later today)

packages/miniflare/src/workers/images/images.worker.ts
```diff
@@ -0,0 +1,221 @@
...
+			filename: options?.filename ?? "uploaded.jpg",
+			uploaded: new Date().toISOString(),
+			requireSignedURLs: options?.requireSignedURLs ?? false,
+			meta: options?.metadata ?? {},
+			variants: ["public"],
+			draft: false,
+			creator: options?.creator,
+		};
+
+		await this.env.IMAGES_STORE.put(id, buffer, { metadata });
```
emily-shen, line 79: how big can these images get? i wonder if it is worth using the disk service directly rather than storing via the KV emulator (still using kv for metadata, but the value can be the path etc.)

just throwing it out there, the code is quite tidy as is

fixtures/vitest-pool-workers-examples/images/test/images.test.ts
emily-shen, line 1: i think it would make more sense to have these tests in miniflare (sorry i realise these were already here, but these fixtures are basically the public docs for our vitest integration so it could be a bit confusing to have tests for miniflare implementation in here.)

packages/miniflare/src/workers/images/images.worker.ts
mglewis: Cloudflare Hosted Images have a max size of 10MB which is less than the 25MB(?) limit of KV. I'm happy to change to use the disk service if you prefer, but I suspect it is probably ok as-is.

packages/miniflare/src/workers/images/images.worker.ts
mglewis: These have now been removed 👍

fixtures/vitest-pool-workers-examples/images/test/images.test.ts
mglewis: I've moved the new tests across and added a new `getImagesBinding()` method.

packages/miniflare/src/plugins/core/proxy/client.ts
```diff
@@ -342,13 +342,16 @@ class ProxyStubHandler<T extends object>
 				},
 			};
 		};
+		const proxy = this.bridge.getProxy(target) as any;
```
dochne, line 345: Did you mean to switch this to `as any` instead of `as ImagesBinding`?

mglewis: The `as any` is deliberate. We get a compile time failure if we do something like `this.bridge.getProxy(target) as ImagesBinding`. The underlying cause is a signature mismatch in `ImagesBinding.input` - a deliberate choice when that code was introduced in order to handle transform chains in a single call.

One alternative would be to destructure `bridge` from `this` which then allows us to  call `bridge.getProxy(target) as ImagesBinding` inline in each method. 

This would look something like:
```
		const { bridge } = this;
		const binding = {
			info: (stream: ReadableStream<Uint8Array>) => {
				// @ts-expect-error The stream types are mismatched
				return (bridge.getProxy(target) as ImagesBinding)["info"](stream);
			},
			input: (stream: ReadableStream<Uint8Array>) => {
				return transformer(bridge.getProxy(target), stream, []);
			},
			get hosted(): ImagesBinding["hosted"] {
				return (bridge.getProxy(target) as ImagesBinding)["hosted"];
			},
		};
```

Happy to make this change, or to add a comment documenting the rationale.

mglewis:
> generally lgtm, could you get a review from the images team too to double check the implementation matches your team's expectations?

I've had a chat with @dochne (Images colleague) and he's happy with the implementation.

.changeset/images-binding-crud-operations.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": minor
+---
+
+Add Hosted Images CRUD operations to Images binding.
```
emily-shen, line 5:
```suggestion
Add Hosted Images CRUD operations to Images binding.

This is an experimental API that only works locally for the moment.
```

or something like that

PR #10174, opened 2025-08-01
make sure we run wrangler e2e tests in nested folders

vicb:
`tools/e2e/runIndividualE2EFiles.ts` was missing test files in nested directories.

That before the PR the top level `test:e2e:wrangler` was not testing the same as `test:e2e` from the wrangler package (the former would miss nested folders).

I also updated the file so that it can be executed from anywhere by using `__dirname` rather than relying on the cwd.

Do we want to update wrangler's `test:e2e` to also use `runIndividualE2EFiles`? (that's how I originally miss that the behavior was not correct on CI)

Nested tests are now executed:

image

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: will make sure the test logs are as expected
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: test runner changer
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a wrangler changer

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1109,7 +1109,7 @@ importers:
         version: 5.8.2
       vitest:
         specifier: ~3.0.7
-        version: 3.0.9(@types/debug@4.1.12)(@types/node@20.19.9)(@vitest/ui@3.2.3)(jiti@2.4.2)(lightningcss@1.29.2)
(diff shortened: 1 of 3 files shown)


petebacondarwin:
> Do we want to update wrangler's test:e2e to also use runIndividualE2EFiles? (that's how I originally miss that the behavior was not correct on CI)


runIndividualE2EFiles actually calls `test:e2e` itself so that would make it circular.

I like that `test:e2e` can be controlled with additional flags and so on

vicb:
Thanks for the review Pete!

PR #10179, opened 2025-08-01
[wrangler] fix: prevent multiple workflows definition with same name

pombosilva:
Fixes WOR-795

Users should not be allowed to define multiple workflows with the same "name" property within a single Wrangler jsonc/toml file. If duplicate names are allowed, only the last-defined workflow would be active, as it would overwrite any earlier workflows with the same name.

To fix this, the code now includes a check for duplicate "name" properties within the workflows array.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s):
  - [x] Documentation not necessary because: only makes a specific scenario not work as intended
- Wrangler V3 Backport
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/10180
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/__tests__/workflows.test.ts
+++ b/packages/wrangler/src/__tests__/workflows.test.ts
@@ -780,5 +780,30 @@ describe("wrangler workflows", () => {
 			await expect(runWrangler("deploy --dry-run")).rejects.toThrow();
 			expect(std.err).toContain('"workflows" bindings should be objects');
 		});
+
+		it("should reject workflows binding with same name", async () => {
+			writeWorkerSource({ format: "ts" });
+			writeWranglerConfig({
+				main: "index.ts",
+				workflows: [
+					{
+						binding: "MY_WORKFLOW",
+						name: "valid-workflow-name",
+						class_name: "MyWorkflow",
+						script_name: "external-script",
+					},
+					{
+						binding: "MY_WORKFLOW_2",
+						name: "valid-workflow-name",
+						class_name: "MyWorkflow2",
+						script_name: "external-script-2",
+					},
+				],
+			});
+			await expect(runWrangler("deploy --dry-run")).rejects.toThrow();
+			expect(std.err).toContain(
+				'"workflows" bindings must have unique "name" values'
+			);
+		});
 	});
 });
--- a/packages/wrangler/src/__tests__/workflows.test.ts
+++ b/packages/wrangler/src/__tests__/workflows.test.ts
@@ -788,22 +788,46 @@ describe("wrangler workflows", () => {
 				workflows: [
 					{
 						binding: "MY_WORKFLOW",
-						name: "valid-workflow-name",
+						name: "duplicate-workflow-name",
 						class_name: "MyWorkflow",
 						script_name: "external-script",
 					},
 					{
 						binding: "MY_WORKFLOW_2",
-						name: "valid-workflow-name",
+						name: "duplicate-workflow-name",
 						class_name: "MyWorkflow2",
 						script_name: "external-script-2",
+					},
+					{
+						binding: "MY_WORKFLOW_3",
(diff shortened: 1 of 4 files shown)


petebacondarwin (changes requested):
Thanks for adding this @pombosilva - can you add a test for it somewhere... I know we discussed the fact that the other workflow stuff is tested separately to the standard config validation code.

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -780,5 +780,30 @@ describe("wrangler workflows", () => {
...
+						name: "valid-workflow-name",
+						class_name: "MyWorkflow2",
+						script_name: "external-script-2",
+					},
+				],
+			});
+			await expect(runWrangler("deploy --dry-run")).rejects.toThrow();
+			expect(std.err).toContain(
+				'"workflows" bindings must have unique "name" values'
+			);
```
petebacondarwin: Consider using an inline snapshot here so that we can also get a check on the diagnostic message (e.g. which names are duplicated):

```suggestion
			expect(std.err).toMatchInlineSnapshot();
```

When you next run the test it will write the snapshot into this call as a parameter.

packages/wrangler/src/config/validation-helpers.ts
```diff
@@ -673,6 +673,29 @@ const isRecord = (
...
+ * Ensure that all bindings in an array have unique `name` properties.
+ */
+export const validateUniqueNameProperty: ValidatorFn = (
+	diagnostics,
+	field,
+	value
+) => {
+	if (Array.isArray(value)) {
+		const names = value.map((entry) => entry.name);
+		const duplicates = names.filter((name, i) => names.indexOf(name) !== i);
```
petebacondarwin: It might not be the case here as the number of items in this array is likely to be small (e.g. <10) but `indexOf` is often not a very efficient way to check for duplicates.
If it were larger then I would use sets or perhaps a map.

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -780,5 +780,30 @@ describe("wrangler workflows", () => {
...
+						class_name: "MyWorkflow",
+						script_name: "external-script",
+					},
+					{
+						binding: "MY_WORKFLOW_2",
+						name: "valid-workflow-name",
+						class_name: "MyWorkflow2",
+						script_name: "external-script-2",
+					},
+				],
```
petebacondarwin, line 819: Maybe include more than one duplicate and some non duplicates so that we can see that how it handles the duplicates and doesn't report incorrectly on non-duplicates.

petebacondarwin (approved):

packages/wrangler/src/__tests__/workflows.test.ts
```diff
@@ -788,22 +789,40 @@ describe("wrangler workflows", () => {
...
+						script_name: "external-script-5",
+					},
 				],
 			});
 			await expect(runWrangler("deploy --dry-run")).rejects.toThrow();
-			expect(std.err).toContain(
-				'"workflows" bindings must have unique "name" values'
-			);
+
+			expect(stripAnsi(std.err)).toMatchInlineSnapshot(`
```
petebacondarwin: I think the stripAnsi is not strictly necessary here?
The `std` props from `mockConsoleMethods()` is already normalized to make is safe to use in snapshots.

pombosilva: Without stripAnsi it expects std.err to have bold/colored chars

petebacondarwin: Does that matter? It's just a snapshot. And also tests that the formatting of the text doesn't change unexpectedly.

PR #10180, opened 2025-08-01
V3 Backport [#10179]: [wrangler] fix: prevent multiple workflows definition with same name

workers-devprod:
This is an automatically opened PR to backport patch changes from #10179 to Wrangler v3


dario-piotrowicz:
Closing since workflows were not GA in v3 and the agreement on v3 backports was only for non-beta features.

PR #10212, opened 2025-08-04
Support packages and virtual modules in the `main` field of the Worker config

jamesopstad:
Fixes #000.

Adds support for package exports and virtual modules in the `main` field of the Worker config. The primary use case is to enable users to directly provide a file exported by a framework as the Worker entry module.

The changes made as part of this PR also prepare for better Rolldown support. By using virtual modules as entry files, it will be easier to adopt Rolldown's [hook filter](https://vite.dev/guide/rolldown.html#hook-filter-feature) feature. Avoiding calling `this.resolve` in `load` or `transform` hooks also prevents errors when using modules that may pass through `@rollup/common-js`, such as package exports.

The entry file name in the build output is now hardcoded to `index.js`. This was necessary because of using a virtual module but is also more reliable than inferring from the first entry chunk as users/frameworks may add additional entry chunks.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): TODO
  - [x] Documentation not necessary because: to be included in relevant framework guides
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: only adds a new experimental feature to Wrangler

### DIFF
--- a/.changeset/eighty-nights-vanish.md
+++ b/.changeset/eighty-nights-vanish.md
@@ -2,4 +2,4 @@
 "wrangler": patch
 ---
 
-Add `experimental.preserveOriginalMain` option to `unstable_readConfig`. This will pass the original `main` value through, without converting it to an absolute path.
+Add `preserveOriginalMain` option to `unstable_readConfig`. This will pass the original `main` value through, without converting it to an absolute path.
--- a/packages/vite-plugin-cloudflare/src/workers-configs.ts
+++ b/packages/vite-plugin-cloudflare/src/workers-configs.ts
@@ -127,7 +127,7 @@ function readWorkerConfig(
 		unstable_readConfig(
 			{ config: configPath, env },
 			// Preserve the original `main` value so that Vite can resolve it
-			{ experimental: { preserveOriginalMain: true } }
+			{ preserveOriginalMain: true }
 		);
 	const raw = structuredClone(config) as RawWorkerConfig;
 
--- a/packages/wrangler/src/config/index.ts
+++ b/packages/wrangler/src/config/index.ts
@@ -118,14 +118,11 @@ export type ReadConfigCommandArgs = NormalizeAndValidateConfigArgs & {
 	script?: string;
 };
 
-export interface ExperimentalReadConfigOptions {
-	// Used by the Vite plugin
-	preserveOriginalMain?: boolean;
-}
-
 export type ReadConfigOptions = ResolveConfigPathOptions & {
 	hideWarnings?: boolean;
-	experimental?: ExperimentalReadConfigOptions;
+	// Used by the Vite plugin
+	// If set to `true`, the `main` field is not converted to an absolute path
+	preserveOriginalMain?: boolean;
 };
 
 export type ConfigBindingOptions = Pick<
@@ -161,7 +158,7 @@ export function readConfig(
 		configPath,
 		userConfigPath,
 		args,
-		options.experimental
+		options.preserveOriginalMain
 	);
 
 	if (diagnostics.hasWarnings() && !options?.hideWarnings) {
--- a/packages/wrangler/src/config/validation.ts
+++ b/packages/wrangler/src/config/validation.ts
@@ -35,7 +35,6 @@ import {
 	validateUniqueNameProperty,
 } from "./validation-helpers";
 import { configFileName, formatConfigSnippet } from ".";
-import type { ExperimentalReadConfigOptions } from ".";
 import type { CfWorkerInit } from "../deployment-bundle/worker";
 import type { Config, DevConfig, RawConfig, RawDevConfig } from "./config";
 import type {
@@ -80,7 +79,7 @@ export function isPagesConfig(rawConfig: RawConfig): boolean {
  * @param configPath The path to the config file
  * @param userConfigPath
  * @param args
- * @param experimentalOptions
+ * @param preserveOriginalMain
  * @returns The normalized `config` and `diagnostics` message
  */
 export function normalizeAndValidateConfig(
@@ -88,7 +87,7 @@ export function normalizeAndValidateConfig(
 	configPath: string | undefined,
 	userConfigPath: string | undefined,
 	args: NormalizeAndValidateConfigArgs,
-	experimentalOptions: ExperimentalReadConfigOptions = {}
+	preserveOriginalMain = false
 ): {
 	config: Config;
 	diagnostics: Diagnostics;
@@ -162,7 +161,7 @@ export function normalizeAndValidateConfig(
 		configPath,
 		rawConfig,
 		isDispatchNamespace,
-		experimentalOptions
+		preserveOriginalMain
 	);
 
 	const isRedirectedConfig = configPath && configPath !== userConfigPath;
@@ -227,7 +226,7 @@ export function normalizeAndValidateConfig(
 					configPath,
 					rawEnv,
 					isDispatchNamespace,
-					experimentalOptions,
+					preserveOriginalMain,
 					envName,
 					topLevelEnv,
 					isLegacyEnv,
@@ -240,7 +239,7 @@ export function normalizeAndValidateConfig(
 					configPath,
 					topLevelEnv, // in this case reuse the topLevelEnv to ensure that nonInherited fields are not removed
 					isDispatchNamespace,
-					experimentalOptions,
+					preserveOriginalMain,
 					envName,
 					topLevelEnv,
 					isLegacyEnv,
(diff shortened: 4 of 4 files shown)


packages/vite-plugin-cloudflare/src/shared.ts
```diff
@@ -11,3 +11,5 @@ export const additionalModuleGlobalRE = new RegExp(
 );
 
 export const WORKER_ENTRY_PATH_HEADER = "__VITE_WORKER_ENTRY_PATH__";
+
+export const VIRTUAL_WORKER_ENTRY = `${virtualPrefix}worker-entry`;
```
jamesopstad, line 15: This name is visible to the user. When a file is updated they will see:

`[vite] hot updated: virtual:vite-plugin-cloudflare/worker-entry`

The last segment may also be used in chunk names e.g. `worker-entry-[hash].js`, depending on how code is split.

Are we happy with `worker-entry` as the name?

.changeset/gold-dodos-stick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": minor
+---
+
+Support package exports and virtual modules in the `main` field of the Worker config. The primary use case is to enable users to directly provide a file exported by a framework as the Worker entry module.
```
edmundhung: By "package exports", do you means the `exports` property on `package.json`? Might be good to include an example package.json / vite config in the changeset

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -199,31 +201,46 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
...
+			resolveId(source) {
 				const workerConfig = getWorkerConfig(this.environment.name);
 
 				if (!workerConfig) {
 					return;
 				}
 
-				const resolvedWorkerEntry = await this.resolve(workerConfig.main);
+				if (source === VIRTUAL_WORKER_ENTRY) {
+					return `\0${VIRTUAL_WORKER_ENTRY}`;
```
edmundhung, line 217: What is the purpose of `\0` here?

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -167,7 +172,10 @@ export function createCloudflareEnvironmentOptions({
 			copyPublicDir: false,
 			ssr: true,
 			rollupOptions: {
-				input: workerConfig.main,
+				input: {
+					// Changes to this file name must be reflected when populating `main` in the `generateBundle` hook in `index.ts`
+					index: VIRTUAL_WORKER_ENTRY,
```
edmundhung: If I understand this correctly, this key must match `workerConfig.main`, right? Should we make it a shared constant too? e.g.

```ts
input: {
 [MAIN_ENTRY]: VIRTUAL_WORKER_ENTRY,
}
```

packages/vite-plugin-cloudflare/src/nodejs-compat.ts
```diff
@@ -150,23 +149,16 @@ export class NodeJsCompat {
...
-		const modified = new MagicString(code);
-		modified.prepend(injectedCode);
-		modified.prepend(polyfillCode);
+			.map((polyfillPath) => `import "${polyfillPath}";`)
+			.join("\n");
 
-		return {
-			code: modified.toString(),
-			map: modified.generateMap({ hires: "boundary", source: id }),
-		};
```
edmundhung, line 169: Why are we not using magic string here anymore? Does vite handle it for us?

packages/vite-plugin-cloudflare/src/shared.ts
edmundhung: Is this the name for primary worker only? How about auxiliary workers?

packages/vite-plugin-cloudflare/src/workers-configs.ts
```diff
@@ -326,30 +331,40 @@ export function getWorkerConfig(
...
 
+/**
+ * If `main` ends with a valid file extension it is resolved to an absolute path.
+ * Else `main` is returned as is so that it can be resolved by Vite.
+ * This enables resolving entry modules relative to the Worker config while also supporting virtual modules and package exports.
+ */
+function maybeResolveMain(main: string, configPath: string): string {
+	if (!ENTRY_MODULE_EXTENSIONS.some((extension) => main.endsWith(extension))) {
+		return main;
+	}
```
edmundhung, line 354: You can technically include an extension in a package entrypoint, though that seems pretty uncommon. Would it be safer to default to treating it as a path, and only consider it a module if it does not exist on the file system?

.changeset/eighty-nights-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Add `experimental.preserveOriginalMain` option to `unstable_readConfig`. This will pass the original `main` value through, without converting it to an absolute path.
```
edmundhung: Do we really need to call it experimental option considering the API is still unstable?

I think `experimental` usually refers to experimental flag on wrangler.

packages/vite-plugin-cloudflare/src/shared.ts
jamesopstad: It's the same for each Worker as they each get their own environment with a separate resolver, build output etc.. Hot updates will also display the environment name.

.changeset/gold-dodos-stick.md
jamesopstad: Yes. Maybe it would be clearer to just say 'Support packages and virtual modules in the `main` field ...'  as it just uses normal package resolution (with or without exports)?

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: This is a convention when working with virtual files. See https://vite.dev/guide/api-plugin.html#virtual-modules-convention. Reading that section again, I think the prefix should probably be `virtual:cloudflare/` rather than `virtual:vite-plugin-cloudflare/`. What do you think?

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
jamesopstad: Good idea. I'll also tighten this up so that we use the name given here to fetch the full file name (including extension) from the bundle in `generateBundle`. We shouldn't assume that the extension in `.js` because this can be overridden in the config.

packages/vite-plugin-cloudflare/src/nodejs-compat.ts
jamesopstad: Previously we were adding this code directly to the user's entry module so we had to modify the existing code. Now that we are using an additional virtual module that reexports the user's module we can just include the code in the virtual module directly.

packages/vite-plugin-cloudflare/src/workers-configs.ts
jamesopstad: The issue then is that it would be a change to the existing behaviour where we error if the file does not exist. There's also a slight risk that the file at the absolute path doesn't exist so we pass it through but then it resolves to a different (and unintended) file that does exist. Part of the problem is that we resolve the `main` field relative to the Worker config file and we treat paths as relative, even if they don't have leading `./`.

Not allowing `package.json` exports that include a file extension seems like an OK tradeoff but open to other ideas here.

.changeset/eighty-nights-vanish.md
jamesopstad: Good point. I'll remove the `experimental` naming.

jamesopstad: https://github.com/cloudflare/workers-sdk/pull/10212/commits/be0d1c467d9d8f943c679ae51eba73084cc9fac1

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
jamesopstad: https://github.com/cloudflare/workers-sdk/pull/10212/commits/acb305142ebf89966f4aaa4d6a53aa2b2b6e7f2d

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: https://github.com/cloudflare/workers-sdk/pull/10212/commits/acb305142ebf89966f4aaa4d6a53aa2b2b6e7f2d

edmundhung (approved):
🔥

PR #10217, opened 2025-08-04
Increase maxBuffer for capnp upload

veggiedefender:
Wrangler breaks when trying to compile a capnp schema large enough that the output is bigger than 1 MB.

```
✘ [ERROR] spawnSync capnp ENOBUFS
```

This happens because spawnSync has a [default buffer size of 1024 * 1024](https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options). The internal data team's message.capnp compiles to a little over a megabyte :(

```
capnp compile -o- schemas/message.capnp | wc -c
 1139904
```

There's likely a better long term fix (async collecting a stream) but this seems like a fine quick fix.

- Tests
  - [ ] Tests included
  - [X] Tests not necessary because: it's a small option tweak..
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: this is an undocumented unsafe feature
- Wrangler V3 Backport
  - [X] Wrangler PR: TODO
  - [ ] Not necessary because:


packages/wrangler/src/deployment-bundle/capnp.ts
```diff
@@ -20,12 +20,11 @@ export function handleUnsafeCapnp(capnp: CfCapnp): Buffer {
 		);
 	}
 	const srcPrefix = resolve(base_path ?? ".");
-	const capnpProcess = spawnSync("capnp", [
-		"compile",
-		"-o-",
-		`--src-prefix=${srcPrefix}`,
-		...capnpSchemas,
-	]);
+	const capnpProcess = spawnSync(
```
vicb, line 23: Could you please add the command you have in the PR description to figure out the size and the current output.

If the size is a little over 1MB, 3MB should be enough below?

veggiedefender: Yeah 3 MB should be enough. Updated.

veggiedefender:
Added a changeset and more details on how to trigger the error!

petebacondarwin:
Skipping description validation as it doesn't seem to like the URL to the backport PR.

petebacondarwin (approved):

petebacondarwin:
Finally got this green and landed! Thanks for your patience.

PR #10232, opened 2025-08-05
validate id in `wrangler containers delete` and also include container API calls in WRANGLER_LOG=debug

emily-shen:
Fixes CC-5810.

Previously we'd see `DELETE method not allowed for the oauth_token authentication scheme` when you try and do something like `wrangler containers images delete <container-name>`, because `<container-name>` didn't look like an ID to API gateway and it would get confused.

Also, added container API calls to the output of `WRANGLER_LOG=debug`, since I wanted to see that when i was debugging.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/containers-shared/src/client/core/request.ts
+++ b/packages/containers-shared/src/client/core/request.ts
@@ -402,7 +402,7 @@ const debugLogRequest = async (
 	config: OpenAPIConfig,
 	url: string,
 	headers: Headers,
-	body: any
+	body: FormData | unknown
 ) => {
 	config.LOGGER?.debug(`-- START CF API REQUEST: ${url}`);
 	const logHeaders = new Headers(headers);
--- a/packages/containers-shared/src/client/core/request.ts
+++ b/packages/containers-shared/src/client/core/request.ts
@@ -352,6 +352,7 @@ export const request = <T>(
 			const formData = getFormData(options);
 			const body = getRequestBody(options);
 			const headers = await getHeaders(config, options);
+			debugLogRequest(config, url, headers, formData ?? body ?? {});
 
 			if (!onCancel.isCancelled) {
 				const response = await sendRequest(
@@ -387,7 +388,7 @@ export const request = <T>(
 						body: responseHeader ?? responseBody,
 					};
 				}
-
+				debugLogResponse(config, result);
 				catchErrorCodes(options, result);
 				resolve(result.body);
 			}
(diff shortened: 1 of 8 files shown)


packages/wrangler/src/containers/containers.ts
```diff
@@ -27,16 +27,20 @@ export function deleteYargs(args: CommonYargsArgv) {
...
 	_config: Config
 ) {
-	if (!deleteArgs.ID) {
+	// API gateway has path restrictions so if someone provides a string that isn't ID shaped, we get a weird error instead of a 404
+	const uuidRegex =
+		/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+	if (!uuidRegex.test(deleteArgs.ID)) {
 		throw new Error(
-			"You must provide an ID. Use 'wrangler containers list` to view your containers."
+			"You must provide the container ID. Use 'wrangler containers list` to view your containers and corresponding IDs."
```
nikitassharma:
```suggestion
			"You must provide the container ID. Use `wrangler containers list` to view your containers and corresponding IDs."
```
since the string is anyways being updated

nikitassharma (approved):

flakey5 (approved):

packages/containers-shared/src/client/core/request.ts
```diff
@@ -396,3 +397,38 @@ export const request = <T>(
 		}
 	});
 };
+
+const debugLogRequest = async (
+	config: OpenAPIConfig,
+	url: string,
+	headers: Headers,
+	body: any
```
petebacondarwin: Can we do better than `any`?

petebacondarwin: If it is really unknown then how about `unknown`??

packages/wrangler/src/containers/containers.ts
```diff
@@ -27,16 +27,20 @@ export function deleteYargs(args: CommonYargsArgv) {
...
 	_config: Config
 ) {
-	if (!deleteArgs.ID) {
+	// API gateway has path restrictions so if someone provides a string that isn't ID shaped, we get a weird error instead of a 404
+	const uuidRegex =
+		/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+	if (!uuidRegex.test(deleteArgs.ID)) {
 		throw new Error(
-			"You must provide an ID. Use 'wrangler containers list` to view your containers."
+			"You must provide the container ID. Use `wrangler containers list` to view your containers and corresponding IDs."
```
petebacondarwin: I think it is good practice to tell them what they actually passed.

Perhaps:

```suggestion
			`Expected a container ID but got \`${deleteArgs.ID}\`. Use \`wrangler containers list\` to view your containers and corresponding IDs."
```

petebacondarwin (approved):
a couple of nits

packages/containers-shared/src/client/core/request.ts
emily-shen: i'll make it unknown. unfortunately this part of the code (generated api client) is full of `any`s...

PR #10237, opened 2025-08-05
update http,https,tls nodejs presets

anonrig:
Addresses the changes in:
- https://github.com/cloudflare/workerd/pull/4690
- https://github.com/cloudflare/workerd/pull/4692

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: Not yet released. Will need to update workerd before landing this PR
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: unenv-preset is not backported to v3


packages/unenv-preset/src/runtime/node/https.ts
petebacondarwin, line 1: Does this mean that https is fully implemented by workerd whether the flag is on or not?

We have logic elsewhere to use this polyfill when the `enable_nodejs_http_server_modules` is on but this polyfill file is now fully pulling through from workerd, right? So the flag is not making any real difference?

packages/unenv-preset/src/runtime/node/tls.ts
```diff
@@ -1,63 +0,0 @@
...
-	getCiphers,
-	rootCertificates,
-	Server,
-} from "unenv/node/tls";
-import type nodeTls from "node:tls";
-
-export {
-	CLIENT_RENEG_LIMIT,
-	CLIENT_RENEG_WINDOW,
-	createSecurePair,
```
petebacondarwin, line 19: The PR to add this to workerd was closed without merging: https://github.com/cloudflare/workerd/pull/4690#issuecomment-3156479136

So I don't think we can just remove this file since that would be a breaking change to users that rely on that EOL function.

petebacondarwin (changes requested):

petebacondarwin:
I think as it stands this PR is blocked on https://github.com/cloudflare/workerd/pull/4706/files

PR #10245, opened 2025-08-06
refactor(wrangler): switch single worker dev to use miniflare dev registry implementation

edmundhung:

_Describe your change..._

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -707,11 +706,11 @@ describe("entrypoints", () => {
...
-		// Wait for error to be thrown
-		await waitFor(() => {
-			const output = session.getOutput();
-			expect(output).toMatch(
-				'The `wrangler dev` session for service "bound" does not support proxying entrypoints. Please upgrade "bound"\'s `wrangler` version.'
+		await waitFor(async () => {
+			let response = await fetch(url);
+			expect(response.status).toBe(503);
+			expect(await response.text()).toBe(
+				'Couldn\'t find a local dev session for the "ThingEntrypoint" entrypoint of service "bound" to proxy to'
```
edmundhung, line 713: The Miniflare dev registry is less precise here, since it cannot distinguish between a worker that is registered without the required entrypoint and one that is not registered at all.

I believe this logic mainly exists to support older Wrangler versions before JSRPC was introduced early last year, so it should be safe to drop for now?

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
```diff
@@ -820,57 +819,6 @@ describe("entrypoints", () => {
...
-		);
-
-		// Wait for error to be thrown
-		await waitFor(() => {
-			const output = session.getOutput();
-			expect(output).toMatch(
-				'Cannot proxy to `wrangler dev` session for service "bound" because it uses HTTPS. Please upgrade "bound"\'s `wrangler` version, or remove the `--local-protocol`/`dev.local_protocol` option.'
-			);
-		});
-	});
```
edmundhung: This test is removed as the miniflare dev registry does support HTTPS (#10281)

packages/miniflare/src/index.ts
```diff
@@ -2570,4 +2570,5 @@ export {
 	type WorkerRegistry,
 	type WorkerDefinition,
 	getDefaultDevRegistryPath,
+	getWorkerRegistry,
```
edmundhung, line 2579: We export this method for Wrangler to read the worker registry for printing the bindings table before miniflare startup.

packages/miniflare/src/shared/external-service.ts
```diff
@@ -212,7 +212,7 @@ export function createOutboundDoProxyService(
                                 const klass = createProxyPrototypeClass(DurableObject, (key) => {
 									const message = ${
 										isProxyEnabled
-											? `\`Cannot access "\${key}" as Durable Object RPC is not yet supported between multiple dev sessions.\``
+											? `\`Cannot access "\${className}#\${key}" as Durable Object RPC is not yet supported between multiple dev sessions.\``
```
edmundhung, line 215: Minor improvement to include the class name in the error message.
e.g. `Cannot access "MyDurableObject#ping" as Durable Object RPC is not yet supported between multiple dev sessions.`

packages/miniflare/src/workers/assets/rpc-proxy.worker.ts
```diff
@@ -22,6 +22,15 @@ export default class RPCProxyWorker extends WorkerEntrypoint<Env> {
...
 	}
 
+	tail(events: TraceItem[]) {
+		// Temporary workaround: the tail events is not serializable over capnproto yet
+		// But they are effectively JSON, so we are serializing them to JSON and parsing it back to make it transferable.
+		// @ts-expect-error FIXME when https://github.com/cloudflare/workerd/pull/4595 lands
+		return this.env.USER_WORKER.tail(
+			JSON.parse(JSON.stringify(events, tailEventsReplacer), tailEventsReviver)
+		);
+	}
```
edmundhung, line 32: This fixes tail handler support over the dev registry in wrangler dev  when static assets are enabled.

packages/wrangler/src/dev/miniflare/index.ts
```diff
@@ -1039,53 +739,28 @@ export function buildMiniflareBindingOptions(
...
-									// doesn't allow IDs from Durable Objects with different unique keys
-									// to be used with each other.
-									unsafeUniqueKey: `${scriptName}-${className}`,
-								},
-							];
+					return [
+						name,
+						{
+							className,
+							scriptName,
```
edmundhung, line 741: Before this change, we skipped passing `scriptName` here likely because we consider those to be external DO which would be handled by the dev registry. But that means wrangler dev multi configs mode doesn't support `scriptName`. This fixes that.

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
penalosa: I agree—I don't think we need to support Wrangler v3 to v4 dev registry connections

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
penalosa: Can we change this to be a test that passes with https configured? Why do we need to remove it?

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -152,7 +154,7 @@ async function resolveDevConfig(
 		testScheduled: input.dev?.testScheduled,
 		// absolute resolved path
 		persist: localPersistencePath,
-		registry: input.dev?.registry,
+		registryPath: input.dev?.registryPath,
```
penalosa: Do we need to rename this? We have other fileds in this object that are paths but aren't named with the suffix `Path` (i.e. `persist` above)

packages/wrangler/src/dev/miniflare/index.ts
```diff
@@ -183,7 +79,7 @@ export interface ConfigBundle {
 	complianceRegion: Config["compliance_region"] | undefined;
 	bindings: CfWorkerInit["bindings"];
 	migrations: Config["migrations"] | undefined;
-	workerDefinitions: WorkerRegistry | undefined | null;
+	devRegistryPath: string | undefined;
```
penalosa: As above—can we just call this `registry` or `devRegistry`?

penalosa (approved):
Fantastic work! 🎉

fixtures/entrypoints-rpc-tests/tests/entrypoints.spec.ts
edmundhung: Added the test back in https://github.com/cloudflare/workers-sdk/pull/10245/commits/b12d3269f47e037fb476e506498cfddef2d84619 and it now passes with https configured

packages/wrangler/src/api/startDevWorker/ConfigController.ts
edmundhung: removed the path suffix in https://github.com/cloudflare/workers-sdk/pull/10245/commits/2835b622163bde8f20285e3bb3c665a77b9c5824

packages/wrangler/src/dev/miniflare/index.ts
edmundhung: Named it `devRegistry` in https://github.com/cloudflare/workers-sdk/commit/2835b622163bde8f20285e3bb3c665a77b9c5824

PR #10248, opened 2025-08-06
fix: re-push image even when only the dockerfile has changed

emily-shen:
Fixes CC-5563

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/src/__tests__/cloudchamber/build.test.ts
+++ b/packages/wrangler/src/__tests__/cloudchamber/build.test.ts
@@ -5,6 +5,7 @@ import {
 	dockerLoginManagedRegistry,
 	getCloudflareContainerRegistry,
 	runDockerCmd,
+	runDockerCmdWithOutput,
 } from "@cloudflare/containers-shared";
 import { UserError } from "../../errors";
 import { mockAccountId, mockApiToken } from "../helpers/mock-account-id";
@@ -17,6 +18,7 @@ vi.mock("@cloudflare/containers-shared", async (importOriginal) => {
 	return Object.assign({}, actual, {
 		dockerLoginManagedRegistry: vi.fn(),
 		runDockerCmd: vi.fn(),
+		runDockerCmdWithOutput: vi.fn(),
 		dockerBuild: vi.fn(() => ({ abort: () => {}, ready: Promise.resolve() })),
 		dockerImageInspect: vi.fn(),
 	});
@@ -37,6 +39,9 @@ describe("buildAndMaybePush", () => {
 			.mockResolvedValueOnce("[]")
 			// return image size and number of layers
 			.mockResolvedValueOnce("53387881 2");
+		vi.mocked(runDockerCmdWithOutput).mockReturnValueOnce(
+			'{"Descriptor":{"digest":"config-sha"}}'
+		);
 		mkdirSync("./container-context");
 
 		writeFileSync("./container-context/Dockerfile", dockerfile);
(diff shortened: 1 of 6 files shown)


packages/wrangler/e2e/containers.test.ts
emily-shen, line 1: argh - hoped github might've preserved the changes when i renamed this file. basically this is the same as the original, except i wanted to add some e2e tests for this image re-pushing behaviour since the docker command we use is experimental and could break under us. although i did simplify some of the dev test setup while i was there too.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -120,25 +132,18 @@ export async function buildAndMaybePush(
...
-				//
-				// A Docker image digest is a unique, cryptographic identifier (SHA-256 hash)
-				// representing the content of a Docker image. Unlike tags, which can be reused
-				// or changed, a digest is immutable and ensures that the exact same image is
-				// pulled every time. This guarantees consistency across different environments
-				// and deployments.
-				// From: https://docs.docker.com/dhi/core-concepts/digests/
-				const parsedDigests = JSON.parse(repoDigests);
-
+				const parsedDigests = JSON.parse(imageInfo.split(" ")[0]);
```
gpanders: Small nit, but instead of `imageInfo.split(" ")[0]`, can we do something like

```typescript
const [digests, imageId] = imageInfo.split(" ");
```

I prefer this format because it makes it clear what the expected format of `imageInfo` is and is more "self-documenting". Bonus points also for not having to call `.split(" ")` twice.

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -102,9 +104,19 @@ export async function buildAndMaybePush(
...
+		 * Get `RepoDigests` and `Id`:
+		 * A Docker image digest (RepoDigest) is a unique, cryptographic identifier (SHA-256 hash)
+		 * representing the content of a Docker image. Unlike tags, which can be reused		or changed, a digest is immutable and ensures that the exact same image is
+		 * pulled every time. This guarantees consistency across different environments
+		 * and deployments. Crucially this is *not* affected by metadata changes (dockerfile only changes).
+		 * From: https://docs.docker.com/dhi/core-concepts/digests/
+		 * The image Id is a sha hash of the image's configuration, so it *does* capture metadata changes.
+		 * We need both to know when to push the image to the managed registry.
+		 */
+		const imageInfo = await dockerImageInspect(pathToDocker, {
```
nikitassharma: Can we move this into the `if (push)` block

nikitassharma (approved):

emily-shen:
For posterity, I've ended up removing the e2e tests that check this behaviour because I can't get them to work in CI :( (regular unit tests still exist for this feature, the e2e tests worked locally too fwiw...) 

I think this is ok because I added the e2e tests because i thought it would be useful to test the actual behaviour of `docker manifest inspect` since that was experimental, but we don't control the docker version in github CI anyway (e.g. currently the runners are using a version from March still). Also, if the command breaks, it doesn't break anyone's flow, it just means slightly worse DX as there is an unnecessary push.

packages/wrangler/src/__tests__/cloudchamber/build.test.ts
```diff
@@ -167,26 +172,23 @@ describe("buildAndMaybePush", () => {
...
-		);
-		expect(runDockerCmd).toHaveBeenNthCalledWith(2, "docker", [
+		expect(runDockerCmdWithOutput).toHaveBeenCalledOnce();
+		expect(runDockerCmdWithOutput).toHaveBeenCalledWith("docker", [
+			"manifest",
+			"inspect",
+			"-v",
+			`${getCloudflareContainerRegistry()}/some-account-id/test-app@sha256:three`,
+		]);
+		expect(runDockerCmd).toHaveBeenCalledTimes(1);
```
petebacondarwin, line 182: NIT: 
```suggestion
		expect(runDockerCmd).toHaveBeenCalledOnce();
```

PR #10249, opened 2025-08-06
Support JSRPC for remote bindings

penalosa:
Fixes https://jira.cfdata.org/browse/DEVX-2042, https://jira.cfdata.org/browse/DEVX-1929, https://jira.cfdata.org/browse/DEVX-1933

Support JSRPC bindings! This uses a vendored version of the upcoming `@cloudflare/jsrpc` library. See code for detailed comments.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: captured by separate ticket
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a V3 feature


packages/miniflare/src/workers/shared/remote-proxy-client.worker.ts
```diff
@@ -23,4 +25,25 @@ export default class Client extends WorkerEntrypoint<{
...
+			get(target, prop) {
+				if (Reflect.has(target, prop)) {
+					return Reflect.get(target, prop);
+				}
+
+				const value = Reflect.get(stub, prop);
+				return value;
+			},
+		});
+	}
```
penalosa: This proxy code is inspired by that written for rpc in local dev with the router worker—thanks @CarmenPopoviciu!

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -213,7 +213,9 @@ export function printBindings(
 						destination_address ||
 						allowed_destination_addresses?.join(", ") ||
 						"unrestricted",
-					mode: getMode({ isSimulatedLocally: true }),
+					mode: getMode({
+						isSimulatedLocally: !emailBinding.experimental_remote,
```
dario-piotrowicz: shouldn't you also check `getFlag("REMOTE_BINDINGS")`?

packages/wrangler/e2e/remote-binding/wrangler-remote-resources.test.ts
```diff
@@ -395,7 +454,16 @@ describe.skipIf(!CLOUDFLARE_ACCOUNT_ID)(
 					const { url } = await worker.waitForReady();
 
 					const response = await fetchText(url);
-					expect(response).toMatch(testCase.expectedResponseMatch);
+					expect(response).toEqual(testCase.expectedResponseMatch);
+
+					// Wait for async logging (e.g. pipeline messages received)
+					await setTimeout(1_000);
```
dario-piotrowicz, line 460: is this reliable? would a `vi.waitFor` be more stable here?

packages/miniflare/src/workers/shared/jsrpc.d.ts
dario-piotrowicz, line 1: could/should this be included in the jsrpc package instead?

packages/miniflare/src/workers/shared/remote-proxy-client.worker.ts
```diff
@@ -23,4 +25,24 @@ export default class Client extends WorkerEntrypoint<{
...
+
+		super(ctx, env);
+
+		return new Proxy(this, {
+			get(target, prop) {
+				if (Reflect.has(target, prop)) {
+					return Reflect.get(target, prop);
+				}
+
+				return Reflect.get(stub, prop);
```
dario-piotrowicz, line 31: could you explain this logic here? 🙏 

does it mean that jsrpc has some limitations/doesn't stub everything?

dario-piotrowicz (commented):
Looks great to me! 😄 

I just left a few comments (some just asking info so that I can understand the implementation better 🙂)

___

However one thing I'd like to ask however is, I can see you'd chosen to use `rpcOverWebSocket`, have you considered generally using `rpcOverHttp` instead?
It would likely be slightly slower and more noisy but since we're not talking about something that needs to handle live traffic that could be acceptable. Using non-persistent http could help us avoid persistence issues that can happen for different reasons (such as sessions restarts, weird remote preview bugs and vite environments reloads).

I am by no means saying that I think `rpcOverHttp` would be more appropriate/better, I'm just curious if you've considered or tested using that instead of the permanent websocket connection 🙂

penalosa:
> have you considered generally using `rpcOverHttp` instead

I was hoping to be able to use that, yes! Unfortunately it requires an explicit `batch.send()` operation to actually send things over the wire, which is a non-starter for us (since we don't control the code interacting with the RPC stub and don't know when it's appropriate to initiate that)

packages/miniflare/src/workers/shared/jsrpc.d.ts
penalosa: It will be once the package is published, this is just for the interim stage where I'm installing the jsrpc package from the filesystem

packages/miniflare/src/workers/shared/remote-proxy-client.worker.ts
penalosa: jsrpc stubs everything, but we don't want to stub anything that's defined directly in the proxy worker. In this case it's just `fetch()` (which doesn't need to go over JSRPC and is more efficient as a direct fetch call). We could check `prop === "fetch"`, but `Reflect.has()` seemed nicer and more flexible

dario-piotrowicz:
> since we don't control the code interacting with the RPC stub and don't know when it's appropriate to initiate that

Couldn't we do that as part of the proxy objects? 🤔 as in, when we return a function in a Proxy getter, we could wrap that function so that after its execution it runs `batch.send()` no?

For example something like:
```js
new Proxy(this, {
   get(target, prop) {
    if (Reflect.has(target, prop)) {
     return Reflect.get(target, prop);
    }

    const stubProp = Reflect.get(stub, prop);
    if (typeof stubProp === 'function') {
      return async (...args) => {
        const resPromise = stubProp(...args);
        batch.send();
        return resPromise;
      };
    }
    return stubProp;
  }
});
```

packages/miniflare/src/workers/shared/jsrpc.d.ts
dario-piotrowicz: ah ok thanks 🙂👍

packages/miniflare/src/workers/shared/remote-proxy-client.worker.ts
dario-piotrowicz: ah ok, thanks 😄 makes sense to me 🙂👍

penalosa:
> Couldn't we do that as part of the proxy objects? 🤔 as in, when we return a function in a Proxy getter, we could wrap that function so that after its execution it runs `batch.send()` no?

There are lots of edge cases here:
- What if the value is a getter? When would we send the batch?
- What if the function returns an `RpcTarget` that the user then interacts with further? We'd have to do another `batch.send()` but there wouldn't be a natural place to do this

dario-piotrowicz:
> > Couldn't we do that as part of the proxy objects? 🤔 as in, when we return a function in a Proxy getter, we could wrap that function so that after its execution it runs `batch.send()` no?
> 
> There are lots of edge cases here:
> 
> * What if the value is a getter? When would we send the batch?
> * What if the function returns an `RpcTarget` that the user then interacts with further? We'd have to do another `batch.send()` but there wouldn't be a natural place to do this

Yeah good points 🤔👍

.gitignore
```diff
@@ -230,3 +230,4 @@ dist/**
 .env*
 !.env.example
 .node-cache/
+!vendor/jsrpc/dist
```
dario-piotrowicz:
```suggestion
!vendor/jsrpc/dist

```

.prettierignore
```diff
@@ -49,3 +49,5 @@ fixtures/redirected-config-worker/build/*
 fixtures/redirected-config-worker-with-environments/build/*
 
 packages/vite-plugin-cloudflare/playground/**/*.d.ts
+
+vendor/jsrpc
```
dario-piotrowicz:
```suggestion
vendor/jsrpc

```

packages/vite-plugin-cloudflare/e2e/remote-bindings.test.ts
```diff
@@ -22,32 +21,37 @@ if (!process.env.CLOUDFLARE_ACCOUNT_ID || !process.env.CLOUDFLARE_API_TOKEN) {
 		//       tests to run sequentially in order to avoid race conditions
 		.sequential("remote bindings tests", () => {
 			const replacements = {
-				"<<REMOTE_WORKER_PLACEHOLDER>>": `tmp-e2e-vite-remote-${randomUUID().split("-")[0]}`,
-				"<<REMOTE_WORKER_PLACEHOLDER_ALT>>": `tmp-e2e-vite-remote-alt-${randomUUID().split("-")[0]}`,
+				"<<REMOTE_WORKER_PLACEHOLDER>>": `preserve-e2e-vite-remote`,
+				"<<REMOTE_WORKER_PLACEHOLDER_ALT>>": `preserve-e2e-vite-remote-alt`,
```
dario-piotrowicz, line 25: why this change? (is this related to Pete's changes for stabilizing the remote bindings e2es?) also if we're not using random values then we can just hardcode them and not use placeholders, no?

packages/wrangler/e2e/remote-binding/wrangler-remote-resources.test.ts
petebacondarwin: My PR has code to make this more resilient

packages/wrangler/package.json
```diff
@@ -80,6 +80,7 @@
 		"@cloudflare/cli": "workspace:*",
 		"@cloudflare/containers-shared": "workspace:*",
 		"@cloudflare/eslint-config-worker": "workspace:*",
+		"@cloudflare/jsrpc": "link:../../vendor/jsrpc",
```
dario-piotrowicz, line 83: I wonder, wouldn't it be cleaner adding `"vendor/*" directory to the packages list in "pnpm-workspace.yaml" and treat jsrpc as any other workspace package? 🤔

packages/wrangler/tsup.config.ts
```diff
@@ -38,6 +38,7 @@ function embedWorkersPlugin({
 						sourcemap: process.env.SOURCEMAPS !== "false",
 						sourcesContent: false,
 						metafile: true,
+						external: ["cloudflare:email", "cloudflare:workers"],
```
petebacondarwin, line 41: NIT: this can take regular expressions. So maybe?

```suggestion
						external: /^cloudflare:/,
```

petebacondarwin (approved):

packages/vite-plugin-cloudflare/e2e/remote-bindings.test.ts
petebacondarwin: I assume that Samuel has created a permanent Worker. But it needs some machinery to keep it there...

penalosa: The machinery is in the beforeAll-it checks whether it's present, and if not deploys it. Happy to remove in favour of whatever the wider solution is for shared workers!

packages/wrangler/templates/remoteBindings/ProxyServerWorker.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/**
+ * For most bindings, we expose them as
+ *  - RPC stubs directly to @cloudflare/jsrpc, or
+ *  - HTTP based fetchers
+ * However, there are some special cases:
+ *  - SendEmail bindings need to take EmailMessage as their first parameter,
+ *    which is not serialisable. As such, we reconstruct it before sending it
+ *    on to the binding. See also packages/miniflare/src/workers/email/email.worker.ts
```
dario-piotrowicz:
```suggestion
 *    on to the binding. See packages/miniflare/src/workers/email/email.worker.ts
```

packages/vite-plugin-cloudflare/e2e/remote-bindings.test.ts
petebacondarwin: Oh I missed that. Cool

packages/miniflare/src/plugins/dispatch-namespace/index.ts
```diff
@@ -67,28 +68,8 @@ export const DISPATCH_NAMESPACE_PLUGIN: Plugin<
...
-					config.remoteProxyConnectionString
-				),
-				worker: remoteProxyClientWorker(
-					config.remoteProxyConnectionString,
-					name
-				),
-			};
-		});
+	async getServices() {
+		return [];
```
dario-piotrowicz, line 72: sorry for the dumb question 😓 

Could you explain this change and why we just return `[]` now? 🙏

packages/vite-plugin-cloudflare/e2e/remote-bindings.test.ts
dario-piotrowicz: > The machinery is in the beforeAll-it checks whether it's present, and if not deploys it. Happy to remove in favour of whatever the wider solution is for shared workers!

I'd keep everything as is and let Pete consistently deal with all the remote bindings resources in one go, but it's up to you two I am happy with whatever works best for you 🙂

petebacondarwin: Happy to land this one as-is.

packages/miniflare/src/plugins/dispatch-namespace/index.ts
penalosa: Previously, we relied on the shared proxy client definition in `remoteProxyClientWorker()` to implement dispatch namespace bindings. The user would be given a wrapped binding, which exposed a `get()` method that ultimately called the fetcher defined in this `getServices()` block. As part of that fetch call, we used an HTTP header to pass the dispatch namespace initialisation properties over the wire to be used on the server.

However, now that we support JSRPC over dispatch namespace bindings, we can't use a header to pass over the initialisation properties, because the user won't necessarily perform a fetch. As such, we now construct the JSRPC client directly in the dispatch namespace wrapped binding, and so don't need the shared service.

packages/wrangler/e2e/remote-binding/wrangler-remote-resources.test.ts
penalosa: I'll leave this as-is pending that change then

packages/wrangler/package.json
penalosa: Potentially? But I wanted to draw a clear difference between this and other packages—e.g. we're not running JSRPC's tests, and we're just including the built output.

penalosa: What would the benefit of treating it as a workspace package be?

packages/wrangler/tsup.config.ts
penalosa: It's type seems to be `BuildOptions.external?: string[] | undefined`?

packages/wrangler/package.json
dario-piotrowicz: I was just wondering it that'd be cleaner, pnpm could also optimize something... but it is not a huge deal I'm happy with the current approach I was just wondering 🙂

packages/miniflare/src/plugins/dispatch-namespace/index.ts
dario-piotrowicz: I see, ok, thanks so very much for the clarification! 🫶

PR #10253, opened 2025-08-06
use remote digest sha if no prev deployment is found

emily-shen:

_Describe your change..._

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


petebacondarwin:
Conflicts ... 😢

packages/wrangler/src/__tests__/containers/deploy.test.ts
```diff
@@ -1338,6 +1319,8 @@ describe("wrangler deploy with containers", () => {
 			"
 		`);
 	});
+
+	// todo it should use the remote digest as image ref if found
```
petebacondarwin: todo when??

petebacondarwin: generally better to create assignable tickets to track this sort of thing, unless it was a todo planned to be added to this PR

packages/wrangler/src/cloudchamber/build.ts
```diff
@@ -77,12 +77,21 @@ export function pushYargs(yargs: CommonYargsArgv) {
 		.positional("TAG", { type: "string", demandOption: true });
 }
 
+/**
+ *
```
petebacondarwin, line 83: I appreciate the interesting bit here is that there are two return types, but a function description would be nice.

petebacondarwin: Actually since this return type is used across the code base, how about you make it a documented type (or union of documented types) and then reuse that everywhere instead. More self documenting in all the different places it appears.

packages/wrangler/src/__tests__/containers/deploy.test.ts
emily-shen: weird that was from the other PR about pushing dockerfile changes. i think i messed up my git history in here, let me fix.

packages/wrangler/src/cloudchamber/deploy.ts
```diff
@@ -8,27 +8,25 @@ import { logger } from "../logger";
...
-			newImageLink: containerConfig.image_uri,
-		};
-	}
+): Promise<ImageRef> {
 	const imageFullName = containerConfig.name + ":" + imageTag.split("-")[0];
 	logger.log("Building image", imageFullName);
 
-	const buildResult = await buildAndMaybePush(
+	// if something already exists on remote, it will be {digest:string} and
+	// otherwise, it will be the new image tag {tag:string}
```
nikitassharma:
```suggestion
	// if something already exists on remote, it will be {remoteDigest:string} and
	// otherwise, it will be the new image tag {newTag:string}
```

emily-shen: removed the comment as its a bit redundant with the return type now having a proper jsdoc comment

packages/wrangler/src/__tests__/containers/deploy.test.ts
petebacondarwin: still there 😄

emily-shen: argh! definitely gone this time

petebacondarwin (approved):

PR #10258, opened 2025-08-06
containers: add option to allow all tiers

nikitassharma:
Fixes CC-5818

Adds the option to allow all tiers when creating a container application.

---

- Tests
  - [X] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [X] Documentation not necessary because: not a documented feature
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/containers-shared/src/types.ts
+++ b/packages/containers-shared/src/types.ts
@@ -67,7 +67,7 @@ export type SharedContainerConfig = {
 	constraints: {
 		regions?: string[];
 		cities?: string[];
-		tier?: number;
+		tier: number | undefined;
 	};
 	observability: { logs_enabled: boolean };
 } & InstanceTypeOrLimits;
--- a/packages/containers-shared/src/types.ts
+++ b/packages/containers-shared/src/types.ts
@@ -67,7 +67,7 @@ export type SharedContainerConfig = {
 	constraints: {
 		regions?: string[];
 		cities?: string[];
-		tier: number;
+		tier?: number;
 	};
 	observability: { logs_enabled: boolean };
 } & InstanceTypeOrLimits;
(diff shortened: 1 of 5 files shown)


flakey5 (approved):

packages/wrangler/src/containers/config.ts
```diff
@@ -58,7 +58,11 @@ export const getNormalizedContainerOptions = async (
 			scheduling_policy: (container.scheduling_policy ??
 				SchedulingPolicy.DEFAULT) as SchedulingPolicy,
 			constraints: {
-				tier: container.constraints?.tier ?? 1,
+				// if the tier is -1, then we allow all tiers
+				tier:
+					container.constraints?.tier === -1
```
emily-shen, line 65: so to double check, in wrangler config  `-1` = all tiers, `undefined` should default to `1` , and in the API `undefined` = all tiers?  can you add a comment saying the API will treat `undefined` as 'all tiers'

packages/containers-shared/src/types.ts
```diff
@@ -67,7 +67,7 @@ export type SharedContainerConfig = {
 	constraints: {
 		regions?: string[];
 		cities?: string[];
-		tier: number;
+		tier?: number;
```
emily-shen:
```suggestion
		tier: number | undefined;
```

packages/wrangler/src/containers/deploy.ts
```diff
@@ -183,7 +183,7 @@ function containerConfigToCreateRequest(
 		// deprecated in favour of max_instances
 		instances: 0,
 		max_instances: containerApp.max_instances,
-		constraints: containerApp.constraints,
+		constraints: stripUndefined(containerApp.constraints),
```
emily-shen: when this gets sent to the API, JSON.stringify will strip undefined keys already. this just means we don't show constraints in the diff, but arguably we _do_ want to show this is in the diff right?

nikitassharma: yup, you're right, i removed this change

PR #10273, opened 2025-08-07
fix(miniflare): support WebSocket proxying over dev registry

edmundhung:
Fixes n/a.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: miniflare dev registry is v4 only

### DIFF
--- a/packages/miniflare/src/shared/dev-registry.worker.ts
+++ b/packages/miniflare/src/shared/dev-registry.worker.ts
@@ -103,7 +103,9 @@ class ProxyServer {
 				serviceProxyTarget.entrypoint
 			);
 
-			this.handleProxy(req, res, address);
+			this.handleProxy(req, res, address, (callback) =>
+				this.subscribe(serviceProxyTarget.service, callback)
+			);
 			return;
 		}
 
@@ -120,7 +122,9 @@ class ProxyServer {
 				return;
 			}
 
-			this.handleProxy(req, res, address);
+			this.handleProxy(req, res, address, (callback) =>
+				this.subscribe(doProxyTarget.scriptName, callback)
+			);
 			return;
 		}
 
@@ -242,7 +246,8 @@ class ProxyServer {
 	private handleProxy(
 		req: http.IncomingMessage,
 		res: http.ServerResponse,
-		target: ProxyAddress
+		target: ProxyAddress,
+		onTargetUpdated: (callback: () => void) => void
 	) {
 		const headers = { ...req.headers };
 		let path = target.path;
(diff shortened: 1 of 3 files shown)


packages/miniflare/src/shared/dev-registry.worker.ts
```diff
@@ -294,6 +297,50 @@ class ProxyServer {
...
+
+		if (req.headers.upgrade?.toLowerCase() === "websocket") {
+			upstream.on("upgrade", (upRes, socket, head) => {
+				// For WebSocket upgrades, we need to respond directly on the original request socket
+				if (req.socket && req.socket.writable) {
+					// Build and write complete HTTP response header
+					const statusLine = `HTTP/1.1 ${upRes.statusCode ?? 101} ${upRes.statusMessage ?? "Switching Protocols"}\r\n`;
+					let headersString = "";
+					for (let i = 0; i < upRes.rawHeaders.length; i += 2) {
+						headersString += `${upRes.rawHeaders[i]}: ${upRes.rawHeaders[i + 1]}\r\n`;
```
petebacondarwin, line 309: OOC is the `\r\n` rather than just `\n` to ensure it is valid on Windows?

packages/miniflare/test/dev-registry.spec.ts
```diff
@@ -126,6 +126,99 @@ test("DevRegistry: fetch to module worker", async (t) => {
...
+				async fetch(request, env, ctx) {
+					const wsResponse = await env.SERVICE.fetch(request.url, {
+						headers: { Upgrade: "websocket" }
+					});
+
+					if (wsResponse.webSocket) {
+						wsResponse.webSocket.accept();
+
+						// Test bidirectional communication
+						wsResponse.webSocket.send("ping");
```
petebacondarwin, line 152: Is there a chance that this ping will cause a response before you have setup the listener?

packages/miniflare/src/shared/dev-registry.worker.ts
edmundhung: It's just because the HTTP/1.1 Spec requires these lines to be terminated with CLRF (`\r\n`): https://datatracker.ietf.org/doc/html/rfc7230#section-3

packages/miniflare/test/dev-registry.spec.ts
edmundhung: Good call. I will move this after the `messagePromise` in a follow up PR.

edmundhung: Fixed in [6b3e487](https://github.com/cloudflare/workers-sdk/commit/6b3e487c8ee59ff52924f0c5f5fec538dc245d86) from #10281

PR #10274, opened 2025-08-07
Wrangler fix queue producer binding

tgarg-cf:
Fixes #[MQ-861](https://jira.cfdata.org/browse/MQ-861)

Removed Delivery_delay setting from queue producer bindings and removed a call to putQueue when deploying a worker with a queue producer binding.

delivery_delay was wrongly a parameter of the queue producer binding. Currently, even though our documentation doesn't show it, we can supply the delivery_delay parameter under the [[queues.producers]] sections in the wrangler.toml/jsonc file. This is not expected behaviour as a worker should not be able to affect queue level settings. Additionally, when having multiple producers, the setting will be determined by the last deployed producer which also leads to issues. 

- Tests
  - [x] Tests included (changed tests are updated)
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: A parameter was mistakenly included. It is not mentioned on the online docs, so we don't need to change them.
- Wrangler V3 Backport
  - [x] Wrangler PR: TBD manually as there are rebase conflicts
  - [ ] Not necessary because: not sure


packages/wrangler/src/deploy/deploy.ts
```diff
@@ -1260,27 +1259,10 @@ export function isAuthenticationError(e: unknown): e is ParseError {
...
-				`Producer for ${producer.queue}`,
-			])
-		);
-	}
-
-	return updateProducers;
+export function updateQueueProducers(config: Config): Promise<string[]>[] {
+	return (config.queues.producers || []).map((producer) =>
+		Promise.resolve([`Producer for ${producer.queue}`])
+	);
```
petebacondarwin, line 1265: This function is no longer updating queue producers. Perhaps it should just be removed?

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -279,7 +279,7 @@ export default async function triggersDeploy(
 	}
 
 	if (config.queues.producers && config.queues.producers.length) {
-		const updateProducers = await updateQueueProducers(config);
+		const updateProducers = updateQueueProducers(config);
 		deployments.push(...updateProducers);
 	}
```
petebacondarwin, line 284: This code is now not actually doing anything so it doesn't make sense to keep it.

petebacondarwin (commented):
A few minor suggestions. Otherwise looks fine on the code side.
There is the question of whether removing the option is a breaking change.

packages/wrangler/src/triggers/deploy.ts
tgarg-cf: @petebacondarwin Should we still be pushing `Producer for ${producer.queue}` into deployments?

petebacondarwin: I think this can be removed.

petebacondarwin:
Hi @tgarg-cf — we're doing some cleanup of stale draft PRs.

This PR has been quiet since 2026-01-07. Could you let us know within **14 days** whether you're still planning to continue?

- If yes: please leave a comment with a status update or push a rebase, and we'll keep it open.
- If no: no need to reply — we'll close it after 14 days, and you can always reopen later if you pick it back up.

Thanks!

PR #10283, opened 2025-08-08
feat(wrangler): support long branch names in preview alias generation

WillTaylorDev:
Updates generatePreviewAlias to truncate long branch names with hash suffixes when they exceed DNS label constraints, instead of returning undefined. This allows preview deployments for branches with longer descriptive names.

Fixes #WC-3840

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: alias generation is a function of WCI, and is best-effort.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: v4 feature only

### DIFF
--- a/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
+++ b/packages/wrangler/src/__tests__/versions/versions.upload.test.ts
@@ -368,71 +368,103 @@ describe("generatePreviewAlias", () => {
 	});
 
 	it("sanitizes branch names correctly", () => {
+		const scriptName = "worker";
 		mockExecSync
 			.mockImplementationOnce(() => {}) // is-inside-work-tree
 			.mockImplementationOnce(() => Buffer.from("feat/awesome-feature"));
 
-		const result = generatePreviewAlias("worker");
+		const result = generatePreviewAlias(scriptName);
 		expect(result).toBe("feat-awesome-feature");
+		expect(result).not.toBeUndefined();
+		expect((scriptName + "-" + result).length).toBeLessThanOrEqual(63);
 	});
 
-	it("returns undefined for long branch names which don't fit within DNS label constraints", () => {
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/__tests__/versions/versions.upload.test.ts
```diff
@@ -376,14 +376,19 @@ describe("generatePreviewAlias", () => {
 		expect(result).toBe("feat-awesome-feature");
 	});
 
-	it("returns undefined for long branch names which don't fit within DNS label constraints", () => {
-		const longBranch = "a".repeat(70);
+	it("truncates and hashes long branch names that don't fit within DNS label constraints", () => {
+		const longBranch = "a".repeat(62);
```
petebacondarwin, line 384: any need to change this?

petebacondarwin (approved):
OOC what happens if the preview alias is already used?

packages/wrangler/src/__tests__/versions/versions.upload.test.ts
WillTaylorDev: Nope, the replacement for this test is now `returns undefined when script name is too long to allow any alias`. For long branch names, we just truncate the branch and suffix with a hash of the full branch name. This (nearly) guarantees alias uniqueness for long branches. We would still return undefined if the script name is really long, as we wouldn't be able to generate an alias to fit within the 63 char dns label limit.

WillTaylorDev:
> OOC what happens if the preview alias is already used?

If a preview alias is already used, EWC will point the alias to the present version being uploaded. The hash being suffixed on the end of the alias ensures that branches with the same long prefix will still get distinct aliases.

PR #10288, opened 2025-08-08
fix(wrangler) remove putQueue call when deploying a worker with queue producer binding

tgarg-cf:
Fixes #[MQ-861](https://jira.cfdata.org/browse/MQ-861)

Removed a call to putQueue when deploying a worker with a queue producer binding. 
This is needed because a worker should not be able to affect queue level settings. Additionally, when having multiple producers, the settings will be determined by the last deployed producer which also leads to issues.

---

- Tests
  - [x] Tests included (changed tests are updated)
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: change does not reflect externally. (optionally start deprecating "delivery_delay" as now, setting it will not do anything
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: @petebacondarwin will create the backport once this is merged.

### DIFF
--- a/.changeset/chatty-pens-sniff.md
+++ b/.changeset/chatty-pens-sniff.md
@@ -2,4 +2,6 @@
 "wrangler": patch
 ---
 
-Remove putQueue call when setting queue producer binding in a worker
+Do not attempt to update queue producer settings when deploying a Worker with a queue binding
+
+Previously, each deployed Worker would update a subset of the queue producer's settings for each queue binding, which could result in broken queue producers or at least conflicts where different Workers tried to set different producer settings on a shared queue.
--- a/.changeset/chatty-pens-sniff.md
+++ b/.changeset/chatty-pens-sniff.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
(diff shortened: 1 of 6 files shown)


.changeset/chatty-pens-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Remove putQueue call when setting queue producer binding in a worker
```
petebacondarwin: It is good to word changesets from the point of view of the user, since they are going to be used to generate the changelog. The user doesn't know about the internal implementation. How about:

```suggestion
Do not attempt to update queue producer settings when deploying a Worker with a queue binding

Previously, each deployed Worker would update a subset of the queue producer's settings for each queue binding, which could result in broken queue producers or at least conflicts where different Workers tried to set different producer settings on a shared queue.
```

petebacondarwin (approved):
Love it!

Made a suggestion about the changeset wording - but up to you whether you want to adopt it or a variant.

Thank you for iterating on this.

PR #10306, opened 2025-08-10
chore: install browser with @puppeteer/browsers

ruifigueira:
Browser installation relied on a postinstall script from puppeteer package which was unreliable with npx-import.

For instance, if we have a workers project with browser binding that doesn't explicitly include `puppeteer`, the first time we run it it with `npx wrangler dev` and trigger the `launch` function, it will install the browser and cache it under `~/.cache/puppeteer`.

But if for some reason we delete the cached browser (e.g., we run `npx puppeteer browsers clear`), and we try to run it the same way as above, it will fail with:

```
Error: Could not find Chrome (ver. 124.0.6367.207). This can occur if either
 1. you did not perform an installation before running the script (e.g. `npx puppeteer browsers install chrome`) or
 2. your cache path is incorrectly configured (which is: /Users/rfigueira/.cache/puppeteer).
For (2), check out our guide on configuring puppeteer at https://pptr.dev/guides/configuration.
    at ChromeLauncher.resolveExecutablePath (/Users/rfigueira/.npm/_npx/9c191fc107e155e1/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ProductLauncher.js:295:27)
    at ChromeLauncher.executablePath (/Users/rfigueira/.npm/_npx/9c191fc107e155e1/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ChromeLauncher.js:209:25)
    at ChromeLauncher.computeLaunchArguments (/Users/rfigueira/.npm/_npx/9c191fc107e155e1/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ChromeLauncher.js:89:37)
    at async ChromeLauncher.launch (/Users/rfigueira/.npm/_npx/9c191fc107e155e1/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ProductLauncher.js:70:28)
    at async #handleLoopback (/Users/rfigueira/dev/tmp/puppeteer-test/node_modules/wrangler/node_modules/miniflare/dist/src/index.js:25418:25)
```

After deleting the browser, and because `npx-import` is already installed, when we run `npxImport("puppeteer@22.8.2")` it will just skip the installation, and therefore it won't call the `postinstall` script associated with that package that actually installs and caches the browser.

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: existing test cover this change
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: [already documented](https://developers.cloudflare.com/changelog/2025-07-22-br-local-dev/)
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: feature improvement


packages/miniflare/src/plugins/browser-rendering/index.ts
```diff
@@ -99,3 +114,117 @@ export const BROWSER_RENDERING_PLUGIN: Plugin<
...
+	const version = "126.0.6478.182";
+	const sessionId = crypto.randomUUID();
+
+	const s = spinner();
+	let startedDownloading = false;
+
+	const { executablePath } = await install({
+		browser,
+		platform,
+		cacheDir: path.join(os.homedir(), ".cache", "miniflare"),
```
ruifigueira: Should I replace this with `xdgAppPaths(".wrangler").cache()` to be in line with `getGlobalWranglerConfigPath`?

petebacondarwin: I guess there is not much value in putting it in `node_modules/.cache` since it would be the same for all projects given a user?

Using XDG would be preferable IMO.

petebacondarwin (approved):
I like this approach! Gets the minimal install (similar to npx-import) for people not using browser-rendering, while having reliable on-demand install for users that do, with caching!

packages/miniflare/src/plugins/browser-rendering/index.ts
ruifigueira: Yes, that browser installation can be shared among multiple projects. I'll change to XDG path.

petebacondarwin (approved):

Issue #10235, opened 2025-08-05
Better Wrangler config diff


penalosa:
### Describe the solution

> For internal context, see https://wiki.cfdata.org/pages/viewpage.action?pageId=1187489699

When using _both_ Wrangler and the dash to update a Worker, it's easy to get into situations where there's a potential conflict. Today, Wrangler warns about those cases in interactive mode, but it doesn't provide much information about _why_ there's a diff.

<img width="767" height="101" alt="Image" src="https://github.com/user-attachments/assets/2db7f381-d257-4a17-b23c-4259d5af1a60" />

We should:
1. Display a rich diff of what the actual config differences are (instead of just a warning that differences exist), so that the user has the opportunity to review. We can re-use the `init --from-dash` codepaths here that can already construct a representation of a Worker's remote state as Wrangler config.
2. If the user chooses _not_ to deploy, we should offer to update their local `wrangler.json` with the remote changes (this will only work for json config files).
3. We should implement an opt-in "strict-mode" that fails deploys with config conflicts in non-interactice mode if a `--force` flag isn't provided. i.e. a user would use `wrangler deploy --strict` as their deploy command.

PR #10314, opened 2025-08-11
add information about dashboard configuration differences on deploys

dario-piotrowicz:
Initial step for https://github.com/cloudflare/workers-sdk/issues/10235

Before:
Screenshot 2025-08-11 at 11 44 57

After:
Screenshot 2025-08-11 at 11 49 26

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: improved DX experience
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: this is (roughly) a new feature so not something that we need to backport


packages/wrangler/src/cloudchamber/helpers/diff.ts
```diff
@@ -50,13 +50,22 @@ type BestPath = {
 	lastComponent?: Result;
 };
 
+// TODO: move this Diff Class to a shared location (since it is not Cloudchamber specific)
```
dario-piotrowicz, line 53: In order not to introduce unnecessary diffs in this PR (and making it much more complex to review) I would move the diff class in a separate followup PR

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+
+	return orderedSource;
+}
+
+/**
+ * Set of config keys that are not represented in the remote representation of a Worker's config.
+ */
+const localOnlyConfigKeys = new Set([
```
dario-piotrowicz: I do not like this.... I think I'd better have a set of the config keys that we include in the remote config representation instead (which are more static and less likely to change)

packages/wrangler/src/init.ts
```diff
@@ -215,20 +215,19 @@ type WorkersDevRes = {
...
+		modified_on: Date;
+	}[];
 };
 
 function isNpm(packageManager: PackageManager) {
 	return packageManager.type === "npm";
 }
 
-async function getWorkerConfig(
+// TODO: move this to a shared location
```
dario-piotrowicz, line 229: Like for `Diff` I'd prefer not to do such refactoring as part of this PR as that would significantly (and with little gain) increase the diffs people would need to review here, if it's ok I'll do it as a followup

.changeset/solid-sloths-drop.md
penalosa, line 1: Can we capitalise sentences here?

packages/wrangler/src/init.ts
```diff
@@ -331,7 +328,7 @@ async function getWorkerConfig(
 					},
 				}
 			: {}),
-		tail_consumers: serviceEnvMetadata.script.tail_consumers,
+		tail_consumers: serviceEnvMetadata.script.tail_consumers ?? undefined,
```
dario-piotrowicz, line 331: note: I've added this because otherwise `tail_consumers` would default to `null` which would not match with the `undefined` default value we have in wrangler

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
```diff
@@ -2026,10 +1997,9 @@ describe("cloudchamber apply", () => {
 			├ EDIT my-container-app
 			│
 			│     [containers.configuration]
-			│     image = \\"${registry}/some-account-id/hello:1.0\\"
+			│     image = \\"registry.cloudflare.com/some-account-id/hello:1.0\\"
```
penalosa, line 2000: Is this intended?

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
```diff
@@ -0,0 +1,186 @@
...
+				tail_consumers: undefined,
+				observability: { enabled: true },
+				limits: undefined,
+			} as unknown as RawConfig,
+			{
+				main: "/Users/dario/Desktop/silent-firefly-dbe3/src/index.js",
+			} as Config
+		);
+
+		expect(`${diff}`).toEqual("");
```
penalosa:
```suggestion
		expect(diff).toEqual("");
```

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
```diff
@@ -0,0 +1,186 @@
+import { getRemoteConfigDiff } from "../../deploy/config-diffs";
+import type { Config, RawConfig } from "../../config";
+
+describe("getRemoteConfigsDiff", () => {
+	it("should handle a very simple diffing scenario (no diffs, random order)", () => {
+		const { diff, onlyAdditionsIfAny } = getRemoteConfigDiff(
+			{
+				name: "silent-firefly-dbe3",
+				main: "/Users/dario/Desktop/silent-firefly-dbe3/src/index.js",
```
penalosa: Is this intended?

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+		expect(`${diff}`).toEqual("");
+		expect(onlyAdditionsIfAny).toBe(true);
+	});
+
+	it("should handle a very simple diffing scenario (some diffs, random order)", () => {
+		const { diff, onlyAdditionsIfAny } = getRemoteConfigDiff(
+			{
+				name: "silent-firefly-dbe3",
+				main: "/Users/dario/Desktop/silent-firefly-dbe3/src/index.js",
```
penalosa: As above

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -372,11 +379,51 @@ export default async function deploy(props: Props): Promise<{
...
+					serviceMetaData.default_environment.environment
 				);
-				if (!(await confirm("Would you like to continue?"))) {
-					return { versionId, workerTag };
+
+				let configDiff: ReturnType<typeof getRemoteConfigDiff> | undefined;
+				let rawConfig: RawConfig | undefined;
+				if (config.configPath) {
+					try {
+						rawConfig = parseRawConfigFile(config.configPath);
```
penalosa: Why is this the raw config? Did diffing against the normalised config present too many problems?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -372,11 +379,51 @@ export default async function deploy(props: Props): Promise<{
...
+					} catch {
+						// We were unable to either read the file or the file is a TOML one
```
penalosa: Why would it throw if the file is TOML?

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -372,11 +379,51 @@ export default async function deploy(props: Props): Promise<{
...
+						);
+					}
+				}
+
+				if (configDiff) {
+					// If there are only additive changes (or no changes at all) there should be no problem,
+					// just using the local config (and override the remote one) should be totally fine
+					if (!configDiff.onlyAdditionsIfAny) {
+						logger.warn(
+							"Your local configuration differs from the remote configuration of your Worker set via the Cloudflare Dashboard, you can see the difference below:" +
```
penalosa:
```suggestion
							"Your local configuration differs from the remote configuration of your Worker set via the Cloudflare Dashboard:" +
```

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -0,0 +1,296 @@
...
+import type { Config, RawConfig } from "../config";
+
+/**
+ * Object representing the difference of two configuration objects.
+ */
+type ConfigDiff = {
+	/** The actual (raw) computed diff of the two objects */
+	diff: Diff;
+	/** Boolean indicating whether the difference is not present or only comprized of the addition of configuration options */
+	onlyAdditionsIfAny: boolean;
```
penalosa: Can we name this a bit more clearly? I think the intention is to mark if this diff is destructive, right? What about `nonDestructive`?

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -0,0 +1,296 @@
...
+ * a Worker's config. This includes resolving some values and setting the default values to some others.
+ *
+ * @param localConfig The local config to normalize
+ * @param localResolvedConfig The resolved local config (used to resolve some values on the raw config)
+ * @returns The local config normalized, ready to be compared to a remote one
+ */
+function normalizeLocalConfigAsRemote(
+	localConfig: RawConfig,
+	localResolvedConfig: Config
+): RawConfig {
```
penalosa: Why does this need a raw config and a resolved one? Can't it just take a `Config` object and transform that?

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -0,0 +1,296 @@
...
+ *    },
+ *	 - compatibility_date: "2025-07-08",
+ *  ```
+ * contains also a modification and a removal
+ *
+ * @param diff The diff object to analyze
+ * @returns true if there are only additions or no diffs at all, false otherwise
+ */
+function configDiffOnlyHasAdditionsIfAny(diff: Diff): boolean {
+	const diffLines = `${diff}`.split("\n");
```
penalosa: Is this implicitly calling `diff.toString()`?

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -0,0 +1,296 @@
...
+ *   +   MY_VAR: "my modified variable",
+ *    },
+ *	 - compatibility_date: "2025-07-08",
+ *  ```
+ * contains also a modification and a removal
+ *
+ * @param diff The diff object to analyze
+ * @returns true if there are only additions or no diffs at all, false otherwise
+ */
+function configDiffOnlyHasAdditionsIfAny(diff: Diff): boolean {
```
penalosa, line 74: Analysing this from the text diff seems fairly complex—can we record this information when actually constructing the diff object?

.changeset/solid-sloths-drop.md
dario-piotrowicz: sure 🙂👍

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
dario-piotrowicz: yes it is intentional, I don't think that using string interpolation in an inline snapshot is correct at all

I can understand the intent but I do not think that it is a correct use of inline snapshots (updating the snapshot will always change this line)

I think that if we do want to use the `registry` variable then we should not use inline snapshots for these checks, or if we do not care about the variable, we can then inline the value as I did here.

Assuming that the Cloudflare registry address is very stable I thought that here it would be correct to inline the value.

@emily-shen do you have some context regarding this by any chance? 🙂

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
dario-piotrowicz: The string interpolation is actually intentional, this check fails without it

Screenshot 2025-08-12 at 12 50 25

if it's clearer I am happy to directly call `diff.toString()` instead here?

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
dario-piotrowicz: mh.... kind of 😅, the value there doesn't really matter so I was thinking of just hardcoding something, but yeah `/Desktop/etc...` is super ugly, I forgot to clean it up a bit 😓

.changeset/solid-sloths-drop.md
```diff
@@ -0,0 +1,12 @@
+---
+"wrangler": minor
+---
+
+add information about dashboard configuration differences on deploys
```
CarmenPopoviciu:
```suggestion
Add information about local vs. dashboard configuration differences on deploys
```
?

packages/wrangler/src/deploy/deploy.ts
dario-piotrowicz: Yes it would likely cause conflict issues with the remote config, but that's not the main reason as to why using the raw config here.

In the diff we present to the users we do want to show them their actual raw configuration and not the values resolved by wrangler.

For example we default various values in the normalized config: https://github.com/cloudflare/workers-sdk/blob/2901224e209b485fb0f4bc5f0bf573b001935821/packages/wrangler/src/config/config.ts#L310-L321
And we do not want to include these in the diff we present to the users. For instance say the user didn't set any KV bindings, if we used the normalized config `kv_namespaces: [],` could likely appear in the diff we present to them even if there hasn't been any KV change, since the diff does include unchanged fields adjacent to changed ones.

TLDR; if we used the normalized config we'd get a bunch of extra stuff (or changed values) that could likely incorrectly end up in the diff output we show to users.

packages/wrangler/src/deploy/deploy.ts
dario-piotrowicz: Sorry this comment was from a previous iteration where I was directly calling `jsonc.parse()`, I updated the code to also include TOMLs but forgot to update the comment, thanks for spotting it! 🙏

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: Sure, let's go with `nonDistructive` 🙂 

it does sound a bit more abstract to me, but I am also not a fan of `onlyAdditionsIfAny` (I could not come up with anything better 😓)

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: Yes, I can directly call `diff.toString()` if you think that that'd be more clear 🙂

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: As mentioned in my other comment (https://github.com/cloudflare/workers-sdk/pull/10314#discussion_r2269623289) I think we need to deal with the raw config so that we can then show a valid diff to users (without extra stuff).

The resolved config gets passed here so that we can resolve the `main` field so that that can be compared with the one the remote config has (and I passed the full `Config` object just in case in the future other resolved fields might be needed).

I am also happy to just remove the `main` field in the check instead (since that doesn't seem super valuable in this local/remote comparison anyways).

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: Yes it is...

the issue here is that I was reusing the pre-existing diffing logic we had, that is very generic and only deals with strings, it is also very lossy (we completely lose the information that two objects are being compared, the diffing logic we have only deals with strings and newlines) so, no I don't think we can record this information when we construct the diff object using the current diffing logic.

I could reimplement the whole diffing logic myself (or use a library that does that) (I've been considering that for a while but then opted into reusing the existing code instead of reinventing the wheel) and avoid the need for this sort of post processing logic.

I am totally happy to reimplement a new "smarter" diffing logic here if preferred (although that would change most of this PR at this point 😅).

.changeset/solid-sloths-drop.md
dario-piotrowicz: sure, thanks 🙂👍

.changeset/solid-sloths-drop.md
```diff
@@ -0,0 +1,12 @@
...
+
+when a user changes the remote configuration of a Worker and then tries to re-deploy using Wrangler they will received a warning mentioning that some config edits have been made via the dashboard and that those will be overridden by the local configuration. They will also be asked if they want to proceed with the deployment or not.
```
CarmenPopoviciu:
```suggestion
When a user changes the remote configuration of a Worker and then tries to re-deploy using Wrangler they will receive a warning mentioning that some config edits have been made via the dashboard and that those will be overridden by the local configuration. They will also be asked if they want to proceed with the deployment or not.
```

this sentence is a bit unclear for me. Is the use case that the local and Dash configs are out of sync, where Dash reflects changes that were made via Dash, and local never "pulled" those changes in, and now attempts to override some of those values?

fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
```diff
@@ -98,7 +98,7 @@ describe("getPlatformProxy - env", () => {
 					"..",
 					"custom-toml",
```
CarmenPopoviciu, line 99: should this also be `custom.toml` then?

.changeset/solid-sloths-drop.md
dario-piotrowicz: Yes, that's exactly the use case...

Sorry I agree that I could have worded this better, I'll try to rewrite it better 🙏

fixtures/get-platform-proxy/tests/get-platform-proxy.env.test.ts
dario-piotrowicz: No, because that's just a directory name 🙂

packages/wrangler/src/deploy/config-diffs.ts
penalosa: Yeah, let's remove the main field from comparison. That will almost certainly be different in most cases

.changeset/solid-sloths-drop.md
dario-piotrowicz: I've updated the text here please have another look and let me know if you find it better 🙏

flakey5 (approved):

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
penalosa: I'd prefer `diff.toString()`, but entirely up to you

packages/wrangler/src/deploy/deploy.ts
penalosa: Sounds good!

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -10692,217 +10694,6 @@ export default{
 		});
 	});
 
-	describe("--outfile", () => {
```
penalosa, line 10695: Why are these tests being removed?

dario-piotrowicz: Sorry I could have done it in a separate PR

But anyways I noticed that this exact same `describe` block is actually present twice in the file:
https://github.com/cloudflare/workers-sdk/blob/1b0b22bca776b4d6b8b50a3e038a78030a49c2dc/packages/wrangler/src/__tests__/deploy.test.ts#L10695
https://github.com/cloudflare/workers-sdk/blob/1b0b22bca776b4d6b8b50a3e038a78030a49c2dc/packages/wrangler/src/__tests__/deploy.test.ts#L10484

As far as I can tell It is the exact same block with the exact same tests, so I assume that this is just a copy-paste mistake, is it not?

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: Removed 🙂

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: renamed 🙂

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
dario-piotrowicz: Fixed 🙂

packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
dario-piotrowicz: ok updated 🙂👍

.changeset/solid-sloths-drop.md
dario-piotrowicz: done 🙂

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: updated to calling `toString()` explicitly

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
dario-piotrowicz: PS: I just checked and saw that we do have the registry address like this here: https://github.com/cloudflare/workers-sdk/blob/2e5b8802b3d9b9b8ff0d44c6b7a5d80e64f92313/packages/wrangler/src/__tests__/containers/deploy.test.ts#L1324c

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -115,57 +109,124 @@ function configDiffOnlyHasAdditionsIfAny(diff: Diff): boolean {
...
+
+		if (!("observability" in localConfig)) {
+			// If `observability` is not in the local config we also remote it from the remote one
+			delete normalizedRemote.observability;
+		}
+
+		if (localConfig.observability === undefined) {
+			// If `observability` is `undefined` in the local config we make sure the same applies to the remote one
+			normalizedRemote.observability = undefined;
+		}
```
penalosa: @Ankcorn can you confirm this is correct? Does wobs now default to on?

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -115,57 +109,124 @@ function configDiffOnlyHasAdditionsIfAny(diff: Diff): boolean {
...
+		localConfig.observability?.logs &&
+		normalizedRemote.observability.logs.enabled === true
 	) {
-		asRemoteConfig.observability.logs.enabled = true;
+		if (!("enabled" in localConfig.observability.logs)) {
+			delete normalizedRemote.observability.logs.enabled;
+		}
+
+		if (localConfig.observability.logs.enabled === undefined) {
+			normalizedRemote.observability.enabled = undefined;
```
penalosa: There's a lot of logic here around observability. Can you add a comment explaining the intention?

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -115,57 +109,124 @@ function configDiffOnlyHasAdditionsIfAny(diff: Diff): boolean {
...
+		if (!("workers_dev" in localConfig)) {
+			// If `workers_dev` is not in the local config we also remote it from the remote one
+			delete normalizedRemote.workers_dev;
+		}
+
+		if (localConfig.workers_dev === undefined) {
+			// If `workers_dev` is `undefined` in the local config we make sure the same applies to the remote one
+			normalizedRemote.workers_dev = undefined;
+		}
 	}
```
penalosa: I don't think this is quite correct. The value of `workers_dev` locally depends on the presence of routes in the local config

penalosa (approved):
Mostly LGTM, couple questions remaining

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: yeah... all of this is pretty ugly... I'll try to clarify things 👍

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: I am not sure I completely follow, could you clarify what you mean? 🙏 

(also note that the local config is the raw user-edited one, not the resolved one)

packages/wrangler/src/__tests__/cloudchamber/apply.test.ts
emily-shen: yeah its fine. strictly speaking registry comes from an env var so there is also a cloudchamber registry url, but i don't think it really matters in this test here.

packages/wrangler/src/__tests__/deploy.test.ts
penalosa: Ah, fair enough!

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: As discussed this is a different issue that we'll have to deal with (most likely via a separate warning)

dario-piotrowicz: https://github.com/cloudflare/workers-sdk/issues/10360 🙂

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -0,0 +1,370 @@
...
+}
+
+/**
+ * Set of config keys that are part of the remote Worker's config representation we construct.
+ *
+ * Or in other words the following contains all the possible fields that can be present in the
+ * object returned from the `downloadWorkerConfig` function
+ *
+ */
+const remoteConfigKeys = new Set<string>([
```
CarmenPopoviciu, line 322: how can we make sure that when new bindings are added, we don't forget to update this list?

CarmenPopoviciu (approved):
amazing work Dario! 🚀

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: that's a really good question.... sorry I don't really have a great answer here... as a followup I can look into improving this and adding some unit tests to make sure that every config option is considered, so to prevent someone from adding a new binding but forgetting to add that to the list, would that work for you? 🙂 

<details><summary>Possible Test</summary>
<p>

I don't really have a super clear picture in mind, but as testing maybe here we could have 2 sets `remoteConfigKey` and `localConfigKeys` and test that they contains all the possible keys that can appear in a resolved config object? or something to that extent? 🤔 (so that if someone adds a new option to the resolved config but forgets to add it to either of the two sets a unit test will fail...)

</p>
</details>

dario-piotrowicz: I've created https://github.com/cloudflare/workers-sdk/issues/10373 to not to forget about this 🙂

PR #10346, opened 2025-08-13
test: further efforts to reduce flakes on email bindings

petebacondarwin:
More efforts to reduce flaky tests

### DIFF
--- a/packages/miniflare/test/plugins/email/index.spec.ts
+++ b/packages/miniflare/test/plugins/email/index.spec.ts
@@ -75,20 +75,21 @@ test("Unbound send_email binding works", async (t) => {
 	);
 	t.is(await res.text(), "ok");
 	t.is(res.status, 200);
-	waitFor(async () =>
-		t.true(
-			log.logs.some(
-				([type, message]) =>
-					type === LogLevel.INFO &&
-					message.includes(
-						"send_email binding called with the following message:"
-					)
-			)
-		)
-	);
-
-	const file = log.logs[1][1].split("\n")[1].trim();
-	t.is(await readFile(file, "utf-8"), email);
+	await waitFor(async () => {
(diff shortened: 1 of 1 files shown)


packages/miniflare/test/plugins/email/index.spec.ts
```diff
@@ -75,20 +75,21 @@ test("Unbound send_email binding works", async (t) => {
...
+				type === LogLevel.INFO &&
+				message.match(/send_email binding called with the following message:\n/)
+		);
+		if (!entry) {
+			throw new Error(
+				"send_email binding log not found in " +
+					JSON.stringify(log.logs, null, 2)
+			);
+		}
+		const file = entry[/* message */ 1].split("\n")[1].trim();
```
petebacondarwin, line 90: Is there a chance that the second line of the message appears in a subsequent log entry?

petebacondarwin: The logs come from here:

https://github.com/cloudflare/workers-sdk/blob/71c5bfd25e820af8443f3d5c8bce349c699e882e/packages/miniflare/src/workers/email/send_email.worker.ts#L86

.github/workflows/test-and-check.yml
```diff
@@ -147,17 +147,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
 
-      - name: Check for Wrangler build cache
```
CarmenPopoviciu, line 150: why are we removing this?

CarmenPopoviciu: talked async. this is ok to remove

petebacondarwin: this causes unwanted stuff to be dumped to the logs

PR #10354, opened 2025-08-13
feat: enable cross-process communication for wrangler dev with multiple configs

edmundhung:
Fixes [DEVX-2125](https://jira.cfdata.org/browse/DEVX-2125)

This enables dev registry support when running `wrangler dev` with multiple config files.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24522
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: miniflare dev registry is v4 only

### DIFF
--- a/fixtures/dev-registry/tests/dev-registry.test.ts
+++ b/fixtures/dev-registry/tests/dev-registry.test.ts
@@ -741,7 +741,10 @@ describe("Dev Registry: vite dev <-> wrangler dev", () => {
 			);
 		});
 
-		await runWranglerDev("wrangler.service-worker.jsonc", devRegistryPath);
+		await runWranglerDev(
+			["wrangler.service-worker.jsonc", "wrangler.worker-entrypoint.jsonc"],
+			devRegistryPath
+		);
 
 		// Test vite dev -> wrangler dev
 		await vi.waitFor(async () => {
@@ -955,7 +958,10 @@ describe("Dev Registry: getPlatformProxy -> wrangler / vite dev", () => {
 		});
 
 		await runWranglerDev(
-			"wrangler.worker-entrypoint-with-assets.jsonc",
+			[
+				"wrangler.worker-entrypoint-with-assets.jsonc",
+				"wrangler.internal-durable-object.jsonc",
+			],
 			devRegistryPath
 		);
 


packages/wrangler/src/dev.ts
```diff
@@ -626,37 +625,41 @@ export async function startDev(args: StartDevOptions) {
...
+			) {
+				process.send(
+					JSON.stringify({
+						event: "DEV_SERVER_READY",
+						ip: url.hostname,
+						port: parseInt(url.port),
+					})
+				);
+			}
+		});
```
edmundhung, line 658: I moved this out so that `runWranglerDev` can wait for the IPC message properly when providing multiple configs.

fixtures/dev-registry/workers/module-worker.ts
```diff
@@ -1,5 +1,17 @@
+import { WorkerEntrypoint } from "cloudflare:workers";
+
 let tailEvents = [];
 
+export class NamedEntrypoint extends WorkerEntrypoint {
+	ping() {
+		return "Pong from Named Entrypoint";
```
dario-piotrowicz, line 7:
```suggestion
		return "Ping from Named Entrypoint";
```

dario-piotrowicz (commented):
Amazing stuff 🫶 

Anyways, I think we don't have any test that shows that multi config wrangler dev workers can communicate with vite workers, do we? Would that by a big lift to test?

fixtures/dev-registry/workers/module-worker.ts
edmundhung: It was meant to be just ping <> pong message but I just added an extra message to include where the pong is from. 🤔 Curious why ping here?

edmundhung:
> Anyways, I think we don't have any test that shows that multi config wrangler dev workers can communicate with vite workers, do we? Would that by a big lift to test?

I felt adding a whole new set of tests just for multi-config would be repetitive, so I updated some of the existing tests in 5caa733 to use a multi-config setup instead. That way we still get a bit more coverage without duplicating too much.

fixtures/dev-registry/workers/module-worker.ts
dario-piotrowicz: The function is called `ping` so I assumed that the message was supposed to be `Ping` no? 🤔 

Mh... sorry probably I just misunderstood the intention here, the called calls the `ping` method and gets a response with `Pong` and that's idea for ping <-> pong right? 🤔

edmundhung: Correct. If you think this confusing, I can change it.

dario-piotrowicz: nono sorry I just I was just confusing myself when reading it 😅

dario-piotrowicz (approved):
Looks good to me 😄

fixtures/dev-registry/workers/module-worker.ts
dario-piotrowicz: (also I noticed that this was already set up 😅 😓)

PR #10357, opened 2025-08-13
Update templates to use default entrypoint handlers for Python.

dom96:
Python Workers has a new compat flag which becomes default on 2025-08-14. At which point the global handlers won't be supported in Python anymore. This PR updates all templates to use the new way of defining the handlers.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24392
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: beta feature

### DIFF
--- a/packages/create-cloudflare/templates/hello-world-durable-object-with-assets/py/src/entry.py
+++ b/packages/create-cloudflare/templates/hello-world-durable-object-with-assets/py/src/entry.py
@@ -49,15 +49,15 @@ class MyDurableObject(DurableObject):
 * @returns {Promise<Response>} The response to be sent back to the client
 """
 class Default(WorkerEntrypoint):
-    async def fetch(self, request, env, ctx):
+    async def fetch(self, request):
         # Create a `DurableObjectId` for an instance of the `MyDurableObject`
         # class named "foo". Requests from all Workers to the instance named
         # "foo" will go to a single globally unique Durable Object instance.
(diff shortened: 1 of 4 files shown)


mikenomitch (approved):

CarmenPopoviciu:
tests seem to be legitimately failing

dom96:
Okay, I think I fixed the failing test. Any ideas why the E2E is failing?

CarmenPopoviciu:
> Okay, I think I fixed the failing test. Any ideas why the E2E is failing?

they are failing with `Uncaught TypeError: Cannot define multiple default entrypoints`. See https://github.com/cloudflare/workers-sdk/actions/runs/16963372772/job/48081306269?pr=10357#step:7:523

dom96:
Oh interesting, thanks for linking. I think that's because the compat date is 2025-08-13. How do these tests determine the compat date to use?

CarmenPopoviciu:
> 2025-08-13.

It's coming from here: https://github.com/cloudflare/workers-sdk/blob/9b097518456fecee5eb0fab1f56d3a269e8bdfc5/packages/create-cloudflare/src/wrangler/config.ts#L197-L209, and based on the logic there you could set a specific compat date in your template if needed. Or perhaps use a compat flag instead, in case that works better? That would be a stronger signal to users that the template relies on a specific compat flag vs compat date.

PR #10380, opened 2025-08-15
chore: improvements to remote-bindings tests to reduce flakes

petebacondarwin:
just test stuff


packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
```diff
@@ -172,4 +179,83 @@ export class WranglerE2ETestHelper {
...
+			{ timeout: 10_000, interval: 500 }
+		);
+
+		const cleanup = async () => {
+			await this.run(`wrangler delete --name ${workerName} --force`);
+		};
+
+		try {
+			onTestFinished(cleanup, 15_000);
+		} catch {
```
dario-piotrowicz: Nit: maybe we could be more selective and only handle the expected error (instead of suppressing them all) 🤔

packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
```diff
@@ -172,4 +179,83 @@ export class WranglerE2ETestHelper {
...
+		extraFlags?: string[];
+	}) {
+		const configOption = configPath ? `-c ${configPath}` : "";
+		const workerNameOption = `--name ${workerName}`;
+		const { stdout } = await this.run(
+			`wrangler deploy ${entryPoint} ${workerNameOption} ${configOption} --compatibility-date 2025-01-01 ${extraFlags.join(" ")}`
+		);
+
+		const match = stdout.match(
+			/(?<url>https:\/\/tmp-e2e-.+?\..+?\.workers\.dev)/
```
dario-piotrowicz: Nit: if we're using a specific convention maybe we should assert that for the name of the worker above 🤔 

(PS: I am just a bit worried of someone at some point calling this function with a non-`tmp-e2e` name and wasting some time understanding why this doesn't work)

(even clearer, maybe could create a regex using `workerName` to be more generic/robust 🤔)

packages/vite-plugin-cloudflare/e2e/remote-bindings.test.ts
```diff
@@ -1,7 +1,8 @@
 import assert from "node:assert";
+import { randomUUID } from "node:crypto";
```
dario-piotrowicz: I think the additions here are incorrect/unintentional, these imports are indeed not being used anyways:
Screenshot 2025-08-21 at 22 09 24

fixtures/vitest-pool-workers-remote-bindings/run-tests.mjs
```diff
@@ -133,3 +173,22 @@ function getAuthenticatedEnv() {
...
+	const start = Date.now();
+	while (true) {
+		try {
+			return callback();
+		} catch (error) {
+			if (Date.now() < start + 15_000) {
+				throw error;
+			}
+		}
+		await setTimeout(1000);
```
dario-piotrowicz, line 194: Nit:
```suggestion
		await setTimeout(1_000);
```

For consistency with the `15_000` above

fixtures/vitest-pool-workers-remote-bindings/run-tests.mjs
```diff
@@ -58,12 +60,19 @@ const deployOut = execSync(
...
+		encoding: "utf8",
 	}
 );
+
 if (!new RegExp(`Deployed\\s+${remoteWorkerName}\\b`).test(`${deployOut}`)) {
 	throw new Error(`Failed to deploy ${remoteWorkerName}`);
 }
 
+const deployedUrl = deployOut.match(
+	/(?<url>https:\/\/tmp-e2e-.+?\..+?\.workers\.dev)/
```
dario-piotrowicz: Like my other comment I wonder if it would be more generic and robust to use `remoteWorkerName` here 🤔

dario-piotrowicz (approved):

packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
petebacondarwin: I switched to an opt-in approach that doesn't rely on catching errors.

packages/wrangler/e2e/helpers/e2e-wrangler-test.ts
```diff
@@ -172,4 +180,114 @@ export class WranglerE2ETestHelper {
...
+		const cleanup = async () => {
+			await this.run(`wrangler delete --name ${workerName} --force`);
+		};
+
+		if (cleanOnTestFinished) {
+			try {
+				onTestFinished(cleanup, 15_000);
+			} catch (e) {
+				throw new Error(
+					`Failed to register cleanup for worker.\nPerhaps you called this outside an \`it\` block?\nIf so, pass \`cleanOnTestFinished: false\` and then use the returned \`cleanup\` helper yourself`,
```
dario-piotrowicz: nit:
```suggestion
					"Failed to register cleanup for worker.\nPerhaps you called this outside an `it` block?\nIf so, pass `cleanOnTestFinished: false` and then use the returned `cleanup` helper yourself",
```

.github/workflows/e2e.yml
```diff
@@ -83,7 +91,8 @@ jobs:
           - os: windows-2022
             description: v20, Windows
             node: 20.19.1
-          - os: ubuntu-22.04-arm
+            # we need to use an amd image to run the containers tests, since we build for linux/amd64
+          - os: ubuntu-22.04
```
dario-piotrowicz: above you bumped `ubuntu-22.04` to `ubuntu-24.04` shall we do the same here?

PR #10393, opened 2025-08-18
use resolved local config for remote deploy config diffing logic

### LINKED ISSUE #10377: In deploy config diffs use resolved local config
https://github.com/cloudflare/workers-sdk/pull/10314 introduced a diffing mechanism that compares the local user authored wrangler config file with the remote representation of the worker's configuration.

There are however a few issues in using the user authored wrangler config for such diffing logic:
 - some fields not present in the file can be resolved to different values (e.g. `workers_dev` or the worker' name when targeting an environment)
 - if the deployment targets a specific environment then the local config file won't be comparable with the local remote configuration (which does target a specific environment)
 - resolved default values that will be deployed (thus will appear in the remote config on the next check) need to be ignored
 - it is difficult to make this logic format independent (e.g. if a toml file is used ad-hoc logic would need to be introduced to handle it, or (as is currently done) the representation will still be json anyways)

For the reasons above it would be better if we updated the diffing logic so that the resolved local config is used instead.

Naturally this should also be clearly conveyed to users.

dario-piotrowicz:
Fixes #10377

This change updates the remote deploy diffing logic to use the local resolved config instead of using the raw one (since that approach turned out to be flawed)

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  the remote deploy diffing is not present in v3

### DIFF
--- a/packages/wrangler/src/deploy/config-diffs.ts
+++ b/packages/wrangler/src/deploy/config-diffs.ts
@@ -114,6 +114,13 @@ function normalizeLocalResolvedConfigAsRemote(
 	return normalizedConfig;
 }
 
+/**
+ * Normalizes an observability config object (either the remote or resolved local one) to a fully filled form, this
+ * helps us resolve any inconsistencies between the local and remote default values.
+ *
+ * @param obs The observability config object to normalize
+ * @returns The normalized observability object
+ */
 function normalizeObservability(
 	obs: RawConfig["observability"]
 ): Config["observability"] {


packages/wrangler/src/__tests__/deploy/config-diffs.test.ts
```diff
@@ -107,18 +106,12 @@ describe("getRemoteConfigsDiff", () => {
 			+     \\"head_sampling_rate\\": 1,
 			+     \\"logs\\": {
 			+       \\"head_sampling_rate\\": 0.95,
-			+       \\"invocation_logs\\": true
+			+       \\"invocation_logs\\": true,
+			+       \\"enabled\\": true
```
edmundhung: ~~I am not familiar with the logic here. Is this line expected? I don't seems to see it showing on any of the provided config above.~~ I understand now.

packages/wrangler/src/deploy/config-diffs.ts
```diff
@@ -108,6 +111,57 @@ function configDiffOnlyHasAdditionsIfAny(diff: Diff): boolean {
...
+			normalizedConfig.observability.logs.head_sampling_rate = 1;
+		}
+
+		if (!("invocation_logs" in normalizedConfig.observability.logs)) {
+			normalizedConfig.observability.logs.invocation_logs = true;
+		}
+	}
+
+	return normalizedConfig;
+}
```
edmundhung: I wonder if this is better be done in `normalizeAndValidateConfig` directly:

https://github.com/cloudflare/workers-sdk/blob/516dc20be624042533af0cd0b4205f92e256bb3f/packages/wrangler/src/config/index.ts#L101C34-L106

That way we can share the same default value as in the dashboard everywhere.

dario-piotrowicz: I think that `workers_dev` shouldn't be done there since its resolution is specific to the deploy command

(
see: https://github.com/cloudflare/workers-sdk/blob/0c04da9b3a8dcf1220b46a0fdd463ba0bad0f9a1/packages/wrangler/src/triggers/deploy.ts#L55 where `routes` comes from:
https://github.com/cloudflare/workers-sdk/blob/0c04da9b3a8dcf1220b46a0fdd463ba0bad0f9a1/packages/wrangler/src/triggers/deploy.ts#L41-L42)

But the other normalizations/default settings yes I think you're right that they should be done as part of `normalizeAndValidateConfig` 🤔 👍

.changeset/tender-beds-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+use resolved local config for remote deploy config diffing logic
```
penalosa:
```suggestion
Use resolved local config for remote deploy config diffing logic
```

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: Done, I've moved the defaulting logic to `normalizeAndValidateConfig` 😄, please have another look 🙏

dario-piotrowicz: I'm working though moving the logic into `normalizeAndValidateConfig`, unfortunately I am hitting a bit of a road block here: https://github.com/cloudflare/workers-sdk/blob/7c339aeb0392e41b9a306c84538950f32c9a0dd4/packages/wrangler/src/__tests__/containers/deploy.test.ts#L1396

I'll try to see if there's a workaround 😓

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -12683,29 +12683,6 @@ export default{
...
-			await runWrangler("deploy index.js");
-			expect(std.out).toMatchInlineSnapshot(`
-				"Total Upload: xx KiB / gzip: xx KiB
-				Worker Startup Time: 100 ms
-				Uploaded test-name (TIMINGS)
-				Deployed test-name triggers (TIMINGS)
-				  https://test-name.test-sub-domain.workers.dev
-				Current Version ID: Galaxy-Class"
-			`);
-		});
```
edmundhung: Why is this test removed?

dario-piotrowicz: because I _think_ that it incorrect, indeed I am updating the observability to default to `enabled`, which, from manually deploying workers it definitely seem to be the case for me.

Instead of removing the test I could update it to check that by default we do enable it 🤔 

In either case I really appreciate the extra look on this... I wonder if the test was wrong or if me trying to default to `enabled` is 🤔

edmundhung: Just for my understanding: I don't see your changes enabling `observability` by default. So if the users don't have it configured, it should still be `enabled: false`, right?

dario-piotrowicz: Sorry I want the completely wrong direction, please ignore these changes 😓

packages/wrangler/src/deploy/config-diffs.ts
dario-piotrowicz: I've updated this logic as it was a bit off.

However regarding moving it to the `normalizeAndValidateConfig` I no longer think that that's necessarily appropriate since we're trying to make the local config similar to the remote one, but doing that in the `normalizeAndValidateConfig` function could potentially change how values are resolved, potentially causing unexpected behaviors.

Also this is very much geared towards aligning the local config to the remote one and trying to follow what the remote one does. If the remote one were to change later on, updating the logic here would be all we'd need to do in order to support that change, but if this logic was embedded into `normalizeAndValidateConfig` updating it could be more risky/potentially breaky.

what do you think? 🤔

dario-piotrowicz: As agreed async we decided to keep this logic only in here 🙂

edmundhung (approved):

PR #10402, opened 2025-08-18
fix: sanitize double-slashes in asset-worker redirects

danielrs:
A redirect rule such as:

```
/foo/* /:splat
```

And a request such as:

```
https://example.com/foo//google.com
```

Would result in a redirect target:

```
//google.com
```

Which would incorrectly resolve to:

```
https://google.com
```

When constructing a URL.

Problem was already patched in the live service.

_Describe your change..._

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a Wrangler change.

### DIFF
--- a/packages/workers-shared/asset-worker/src/utils/rules-engine.ts
+++ b/packages/workers-shared/asset-worker/src/utils/rules-engine.ts
@@ -140,10 +140,23 @@ export const generateRedirectsMatcher = (
 		configuration.redirects.version === REDIRECTS_VERSION
 			? configuration.redirects.rules
 			: {},
-		({ status, to }, replacements) => ({
-			status,
-			to: replacer(to, replacements),
-		})
+		({ status, to }, replacements) => {
+			const target = replacer(to, replacements).trim();
+			const protoPattern = /^(\w+:\/\/)/;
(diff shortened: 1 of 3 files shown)


packages/workers-shared/asset-worker/src/utils/rules-engine.ts
```diff
@@ -140,10 +140,23 @@ export const generateRedirectsMatcher = (
 		configuration.redirects.version === REDIRECTS_VERSION
 			? configuration.redirects.rules
 			: {},
-		({ status, to }, replacements) => ({
-			status,
-			to: replacer(to, replacements),
-		})
+		({ status, to }, replacements) => {
+			const target = replacer(to, replacements).trim();
+			const protoPattern = /^(\w+:\/\/)(.*)/;
```
dario-piotrowicz:
```suggestion
			const protoPattern = /^\w+:\/\//;
```
I think the regex can be simplified 🤔

danielrs: Good catch!

PR #10414, opened 2025-08-20
Update config after creation commands

penalosa:
Fixes https://github.com/cloudflare/workers-sdk/issues/9730

When running a resource creation command (like `wrangler kv namespace create NAME`), Wrangler will now offer to update your config file for you. Some notes:
- This does not work for preview KV namespaces, as we don't know the right main KV ID to use
- This _does_ work for environments, and will update the relevant resource field in the given environment (i.e. `--env test`)
- Wrangler will prompt with 3 options: Yes, Yes but change the binding name (which will then present a second dialog to choose the binding name), or No
- This will not work in non-interactive contexts like CI
- This will not work when running Wrangler in a context that can't find a config file
- Updating the config file only works for jsonc format

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: this doesn't need additional documentation
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/packages/wrangler/src/__tests__/__snapshots__/r2.test.ts.snap
+++ b/packages/wrangler/src/__tests__/__snapshots__/r2.test.ts.snap
@@ -13,7 +13,7 @@ To access your new R2 Bucket in your Worker, add the following snippet to your c
   ]
 }
 ? Would you like Wrangler to add it on your behalf?
-🤖 Using fallback value in non-interactive context: no"
+🤖 Using fallback value in non-interactive context: No"
 `;
 
 exports[`r2 > bucket > create > wrangler.json > should create a bucket with the expected default storage class 1`] = `
@@ -29,7 +29,7 @@ To access your new R2 Bucket in your Worker, add the following snippet to your c
   ]
 }
 ? Would you like Wrangler to add it on your behalf?
-🤖 Using fallback value in non-interactive context: no"
+🤖 Using fallback value in non-interactive context: No"
 `;
 
 exports[`r2 > bucket > create > wrangler.json > should create a bucket with the expected jurisdiction 1`] = `
@@ -45,7 +45,7 @@ To access your new R2 Bucket in your Worker, add the following snippet to your c
   ]
 }
 ? Would you like Wrangler to add it on your behalf?
-🤖 Using fallback value in non-interactive context: no"
+🤖 Using fallback value in non-interactive context: No"
 `;
 
 exports[`r2 > bucket > create > wrangler.json > should create a bucket with the expected location hint 1`] = `
@@ -61,7 +61,7 @@ To access your new R2 Bucket in your Worker, add the following snippet to your c
   ]
 }
 ? Would you like Wrangler to add it on your behalf?
-🤖 Using fallback value in non-interactive context: no"
+🤖 Using fallback value in non-interactive context: No"
 `;
 
 exports[`r2 > bucket > create > wrangler.toml > should create a bucket & check request inputs 1`] = `
--- a/packages/wrangler/src/__tests__/__snapshots__/r2.test.ts.snap
+++ b/packages/wrangler/src/__tests__/__snapshots__/r2.test.ts.snap
@@ -3,9 +3,7 @@
 exports[`r2 > bucket > create > wrangler.json > should create a bucket & check request inputs 1`] = `
 "Creating bucket 'test-bucket'...
 ✅ Created bucket 'test-bucket' with default storage class of Standard.
-
-Configure your Worker to write objects to this bucket:
-
+To access your new R2 Bucket in your Worker, add the following snippet to your configuration file:
 {
   \\"r2_buckets\\": [
     {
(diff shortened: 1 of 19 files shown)


packages/wrangler/src/__tests__/__snapshots__/r2.test.ts.snap
```diff
@@ -3,73 +3,71 @@
...
 {
   \\"r2_buckets\\": [
     {
       \\"bucket_name\\": \\"test-bucket\\",
       \\"binding\\": \\"test_bucket\\"
     }
   ]
-}"
+}
+? Would you like Wrangler to add it on your behalf?
```
dario-piotrowicz, line 15: Sorry... this is completely completely unrelated, but probably as a followup, I think it's be nicer to use an emoji for the question mark (like❔ for example), what do you think? 🙂

packages/wrangler/src/__tests__/update-config-file.test.ts
```diff
@@ -0,0 +1,282 @@
...
+		clearDialogs();
+	});
+	// mockSelect({
+	// 	text: "Would you like Wrangler to add it on your behalf?",
+	// 	result: "yes-but",
+	// });
+	// mockPrompt({
+	// 	text: "What binding name would you like to use?",
+	// 	result: "HELLO",
+	// });
```
dario-piotrowicz: this is not intentional, is it?

packages/wrangler/src/dialogs.ts
```diff
@@ -104,16 +105,17 @@ export async function select<Values extends string>(
 	options: SelectOptions<Values>
 ): Promise<Values> {
 	if (isNonInteractiveOrCI()) {
-		if (options?.defaultOption === undefined) {
+		const fallback = options.fallbackOption ?? options.defaultOption;
```
dario-piotrowicz: I find having both the `fallbackOption` and the `defaultOption` a bit confusing 😓

what about instead having `initialOption?: string` and `fallbackOption?: string` and use `initialOption` on line 126 and just the `fallbackOption` one here?

penalosa: The `fallbackOption` is following the pattern we use for `confirm()` above, where the default value can be different from what should be selected in CI. The only reason I added `?? options.defaultOption` here was to not break any callers who relied on the old behaviour, but I suppose I could go through and update them all instead?

penalosa:
The TOML case is tested here: https://github.com/cloudflare/workers-sdk/blob/cc24a5827b43d9304de8b8d9cdd7a02971fe557d/packages/wrangler/src/__tests__/update-config-file.test.ts#L194, howing no prompts & no config updates. I could add an additional test for a file with comments, but as you say I think it might be a bit redundant given `experimental_patchConfig()` is already tested with that case.

packages/wrangler/src/dialogs.ts
dario-piotrowicz: Ah ok, thanks for the context, I agree that it's good to stay consistent with the pattern in `confirm` 🤔

I do find the names `default` and `fallback` a bit confusing (as they to me sound like almost synonyms), but it's not a huge deal, I am happy to keep them as they are 🙂

dario-piotrowicz:
> The TOML case is tested here:

Sorry, somehow I missed that! 🤦 🙇 

> I could add an additional test for a file with comments, but as you say I think it might be a bit redundant given `experimental_patchConfig()` is already tested with that case.

I don't have a strong opinion on it either, if you think it's unnecessary I'm ok avoiding it 🙂👍

packages/wrangler/src/config/index.ts
```diff
@@ -61,6 +65,54 @@ export function formatConfigSnippet(
...
+) {
+	const resource = Object.keys(snippet())[0] as keyof CfWorkerInit["bindings"];
+	const envString = env ? ` in the "${env}" environment` : "";
+	logger.log(
+		`To access your new ${friendlyBindingNames[resource]} in your Worker, add the following snippet to your configuration file${envString}:`
+	);
+
+	logger.log(formatConfigSnippet(snippet(), configPath));
+
+	if (configPath && offerToUpdate && configFormat(configPath) === "jsonc") {
```
emily-shen, line 84: why not `json`?

packages/wrangler/src/config/index.ts
```diff
@@ -61,6 +65,54 @@ export function formatConfigSnippet(
...
+	snippet: (
+		bindingName?: string
+	) => Partial<{ [K in keyof CfWorkerInit["bindings"]]: RawConfig[K] }>,
+	configPath: Config["configPath"],
+	env: string | undefined,
+	offerToUpdate: boolean = true
+) {
+	const resource = Object.keys(snippet())[0] as keyof CfWorkerInit["bindings"];
+	const envString = env ? ` in the "${env}" environment` : "";
+	logger.log(
```
emily-shen, line 78: should we log this after we offer to update and we can't/are declined?

packages/wrangler/src/config/index.ts
emily-shen: oh sorry i see that includes json as well

packages/wrangler/src/config/index.ts
penalosa: I went with before just so we're giving people as much info as possible up front, but can definitely change if you think that's too much. What would be the benefit of logging it later?

emily-shen: eh actually i think i like the way it is currently. carry on.

emily-shen (approved):

PR #10416, opened 2025-08-20
feat(vite-plugin): Add support for email triggers

CarmenPopoviciu:
Fixes DEVX-2114

Screenshot 2025-08-20 at 14 17 16

Screenshot 2025-08-22 at 10 30 02

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24618
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -1080,6 +1080,39 @@ if (import.meta.hot) {
...
+			name: "vite-plugin-cloudflare:trigger-handlers",
+			async configureServer(viteDevServer) {
+				assertIsNotPreview(resolvedPluginConfig);
+
+				const entryWorkerConfig = getEntryWorkerConfig(resolvedPluginConfig);
+				assert(entryWorkerConfig, `No entry Worker config`);
+
+				const entryWorkerName = entryWorkerConfig.name;
+
+				viteDevServer.middlewares.use((req, res, next) => {
```
```suggestion
				viteDevServer.middlewares.use("/cdn-cgi/", (req, res, next) => {
```

If we add the [route](https://github.com/senchalabs/connect?tab=readme-ov-file#appuseroute-fn) here then this middleware will only be invoked for routes that start with `/cdn-cgi/`. Adding the header and forwarding the request should then work for `email` and `scheduled` handlers.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -1080,6 +1080,39 @@ if (import.meta.hot) {
...
+					const requestHandler = createRequestHandler((request) => {
+						assert(miniflare, `Miniflare not defined`);
+
+						// email triggers
+						if (url.pathname === "/cdn-cgi/handler/email") {
+							// set the target service that handles these requests
+							// to point to the User Worker
+							// see `getTargetService` fn in packages/miniflare/src/workers/core/entry.worker.ts
+							request.headers.set(CoreHeaders.ROUTE_OVERRIDE, entryWorkerName);
+						}
```
```suggestion
							// set the target service that handles these requests
							// to point to the User Worker
							// see `getTargetService` fn in packages/miniflare/src/workers/core/entry.worker.ts
							request.headers.set(CoreHeaders.ROUTE_OVERRIDE, entryWorkerName);
```

Don't think we need the condition here as this should always apply.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -1080,6 +1080,39 @@ if (import.meta.hot) {
 				}
 			},
 		},
+		// Plugin to handle cron/email/etc triggers
+		{
+			name: "vite-plugin-cloudflare:trigger-handlers",
```
jamesopstad, line 1071:
```suggestion
			name: "vite-plugin-cloudflare:trigger-handlers",
			enforce: "pre",
```

We need `enforce: "pre"` here so that this runs ahead of the other middleware.

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: that is what I originally did, but ran into the following issue...It seems that when we add middleware for a specific route, vite "extracts" that route from the request url, making the `req.url` value `http://localhost:5173/handler/email` as opposed to `http://localhost:5173/cdn-cgi/handler/email` what itoriginally was. It's worth noting here that the original pathname is reflected in `req.originalUrl` which is `/cdn-cgi/handler/email` in our case.

That means that when we invoke `createRequestHandler`, which in turn calls `const request = createRequest(req, res);`, the resulting request that is then passed to Miniflare, incorrectly reflects the `http://localhost:5173/handler/email` value. `createRequest` is imported from `@mjackson/node-fetch-server`, and I'm assuming that lib looks at `req.url` not `req.originalUrl`, which gets us into the situation I described.

I'm not sure if we can tell vite not to extract that route from the req url...I couldn't find anything in the docs, so I eventually resorted to the logic I have in my PR now. But I'm all ears if I missed smth and we could actually make your suggestion work

jamesopstad: Good point. Can you just do `req.url = req.originalUrl` at the top of the function?

CarmenPopoviciu: we can, but I'd have to reconstruct the whole url since `originalUrl` only reflects the `pathname`, and not the host, port and protocol :/. It's a strange decision to have `.url` reflect the whole url, but `.originalUrl` to reflect only the pathname

jamesopstad: I think both only include the pathname.

CarmenPopoviciu: oh...you are right...my mistake. I was logging too many things, and misread the url value 😓 . Thx for catching that!

CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/src/index.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -1080,6 +1080,43 @@ if (import.meta.hot) {
...
+						// to point to the User Worker (see `getTargetService` fn in
+						// `packages/miniflare/src/workers/core/entry.worker.ts`)
+						request.headers.set(CoreHeaders.ROUTE_OVERRIDE, entryWorkerName);
+						return miniflare.dispatchFetch(request, { redirect: "manual" });
+					});
+
+					// `req.url` is the URL of the request relative to the middleware
+					// mount path. Here we ensure that miniflare receives a request that
+					// reflects the original request url
+					// 🚨🚨🚨 A potential side effect of mutating the original request is
```
CarmenPopoviciu: not sure how concerning this is, but added a comment for posterity explaining why we do this and what possible side effects are. Another option would have been to create a new `IncomingMessage`-like object, but that seemed to be rather complex, so I chose for this option instead. Happy to switch approach though if there are major concerns

.changeset/wide-donuts-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": minor
+---
+
+Add support for email triggers in `vite dev` and `vite preview`
```
jamesopstad:
```suggestion
Add support for testing [Cron Triggers](https://developers.cloudflare.com/workers/configuration/cron-triggers/#test-cron-triggers-locally) and [Email Workers](https://developers.cloudflare.com/email-routing/email-workers/local-development/) in `vite dev` and `vite preview`.
```

Also, it does look like these docs should be updated to mention `vite dev` alongside `wrangler dev` if you have time.

packages/vite-plugin-cloudflare/playground/cron-triggers/package.json
```diff
@@ -0,0 +1,19 @@
...
+	"scripts": {
+		"build": "vite build",
+		"check:types": "tsc --build",
+		"dev": "vite dev",
+		"preview": "vite preview"
+	},
+	"devDependencies": {
+		"@cloudflare/vite-plugin": "workspace:*",
+		"@cloudflare/workers-tsconfig": "workspace:*",
+		"@cloudflare/workers-types": "^4.20250803.0",
```
jamesopstad:
```suggestion
		"@cloudflare/workers-types": "catalog:default",
```

packages/vite-plugin-cloudflare/playground/cron-triggers/wrangler.jsonc
```diff
@@ -0,0 +1,9 @@
+{
+	"name": "cron-trigger-worker",
+	"main": "src/index.ts",
+	"compatibility_date": "2025-02-14",
+	"compatibility_flags": ["nodejs_compat"],
```
jamesopstad:
```suggestion
```

packages/vite-plugin-cloudflare/playground/email-worker/package.json
```diff
@@ -0,0 +1,22 @@
...
+	"scripts": {
+		"build": "vite build",
+		"check:types": "tsc --build",
+		"dev": "vite dev",
+		"preview": "vite preview"
+	},
+	"devDependencies": {
+		"@cloudflare/vite-plugin": "workspace:*",
+		"@cloudflare/workers-tsconfig": "workspace:*",
+		"@cloudflare/workers-types": "^4.20250803.0",
```
jamesopstad:
```suggestion
		"@cloudflare/workers-types": "catalog:default",
```

packages/vite-plugin-cloudflare/playground/email-worker/wrangler.jsonc
```diff
@@ -0,0 +1,12 @@
+{
+	"name": "email-worker",
+	"main": "src/index.ts",
+	"compatibility_date": "2025-02-14",
+	"compatibility_flags": ["nodejs_compat"],
```
jamesopstad, line 5:
```suggestion
```

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -1080,6 +1080,45 @@ if (import.meta.hot) {
...
+						});
+
+						// `req.url` is the URL of the request relative to the middleware
+						// mount path. Here we ensure that miniflare receives a request that
+						// reflects the original request url
+						// 🚨🚨🚨 A potential side effect of mutating the original request is
+						// that other middleware functions in the chain will no longer see
+						// the original, relative `req.url`. If a subsequent middleware
+						// depends on the relative path for its logic, this change could
+						// break it
```
```suggestion
```

I don't think this matters in this case because we don't call `next()` in this middleware so other middleware won't receive the request.

packages/vite-plugin-cloudflare/playground/cron-triggers/__tests__/cron-triggers.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { expect, test } from "vitest";
+import { getTextResponse } from "../../__test-utils__";
+
+test("Email triggers", async () => {
+	const rootResponse = await getTextResponse("/");
+	expect(rootResponse).toBe("Hello cron trigger Worker playground!");
+
```
jamesopstad:
```suggestion
```

Don't think this is needed but if it is it would be nicer as a separate test.

packages/vite-plugin-cloudflare/playground/cron-triggers/__tests__/cron-triggers.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { expect, test } from "vitest";
+import { getTextResponse } from "../../__test-utils__";
+
+test("Email triggers", async () => {
```
jamesopstad:
```suggestion
test("supports testing Cron Triggers at /cdn-cgi/handler/scheduled route", async () => {
```

packages/vite-plugin-cloudflare/playground/email-worker/__tests__/email-triggers.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { expect, test, vi } from "vitest";
+import { getTextResponse, viteTestUrl } from "../../__test-utils__";
+
+test("Email triggers", async () => {
```
jamesopstad:
```suggestion
test("Supports sending email via the email binding", async () => {
```

packages/vite-plugin-cloudflare/playground/email-worker/__tests__/email-triggers.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+Hi there`,
+				}
+			);
+			expect(await fetchResponse.text()).toBe(
+				"Worker successfully processed email"
+			);
+		},
+		{ timeout: 2000, interval: 500 }
+	);
```
jamesopstad: This would be nicer in a separate test e.g. 'supports testing Email Workers at /cdn-cgi/handler/scheduled route'.

packages/vite-plugin-cloudflare/playground/email-worker/__tests__/email-triggers.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { expect, test, vi } from "vitest";
+import { getTextResponse, viteTestUrl } from "../../__test-utils__";
+
+test("Email triggers", async () => {
+	const rootResponse = await getTextResponse("/");
+	expect(rootResponse).toBe("Hello Email Workers playground!");
+
```
jamesopstad:
```suggestion
```

packages/vite-plugin-cloudflare/playground/email-worker/__tests__/email-triggers.spec.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/playground/cron-triggers/__tests__/cron-triggers.spec.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/playground/email-worker/__tests__/email-triggers.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+			body: `From: "John" <sender@example.com>
+Reply-To: sender@example.com
+To: recipient@example.com
+Subject: Testing Email Workers Local Dev
+Content-Type: text/html; charset="windows-1252"
+X-Mailer: Curl
+Date: Tue, 27 Aug 2024 08:49:44 -0700
+Message-ID: <6114391943504294873000@ZSH-GHOSTTY>
+
+Hi there`,
```
dario-piotrowicz: Very Nit: I am generally not a fan of this sort of multiline strings breaking indentation, I would consider using dedent, create the string manually via `\n`s and `+`s or something like that instead.

Not a huge deal anyways 🙂

packages/vite-plugin-cloudflare/playground/email-worker/src/index.ts
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	async email(message: ForwardableEmailMessage) {
+		const parser = new PostalMime.default();
+		const rawEmail = new Response(message.raw);
+		const email = await parser.parse(await rawEmail.arrayBuffer());
+		console.log(
+			`Received email from ${email.from.address} on ${email.date} with following message:\n\n ${email.html}`
+		);
+	}
```
dario-piotrowicz, line 49: This logic doesn't seem to be tested in any way, is that intentional?

If I replace this with just
```ts
async email() {}
```
the tests still pass without issues

packages/vite-plugin-cloudflare/playground/email-worker/src/index.ts
dario-piotrowicz, line 1: Locally I'm seeing typescript issues, are you not getting them on your end?

Screenshot 2025-08-22 at 10 38 06

(and is CI not detecting them!? 😮)

packages/vite-plugin-cloudflare/playground/cron-triggers/src/index.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export default class extends WorkerEntrypoint {
+	async fetch(): Promise<Response> {
+		return new Response("Hello cron trigger Worker playground!");
+	}
+
+	async scheduled() {
+		console.log("Cron processed");
+	}
+}
```
dario-piotrowicz: The logic here, as far as I can tell is not being tested, if I update this to be:
```ts
export default class extends WorkerEntrypoint {
	async fetch(): Promise<Response> {
		throw new Error();
	}

	async scheduled() {}
}
```
the tests still pass, is this correct? 🤔

packages/vite-plugin-cloudflare/playground/cron-triggers/src/index.ts
dario-piotrowicz, line 1: I'm seeing typescript issues locally for this file? are you not?

Screenshot 2025-08-22 at 10 45 13

dario-piotrowicz (approved):
Overall looks good to me 😄 

But, unless I'm mistaken there are a few small issues that could ideally be fixed 🤔

packages/vite-plugin-cloudflare/playground/email-worker/src/index.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/playground/cron-triggers/src/index.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/playground/email-worker/src/index.ts
CarmenPopoviciu: oh dear...it's because the script name is `check:types` not `check:type`..notice the extra `s`

CarmenPopoviciu: turns out all playground projects use `check:types` 😿 . I'll address that in a separate PR

CarmenPopoviciu: in the meanwhile, fixed these particular TS issues ✅

packages/vite-plugin-cloudflare/playground/cron-triggers/src/index.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/playground/email-worker/__tests__/email-triggers.spec.ts
CarmenPopoviciu: ✅

packages/vite-plugin-cloudflare/playground/email-worker/__tests__/email-triggers.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+			body: dedent`From: "John" <sender@example.com>
+Reply-To: sender@example.com
+To: recipient@example.com
+Subject: Testing Email Workers Local Dev
+Content-Type: text/html; charset="windows-1252"
+X-Mailer: Curl
+Date: Tue, 27 Aug 2024 08:49:44 -0700
+Message-ID: <6114391943504294873000@ZSH-GHOSTTY>
+
+Hi there`,
```
dario-piotrowicz: you added the `dedent` but kept the same indentation? 🥲 

sorry what I meant is what with `dedent` then you can add extra indentation and that will be ignored 🙂 
(oh.... now hat I think about it... is the indentation actually important here? does miniflare not cope with the formdata if indented? 🤔 )

CarmenPopoviciu: updating as per our async convo ✅

PR #10419, opened 2025-08-20
Update C3 templates to use Durable Object getByName

joshthoward:

Update C3 templates to use Durable Object getByName

- Tests
  - [ ] Tests included
  - [X] Tests not necessary because: I assume we already have tests that C3 templates render
- Public documentation
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24586
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [X] Not necessary because: The only templates work just fine.

### DIFF
--- a/.changeset/curvy-feet-swim.md
+++ b/.changeset/curvy-feet-swim.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": minor
+---
+
+Update create-cloudflare hello-world Durable Objects templates to use `getByName()`
--- a/packages/create-cloudflare/templates/hello-world-durable-object-with-assets/py/src/entry.py
+++ b/packages/create-cloudflare/templates/hello-world-durable-object-with-assets/py/src/entry.py
@@ -50,17 +50,15 @@ class MyDurableObject(DurableObject):
 """
 class Default(WorkerEntrypoint):
     async def fetch(self, request):
-        # Create a `DurableObjectId` for an instance of the `MyDurableObject`
-        # class named "foo". Requests from all Workers to the instance named
-        # "foo" will go to a single globally unique Durable Object instance.
-        id = self.env.MY_DURABLE_OBJECT.idFromName("foo")
+		# Create a stub to open a communication channel with the Durable Object
+		# instance named "foo".
+		#
+		# Requests from all Workers to the Durable Object instance named "foo"
+		# will go to a single remote Durable Object instance.
+        stub = self.env.MY_DURABLE_OBJECT.getByName("foo")
 
-        # Create a stub to open a communication channel with the Durable
-        # Object instance.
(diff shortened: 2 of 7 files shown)


packages/create-cloudflare/templates/hello-world-durable-object-with-assets/py/src/entry.py
```diff
@@ -50,17 +50,15 @@ async def say_hello(self, name):
...
+        stub = self.env.MY_DURABLE_OBJECT.getByName("foo")
 
-        # Create a stub to open a communication channel with the Durable
-        # Object instance.
-        stub = self.env.MY_DURABLE_OBJECT.get(id)
-
-        # Call the `say_hello()` RPC method on the stub to invoke the method on
-        # the remote Durable Object instance
+		# Call the `sayHello()` RPC method on the stub to invoke the method on
+		# the remote Durable Object instance.
```
dario-piotrowicz: shouldn't you also update the method itself? (see on the next line `say_hello` still being used there)

.changeset/curvy-feet-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-cloudflare": minor
+---
+
+Update create-cloudflare Durable Objects templates to use `getByName()`
```
dario-piotrowicz:
```suggestion
Update create-cloudflare hello-world Durable Objects templates to use `getByName()`
```

packages/create-cloudflare/templates/hello-world-durable-object/py/src/entry.py
```diff
@@ -50,17 +50,15 @@ class Default(WorkerEntrypoint):
...
 
-        # Create a stub to open a communication channel with the Durable
-        # Object instance.
-        stub = self.env.MY_DURABLE_OBJECT.get(id)
-
-        # Call the `say_hello()` RPC method on the stub to invoke the method on
-        # the remote Durable Object instance
+		# Call the `sayHello()` RPC method on the stub to invoke the method on
+		# the remote Durable Object instance.
         greeting = await stub.say_hello("world")
```
dario-piotrowicz, line 62: as my other comment, here we still have `say_hello`, so this name should be updated I believe

dario-piotrowicz (commented):
Love the change! 🚀 

I wasn't aware of this new `getByName` API, it looks so much better! 🫶 

Not approving only because I think that some `say_hello` methods need to be renamed to `sayHello`, please have a quick look 🙏

dario-piotrowicz (approved):
Ah right, sorry I forgot that in python camel case is the convention, thanks for reverting the comment 🙏 

Looks good to me! 😄

PR #10422, opened 2025-08-20
Always replace `process.env.NODE_ENV`

jamesopstad:
Fixes #000.

Statically replace the value of `process.env.NODE_ENV` in development when the `nodejs_compat` compatibility flag is enabled.
Previously, this was replaced at build time when `nodejs_compat` was enabled and at dev and build time when `nodejs_compat` was not enabled.

I opted to always set `keepProcessEnv: true` and manually set the replacements as it is easier to reason about.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -124,11 +124,21 @@ export const cloudflareBuiltInModules = [
...
+export function createCloudflareEnvironmentOptions({
+	workerConfig,
+	userConfig,
+	mode,
+	environmentName,
+	isEntryWorker,
+}: {
+	workerConfig: WorkerConfig;
+	userConfig: vite.UserConfig;
+	mode: string;
```
dario-piotrowicz: I'm not sure if it makes things much better, but potentially this could be: `mode: vite.ConfigEnv['mode'];` to indicate more clearly that we're reusing `mode` from vite? 🤔

dario-piotrowicz: The same applies to `getProcessEnvReplacements`

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
```diff
@@ -182,13 +193,43 @@ export function createCloudflareEnvironmentOptions(
...
+		"globalThis.process.env.NODE_ENV": JSON.stringify(nodeEnv),
+	};
+
+	return isNodeCompat(workerConfig)
+		? nodeEnvReplacements
+		: {
+				...nodeEnvReplacements,
+				"process.env": "{}",
+				"global.process.env": "{}",
+				"globalThis.process.env": "{}",
```
dario-piotrowicz, line 229: could you explain the logic here? 🙂 

Why setting the `process.env.NODE_ENV` but then setting all the rest to `{}`, is it really necessary for force `process.env` to be `{}`? (for example I wonder, if some user code / library were to polyfill `process.env` could this break them? 🤔)

jamesopstad: This copies the logic from [here](https://github.com/vitejs/vite/blob/d52cf887fddcfe9e19b02d20de302fae4bbffd1f/packages/vite/src/node/plugins/define.ts#L24-L26). I think the rationale is that it's safer for `process.env` to be an empty object than left as is in a Worker without `nodejs_compat` enabled. This is because if code tries to access a property on it then it will get `undefined` rather than throwing `Cannot read properties of undefined`.

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
jamesopstad: https://github.com/cloudflare/workers-sdk/pull/10422/commits/d6422944217afd95f8976fc690aa4620e162cc6c

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
jamesopstad: Also, this is consistent with the previous behaviour as this is what you get with `keepProcessEnv: false`.

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -788,18 +786,6 @@ if (import.meta.hot) {
...
 				if (isNodeCompat(getWorkerConfig(name))) {
 					return {
-						build: {
-							rollupOptions: {
-								plugins: [
-									replace({
-										"process.env.NODE_ENV": JSON.stringify(
-											process.env.NODE_ENV ?? "production"
-										),
-										preventAssignment: true,
```
edmundhung, line 798: It looks like the replace plugin does a bit more than just setting `define`?

edmundhung (commented):
Is that means we are replacing `process.env.NODE_ENV` at dev and build time regardless of `nodejs_compat` is enabled?

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: Can you elaborate? Are there things that we are now missing that are needed?

jamesopstad:
> Is that means we are replacing `process.env.NODE_ENV` at dev and build time regardless of `nodejs_compat` is enabled?

We were before as well with the difference that it wasn't being replaced at dev time when `nodejs_compat` was enabled.

packages/vite-plugin-cloudflare/src/cloudflare-environment.ts
dario-piotrowicz: ok thanks 🙂 

Yeah, agreed, I do think that this is fine 👍

packages/vite-plugin-cloudflare/src/index.ts
edmundhung: I am just wondering whether `define` has the same concept of `preventAssignment` like the replace plugin did, or we might run into the issue as described in its readme [here](https://github.com/rollup/plugins/blob/master/packages/replace/README.md#preventassignment): 

```
// Input
process.env.DEBUG = false;
if (process.env.DEBUG == true) {
  //
}
// Without `preventAssignment`
false = false; // this throws an error because false cannot be assigned to
if (false == true) {
  //
}
// With `preventAssignment`
process.env.DEBUG = false;
if (false == true) {
  //
}
```

jamesopstad: Hmm, I think assigning to `process.env.NODE_ENV` at runtime would be very unusual. That's the only potential behaviour change related to `preventAssignment` here because `@rollup/plugin-replace` was only being used when `nodejs_compat` was enabled before and we don't replace `process.env` in that scenario.

edmundhung: Yeah, it is definitely uncommon. The debug package did that (which is likely why this example come up), but I guess it should be relatively safe with `NODE_ENV`?

edmundhung (approved):

PR #10424, opened 2025-08-21
Expose global flags from `experimental_getWranglerCommands()`

penalosa:
---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal API
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: experimental

### DIFF
--- a/packages/wrangler/src/experimental-commands-api.ts
+++ b/packages/wrangler/src/experimental-commands-api.ts
@@ -7,7 +7,10 @@ import type { DefinitionTreeNode } from "./core/types";
  *
  * @returns The complete command tree structure with all metadata
  */
-export function experimental_getWranglerCommands(): DefinitionTreeNode {
-	const { registry } = createCLIParser([]);
-	return registry.getDefinitionTreeRoot();
+export function experimental_getWranglerCommands(): {
+	registry: DefinitionTreeNode;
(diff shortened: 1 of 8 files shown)


packages/wrangler/src/experimental-commands-api.ts
```diff
@@ -7,7 +7,10 @@ import type { DefinitionTreeNode } from "./core/types";
  *
  * @returns The complete command tree structure with all metadata
  */
-export function experimental_getWranglerCommands(): DefinitionTreeNode {
```
dario-piotrowicz, line 10: Shouldn't the `returns` tsdoc comment above be updated?

dario-piotrowicz (commented):
Shouldn't this PR include a (patch) changeset? 🤔 

Also completely out of curiosity, but what's the motivation for this change? 🙂

penalosa:
@dario-piotrowicz changesets added, and motivation-wise this is to support cloudflare-docs, who need this when documenting Wrangler

dario-piotrowicz:
Thanks @penalosa 😄 

> motivation-wise this is to support cloudflare-docs, who need this when documenting Wrangler

Oh right... sorry yeah that makes sense 😅 (I forgot that this utility is primarily used for that 👍)

PR #10425, opened 2025-08-21
Fix debugging logs not including headers for CF API requests and responses

### LINKED ISSUE #10421: Wrangler logger with debug sanitization off doesn't correctly print CF HEADERS

Wrangler 4.31.0

_No response_

### Describe the Bug

The wrangler debugging logging with sanitization turned off, implemented in:
https://github.com/cloudflare/workers-sdk/blob/2fd888ffec31e1656a39562308ccc75774b8b0e0/packages/wrangler/src/cfetch/internal.ts#L45
(and in a few other places)

Is flawed and always simply prints headers as `{}`

(For context having sanitization on results in: `HEADERS: omitted; set WRANGLER_LOG_SANITIZE=false to include sanitized data`)

For example if I run ` WRANGLER_LOG=DEBUG WRANGLER_LOG_SANITIZE=false npx wrangler deploy` in the results all the headers look to like they are empty (even if they aren't):
Image

The problem being that wrangler simply prints using `JSON.stringify(logHeaders, 2, null)` while that is not enough, for example see:
Image

_No response_

dario-piotrowicz:
Fixes #10421

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/real-rabbits-enjoy.md
+++ b/.changeset/real-rabbits-enjoy.md
@@ -1,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+Fix debugging logs not including headers for CF API requests and responses
+
+Fix the fact that `wrangle` when run with the `WRANGLER_LOG=DEBUG` and `WRANGLER_LOG_SANITIZE=false` environment variables displays `{}` instead of the actual headers for requests and responses for CF API fetches
--- a/packages/wrangler/src/__tests__/helpers/assert-request.ts
+++ b/packages/wrangler/src/__tests__/helpers/assert-request.ts
@@ -29,7 +29,7 @@ export function makeApiRequestAsserter(
 				/HEADERS: (?<headers>(.|\n)*?)\nINIT: (?<init>(.|\n)*?)\n(BODY: (?<bodyMatch>(.|\n)*?)\n)?-- END CF API REQUEST/
 			);
 		const {
-			headers: _headers,
+			headers: headersStr,
 			init: _init,
 			bodyMatch,
 		} = requestDetails?.groups ?? {};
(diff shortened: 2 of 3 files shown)


dario-piotrowicz:
I think the debugging logs could use some more tests (for example I don't think we have any tests for debug logs for api responses), but just not to go out of scope here I just added a simple check to make sure that headers are indeed included in the non sanitized debug logs

dario-piotrowicz:
> [!Note]
> I've added the `skip-v3-pr` label just to silence the backport workflow

.changeset/real-rabbits-enjoy.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": patch
+---
+
+Fix debugging logs not including headers for CF API requests and responses
+
+Fix the fact that `wrangle` when run with the `WRANGLER_LOG=DEBUG` and `WRANGLER_LOG_SANITIZE=false` environment variables displays `{}` instead of the actual headers for requests and responses for CF API fetches
```
CarmenPopoviciu:
```suggestion
Fix the fact that `wrangler`, when run with the `WRANGLER_LOG=DEBUG` and `WRANGLER_LOG_SANITIZE=false` environment variables, displays `{}` instead of the actual headers for requests and responses for CF API fetches
```

CarmenPopoviciu (approved):

.changeset/real-rabbits-enjoy.md
dario-piotrowicz: 🤦 👍

PR #10427, opened 2025-08-21
Simplify ENOENT debug logs for `.env` files

dario-piotrowicz:
Fixes https://github.com/cloudflare/workers-sdk/issues/7042

This PR reintroduces the changes made in https://github.com/cloudflare/workers-sdk/pull/7272 that were incorrectly removed in https://github.com/cloudflare/workers-sdk/pull/9914

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: simple UX improvement
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: The `.env` change that caused this regression is not in v3

### DIFF
--- a/packages/wrangler/src/config/dot-env.ts
+++ b/packages/wrangler/src/config/dot-env.ts
@@ -49,7 +49,7 @@ export function loadDotEnv(
 			override: true,
 		});
 		if (error) {
-			if ('code' in error && error.code === "ENOENT") {
+			if ("code" in error && error.code === "ENOENT") {
 				logger.debug(
 					`.env file not found at "${envPath}". Continuing... For more details, refer to https://developers.cloudflare.com/workers/wrangler/system-environment-variables/`
 				);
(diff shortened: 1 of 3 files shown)


packages/wrangler/src/config/dot-env.ts
```diff
@@ -49,7 +49,13 @@ export function loadDotEnv(
 			override: true,
 		});
 		if (error) {
-			logger.debug(`Failed to load .env file "${envPath}":`, error);
+			if ((error as unknown as { code: string }).code === "ENOENT") {
```
anonrig: Nit

```suggestion
			if ('code' in error && error.code === "ENOENT") {
```

dario-piotrowicz: good call! 😄 👍

PR #10437, opened 2025-08-22
Loosen validation around different configurations for Durable Object

dario-piotrowicz:
The changes in this PR allow durable objects to have `enableSql`, `unsafeUniqueKey` and `unsafePreventEviction` configurations set to `undefined` even if they same durable objects are defined with those configurations set to a different value (this would allow workers using external durable objects not to have to duplicate such configurations in their options)

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/10465
  - [ ] Not necessary because:

### DIFF
--- a/.changeset/loose-glasses-poke.md
+++ b/.changeset/loose-glasses-poke.md
@@ -4,4 +4,4 @@
 
 Loosen validation around different configurations for Durable Object
 
-Allow durable objects to have `enableSql`, `unsafeUniqueKey` and `unsafePreventEviction` configurations set to `undefined` even if they same durable objects are defined with those configurations set to a different value (this would allow workers using external durable objects not to have to duplicate such configurations in their options)
+Allow durable objects to have `enableSql`, `unsafeUniqueKey` and `unsafePreventEviction` configurations set to `undefined` even if the same durable objects are defined with those configurations set to different values (this allows workers using external durable objects not to have to duplicate such configurations in their options)


packages/miniflare/src/index.ts
```diff
@@ -355,13 +355,27 @@ function getDurableObjectClassNames(
...
+					if (!potentiallyValidSqlDiff) {
+						// We only want to throw this error is we know that the DO configuration cannot
+						// be potentially valid
+						throw new MiniflareCoreError(
+							"ERR_DIFFERENT_STORAGE_BACKEND",
+							`Different storage backends defined for Durable Object "${className}" in "${serviceName}": ${JSON.stringify(
+								enableSql
+							)} and ${JSON.stringify(existingInfo?.enableSql)}`
+						);
+					}
```
edmundhung: You might need to update the map with regards to the `enableSql` if `existingInfo.enableSql` is `undefined` but `enableSql` is `true`. I think we will use the map for some follow up operations (?)

Should we apply the same check for `unsafeUniqueKey` and `unsafePreventEviction`?

dario-piotrowicz: > You might need to update the map with regards to the enableSql if existingInfo.enableSql is undefined but enableSql is true. I think we will use the map for some follow up operations (?)

I don't think that we use `existingInfo` for sql logic besides this? 🤔 Also via manual testing my changes seemed enough in this regard, so unless something would break I think this should be ok? do you see/think of anything that could specifically break by not updating the map? 🤔

dario-piotrowicz: > Should we apply the same check for unsafeUniqueKey and unsafePreventEviction?

mh.... good question.... yeah I might just as well! 🤔👍

dario-piotrowicz: @edmundhung I've addressed the other configs checking, but not updating the map, please have another look and let me know if this works for you 🙇

packages/miniflare/test/plugins/do/index.spec.ts
```diff
@@ -482,3 +482,112 @@ test("colo-local actors", async (t) => {
 	res = await stub.fetch("http://localhost");
 	t.is(await res.text(), "body:thing2");
 });
+
+test("multiple workers with DO conflicting useSQLite booleans cause options error", async (t) => {
```
dario-piotrowicz, line 486: I've added tests only for `useSQLite` since adding tests for all 3 fields seems overkill and unncessary to me, if someone disagree please just let me know and I'll be happy to add some for the other 2 fields as well 🙂

packages/miniflare/src/index.ts
```diff
@@ -340,51 +340,83 @@ function getDurableObjectClassNames(
...
+					// want to consider this as an acceptable difference since we might be in a potentially valid situation in
+					// which worker A defines a DO with a config, while worker B simply uses the DO from worker A but without
+					// providing the configuration (thus leaving it `undefined`) (this for example is exactly what Wrangler does
+					// with the implicitly defined `enableSql` flag)
+					if (oneIsUndefined) {
+						return false;
+					}
+
+					return true;
+				};
```
edmundhung: Nit: I think one minor issue with this check is that if the `existingInfo[field]` is undefined, even when later there are DOs with `doConfigs[field]` being `true` and some being `false`, this is still being accepted.

Ideally, it would be great to only accept `true | undefined` or `false | undefined` among a set of DO configs.

edmundhung: In reality this shouldn't really happen as the config will be `true` or `false` only from the worker that define the DO, the rest of the workers would only be `undefined`. So this is just a nitpick :)

dario-piotrowicz: I understand what you are suggesting, and I can make the change (you are suggesting to collect all the values and check them together right?), but I am not sure if that is necessary, because this check applies to all DO configuration so let's say that as some point we have `existingInfo[field] === undefined` and `doConfigs[field] === true`, this will pass, if later all the `doConfigs[field]` values are `true` then everything will work fine, if one will be `false` we will detect that then (and vice versa).

So either way if there is a conflict between two DOs we are going to detect it sooner or later and the order of DOs doesn't even matter, no?

dario-piotrowicz: As discussed async you were right, sorry I misremembered how the `existingInfo` are being set 🙇 

d81c85e72659b6348784f844d531ee6d9f5cc0f6 should address your concerns 🙂

packages/miniflare/src/index.ts
```diff
@@ -331,63 +331,114 @@ function getDurableObjectClassNames(
...
-					);
+			);
+		})
+		// We sort the list of durable objects because we want the durable objects without a scriptName or a scriptName
+		// that matches the raw worker's name (meaning that they are defined within their worker) to be processed first
+		.sort(({ doInfo, workerRawName }) =>
+			doInfo.scriptName === undefined || doInfo.scriptName === workerRawName
+				? -1
+				: 0
+		)
```
edmundhung, line 359: I think the default scriptName logic might have broken the sorting?

edmundhung (approved):

.changeset/loose-glasses-poke.md
```diff
@@ -0,0 +1,7 @@
+---
+"miniflare": patch
+---
+
+Loosen validation around different configurations for Durable Object
+
+Allow durable objects to have `enableSql`, `unsafeUniqueKey` and `unsafePreventEviction` configurations set to `undefined` even if they same durable objects are defined with those configurations set to a different value (this would allow workers using external durable objects not to have to duplicate such configurations in their options)
```
dario-piotrowicz:
```suggestion
Allow durable objects to have `enableSql`, `unsafeUniqueKey` and `unsafePreventEviction` configurations set to `undefined` even if the same durable objects are defined with those configurations set to different values (this allows workers using external durable objects not to have to duplicate such configurations in their options)
```

PR #10442, opened 2025-08-22
handle hybrid polyfill on node:https

anonrig:
Handles the hybrid polyfills on node:https

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Will do in bulk
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Unenv doesn't exist on v3

### DIFF
--- a/packages/unenv-preset/src/preset.ts
+++ b/packages/unenv-preset/src/preset.ts
@@ -184,7 +184,7 @@ function getHttpOverrides({
 			"_http_outgoing",
 			...(httpServerEnabled ? ["_http_server", "https"] : []),
 		],
-		hybridModules: httpServerEnabled ? ["http"] : ["http", "https"],
+		hybridModules: ["http", "https"],
 	};
 }
 
--- a/packages/unenv-preset/src/runtime/node/https.ts
+++ b/packages/unenv-preset/src/runtime/node/https.ts
@@ -1,13 +1,26 @@
 // This hybrid polyfill is used only when the native implementation is not available
 // See `src/preset.ts` for details
 
-import { createServer, Server } from "unenv/node/https";
+import {
+	createServer as unenvCreateServer,
+	Server as unenvServer,
+} from "unenv/node/https";
 import type nodeHttps from "node:https";
-
-export { Server, createServer } from "unenv/node/https";
 
 const workerdHttps = process.getBuiltinModule("node:https");
 
+// Use the workerd implementation of server APIs when the
+// `enable_nodejs_http_server_modules` flag is on.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const isWorkerdServerEnabled = (globalThis as any).Cloudflare.compatibilityFlags
+	.enable_nodejs_http_server_modules;
+
+export const createServer = isWorkerdServerEnabled
+	? workerdHttps.createServer
+	: unenvCreateServer;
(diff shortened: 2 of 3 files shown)


.changeset/fruity-horses-divide.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cloudflare/unenv-preset": patch
+"wrangler": patch
+---
+
+Fix hybrid polyfill on node:https
```
vicb, line 6: Can you expand on what this PR is fixing?

packages/unenv-preset/src/runtime/node/https.ts
```diff
@@ -1,13 +1,26 @@
 // This hybrid polyfill is used only when the native implementation is not available
```
vicb, line 1: The comment does not look applicable any more?

packages/unenv-preset/src/preset.ts
```diff
@@ -184,7 +184,7 @@ function getHttpOverrides({
 			"_http_outgoing",
 			...(httpServerEnabled ? ["_http_server", "https"] : []),
 		],
-		hybridModules: httpServerEnabled ? ["http"] : ["http", "https"],
+		hybridModules: ["http", "https"],
```
vicb, line 187: Why this chnage?

It is better to avoid hybrid module (in favor of native) if we can. Not sure why the hybrid polyfill is always used after this PR?

.changeset/fruity-horses-divide.md
anonrig: Https is enabled when you have enable_nodejs_http_modules, and https.Server is enabled when you have enable_nodejs_http_modules and enable_nodejs_http_server_modules. Just like node:http. So I think these changes are correct. It should behave the same way as node:http

vicb: We always need an hybrid polyfill for `http` because some APIs are not implemented by workerd even when `enable_nodejs_http_modules` is used:

https://github.com/cloudflare/workers-sdk/blob/6ff6b3e1dd5d499ba9f0f84b1ca156280e12069d/packages/unenv-preset/src/runtime/node/http.ts#L14-L21

For `https`:

- if `enable_nodejs_http_modules` is used without  `enable_nodejs_http_server_modules` then we get the server APIs from unenv and the rest from workerd.
- when both flags are used, all of the APIs are implemented by workerd and we can then use the native module. There is no need for an hybrid polyfill in this case.

Your code should work but it will always use an hybrid polyfill and you can see that nothing is ever used from unenv if both flags are set. In such a case it is better not to polyfill and use the native module directly which is what the current code does.

anonrig: These are all implemented at the runtime: https://github.com/cloudflare/workerd/blob/main/src/node/http.ts#L36

vicb: > These are all implemented at the runtime: https://github.com/cloudflare/workerd/blob/main/src/node/http.ts#L36

Then we should be able to drop the http polyfill and make sure to bump the workerd peer dep to when https://github.com/cloudflare/workerd/pull/4692 was released if it is anytime before.

Feel free to close the current PR as I don't think it's needed.

PR #10462, opened 2025-08-25
Update mime dependency to 4.0

WillTaylorDev:
Javascript files should have content type as `text/javascript`.
https://github.com/jshttp/mime-db/issues/194#issuecomment-2226429582

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: n/a


WalshyDev (approved):

packages/workers-shared/utils/helpers.ts
```diff
@@ -1,7 +1,7 @@
 import { readFileSync } from "node:fs";
 import { isAbsolute, resolve, sep } from "node:path";
 import ignore from "ignore";
-import { getType } from "mime";
```
matthewdavidrodgers, line 4: any reason for the change in imports? looks identical to me

matthewdavidrodgers (approved):

packages/workers-shared/utils/helpers.ts
WalshyDev: getType isn't exported as it's own func now, they export a class of mime which has the getType func
https://github.com/broofa/mime/blob/main/src/index.ts

PR #10478, opened 2025-08-27
Wrangler preview urls default to disabled

danielrs:
_Describe your change..._

This change makes preview_urls disabled by default when it's not provided, making
the feature opt-in instead of opt-out.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24756
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: No big changes on the feature, just the default value.

Internal tracking: https://jira.cfdata.org/browse/RM-24647

### DIFF
--- a/packages/wrangler/src/triggers/deploy.ts
+++ b/packages/wrangler/src/triggers/deploy.ts
@@ -336,6 +336,28 @@ async function subdomainDeploy(
 	const previewsInSync = wantPreviews === currPreviews;
 	const allInSync = [workersDevInSync, previewsInSync].every((v) => v);
 
+	// Warn about mismatching config and current values.
+
+	if (config.workers_dev == undefined && !workersDevInSync) {
+		const currWorkersDevStatus = currWorkersDev ? "enabled" : "disabled";
+		logger.warn(
+			[
+				`Worker has workers.dev ${currWorkersDevStatus}, but 'workers_dev' is not in the config.`,
+				`Using fallback value 'workers_dev = ${wantWorkersDev}'.`,
+			].join("\n")
+		);
+	}
+
+	if (config.preview_urls == undefined && !previewsInSync) {
+		const currPreviewsStatus = currPreviews ? "enabled" : "disabled";
+		logger.warn(
+			[
+				`Worker has preview URLs ${currPreviewsStatus}, but 'preview_urls' is not in the config.`,
+				`Using fallback value 'preview_urls = ${wantPreviews}'.`,
+			].join("\n")
+		);
+	}
+
 	// workers.dev URL is only set if we want to deploy to workers.dev.
 
 	let workersDevURL: string | undefined;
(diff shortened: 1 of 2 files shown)


packages/wrangler/src/triggers/deploy.ts
```diff
@@ -340,3 +291,73 @@ export default async function triggersDeploy(
 		logger.log("No deploy targets for", workerName, formatTime(deployMs));
 	}
 }
+
+async function subdomainDeploy(
```
danielrs, line 311: Function needs some work to make all tests pass, but should be clearer what's going on with subdomains.

vicb:
> This change makes preview_urls disabled by default when it's not provided, making
> the feature opt-in instead of opt-out.

Hey @danielrs thanks for the PR.

Could you expand on the rationale of this change?
Has this been discussed with the ANT team before?
_Asking because this could be seen as a breaking change._

danielrs:
@vicb We think that this feature is better off disabled by default, users can explicitly opt-in depending on their use-case.

packages/wrangler/src/__tests__/helpers/mock-upload-worker.ts
```diff
@@ -236,9 +241,27 @@ export function mockUploadWorkerRequest(
...
+		}
+	}
+	let routes: Route[] = config.routes ?? [];
+	if (env) {
+		routes = config.env?.[env]?.routes ?? [];
+	}
+	const { workers_dev: defaultWorkersDev, preview_urls: defaultPreviewUrls } =
+		getSubdomainDefaults(routes);
 	mockGetWorkerSubdomain({
-		enabled: true,
```
danielrs, line 241: Always setting this value to enabled is actually wrong, the correct value depends on whenever we have routes or not. The function `getSubdomainDefaults` should help with that.

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -94,55 +83,17 @@ export default async function triggersDeploy(
...
+		props,
+		accountId,
+		scriptName,
+		envName,
+		workerUrl,
+		routes,
+		deployments
+	);
+
+	if (!wantWorkersDev && workersDevInSync && routes.length !== 0) {
```
danielrs, line 96: There was a bug in our testing suite that was always setting `workers_dev = true`, causing this block to never execute. Now that it does execute, tests mocks were updated accordingly. See my related comment here: https://github.com/cloudflare/workers-sdk/pull/10478/files#r2309034991.

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -349,3 +291,111 @@ export default async function triggersDeploy(
...
+	if (!allInSync) {
+		await fetchResult(config, `${workerUrl}/subdomain`, {
+			method: "POST",
+			body: JSON.stringify({
+				enabled: wantWorkersDev,
+				previews_enabled: wantPreviews,
+			}),
+			headers: {
+				"Content-Type": "application/json",
+				"Cloudflare-Workers-Subdomain-Compatibility-Date": "2025-08-01",
```
danielrs: This will be used in the API for handling older Wrangler versions.

aaronlisman:
we went over these changes together and they look good in conjunction with the API changes ✅

MattIPv4:
👋 Just filed #10722 for this, but with this change, wrangler is now forcibly disabling preview URLs even if you enable them in the dashboard, if you call `wrangler deploy` without a config...

PR #10479, opened 2025-08-27
[R2 Data Catalog] Add wrangler commands for the R2 Data Catalog compaction

nagraham:
Closes internal jira IGLOO-102.

This change adds two new Wrangler sub-commands within R2 Data Catalog to enable/disable automatic compaction for an Iceberg Table in their catalog. Compaction optimizes the size of data files to improve read performance.

The feature is still in private-beta.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: This is forthcoming
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: new feature.

### DIFF
--- a/packages/wrangler/src/r2/catalog.ts
+++ b/packages/wrangler/src/r2/catalog.ts
@@ -5,7 +5,13 @@ import { logger } from "../logger";
 import { APIError } from "../parse";
 import { requireAuth } from "../user";
 import formatLabelledValues from "../utils/render-labelled-values";
-import { disableR2Catalog, enableR2Catalog, getR2Catalog } from "./helpers";
+import {
+	disableR2Catalog,
+	disableR2CatalogCompaction,
+	enableR2Catalog,
+	enableR2CatalogCompaction,
+	getR2Catalog,
+} from "./helpers";
 
 export const r2BucketCatalogNamespace = createNamespace({
 	metadata: {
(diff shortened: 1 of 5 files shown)


packages/wrangler/src/r2/catalog.ts
```diff
@@ -152,3 +158,104 @@ export const r2BucketCatalogGetCommand = createCommand({
...
+		namespace: {
+			describe: "The namespace containing the table",
+			type: "string",
+			demandOption: true,
+		},
+	},
+	async handler(args, { config }) {
+		const accountId = await requireAuth(config);
+
+		await enableR2CatalogCompaction(
```
penalosa, line 199: If this API is documented, could we use the SDK here instead? It's available as `sdk` on the second argument of the handler function

nagraham: I don't think it's yet documented. I don't see the commands available on the SDK.

penalosa: Is there any blocker for this being documented? It would be nice to have terraform/API support at launch, and that would enable using the SDK in Wrangler, as well

nagraham: It's not so much as blocker as it is competing priorities and limited time. We would like to get the API doc changes in, but it won't happen before our internal deadlines for wrangler changes. We need to ship a working version of these commands before that deadline. Using the sdk isn't a strict requirement and yet it requires additional time/effort. So it adds risk. I'd like to ship the commands first to get this in the hands of users. At a later point in time, we'll ship the docs and then switch over to the sdk.

PR #10482, opened 2025-08-28
Pass the `compatibility_date` and `compatibility_flags` to the `unenv` preset

### LINKED ISSUE #10079: [Feature] The vite plugin should pass the compatibility date and flags to the preset
Some of the newly implemented native modules (http & process for now) are behind compatibility date and flags.

The unenv preset was updated to return a different set of polyfills according to the date and time, see https://github.com/cloudflare/workers-sdk/pull/10048

https://github.com/cloudflare/workers-sdk/pull/10078 takes advantage of the feature for the http module. 

However this only works with wrangler for now.

That's because the Vite plugin only creates the preset once at top level:

https://github.com/cloudflare/workers-sdk/blob/2df1d066cfe376b831ff0b29b656437d869791e5/packages/vite-plugin-cloudflare/src/node-js-compat.ts#L12-L21

That's problematic for a couple reason:
- the date and flags are not known at top level (i.e. before a "build")
- the date and flags can change when the wrangler configuration is update.

For the Vite plugin to take advantage of the "dynamic" preset, it will need to (re)create the preset when either the date or flag changes.

I'm not familiar with the architecture of the Vite plugin and I would appreciate someone with more context to take care of this update. The first thing to figure out would be where to create to preset knowing the lifecycle of the plugin).

@jamesopstad  please add any more info that would be needed by whoever will implement - you probably have an idea of who can help with that.

Note: The Vite plugin works even if it does not use the "dynamic" preset - it will only always use the `unenv` polyfill instead of the native implemenations where possible (for now, only the `http` modules are impacted)

jamesopstad:
Fixes #10079.
Fixes [ #3820](https://github.com/cloudflare/workerd/issues/3820).

Pass the `compatibility_date` and `compatibility_flags` to the `unenv` preset. This enables support for the `node:http` and `node:https` modules.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -1107,4 +1128,10 @@ if (import.meta.hot) {
 			? resolvedPluginConfig.workers[environmentName]
 			: undefined;
 	}
+
+	function getNodeJsCompat(environmentName: string) {
```
vicb, line 1135: nit: add JSDoc to this function?

packages/vite-plugin-cloudflare/src/index.ts
```diff
@@ -163,6 +157,11 @@ export function cloudflare(pluginConfig: PluginConfig = {}): vite.Plugin[] {
 															resolvedPluginConfig.type === "workers" &&
 															environmentName ===
 																resolvedPluginConfig.entryWorkerEnvironmentName,
+														hasNodeJsCompat:
+															resolvedPluginConfig.type === "workers" &&
+															resolvedPluginConfig.nodeJsCompatMap.has(
+																environmentName
+															),
```
vicb: nit: re-use the helper function

```suggestion
														hasNodeJsCompat: getNodeJsCompat(environmentName) !== undefined
```

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: [67cfeba](https://github.com/cloudflare/workers-sdk/pull/10482/commits/67cfebaffa2fa2424b6e2285606df906184f9b93)

packages/vite-plugin-cloudflare/src/index.ts
jamesopstad: [e43e7db](https://github.com/cloudflare/workers-sdk/pull/10482/commits/e43e7db43de161e6d1424ec6fe9842a4503adf4f)

PR #10485, opened 2025-08-28
feat(vite-plugin-cloudflare): enable cross-process named entrypoint binding support

### LINKED ISSUE #9758: vite plugin with named entrypoints does not work
When trying to run RPC locally with the React + Vite framework (https://developers.cloudflare.com/workers/framework-guides/web-apps/react/)  with a Service Binding with named entrypoints, you get the error below. The RPC works fine once you publish and deploy the worker, but on local dev using vite , it throws the error below. This requires us to comment out the RPC code in local dev or else we can't run the application. We are familiar with RPC and can run it locally with regular workers, so we have done this before. This error only occurs when trying to use vite with the cloudflare plugin locally. So I think this is a bug in the cloudflare plugin for Vite because it doesn't work with RPC.

error when starting dev server:
AssertionError [ERR_ASSERTION]: <service> does not match a worker name.
    at getWorkerToWorkerEntrypointNamesMap (node_modules/@cloudflare/vite-plugin/dist/index.js:13386:9)
    at getDevMiniflareOptions (node_modules/@cloudflare/vite-plugin/dist/index.js:13624:44)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async configureServer (node_modules/@cloudflare/vite-plugin/dist/index.js:14522:37)
    at async _createServer (node_modules/vite/dist/node/chunks/dep-DBxKXgDP.js:38657:20)
    at async CAC.<anonymous> (fnode_modules/vite/dist/node/cli.js:753:20)

BTW our binding in the wrangler.jsonc looks like this (names changed). It works perfectly fine in production, just doesn't work in local development. The service bindings just don't get picked up. Our set up is exactly the same as what you get when you run the CLI from the link above for React + vite framework
```
"services": [
	{
			"binding": "OUR_SERVICE",
			"service": "our_worker_service_name",
			"entrypoint": "ENTRY_CLASS"
		}
	],
```

edmundhung:
Fixes #9758.

This PR introduces a new mechanism that allows miniflare to register the worker with its dependency information, which will be used by the vite plugin to discover whether a named entrypoint should be exposed over the dev registry.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: This limitation was not documented. So no need to update the docs.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Miniflare Dev Registry is a v4 only feature

### DIFF
--- a/packages/miniflare/src/shared/dev-registry.ts
+++ b/packages/miniflare/src/shared/dev-registry.ts
@@ -84,7 +84,9 @@ export class DevRegistry {
 		this.externalServicesByWorker = new Map(servicesByWorker);
 
 		if (!this.watcher) {
-			this.watcher = watch(this.registryPath).on("all", () => this.refresh());
+			this.watcher = watch(this.registryPath, { ignoreInitial: true })
+				.on("ready", () => this.refresh())
+				.on("all", () => this.refresh());
 			this.refresh();
 		}
 	}


danieljvdm:
Any updates on this? It'd be really nice to use named entrypoints with the vite plugin....

edmundhung:
Hi @danieljvdm, sorry this takes longer than we hoped. It should be supported now with `@cloudflare/vite-plugin` v1.16.0+

PR #10489, opened 2025-08-28
Up max asset count to 100k

WalshyDev:
Fixes N/A

Allows up to 100k static assets for Workers Assets. Further work needed in the future to pull the limit from the API and use that instead of having a hardcoded limit.
We could remove the limit totally in Wrangler but having it for dev is nice so this is the best current path currently.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24842
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: assets wasn't GA till v4

### DIFF
--- a/.changeset/clear-views-scream.md
+++ b/.changeset/clear-views-scream.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Allow Wrangler to upload 100,000 assets inline with the newly increased Workers Paid limit.
--- a/packages/workers-editor-shared/package.json
+++ b/packages/workers-editor-shared/package.json
@@ -42,6 +42,7 @@
 		"eslint": "^8.57.1",
 		"react": "^18.3.1",
 		"react-dom": "^18.3.1",
+		"typescript": "catalog:default",
 		"vite": "catalog:default",
 		"vite-plugin-dts": "^4.0.1"
 	},
--- a/packages/workers-shared/utils/constants.ts
+++ b/packages/workers-shared/utils/constants.ts
@@ -23,9 +23,12 @@ export const ENTRY_SIZE = PATH_HASH_SIZE + CONTENT_HASH_SIZE + TAIL_SIZE;
 
 // -- Manifest creation constants --
 // used in wrangler dev and deploy
-/** Maximum number of assets that can be deployed with a worker */
-export const MAX_ASSET_COUNT = 20_000;
-/** Maximum size per asset that can be deployed with a worker */
+/**
+ * Maximum number of assets that can be deployed with a Worker; this is a global
+ * ceiling, and may vary by the user's subscription.
+ */
+export const MAX_ASSET_COUNT = 100_000;
+/** Maximum size per asset that can be deployed with a Worker */
 export const MAX_ASSET_SIZE = 25 * 1024 * 1024;
 
 export const CF_ASSETS_IGNORE_FILENAME = ".assetsignore";
--- a/packages/wrangler/src/assets.ts
+++ b/packages/wrangler/src/assets.ts
@@ -6,7 +6,6 @@ import { parseStaticRouting } from "@cloudflare/workers-shared/utils/configurati
 import {
 	CF_ASSETS_IGNORE_FILENAME,
 	HEADERS_FILENAME,
-	MAX_ASSET_COUNT,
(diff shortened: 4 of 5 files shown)


packages/workers-shared/utils/constants.ts
```diff
@@ -24,7 +24,7 @@ export const ENTRY_SIZE = PATH_HASH_SIZE + CONTENT_HASH_SIZE + TAIL_SIZE;
 // -- Manifest creation constants --
 // used in wrangler dev and deploy
 /** Maximum number of assets that can be deployed with a worker */
```
WillTaylorDev:
```suggestion
/** Maximum number of assets that can be deployed with a worker; this is a global
* ceiling, and may vary by the user's subscription.
*/
```

packages/workers-editor-shared/package.json
```diff
@@ -42,6 +42,7 @@
 		"eslint": "^8.57.1",
 		"react": "^18.3.1",
 		"react-dom": "^18.3.1",
+		"typescript": "catalog:default",
```
WalshyDev, line 45: fixes typescript errors for me when building - previously it was trying to use a non-installed ts ver

packages/wrangler/src/assets.ts
```diff
@@ -287,15 +285,6 @@ const buildAssetManifest = async (dir: string) => {
 					assetsIgnoreFilePresent
 				);
 
-				if (counter >= MAX_ASSET_COUNT) {
```
WalshyDev, line 290: We wan the API to handle this

packages/wrangler/src/assets.ts
```diff
@@ -340,7 +328,7 @@ function logAssetUpload(line: string, diffCount: number) {
 			"   (truncating changed assets log, set `WRANGLER_LOG=debug` environment variable to see full diff)";
 		logger.info(chalk.dim(msg));
 	}
-	return diffCount++;
+	return ++diffCount;
```
WalshyDev, line 331: This was never incrementing and thus would print out 100k lines, fixed the limiting

WillTaylorDev:
Looks good to me, thanks Walshy!!

packages/workers-shared/utils/constants.ts
```diff
@@ -23,8 +23,11 @@ export const ENTRY_SIZE = PATH_HASH_SIZE + CONTENT_HASH_SIZE + TAIL_SIZE;
 
 // -- Manifest creation constants --
 // used in wrangler dev and deploy
-/** Maximum number of assets that can be deployed with a worker */
-export const MAX_ASSET_COUNT = 20_000;
+/**
+ * Maximum number of assets that can be deployed with a worker; this is a global
```
CarmenPopoviciu:
```suggestion
 * Maximum number of assets that can be deployed with a Worker; this is a global
```

packages/wrangler/src/assets.ts
CarmenPopoviciu: why not fail early tho?

.changeset/clear-views-scream.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
dario-piotrowicz:
```suggestion
"wrangler": patch
```

packages/wrangler/src/assets.ts
WalshyDev: what do you mean? the api handles the limit, this will fail on that api call.

CarmenPopoviciu: right, but we could already fail in the cli, without making the API call, like it was until now. Totally not a blocker, was just trying to understand why we'd rather make an API call that we already know will fail because of that limit

WalshyDev: Because we don't know what the limit is and will be. For some users it'll be 20k, for some 100k, some in the future may have 200k.

Wrangler should really _never_ hardcode limits, they can always change. The API is the source of truth.

I suspect we'll have a /limits endpoint in the future which we can pull from but for now, we'll do this.

CarmenPopoviciu: cool cool! got it! thx Walshy 🙏

PR #10494, opened 2025-08-28
[vitest-pool-workers] Workflows test support

pombosilva:
Fixes WOR-355 (related to WOR-854,  WOR-855, WOR-856,  WOR-857, WOR-858, WOR-859, WOR-860, WOR-861 and WOR-862).

### New testing features in vitest-pool-workers:
This provides two APIs for testing Workflows with vitest-pool-workers. These are the main entry points into changing Workflow instances behavior during tests:
- **introspectWorkflowInstance**(env.WORKFLOW, "id123"): when the instance id is known beforehand, useful for Workflow unit tests
    - returns `WorkflowInstanceIntrospector` that allows to call:
        - `modify(fn)`: function to where modifications are applied (calls rpc functions to workflows-shared, where changes are made to DO storage)
        - `waitFoRStepResult(step, result)`: waits until a step resolves into a result or rejects into an error
        - `waitForStatus(status)`: waits until the provided workflow instance reaches a given state
        - `dispose()`: disposes the workflow instance, so state does not get persisted across tests using the same workflow instance (it is actually deleting the DO storage and then aborting it)

- **introspectWorkflow**(env.WORKFLOW): when the instance id is not known beforehand, useful for integration tests
    - proxies the Workflow to intercept .create() and .createBatch() calls 
    - returns `WorkflowIntrospector`  that allows to call:
        - `modifyAll(fn)`: same as `modify(fn)` but applies modifications to all workflow instances created after the `introspectWorkflow` call until the introspector `cleanUp` call
        - `get()`: returns a list of `WorkflowInstanceIntrospector` of all created workflow instances after the `introspectWorkflow` call
        - `dispose()`: disposes the introspector, meaning that no more Workflow calls will be intercepted, and also disposes all workflow instance introspector created

Both APIs support disposal of their returned objects by using the `using` keyword. This will automatically call cleanUp() at the end of the code block (at the end of the test).

### Miniflare:
- ~~Added an `unsafeScriptName` to DurableObjectsOptionsSchema of DO plugin, to allow defining the DO serviceName without the 'core:user:' prefix.~~

### workflows-shared:
- [new] Added unsafe functions to Workflow binding to be able to access Engine DOs from vitest-pool-workers
- Added an RpcTarget class with the modifier functions that update the DO storage
- Added logic into the Engine/Context to verify the DO storage to see if behavior needs to be changed accordingly

### vitest-pool-workers:
- ~~For each Workflow binding present in a wrangler file, added a DO namespace binding to Runner Worker with unsafeScriptName (these are the bindings to the Engines DOs)~~
- Changed the bundle.mjs file (with @edmundhung help) to make it possible importing from `worklfows-shared` (to avoid type duplication) by using explicit external list to customize what get bundled
- Added workflows file to `src/worker` with the new APIs
- Exposed the new APIs and necessary interfaces with some documentation

### Fixtures:
- Updated the tests with the new APIs in vitest-pool-workers-examples workflows 

### Changes to files after running tests:
If I successfully run some tests with these new APIs, then make a change to the test file (can as little as just adding a space) and then save it (ctrl+s), the tests will automatically run again (if the vitest interactive process on the terminal does not get killed) and they will fail.
I suspect this is due to the instance cleanUp() function that is effectively aborting the instances DOs, and whenever the tests try to run again in the same vitest process lifetime, it does not construct the DOs again.

### Workerd errors appear on terminal:
I've noticed that if I try to run a test that makes an error be thrown, workerd prints it into the console. Let's say I'm doing a test where I am making a step fail (through mockStepError), since the step failed and an error was thrown, workerd will print that out. This might be normal? The tests run accordingly, but can be confusing to see those logs appearing.

---

- Tests
  - [x] Tests included: in fixtures
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/24832
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Workflows were not GA in v3

### DIFF
--- a/fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
+++ b/fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
@@ -5,12 +5,11 @@ const STATUS_COMPLETE = "complete";
 const STEP_NAME = "AI content scan";
 const mockResult = { violationScore: 0 };
 
+// This example implicitly disposes the Workflow instance
 it("workflow should be able to reach the end and be successful", async () => {
-	// This example shows how to implicitly cleanUp Workflow instances
-
-	// CONFIG with `using` to ensure Workflow instances cleanup:
+	// CONFIG with `await using` to ensure Workflow instances cleanup:
 	await using introspector = await introspectWorkflow(env.MODERATOR);
-	introspector.modifyAll(async (m) => {
+	await introspector.modifyAll(async (m) => {
 		await m.disableSleeps();
 		await m.mockStepResult({ name: STEP_NAME }, mockResult);
 	});
@@ -25,36 +24,39 @@ it("workflow should be able to reach the end and be successful", async () => {
 	expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
 		mockResult
 	);
-	await instance.waitForStatus(STATUS_COMPLETE);
+	await expect(instance.waitForStatus(STATUS_COMPLETE)).resolves.not.toThrow();
 
-	// CLEANUP: ensured by `using`
+	// DISPOSE: ensured by `await using`
 });
 
+// This example explicitly disposes the Workflow instances
 it("workflow batch should be able to reach the end and be successful", async () => {
-	// This example shows how to explicitly cleanUp Workflow instances
-
 	// CONFIG:
 	let introspector = await introspectWorkflow(env.MODERATOR);
-	introspector.modifyAll(async (m) => {
-		await m.disableSleeps();
-		await m.mockStepResult({ name: STEP_NAME }, mockResult);
-	});
+	try {
+		await introspector.modifyAll(async (m) => {
+			await m.disableSleeps();
+			await m.mockStepResult({ name: STEP_NAME }, mockResult);
+		});
 
-	await SELF.fetch(`https://mock-worker.local/moderate-batch`);
+		await SELF.fetch(`https://mock-worker.local/moderate-batch`);
 
-	const instances = introspector.get();
-	expect(instances.length).toBe(3);
+		const instances = introspector.get();
+		expect(instances.length).toBe(3);
 
-	// ASSERTIONS:
-	for (const instance of instances) {
-		expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
-			mockResult
-		);
-		await instance.waitForStatus(STATUS_COMPLETE);
+		// ASSERTIONS:
+		for (const instance of instances) {
+			expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
+				mockResult
+			);
+			await expect(
+				instance.waitForStatus(STATUS_COMPLETE)
+			).resolves.not.toThrow();
+		}
+	} finally {
+		// DISPOSE:
+		// Workflow introspector should be disposed the end of each test, if no `await using` dyntax is used
+		// Also disposes all intercepted instances
+		await introspector.dispose();
 	}
-
-	// CLEANUP:
-	// Workflow introspector should be cleaned at the end of/after each test, if no `using` keyword is used for the introspector
-	// Cleans up all intercepted instances
-	await introspector.cleanUp();
 });
--- a/fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
+++ b/fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
@@ -6,9 +6,11 @@ const STATUS_COMPLETE = "complete";
 const STATUS_ERROR = "errored";
 const STEP_NAME = "AI content scan";
 
+// This example implicitly disposes the Workflow instance
 it("should mock a non-violation score and complete", async () => {
 	const mockResult = { violationScore: 0 };
 
+	// CONFIG with `await using` to ensure Workflow instances cleanup:
 	await using instance = await introspectWorkflowInstance(
 		env.MODERATOR,
 		INSTANCE_ID
@@ -22,6 +24,7 @@ it("should mock a non-violation score and complete", async () => {
 		id: INSTANCE_ID,
 	});
 
+	// ASSERTIONS:
 	expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
 		mockResult
 	);
@@ -29,38 +32,50 @@ it("should mock a non-violation score and complete", async () => {
 		await instance.waitForStepResult({ name: "auto approve content" })
 	).toEqual({ status: "auto_approved" });
 
-	await instance.waitForStatus(STATUS_COMPLETE);
+	await expect(instance.waitForStatus(STATUS_COMPLETE)).resolves.not.toThrow();
+
+	// DISPOSE: ensured by `await using`
 });
 
+// This example explicitly disposes the Workflow instance
 it("should mock the violation score calculation to fail 2 times and then complete", async () => {
 	const mockResult = { violationScore: 0 };
 
-	await using instance = await introspectWorkflowInstance(
-		env.MODERATOR,
-		INSTANCE_ID
-	);
-	await instance.modify(async (m) => {
-		await m.disableSleeps();
-		await m.mockStepError(
-			{ name: STEP_NAME },
-			new Error("Something went wrong!"),
-			2
+	// CONFIG:
+	const instance = await introspectWorkflowInstance(env.MODERATOR, INSTANCE_ID);
+
+	try {
+		await instance.modify(async (m) => {
+			await m.disableSleeps();
+			await m.mockStepError(
+				{ name: STEP_NAME },
+				new Error("Something went wrong!"),
+				2
+			);
+			await m.mockStepResult({ name: STEP_NAME }, mockResult);
+		});
+
+		await env.MODERATOR.create({
+			id: INSTANCE_ID,
+		});
+
+		// ASSERTIONS:
+		expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
+			mockResult
 		);
-		await m.mockStepResult({ name: STEP_NAME }, mockResult);
-	});
+		expect(
+			await instance.waitForStepResult({ name: "auto approve content" })
+		).toEqual({ status: "auto_approved" });
 
-	await env.MODERATOR.create({
-		id: INSTANCE_ID,
-	});
-
-	expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
-		mockResult
-	);
-	expect(
-		await instance.waitForStepResult({ name: "auto approve content" })
-	).toEqual({ status: "auto_approved" });
-
-	await instance.waitForStatus(STATUS_COMPLETE);
+		await expect(
+			instance.waitForStatus(STATUS_COMPLETE)
+		).resolves.not.toThrow();
+	} finally {
+		// DISPOSE:
+		// Workflow introspector should be disposed the end of each test, if no `await using` dyntax is used
+		// Also disposes all intercepted instances
+		await instance.dispose();
+	}
 });
 
 it("should mock a violation score and complete", async () => {
@@ -86,7 +101,7 @@ it("should mock a violation score and complete", async () => {
 		await instance.waitForStepResult({ name: "auto reject content" })
 	).toEqual({ status: "auto_rejected" });
 
-	await instance.waitForStatus(STATUS_COMPLETE);
+	await expect(instance.waitForStatus(STATUS_COMPLETE)).resolves.not.toThrow();
 });
 
 it("should be reviewed, accepted and complete", async () => {
@@ -116,7 +131,7 @@ it("should be reviewed, accepted and complete", async () => {
 		await instance.waitForStepResult({ name: "apply moderator decision" })
 	).toEqual({ status: "moderated", decision: "approve" });
 
-	await instance.waitForStatus(STATUS_COMPLETE);
+	await expect(instance.waitForStatus(STATUS_COMPLETE)).resolves.not.toThrow();
 });
 
 it("should force human review to timeout and error", async () => {
@@ -140,5 +155,5 @@ it("should force human review to timeout and error", async () => {
 		mockResult
 	);
 
-	await instance.waitForStatus(STATUS_ERROR);
+	await expect(instance.waitForStatus(STATUS_ERROR)).resolves.not.toThrow();
 });
--- a/packages/vitest-pool-workers/src/worker/workflows.ts
+++ b/packages/vitest-pool-workers/src/worker/workflows.ts
@@ -20,9 +20,14 @@ export interface WorkflowInstanceIntrospector {
 
 	waitForStatus(status: string): Promise<void>;
 
-	cleanUp(): Promise<void>;
+	dispose(): Promise<void>;
 }
 
+// Note(osilva): `introspectWorkflowInstance()` doesn’t need to be async, but we keep it that way
+// to avoid potential breaking changes later and to stay consistent with `introspectWorkflow`.
+
+// In the "cloudflare:test" module, the exposed type is `Workflow`. Here we use `WorkflowBinding`
+// (which implements `Workflow`) to access unsafe functions.
 export async function introspectWorkflowInstance(
 	workflow: WorkflowBinding,
 	instanceId: string
@@ -85,24 +90,29 @@ class WorkflowInstanceIntrospectorHandle
 		await this.#workflow.unsafeWaitForStatus(this.#instanceId, status);
 	}
 
-	async cleanUp(): Promise<void> {
-		await this.#workflow.unsafeAbort(this.#instanceId, "Instance clean up");
+	async dispose(): Promise<void> {
+		await this.#workflow.unsafeAbort(this.#instanceId, "Instance dispose");
 	}
 
 	async [Symbol.asyncDispose](): Promise<void> {
-		await this.cleanUp();
+		await this.dispose();
 	}
 }
 
 // See public facing `cloudflare:test` types for docs
 export interface WorkflowIntrospector {
-	modifyAll(fn: ModifierCallback): void;
+	modifyAll(fn: ModifierCallback): Promise<void>;
 
 	get(): WorkflowInstanceIntrospector[];
 
-	cleanUp(): Promise<void>;
+	dispose(): Promise<void>;
 }
 
+// Note(osilva): `introspectWorkflow` could be sync with some changes, but we keep it async
+// to avoid potential breaking changes later.
+
+// In the "cloudflare:test" module, the exposed type is `Workflow`. Here we use `WorkflowBinding`
+// (which implements `Workflow`) to access unsafe functions.
 export async function introspectWorkflow(
 	workflow: WorkflowBinding
 ): Promise<WorkflowIntrospectorHandle> {
@@ -189,7 +199,7 @@ export async function introspectWorkflow(
 		};
 	};
 
-	const cleanup = () => {
+	const dispose = () => {
 		internalEnv[bindingName] = internalOriginalWorkflow;
 		env[bindingName] = externalOriginalWorkflow;
 	};
@@ -209,7 +219,7 @@ export async function introspectWorkflow(
 		workflow,
 		modifierCallbacks,
 		instanceIntrospectors,
-		cleanup
+		dispose
 	);
 }
 
@@ -217,21 +227,21 @@ class WorkflowIntrospectorHandle implements WorkflowIntrospector {
 	workflow: WorkflowBinding;
 	#modifierCallbacks: ModifierCallback[];
 	#instanceIntrospectors: WorkflowInstanceIntrospector[];
-	#cleanupCallback: () => void;
+	#disposeCallback: () => void;
 
 	constructor(
 		workflow: WorkflowBinding,
 		modifierCallbacks: ModifierCallback[],
 		instanceIntrospectors: WorkflowInstanceIntrospector[],
-		cleanupCallback: () => void
+		disposeCallback: () => void
 	) {
 		this.workflow = workflow;
 		this.#modifierCallbacks = modifierCallbacks;
 		this.#instanceIntrospectors = instanceIntrospectors;
-		this.#cleanupCallback = cleanupCallback;
+		this.#disposeCallback = disposeCallback;
 	}
 
-	modifyAll(fn: ModifierCallback): void {
+	async modifyAll(fn: ModifierCallback): Promise<void> {
 		this.#modifierCallbacks.push(fn);
 	}
 
@@ -239,17 +249,17 @@ class WorkflowIntrospectorHandle implements WorkflowIntrospector {
 		return this.#instanceIntrospectors;
 	}
 
-	async cleanUp(): Promise<void> {
-		// also cleans all instance introspectors
+	async dispose(): Promise<void> {
+		// also disposes all instance introspectors
 		await Promise.all(
-			this.#instanceIntrospectors.map((introspector) => introspector.cleanUp())
+			this.#instanceIntrospectors.map((introspector) => introspector.dispose())
 		);
 		this.#modifierCallbacks = [];
 		this.#instanceIntrospectors = [];
-		this.#cleanupCallback();
+		this.#disposeCallback();
 	}
 
 	async [Symbol.asyncDispose](): Promise<void> {
-		await this.cleanUp();
+		await this.dispose();
 	}
 }
--- a/packages/vitest-pool-workers/types/cloudflare-test.d.ts
+++ b/packages/vitest-pool-workers/types/cloudflare-test.d.ts
@@ -129,7 +129,7 @@ declare module "cloudflare:test" {
 
 	/**
 	 * Creates an introspector for a specific Workflow instance, used to
-	 * modify its behavior, await outcomes, and clean up its state during tests.
+	 * modify its behavior and await outcomes.
 	 * This is the primary entry point for testing individual Workflow instances.
 	 *
 	 * @param workflow - The Workflow binding, e.g., `env.MY_WORKFLOW`.
@@ -137,35 +137,35 @@ declare module "cloudflare:test" {
 	 * @returns A `WorkflowInstanceIntrospector` to control the instance's behavior.
 	 *
 	 * @remarks
-	 * ### Cleanup
+	 * ### Dispose
 	 *
-	 * The introspector must be cleaned up after the test to remove mocks and release
+	 * The introspector must be disposed after the test to remove mocks and release
 	 * resources. This can be handled in two ways:
 	 *
-	 * 1.  **Implicit Cleanup (Recommended)**: With the `await using` syntax.
+	 * 1.  **Implicit dispose**: With the `await using` syntax.
 	 * `await using instance = await introspectWorkflowInstance(...)`
 	 *
-	 * 2.  **Explicit Cleanup**: Manually call `await instance.cleanUp()` at the end of the
+	 * 2.  **Explicit dispose**: Manually call `await instance.dispose()` at the end of the
 	 * test.
 	 *
 	 * @example
-	 * // Full test of a Workflow instance using implicit cleanup (await using).
+	 * // Full test of a Workflow instance using implicit dispose
 	 * it("should disable all sleeps and complete", async () => {
-	 * // 1. CONFIGURATION
-	 * // `await using` ensures .cleanUp() is automatically called at the end of the block.
-	 * await using instance = await introspectWorkflowInstance(env.MY_WORKFLOW, "123456");
+	 *   // 1. CONFIGURATION
+	 *   // `await using` ensures .dispose() is automatically called at the end of the block.
+	 *   await using instance = await introspectWorkflowInstance(env.MY_WORKFLOW, "123456");
 	 *
-	 * await instance.modify(async (m) => {
-	 * await m.disableSleeps();
-	 * });
+	 *   await instance.modify(async (m) => {
+	 *     await m.disableSleeps();
+	 *   });
 	 *
-	 * // 2. EXECUTION
-	 * await env.MY_WORKFLOW.create({ id: "123456" });
+	 *   // 2. EXECUTION
+	 *   await env.MY_WORKFLOW.create({ id: "123456" });
 	 *
-	 * // 3. ASSERTION
-	 * await instance.waitForStatus("complete");
+	 *   // 3. ASSERTION
+	 *   await instance.waitForStatus("complete");
 	 *
-	 * // 4. CLEANUP is implicit and automatic here.
+	 *   // 4. DISPOSE is implicit and automatic here.
 	 * });
 	 */
 	export function introspectWorkflowInstance(
@@ -212,26 +212,27 @@ declare module "cloudflare:test" {
 		waitForStatus(status: InstanceStatus["status"]): Promise<void>;
 
 		/**
-		 * Cleans the Workflow instance introspector.
+		 * Disposes the Workflow instance introspector.
 		 *
 		 * This is crucial for ensuring test isolation by preventing state from
-		 * leaking between tests. It should be called at the end or after each test.
+		 * leaking between tests. It should be called at the end of each test.
 		 */
-		cleanUp(): Promise<void>;
+		dispose(): Promise<void>;
 
 		/**
-		 * An alias for {@link cleanUp} to support automatic disposal with the `using` keyword.
-		 * This is an alternative to calling `cleanUp()` in an `afterEach` hook.
+		 * An alias for {@link dispose} to support automatic disposal with the `using` keyword.
 		 *
-		 * @see {@link cleanUp}
+		 * @see {@link dispose}
 		 * @example
+		 * ```ts
 		 * it('my workflow test', async () => {
-		 * await using instance = await introspectWorkflowInstance(env.WORKFLOW, "123456");
+		 *   await using instance = await introspectWorkflowInstance(env.WORKFLOW, "123456");
 		 *
-		 * // ... your test logic ...
+		 *   // ... your test logic ...
 		 *
-		 * // .cleanUp() is automatically called here at the end of the scope
+		 *   // .dispose() is automatically called here at the end of the scope
 		 * });
+		 * ```
 		 */
 		[Symbol.asyncDispose](): Promise<void>;
 	}
@@ -248,14 +249,16 @@ declare module "cloudflare:test" {
 		 * @example Disable all sleeps:
 		 * ```ts
 		 * await instance.modify(m => {
-		 * m.disableSleeps();
+		 *   m.disableSleeps();
 		 * });
+		 * ```
 		 *
 		 * @example Disable a specific set of sleeps by their step names:
 		 * ```ts
 		 * await instance.modify(m => {
-		 * m.disableSleeps([{ name: "sleep1" }, { name: "sleep5" }, { name: "sleep7" }]);
+		 *   m.disableSleeps([{ name: "sleep1" }, { name: "sleep5" }, { name: "sleep7" }]);
 		 * });
+		 * ```
 		 *
 		 * @param steps - Optional array of specific steps to disable sleeps for.
 		 * If omitted, **all sleeps** in the Workflow will be disabled.
@@ -279,10 +282,10 @@ declare module "cloudflare:test" {
 		 * @example Mock the result of the third step named "fetch-data":
 		 * ```ts
 		 * await instance.modify(m => {
-		 * m.mockStepResult(
-		 * { name: "fetch-data", index: 3 },
-		 * { success: true, data: [1, 2, 3] }
-		 * );
+		 *   m.mockStepResult(
+		 *     { name: "fetch-data", index: 3 },
+		 *     { success: true, data: [1, 2, 3] }
+		 *   );
 		 * });
 		 * ```
 		 */
@@ -300,16 +303,17 @@ declare module "cloudflare:test" {
 		 * ```ts
 		 * // This example assumes the "fetch-data" step is configured with at least 3 retries.
 		 * await instance.modify(m => {
-		 * m.mockStepError(
-		 * { name: "fetch-data" },
-		 * new Error("Failed!"),
-		 * 3
-		 * );
-		 * m.mockStepResult(
-		 * { name: "fetch-data" },
-		 * { success: true, data: [1, 2, 3] }
-		 * );
+		 *   m.mockStepError(
+		 *     { name: "fetch-data" },
+		 *     new Error("Failed!"),
+		 *     3
+		 *   );
+		 *   m.mockStepResult(
+		 *     { name: "fetch-data" },
+		 *     { success: true, data: [1, 2, 3] }
+		 *   );
 		 * });
+		 * ```
 		 *
 		 * @param step - An object specifying the step `name` and optional `index` (1-based).
 		 * If multiple steps share the same name, `index` targets a specific one.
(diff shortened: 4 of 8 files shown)


.changeset/public-women-exist.md
```diff
@@ -0,0 +1,13 @@
...
+"miniflare": patch
+---
+
+Workflows test support to "cloudflare:test" module.
+
+"cloudfalre:test" module has two new APIs: `introspectWorkflowInstance`and `itrospectWorkflow` to allow changing the behavior of one/multiple Workflow instances created during tests.
+
+workflows-shared has rpc functions to allow workflows instances chnages during tests.
+
+Miniflare DO plugin has an `unsafeScriptName` in DurableObjectsOptionsSchema, to allow defining a DO serviceName without the 'core:user:' prefix.
```
edmundhung: Can you split this to several changeset files? It's bit hard to follow as this will be duplicated on the changelogs of all 3 packages.

fixtures/vitest-pool-workers-examples/workflows/src/index.ts
```diff
@@ -7,22 +7,77 @@ import {
...
 			return Response.json(await instance.status());
 		}
 
-		const workflow = await this.env.TEST_WORKFLOW.create();
+		if (request.url.endsWith("/long-workflow")) {
+			const workflow = await env.TEST_LONG_WORKFLOW.create();
+			return new Response(JSON.stringify({ id: workflow.id }));
+		}
+
+		if (request.url.endsWith("/long-workflow-batch")) {
```
edmundhung: Nit: You can make the example cleaner with a shared URL object like this:

```
const maybeId = url.searchParams.get("id");

// ...

if (url.pathname === "/long-workflow") {
  // ...
}

// ...
```

fixtures/vitest-pool-workers-examples/workflows/src/index.ts
```diff
@@ -7,22 +7,77 @@ import {
...
+			const workflows = await env.TEST_LONG_WORKFLOW.createBatch([
+				{},
+				{ id: "321" },
+				{},
+			]);
+			const ids = workflows.map((workflow) => workflow.id);
+			return new Response(JSON.stringify({ ids: ids }));
```
edmundhung: Nit: You can use `Response.json()` to make the example cleaner

```
return Response.json({ ids: ids });
```

fixtures/vitest-pool-workers-examples/workflows/test/env.d.ts
```diff
@@ -1,6 +1,4 @@
 declare module "cloudflare:test" {
 	// Controls the type of `import("cloudflare:test").env`
-	interface ProvidedEnv extends Env {
-		TEST_WORKFLOW: Workflow;
-	}
```
edmundhung, line 5: Why are we removing the types here?

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+	it("workflow should reach the end and be successful", async () => {
+		const res = await SELF.fetch("https://mock-worker.local");
+
+		const json = await res.json<{ id: string }>();
+
+		await vi.waitUntil(async () => {
+			const res = await SELF.fetch(`https://mock-worker.local?id=${json.id}`);
+
+			const statusJson = await res.json<{ status: string }>();
+			console.log(statusJson);
```
edmundhung: Is the `console.log()` here intentional? I see a few more of them in the example.

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+	afterEach(async () => {
+		// CLEANUP:
+		// Instance introspectors should be clean to prevent persisted state across tests
+		for (const instance of instances) {
+			instance.cleanUp();
+		}
+
+		// Workflow introspector should be cleaned at the end of/after each test
+		introspector.cleanUp();
```
edmundhung: Can `introspector.cleanUp()` clean up the instance as well?

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+		}, 1000);
+	});
+
+	it("workflow should reach the end and be successful with introspector", async () => {
+		const introspector = await introspectWorkflow(env.TEST_WORKFLOW);
+
+		await SELF.fetch("https://mock-worker.local");
+
+		const instances = introspector.get();
+		assert(instances.length === 1);
```
edmundhung: Nit: It's a bit uncommon to use both `assert` and `expect` in the same test suite. Should we standardize it to `expect` only?

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
```diff
@@ -0,0 +1,730 @@
...
+					instance.waitForStepResult({ name: STEP_NAME })
+				).rejects.toThrow();
+				await instance.waitForStatus("errored");
+				return (await createdInstance.status()).status === "errored";
+			},
+			{
+				timeout: 1000,
+				interval: 1000,
+			}
+		);
```
edmundhung: Do you meant to create an instance within the `waitUntil`? This will create a new instance if it retries.

I would also recommend asserting the status directly, e.g.

```ts
const instanceStatus = await createdInstance.status();
expect(instanceStatus.status).toBe("errored");
```

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
```diff
@@ -0,0 +1,730 @@
...
+				expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
+					mockResult
+				);
+				await instance.waitForStatus("complete");
+				return (await createdInstance.status()).status === "complete";
+			},
+			{
+				timeout: 1000,
+				interval: 1000,
+			}
```
edmundhung: This seems unnecessary as the default timeout time is 1s unless you want to demonstrate something specifically.

packages/workflows-shared/src/engine.ts
```diff
@@ -191,12 +222,168 @@ export class Engine extends DurableObject<Env> {
...
+		reachedStatus: number,
+		unreachableStatuses: number[]
+	): void {
+		if (unreachableStatuses) {
+			for (const unreachableStatus of unreachableStatuses) {
+				const waiter = this.statusWaiters.get(unreachableStatus);
+				if (waiter) {
+					waiter.reject(
+						new Error(
+							`[WorkflowIntrospector] The Wokflow instance ${this.instanceId} has reached status '${instanceStatusName(reachedStatus)}'. This is a finite status that prevents it from ever reaching the expected status of '${instanceStatusName(unreachableStatus)}'.`
```
edmundhung:
```suggestion
							`[WorkflowIntrospector] The Workflow instance ${this.instanceId} has reached status '${instanceStatusName(reachedStatus)}'. This is a finite status that prevents it from ever reaching the expected status of '${instanceStatusName(unreachableStatus)}'.`
```

packages/miniflare/src/plugins/do/index.ts
```diff
@@ -34,6 +34,7 @@ export const DurableObjectsOptionsSchema = z.object({
 					unsafeUniqueKey: z
 						.union([z.string(), z.literal(kUnsafeEphemeralUniqueKey)])
 						.optional(),
+					unsafeScriptName: z.string().optional(),
```
edmundhung: I wonder there might be some side effects with this setup as the `scriptName` returned will be undefined when you are using `unsafeScriptName` only.

What about making it a boolean that decide whether the service name should add a user service prefix?

packages/miniflare/src/plugins/workflows/index.ts
```diff
@@ -106,7 +106,9 @@ export const WORKFLOWS_PLUGIN: Plugin<
 								className: "Engine",
 								enableSql: true,
 								uniqueKey,
-								preventEviction: true,
+								// Note(osilva): the engine should not be prevented from eviction if we wish
+								// to abort it during tests (vitest-pool-workers)
+								// preventEviction: true,
```
edmundhung: Would this have any impact to development?

packages/vitest-pool-workers/src/worker/workflows.ts
```diff
@@ -0,0 +1,258 @@
...
+	WorkflowInstanceModifier,
+} from "@cloudflare/workflows-shared/src/modifier";
+
+type ModifierCallback = (m: WorkflowInstanceModifier) => Promise<void>;
+
+// See public facing `cloudflare:test` types for docs
+export interface WorkflowInstanceIntrospector {
+	modify(fn: ModifierCallback): Promise<WorkflowInstanceIntrospector>;
+
+	waitForStepResult(step: StepSelector): Promise<unknown>;
```
edmundhung, line 19: A random last minute thought 😅 

Did we consider `waitForStepResult(name: string, index?: number): Promise<unknown>;` instead? Just feeling this would be a bit more compact to write and still readable. But maybe we have plan to introduce more options to the `StepSelector` in the future?

(Raised purely out of curiosity, no changes needed)

packages/vitest-pool-workers/src/worker/workflows.ts
```diff
@@ -0,0 +1,258 @@
...
+
+						await Promise.all(
+							argArray[0].map((options: { id: string }) =>
+								introspectAndModifyInstance(options.id)
+							)
+						);
+
+						const createPromises = (argArray[0] ?? []).map(
+							(arg: WorkflowInstanceCreateOptions) => target["create"](arg)
+						);
```
edmundhung, line 192: Why are we creating the first instance only? Should it call `createBatch` with the patched arguments?

packages/vitest-pool-workers/src/worker/workflows.ts
```diff
@@ -0,0 +1,258 @@
...
+
+	get(): WorkflowInstanceIntrospector[] {
+		return this.#instanceIntrospectors;
+	}
+
+	cleanUp(): void {
+		this.#modifierCallbacks = [];
+		this.#instanceIntrospectors = [];
+		this.#cleanupCallback();
+	}
```
edmundhung: A few suggestions:

- I would recommend calling `cleanUp()` on all introspectors instead of asking users to call it manually, unless there are some other usecases you have in mind.
- It might be worth supporting the `using` keyword here, like this:
  ```ts
  [Symbol.dispose]() {
    this.cleanUp();
  }
  ```

  This will remove the need to create and clean up introspector in the lifecycle hooks:
  ```ts
  test('workflow', () => {
      using instance = await introspectWorkflowInstance(
          env.TEST_LONG_WORKFLOW,
          INSTANCE_ID
      );
  
      await instance.modify(async (m) => {
          await m.disableSleeps();
      });
  
      const createdInstance = await env.TEST_LONG_WORKFLOW.create({
          id: INSTANCE_ID,
      });
  
      await vi.waitUntil(async () => {
          await instance.waitForStatus("complete");
          const instanceStatus = await createdInstance.status();
          expect(instanceStatus.status).toBe("complete");
      });
  });
  ```
- This is just a nitpick, but I thought I could still introspect the workflow after clean up as it looks like resetting it to an empty state, but it doesn't as the proxied workflow binding are pushing to the old array instead of `this.#instanceIntrospectors` after you have assigned it a new array. This is not an issue as it is not an intended usage, but might be worth adding some warnings in case people try to reuse the introspector.

packages/vitest-pool-workers/src/worker/workflows.ts
```diff
@@ -0,0 +1,258 @@
...
+
+	cleanUp(): void;
+}
+
+export async function introspectWorkflow(
+	workflow: Workflow
+): Promise<WorkflowIntrospectorHandle> {
+	if (!workflow) {
+		throw new Error("[WorkflowIntrospector] Workflow binding is required.");
+	}
```
edmundhung: Another random last minute thought. Feel free to ignore 😅 

I can't stop thinking whether `introspectWorkflow` and `introspectWorkflowInstance` should be just one API instead:

```ts
export async function introspectWorkflow(
	workflow: Workflow
): Promise<WorkflowIntrospectorHandle>;
export introspectWorkflow(
	workflow: Workflow,
  instanceId: string
): Promise<WorkflowInstanceIntrospector>;
export async function introspectWorkflow(
  workflow: Workflow,
  instanceId?: string
): Promise<WorkflowIntrospectorHandle> | Promise<WorkflowInstanceIntrospector> {
  if (typeof instanceId === 'string') {
    return introspectWorkflowInstance(workflow, instanceId);
  }

  // ... original implementation
}
```

packages/workflows-shared/src/context.ts
```diff
@@ -29,9 +29,9 @@ const defaultConfig: Required<WorkflowStepConfig> = {
 	retries: {
 		limit: 5,
 		delay: 1000,
-		backoff: "constant",
+		backoff: "exponential",
 	},
-	timeout: "15 minutes",
+	timeout: "10 minutes",
```
edmundhung, line 34: Why are we changing the defaults here? Do they match the defaults on production?

edmundhung (commented):
Great work! This is looking good so far! I don't have a deep understanding of the `workflow-shared` package, so it would be good for the workflow team to take a closer look at those changes.

I will give it another pass later this week 👍🏼

packages/workflows-shared/src/context.ts
pombosilva: Yes, now they match the production ones

packages/miniflare/src/plugins/workflows/index.ts
pombosilva: I commented this when I thought we could leverage abortAllDurableObjects. Since we haven't found a way of using that, I will remove the comment, since it doesn't affect the custom abortion being done in cleanUp()

packages/miniflare/src/plugins/do/index.ts
pombosilva: It's not only about removing the user prefix, but also about adding the "workflows:" prefix. 
But I can make it a boolean that decides if the service name should be the script name or the script name with user prefix. Then, when defining the DO simply put the script name with the "workflows:" prefix.

What do you think?

packages/vitest-pool-workers/src/worker/workflows.ts
pombosilva: We're not creating the first instance only. argArray in the case of createBatch() a list with only 1 position filled - that for createBatch means a list with WorkflowInstanceCreateOptions. That is the only argument ever being passed to createBatch.

I was calling createBatch directly, however Windows tests were crashing for createBatch tests, and changing that fixed it.

packages/miniflare/src/plugins/do/index.ts
edmundhung: Right, that would also result in an invalid `scriptName`. I think what you have is fine. Do you think if naming it `unsafeServiceName` would make its purpose more clear?

packages/vitest-pool-workers/src/worker/workflows.ts
pombosilva: This makes this function match the argument types we have for all the WorkflowInstanceModifier functions that receive a step identification and other arguments. And yes, if we do introduce more options, its way easier to apply changes 😅

fixtures/vitest-pool-workers-examples/workflows/test/env.d.ts
pombosilva: They are in src/env.d.ts. Just though of making it look cleaner since here test bindings don't differ from the bindings defined in Wrangler config

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
pombosilva: This console.log was here before but will be removed 👍

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
pombosilva: Yes 👍

.changeset/public-women-exist.md
```diff
@@ -0,0 +1,13 @@
+---
+"@cloudflare/vitest-pool-workers": minor
+"@cloudflare/workflows-shared": minor
+"miniflare": patch
+---
+
+Workflows test support to "cloudflare:test" module.
```
penalosa:
```suggestion
Add Workflows test support to the `cloudflare:test` module.
```

.changeset/public-women-exist.md
```diff
@@ -0,0 +1,13 @@
...
+
+"cloudfalre:test" module has two new APIs: `introspectWorkflowInstance`and `itrospectWorkflow` to allow changing the behavior of one/multiple Workflow instances created during tests.
```
penalosa:
```suggestion
The `cloudflare:test` module has two new APIs: 
- `introspectWorkflowInstance`
- `introspectWorkflow` 
which allow changing the behavior of one or multiple Workflow instances created during tests.
```

.changeset/public-women-exist.md
```diff
@@ -0,0 +1,13 @@
...
+
+workflows-shared has rpc functions to allow workflows instances chnages during tests.
```
penalosa:
```suggestion
```
this is not user facing

fixtures/vitest-pool-workers-examples/workflows/test/env.d.ts
edmundhung: Ah, I didnt notice that. Makes sense.

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
```diff
@@ -0,0 +1,107 @@
...
+		const res = await SELF.fetch("https://mock-worker.local");
+
+		const json = await res.json<{ id: string }>();
+
+		await vi.waitUntil(async () => {
+			const res = await SELF.fetch(`https://mock-worker.local?id=${json.id}`);
+
+			const statusJson = await res.json<{ status: string }>();
+			console.log(statusJson);
+			return statusJson.status === "complete";
```
penalosa:
```suggestion
			expect(statusJson.status).toBe("complete");
```

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
```diff
@@ -0,0 +1,730 @@
...
+		});
+
+		await vi.waitUntil(
+			async () => {
+				const createdInstance = await env.TEST_LONG_WORKFLOW.create({
+					id: INSTANCE_ID,
+				});
+
+				await instance.waitForStatus("complete");
+				return (await createdInstance.status()).status === "complete";
```
penalosa: This should be an `expect()`

packages/vitest-pool-workers/scripts/bundle.mjs
```diff
@@ -85,8 +85,32 @@ const commonOptions = {
...
+		"zod",
+		"undici",
+		"undici/*",
+		// Peer dependencies
+		"vitest",
+		"vitest/*",
+		"@vitest/runner",
+		"@vitest/snapshot",
+		"@vitest/snapshot/*",
+	],
```
penalosa, line 113: Could you expand on the motivation for this change?

packages/vitest-pool-workers/src/pool/index.ts
```diff
@@ -532,6 +533,16 @@ Workflows defined in project: ${workflowClassNames.join(", ")}`);
...
 
+	// Add Workflows Engines DOs bindings to the Runner Worker
+	for (const value of Object.values(runnerWorker.workflows ?? {})) {
+		const engineName = `${WORKFLOW_ENGINE_BINDING}${value.name.toUpperCase()}`;
+		runnerWorker.durableObjects[engineName] = {
+			className: "Engine",
+			unsafeScriptName: `workflows:${value.name}`,
+			unsafeUniqueKey: `miniflare-workflows-${value.name}`,
+		};
+	}
```
penalosa: I'm a bit worried about this strategy of connecting additional DOs to the runner worker in order to support accessing the workflow engine DO. Did we consider adding an unsafe method to a workflow binding to access the underlying engine instead? i.e. `env.WORKFLOW.unsafeGetEngine()`?

penalosa (commented):
Two top level comments ans some initial review comments:
- Is `introspect` the right name for this? I feel like `inspect` or `mock` would be simpler and more clearly represent what this does.
- We're building a _lot_ of custom APIs here. To what extent have we considered leaning into Vitest's mocking support instead of/in addition to some of these custom mocking APIs?

packages/vitest-pool-workers/src/worker/workflows.ts
pombosilva: Although you have a great point as this makes it cleaner, we (workflows team) think we should leave it as it is because these functions do have different behaviors. We like how it's more clear what each API does. But if we notice that users will prefer the other way around, we'll support it.

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
pombosilva: Yes it can clean all the instances intercepted. Will do that 👍

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
pombosilva: Since `timeout` is 1 second and `interval` is also 1 second, it will not retry at all.

Will add the assertion 👍

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
pombosilva: I could remove `timeout`, but `interval` default is 50, and I just want this to try once

packages/vitest-pool-workers/scripts/bundle.mjs
pombosilva: We couldn't import from workflows-shared due to bundling issues, and @edmundhung helped with this workaround. We just did not want to be copying types from workflows-shared into vitest-pool-workers, but if this change is not ideal I guess we could revert into copying what we need.

packages/vitest-pool-workers/src/pool/index.ts
pombosilva: I think we considered that at first and it did not work - we can't serialize DOs. So, this was the strategy that worked and we all agreed that it was ok to use.

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
pombosilva: I'll change every waitUntil into a timeout associated with the test, it's easier and serves the purpose!

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
pombosilva: Note: will actually remove the waitUntils and use the test timeout

packages/miniflare/src/plugins/do/index.ts
pombosilva: Like you said, if I have a boolean "unsafeScriptName" set to true, and the `scriptName` = 
"workflows:${name_of_workflow}", it would work. `scriptName` would not be invalid

packages/vitest-pool-workers/src/worker/workflows.ts
pombosilva: > I would recommend calling cleanUp() on all introspectors instead of asking users to call it manually, unless there are some other usecases you have in mind.

We require users to call cleanUp() because we don't know when they want to stop instrospecting. They could introspect in the first test and maintain the state of those workflow instances for the next tests... that a possible outcome.

> It might be worth supporting the using keyword here

That is clever, that way we can support both "automatic" cleanup with `using`, or manual without it. Might be worth it also adding this to WorkflowInstanceIntrospector. Thank you! Should we keep the cleanUp function or require to call `await introspector[Symbol.asyncDispose]();`?

> This is just a nitpick...

There is no trivial way of adding warnings, but I can document it!

pombosilva: Just figured if you have a big file of unit tests where you wish to cleanup/dispose instances/introspectors, the `using` keyword is block scoped, so it makes perfect sense to provide the `cleanUp` function as well. 
I know users can just call `await introspector[Symbol.asyncDispose]();`, but having a `cleanUp()` functions makes it cleaner 🤣 what do you think?

pombosilva:
> Two top level comments ans some initial review comments:
> 
> * Is `introspect` the right name for this? I feel like `inspect` or `mock` would be simpler and more clearly represent what this does.

It was the name we decided as a team!
`inspect` is not the best name because we can change the instance behavior
`mock` is not the best fit because we could just want to wait for step results and don't change the instance behavior at all

> * We're building a _lot_ of custom APIs here. To what extent have we considered leaning into Vitest's mocking support instead of/in addition to some of these custom mocking APIs?

We are able to make the workflow instances explicitly change their behavior. Without manipulating the Engine DO (that users shouldn't even have to worry about that it existis or not), users can't make any mocks

.changeset/public-women-exist.md
pombosilva: since I am spliting this into several changeset files, I don't add a workflows-shared changeset file?

packages/vitest-pool-workers/scripts/bundle.mjs
edmundhung: For context, `@cloudflare/workflows-shared` wasn't built or published. The `vitest-pool-workers` build script sets `packages: "external"`, which marks all dependencies as external. So the only way I can see to bundle a dependency here is to remove that setting and provide an explicit external list.

packages/vitest-pool-workers/src/pool/index.ts
edmundhung: Just discussed with @pombosilva and we are looking into exposing some engine methods through the workflow bindings directly.

pombosilva: Yes! Done 👍 No more Engine DO bindings to the runner worker

.changeset/public-women-exist.md
pombosilva: I now just have a changeset for vitest-pool-workers, since miniflare changes disapeared, and workflows-shared changes are not user facing. Is this ok?

packages/workflows-shared/src/context.ts
```diff
@@ -242,7 +242,11 @@ export class Context extends RpcTarget {
 			try {
 				const timeoutPromise = async () => {
 					const priorityQueueHash = `${cacheKey}-${stepState.attemptedCount}`;
-					const timeout = ms(config.timeout);
+					// inserir aqui
```
avenceslau: Please remove this comment

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const STATUS_COMPLETE = "complete";
+const STEP_NAME = "AI content scan";
+const mockResult = { violationScore: 0 };
+
+it("workflow should be able to reach the end and be successful", async () => {
+	// This example shows how to implicitly cleanUp Workflow instances
+
+	// CONFIG with `using` to ensure Workflow instances cleanup:
+	await using introspector = await introspectWorkflow(env.MODERATOR);
```
LuisDuarte1, line 11: non-blocking: is `await using` needed?  I guess it symbolizes `Symbol.asyncDispose` - but I wonder if there's a good way to doing it "synchronously" to have better syntax/DX? Maybe the runner DO can store the list of disposes and dispose it at the end of the test?

packages/vitest-pool-workers/src/worker/workflows.ts
```diff
@@ -0,0 +1,255 @@
...
+			"[WorkflowIntrospector] Workflow binding and instance id are required."
+		);
+	}
+	return new WorkflowInstanceIntrospectorHandle(workflow, instanceId);
+}
+
+class WorkflowInstanceIntrospectorHandle
+	implements WorkflowInstanceIntrospector
+{
+	#workflow: Workflow;
```
LuisDuarte1: to avoid these `@ts-except-errors` - maybe it's worth to extend the Workflows type here, like:  `Workflows & {unsafeWaitForStepResult: (...) => Promise<unknown> ... }`, wdyt? you might need to assert it, but in this context we know it's fine
Or get the "true" type from `workflows-shared`

packages/workflows-shared/src/modifier.ts
```diff
@@ -0,0 +1,165 @@
...
+	async mockStepResult(step: StepSelector, stepResult: unknown): Promise<void> {
+		const valueKey = await this.#getStepCacheKey(step);
+
+		if (await this.#state.storage.get(`replace-result-${valueKey}`)) {
+			throw new Error(
+				`[WorkflowIntrospector] Trying to mock step '${step.name}' multiple times!`
+			);
+		}
+
+		await this.#state.storage.put(`replace-result-${valueKey}`, stepResult);
```
LuisDuarte1, line 94: Took me a while to realize why you couldn't just put the correct value in `valueKey`  :sweat_smile:  - if I got it right it's because we need to `writeLog`s in a certain order and putting the value directly in `valueKey` would skip the log writes? 

maybe it's useful to add some documentation regarding the internal flow + maybe _why_ it's needed, wdyt?

packages/workflows-shared/src/engine.ts
```diff
@@ -191,12 +209,168 @@ export class Engine extends DurableObject<Env> {
...
+
+				this.rejectUnreachableStatus(status, unreachableStatuses);
+				break;
+			}
+			case InstanceStatus.Complete: {
+				// if it reaches final status "complete", then it can't be waiting for it to terminate or error
+				const unreachableStatuses = [
+					InstanceStatus.Terminated,
+					InstanceStatus.Errored,
+				];
```
LuisDuarte1, line 266: what if people wait for status that are _techinically not_ implemented by local dev? (but they still do exist) e.g: paused, queued, waiting. 

would personally prefer `unreachableStatuses` to be exhaustive, if possible, wdyt?

packages/vitest-pool-workers/src/worker/workflows.ts
pombosilva: Yep, way better. Will import WorkflowBinding

packages/workflows-shared/src/modifier.ts
pombosilva: Yes, if I put the value directly, it will assume it is a cached value and won't write any logs.

Will do, thanks 👍

packages/workflows-shared/src/engine.ts
pombosilva: This was intended for finite states only. For existing non-implemented status the user test would eventually timeout. We could eventually think of a better approach

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
pombosilva: @edmundhung can we do this?

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
```diff
@@ -0,0 +1,60 @@
...
+		expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
+			mockResult
+		);
+		await instance.waitForStatus(STATUS_COMPLETE);
+	}
+
+	// CLEANUP:
+	// Workflow introspector should be cleaned at the end of/after each test, if no `using` keyword is used for the introspector
+	// Cleans up all intercepted instances
+	await introspector.cleanUp();
```
edmundhung: Please use try-finally clause in this case

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
```diff
@@ -0,0 +1,144 @@
...
+	});
+
+	expect(await instance.waitForStepResult({ name: STEP_NAME })).toEqual(
+		mockResult
+	);
+	expect(
+		await instance.waitForStepResult({ name: "auto approve content" })
+	).toEqual({ status: "auto_approved" });
+
+	await instance.waitForStatus(STATUS_COMPLETE);
```
edmundhung: Nit: You can consider asserting this by `expect(() => instance.waitForStatus(STATUS_COMPLETE)).not.toThrow()`

packages/vitest-pool-workers/src/worker/workflows.ts
```diff
@@ -0,0 +1,255 @@
...
+
+// See public facing `cloudflare:test` types for docs
+export interface WorkflowInstanceIntrospector {
+	modify(fn: ModifierCallback): Promise<WorkflowInstanceIntrospector>;
+
+	waitForStepResult(step: StepSelector): Promise<unknown>;
+
+	waitForStatus(status: string): Promise<void>;
+
+	cleanUp(): Promise<void>;
```
edmundhung: Nit: Consider `dispose`

packages/vitest-pool-workers/types/cloudflare-test.d.ts
```diff
@@ -127,6 +127,363 @@ declare module "cloudflare:test" {
...
+		 * ```ts
+		 * await instance.modify(m => {
+		 * m.disableSleeps();
+		 * });
+		 *
+		 * @example Disable a specific set of sleeps by their step names:
+		 * ```ts
+		 * await instance.modify(m => {
+		 * m.disableSleeps([{ name: "sleep1" }, { name: "sleep5" }, { name: "sleep7" }]);
+		 * });
```
edmundhung: Formatting issue

packages/vitest-pool-workers/types/cloudflare-test.d.ts
```diff
@@ -127,6 +127,363 @@ declare module "cloudflare:test" {
...
+	 * **all subsequently created instances**.
+	 *
+	 * This is the primary entry point for testing Workflow instances where the id is unknown before their creation.
+	 *
+	 * @param workflow - The Workflow binding, e.g., `env.MY_WORKFLOW`.
+	 * @returns A `WorkflowIntrospector` to control the instances behavior.
+	 *
+	 *  * @remarks
+	 * ### Cleanup
+	 *
```
edmundhung: Formatting issue

packages/workflows-shared/src/binding.ts
```diff
@@ -78,6 +79,47 @@ export class WorkflowBinding extends WorkerEntrypoint<Env> implements Workflow {
 
 		return await Promise.all(batch.map((val) => this.create(val)));
 	}
+
+	public unsafeGetBindingName(): string {
+		return this.env.BINDING_NAME;
+	}
+
+	public unsafeGetInstanceModifier(id: string): unknown {
```
edmundhung:
```suggestion
	public unsafeGetInstanceModifier(instanceId: string): unknown {
```

packages/workflows-shared/src/engine.ts
```diff
@@ -133,10 +137,24 @@ export class Engine extends DurableObject<Env> {
...
+	}
+
+	readLogsFromStep(cacheKey: string): RawInstanceLog[] {
+		return [
+			...this.ctx.storage.sql.exec(
+				"SELECT * FROM states WHERE groupKey = ? ORDER BY id",
+				cacheKey
+			),
+		] as RawInstanceLog[];
 	}
```
edmundhung: Please check if this is still needed.

packages/workflows-shared/src/engine.ts
```diff
@@ -364,6 +542,9 @@ export class Engine extends DurableObject<Env> {
 			});
 		};
 		this.isRunning = true;
+
+		const stubStep = new Context(this, this.ctx);
+
```
edmundhung: Check if moving this down is intended.

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
edmundhung: Or using [onTestFinished](https://vitest.dev/api/#ontestfinished), e.g.

```ts
const introspector = await introspectWorkflow(env.MODERATOR);

onTestFinished(() => introspector.cleanUp())
```

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
pombosilva: Decided on not worrying about this for now. Will be an improvement to be done later

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
pombosilva: try-finally clause is better, because the `onTestFinished` vitest hook "is always called after the test has finished running", leading the test to fail with isolated storage.

packages/workflows-shared/src/engine.ts
pombosilva: It does not appear to make any change (although it would match the order in internal code). But will revert

fixtures/vitest-pool-workers-examples/workflows/test/unit.test.ts
pombosilva: Must be `await expect(instance.waitForStatus(STATUS_COMPLETE)).resolves.not.toThrow();` because it is async

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
jahands: I actually prefer `await using` for stuff like this because I've found them to be more reliable than `afterEach` hooks (which I'm guessing would be used in the sync version proposed?)

By reliable, I mean I've had tests hang/give weird errors when doing certain things in `afterEach` that went away when I switched to `await using`

packages/vitest-pool-workers/src/worker/workflows.ts
```diff
@@ -0,0 +1,265 @@
...
+
+	dispose(): Promise<void>;
+}
+
+// Note(osilva): `introspectWorkflowInstance()` doesn’t need to be async, but we keep it that way
+// to avoid potential breaking changes later and to stay consistent with `introspectWorkflow`.
+
+// In the "cloudflare:test" module, the exposed type is `Workflow`. Here we use `WorkflowBinding`
+// (which implements `Workflow`) to access unsafe functions.
+export async function introspectWorkflowInstance(
```
jahands, line 31: Could we include an example in each helper method where users should use `await using`? That'd make it a lot easier to remember where it should be used.

Doesn't have to be extensive, something like:
```ts
/**
 * @example
 * ```ts
 * // IMPORTANT: Always use `await using`
 * await using introspectWorkflowInstance(...)
 * ```
 */
```

pombosilva: We have examples on the file that exports the functions [here](https://github.com/cloudflare/workers-sdk/pull/10494/files#diff-d2ee7735ffb5a70b9cdf56c0e44264f7056fd853bcea83ee905e302c8306f25c) 

I can change how I am explaining it if its not that trivial to understand, wdyt?

```js
* @example
* // Full test of a Workflow instance using implicit dispose
* it("should disable all sleeps and complete", async () => {
*   // 1. CONFIGURATION
*   // `await using` ensures .dispose() is automatically called at the end of the block.
*   await using instance = await introspectWorkflowInstance(env.MY_WORKFLOW, "123456");
*
*   await instance.modify(async (m) => {
*     await m.disableSleeps();
*   });
*
*   // 2. EXECUTION
*   await env.MY_WORKFLOW.create({ id: "123456" });
*
*   // 3. ASSERTION
*   await instance.waitForStatus("complete");
*
*   // 4. DISPOSE is implicit and automatic here.
* });
```

fixtures/vitest-pool-workers-examples/workflows/test/integration.test.ts
pombosilva: The problem with `afterEach` is that it would break with isolated storage - without isolated storage works just fine. Isolated storage makes its assertions at the end of each test and therefore we need to enforce dispose before the test ends to avoid errors

jahands: TIL! Thank you for explaining

packages/vitest-pool-workers/src/worker/workflows.ts
jahands: oh didn't see that file! Looks great :)

PR #10497, opened 2025-08-29
Python templates: include pywrangler dev dep in pyproject.toml

hoodmane:
I also renamed the project from `"<TBD>"` to `"TBD"` because otherwise we get errors like:
```
Using CPython 3.12.9
error: Failed to parse `uv.lock`
  Caused by: TOML parse error at line 4, column 1
  |
4 | [[package]]
  | ^^^^^^^^^^^
Not a valid package or extra name: "<TBD>". Names must start and end with a letter or digit and may only contain -, _, ., and alphanumeric characters.
```

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: Just a template update
- Public documentation
  - [ ] Cloudflare docs PR(s):
  - [x] Documentation not necessary because: Just made the template a little better.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: beta feature

### DIFF
--- a/packages/create-cloudflare/src/templates.ts
+++ b/packages/create-cloudflare/src/templates.ts
@@ -821,6 +821,24 @@ export const downloadRemoteTemplate = async (
 	}
 };
 
+function updatePythonPackageName(path: string, projectName: string) {
+	const pyprojectTomlPath = resolve(path, "pyproject.toml");
+	const uvLockPath = resolve(path, "uv.lock");
+	if (!existsSync(pyprojectTomlPath)) {
+		// Not a python template
+		return;
+	}
+	const s = spinner();
+	s.start("Updating name in `pyproject.toml`");
+	let pyprojectTomlContents = readFile(pyprojectTomlPath);
+	pyprojectTomlContents = pyprojectTomlContents.replace('"TBD"', projectName);
+	writeFile(pyprojectTomlPath, pyprojectTomlContents);
+	let uvLockContents = readFile(uvLockPath);
+	uvLockContents = uvLockContents.replace('"tbd"', projectName.toLowerCase());
+	writeFile(uvLockPath, uvLockContents);
+	s.stop(`${brandColor("updated")} ${dim("`pyproject.toml`")}`);
+}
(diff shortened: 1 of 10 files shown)


dom96 (approved):

hoodmane:
Can someone explain changesets to me? I do not understand... @cloudflare/wrangler

dom96:
@hoodmane just run `npx changeset`

dom96:
> I also renamed the project from "\<TBD\>" to "TBD" because otherwise we get errors like

Doesn't wrangler specifically replace `<TBD>` though? this might break it

hoodmane:
Apparently it isn't replacing it.

hoodmane:
Maybe the problem is the uv.lock file.

packages/create-cloudflare/src/templates.ts
```diff
@@ -821,6 +821,24 @@ export const downloadRemoteTemplate = async (
...
+	const pyprojectTomlPath = resolve(path, "pyproject.toml");
+	const uvLockPath = resolve(path, "uv.lock");
+	if (!existsSync(pyprojectTomlPath)) {
+		// Not a python template
+		return;
+	}
+	const s = spinner();
+	s.start("Updating name in `pyproject.toml`");
+	let pyprojectTomlContents = readFile(pyprojectTomlPath);
+	pyprojectTomlContents = pyprojectTomlContents.replace('"TBD"', projectName);
```
jamesopstad, line 834: Why not use `<TBD>` here for consistency with the replacements in `package.json` and `wrangler.jsonc`?

hoodmane: This way the files are valid before the replacement is applied which is convenient. With the name as `<TBD>` various tools seem to get angry that it's not a valid name. But it would work fine to use `<TBD>` too...

hoodmane:
I don't seem to have the right to merge PRs to workers-sdk, so if you are happy with it @jamesopstad I'd appreciate if you could merge it.

PR #10500, opened 2025-08-29
Return 403 for /_image with href redirect using double slash

WillTaylorDev:

_Describe your change..._

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: time sensitive fix
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: n/a


packages/workers-shared/router-worker/src/worker.ts
```diff
@@ -128,6 +128,26 @@ export default {
 						}
 					}
 
+					if (url.pathname === "/_image") {
+						const hrefParam = url.searchParams.get("href");
+						if (hrefParam && hrefParam.match(/^\/\/.+/)) {
+							try {
+								const hrefUrl = new URL(
+									hrefParam.startsWith("//") ? "https:" + hrefParam : hrefParam
+								);
```
vicb: This code is only executed if `hrefParam.match(/^\/\/.+/)` (l133)

```suggestion
								const hrefUrl = new URL("https:" + hrefParam);
```

Also  `hrefParam.length > 2 && hrefParam.startsWith("//")` could fe faster than `hrefParam.match(/^\/\/.+/)` - if you decide the keep the regexp, the final `+` should not be needed?

vicb (approved):
LGTM with nits

PR #10503, opened 2025-08-29
Support setting container affinities

ichernetsky-cf:
Fixes CC-5971.

Allow setting container affinities on container creation and modification.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: because it is intended for internal use for now.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: AFAIU, containers are not supported with the previous version.

### DIFF
--- a/packages/wrangler/src/cloudchamber/apply.ts
+++ b/packages/wrangler/src/cloudchamber/apply.ts
@@ -42,6 +42,8 @@ import type {
 } from "../yargs-types";
 import type {
 	Application,
+	ApplicationAffinities,
+	ApplicationAffinityColocation,
 	ApplicationID,
 	ApplicationName,
 	CreateApplicationRequest,
@@ -50,6 +52,7 @@ import type {
 	Observability as ObservabilityConfiguration,
 	UserDeploymentConfiguration,
 } from "@cloudflare/containers-shared";
+import type { ApplicationAffinityHardwareGeneration } from "@cloudflare/containers-shared/src/client/models/ApplicationAffinityHardwareGeneration";
 import type { JsonMap } from "@iarna/toml";
 
 function mergeDeep<T>(target: T, source: Partial<T>): T {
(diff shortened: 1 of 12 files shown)


packages/wrangler/src/containers/config.ts
```diff
@@ -58,6 +61,26 @@ export const getNormalizedContainerOptions = async (
...
+
+		if (container.affinities?.hardware_generation !== undefined) {
+			if (affinities === undefined) {
+				affinities = {};
+			}
+
+			affinities.hardware_generation = container.affinities
+				.hardware_generation as ApplicationAffinityHardwareGeneration;
+		}
+
```
nikitassharma: can this be a function?

packages/wrangler/src/cloudchamber/apply.ts
```diff
@@ -245,6 +248,26 @@ function containerAppToCreateApplication(
 		});
 	}
 
+	let affinities: ApplicationAffinities | undefined = undefined;
+
+	if (containerApp.affinities?.colocation !== undefined) {
+		if (affinities === undefined) {
```
nikitassharma: i think this condition will always be true?

flakey5 (approved):

packages/wrangler/src/containers/config.ts
ichernetsky-cf: Absolutely! Thanks!

packages/wrangler/src/cloudchamber/apply.ts
ichernetsky-cf: The pattern is universal thus allowing for reordering of this `if`s, however either way works for me :)

nikitassharma: makes sense! thanks for explaining, i'm not particular either way, was just wondering

packages/wrangler/src/containers/config.ts
```diff
@@ -9,11 +9,42 @@ import {
...
+		if (affinities === undefined) {
+			affinities = {};
+		}
+
+		affinities.hardware_generation = container.affinities
+			.hardware_generation as ApplicationAffinityHardwareGeneration;
+	}
+
+	return affinities;
+}
```
emily-shen: nit but this can probably be simplified to:
```
	if(container.affinities?.colocation === undefined && container.affinities?.hardware_generation === undefined) {
		return undefined;
	}
	return {
		colocation: container.affinities?.colocation,
		hardware_generation: container.affinities?.hardware_generation
	} as ApplicationAffinityColocation;
```
since undefined keys are stripped when we make the request anyway.

ichernetsky-cf: :saluting_face:

PR #10507, opened 2025-08-30
Add strict mode for the `wrangler deploy` command

dario-piotrowicz:
Addresses the third point of #10235

This PR adds a new "strict mode" for the `wrangler deploy` command, for more details please check the changelog in the file changes

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: this is a new feature, so not something needing backporting

### DIFF
--- a/.changeset/calm-camels-return.md
+++ b/.changeset/calm-camels-return.md
@@ -1,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+Add strict mode for the `wrangler deploy` command
+
+Add a new flag: `--strict` that makes the `wrangler deploy` command be more strict/prudent and not deploy workers when such deployments can be potentially problematic. This "strict mode" currently only effects non-interactive sessions where conflicts with the remote settings for the worker (for example when the worker has been re-deployed via the dashboard) will cause the deployment to fail instead of automatically overriding the remote settings.
--- a/packages/wrangler/src/deploy/deploy.ts
+++ b/packages/wrangler/src/deploy/deploy.ts
@@ -23,12 +23,13 @@ import {
 import { noBundleWorker } from "../deployment-bundle/no-bundle-worker";
 import { validateNodeCompatMode } from "../deployment-bundle/node-compat";
 import { loadSourceMaps } from "../deployment-bundle/source-maps";
-import { confirm } from "../dialogs";
+import { confirm as genericConfirm } from "../dialogs";
 import { getMigrationsToUpload } from "../durable";
 import { getDockerPath } from "../environment-variables/misc-variables";
 import { UserError } from "../errors";
 import { getFlag } from "../experimental-flags";
 import { downloadWorkerConfig } from "../init";
+import { isNonInteractiveOrCI } from "../is-interactive";
 import { logger } from "../logger";
 import { getMetricsUsageHeaders } from "../metrics";
 import { isNavigatorDefined } from "../navigator-user-agent";
@@ -115,6 +116,8 @@ type Props = {
 	experimentalAutoCreate: boolean;
 	metafile: string | boolean | undefined;
 	containersRollout: "immediate" | "gradual" | undefined;
+	strict: boolean | undefined;
+	force: boolean | undefined;
 };
 
 export type RouteObject = ZoneIdRoute | ZoneNameRoute | CustomDomainRoute;
@@ -314,7 +317,7 @@ export async function publishCustomDomains(
 			const message = `Custom Domains already exist for these domains:
 ${existingRendered}
 Update them to point to this script instead?`;
-			if (!(await confirm(message))) {
+			if (!(await genericConfirm(message))) {
 				return fail();
 			}
 			options.override_existing_origin = true;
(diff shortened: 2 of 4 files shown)


packages/wrangler/src/deploy/index.ts
```diff
@@ -232,11 +232,24 @@ export const deployCommand = createCommand({
...
 			hidden: true,
 			alias: ["x-remote-diff-check"],
 		},
+		strict: {
+			describe:
+				"Enables strict mode for the deploy command, this prevents deployments to occur when there are even small potential risks.",
+			type: "boolean",
+			default: false,
+		},
+		// TODO: check, if `--force` really necessary? users can just provide or not `--strict`, no?
```
dario-piotrowicz: Question for reviewers 🙂 

(This todo should not be merged)

.changeset/calm-camels-return.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+Add strict mode for the `wrangler deploy` command
+
+Add a new flag: `--strict` that makes the `wrangler deploy` command be more strict/prudent and not deploy workers when such deployments can be potentially problematic. This "strict mode" currently only effects non-interactive sessions where conflicts with the remote settings for the worker (for example when the worker has been re-deployed via the dashboard) will cause the deployment to fail instead of automatically overriding the remote settings.
```
penalosa:
```suggestion
Add a new flag: `--strict` that makes the `wrangler deploy` command be more strict/prudent and not deploy workers when such deployments can be potentially problematic. This "strict mode" currently only affects non-interactive sessions where conflicts with the remote settings for the worker (for example when the worker has been re-deployed via the dashboard) will cause the deployment to fail instead of automatically overriding the remote settings.
```

packages/wrangler/src/deploy/index.ts
penalosa: That sounds pretty reasonable to me, yeah! (just having `--strict` and removing `--force`). @irvinebroque @mikenomitch do you have thoughts from a product perspective?

.changeset/calm-camels-return.md
dario-piotrowicz: thanks 🙏

packages/wrangler/src/deploy/index.ts
dario-piotrowicz: I added `--force` since it was included in the issue: https://github.com/cloudflare/workers-sdk/issues/10235

I don't have the full context on why `--force` was considered, I assume that it could be valuable for some systems to call always `wrangler deploy --strict` and then based on some condition add or not `--force`, but in such a system I do imagine that adding or not `--strict` would require pretty much the same effort anyways 🤔 

The other thing that comes to mind is that we could want `--force` if strict mode would become the default, but the flag could added when and if strict come does become the default (in a major bump)

penalosa: To be honest, I think I probably just didn't think about it that much when writing the issue. It was probably coming from a place of strict mode becoming the default in a future major, but you're right, we can add a force flag at that point

dario-piotrowicz: I've removed the `--force` flag as we've discussed async 🙂

.changeset/calm-camels-return.md
```diff
@@ -0,0 +1,7 @@
+---
+"wrangler": minor
+---
+
+Add strict mode for the `wrangler deploy` command
+
+Add a new flag: `--strict` that makes the `wrangler deploy` command be more strict/prudent and not deploy workers when such deployments can be potentially problematic. This "strict mode" currently only affects non-interactive sessions where conflicts with the remote settings for the worker (for example when the worker has been re-deployed via the dashboard) will cause the deployment to fail instead of automatically overriding the remote settings.
```
penalosa:
```suggestion
Add a new flag: `--strict` that makes the `wrangler deploy` command be more strict and not deploy workers when the deployment could be potentially problematic. This "strict mode" currently only affects non-interactive sessions where conflicts with the remote settings for the worker (for example when the worker has been re-deployed via the dashboard) will cause the deployment to fail instead of automatically overriding the remote settings.
```

packages/wrangler/src/deploy/deploy.ts
```diff
@@ -1430,3 +1434,21 @@ export async function updateQueueConsumers(
...
+
+function getDeployConfirmFunction(
+	strictMode = false
+): (text: string) => Promise<boolean> {
+	const nonInteractive = isNonInteractiveOrCI();
+
+	if (nonInteractive && strictMode) {
+		return async () => {
+			logger.error(
+				"Aborting the deployment operation (due to strict mode, to prevent this failure either remove the `--strict` flag)"
```
penalosa:
```suggestion
				"Aborting the deployment operation because of conflicts. To override and deploy anyway remove the `--strict` flag"
```

PR #10536, opened 2025-09-02
use the native `node:http2` when available

vicb:
`node:http2` was implemented in https://github.com/cloudflare/workerd/pull/4765, released in https://github.com/cloudflare/workerd/releases/tag/v1.20250814.0

unenv has a peer dep on workerd ^1.20250828.1 which includes this change.

Notes: 
- it is a deliberate choice not to fold `getHttp2Overrides` into `getHttpOverrides` in order to keep the preset code simple. The later already handle http modules and server modules and the dependency on server module to the base module
- ~tests are expected to fail until workerd is >= 20250901 in wrangler and vite - needed to test the default enabled date.~ workerd was updated to 2050902 in #10535 

/cc @anonrig @jasnell 

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: will be documented by the runtime team
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: unenv updates are not backported

### DIFF
--- a/.changeset/tiny-hotels-win.md
+++ b/.changeset/tiny-hotels-win.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/unenv-preset": patch
+---
+
+Use the native `node:http2` when available.
+
+It is enabled starting on 2025-09-01 or when the `enable_nodejs_http2_module` compatibility flag is set.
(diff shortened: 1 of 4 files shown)


jamesopstad (approved):

vicb:
Thanks for the review @jamesopstad

PR #10541, opened 2025-09-03
wrangler: move ratelimit binding to stable

qjex:
[Rate Limiting in Workers ](https://developers.cloudflare.com/workers/runtime-apis/bindings/rate-limit/) is now generally available. We can now remove `ratelimit` from unsafe bindings.

Old unsafe bindings should continue to work as before for now.  

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [x] Wrangler PR: TBD
  - [] Not necessary because:


packages/wrangler/src/config/environment.ts
```diff
@@ -1119,6 +1119,29 @@ export interface EnvironmentNonInheritable {
...
+	 *
+	 * NOTE: This field is not automatically inherited from the top level environment,
+	 * and so must be specified in every named environment.
+	 *
+	 * @default []
+	 * @nonInheritable
+	 */
+	ratelimit: {
+		/** The binding name used to refer to the rate limiter in the Worker. */
+		name: string;
```
penalosa, line 1163: I won't block on this because we're not 100% consistent, but I'd lean towards this being called `binding` rather than name

packages/wrangler/src/config/environment.ts
```diff
@@ -1119,6 +1119,29 @@ export interface EnvironmentNonInheritable {
...
+		/** The namespace ID for this rate limiter. */
+		namespace_id: string;
```
penalosa, line 1165: Where does this come from? Is there a CRUD API for creating these? If so, we should have Wrangler commands to speak to it (can be a followup)

packages/wrangler/src/config/environment.ts
```diff
@@ -1119,6 +1119,29 @@ export interface EnvironmentNonInheritable {
...
+	/**
+	 * Specifies rate limit bindings that are bound to this Worker environment.
+	 *
+	 * NOTE: This field is not automatically inherited from the top level environment,
+	 * and so must be specified in every named environment.
+	 *
+	 * @default []
+	 * @nonInheritable
+	 */
+	ratelimit: {
```
penalosa: I would lean towards this being `ratelimits` (plural)

packages/wrangler/src/type-generation/index.ts
```diff
@@ -454,6 +455,12 @@ export async function generateEnvTypes(
 		}
 	}
 
+	if (configToDTS.ratelimit) {
+		for (const ratelimit of configToDTS.ratelimit) {
+			envTypeStructure.push([constructTypeKey(ratelimit.name), "RateLimit"]);
```
penalosa, line 460: Has this type already been added to `workerd`?

penalosa (commented):
This should have a docs PR, and some tests. Probably a fixture test? I think there's an existing ratelimit one, so that could be extended

packages/wrangler/src/type-generation/index.ts
qjex: Yes, this PR just creates a safe version of the already existing unsafe binding for the Rate Limiting

packages/wrangler/src/config/environment.ts
qjex: No, it's just an arbitrary number specifying the universe of counters configured by a user. Technically, a use can have multiple bindings referring the same universe in order to count the same thing using from two bindings

penalosa: Is it a number or a string? This type seems to indicate a string

qjex: oh yeah, sure, it's a string and supposed to be a string. Think of it as an identity of that universe.

packages/wrangler/src/config/environment.ts
qjex: Leaving it as is to make it a drop-in replacement for the unsafe binding

qjex:
Thanks @penalosa for the review. I added a fixture test, and linked the docs PR

.changeset/five-drinks-stick.md
```diff
@@ -0,0 +1,8 @@
+---
+"wrangler": patch
+---
+
+stable `ratelimit` binding
+
+[Rate Limiting in Workers ](https://developers.cloudflare.com/workers/runtime-apis/bindings/rate-limit/) is now generally available, `ratelimit` can be removed from unsafe bindings.
```
petebacondarwin, line 7: The space here causes formatting checks to fail locally.
Not sure how that was not caught in CI?

petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/10630

PR #10542, opened 2025-09-03
[unenv] refactor http modules polyfill

vicb:
For historical reasons, the handling of the http modules (`http`, `https`, `_http_...`) was quite complex:

- the first implementation of the http modules did not include server related functions. Before [they were added](https://github.com/cloudflare/workerd/pull/4591), server APIs exports were missing from the workerd implementation. That is one of the reason why we needed an hybrid polyfill: the missing workerd exports were retrieved from `unenv`
- the first implementation of the `node:http` module was missing some of the APIs present in `unenv` They were added [at a later time](https://github.com/cloudflare/workerd/pull/4692). Note that all of the APIs exported from `node:https` have been present in `workerd` since its first implementation - it explains why the handling of `http` vs `https` was different before that PR 
- lastly the node modules and the node server modules used to be experimental.

The situation as of now (all changes are in workerd 20250828 which is the min peer dep of unenv) is much more simple:
- The experimental flag has been removed, all users can enable the http modules (depending on their compatibility date and flags)
- Both the native impl of `node:http` and `node:https` implement all of the node APIs

As such we do not need the hybrid polyfill any more but we can use the native workerd implementation directly - when the modules are enabled by compatibility date or flag.

Note: if the http modules are enabled (by date/flag) but the *server* modules are disabled (by date/flag) then workerd will throw "not implemented" when using any of the server APIs. 

/cc @anonrig @jasnell 

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: the existing tests already cover the http modules
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: unenv changes are not backported


anonrig (approved):

jamesopstad (approved):
The best PRs are the ones where you get to delete lots of stuff 😀.

vicb:
Thanks for the reviews Yagiz and James 🙏

PR #10545, opened 2025-09-03
Adds dummy workflow entrypoint field

Caio-Nogueira:
Fixes WOR-865.

Local dev for Python Workflows stopped working with the default entrypoint class. The reason for this is that named entrypoints wrap the `env` parameter to perform type conversions. Consequently the workflow binding was not considered convertible.

For example: 
```python
class Default(WorkerEntrypoint):
    async def fetch(self, request):
        await self.env.MY_WORKFLOW.create() # this would throw
```

We need to preserve the workflow as a `JsProxy`. For that we need to know that the object itself is a workflow binding. This PR accomplishes that.

---

- Tests
  - [ ] Tests included
  - [ ] Tests not necessary because: no need in asserting the added field
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [ ] Documentation not necessary because: this is merely for internal purposes
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because: not a wrangler change

### DIFF
--- a/packages/workflows-shared/src/binding.ts
+++ b/packages/workflows-shared/src/binding.ts
@@ -14,6 +14,7 @@ type Env = {
 
 // this.env.WORKFLOW is WorkflowBinding
 export class WorkflowBinding extends WorkerEntrypoint<Env> implements Workflow {
+	public readonly __workflow_entrypoint = true;
 	public async create({
 		id = crypto.randomUUID(),
 		params = {},


packages/workflows-shared/src/binding.ts
```diff
@@ -14,6 +14,7 @@ type Env = {
 
 // this.env.WORKFLOW is WorkflowBinding
 export class WorkflowBinding extends WorkerEntrypoint<Env> implements Workflow {
+	public readonly __workflow_entrypoint = true;
```
LuisDuarte1, line 17: nit: maybe call it `__workflow_binding` to not confuse ourselves with an actual `WorkflowEntrypoint`, wdyt?

petebacondarwin:
https://github.com/cloudflare/workerd/pull/4962#discussion_r2322599219

This comment suggests that this is not a viable fix?

vicb:
@Caio-Nogueira could you please answer Pete's comment above?

petebacondarwin:
Closing for now as this seems to have been forgotten and is stale and has conflicts.

PR #10546, opened 2025-09-03
feat: tag services and environments

1000hz:
This PR adds tagging behavior that applies appropriate service and environment tags to Workers with multiple environments. This metadata will be used by the Cloudflare Dashboard to group these Workers together in a more user-friendly way.

It also includes a new `definedEnvironments` computed field to Wrangler config that can be used to more easily determine whether a Worker uses environments.

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): _Pending..._
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: Not scoped for v3


packages/wrangler/src/config/config.ts
```diff
@@ -48,6 +50,10 @@ export interface ComputedFields {
 	 * It can be useful to know what the top-level name was before the flattening.
 	 */
 	topLevelName: string | undefined;
+	/** A list of environment names declared in the raw configuration. */
+	definedEnvironments: string[] | undefined;
```
penalosa, line 54: Why do we need `definedEnvironments`? It looks like it's only used to decide whether or not to apply environment tags, but surely the presence of a target environment is enough to decide that?

packages/wrangler/src/__tests__/deploy.test.ts
```diff
@@ -12904,6 +12905,428 @@ export default{
...
+				env: {
+					production: {},
+				},
+			});
+
+			expect.assertions(1);
+			let requestCount = 0;
+			mockPatchScriptSettings(async ({ request }) => {
+				requestCount++;
+				if (requestCount === 2) {
```
penalosa, line 13314: Why are we sending three requests to this endpoint?

packages/wrangler/src/config/config.ts
1000hz: target environment alone is not sufficient. we need to apply service tags as long as there are any environments defined in the config file, even when you're targeting the top-level environment (i.e. target environment would be empty).

packages/wrangler/src/__tests__/deploy.test.ts
1000hz: we are sending two. why not bundle setting the tags into the existing request? there's a chance updating the tags will fail if the additional tags push the worker beyond the tag limit, so I didn't want to jeopardize syncing the other script settings.

packages/wrangler/src/config/config.ts
penalosa: That makes sense, but to play devils advocate for a second, did you consider just always adding the service name tag, regardless of whether the user has environments in their config?

packages/wrangler/src/__tests__/deploy.test.ts
penalosa: Is there no way we can exempt these specific tags from the tag limit server side? I'm not going to block on this because we already do this all over the place in Wrangler, but I'd definitely prefer we get closer to 1 API call being a deploy, rather than further away

1000hz: yeah, now that we're qualifying these tags with a `cf:` prefix, we probably should exempt these two specific tags from the limit.

packages/wrangler/src/config/config.ts
1000hz: We potentially could, but there'd be some undesirable implications on other parts of the system that we'd need to handle (i.e. service grouping) and it'd be better to only apply it where necessary for now.

1000hz:
Spoke with @penalosa and decided I'll follow up this PR with another one to add additional test coverage

penalosa:
For posterity, after a lot of flakes this is now failing on Windows because of a test timeout. I'm going to merge this as-is in hopes that the caching of test passes on VP will enable Windows CI to pass

PR #10549, opened 2025-09-03
Adds the `--force-subdomain-deploy` flag to the `wrangler triggers deploy` command

danielrs:
Wrangler will sometimes skip the API call if the config file matches the remote state. The flag lets us skip this behavior, and force the subdomain deployment.

Follow up of: https://github.com/cloudflare/workers-sdk/pull/10478

_Describe your change..._

Adds `--force-subdomain-deploy` flag to `wrangler triggers deploy`.

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: experimental command.
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: experimental command.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: experimental command.

### DIFF
--- a/.changeset/moody-goats-cry.md
+++ b/.changeset/moody-goats-cry.md
@@ -1,0 +1,9 @@
+---
+"wrangler": minor
+---
+
+Adds the `--force-subdomain-deploy` flag to the `wrangler triggers deploy` command.
+
+Wrangler will sometimes skip the API call if the config file matches
+the remote state. The flag lets us skip this behavior, and force the
+subdomain deployment.
--- a/packages/wrangler/src/triggers/deploy.ts
+++ b/packages/wrangler/src/triggers/deploy.ts
@@ -30,6 +30,7 @@ type Props = {
 	legacyEnv: boolean | undefined;
 	dryRun: boolean | undefined;
 	assetsOptions: AssetsOptions | undefined;
+	forceSubdomainDeploy?: boolean;
 };
 
 export default async function triggersDeploy(
@@ -90,7 +91,8 @@ export default async function triggersDeploy(
 		envName,
 		workerUrl,
 		routes,
-		deployments
+		deployments,
+		props.forceSubdomainDeploy ?? false
 	);
 
 	if (!wantWorkersDev && workersDevInSync && routes.length !== 0) {
@@ -315,7 +317,8 @@ async function subdomainDeploy(
 	envName: string,
 	workerUrl: string,
 	routes: Route[],
-	deployments: Array<Promise<string[]>>
+	deployments: Array<Promise<string[]>>,
+	forceSubdomainDeploy: boolean
 ) {
 	const { config } = props;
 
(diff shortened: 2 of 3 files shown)


.changeset/moody-goats-cry.md
```diff
@@ -0,0 +1,5 @@
+---
```
vicb, line 1: Do we need a minor (vs a patch)?

It's not clear to me what this PR does by reading

> Adds `wrangler triggers deploy --force-subdomain-deploy` flag.

vicb: Ok it's more clear now by looking at the code.

Maybe update the description to read something like:

```
Adds a `--force-subdomain-deploy` flag to the `wrangler triggers deploy` command.

[Describe what the flag does/is used for]
```

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -315,7 +317,8 @@ async function subdomainDeploy(
 	envName: string,
 	workerUrl: string,
 	routes: Route[],
-	deployments: Array<Promise<string[]>>
+	deployments: Array<Promise<string[]>>,
+	forceSubdomainDeploy: boolean
```
vicb, line 321: Do we really need to add an arg, given that we pass `props.forceSubdomainDeploy` and props is already the first arg?

danielrs: I feel like it is good to be explicit about forcing here. I want to get rid of `props` arg eventually.

.changeset/moody-goats-cry.md
danielrs: Good catch, adding a description!

.changeset/moody-goats-cry.md
```diff
@@ -0,0 +1,9 @@
...
+"wrangler": minor
+---
+
+Adds `wrangler triggers deploy --force-subdomain-deploy` flag.
```
vicb: nit: still think the following wording better express what's new (`force-subdomain-deploy`) and being added to the existing `wrangler triggers deploy` command.

```suggestion
Adds a `--force-subdomain-deploy` flag to the `wrangler triggers deploy` command.
```

packages/wrangler/src/triggers/deploy.ts
vicb: Could you please reconsider this for now?

You can still explicitly pass `props.forceSubdomainDeploy` later at the time you refactor the code.

.changeset/moody-goats-cry.md
danielrs: Gotcha, updating.

.changeset/moody-goats-cry.md
edmundhung: Is there any docs or internal reference on what this does exactly? It wasn't clear to me why I want to force the subdomain deployment.

edmundhung (changes requested):
We need tests even for experimental command

danielrs:
We did some changes in the API that should make this flag less urgent.
Closing for now.

PR #10551, opened 2025-09-04
Auto-generate DO bindings and migrations from containers config

Talador12:
Fixes WOR-840

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  ----- this change will require changes to /containers first, and then we will update docs
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:  
  - [x] unknown


fixtures/container-app/README.md
```diff
@@ -1,3 +1,24 @@
 # HTTP fetch to a container
 
 This example shows a simple container setup where a DO passes requests through to a node server.
+
+## Simplified Container Configuration
+
+Only need to define the `[[containers]]` section in `wrangler.toml`. The Durable Object bindings and migrations are automatically generated:
```
vicb, line 7: The PR mentions "wrangler.toml" at multiple location, could you please use "wrangler configuration" instead (i.e. line 9 below is jsonc)

packages/wrangler/src/containers/config.ts
```diff
@@ -46,6 +46,81 @@ function convertContainerAffinitiesForApi(
  * We set defaults here too, because we can assume that if the value is undefined,
  * we want to revert to the default rather than inheriting from the prev deployment
  */
+
+
+/**
+ * Auto-generate Durable Object bindings and migrations for containers
+ * This reduces wrangler.toml boilerplate by creating necessary DO config automatically
+ */
+export const autoGenerateContainerDurableObjects = (config: Config): void => {
```
vicb, line 55: IMO one important thing is that this function *updates* the passed config.

Could you make that clear:
- In the JSDoc above
- in the function name (i.e. something like `addContainerDurableObjectsToConfig`)

packages/wrangler/src/containers/config.ts
```diff
@@ -46,6 +46,81 @@ function convertContainerAffinitiesForApi(
...
+
+	for (const container of config.containers) {
+		// Auto-generate class_name if not provided
+		if (!container.class_name) {
+			// Convert container name to PascalCase for class name
+			container.class_name = container.name
+				.split(/[-_]/)
+				.map(word => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase())
+				.join('') + 'Container';
+		}
```
vicb, line 81: IIUC this is a naming convention that users have to respect. If they name their class anything else than `nameContainer`, it won't work.

If that is the case, could you please document that convetion in the JS doc for this function and the README.

packages/wrangler/src/containers/config.ts
```diff
@@ -46,6 +46,81 @@ function convertContainerAffinitiesForApi(
...
+
+		// Auto-generate DO binding if it doesn't exist
+		const existingBinding = config.durable_objects.bindings.find(
+			binding => binding.class_name === container.class_name
+		);
+		
+		if (!existingBinding) {
+			// Generate binding name from container name (SCREAMING_SNAKE_CASE)
+			const bindingName = container.name.toUpperCase().replace(/[-]/g, '_');
```
vicb, line 90: nit
```suggestion
			const bindingName = container.name.toUpperCase().replace(/-/g, '_');
```

packages/wrangler/src/containers/config.ts
```diff
@@ -46,6 +46,81 @@ function convertContainerAffinitiesForApi(
...
+			
+			config.durable_objects.bindings.push({
+				class_name: container.class_name,
+				name: bindingName,
+			});
+		}
+
+		// Auto-generate migration if class doesn't exist in migrations
+		const hasExistingMigration = config.migrations.some(migration =>
+			migration.new_sqlite_classes?.includes(container.class_name!)
```
vicb, line 100: is the trailing `!` actually required here?

packages/wrangler/src/containers/config.ts
```diff
@@ -46,6 +46,81 @@ function convertContainerAffinitiesForApi(
...
+		);
+
+		if (!hasExistingMigration) {
+			// Find existing v1 migration or create new one
+			let v1Migration = config.migrations.find(migration => migration.tag === 'v1');
```
vicb, line 105: IIUC `v1` is another convention and users might have an existing valid migration with a different label.

If that's correct, please document the convention your code is using

vicb (commented):
I have mixed feelings about this PR

On one hand it's nice not to have to write the full config.

On the other hand the generated config is kind of brittle because it is based on convention (class name, tag name) and will break if users use different naming scheme.

IMO displaying how the config should be updated when the DO config is missing could be enough.

What do others think?

kentonv (changes requested):
Sorry, this is the first I've seen this, but I think this design conflicts with other work going on, and needs more review with the DO team:

1. `ctx.exports` will already make it unnecessary to declare DO bindings in the near future. We should focus on `ctx.exports` as the general solution and not design a second solution.
2. Similarly regarding migrations, we should design a solution that makes it better for durable objects in general, not a solution that is specific to containers. Note that making storage provisioning automatic is dangerous. The design of migrations was intended to avoid real pitfalls -- e.g., if you rename your class and redeploy, does all your data disappear? These pitfalls apply to containers too. Everyone agrees that the design we ended up with seems to have missed the mark, but we need to carefully consider the pitfalls in any new design.

PR #10554, opened 2025-09-04
Pin the `vercel` and `next` dependencies to their latest supported for C3 next-on-pages applications

dario-piotrowicz:
We've deprecated next-on-pages in https://github.com/cloudflare/next-on-pages/pull/986 and pinned the `vercel` and `next` dependencies to their current latest stable releases

As a result we also need to update C3 not to use any newer version of `vercel` and `next`, which is what this PR is doing

Additionally there's the question whether we should then remove next-on-pages as a C3 option, that will be discussed with the team and addressed separately

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: this functionality is already all tested
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not something needing documentation
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a wrangler feature

### DIFF
--- a/.changeset/hungry-chairs-invite.md
+++ b/.changeset/hungry-chairs-invite.md
@@ -1,0 +1,5 @@
+---
+"create-cloudflare": patch
+---
+
+Pin the `vercel` and `next` dependencies to their latest supported for next-on-pages applications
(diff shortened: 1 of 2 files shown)


dario-piotrowicz:
Closing since maybe we might just as well skip this step and remove the next-on-pages entry instead

vicb (approved):
LGTM.

As of today C3 e2e (npm/linux) are broken and I believe this would solve that and allow merging PRs top branch.

IMO we should merge this for now and keep discussing what we cant for c3/next-no-pages

PR #10558, opened 2025-09-05
feat: add R2 SQL command

laplab:
Fixes https://jira.cfdata.org/browse/KRB-49.

This change adds one command: `wrangler r2 sql query` - send an SQL query targeting an R2 Data Catalog.

This command relies on `CLOUDFLARE_API_TOKEN` env variable to provide an API token. For now, a separate token with specific permissions is required to connect to the R2 SQL service.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/issues/24988
- Wrangler V3 Backport
  - [x] Not necessary because: change is related to a new product

### DIFF
--- a/packages/wrangler/src/logger.ts
+++ b/packages/wrangler/src/logger.ts
@@ -126,19 +126,21 @@ export class Logger {
 		data: TableRow<Keys>[],
 		options?: { wordWrap: boolean; head?: Keys[] }
 	) {
+		const derivedHead =
+			data.length === 0 ? [] : (Object.keys(data[0]) as Keys[]);
+		const wordWrap = options?.wordWrap ?? false;
+		const head = options?.head ?? derivedHead;
+
 		const tableOptions = {
 			style: {
 				head: chalk.level ? ["blue"] : [],
 				border: chalk.level ? ["gray"] : [],
 			},
-			...options,
+			wordWrap,
+			head,
 		};
-		if (!tableOptions.head) {
-			tableOptions.head =
-				data.length === 0 ? [] : (Object.keys(data[0]) as Keys[]);
-		}
 		const t = new CLITable(tableOptions);
-		t.push(...data.map((row) => tableOptions.head!.map((k) => row[k])));
+		t.push(...data.map((row) => head.map((k) => row[k])));
 		return this.doLog("log", [t.toString()]);
 	}
 	console<M extends Exclude<keyof Console, "Console">>(


packages/wrangler/src/__tests__/r2/sql.test.ts
```diff
@@ -0,0 +1,369 @@
...
+		});
+	});
+
+	describe("query", () => {
+		const mockWarehouse = "account123_mybucket";
+		const mockQuery = "SELECT * FROM data";
+		const mockToken = "test-token-123";
+
+		beforeEach(() => {
+			process.env.CLOUDFLARE_R2_SQL_TOKEN = mockToken;
```
edmundhung: You can use [vi.stubEnv()](https://vitest.dev/api/vi.html#vi-stubenv) instead of manipulating the `process.env` directly.

```ts
// Override 
vi.stubEnv('CLOUDFLARE_R2_SQL_TOKEN', mockToken);
// Delete
vi.stubEnv('CLOUDFLARE_R2_SQL_TOKEN', undefined);
```

packages/wrangler/src/__tests__/r2/sql.test.ts
```diff
@@ -0,0 +1,369 @@
...
+					column_order: [],
+					rows: [],
+				},
+			};
+
+			const originalFetch = global.fetch;
+			global.fetch = vi.fn().mockResolvedValue({
+				status: 200,
+				text: async () => JSON.stringify(mockResponse),
+			});
```
edmundhung: Why? Can you use `msw` to mock the request?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,227 @@
...
+				headers: {
+					"Content-Type": "application/json",
+				},
+				body: JSON.stringify({}),
+			});
+
+			logger.log(`✅ R2 SQL is enabled for your account
+
+Try sending a query with \`wrangler r2 sql query <warehouse name> <SQL query>\`
+`);
```
edmundhung: What happens if it is already enabled?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,227 @@
...
+		const { warehouse, query } = args;
+
+		const token = getR2SqlAPITokenFromEnv();
+		if (!token) {
+			// TODO: provide documentation link.
+			throw new UserError(
+				"CLOUDFLARE_R2_SQL_TOKEN environment variable is not set. " +
+					"Please set it to authenticate with the R2 SQL query service."
+			);
+		}
```
edmundhung: Why can't we use the API token from `wrangler login`?

If you can't use `fetchResult` directly, you can call `requireApiToken()` to get the API token.

sejoker: We need API token with R2 SQL, R2 Data Catalog and R2 admin permissions to be able to perform necessary operations.

sejoker:
@laplab please update PR description with a brief intro about CLOUDFLARE_R2_SQL_TOKEN env variable.

packages/wrangler/src/r2/sql.ts
edmundhung: Can you extend the wrangler oauth scope to cover all the missing permission? This is not an established pattern in wrangler, so our users might find this additional token  confusing.

With `requireApiToken()`, it also allows people to override it with `CLOUDFLARE_API_KEY` env variable just like all other commands.

packages/wrangler/src/r2/sql.ts
laplab: It stays enabled

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,227 @@
+import CLITable from "cli-table3";
```
penalosa: You shouldn't need to import this directly. Instead, use the `logger.table()` utility

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,227 @@
...
+		status: "open-beta",
+		owner: "Product: R2 SQL",
+	},
+	async handler(args, { config }) {
+		const accountId = await requireAuth(config);
+
+		logger.log("Enabling R2 SQL for your account...");
+
+		try {
+			await fetchResult(config, `/accounts/${accountId}/dqe/enable`, {
```
penalosa: Not a blocker, but is this endpoint available in the SDK? If so, it would be great to use that

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,227 @@
...
+			warehouse.slice(splitIndex + 1),
+		];
+
+		let responseStatus = null;
+		let text = null;
+		let duration = null;
+		try {
+			const start = Date.now();
+			const response = await fetch(
+				`https://api.dqe.cloudflarestorage.com/api/v1/accounts/${accountId}/dqe/query/${bucketName}`,
```
penalosa: Why is this not behind API gateway?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,227 @@
...
+						"Content-Type": "application/json",
+					},
+					body: JSON.stringify({
+						warehouse,
+						query,
+					}),
+				}
+			);
+			responseStatus = response.status;
+			text = await response.text();
```
penalosa, line 127: Given this is being parsed to JSON, this could just be `response.json()`?

packages/wrangler/src/r2/sql.ts
laplab: This is a technical limitation of our auth architecture. Even with correct permissions, API token from Wrangler won't be enough for R2 SQL. Added some more details on this in the PR description.

packages/wrangler/src/r2/sql.ts
laplab: Two reasons:
- Mention text in the error message if it's not a valid json. Calling `response.json()` and then `response.text()` in case of exception is not safe because we might have already read all the text from the connection in `response.json()`
- To separate "connection errors" and "invalid json errors" in exceptions we throw. This might not be that useful

packages/wrangler/src/r2/sql.ts
laplab: Because of technical limitations of our auth platform.

packages/wrangler/src/r2/sql.ts
laplab: I don't think this works for us because the order of columns is important to us. `logger.table()` just takes whatever order `Object.keys()` returns. We have a field in the API that dictates what the order of columns should be.

packages/wrangler/src/r2/sql.ts
laplab: It's not available in the SDK. We might add it in the future

packages/wrangler/src/user/auth-variables.ts
```diff
@@ -111,3 +111,7 @@ export const getCloudflareAccessToken = async () => {
 
 	return getAccessToken(getAuthDomainFromEnv());
 };
+
+export const getR2SqlAPITokenFromEnv = getEnvironmentVariableFactory({
+	variableName: "CLOUDFLARE_R2_SQL_TOKEN",
+});
```
penalosa: This doesn't need an additional token. Instead, what about requiring users to provide `CLOUDFLARE_API_TOKEN`, which is our existing documented mechanism for accepting tokens in Wrangler?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,223 @@
...
+		if (splitIndex === -1) {
+			throw new UserError("Warehouse name has invalid format");
+		}
+		const [accountId, bucketName] = [
+			warehouse.slice(0, splitIndex),
+			warehouse.slice(splitIndex + 1),
+		];
+
+		const s = spinner();
+		s.start("Query in progress");
```
penalosa, line 106: How long do you expect these queries to take? We don't have spinners for any other network requests in Wrangler

packages/wrangler/src/r2/sql.ts
penalosa: In that case, what about adding an extra second parameter to `logger.table()` that lets you override the keys?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,170 @@
...
+			type: "string",
+			demandOption: true,
+		},
+	},
+	async handler(args, { config: _config }) {
+		const { warehouse, query } = args;
+
+		const token = getCloudflareAPITokenFromEnv();
+		if (!token) {
+			// TODO: provide documentation link.
```
penalosa: Can you add this before merging?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,170 @@
...
+			throw new UserError("Warehouse name has invalid format");
+		}
+		const [accountId, bucketName] = [
+			warehouse.slice(0, splitIndex),
+			warehouse.slice(splitIndex + 1),
+		];
+
+		const s = spinner();
+		s.start("Query in progress");
+		const apiUrl = `https://api.sql.cloudflarestorage.com/api/v1/accounts/${accountId}/r2-sql/query/${bucketName}`;
```
penalosa, line 107: How does this interact with fed? Most other API requests Wrangler makes can have a different base URL depending on whether fed is being used or not. Does this one?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,170 @@
...
+			demandOption: true,
+		},
+		query: {
+			describe: "The SQL query to execute",
+			type: "string",
+			demandOption: true,
+		},
+	},
+	async handler(args, { config: _config }) {
+		const { warehouse, query } = args;
```
penalosa:
```suggestion
	async handler({ warehouse, query }) {
```

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,170 @@
...
+
+		const token = getCloudflareAPITokenFromEnv();
+		if (!token) {
+			// TODO: provide documentation link.
+			throw new UserError(
+				"CLOUDFLARE_API_TOKEN environment variable is not set. " +
+					"Please set it to authenticate with the R2 SQL query service."
```
penalosa: Could you add some more explanation/docs link?

packages/wrangler/src/r2/sql.ts
penalosa: I see, makes sense. This mostly looks okay, but we already have a utility for making fetch requests in Wrangler (`fetchResult()`) that handles all this + debug logging etc... Can that be re-used at all? Perhaps with an additional parameter for overriding the base URL?

packages/wrangler/src/r2/sql.ts
laplab: Query can take anywhere from 5 seconds to 5 minutes

packages/wrangler/src/r2/sql.ts
laplab: We don't have a fed deployment at the moment

packages/wrangler/src/r2/sql.ts
laplab: I added a link and a more detailed explanation. The link itself is 404 right now, but we will add the docs before the official release.

packages/wrangler/src/r2/sql.ts
laplab: Done, see other thread.

packages/wrangler/src/r2/sql.ts
laplab: I don't think this can be done easily. `fetchResult` requires the user to login and passes the API token in the headers. We technically don't need the user to login and it can be confusing to show login prompt and then still ask for the token. `fetchResult` also adds a bunch of headers we do not need. It can still be done, but I think it will complicated already quite involved `fetchResult` and won't save us much effort for this specific command.

packages/wrangler/src/r2/sql.ts
penalosa: 5 minutes is likely to trip both API gateway and undici request timeout limits. Do you have mitigations in place for that?

packages/wrangler/src/r2/sql.ts
```diff
@@ -0,0 +1,163 @@
...
+		const s = spinner();
+		s.start("Query in progress");
+		const apiUrl = `https://api.sql.cloudflarestorage.com/api/v1/accounts/${accountId}/r2-sql/query/${bucketName}`;
+		let responseStatus = null;
+		let statusText = null;
+		let text = null;
+		let duration = null;
+		try {
+			const start = Date.now();
+			const response = await fetch(apiUrl, {
```
penalosa, line 114: This should be imported from `undici`

packages/wrangler/src/r2/sql.ts
penalosa: Fair enough

packages/wrangler/src/r2/sql.ts
laplab: We don't use API Gateway. Will tune undici timeout

PR #10561, opened 2025-09-05
Wrangler preview urls e2e fixes and warning follow up

danielrs:
Follow up of: https://github.com/cloudflare/workers-sdk/pull/10478

_Describe your change..._

- Fixes E2E tests.
- Only shows subdomain mismatch warning on redeploys, not first deploy.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Just test and log changes.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a feature in Wrangler V3.

### DIFF
--- a/packages/wrangler/e2e/deploy.test.ts
+++ b/packages/wrangler/e2e/deploy.test.ts
@@ -1,0 +1,130 @@
+import dedent from "ts-dedent";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { CLOUDFLARE_ACCOUNT_ID } from "./helpers/account-id";
+import { WranglerE2ETestHelper } from "./helpers/e2e-wrangler-test";
+import { generateResourceName } from "./helpers/generate-resource-name";
+import { normalizeOutput } from "./helpers/normalize";
+
+const TIMEOUT = 50_000;
+const workerName = generateResourceName();
+const normalize = (str: string) =>
+	normalizeOutput(str, {
+		[CLOUDFLARE_ACCOUNT_ID]: "CLOUDFLARE_ACCOUNT_ID",
+	}).replaceAll(/^Author:.*$/gm, "Author:      person@example.com");
+
+describe.skipIf(!CLOUDFLARE_ACCOUNT_ID)("deploy", { timeout: TIMEOUT }, () => {
+	let helper: WranglerE2ETestHelper;
+	beforeAll(async () => {
+		helper = new WranglerE2ETestHelper();
+	});
+
+	describe("subdomain warnings", () => {
+		beforeAll(async () => {
+			await helper.seed({
+				"wrangler.toml": dedent`
+                    name = "${workerName}"
+                    main = "src/index.ts"
+                    compatibility_date = "2023-01-01"
+                `,
+				"src/index.ts": dedent`
+                    export default {
+                        fetch(request) {
+                            return new Response("Hello World!")
+                        }
+                    }
+                `,
+				"package.json": dedent`
+                    {
+                        "name": "${workerName}",
+                        "version": "0.0.0",
+                        "private": true
+                    }
+                `,
+			});
+		});
+
+		afterAll(async () => {
+			await helper.run(`wrangler delete`);
+		});
+
+		it("omit subdomain warnings on 1st deploy", async () => {
+			const deploy = await helper.run("wrangler deploy");
+			expect(normalize(deploy.stdout)).toMatchInlineSnapshot(`
+				"Total Upload: xx KiB / gzip: xx KiB
+				Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
+				Deployed tmp-e2e-worker-00000000-0000-0000-0000-000000000000 triggers (TIMINGS)
+				  https://tmp-e2e-worker-00000000-0000-0000-0000-000000000000.SUBDOMAIN.workers.dev
+				Current Version ID: 00000000-0000-0000-0000-000000000000"
+			`);
+			expect(normalize(deploy.stderr)).toMatchInlineSnapshot(`""`);
+		});
+
+		it("show subdomain warnings on 2nd deploy, remote enabled", async () => {
+			// Set remote state using `wrangler triggers deploy`.
+			await helper.seed({
+				"wrangler.toml": dedent`
+                    name = "${workerName}"
+                    main = "src/index.ts"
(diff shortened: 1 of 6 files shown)


for-the-kidz (approved):

vicb:
> * Fixes E2E tests.

That sounds bad!

What the initial PR merged with failing E2E?

danielrs:
> > * Fixes E2E tests.
> 
> That sounds bad!
> 
> What the initial PR merged with failing E2E?

Seems like it was somehow skipped? Just fixed some snapshot lines.

packages/wrangler/e2e/versions.test.ts
```diff
@@ -70,7 +70,6 @@ describe.skipIf(!CLOUDFLARE_ACCOUNT_ID)(
 			Worker Startup Time: (TIMINGS)
 			Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
 			Worker Version ID: 00000000-0000-0000-0000-000000000000
-			Version Preview URL: https://tmp-e2e-worker-PREVIEW-URL.SUBDOMAIN.workers.dev
```
edmundhung, line 73: Is it fixing the e2e tests issue we discussed last week? If yes, I think the root clause is here: https://github.com/cloudflare/workers-sdk/pull/10547/files#diff-866c69b793109cd9eaa86031e90846e5806265a581fdba5f5eac984523d44c1eR40

#10547 is already merged to `main`.

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -348,7 +351,7 @@ async function subdomainDeploy(
 		);
 	}
 
-	if (config.preview_urls == undefined && !previewsInSync) {
+	if (!firstDeploy && config.preview_urls == undefined && !previewsInSync) {
```
edmundhung, line 354: Can we add an e2e test to confirm whether the warning is logged only after first deploy?

packages/wrangler/e2e/versions.test.ts
danielrs: This PR adds a new case, so we have e2e tests for `preview_urls` missing, and `preview_urls=true`.

packages/wrangler/src/triggers/deploy.ts
danielrs: Added!

packages/wrangler/src/triggers/deploy.ts
```diff
@@ -30,6 +30,7 @@ type Props = {
 	legacyEnv: boolean | undefined;
 	dryRun: boolean | undefined;
 	assetsOptions: AssetsOptions | undefined;
+	firstDeploy?: boolean;
```
petebacondarwin: There should be no need for this to be an optional parameter since it is internal and we have control over all the call sites.
It is safer to make it `boolean` or `boolean|undefined` and then make sure that all the call sites have been checked and updated accordingly.

packages/wrangler/e2e/deploy.test.ts
```diff
@@ -0,0 +1,130 @@
...
+				"wrangler.toml": dedent`
+                    name = "${workerName}"
+                    main = "src/index.ts"
+                    compatibility_date = "2023-01-01"
+                    workers_dev = false
+                    preview_urls = false
+                `,
+			});
+			await helper.run("wrangler triggers deploy");
+			// Run deploy.
```
petebacondarwin:
```suggestion
			// Run deploy.
			await helper.run("wrangler triggers deploy");
```

packages/wrangler/e2e/deploy.test.ts
```diff
@@ -0,0 +1,130 @@
...
+			await helper.seed({
```
petebacondarwin, line 109:
```suggestion
			// Now remove the `workers_dev` and `preview_urls` props and redeploy.
			await helper.seed({
```

packages/wrangler/e2e/deploy.test.ts
```diff
@@ -0,0 +1,130 @@
...
+				"Total Upload: xx KiB / gzip: xx KiB
+				Uploaded tmp-e2e-worker-00000000-0000-0000-0000-000000000000 (TIMINGS)
+				Deployed tmp-e2e-worker-00000000-0000-0000-0000-000000000000 triggers (TIMINGS)
+				  https://tmp-e2e-worker-00000000-0000-0000-0000-000000000000.SUBDOMAIN.workers.dev
+				Current Version ID: 00000000-0000-0000-0000-000000000000"
+			`);
+			expect(normalize(deploy.stderr)).toMatchInlineSnapshot(`
+				"▲ [WARNING] Worker has workers.dev disabled, but 'workers_dev' is not in the config.
+				  Using fallback value 'workers_dev = true'."
+			`);
```
petebacondarwin: I have to say that I find this warning message confusing.
Could we provide some more context as to why we are overriding what appears to be the current setting?

Should it be saying that the previous deployment had workers_dev disabled but since this config doesn't specify a value we are now going to default to enabling that?

packages/wrangler/e2e/deploy.test.ts
danielrs: Updated comment for clarity instead, `Run deploy` -> `Deploy Worker`. Deploying triggers is actually separate, and a requirement for this section.

packages/wrangler/e2e/deploy.test.ts
danielrs: Updated this for clearer messaging, e.g.
```
"▲ [WARNING] Worker has preview URLs enabled, but 'preview_urls' is not in the config.
	Using default config 'preview_urls = false', current status will be overwritten."
```

packages/wrangler/e2e/deploy.test.ts
```diff
@@ -0,0 +1,130 @@
...
+                        "name": "${workerName}",
+                        "version": "0.0.0",
+                        "private": true
+                    }
+                `,
+			});
+		});
+
+		afterAll(async () => {
+			await helper.run(`wrangler delete`);
```
dario-piotrowicz, line 47: ultra nit
```suggestion
			await helper.run("wrangler delete");
```

dario-piotrowicz (approved):
LGTM 🙂

petebacondarwin (approved):

PR #10570, opened 2025-09-07
[wrangler/d1] Support migrations in a directory (multi-file migrations)

Ehbraheem:
#### Summary

Allow a migration to be either:
- a top-level **.sql** file (unchanged), or
- a directory whose name is the migration name and which contains one or more **.sql** files (executed in deterministic order).

<br>

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal behavior change; README/docs follow-up can be added separately.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: change is contained to D1 migration logic and tests in this repo.

### DIFF
--- a/packages/wrangler/src/__tests__/d1/migrate.test.ts
+++ b/packages/wrangler/src/__tests__/d1/migrate.test.ts
@@ -1,3 +1,4 @@
+import fs from "node:fs";
 import { http, HttpResponse } from "msw";
 import { reinitialiseAuthTokens } from "../../user";
 import { mockAccountId, mockApiToken } from "../helpers/mock-account-id";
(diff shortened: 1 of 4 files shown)


Ehbraheem:
@penalosa, please help take a look at this PR when you're free. 

This change will help unblock dev work

packages/wrangler/src/__tests__/d1/migrate.test.ts
```diff
@@ -189,6 +190,202 @@ Your database may not be available to serve requests during the migration, conti
...
+				"CREATE TABLE top (id INTEGER PRIMARY KEY);"
+			);
+			fs.writeFileSync(
+				"migrations/002_complex/01_schema.sql",
+				"CREATE TABLE users (id INTEGER PRIMARY KEY);"
+			);
+			fs.writeFileSync(
+				"migrations/002_complex/02_seed.sql",
+				"INSERT INTO users (id) VALUES (1);"
+			);
```
alsuren, line 374: Hi @Ehbraheem.

Looking at this example, I think this would pattern would be better written as a flat list of files (potentially using a common prefix to visually group migrations that belong together). For example:

```
migrations/001_init.sql
migrations/002_complex_01_schema.sql
migrations/003_complex_02_seed.sql
```

Keeping things as a flat list of files makes things simpler and more consistent, and means that migrations will always be sorted correctly by all file managers + IDE + git forges.

Out of interest, what migration framework are you coming from that arranges migrations like this?

Ehbraheem: Hi @alsuren 

Short answer on frameworks that use per-migration directories: Prisma Migrate, Hasura migrations, and Sqitch (Liquibase is also commonly used in multi-file setups). We also run a small internal extension on top of Prisma/D1 that organizes a logical migration as a directory.

Your flat-file suggestion is sensible and works well for most cases (the current change continues to support that pattern). The reason directories are useful is when a single logical migration actually requires multiple coordinated artifacts (schema DDL, seeds, views, functions, helper queries, vendor SQL). Putting those files in one directory keeps them grouped, preserves local ordering, and makes the intent explicit; flattening works but quickly becomes noisy and error-prone as files proliferate.

Examples:

Flat (works, but verbose):
````text
migrations/002_complex_01_schema.sql
migrations/002_complex_02_views.sql
migrations/002_complex_03_seed.sql
migrations/002_complex_04_functions.sql
````

Directory (clear grouping, easier to reason about):
````text
migrations/002_complex/
  01_schema.sql
  02_views.sql
  03_seed.sql
  04_functions.sql
````

Compatibility note: this PR supports both patterns, single-file flat migrations remain unchanged, and directory-based multi-file migrations are optional. If you prefer, we can recommend a default (e.g., prefer flat files unless a migration needs multiple coordinated files) and add a short README note (numeric prefixes for ordering) to the docs.

Happy to add that doc note or change the recommendation if you prefer.

alsuren:
Sorry for leaving you hanging for so long.

After chatting internally, we have decided that we don't want to support this. Keeping things simple and uniform on our side means there is less to explain to beginners/llms. For users with more complex needs, it should be possible to maintain a hierarchical directory of migrations in git and write a script to flatten them into a temporary dir before invoking wrangler.

Ehbraheem:
> Sorry for leaving you hanging for so long.
> 
> After chatting internally, we have decided that we don't want to support this. Keeping things simple and uniform on our side means there is less to explain to beginners/llms. For users with more complex needs, it should be possible to maintain a hierarchical directory of migrations in git and write a script to flatten them into a temporary dir before invoking wrangler.

Hi @alsuren,

Closing this PR is premature and unjustified.

If you haven’t had time to review the implementation in detail, that’s fine, but closing it without even a minimal review or discussion is lazy, dismissive, and sends a bad signal. It implies that D1 isn’t serious about supporting real production workflows outside the narrow, Cloudflare-centric comfort zone. That kind of posture kills adoption and drives away contributors who are actually trying to make the platform better.

Per-migration directories aren’t some fringe idea; they’re standard practice in **Prisma Migrate, Hasura, Sqitch**, and countless production setups. The implementation here is completely backward-compatible. Nothing breaks, nothing changes for existing users. It just adds a sane, optional way to organize migrations with related files (DDL, seeds, views, functions, helper queries, vendor SQL).

Already addressed:
- Directories are treated as migrations only if they exist and contain .sql files.
- Files are executed deterministically (lexicographic order; numeric prefixes recommended).
- Single-file migrations remain the default.
- Directory support can easily be opt-in through config or CLI.

Closing this PR without a single technical objection or review sends the message that D1 isn’t open to evolving beyond its internal use cases. That’s a terrible signal for a platform that claims to be developer-focused.

I’m open to tightening scope, improving docs, or making directory support opt-in, but “close and ignore” isn’t a serious response.

lambrospetrou:
> If you haven’t had time to review the implementation in detail, that’s fine, but closing it without even a minimal review or discussion is lazy, dismissive, and sends a bad signal. 

@Ehbraheem  First of all, please avoid offensive language. 

There was already a discussion on the pull request changes, so you know it has been reviewed.

Any extensions we add is something that our team needs to maintain, even after you or whoever contributes them moves to different frameworks or platforms, so it needs to make sense for the overall product and not add every single thing.

In addition, according to https://hasura.io/docs/2.0/migrations-metadata-seeds/manage-migrations/ their structure does not work the way this PR is implemented. Hasura for example only has `<directory>/{up,down}.sql` files, and only one of them is executed.
Also, Prisma migrate does the same, only a single `<directory>/migration.sql` file is generated according to https://www.prisma.io/docs/orm/prisma-migrate/understanding-prisma-migrate/migration-histories.

Whereas, in this PR implementation all `.sql` files inside each directory are concatenated and executed as a single transaction (https://github.com/cloudflare/workers-sdk/pull/10570/files#diff-45e01a1e09bf07abd0415a7302b12340268b6268d505b7341d0e3aa619fbc180R141). This is not intuitive to see right off the bat, and it also has other UX issues since any error thrown from the D1 service (or SQLite itself if the statements are wrong) will refer to a single request whereas the users might have multiple files and they won't know where the error is coming from without wrangler having a way to tell them either. This makes the experience even worse.

Finally, this is a breaking change, since if someone has a directory of SQL files in their existing migrations directory those files are not executed, but after this change they will be and it can cause data corruption or loss potentially. So, any kind of this change needs to be behind a new CLI option.

We are going to think if there is a better way to achieve what you want, but as is the PR is not ready.

If we added generic support for executing SQL files inside sub-directories (recursively) all in alphabetical order, but each file would be its own transaction/migration, would that work for you? It's not what this PR is doing but maybe it works for you still.

Ehbraheem:
> > If you haven’t had time to review the implementation in detail, that’s fine, but closing it without even a minimal review or discussion is lazy, dismissive, and sends a bad signal.
> 
> @Ehbraheem First of all, please avoid offensive language.
> 
> There was already a discussion on the pull request changes, so you know it has been reviewed.
> 
> Any extensions we add is something that our team needs to maintain, even after you or whoever contributes them moves to different frameworks or platforms, so it needs to make sense for the overall product and not add every single thing.
> 
> In addition, according to https://hasura.io/docs/2.0/migrations-metadata-seeds/manage-migrations/ their structure does not work the way this PR is implemented. Hasura for example only has `<directory>/{up,down}.sql` files, and only one of them is executed. Also, Prisma migrate does the same, only a single `<directory>/migration.sql` file is generated according to https://www.prisma.io/docs/orm/prisma-migrate/understanding-prisma-migrate/migration-histories.
> 
> Whereas, in this PR implementation all `.sql` files inside each directory are concatenated and executed as a single transaction (https://github.com/cloudflare/workers-sdk/pull/10570/files#diff-45e01a1e09bf07abd0415a7302b12340268b6268d505b7341d0e3aa619fbc180R141). This is not intuitive to see right off the bat, and it also has other UX issues since any error thrown from the D1 service (or SQLite itself if the statements are wrong) will refer to a single request whereas the users might have multiple files and they won't know where the error is coming from without wrangler having a way to tell them either. This makes the experience even worse.
> 
> Finally, this is a breaking change, since if someone has a directory of SQL files in their existing migrations directory those files are not executed, but after this change they will be and it can cause data corruption or loss potentially. So, any kind of this change needs to be behind a new CLI option.
> 
> We are going to think if there is a better way to achieve what you want, but as is the PR is not ready.
> 
> If we added generic support for executing SQL files inside sub-directories (recursively) all in alphabetical order, but each file would be its own transaction/migration, would that work for you? It's not what this PR is doing but maybe it works for you still.

Thanks for taking the time to look into this, @lambrospetrou.

The reason for concatenating all .sql files in a directory is that file naming can’t be reliably predicted across different setups and tooling. Concatenation provides a best-effort, framework-agnostic way to ensure no valid SQL artifact is ignored; it’s not perfect, but it avoids locking the implementation into a single convention.

You’re right that Hasura handles things differently with paired up and down files, and that could justify adding a clear console warning to explain the execution behavior when a directory is detected.

I also agree with you on the potential maintenance overhead; that’s a valid concern.

Given we’ve established that composite directory migrations are a legitimate workflow, it seems reasonable to gate this behind a flag and include a clear warning about the concatenation behavior. That keeps it opt-in, transparent, and safe while still allowing advanced users to support more complex migration setups

lambrospetrou:
> Given we’ve established that composite directory migrations are a legitimate workflow

We haven't though. The concatenation approach is not what other frameworks do, and is very confusing to someone new as well.

If we will do a directory based approach it would be for each file to be its own transaction, and directories only used for visual/organizational grouping.

Would that work for you?

Ehbraheem:
> > Given we’ve established that composite directory migrations are a legitimate workflow
> 
> We haven't though. The concatenation approach is not what other frameworks do, and is very confusing to someone new as well.
> 
> If we will do a directory based approach it would be for each file to be its own transaction, and directories only used for visual/organizational grouping.
> 
> Would that work for you?

Yes, that works for me.

PR #10571, opened 2025-09-07
add missing type for `send_email`'s `experimental_remote` field

dario-piotrowicz:
I noticed that we're missing the `experimental_remote` field in the `send_email` wrangler config:
Screenshot 2025-09-07 at 22 45 18

Although as you can see from e2e tests this is an expected field:
https://github.com/cloudflare/workers-sdk/blob/dceb550cd817b792b7f7ec04df2b8f1f5147527d/packages/wrangler/e2e/remote-binding/wrangler-remote-resources.test.ts#L395-L400

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: the functionality using the field is already tested and we don't have type-specific tests
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


packages/wrangler/src/config/environment.ts
```diff
@@ -692,6 +692,8 @@ export interface EnvironmentNonInheritable {
 		destination_address?: string;
 		/** If this binding should be restricted to a set of verified addresses */
 		allowed_destination_addresses?: string[];
+		/** Whether the binding should be remote or not (only available under `--x-remote-bindings`) */
+		experimental_remote?: boolean;
```
edmundhung, line 696: I am curious why this doesn't get caught by the type checker. Didn't we use this config to infer the miniflare setup?

dario-piotrowicz: We check the `remote` field here:
Screenshot 2025-09-08 at 15 11 38

But as you can see the type is just `object` so there isn't a strong type checking here

After that we do filter the bindings that need to be remote, but we check if `"experimental_remote" in binding`:
Screenshot 2025-09-08 at 15 14 33

Which again would not fail if `experimental_remote` wasn't part of the type

So we simply miss the fact that the field wasn't defined in the type 🤔

PR #10574, opened 2025-09-08
Add support for WVPC connectivity services CRUD operations

efalcao:
```
wrangler vpc service

🔗 Manage VPC connectivity services

COMMANDS
  wrangler vpc service create <name>        Create a new VPC connectivity service
  wrangler vpc service delete <service-id>  Delete a VPC connectivity service
  wrangler vpc service get <service-id>     Get a VPC connectivity service
  wrangler vpc service list                 List VPC connectivity services
  wrangler vpc service update <service-id>  Update a VPC connectivity service
```

Implements WVPC-61

Workers VPC service creation that is similar to creating Hyperdrive configs.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/25308
  - [ ] Documentation not necessary because: will be followup
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a v3 feature

### DIFF
--- a/.changeset/orange-swans-talk.md
+++ b/.changeset/orange-swans-talk.md
@@ -2,4 +2,4 @@
 "wrangler": minor
 ---
 
-Add support for WVPC services CRUD via `wrangler wvpc service`
+Add support for VPC services CRUD via `wrangler vpc service`


packages/wrangler/src/index.ts
```diff
@@ -1373,6 +1379,33 @@ export function createCLIParser(argv: string[]) {
 	]);
 	registry.registerNamespace("pipelines");
 
+	// wvpc
```
penalosa:
```suggestion
```

packages/wrangler/src/index.ts
```diff
@@ -1373,6 +1379,33 @@ export function createCLIParser(argv: string[]) {
...
+	registry.define([
+		{ command: "wrangler wvpc", definition: wvpcNamespace },
+		{ command: "wrangler wvpc service", definition: wvpcServiceNamespace },
```
penalosa: Why is everything under `wvpc service`? Do you expect something else under `wvpc` in the future?

packages/wrangler/src/wvpc/update.ts
```diff
@@ -0,0 +1,85 @@
...
+		"service-id": {
+			type: "string",
+			demandOption: true,
+			description: "The ID of the connectivity service to update",
+		},
+		...serviceOptions(),
+	},
+	positionalArgs: ["service-id"],
+	async handler(args, { config }) {
+		logger.log(wvpcBetaWarning);
```
penalosa: Could you remove any beta warnings like this? Adding `status: "private-beta"` to the command definition (which you have) should be sufficient

packages/wrangler/src/wvpc/list.ts
```diff
@@ -0,0 +1,33 @@
...
+	metadata: {
+		description: "List WVPC connectivity services",
+		status: "private-beta",
+		owner: "Product: WVPC",
+	},
+	args: {
+		"service-type": {
+			type: "string",
+			choices: ["tcp", "http"],
+			description: "Filter by service type (tcp or http)",
```
penalosa:
```suggestion
			description: "Filter by service type",
```

packages/wrangler/src/wvpc/list.ts
```diff
@@ -0,0 +1,33 @@
...
+		},
+	},
+	async handler(args, { config }) {
+		logger.log(wvpcBetaWarning);
```
penalosa:
```suggestion
```

packages/wrangler/src/wvpc/index.ts
```diff
@@ -0,0 +1,135 @@
...
+	http_port?: number;
+	https_port?: number;
+	host: ServiceHost;
+}
+
+export interface ConnectivityServiceListParams {
+	service_type?: ServiceType;
+}
+
+export const serviceOptions = () =>
```
penalosa: Why is this a function rather than just an object? Additionally, it seems like ther are groups of options that need to be provided together—could you use the `conflicts` and `group` arguments? e.g. https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/src/hyperdrive/index.ts#L93-L100

packages/wrangler/src/wvpc/get.ts
```diff
@@ -0,0 +1,71 @@
...
+	},
+	args: {
+		"service-id": {
+			type: "string",
+			demandOption: true,
+			description: "The ID of the connectivity service",
+		},
+	},
+	positionalArgs: ["service-id"],
+	async handler(args, { config }) {
```
penalosa: Is this service documented in the SDK yet? If so, you have access to an SDK instance in the second parameter here, which should be used instead of fetchResult where possible

packages/wrangler/src/wvpc/create.ts
```diff
@@ -0,0 +1,80 @@
...
+	args: {
+		...serviceOptions(),
+	},
+	positionalArgs: ["name"],
+	async handler(args, { config }) {
+		logger.log(wvpcBetaWarning);
+		logger.log(`🚧 Creating WVPC connectivity service '${args.name}'`);
+
+		// Validate arguments and build request
+		const request = validateAndBuildRequest({
```
penalosa: In general, validation should happen in `validateArgs()` rather than `handler()`

packages/wrangler/src/index.ts
efalcao: yes we expect lots of further development here. `network`, `host`, `proxy` are coming

PR #10575, opened 2025-09-08
fix: ensure that pinned versions are not converted to ranges when using npm

petebacondarwin:
- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  create-cloudflare

### DIFF
--- a/packages/create-cloudflare/src/helpers/__tests__/packages.test.ts
+++ b/packages/create-cloudflare/src/helpers/__tests__/packages.test.ts
@@ -1,15 +1,23 @@
-import { existsSync } from "fs";
+import { existsSync } from "node:fs";
+import { resolve } from "node:path";
 import { runCommand } from "helpers/command";
 import { installPackages, installWrangler, npmInstall } from "helpers/packages";
 import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 import whichPMRuns from "which-pm-runs";
 import { createTestContext } from "../../__tests__/helpers";
+import * as files from "../files";
 import { mockPackageManager } from "./mocks";
 
 vi.mock("fs");
 vi.mock("which-pm-runs");
 vi.mock("which-pm-runs");
 vi.mock("helpers/command");
+vi.mock("../files", () => ({
+	readJSON: vi.fn(),
+	writeJSON: vi.fn(),
+}));
+const mockReadJSON = vi.mocked(files.readJSON);
+const mockWriteJSON = vi.mocked(files.writeJSON);
 
 describe("Package Helpers", () => {
 	beforeEach(() => {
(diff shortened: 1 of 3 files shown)


packages/create-cloudflare/src/helpers/packages.ts
```diff
@@ -61,6 +67,27 @@ export const installPackages = async (
 			silent: true,
 		},
 	);
+
+	if (npm === "npm") {
```
penalosa, line 71: Won't this not update `package-lock.json`?

petebacondarwin:
~~@penalosa - regarding the question about the lockfile... I think it might be that after the package.json is updated we might need to run `npm i` again to ensure the lockfile is completely in sync. But it definitely wouldn't be broken.~~

Actually just checked and the npm package-lock.json doesn't appear to include the `^` for these pinned deps, so all good.

packages/create-cloudflare/src/helpers/packages.ts
```diff
@@ -61,6 +67,27 @@ export const installPackages = async (
...
+		// Npm install will update the package.json with a hat-range rather than the exact version/range we asked for.
```
vicb, line 72: nit: should hat-range be caret range (I don't remember ever encountering the former)

```suggestion
		// `npm install` will update the `package.json` with a caret range rather than the exact version/range we asked for.
```

petebacondarwin: Ah yes. I always call `^` a hat but probably not that well used.

packages/create-cloudflare/src/helpers/__tests__/packages.test.ts
```diff
@@ -87,18 +121,47 @@ describe("Package Helpers", () => {
...
+				if (pm === "npm") {
+					// Check that package.json was updated for npm
+					expect(mockReadJSON).toHaveBeenCalledWith(
+						resolve(process.cwd(), "package.json"),
+					);
+					expect(mockWriteJSON).toHaveBeenCalledWith(
+						resolve(process.cwd(), "package.json"),
+						expect.objectContaining({
+							["devDependencies"]: {
+								foo: "^1.0.0",
```
vicb, line 148: wait don't we want to have exact version here?

petebacondarwin: I believe it is normal for all package managers to create a "caret" range for deps that have no version specified...

vicb: OOps sorry, I was looking at the wrong line - l 131 has a fixed version for `baz` and it's on l 150.

Looks good!

petebacondarwin: E.g. `pnpm add -D foo` results in

```
"foo": "^0.0.7",
```

being added to package.json

packages/create-cloudflare/src/helpers/packages.ts
vicb: Also maybe add a comment to explain why `--save-exact` can not be used.

petebacondarwin: https://github.com/cloudflare/workers-sdk/pull/10576

PR #10577, opened 2025-09-08
Use workerd `node:process` v2 when available

vicb:
This PR enabled using workerd `node:process` v2 when it is enabled (by flag or date).

It depends on the implemntation being available:
- implemented in https://github.com/cloudflare/workerd/pull/4960
- releases in [workerd v1.20250904.0](https://github.com/cloudflare/workerd/releases/tag/v1.20250904.0)
- depends on https://github.com/cloudflare/workerd/pull/5040  released in https://github.com/cloudflare/workerd/releases/tag/v1.20250911.0
- depends on https://github.com/cloudflare/workerd/pull/5024 released in https://github.com/cloudflare/workerd/releases/tag/v1.20250912.0

---

Outdated comment:

<details>

A few notes:

This PR still fails the remote tests as the v2 changes have not reached production yet.

Some of [the v2 APIs are `undefined`](https://github.com/cloudflare/workerd/blob/e8aad2a03def0f5a18344c0e96b17a6dfafefdee/src/node/internal/public_process.ts#L204-L227).

For that reason we still need an hybrid polyfill pulling their implementation from `unenv`. Otherwise the following code might break:

```ts
import { process } from "node:process";

// This would break with v2 as `process.release` is `undefined.
console.log(process.release.name);
```

Because we have to use an hybrid polyfill anyway, we keep all the APIs that are implemented in `unenv` and missing from `workerd`:
- `_...` APIs (`_channel`, `_debugEnd`, ...)
- some other APIs as `addListener`, `assert`, ...

`addListener`, `assert`, ... are not all documented on nodejs.org and a few are deprecated. But they all are available for Node 22.16.0 (tested in the REPL).

Maybe we can refine and better handle deprecated APIs now that we have a dedicated workerd flag. Comments welcome.

@petebacondarwin @guybedford could you please review the PR, the API surface is quite large and having more eyes would be helpful.

<details>

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: will be documented by the runtime team
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: unenv changes are not backported

### DIFF
--- a/packages/unenv-preset/src/runtime/node/process.ts
+++ b/packages/unenv-preset/src/runtime/node/process.ts
@@ -38,7 +38,7 @@ const unenvProcess = new UnenvProcess({
 // Note that `env`, `hrtime` and `nextTick` are always retrieved from `unenv`
 export const { exit, features, platform } = workerdProcess;
 
-// APIs that can be implemented by either by `unenv` or `workerd`.
+// APIs that can be implemented by either `unenv` or `workerd`.
 // They are always retrieved from `unenv` which might use their `workerd` implementation.
 export const {
 	// Always implemented by workerd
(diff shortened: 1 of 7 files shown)


guybedford:
I believe that in most of the cases where unenv is providing an implementation of a function like `process.binding()` and where we are providing `undefined` - the implementation is simply a throwing stub. Simply copying over the throwing stub should be just a couple of lines and might get far to enabling a non-hybrid approach.

For undocumented APIs, I notice `_debugEnd` is implemented by Bun but not Deno. Do we definitely need these?

For non-function bindings, like `process.release` - having an upstream issue to track their support would be a huge help. Happy to fill gaps as necessary as well.

vicb:
> I believe that in most of the cases where unenv is providing an implementation of a function like `process.binding()` and where we are providing `undefined` - the implementation is simply a throwing stub. Simply copying over the throwing stub should be just a couple of lines and might get far to enabling a non-hybrid approach.

When you say "Simply copying over the throwing stub [...]" do you mean copying the implementation to workerd?

If you rather mean copying the implementation to the [unev preset in this repo](https://github.com/cloudflare/workers-sdk/blob/main/packages/unenv-preset/src/runtime/node/process.ts) then that would still be an hybrid polyfill and I don't really see the point?

> For undocumented APIs, I notice `_debugEnd` is implemented by Bun but not Deno. Do we definitely need these?

I agree that the likeness those `_...` APIs are used is low but we have no way to definitely rule out that they are not.
And because we need an hybrid polyfill anyway there should be no good reason to remove them?
 
> For non-function bindings, like `process.release` - having an upstream issue to track their support would be a huge help. Happy to fill gaps as necessary as well.

Sorry, I'm not exactly sure where you'd want to create the issue and what the content should be. Feel free to start someone and I'll contribute if needed.

guybedford:
Thanks again for working on this. Happy to review further. We just landed comprehensive process stubs in https://github.com/cloudflare/workerd/pull/5040 which may help the transition here when that is available.

vicb:
I did sync after the update of workerd in wrangler.

@petebacondarwin and @guybedford could you please review this PR?

petebacondarwin (approved):

packages/unenv-preset/src/runtime/node/process.ts
```diff
@@ -14,121 +14,171 @@ const globalProcess: NodeJS.Process = (globalThis as any)["pro" + "cess"];
...
+	// @ts-expect-error `domain` is missing typings
+	_preload_modules,
+	// @ts-expect-error `domain` is missing typings
+	_rawDebug,
+	// @ts-expect-error `domain` is missing typings
+	_startProfilerIdleNotifier,
+	// @ts-expect-error `domain` is missing typings
+	_stopProfilerIdleNotifier,
+	// @ts-expect-error `domain` is missing typings
+	_tickCallback,
```
guybedford: These are now all implemented in https://github.com/cloudflare/workerd/pull/5040.

packages/unenv-preset/src/runtime/node/process.ts
```diff
@@ -14,121 +14,171 @@ const globalProcess: NodeJS.Process = (globalThis as any)["pro" + "cess"];
...
-	setUncaughtExceptionCaptureCallback,
-	loadEnvFile,
-	sourceMapsEnabled,
 	arch,
 	argv,
 	argv0,
+	availableMemory,
+	// @ts-expect-error `domain` is missing typings
+	binding,
+	channel,
```
guybedford: These are also now added in https://github.com/cloudflare/workerd/pull/5040.

packages/unenv-preset/src/runtime/node/process.ts
```diff
@@ -14,121 +14,171 @@ const globalProcess: NodeJS.Process = (globalThis as any)["pro" + "cess"];
...
+	_events,
+	_eventsCount,
+	_handleQueue,
+	_maxListeners,
+	_pendingMessage,
+	_send,
+	assert,
+	disconnect,
+	mainModule,
+} = unenvProcess;
```
guybedford, line 66: I think it would be fine to leave these ones out at this point, or have a path to test that. In Node.js,

```
import * as process from 'node:process';
console.log('process._channel', '_channel' in process, process._channel);
// ...
```

Gives:

```
process._channel false undefined
process._disconnect false undefined
process._events true [Object: null prototype] {
  newListener: [ [Function (anonymous)], [Function: startListeningIfSignal] ],
  removeListener: [ [Function (anonymous)], [Function: stopListeningIfSignal] ],
  warning: [Function: onWarning],
  SIGWINCH: [
    [Function: refreshStdoutOnSigWinch],
    [Function: refreshStderrOnSigWinch]
  ]
}
process._eventsCount true 3
process._handleQueue false undefined
process._maxListeners true undefined
process._pendingMessage false undefined
process._send false undefined
process.assert false undefined
process.disconnect false undefined
process.mainModule false undefined
```

Maintaining compat that is _just not Node.js_ doesn't make sense to me.

Where `_eventsCount`, `_maxListeners` and `_events` are thus the only ones worth even considering keeping here.

packages/wrangler/e2e/unenv-preset/preset.test.ts
```diff
@@ -188,6 +188,34 @@ const testConfigs: TestConfig[] = [
...
+	// node:process v2
+	[
+		{
+			name: "process v1  by date",
+			compatibilityDate: "2024-09-23",
+			expectRuntimeFlags: {
+				enable_nodejs_process_v2: false,
+			},
+		},
+		// TODO: add a config when v2 is enabled by default (>= 2025-09-15)
```
guybedford, line 200: Does this need to be done now, or has this already been implemented?

packages/unenv-preset/src/runtime/node/process.ts
vicb: > In Node.js

In what version of Node.js ?

At least we know that they were there before so let's merge this now and cleanup later.

WDYT?

packages/unenv-preset/src/runtime/node/process.ts
```diff
@@ -14,121 +14,171 @@ const globalProcess: NodeJS.Process = (globalThis as any)["pro" + "cess"];
...
+
+// API that are only implemented starting from v2 of workerd process
+// They are retrieved from unenv when process v1 is used
+export const {
+	// @ts-expect-error `domain` is missing typings
```
vicb: `domain` should be updated here and below

vicb:
> Left an initial review. I'll look into an upstream follow-up for `_eventsCount`, `_maxListeners` and `_events` as well.

I think it's ok to address those in a follow up PR because this should be merged soonish (merging in workerd, doing a release, updating workerd in this repo will take 2/3 days if everything goes well).

Any objection?

packages/wrangler/e2e/unenv-preset/preset.test.ts
vicb: This can only be done when worked >= 2025-09-15 both in this repo and in prod.

packages/unenv-preset/src/runtime/node/process.ts
vicb: See the comment line 64 and line 181

```ts
// API that are only implemented starting from v2 of workerd process
// They are retrieved from unenv when process v1 is used
export const {
[...]
} = isWorkerdProcessV2 ? workerdProcess : unenvProcess;
```

I think the code is right. Am I missing something?

packages/unenv-preset/src/runtime/node/process.ts
vicb: Same comment as on line 106

guybedford (approved):
Sure, absolutely makes sense to roll this one out sooner rather than later.

vicb:
Rebased + added a fixup commit for comments only

PR #10582, opened 2025-09-09
feat: `containers ssh` command

flakey5:
Closes CC-4634

Adds a `containers ssh` command + `ssh` and `authorized_keys` properties to a container's configuration. This allows for ssh'ing into a container through wrangler. 

- [ ] Currently not working since it relies on pending api gateway PRs

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/containers-shared/src/client/models/SSHPublicKeyItemV3.ts
```diff
@@ -0,0 +1,13 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
```
vicb, line 4: Is this really needed?
If not, can this be removed from all the added files?

packages/wrangler/src/__tests__/containers/deploy.test.ts
```diff
@@ -1635,6 +1635,109 @@ describe("wrangler deploy with containers", () => {
...
+
+			Uploaded test-name (TIMINGS)
+			Deployed test-name triggers (TIMINGS)
+			  https://test-name.test-sub-domain.workers.dev
+			Current Version ID: Galaxy-Class"
+		`);
+		// no deprecation warnings should show up on this run
+		expect(std.warn).toMatchInlineSnapshot(`""`);
+		expect(std.err).toMatchInlineSnapshot(`""`);
+
```
vicb: nit: there should be no need for using a snapshot here and using `toBe` will make sure we do not inadvertently change the empty string to something else when updating the snapshots

```suggestion
		expect(std.warn).toBe("");
		expect(std.err).toBe("");

```

vicb: Note there are other ocurences of `.toMatchInlineSnapshot(`""`)` in this PR, can you please update all of them?

packages/wrangler/src/containers/containers.ts
```diff
@@ -147,6 +156,147 @@ export async function listCommand(
...
+export function sshYargs(args: CommonYargsArgv) {
+	return (
+		args
+			.positional("ID", {
+				describe: "id of the container instance",
+				type: "string",
+				demandOption: true,
+			})
+			// Following are SSH flags that should be directly passed in
+			.option("c", {
```
vicb: Wrangler has `--config` / `-c` and `--env` / `-e`, can there be a conflict here?

.changeset/spotty-plums-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+"@cloudflare/containers-shared": minor
+"wrangler": minor
```
emily-shen:
```suggestion
"wrangler": patch
```
feats on experimental stuff is a patch here usually

packages/containers-shared/src/client/models/SSHPublicKeyItemV3.ts
emily-shen: this is all auto-generated code that's being copy-pasted over, we should remove these ignores but i can do that in a separate PR - its not really on @flakey5  :')

packages/wrangler/src/containers/containers.ts
```diff
@@ -147,6 +156,147 @@ export async function listCommand(
...
+		}
+
+		if (err instanceof ApiError) {
+			if (err.status === 400 || err.status === 404) {
+				throw new UserError(
+					`There has been an error when trying to SSH into the container.\n${err.body.error}`
+				);
+			}
+
+			throw new Error(
```
emily-shen:
```suggestion
			throw new APIError(
```
that's wrangler's APIError, with caps - means it'll get reported to sentry, vs usererror which doesn't.

packages/workers-utils/src/config/environment.ts
```diff
@@ -123,6 +123,23 @@ export type ContainerApp = {
...
+		enabled: boolean;
+		/**
+		 * Port that the SSH service is running on
+		 * @defaults to 22
+		 */
+		port?: number;
+	};
+
+	/**
+	 * SSH public keys to put in the container's authorized_keys file.
```
emily-shen, line 189: any reason not to nest this under `ssh` as well?

packages/wrangler/src/config/validation.ts
```diff
@@ -2725,6 +2727,43 @@ function validateContainerApp(
...
+			}
+
+			if ("authorized_keys" in containerAppOptional) {
+				if (!Array.isArray(containerAppOptional.authorized_keys)) {
+					diagnostics.errors.push(`${field}.authorized_keys must be an array`);
+				} else {
+					for (
+						let i = 0;
+						i < containerAppOptional.authorized_keys.length;
+						i++
```
emily-shen:
```suggestion
						const item of containerAppOptional.authorized_keys
```
nit

packages/workers-utils/src/config/environment.ts
```diff
@@ -123,6 +123,23 @@ export type ContainerApp = {
 				disk_mb?: number;
 		  };
 
+	ssh?: {
+		/**
+		 * If enabled, those with write access to a container will be able to SSH into it through Wrangler.
+		 */
+		enabled: boolean;
```
emily-shen, line 180: can you add the default to the comment?

packages/wrangler/src/__tests__/containers/ssh.test.ts
```diff
@@ -0,0 +1,115 @@
...
+	// against, but everything up until that point is covered.
+	it("should try ssh'ing into a container", async () => {
+		const instanceId = "a".repeat(64);
+		const wsUrl = "ws://localhost:1234";
+		const sshJwt = "asd";
+
+		setWranglerConfig({});
+		msw.use(
+			http.get(`*/instances/:instanceId/ssh`, async ({ request }) => {
+				expect(request.url.endsWith(`${instanceId}/ssh`)).toBeTruthy();
```
emily-shen:
```suggestion

```

won't hit the mock if it doesn't fulfil that, afaik

packages/wrangler/src/__tests__/containers/ssh.test.ts
```diff
@@ -0,0 +1,115 @@
...
+	it("should reject invalid container ID format", async () => {
+		setWranglerConfig({});
+		await expect(
+			runWrangler("containers ssh invalid-id")
+		).rejects.toMatchInlineSnapshot(
+			`[Error: Expected an instance ID but got invalid-id]`
+		);
+	});
+
+	it("should handle 500s when getting ssh jwt", async () => {
```
emily-shen, line 62: i don't think this test does anything currently 😅

packages/wrangler/src/containers/deploy.ts
emily-shen, line 1: it would really nice if there was a way we can verify ssh is configured correctly at deploy time

packages/wrangler/src/containers/containers.ts
emily-shen: womp yeah wrangler's flags conflict with the actual ssh flags, and i don't think we could not include the global flags here because it requires auth and we might need to read an account id from a config file :(

packages/wrangler/src/containers/containers.ts
```diff
@@ -147,6 +156,147 @@ export async function listCommand(
...
+
+		child.on("error", reject);
+
+		child.on("exit", (code) => {
+			// Ssh errors exit with code 255
+			if (code !== 255) {
+				resolve(undefined);
+			} else {
+				reject(
+					new Error(`ssh exited unsuccessfully. Is the container running?`)
```
emily-shen: what should the user do if the container isn't running?

packages/wrangler/src/containers/containers.ts
```diff
@@ -147,6 +156,147 @@ export async function listCommand(
...
+	return (
+		args
+			.positional("ID", {
+				describe: "id of the container instance",
+				type: "string",
+				demandOption: true,
+			})
+			// Following are SSH flags that should be directly passed in
+			.option("c", {
+				describe: "cipher_spec",
```
emily-shen:
```suggestion
				describe: "ssh option for  cipher_spec",
```

and for the rest too

packages/wrangler/src/__tests__/containers/ssh.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+			SSH into a container
+
+			POSITIONALS
+			  ID  id of the container instance  [string]
+
+			GLOBAL FLAGS
+			  -c, --config    cipher_spec  [string]
+			      --cwd       Run as if Wrangler was started in the specified directory instead of the current working directory  [string]
+			  -e, --env       escape_char  [string]
```
emily-shen: womp yargs has consolidated this in a weird way

packages/wrangler/src/__tests__/containers/deploy.test.ts
```diff
@@ -1635,6 +1635,109 @@ describe("wrangler deploy with containers", () => {
 			`);
 		});
 	});
+
+	it("should enable ssh when provided for new container", async () => {
```
emily-shen, line 1783: add a test for modify? or add ssh to one of the existing modify tests maybe

Talador12:
This PR is a requirement for SSH for containers
https://github.com/cloudflare/containers-demos/pull/17
https://github.com/cloudflare/containers/pull/86

After this is merged, we will update this PR

packages/workers-utils/src/config/environment.ts
flakey5: `authorized_keys` isn't solely tied to wrangler ssh, it's pretty much just the ssh key logic for v2 containers but meant for v3 containers

packages/wrangler/src/containers/containers.ts
flakey5: They should start it themselves, we want to avoid starting it for them here since we don't actually know what the container is doing (ex/ containers that create backups, cleanup tasks, ...)

packages/workers-utils/src/config/environment.ts
emily-shen: so is `ssh` solely applicable to wrangler ssh? should we rename it `wrangler_ssh` like it is in the api?

packages/wrangler/src/config/validation.ts
```diff
@@ -2725,6 +2727,84 @@ function validateContainerApp(
...
+							diagnostics.errors.push(`${fieldPath}.name must be a string`);
+						}
+
+						if (!isRequiredProperty(key, "public_key", "string")) {
+							diagnostics.errors.push(
+								`${fieldPath}.public_key must be a string`
+							);
+						}
+
+						if (key.public_key.toLowerCase().startsWith("ssh-ed25519")) {
```
flakey5: Not sure if we wanna have this check here, it's not unlikely that we end up supporting more key types and this would make it so you'd need to update wrangler in order to use them

packages/wrangler/src/config/validation.ts
```diff
@@ -2746,6 +2748,84 @@ function validateContainerApp(
...
+						`${field}.authorized_keys must be provided if wrangler ssh is enabled`
+					);
+				}
+			}
+
+			if ("authorized_keys" in containerAppOptional) {
+				if (!Array.isArray(containerAppOptional.authorized_keys)) {
+					diagnostics.errors.push(`${field}.authorized_keys must be an array`);
+				} else {
+					for (const item in containerAppOptional.authorized_keys) {
```
```suggestion
					for (const index in containerAppOptional.authorized_keys) {
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2746,6 +2748,84 @@ function validateContainerApp(
...
+						const fieldPath = `${field}.authorized_keys[${item}]`;
+						const key = containerAppOptional.authorized_keys[item];
+
+						validateAdditionalProperties(
+							diagnostics,
+							fieldPath,
+							Object.keys(key),
+							["name", "public_key"]
+						);
```
nikitassharma: this could be an error instead of just a warning?

packages/wrangler/src/containers/config.ts
```diff
@@ -118,6 +118,13 @@ export const getNormalizedContainerOptions = async (
 					config.observability?.logs?.enabled ??
 					config.observability?.enabled === true,
 			},
+			wrangler_ssh: container.wrangler_ssh
+				? {
+						enabled: container.wrangler_ssh.enabled,
+						port: container.wrangler_ssh.port,
+					}
+				: undefined,
```
nikitassharma:
```suggestion
			wrangler_ssh: container.wrangler_ssh,
```

packages/wrangler/src/containers/containers.ts
```diff
@@ -147,6 +157,147 @@ export async function listCommand(
 	await listCommandHandle(listArgs, config);
 }
 
+export function sshYargs(args: CommonYargsArgv) {
```
nikitassharma: The `SSH option` prefix for the descriptions might  be necessary since they're all options for SSH

Also wondering if it might be a bit confusing to have the ssh aliases (i.e. `(-c)`) included in the description. Maybe instead having the descriptions pulled from `ssh`?

packages/wrangler/src/containers/containers.ts
```diff
@@ -147,6 +157,147 @@ export async function listCommand(
...
+				type: "string",
+				demandOption: true,
+			})
+			// Following are SSH flags that should be directly passed in
+			.option("cipher", {
+				describe: "SSH option for cipher_spec (-c)",
+				type: "string",
+			})
+			.option("log-file", {
+				describe: "SSH option for log_file (-c)",
```
nikitassharma:
```suggestion
				describe: "SSH option for log_file (-E)",
```

packages/wrangler/src/containers/deploy.ts
```diff
@@ -180,6 +180,13 @@ function containerConfigToCreateRequest(
 				containerApp.observability.logs_enabled,
 				prevApp?.configuration.observability
 			),
+			wrangler_ssh: containerApp.wrangler_ssh
+				? {
+						enabled: containerApp.wrangler_ssh.enabled,
+						port: containerApp.wrangler_ssh.port,
+					}
+				: undefined,
```
nikitassharma:
```suggestion
			wrangler_ssh: containerApp.wrangler_ssh,
```

packages/wrangler/src/config/validation.ts
```diff
@@ -2746,6 +2748,84 @@ function validateContainerApp(
...
+				}
+			}
+
+			if ("authorized_keys" in containerAppOptional) {
+				if (!Array.isArray(containerAppOptional.authorized_keys)) {
+					diagnostics.errors.push(`${field}.authorized_keys must be an array`);
+				} else {
+					for (const item in containerAppOptional.authorized_keys) {
+						const fieldPath = `${field}.authorized_keys[${item}]`;
+						const key = containerAppOptional.authorized_keys[item];
```
flakey5:
```suggestion
						const fieldPath = `${field}.authorized_keys[${index}]`;
						const key = containerAppOptional.authorized_keys[index];
```

packages/wrangler/src/containers/containers.ts
flakey5: Ssh doesn't really have great descriptions for the flags other than what's here already:
```
usage: ssh [-46AaCfGgKkMNnqsTtVvXxYy] [-B bind_interface] [-b bind_address]
           [-c cipher_spec] [-D [bind_address:]port] [-E log_file]
           [-e escape_char] [-F configfile] [-I pkcs11] [-i identity_file]
           [-J destination] [-L address] [-l login_name] [-m mac_spec]
           [-O ctl_cmd] [-o option] [-P tag] [-p port] [-R address]
           [-S ctl_path] [-W host:port] [-w local_tun[:remote_tun]]
           destination [command [argument ...]]
       ssh [-Q query_option]
```

Still might be able to write up better ones?

nikitassharma: Oh I meant the descriptions from the man page:
```
       -c cipher_spec
               Selects the cipher specification for encrypting the session.  cipher_spec is a comma-separated list of ciphers listed in order  of  prefer‐
               ence.  See the Ciphers keyword in ssh_config(5) for more information.
```

So for `cipher_spec`, for example, either the first 1 or 2 sentences?

packages/containers-shared/src/client/models/SSHPublicKeyItemV3.ts
vicb: I don't think we should land such code in this repo.
Ok to accept this once if there is a cleanup afterwards but going forward, please make sure to prune such comments before submitting a PR.
Thanks!

vicb:
What is the status here?
Were the force push only rebases (it would help if a comment is added)?

flakey5:
> What is the status here?

I don't believe I have write access to the branch anymore, but this feature shouldn't need any more changes on the container's side of things so this pr should be the only thing left to get this feature added in iirc

emily-shen:
@flakey5 did you manage to test this e2e before you left? i'm having trouble getting it to work with various `host key verification failed` and `connection closed` - @gabivlj is going to debug.  

also @vicb - for context @flakey5's internship finished, so I can take over making sure the PR is ready to be merged (that's why I rebased) - there shouldn't be any new changes needed.

flakey5:
> did you manage to test this e2e before you left?

I was able to get it working, `connection closed` might just be from the container exiting? The `host key verification failed` is interesting though, what I'm assuming is that you got the same port for the tcp proxy 2+ times but for different instances. So the host keys for the subsequent requests were different from the first time you ssh'd, but the key from the first instance was still saved in `~/.ssh/known_hosts`

emily-shen:
the plan is to merge this mostly as is but hidden as experimental (which i have done here [a29c077](https://github.com/cloudflare/workers-sdk/pull/10582/commits/a29c0771f799211c212ff9ede1955395efce79b3)).

packages/wrangler/src/containers/ssh.ts
```diff
@@ -0,0 +1,270 @@
...
+				type: "string",
+			})
+			.option("escape-char", {
+				describe:
+					"SSH option: Sets the escape character for sessions with a pty (default: ‘~’)",
+				type: "string",
+			})
+			.option("config-file", {
+				describe:
+					"SSH option: Specifies an alternative per-user configuration file",
```
jamesopstad: Is this a path to a configuration file? If so, that should be in the description.

packages/wrangler/src/containers/ssh.ts
```diff
@@ -0,0 +1,270 @@
...
+				type: "string",
+			})
+			.option("mac-spec", {
+				describe:
+					"SSH option: A comma-separated list of MAC (message authentication code) algorithms, specified in order of preference",
+				type: "string",
+			})
+			.option("option", {
+				describe:
+					"SSH option: Can be used to give options in the format used in the configuration file",
```
jamesopstad: This seems quite vague. What configuration file does this refer to and where is the format documented?

packages/wrangler/src/containers/ssh.ts
```diff
@@ -24,47 +24,47 @@ export function sshYargs(args: CommonYargsArgv) {
...
-					"SSH option: Can be used to give options in the format used in the configuration file",
+					"Sets `ssh -o`: Can be used to give options in the format used in the ssh configuration file",
 				type: "string",
 			})
 			.option("tag", {
 				describe:
-					"SSH option: Specify a tag name that may be used to select configuration in ssh_config(5)",
+					"Sets `ssh -P`: Specify a tag name that may be used to select configuration in ssh_config",
 				type: "string",
 			})
```
gpanders, line 69: I think all of these options should accept single character aliases (where possible) that match the flags in SSH, which would make the ``"Sets `ssh -X`"`` text redundant. I had planned to do that in a follow up PR.

Up to you if you think it's worth changing now before merging. Personally I'd like to get this merged sooner to unblock other work and fixup minor nits and papercuts in follow ups.

emily-shen: pasting this here as well posterity: i'm going to merge this pr as is once CI passes, and then you can do a followup PR?

the short flags is what aaron had originally, but some of them conflicted with wrangler's global flags. i think for consistency aaron went with just the full option names for everything. wdyt?

gpanders: Hmm I do think at the very least we should provide aliases for `-o`, `-F`, and `-i`, as those are fairly common when using the "real" `ssh` and it would be nice to support those in our wrapper too.

AFAICT `-c` and `-e` are the only conflicting flags, which correspond to `--cipher` and `--escape-char`. I'm perfectly ok with not providing aliases for those options since they're relatively more rare. That does add a bit of inconsistency since not all of the options have short flags, but that's also not exactly unusual with CLI tools.

emily-shen: that's fair, lets add them back in then!  idk what the best phrasing is, but the "Sets `ssh -o`" prefixes was trying to convey that these flags just pass through to the 'real ssh' flags, so it would be good to have something that indicates that still.

PR #10584, opened 2025-09-09
fix: Improve child process handling by separating command and arguments

kreeksec:
fix this, change the way the child process is launched so arguments are passed as an array to avoid shell interpretation, rather than composing a single shell command string. Specifically, rewrite `runCommand` to accept both a command and an argument array, and use `childProcess.execFileSync` instead of `execSync`. Update `getWranglerCommand` (and callers) to return or build the executable name and argument array separately, rather than interpolating command strings together.

child processes are executed in three files to avoid the risks and limitations associated with passing shell commands as strings to `execSync`. By using `execFileSync` with a separate command and argument array, we eliminate shell interpretation, which helps prevent issues with special characters and improves overall safety and reliability.

**Summary of changes:**

1. Replace `child_process.execSync` with `child_process.execFileSync` to avoid shell interpretation of dynamic arguments.
2. Refactor argument handling to pass the command and its arguments as separate parameters, rather than interpolating a full shell command string.
3. Ensure that dynamic values are passed strictly as arguments to avoid injection risks or failures due to shell metacharacters.

**Detailed changes by file:**

**In `packages/vite-plugin-cloudflare/e2e/helpers.ts`:**

* Refactored the `getWranglerCommand(command: string)` function to return an object containing `{ file: string, args: string[] }` instead of a single command string.
* Updated `runCommand` to accept a command file and an argument array, with fallback support for existing usage patterns.
* Modified `runWrangler` and other related functions to pass the command and arguments separately.
* Replaced `execSync` with `execFileSync`, ensuring all arguments are passed safely and correctly.

**In `fixtures/pages-workerjs-directory/tests/index.test.ts`:**

* Replaced usage of `execSync` with `execFileSync`.
* Separated the executable and its arguments instead of using string interpolation.
* Ensured that the `file` variable is passed as an argument, not embedded within a shell command.
* Added the required import for `execFileSync`.

**Other changes:**

* Replaced string interpolations like `"--bindings=..."` with argument arrays, such as `["--bindings", value]`.
* No functional logic has been changed beyond this adjustment.
* No additional dependencies were added.

This refactor improves cross-platform compatibility, eliminates potential security concerns, and aligns the code with best practices for running child processes.

---

#### Refactor child process execution to use execFileSync for improved safety and correctness

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/fixtures/pages-d1-shim/tests/index.test.ts
+++ b/fixtures/pages-d1-shim/tests/index.test.ts
@@ -1,4 +1,4 @@
-import { execSync } from "child_process";
+import { execSync, execFileSync } from "child_process";
 import { mkdtempSync, readFileSync, realpathSync } from "fs";
 import { tmpdir } from "os";
 import * as path from "path";
@@ -14,8 +14,18 @@ describe("Pages D1 shim", () => {
 		);
 		const file = join(tempDir, "./d1-pages.js");
 
-		execSync(
-			`npx wrangler pages functions build --outfile ${file} --bindings="{\\"d1_databases\\":{\\"FOO\\":{}}}"`,
+		execFileSync(
+			"npx",
+			[
+				"wrangler",
+				"pages",
+				"functions",
+				"build",
+				"--outfile",
+				file,
+				"--bindings",
+				'{"d1_databases":{"FOO":{}}}'
+			],
 			{
 				cwd: path.resolve(__dirname, ".."),
 			}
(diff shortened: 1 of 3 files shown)


vicb:
It looks like the lint test fails.
Could you please run `pnpm fix` and push a fixup commit?

PR #10593, opened 2025-09-09
fix: restore original url before passing request to miniflare

### LINKED ISSUE #9567: Request url received by worker is incorrect when using vite plugin + base option

vite: 6.3.5,   wrangler: 4.19.2, @cloudflare/vite-plugin: ^1.5.1,

https://github.com/cloudkite/vite-plugin-test

### Describe the Bug

When specifying the base option:

```js
// vite.config.ts
import { cloudflare } from "@cloudflare/vite-plugin";
import react from "@vitejs/plugin-react";
import { defineConfig } from "vite";

export default defineConfig({
  base: "/subpath",
  plugins: [react(), cloudflare()],
});
```

```js
// worker.ts
export default {
  fetch(request: Request) {
    console.log(request.url);
    return new Response(null, { status: 200 });
  },
};
```

```jsx
// main.tsx
import { StrictMode } from "react";
import { createRoot } from "react-dom/client";

function App() {
  return (
    <button
      type="button"
      onClick={() => {
        fetch("/subpath/api/")
      }}
    >
      Click me
    </button>
  );
}

createRoot(document.getElementById("root")).render(
  <StrictMode>
    <App />
  </StrictMode>,
);
```

When the button is clicked and `fetch("/subpath/api/")` is invoked, the base seems to be stripped from the URL.  The worker console.log prints:
http://localhost/api

I would expect the worker to receive the full url http://localhost/subpath/api

BlankParticle:
Fixes #9567

When using the `base` config with vite, vite dev server strips out the base part from the url
this means the worker receives a url with `/api/test` in production but `/test` in vite dev

we already has to fix this behaviour when working with `/cdn-cgi` urls for triggers and restore the originalUrl
so this PR actually restores that URL for all requests that are passed into miniflare

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


jamesopstad:
Hi @BlankParticle. Thanks for the PR. The `base` option still won't work correctly for Workers with assets (see https://github.com/cloudflare/workers-sdk/issues/9567#issuecomment-3019215109) but it's worth getting this fix in anyway.

Please could you add a test in `playground/worker` that uses a different Vite config with the `base` option e.g. `vite.config.base`. You can see `playground/assets` for examples of tests using different config files. Thanks.

BlankParticle:
I wasn't aware this was a problem with assets too, but yeah we can probably figure out something for that
maybe changing the client output will help, like this `/api` -> `dist/client/api`
but thats a separate issue

I will add the tests for now

aroman:
Sorry to necro-bump this, but it was unfortunately the source of quite some frustration for me this afternoon.

edit: I'm pretty sure this change broke SPAs with the `base` URL, as this change causes the request for `index.html` to not be served by vite when a `base` URL is set. I filed a bug with repro steps here:  https://github.com/cloudflare/workers-sdk/issues/11857

As @jamesopstad [noted](https://github.com/cloudflare/workers-sdk/pull/10593#issuecomment-3292891488), this change means that the vite plugin is now not only different from production, is it now _inconsistently different_ from production, because it works differently for assets and workers routes. Perhaps I'm overlooking something, but I believe this makes it significantly more difficult to get consistent behavior with a simple proxy rule, whereas before you could.

For our app, the original behavior _matched_ production, so now I'm needing to work around it the other way around. But I acknowledge that our setup is probably the more atypical one.

Is anyone aware of either:
A) a workaround to re-enable the previous behavior 
B) a timeline for when workers assets will support a "base URL" concept

Appreciate any insight anyone might have here!

(As an aside, this PR introduced a breaking change (how URLs are constructed is about as "breaking" as you can get for a development server IMO) without being acknowledged as such in the release notes — I had to dig very far back in the changelog to find the "patch" mentioning it, and even then, it doesn't mention the keyword "base", which is what I was grepping for, and is the name of the behavior it changed.)

PR #10596, opened 2025-09-09
Add Miniflare & Wrangler support for unbound Durable Objects

### LINKED ISSUE #10294: wrangler doesn't recognize a Durable Object namespace exists without a binding

wrangler 4.28.1 on Linux

_No response_

### Describe the Bug

Say I have this `wrangler.jsonc`:

```
{
  "name": "worker",
  "main": "src/index.ts",
  "compatibility_date": "2025-07-01",
  "compatibility_flags": ["experimental"],
  "migrations": [
    {
      "tag": "v0",
      "new_sqlite_classes": [ "MyDurableObject" ]
    }
  ]
}
```

I've declared via `migrations` that `MyDurableObject` is a Durable Object implementation and should have storage attached.

However, wrangler does not actually configure workerd to attach storage to this DO class unless I explicitly declare a binding for it.

With the upcoming `ctx.exports` API (currently available with the "experimental" compat flag), it will no longer be necessary to declare bindings. But, if workerd does not configure storage for the DO, then it won't show up in `ctx.exports`. So, currently as a work-around I have to declare a fake binding just to convince Wrangler to configure the storage. Once I do, `ctx.exports` works.

_No response_

penalosa:
Fixes #10294

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: internal implementation detail
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a patch

### DIFF
--- a/fixtures/unbound-durable-object/tests/index.test.ts
+++ b/fixtures/unbound-durable-object/tests/index.test.ts
@@ -1,0 +1,32 @@
+import { join, resolve } from "path";
+import { afterAll, beforeAll, describe, it } from "vitest";
+import { unstable_startWorker } from "wrangler";
+
+const basePath = resolve(__dirname, "..");
+
+describe("Unbound DO is available through `ctx.exports`", () => {
+	let worker: Awaited<ReturnType<typeof unstable_startWorker>>;
+
+	beforeAll(async () => {
+		worker = await unstable_startWorker({
+			config: join(basePath, "wrangler.jsonc"),
+		});
+	});
+
+	afterAll(async () => {
+		await worker.dispose();
+	});
+
+	it("storage operations don't throw", async ({ expect }) => {
+		const doName = crypto.randomUUID();
+		let response = await worker.fetch(`http://example.com?name=${doName}`);
+		let content = await response.text();
+		expect(content).toMatchInlineSnapshot(`"count: 0"`);
+
+		response = await worker.fetch(
+			`http://example.com/increment?name=${doName}`
+		);
+		content = await response.text();
+		expect(content).toMatchInlineSnapshot(`"count: 1"`);
+	});
+});
--- a/fixtures/unbound-durable-object/tests/index.test.ts
+++ b/fixtures/unbound-durable-object/tests/index.test.ts
@@ -17,7 +17,7 @@ describe("Unbound DO is available through `ctx.exports`", () => {
 		await worker.dispose();
 	});
 
-	it("storage operations don't throw", async ({ expect }) => {
+	it("can execute storage operations", async ({ expect }) => {
 		const doName = crypto.randomUUID();
 		let response = await worker.fetch(`http://example.com?name=${doName}`);
 		let content = await response.text();
--- a/packages/miniflare/src/index.ts
+++ b/packages/miniflare/src/index.ts
@@ -338,23 +338,10 @@ function getDurableObjectClassNames(
 		.flatMap((workerOpts) => {
 			const workerServiceName = getUserServiceName(workerOpts.core.name);
 
-			const boundDOs = Object.values(workerOpts.do.durableObjects ?? {}).map(
-				(workerDODesignator) => {
-					const doInfo = normaliseDurableObject(workerDODesignator);
-					if (doInfo.serviceName === undefined) {
-						// Fallback to current worker service if name not defined
-						doInfo.serviceName = workerServiceName;
-					}
-					return {
-						doInfo,
-						workerRawName: workerOpts.core.name,
-					};
-				}
-			);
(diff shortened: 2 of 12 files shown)


fixtures/unbound-durable-object/tests/index.test.ts
```diff
@@ -0,0 +1,32 @@
...
+		worker = await unstable_startWorker({
+			config: join(basePath, "wrangler.jsonc"),
+		});
+	});
+
+	afterAll(async () => {
+		await worker.dispose();
+	});
+
+	it("storage operations don't throw", async ({ expect }) => {
```
jamesopstad:
```suggestion
	it("can execute storage operations", async ({ expect }) => {
```

packages/miniflare/src/index.ts
```diff
@@ -351,6 +351,22 @@ function getDurableObjectClassNames(
...
+				const doInfo = normaliseDurableObject(workerDODesignator);
+				if (doInfo.serviceName === undefined) {
+					// Fallback to current worker service if name not defined
+					doInfo.serviceName = workerServiceName;
+				}
+				return {
+					doInfo,
+					workerRawName: workerOpts.core.name,
+				};
+			});
```
jamesopstad: This is exactly the same as the `map` callback above. Could we combine before mapping e.g. `[...Object.values(workerOpts.do.durableObjects ?? {}), ...(workerOpts.do.additionalUnboundDurableObjects ?? [])].map()`. Either that or extract the callback into a function that's reused.

packages/wrangler/src/dev/miniflare/index.ts
```diff
@@ -647,6 +648,37 @@ export function buildMiniflareBindingOptions(
 		}
 	}
 
+	/**
+	 * The `durableObjects` variable contains all DO bindings. However, this
+	 * may not represent all DOs defined in the app, because DOs can be defined
+	 * without being bound (accessible via ctx.exports).
+	 * To get a list of all configured DOS, we need all DOs provisioned via migrations,
+	 * wich we already have in the form of `classNameToUseSQLite`
```
jamesopstad:
```suggestion
	 * which we already have in the form of `classNameToUseSQLite`
```

packages/wrangler/src/dev/miniflare/index.ts
```diff
@@ -647,6 +648,37 @@ export function buildMiniflareBindingOptions(
...
+	 * As such, this code extends the list of bound DOs with configured DOs that
+	 * aren't already referenced. The outcome is that `additionalUnboundDurableObjects` will
+	 * contain DOs configured via migrations that are not bound.
+	 */
+	const additionalUnboundDurableObjects: WorkerOptionsBindings["additionalUnboundDurableObjects"] =
+		[];
+
+	for (const [className, useSQLite] of classNameToUseSQLite) {
```
jamesopstad, line 655: Is this feature only available if using SQLite-backed DOs? Seems reasonable but just checking that's the constraint.

packages/miniflare/src/plugins/do/index.ts
```diff
@@ -18,32 +18,30 @@ export const DOContainerOptionsSchema = z.object({
...
+	remoteProxyConnectionString: z
+		.custom<RemoteProxyConnectionString>()
 		.optional(),
+	container: z.custom<DOContainerOptions>().optional(),
+});
+
+export const DurableObjectsOptionsSchema = z.object({
+	durableObjects: z.record(z.union([z.string(), DurableObject])).optional(),
+	// Not all DOs are configured as bindings! Include these in a different key
+	// These might just be configured via migrations, but should still be allocated storage for e..g ctx.exports support
```
jamesopstad:
```suggestion
	// These might just be configured via migrations, but should still be allocated storage for e.g. ctx.exports support
```

packages/wrangler/src/dev/miniflare/index.ts
penalosa: That variable is perhaps poorly named. It's a mapping of all DOs found in migrations to whether or not they use SQLite. So the feature will support non-sqlite DOs as well

penalosa:
> Looks good to me. Non-blocking question: The alternative approach would be to collect all the DOs in the same way (by storage) and then collect the bindings for those that have them. Would that also work or not?

In theory, yes, and I initially tried that approach. However, it ended up being much more invasive across Miniflare/Vite/Vitest since all of those assume in various places that the durable object config is a record of binding name to config. This approach seemed like a much simpler option to reduce the added complexity

PR #10605, opened 2025-09-10
[draft] add npx wrangler containers registry configure command

emily-shen:
wip 
CC-5835

this is to support configuring credentials for external (non cloudflare) registries.

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/wrangler/src/__tests__/containers/registries.test.ts
```diff
@@ -0,0 +1,310 @@
...
+	mockApiToken();
+	beforeEach(() => {
+		mockAccount();
+	});
+
+	afterEach(() => {
+		clearDialogs();
+	});
+
+	it("should not show in help", async () => {
```
nikitassharma: should `wrangler containers registry --help` show help?

packages/containers-shared/src/login.ts
```diff
@@ -1,38 +1,31 @@
...
 import { UserError } from "./error";
-import { getCloudflareContainerRegistry } from "./knobs";
 
 /**
  * Gets push and pull credentials for Cloudflare's managed image registry
  * and runs `docker login`, so subsequent image pushes or pulls are
  * authenticated
  */
-export async function dockerLoginManagedRegistry(pathToDocker: string) {
+export async function dockerLoginManagedRegistry(
```
nikitassharma:
```suggestion
export async function dockerLoginImageRegistry(
```

packages/containers-shared/src/login.ts
```diff
@@ -1,38 +1,31 @@
...
 	const child = spawn(
 		pathToDocker,
-		[
-			"login",
-			"--password-stdin",
-			"--username",
-			"v1",
-			getCloudflareContainerRegistry(),
-		],
+		["login", "--password-stdin", "--username", "v1", domain],
```
nikitassharma:
```suggestion
		["login", "--password-stdin", "--username", credentials.username, domain],
```

I think this also needs to be changed in `cloudchamber/images/images.ts:getCreds`

emily-shen: i've intentionally left the images commands unchanged since people shouldn't be using wrangler commands to interact with images that aren't in the managed registry.

packages/wrangler/src/__tests__/containers/registries.test.ts
emily-shen: yup that works, but this test is just asserting its hidden from the top level help so no one stumbles on this and gets confused whlie we're still working on it

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,261 @@
...
+			ImageRegistriesService.deleteImageRegistry(deleteArgs.DOMAIN)
+		);
+	} catch (e) {
+		if (e instanceof ApiError) {
+			if (e.status === 404) {
+				throw new UserError(
+					`The registry ${deleteArgs.DOMAIN} does not exist.`
+				);
+			}
+			throw new APIError(e.body.error ?? "Unknown API error");
```
nikitassharma: `e.status === 403` indicates an attempt to delete a managed registry

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,261 @@
...
+		} else {
+			let counter = 0;
+			for (const registry of res) {
+				if (counter < res.length - 1) {
+					updateStatus(registry.domain);
+				} else {
+					endSection(registry.domain);
+				}
+				counter++;
+				// do we want any other info?
```
nikitassharma: I think `kind` would be nice to have, but that will need to be updated in coordinator first

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,261 @@
...
+			logger.json(res);
+		} else if (res.length === 0) {
+			endSection("No external registries configured for this account");
+		} else {
+			let counter = 0;
+			for (const registry of res) {
+				if (counter < res.length - 1) {
+					updateStatus(registry.domain);
+				} else {
+					endSection(registry.domain);
```
nikitassharma: not very familiar with this api so might be wrong, but does this need to be `updateStatus` so that the last registry isn't formatted differently from the rest?

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	try {
+		const res = await promiseSpinner(
+			ImageRegistriesService.listImageRegistries()
+		);
+		if (listArgs.json || isNonInteractiveOrCI()) {
+			logger.json(res);
+		} else if (res.length === 0) {
+			endSection("No external registries configured for this account");
```
nikitassharma:
```suggestion
			endSection("No registries configured for this account");
```
nit: this will list all registries

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,261 @@
...
+				auth: JSON.stringify(credentials),
+				kind: registryType.type,
+			})
+		);
+	} catch (e) {
+		if (e instanceof ApiError) {
+			if (e.status === 409) {
+				throw new UserError(
+					`A registry with the domain ${configureArgs.DOMAIN} already exists. Use "wrangler containers registry delete ${configureArgs.DOMAIN}" to delete it first if you want to reconfigure it.`
+				);
```
nikitassharma, line 207: `e.status === 403` account doesn't have permission to create an external registry

packages/wrangler/src/containers/registries.ts
emily-shen: oh yeah i forgot about that, its why the last registry was orange.

packages/wrangler/src/containers/registries.ts
emily-shen: Do you think its more or less confusing to filter out the managed registries?

nikitassharma: I think okay to keep managed registries. Especially if someone were trying to confirm the domain when trying to configure an image from their managed registry? i.e. an image can be a dockerfile or from a domain that's returned from `registries list` (ignoring public images for now)

packages/wrangler/src/containers/index.ts
```diff
@@ -83,5 +84,12 @@ export const containers = (
 					deleteCommand,
 					containersScope
 				)(args)
+		)
+		.command(
+			"registry",
```
nikitassharma:
```suggestion
			"registries",
```
I'm not very tied to `registries`, I more just wanted to bring it up because it would match `containers` and `images`. It felt a little odd to use `wrangler registry list` and get back multiple registries. But I don't feel very strongly here, just wanted to make sure we think about this first

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,262 @@
...
+		describe: "domain to configure for the registry",
+		type: "string",
+		demandOption: true,
+	});
+}
+
+async function registryPutCommand(
+	configureArgs: StrictYargsOptionsToInterface<typeof registryPutYargs>
+) {
+	startSection("Configure container registry");
```
nikitassharma:
```suggestion
	startSection("Configure a container registry");
```

packages/wrangler/src/containers/registries.ts
emily-shen: i think a 403 could come back from api gateway, so we can't assume its a missing capability. i think coordinator returns a decent error message in that case, which we'll show. i'm overriding the 409 specifically with a more verbose error message since the command is 'put' but it can't actually put for now 🙈

packages/wrangler/src/containers/index.ts
emily-shen: yeah i agree, will change

.changeset/rare-bushes-prove.md
```diff
@@ -0,0 +1,8 @@
+---
+"@cloudflare/containers-shared": patch
+"wrangler": patch
```
petebacondarwin, line 3: For release review... this is an experimental feature so it is marked as patch rather than minor.

packages/wrangler/src/__tests__/containers/registries.test.ts
```diff
@@ -0,0 +1,320 @@
...
+			{ domain: "123456789012.dkr.ecr.us-west-2.amazonaws.com" },
+			{ domain: "987654321098.dkr.ecr.eu-west-1.amazonaws.com" },
+		];
+		mockListRegistries(mockRegistries);
+		await runWrangler("containers registries list");
+		expect(std.out).toMatchInlineSnapshot(`
+			"
+			 ⛅️ wrangler x.x.x
+			──────────────────"
+		`);
```
petebacondarwin, line 309: Do we care about this check?

petebacondarwin: Oh I see. It is showing that we display the header when not in JSON mode

packages/wrangler/src/__tests__/containers/registries.test.ts
```diff
@@ -0,0 +1,320 @@
...
+
+	afterEach(() => {
+		clearDialogs();
+	});
+
+	it("should delete a registry with confirmation", async () => {
+		setIsTTY(true);
+		const domain = "123456789012.dkr.ecr.us-west-2.amazonaws.com";
+		mockConfirm({
+			text: `Are you sure you want to delete the registry ${domain}? This action cannot be undone.`,
```
petebacondarwin: We are deleting from the configured list of registries, rather than deleting the registry itself, right?
I wonder if this message might cause concern?

packages/wrangler/src/__tests__/containers/registries.test.ts
```diff
@@ -0,0 +1,320 @@
...
+
+			"
+		`);
+		expect(std.out).toMatchInlineSnapshot(`
+			"? Are you sure you want to delete the registry 123456789012.dkr.ecr.us-west-2.amazonaws.com? This action cannot be undone.
+			🤖 Using fallback value in non-interactive context: yes"
+		`);
+	});
+
+	it("should delete a registry with --skip-confirmation flag", async () => {
```
petebacondarwin:
```suggestion
	it("should delete a registry in interactive mode with --skip-confirmation flag", async () => {
```

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,262 @@
...
+			log(
+				"You do not need to configure credentials for Cloudflare managed registries.\n"
+			);
+			endSection("No configuration required");
+			return;
+		case ExternalRegistryKind.ECR:
+			credentials = await configureAwsEcrRegistry(configureArgs.DOMAIN);
+			break;
+		default:
+			throw new Error(`Unhandled registry type: ${registryType.type}`);
```
petebacondarwin: UserError?

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,262 @@
...
+			})
+		);
+	} catch (e) {
+		if (e instanceof ApiError) {
+			if (e.status === 409) {
+				throw new UserError(
+					`A registry with the domain ${configureArgs.DOMAIN} already exists. Use "wrangler containers registries delete ${configureArgs.DOMAIN}" to delete it first if you want to reconfigure it.`
+				);
+			}
+			throw new FatalError(
```
petebacondarwin: The only difference between UserError and FatalError is that the latter lets you set a process exit code, which you haven't done here, so you may as well just throw UserError

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,262 @@
...
+	} = {};
+	if (isNonInteractiveOrCI()) {
+		// Non-interactive mode: expect JSON input via stdin
+		log("Reading AWS credentials from stdin...\n");
+
+		const stdinInput = trimTrailingWhitespace(await readFromStdin());
+		if (!stdinInput) {
+			throw new UserError(
+				"No input provided. In non-interactive mode, please pipe AWS credentials as JSON:\n" +
+					'echo \'{"AWS_ACCESS_KEY_ID":"...","AWS_SECRET_ACCESS_KEY":"..."}\' | wrangler containers registries put ' +
```
petebacondarwin: Perhaps a simpler example would be:


```suggestion
					`wrangler containers registries put ${domain} < credentials.json\n` +
					"where credentials.json looks like\n" +
					`{"AWS_ACCESS_KEY_ID":"...","AWS_SECRET_ACCESS_KEY":"..."}`
```

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,262 @@
...
+			endSection("No registries configured for this account");
+		} else {
+			res.forEach((registry) => {
+				updateStatus(registry.domain);
+			});
+			endSection(`End`);
+		}
+	} catch (e) {
+		if (e instanceof ApiError) {
+			throw new FatalError(
```
petebacondarwin:
```suggestion
			throw new UserError(
```

packages/wrangler/src/__tests__/containers/registries.test.ts
nikitassharma: I agree with this, maybe "delete the registry configuration for ..."
Also, would it make sense to disallow deleting the managed registry in wrangler? Coordinator will block it, but I'm not sure if it's worth blocking it at the wrangler level as well.

nikitassharma (approved):

packages/wrangler/src/containers/registries.ts
emily-shen: making them APIErrors instead

Hafiz613:
Nikita Sharma : ***@***.***

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,334 @@
...
+		if (stores.length === 0) {
+			const check = await confirm(
+				`No existing secret stores found. Create a secret store to store your registry credentials?`
+			);
+			if (!check) {
+				endSection("Cancelled.");
+				return;
+			}
+			const res = await promiseSpinner(
+				// should we allow users to specify the name of the store?
```
nikitassharma: fwiw, when I interacted with AI Gateway in the dash, a secrets store store was created without prompting with the name `default_secrets_store`. I think it's okay if we don't give the option to specify?

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,334 @@
...
+
+	log("\n");
+	log("Setting up integration with Secrets Store...\n");
+	const accountId = await getAccountId(config);
+	let secretStoreId = configureArgs.secretStoreId;
+	if (!secretStoreId) {
+		const stores = await listStores(config, accountId);
+		if (stores.length === 0) {
+			const check = await confirm(
+				`No existing secret stores found. Create a secret store to store your registry credentials?`
```
nikitassharma:
```suggestion
				`No existing Secrets Store stores found. Create a Secrets Store store to store your registry credentials?`
```
I'm not sure I like the number of times this would use the word store, but I'm wondering if it's worth it to be more consistent with how the secrets store commands work?

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,334 @@
...
+		}
+	}
+	log("\n");
+	let secretName = configureArgs.secretName;
+	while (!secretName) {
+		try {
+			const res = await prompt(
+				`Please provide a name for the secret to store the registry credentials:`,
+				{ defaultValue: `${registryType.secretName?.replaceAll(" ", "_")}` }
+			);
```
nikitassharma: I'm wondering if we should automatically generate the name using something like the domain given that's unique? The prompts feel like they might be introducing quite a bit of friction when configuring a registry

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,334 @@
...
+	);
+	log(`Container-scoped secret ${secretName} created in Secrets Store.\n`);
+
+	try {
+		await promiseSpinner(
+			ImageRegistriesService.createImageRegistry({
+				domain: configureArgs.DOMAIN,
+				is_public: false,
+				auth: {
+					identifier: configureArgs.identifier,
```
nikitassharma: Should their be an interactive option for the identifier? (I don't have a strong opinion on this, I more care that there is a way to set it up non-interactively which there already is)

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,334 @@
...
+		// Non-interactive mode: expect JSON input via stdin
+		const stdinInput = trimTrailingWhitespace(await readFromStdin());
+		if (!stdinInput) {
+			throw new UserError(
+				"No input provided. In non-interactive mode, please pipe in the secret."
+			);
+		}
+		return stdinInput;
+	}
+	const secret = await prompt(`Enter secret:`, {
```
nikitassharma: can this function take registry type as an argument? that way the prompt can be `Enter AWS_SECRET_ACCESS_KEY`

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,334 @@
...
+		);
+		if (!yes) {
+			cancel("The operation has been cancelled");
+			return;
+		}
+	}
+
+	try {
+		await promiseSpinner(
+			ImageRegistriesService.deleteImageRegistry(deleteArgs.DOMAIN)
```
nikitassharma, line 313: I wonder if this api call should be updated to return the secrets store reference so that the secret can be deleted? or is that something we would even want to do?

packages/wrangler/src/containers/registries.ts
emily-shen: i'll match that name

packages/wrangler/src/containers/registries.ts
emily-shen: i tried that but its hard to make it readable given `.` is not an allowed character. i think the friction here is acceptable given there is a default falue

packages/wrangler/src/containers/registries.ts
emily-shen: good shout i should capitalise but the extra store is a bit tongue twistery 😅

packages/wrangler/src/containers/registries.ts
emily-shen: ah. yeah it probably should, but we would need to check if anything else is using it. i'll create a ticket to consider this in followup

packages/wrangler/src/__tests__/index.test.ts
```diff
@@ -59,6 +59,7 @@ describe("wrangler", () => {
 				  wrangler cert                   🪪 Manage client mTLS certificates and CA certificate chains used for secured connections [open-beta]
 				  wrangler pages                  ⚡️ Configure Cloudflare Pages
 				  wrangler mtls-certificate       🪪 Manage certificates used for mTLS connections
+				  wrangler containers             📦 Manage Containers
```
nikitassharma: 🎉

packages/wrangler/src/containers/index.ts
```diff
@@ -83,5 +84,12 @@ export const containers = (
 					deleteCommand,
 					containersScope
 				)(args)
+		)
+		.command(
+			"registries",
+			// hide for now so it doesn't show up in help while we test
```
nikitassharma:
```suggestion
			// hide for now so it doesn't show up in help while not publicly available
```

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,332 @@
...
+				describe: "Domain to configure for the registry",
+				type: "string",
+				demandOption: true,
+			})
+			.option("public-credential", {
+				type: "string",
+				description:
+					"The public part of the registry credentials, e.g. `AWS_ACCESS_KEY_ID` for ECR",
+				demandOption: true,
+				alias: ["id", "aws-access-key-id"],
```
nikitassharma:
```suggestion
				alias: ["aws-access-key-id"],
```
might risk confusion with secrets store id? I don't have a strong opinion here though

packages/wrangler/src/containers/registries.ts
nikitassharma: thanks!

packages/wrangler/src/secrets-store/commands.ts
```diff
@@ -746,3 +752,12 @@ export const secretsStoreSecretDuplicateCommand = createCommand({
 		logger.table(prettierSecret);
 	},
 });
+
+export const validateSecretName = (name: string) => {
+	const validName = /^[A-Za-z0-9_-]+$/;
```
nikitassharma: I _think_ this would be a breaking change, the [current regex](https://sourcegraph-oss.cfdata.org/gitlab.cfdata.org/cloudflare/https/secrets-store/-/blob/cmd/secrets-api/req/validation.go?L42) is more permissive

packages/wrangler/src/containers/registries.ts
nikitassharma: makes sense

packages/wrangler/src/containers/registries.ts
emily-shen: ah that was left over from when the flag was identifier. removing

packages/containers-shared/src/images.ts
```diff
@@ -173,3 +166,66 @@ export function resolveImageName(accountId: string, image: string): string {
...
+	}
+	let url: URL;
+	try {
+		url = new URL(`http://${domain}`);
+	} catch (e) {
+		if (e instanceof Error) {
+			throw new Error(`${domain} is invalid:\n${e.message}`);
+		}
+		throw e;
+	}
```
ascorbic, line 190: You could use `URL.canParse` here

packages/wrangler/src/index.ts
```diff
@@ -1214,7 +1214,7 @@ export function createCLIParser(argv: string[]) {
 	});
 
 	// containers
-	wrangler.command("containers", false, (containersArgs) => {
+	wrangler.command("containers", "📦 Manage Containers ", (containersArgs) => {
```
ascorbic:
```suggestion
	wrangler.command("containers", "📦 Manage Containers", (containersArgs) => {
```

Not sure if this will break all the snapshots!

packages/wrangler/src/secrets-store/commands.ts
```diff
@@ -746,3 +752,12 @@ export const secretsStoreSecretDuplicateCommand = createCommand({
 		logger.table(prettierSecret);
 	},
 });
+
+export const validateSecretName = (name: string) => {
+	const validName = /^[A-z0-9-_]+$/;
```
vicb, line 757: This looks like a valid concern, maybe switch to `/^[\w-]+$/` (`\w` include `_`)

packages/containers-shared/src/client/models/CreateImageRegistryRequestBody.ts
```diff
@@ -3,17 +3,18 @@
 /* eslint-disable */
 
 import type { Domain } from "./Domain";
+import type { ExternalRegistryKind } from "./ExternalRegistryKind";
+import type { ImageRegistryAuth } from "./ImageRegistryAuth";
 
 /**
  * Request body for creating a new image registry configuration
  */
```
vicb, line 11: Why is this comment removed?
Should the new fields be documented?

packages/containers-shared/src/client/models/ExternalRegistryKind.ts
```diff
@@ -0,0 +1,10 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
```
vicb: Could this be removed everywhere it is added (not applicable to this repo)?

packages/containers-shared/src/client/models/ExternalRegistryKind.ts
```diff
@@ -0,0 +1,10 @@
...
+/**
+ * The type of external registry that is being configured.
+ */
+export enum ExternalRegistryKind {
+	ECR = "ECR",
```
vicb, line 9: nit: any particular reason to give this an explicit name (` = "ECR"`)?

packages/containers-shared/src/client/models/ImageRegistryAuth.ts
```diff
@@ -0,0 +1,14 @@
+/* istanbul ignore file */
+/* tslint:disable */
+/* eslint-disable */
+
+/**
+ * A JSON string that encodes the auth required to authenticate with an external image registry. The format of the JSON object is determined by the registry being configured.
```
vicb, line 6: string ?

packages/containers-shared/src/images.ts
```diff
@@ -18,7 +16,8 @@ export async function pullImage(
 	dockerPath: string,
 	options: Exclude<ContainerDevOptions, DockerfileConfig>
 ): Promise<{ abort: () => void; ready: Promise<void> }> {
-	await dockerLoginManagedRegistry(dockerPath);
+	const domain = new URL(`http://${options.image_uri}`).hostname;
```
vicb, line 19: This could throw, should the error be handled?

packages/containers-shared/src/images.ts
```diff
@@ -173,3 +166,66 @@ export function resolveImageName(accountId: string, image: string): string {
 	// is managed registry and doesn't have the account id,add it to the path
 	return `${url.hostname}/${accountId}${url.pathname}`;
 }
+
+/**
+ * get type of container registry, and validate
```
vicb: Add uppercase, punctuation.

"mananged" -> "managed"

packages/wrangler/src/secrets-store/commands.ts
emily-shen: this is the same regex that secret store applies api side

packages/containers-shared/src/images.ts
```diff
@@ -173,3 +166,66 @@ export function resolveImageName(accountId: string, image: string): string {
...
+		url = new URL(`http://${domain}`);
+	} catch (e) {
+		if (e instanceof Error) {
+			throw new Error(`${domain} is invalid:\n${e.message}`);
+		}
+		throw e;
+	}
+
+	const acceptedRegistries: RegistryPattern[] = [
+		{
```
vicb, line 193: Would it make sense to consolidate this object (for ext registries) and type in an `ExternalRegistry.ts` file that would also have the content from the current `ExternalRegistryKind.ts`

packages/wrangler/src/__tests__/containers/config.test.ts
```diff
@@ -650,4 +650,101 @@ describe("getNormalizedContainerOptions", () => {
 		expect(result).toHaveLength(1);
 		expect(result[0].rollout_step_percentage).toBe(100);
 	});
+
+	describe("image validation and resolution", async () => {
```
vicb, line 654: Is "image validation" accurate here?
It looks like this is validating registries instead?

packages/containers-shared/src/images.ts
emily-shen: Everything in `containers-shared/src/client/` is autogenerated from an open api spec in the cloudchamber internal repo, and I would prefer to not mix in hand written code with the autogenerated code. also the autogenerated client will be removed soon anyway in favour of the cloudflare ts sdk :)

packages/wrangler/src/containers/index.ts
```diff
@@ -83,5 +84,12 @@ export const containers = (
 					deleteCommand,
 					containersScope
 				)(args)
+		)
+		.command(
+			"registries",
+			// hide for now so it doesn't show up in help while we not publicly available
```
vicb, line 90:
```suggestion
			// hide for now so it doesn't show up in help while not publicly available
```

packages/containers-shared/src/images.ts
emily-shen: this will already be validated by this point so it shouldn't, adding a comment to explain

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,332 @@
...
+				describe: "Domain to configure for the registry",
+				type: "string",
+				demandOption: true,
+			})
+			.option("public-credential", {
+				type: "string",
+				description:
+					"The public part of the registry credentials, e.g. `AWS_ACCESS_KEY_ID` for ECR",
+				demandOption: true,
+				alias: ["aws-access-key-id"],
```
vicb, line 80: Isn't that going to be a problem when we add more registry types?
i.e. you'll be able to use `aws-access-key-id` to configure GCP containers

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,332 @@
...
+		case ExternalRegistryKind.ECR:
+			log(`Getting ${registryType.secretType}...\n`);
+			secret = await getSecret(registryType.secretType);
+			break;
+		default:
+			throw new UserError(`Unhandled registry type: ${registryType.type}`);
+	}
+
+	log("\n");
+	log("Setting up integration with Secrets Store...\n");
```
vicb, line 130:
```suggestion
	log("\nSetting up integration with Secrets Store...\n");
```

packages/containers-shared/src/images.ts
emily-shen: alas we still support node 20 where this doesn't exist

packages/wrangler/src/containers/registries.ts
```diff
@@ -0,0 +1,332 @@
...
+	const accountId = await getAccountId(config);
+	let secretStoreId = configureArgs.secretStoreId;
+	if (!secretStoreId) {
+		const stores = await listStores(config, accountId);
+		if (stores.length === 0) {
+			const defaultStoreName = "default_secret_store";
+			const check = await confirm(
```
vicb, line 137: nit: use `yes`/`create` instead? The former is consistent with other usage in this PR, create is also nice to describe what is being done/asked.

vicb (approved):
The regexp should be updated.

Also added other minor comments.

LGTM with the changes.

packages/wrangler/src/__tests__/containers/config.test.ts
emily-shen: it is also validating how we resolve image names (e.g. adding account ids)

packages/containers-shared/src/client/models/ExternalRegistryKind.ts
emily-shen: everything in containers-shared/src/client/ is autogenerated from an openapi spec in the cloudchamber repo, and this is unfortunately full of 'any' uses so we do actually need to disable eslint. 
we could remove the others but again i'd like to avoid unnecessarily increasing the diff here.

packages/containers-shared/src/images.ts
```diff
@@ -173,3 +166,66 @@ export function resolveImageName(accountId: string, image: string): string {
 	// is managed registry and doesn't have the account id,add it to the path
 	return `${url.hostname}/${accountId}${url.pathname}`;
 }
+
+/**
+ * get type of container registry, and validate
+ * currently we support cloudflare managed registries and AWS ECR
+ * when using cloudflare mananged registries we expect CLOUDFLARE_CONTAINER_REGISTRY to be set
```
petebacondarwin, line 173:
```suggestion
 * when using cloudflare managed registries we expect CLOUDFLARE_CONTAINER_REGISTRY to be set
```

packages/containers-shared/src/images.ts
petebacondarwin, line 171:
```suggestion
 * Get type of container registry, and validate
```

emily-shen:
merging this as-is because the remaining comments are nits, this is hidden and i am going to do a follow-up pr immediately after

PR #10606, opened 2025-09-10
Use workerd `node:console` when available

petebacondarwin:
This changes the unenv-preset so that it only applies the unenv polyfills for `console` when the native implementation is not turned on (via compat flag or compat date).

The native code was merged into workerd at https://github.com/cloudflare/workerd/releases/tag/v1.20250911.0
The tests would fail if the production edgeworker doesn't include the correct native code.

All the members of the console module are the same between the polyfill and native except for the following:

- `_stderr` - in workerd it is `process.stderr`, in unenv it is `new Writable()`
- `_stdout` - in workerd it is `process.stdout`, in unenv it is `new Writable()`
- `_stdoutErrorHandler` - in workerd it throws an error, in unenv it is a noop
- `_stderrErrorHandler` - in workerd it throws an error, in unenv it is a noop

I think the first two are not breaking changes. In fact it provides a more full implementation natively than unenv.
The second two are breaking changes from unenv.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: covered by workerd docs
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  unenv-preset is not backported


packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -562,3 +615,39 @@ export const WorkerdTests: Record<string, () => void> = {
 		);
 	},
 };
+
+async function getPropertyTester<ImportType, GlobalType>(
+	importSpecifier: string,
+	importNamespace: ImportType,
+	globalProperty: keyof typeof globalThis,
+	globalObject: GlobalType
+) {
```
vicb: Could we:
- Converter parameters to an obj (line 71 is not vary readable)
- Add a return type
- Add JS Docs
- param 2 and 4 can probably be inferred from 1 and 3

packages/wrangler/turbo.json
```diff
@@ -70,6 +70,7 @@
 				"WRANGLER",
 				"WRANGLER_IMPORT",
 				"MINIFLARE_IMPORT",
+				"MINIFLARE_WORKERD_PATH",
```
vicb: I don't think this belongs in this PR.

It's a nice addition and we do not want to revert this if we revert the PR.

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: Unfortunately we can't infer the namespace from the import specifier because then esbuild treats it as dynamic and the unenv polyfill doesn't get used.

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -59,7 +59,60 @@ function generateTestListResponse(testName: string): Response {
...
+		testPropertyType("profile", "function");
+		testPropertyType("profileEnd", "function");
+		testPropertyType("table", "function");
+		testPropertyType("time", "function");
+		testPropertyType("timeEnd", "function");
+		testPropertyType("timeLog", "function");
+		testPropertyType("trace", "function");
+		testPropertyType("warn", "function");
+
+		// The undocumented APIs are supported in workerd natively.
```
vicb: Reword that comment?
The tests might use the unenv impl

packages/wrangler/turbo.json
petebacondarwin: Actually it was already here on line 69

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -22,7 +22,7 @@ export default {
 
 			default: {
 				// `/<test name>` executes the test or returns an html list of tests when not found
-				const testName = url.pathname.slice(1);
+				const testName = url.pathname.slice(1) as keyof typeof WorkerdTests;
```
vicb: is this needed?

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -32,7 +32,7 @@ export default {
 					await test();
 					return new Response("passed");
 				} catch (e) {
-					return new Response(`failed\n${e}`);
+					return new Response(`failed\n${(e as Error).stack ?? e}`);
```
vicb: Could you please move all changes that are not related to console to a different PR?

packages/wrangler/e2e/unenv-preset/worker/index.ts
vicb: More generally I'm questioning the need for this function (/see WET)

i.e. I could never guess what `testPropertyType` does without looking at the implementation.

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: This gives us type safety on the testName.

petebacondarwin: Otherwise `test` is just `any`.

.changeset/calm-carrots-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/unenv-preset": patch
+---
+
+Use workerd `node:console` when it is available
```
vicb, line 5: Could you please add details about the default enable compat date and the compat flag

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: And it removes the need to type the `workerdTests` with `Record<string, () => void>`

vicb: but it's wrong because `url.pathname.slice(1)` could be anything

vicb (dismissed):
We should check all the APIs available in unenv and make sure they have a similar behavior in the native implementation.

I stopped after seeing that `_stdoutErrorHandler` and `_stdoutErrorHandler` do not.

Could you please note all the differences in the PR description and then we can decide if we want to integrate that breaking change or not.

petebacondarwin:
> We should check all the APIs available in unenv and make sure they have a similar behavior in the native implementation.
> 
> I stopped after seeing that `_stdoutErrorHandler` and `_stdoutErrorHandler` do not.
> 
> Could you please note all the differences in the PR description and then we can decide if we want to integrate that breaking change or not.

Noted, PTAL

petebacondarwin:
Blocked on https://github.com/cloudflare/workerd/pull/5610

petebacondarwin:
https://github.com/cloudflare/workerd/pull/5610 has landed so this should become unblocked when that is released tomorrow.

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -59,6 +59,59 @@ function generateTestListResponse(testName: string): Response {
...
+		const importNamespace = await import("node:console");
+		const globalObject = globalThis.console;
+
+		assert.strictEqual(
+			globalThis.console,
+			importNamespace.default,
+			"expected `console` to be the same as `consoleImport.default`"
+		);
+
+		assertTypeOf(importNamespace.default, "object", "consoleImport.default");
```
vicb:
```suggestion
		assertTypeOf(importNamespace, "default", "object");
```

vicb:
rebased + bump workerd to 1202 where the console implementation has been merged

PR #10612, opened 2025-09-11
Added new pipelines commands (pipelines, streams, sinks, setup) [do not merge until 9/22]

jonesphillip:
Added new pipelines commands (pipelines, streams, sinks, setup), moved old pipelines commands behind --legacy

Approval: CR-1299934

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/25319
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: change is related to new product

### DIFF
--- a/packages/wrangler/src/pipelines/cli/get.ts
+++ b/packages/wrangler/src/pipelines/cli/get.ts
@@ -18,10 +18,10 @@ export const pipelinesGetCommand = createCommand({
 			describe: "The ID of the pipeline to retrieve",
 			demandOption: true,
 		},
-		format: {
-			choices: ["pretty", "json"],
-			describe: "The output format for pipeline",
-			default: "pretty",
+		json: {
+			describe: "Output in JSON format",
+			type: "boolean",
+			default: false,
 		},
 	},
 	positionalArgs: ["pipeline"],
@@ -42,7 +42,7 @@ export const pipelinesGetCommand = createCommand({
 					config,
 					accountId,
 					pipelineId,
-					args.format as "pretty" | "json"
+					args.json ? "json" : "pretty"
 				);
 
 				if (foundInLegacy) {
@@ -52,7 +52,7 @@ export const pipelinesGetCommand = createCommand({
 			throw error;
 		}
 
-		if (args.format === "json") {
+		if (args.json) {
 			logger.json(pipeline);
 			return;
 		}


packages/wrangler/src/pipelines/legacy-client.ts
penalosa, line 1: Could you make sure everything in here is legacy only? For instance, `generateR2ServiceToken` appears to be used in the new flows

packages/wrangler/src/pipelines/cli/sinks/create.ts
```diff
@@ -0,0 +1,277 @@
...
+			describe: "Time partition pattern (r2 sinks only)",
+			type: "string",
+		},
+		"roll-size": {
+			describe: "Roll file when size reaches (e.g., 100MB, 1GB)",
+			type: "string",
+			default:
+				SINK_DEFAULTS.rolling_policy.file_size_bytes === 0
+					? undefined
+					: `${SINK_DEFAULTS.rolling_policy.file_size_bytes}`,
```
penalosa: `SINK_DEFAULTS.rolling_policy.file_size_bytes` is a static value that's in the codebase, so why is this ternary needed?

packages/wrangler/src/pipelines/cli/sinks/create.ts
```diff
@@ -0,0 +1,277 @@
...
+		},
+		"roll-interval": {
+			describe: "Roll file when time reaches (e.g., 5m, 1h)",
+			type: "string",
+			default: `${SINK_DEFAULTS.rolling_policy.interval_seconds}s`,
```
penalosa: `interval_seconds` is a number without a unit, but the examples include a unit (e.g. `5m`). Will this work?

packages/wrangler/src/pipelines/cli/sinks/create.ts
```diff
@@ -0,0 +1,277 @@
...
+			}
+			if (!args.table) {
+				throw new UserError("--table is required for r2-data-catalog sinks");
+			}
+			if (!args.catalogToken) {
+				throw new UserError(
+					"--catalog-token is required for r2-data-catalog sinks"
+				);
+			}
+		}
```
penalosa, line 135: Validation of arguments should happen in a validate callback (not handler), and you can also make use of the yargs `conflicts` and `group` arguments to handle some validation for you

packages/wrangler/src/pipelines/cli/sinks/create.ts
```diff
@@ -0,0 +1,277 @@
...
+			}
+			if (args.rollInterval) {
+				// Parse interval (e.g., "300s" or "5m" -> seconds)
+				const intervalMatch = args.rollInterval.match(/^(\d+)([sm])?$/i);
+				if (intervalMatch) {
+					const interval = parseInt(intervalMatch[1]);
+					const unit = intervalMatch[2]?.toLowerCase() || "s";
+					interval_seconds = unit === "m" ? interval * 60 : interval;
+				}
+			}
```
penalosa, line 196: One of your example is `1h`, which this replacement strategy does not support

packages/wrangler/src/pipelines/cli/sinks/create.ts
```diff
@@ -0,0 +1,277 @@
...
+
+		if (args.rollSize || args.rollInterval) {
+			let file_size_bytes: number =
+				SINK_DEFAULTS.rolling_policy.file_size_bytes;
+			let interval_seconds: number =
+				SINK_DEFAULTS.rolling_policy.interval_seconds;
+
+			if (args.rollSize) {
+				// Parse file size (e.g., "100MB" -> bytes)
+				const sizeMatch = args.rollSize.match(/^(\d+)(MB|GB)?$/i);
```
penalosa: This isn't very robust. What about KB? What about MiB?

packages/wrangler/src/pipelines/cli/sinks/create.ts
penalosa: It also won't support fairly natural things like `1m30s`. I think Wrangler already includes a library to do this parsing, so this should probably use that?

packages/wrangler/src/pipelines/cli/sinks/create.ts
```diff
@@ -0,0 +1,277 @@
...
+					errorMessage = errorObj.notes[0].text;
+				} else if (error instanceof Error) {
+					errorMessage = error.message;
+				}
+			}
+
+			throw new UserError(
+				`Failed to create sink '${sinkName}': ${errorMessage}`
+			);
+		}
```
penalosa, line 259: If the API returns a sensible error message it'll be handled by Wrangler automatically (so there's no need for this try catch). If it doesn't (and returns "Unknown error", we should fix the API

packages/wrangler/src/pipelines/cli/sinks/delete.ts
```diff
@@ -0,0 +1,48 @@
...
+		},
+	},
+	async handler(args, { config }) {
+		await requireAuth(config);
+
+		const sink = await getSink(config, args.sink);
+
+		if (!args.force) {
+			const confirmedDelete = await confirm(
+				`Are you sure you want to delete the sink '${sink.name}' (${args.sink})?`
```
penalosa: This needs `fallbackValue: false` in order to not automatically be accepted in CI

packages/wrangler/src/pipelines/cli/sinks/get.ts
```diff
@@ -0,0 +1,49 @@
...
+			default: false,
+		},
+	},
+	async handler(args, { config }) {
+		await requireAuth(config);
+
+		const rawSink = await getSink(config, args.sink);
+
+		if (args.json) {
+			logger.log(JSON.stringify(rawSink, null, 2));
```
penalosa: This should use `logger.json`

packages/wrangler/src/pipelines/cli/sinks/list.ts
```diff
@@ -0,0 +1,67 @@
...
+		await requireAuth(config);
+
+		const sinks = await listSinks(config, {
+			page: args.page,
+			per_page: args.perPage,
+			pipeline_id: args.pipelineId,
+		});
+
+		if (args.json) {
+			logger.log(JSON.stringify(sinks, null, 2));
```
penalosa: As above re `logger.json`

penalosa (commented):
Just sinks—still need to review streams

packages/wrangler/src/pipelines/cli/streams/create.ts
```diff
@@ -0,0 +1,115 @@
...
+		await requireAuth(config);
+		const streamName = args.stream;
+
+		let corsOrigins: string[] | undefined;
+		if (args.corsOrigins) {
+			corsOrigins =
+				args.corsOrigins === "*"
+					? ["*"]
+					: args.corsOrigins.split(",").map((origin) => origin.trim());
+		}
```
penalosa: This should be an array argument (e.g. `--cors-origin example.com --cors-origin google.com`)

packages/wrangler/src/pipelines/cli/streams/create.ts
```diff
@@ -0,0 +1,115 @@
...
+				schema = parsedSchema;
+			} catch (error) {
+				throw new UserError(
+					`Failed to read or parse schema file '${args.schemaFile}': ${error instanceof Error ? error.message : String(error)}`
+				);
+			}
+		} else {
+			// No schema file provided - confirm with user
+			const confirmNoSchema = await confirm(
+				"No schema file provided. Create stream without a schema (unstructured JSON)?",
```
penalosa: Can you make this a question? i.e. `Do you want to create a stream...`

packages/wrangler/src/pipelines/cli/streams/create.ts
```diff
@@ -0,0 +1,115 @@
...
+
+				if (!parsedSchema || !Array.isArray(parsedSchema.fields)) {
+					throw new UserError("Schema file must contain a 'fields' array");
+				}
+
+				schema = parsedSchema;
+			} catch (error) {
+				throw new UserError(
+					`Failed to read or parse schema file '${args.schemaFile}': ${error instanceof Error ? error.message : String(error)}`
+				);
```
penalosa: This seems to be throwing a `UserError` which will then be caught and turned into a different `UserError`. Could you make the `try-catch` a little more targeted?

packages/wrangler/src/pipelines/cli/streams/utils.ts
```diff
@@ -0,0 +1,256 @@
...
+	logger.log("\nWorker Integration:");
+	logger.log("Add to wrangler.json:");
+	logger.log(`{
+  "pipelines": {
+    "streams": [{
+      "stream": "${stream.name}",
+      "binding": "${bindingName}"
+    }]
+  }
+}`);
```
penalosa: This should use the `updateConfigFile()` utility, which handles both TOML & JSON config formats (both of which are supported by Wrangler), and also offers to update the config file on the users behalf

packages/wrangler/src/pipelines/cli/create.ts
```diff
@@ -29,274 +16,132 @@ export const pipelinesCreateCommand = createCommand({
...
-			destination.credentials.access_key_id = auth.accessKeyId;
-			destination.credentials.secret_access_key = auth.secretAccessKey;
+		if (!args.sql && !args.sqlFile) {
+			throw new UserError("Either --sql or --sql-file must be provided");
 		}
-
-		if (!destination.credentials.access_key_id) {
-			throw new FatalError("Requires a r2 access key id");
+		if (args.sql && args.sqlFile) {
+			throw new UserError("Cannot specify both --sql and --sql-file");
```
penalosa: You can specify this using the `conflicts` yargs option

packages/wrangler/src/pipelines/cli/create.ts
```diff
@@ -29,274 +16,132 @@ export const pipelinesCreateCommand = createCommand({
...
+				sql = readFileSync(args.sqlFile, "utf-8").trim();
+			} catch (error) {
+				throw new UserError(
+					`Failed to read SQL file '${args.sqlFile}': ${error instanceof Error ? error.message : String(error)}`
+				);
+			}
+		} else {
+			// This should never happen due to earlier validation
+			throw new UserError("Either --sql or --sql-file must be provided");
 		}
```
penalosa, line 53: This isn't needed

packages/wrangler/src/pipelines/cli/create.ts
```diff
@@ -29,274 +16,132 @@ export const pipelinesCreateCommand = createCommand({
...
+					Array.isArray(errorObj.notes) &&
+					errorObj.notes[0]?.text
+				) {
+					errorMessage = errorObj.notes[0].text;
+				} else if (error instanceof Error) {
+					errorMessage = error.message;
 				}
 			}
+			throw new UserError(`SQL validation failed: ${errorMessage}`);
 		}
```
penalosa, line 93: This should not be needed. The API Error will be displayed by Wrangler and should have enough detail

packages/wrangler/src/pipelines/cli/create.ts
```diff
@@ -29,274 +16,132 @@ export const pipelinesCreateCommand = createCommand({
...
+		logger.log(`🌀 Creating pipeline '${pipelineName}'...`);
+
+		let pipeline;
+		try {
+			pipeline = await createPipeline(config, pipelineConfig);
+		} catch (error) {
+			if (error instanceof APIError && error.code === 10000) {
+				// Show error when no access to v1 Pipelines API
+				throw new UserError(
+					`An authentication error has occurred creating a pipeline with this version of Wrangler. Please try: npm install wrangler@4.36.0`
```
penalosa: This isn't very clear. Could you include the entire command (using `npx`) and explain why the error occurred?

packages/wrangler/src/pipelines/cli/create.ts
```diff
@@ -29,274 +16,132 @@ export const pipelinesCreateCommand = createCommand({
...
+				logger.log(
+					`\nRun 'wrangler pipelines get ${pipeline.id}' to view full details.`
+				);
+			}
+		} catch {
+			logger.warn("Could not fetch pipeline details for usage examples.");
+			logger.log(
+				`\nRun 'wrangler pipelines get ${pipeline.id}' to view full details.`
+			);
 		}
```
penalosa, line 139: Why is this error being suppressed?

packages/wrangler/src/pipelines/cli/get.ts
```diff
@@ -26,19 +27,77 @@ export const pipelinesGetCommand = createCommand({
...
-		switch (args.format) {
-			case "json":
-				logger.log(JSON.stringify(pipeline, null, 2));
-				break;
-			case "pretty":
-				logger.log(formatPipelinePretty(pipeline));
-				break;
+				if (foundInLegacy) {
+					return;
+				}
```
penalosa, line 50: Does this ever get shown to the user?

packages/wrangler/src/pipelines/cli/get.ts
```diff
@@ -26,19 +27,77 @@ export const pipelinesGetCommand = createCommand({
...
+			}
+			throw error;
+		}
+
+		if (args.format === "json") {
+			logger.log(JSON.stringify(pipeline, null, 2));
```
penalosa: Should be logger.json

packages/wrangler/src/pipelines/cli/setup.ts
```diff
@@ -0,0 +1,809 @@
...
+	sinkConfig: CreateSinkRequest;
+	pipelineConfig?: CreatePipelineRequest;
+}
+
+export const pipelinesSetupCommand = createCommand({
+	metadata: {
+		description: "Interactive setup for a complete pipeline",
+		owner: "Product: Pipelines",
+		status: "open-beta",
+	},
```
penalosa, line 48: How does this differ from `create`?

packages/wrangler/src/pipelines/cli/setup.ts
```diff
@@ -0,0 +1,809 @@
...
+	args: {
+		name: {
+			describe: "Pipeline name (optional - will prompt if not provided)",
```
penalosa:
```suggestion
			describe: "Pipeline name",
```

packages/wrangler/src/pipelines/cli/create.ts
jonesphillip: Don't want to fail command  in this case

packages/wrangler/src/pipelines/cli/get.ts
jonesphillip: yes it does

packages/wrangler/src/pipelines/cli/sinks/create.ts
jonesphillip: This would fail silently for me with no error, so catching instead to surface

packages/wrangler/src/pipelines/cli/streams/utils.ts
```diff
@@ -203,23 +205,26 @@ function generateExampleData(stream: Stream): string {
...
+		pipelines: [
+			{
+				pipeline: stream.name,
+				binding: bindingName,
+			},
+		],
+	};
+
+	logger.log("Add to your configuration file:");
+	logger.log(formatConfigSnippet(configSnippet, config.configPath));
```
penalosa: This should be using the `updateConfigFile()` helper

packages/wrangler/src/pipelines/cli/streams/utils.ts
penalosa: This was resolved, but doesn't seem to have been addressed?

vicb:
Validate PR Description has a legit failure

image

lrapoport-cf:
as discussed with team, please wait to land till monday (9/22)

cc @jonesphillip @dario-piotrowicz @oliy

packages/wrangler/src/pipelines/cli/streams/utils.ts
```diff
@@ -214,17 +218,19 @@ export function displayUsageExamples(stream: Stream, config: Config) {
...
+			pipelines: [
+				{
+					pipeline: stream.name,
+					binding: customBindingName ?? bindingName,
+				},
+			],
+		}),
+		config.configPath,
+		args.env,
+		false // Don't offer to update automatically
```
penalosa, line 232: This should be set to true, ideally. Why don't you want the config file updated automatically?

PR #10618, opened 2025-09-11
[unenv] Use the native workerd perf_hooks modules when available

petebacondarwin:
Previously, the unenv-preset always used a polyfill for perf_hooks, which was provided in the unenv libray. This polyfill used sniffing of the `performance.addEventListener` method to determine if the environment supported the perf_hooks API natively, and if so, it would use the native implementation.

However, now that the native `performance` object has been updated to provide stubs for many of these methods, the sniffing approach no longer works.

This change moves the perf_hooks polyfill into the unenv-preset package, and removes the sniffing logic. The decision whether to use the native or polyfill implementation is now made based on whether the `enable_nodejs_perf_hooks_module` flag is set.

---

There are two compat flags of note:
-  `enable_global_performance_classes` ("class") - only enables some standard performance classes and can be used without nodejs_compat.
-  `enable_nodejs_perf_hooks_module` ("module") - enables node:perf_hooks and also ensures that the performance global classes are avaiable (regardless of the `enable_global_performance_classes` flag).

---

- Tests
  - [x] Tests included/updated
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  unenv preset not backported

---

<a href="https://app.devin.ai/review/cloudflare/workers-sdk/pull/10618" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/packages/wrangler/e2e/unenv-preset/worker/index.ts
+++ b/packages/wrangler/e2e/unenv-preset/worker/index.ts
@@ -1105,38 +1105,6 @@ export const WorkerdTests: Record<string, () => Promise<void>> = {
 		assert.doesNotThrow(() =>
 			(perfHooks.performance.markResourceTiming as unknown as () => void)()
 		);
-
-		// Test all 24 constants exist
-		assertTypeOfProperties(perfHooks.constants, {
-			// GC type constants
-			NODE_PERFORMANCE_GC_MAJOR: "number",
-			NODE_PERFORMANCE_GC_MINOR: "number",
-			NODE_PERFORMANCE_GC_INCREMENTAL: "number",
-			NODE_PERFORMANCE_GC_WEAKCB: "number",
-			// GC flags constants
-			NODE_PERFORMANCE_GC_FLAGS_NO: "number",
-			NODE_PERFORMANCE_GC_FLAGS_CONSTRUCT_RETAINED: "number",
-			NODE_PERFORMANCE_GC_FLAGS_FORCED: "number",
-			NODE_PERFORMANCE_GC_FLAGS_SYNCHRONOUS_PHANTOM_PROCESSING: "number",
-			NODE_PERFORMANCE_GC_FLAGS_ALL_AVAILABLE_GARBAGE: "number",
-			NODE_PERFORMANCE_GC_FLAGS_ALL_EXTERNAL_MEMORY: "number",
-			NODE_PERFORMANCE_GC_FLAGS_SCHEDULE_IDLE: "number",
-			// Entry type constants
-			NODE_PERFORMANCE_ENTRY_TYPE_GC: "number",
-			NODE_PERFORMANCE_ENTRY_TYPE_HTTP: "number",
-			NODE_PERFORMANCE_ENTRY_TYPE_HTTP2: "number",
-			NODE_PERFORMANCE_ENTRY_TYPE_NET: "number",
-			NODE_PERFORMANCE_ENTRY_TYPE_DNS: "number",
-			// Milestone constants
-			NODE_PERFORMANCE_MILESTONE_TIME_ORIGIN_TIMESTAMP: "number",
-			NODE_PERFORMANCE_MILESTONE_TIME_ORIGIN: "number",
-			NODE_PERFORMANCE_MILESTONE_ENVIRONMENT: "number",
-			NODE_PERFORMANCE_MILESTONE_NODE_START: "number",
-			NODE_PERFORMANCE_MILESTONE_V8_START: "number",
-			NODE_PERFORMANCE_MILESTONE_LOOP_START: "number",
-			NODE_PERFORMANCE_MILESTONE_LOOP_EXIT: "number",
-			NODE_PERFORMANCE_MILESTONE_BOOTSTRAP_COMPLETE: "number",
-		});
 	},
 };
 


petebacondarwin:
Wrangler e2e failures are because the `enable_nodejs_perf_hooks_module` and `disable_nodejs_perf_hooks_module` compat flags are not yet available in the current workerd version we are using here.

vicb:
Could you please address [the points I commented on your other PR](https://github.com/cloudflare/workers-sdk/pull/10621#pullrequestreview-3215092224)

packages/unenv-preset/src/runtime/node/internal/performance.ts
```diff
@@ -0,0 +1,360 @@
...
+		if (this.timeOrigin === _timeOrigin) {
+			return _performanceNow();
+		}
+		return Date.now() - this.timeOrigin;
+	}
+
+	clearMarks(markName?: string | undefined): void {
+		this._entries = markName
+			? this._entries.filter((e) => e.name !== markName)
+			: this._entries.filter((e) => e.entryType !== "mark");
```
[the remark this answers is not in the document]
petebacondarwin: This bug is inherited from the unenv project itself, since this file was mostly copied over from there.

petebacondarwin:
Actually I realised there are a few more discrepancies in workerd that need tidying up so blocking on https://github.com/cloudflare/workerd/pull/6107

.changeset/lucky-chairs-relate.md
petebacondarwin, line 1: This change is not related to this PR but `pnpm fix` keeps wanting to remove the excess whitespace. So I am including it here.

packages/unenv-preset/src/preset.ts
```diff
@@ -5,7 +5,7 @@ import type { Preset } from "unenv";
 //
 // https://developers.cloudflare.com/workers/runtime-apis/nodejs/
 // https://github.com/cloudflare/workerd/tree/main/src/node
-const nativeModules = [
+const standardNativeModules = [
```
vicb: What does "standard" means?

Maybe expand the JS Doc. IIUC you mean native modules that are not behind any other compat flag that `nodejs_compat(_v2)`

Would `nodeJsCompatNativeModules` be a better name?

packages/unenv-preset/src/preset.ts
```diff
@@ -1096,3 +1099,49 @@ function getReadlineOverrides({
...
+
+/**
+ * Returns the overrides for `node:perf_hooks` (unenv or workerd)
+ *
+ * The native performance implementation:
+ * - is enabled starting from 2025-09-21
+ * - can be enabled with the "enable_nodejs_perf_hooks_module" flag
+ * - can be disabled with the "disable_nodejs_perf_hooks_module" flag
+ */
+function getPerfHooksOverrides({
```
vicb, line 977: What about `getPerfHookAndPerfomanceOverride` + updating the JSDoc

But there seems to be no dependency across the 2 so maybe 2 different functions?

packages/unenv-preset/src/preset.ts
```diff
@@ -1096,3 +1099,49 @@ function getReadlineOverrides({
...
+	);
+
+	const enabledModulesByFlag =
+		compatibilityFlags.includes("enable_nodejs_perf_hooks_module") &&
+		compatibilityFlags.includes("experimental");
+	const enabledGlobalsByFlag =
+		compatibilityFlags.includes("enable_global_performance_classes") &&
+		compatibilityFlags.includes("experimental");
+
+	const enabledModules = enabledModulesByFlag && !disabledModulesByFlag;
```
vicb: This is missing the enabled by date part.

packages/unenv-preset/src/preset.ts
```diff
@@ -1096,3 +1099,49 @@ function getReadlineOverrides({
...
+	const enabledGlobals = enabledGlobalsByFlag && !disabledGlobalsByFlag;
```
vicb: This is missing a `// TODO` to add `enabledByDate`

packages/unenv-preset/src/preset.ts
```diff
@@ -1096,3 +1099,49 @@ function getReadlineOverrides({
...
+
+	const nativeModules = enabledModules ? ["perf_hooks"] : [];
+	const hybridModules = enabledModules ? [] : ["perf_hooks"];
```
vicb: Could you add a comment here about why we need an hybrid module?

I think the reason is to get the `perf_hooks.ts` added in this PR.

This is different from other overrides where we do not add an explicit hybrid in order to use what unenv provides.

IMO worth explaining

packages/unenv-preset/src/preset.ts
```diff
@@ -1096,3 +1099,49 @@ function getReadlineOverrides({
...
+	const polyfills = !enabledGlobals
+		? ["@cloudflare/unenv-preset/polyfill/performance"]
+		: [];
+
+	// The native perf_hooks module should implement all the APIs.
+	// It can then be used as a native module.
+	return {
```
vicb: Could we use the same syntax/formatting as the other modules for consistency

packages/wrangler/e2e/unenv-preset/preset.test.ts
```diff
@@ -748,6 +748,39 @@ const localTestConfigs: TestConfig[] = [
 			},
 		},
 	],
+	// node:perf_hooks
+	[
```
vicb, line 755: The tests are testing when both flags are enabled or both flags are disabled.

We should also test the 2 cases where only one flag is enabled.

packages/unenv-preset/src/runtime/node/internal/performance.ts
```diff
@@ -0,0 +1,358 @@
...
+	name: string;
+	startTime: number;
+	constructor(name?: string, options?: PerformanceMarkOptions) {
+		if (name === undefined) {
+			throw new TypeError(
+				"Failed to construct 'PerformanceEntry': the `name` parameter is required."
+			);
+		}
+		this.name = name;
+		this.startTime = options?.startTime || _performanceNow();
```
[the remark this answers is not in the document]
vicb: Looks like tthe fix was reverted somehow?

packages/unenv-preset/src/runtime/node/perf_hooks.ts
```diff
@@ -0,0 +1,56 @@
+import {
```
vicb: Please add JSDoc about what this and when it is used (wrt the flags) and why this is used over the unenv version

packages/unenv-preset/src/runtime/node/internal/performance.ts
vicb, line 1: Please add JSDoc about what this and when it is used (wrt the flags) and why this is used over the unenv version

packages/unenv-preset/src/runtime/node/internal/performance.ts
```diff
@@ -0,0 +1,366 @@
+import { createNotImplementedError } from "unenv/_internal/utils";
+import type nodePerfHooks from "node:perf_hooks";
+
+const _timeOrigin = globalThis.performance?.timeOrigin ?? Date.now();
```
vicb: I guess we should know `globalThis.performance?` given the runtime value of the flag to enable global perf.

It would probably be clearer to use it. If not at least comment why `?`

vicb (commented):
Added more comments, waiting on answers before reviewing further

petebacondarwin:
Just for the record, I can't actually remember why there is cloudflare specific polyfill code in this PR rather than just inheriting the stuff from unenv itself. I'll check what the difference is. It might not be needed.

packages/unenv-preset/src/preset.ts
petebacondarwin: I basically just wanted to use the `nativeModules` variable further down, so needed a new name for this one. Don't really care what it is. Happy to go with your preference.
It's basically the non-hybrid/dynamic modules

packages/unenv-preset/src/preset.ts
petebacondarwin: In my other PR on workerd it will make the global flag implied by the module one.
So let's keep one fn but rename it

packages/unenv-preset/src/preset.ts
petebacondarwin: There isn't a compatibility date yet, right?

packages/unenv-preset/src/preset.ts
vicb: Even if implied it can still be disabled explicitly using a flag as I commented [here](https://github.com/cloudflare/workers-sdk/pull/10618#pullrequestreview-3865576646)

packages/unenv-preset/src/preset.ts
vicb: Oh right.
I was 
- trusting the JSDoc
- mixing module and global

So it looks like the JSDoc needs to be fixed and 2 TODOs added

packages/unenv-preset/src/preset.ts
vicb: Oh right.

I was 
- trusting the JSDoc
- mixing module and global

So it looks like the JSDoc needs to be fixed and 2 TODOs added

petebacondarwin:
> Just for the record, I can't actually remember why there is cloudflare specific polyfill code in this PR rather than just inheriting the stuff from unenv itself. I'll check what the difference is. It might not be needed.

Ah, I re-read from my PR description: 

> and removes the sniffing logic

So that is why we have moved the polyfill code into this preset.

packages/unenv-preset/src/preset.ts
petebacondarwin: Changing to `defaultNativeModules` and updated the JSDoc.

petebacondarwin:
The original PR provided local polyfills for performance because the unenv ones sniff the `performance.addEventListener` and won't polyfill if that is defined, which was previously broken since the native workerd version did define this but failed to define other parts of the `performance` global.

Now that the inheritance of `Performance` from `EventTarget` is conditional on the compat flag being enabled, which will also turn on the other necessary features on the performance global, we can rely on the compat flags controlling the code appropriately.

In other words

- when the flags are off, the performance global is missing essential features, including the `addEventListener` so unenv will correctly sniff and polyfill it.
- when the flags are on, the performance global has all the necessary features, so it is safe to turn off the polyfill altogether and the sniffing is irrelevant.

Therefore we no longer need to provide local copies of the unenv polyfills (that do not sniff) for this feature.

petebacondarwin:
There are a couple of very minor additions needed for workerd's Performance class: https://github.com/cloudflare/workerd/pull/6199. But I don't that is blocking this work.

packages/unenv-preset/src/preset.ts
```diff
@@ -63,117 +63,89 @@ export function getCloudflarePreset({
...
-		...childProcessOverrides.hybridModules,
-		...workerThreadsOverrides.hybridModules,
-		...readlineOverrides.hybridModules,
+	// The native workerd modules that are enabled for the current compat date and flags.
+	const nativeModules = [
+		...defaultNativeModules,
+		...overrides.flatMap((o) => o.nativeModules),
 	];
 
+	// The modules are are provided by unenv for the current compat date and flags.
```
petebacondarwin:
```suggestion
	// The modules that are provided by unenv for the current compat date and flags.
```

packages/unenv-preset/src/preset.ts
petebacondarwin: done

packages/wrangler/e2e/unenv-preset/preset.test.ts
petebacondarwin: only relying on one flag now

petebacondarwin:
Further investigation shows that all the required code is turned on by just the perf-hooks module compat flag, whether or not the global Performance classes compat flag is enabled/disabled. So we don't actually care about that flag at all.

Therefore, the code here is now back to almost trivial... it just needs to polyfill when the perf-hooks module is enabled.
That being said, I will add in some tests around the global classes flag just in case.

packages/unenv-preset/src/preset.ts
```diff
@@ -63,117 +63,89 @@ export function getCloudflarePreset({
...
-		...dgramOverrides.nativeModules,
-		...streamWrapOverrides.nativeModules,
-		...replOverrides.nativeModules,
-		...processOverrides.nativeModules,
-		...v8Overrides.nativeModules,
-		...ttyOverrides.nativeModules,
-		...childProcessOverrides.nativeModules,
-		...workerThreadsOverrides.nativeModules,
-		...readlineOverrides.nativeModules,
+	const overrides: {
```
vicb: nit: maybe create an explicit type an reuse it on all of the function too?

packages/unenv-preset/src/preset.ts
```diff
@@ -63,117 +63,89 @@ export function getCloudflarePreset({
...
+	const hybridModules = overrides.flatMap((o) => o.hybridModules);
+
+	// The globals that are injected from unenv for the current compat date and flags.
+	const hybridInjects = Object.assign(
+		{},
+		...overrides.map((o) => o.inject ?? {})
+	);
+
+	// The side-effect polyfill modules that are provided from unenv for the current compat date and flags.
+	const hybridPolyfills = overrides.flatMap((o) => o.polyfills ?? []);
```
vicb: could be be just "polyfills"?

```suggestion
	const polyfills = overrides.flatMap((o) => o.polyfills ?? []);
```

packages/unenv-preset/src/preset.ts
```diff
@@ -1096,3 +1068,39 @@ function getReadlineOverrides({
 				hybridModules: [],
 			};
 }
+
+/**
+ * Returns the overrides for `node:perf_hooks` (unenv or workerd)
+ *
+ * The native performance implementation:
+ * - is enabled starting from 2025-09-21
```
vicb: still not

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -28,12 +28,10 @@ export default {
 					return generateTestListResponse(testName);
 				}
 
-				try {
-					await test();
-					return new Response("passed");
-				} catch (e) {
-					return new Response(`failed\n${e}`);
-				}
+				return test().then(
```
vicb: IMO left is more readable

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -1062,6 +1060,105 @@ export const WorkerdTests: Record<string, () => void> = {
...
+		assertTypeOf(globalThis, "performance", "object");
+		assertTypeOf(globalThis, "PerformanceObserver", "function");
+
+		assert.doesNotThrow(() => globalThis.performance.now());
+		assert.doesNotThrow(() => globalThis.performance.mark("test"));
+		assert.doesNotThrow(() => globalThis.performance.measure("test"));
+
+		const perfHooks = await import("node:perf_hooks");
+
+		assertTypeOf(perfHooks, "performance", "object");
```
vicb: use the helper to avoid duplication here and below

vicb:
> Further investigation shows that all the required code is turned on by just the perf-hooks module compat flag, whether or not the global Performance classes compat flag is enabled/disabled. So we don't actually care about that flag at all.

Could you add the rationale of that to the PR description, i.e. what exactly `enable_global_performance_classes` does and why it is not needed.

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: The previous code didn't complain that the type of `test` was `() => void`, since you can await void. Whereas it is actually `() => Promise<void>`.
But happy to move back to an async function now I have fixed the type.

petebacondarwin:
Doh! It turns out that my analysis was not correct 😢 
The `Performance` classes are indeed not exposed on the global with the global classes flag.
They are available on the `perf_hooks` module, and they have the correct extra method when the `perf_hooks` flag is enabled.
So I think we will need to handle the combinations of flags ...

petebacondarwin:
But the good news is that the tests actually caught that!!

petebacondarwin:
Further update, we are going to fix this in workerd so that the perf_hooks flag will also cause these classes to be on the global when nodejs compat is on

packages/unenv-preset/src/preset.ts
petebacondarwin: we're fixing the complexity via workerd so this won't be a problem. Happy to rename this to `getPerformanceOverrides()` though?

petebacondarwin:
Currently blocked on https://github.com/cloudflare/workerd/pull/6206.

packages/unenv-preset/src/preset.ts
petebacondarwin: updated

packages/unenv-preset/src/preset.ts
petebacondarwin: Actually since we are no longer dealing with the performance globals flag, I think this can stay as `getPerfHooksOverrides`

packages/unenv-preset/src/preset.ts
petebacondarwin: done

packages/unenv-preset/src/preset.ts
petebacondarwin: done

packages/unenv-preset/src/preset.ts
petebacondarwin: removed date and added experimental

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: reverted

petebacondarwin:
> Currently blocked on [cloudflare/workerd#6206](https://github.com/cloudflare/workerd/pull/6206).

This is now merged but we'll need to bump the peer dependency once it is released

vicb:
> > Currently blocked on [cloudflare/workerd#6206](https://github.com/cloudflare/workerd/pull/6206).
> 
> This is now merged but we'll need to bump the peer dependency once it is released

And even beyond.
I think the later release ATM is 0305.0 so we should use at least 0305.1 when this goes out. 
If you use anything below that (which is the schedule we're back on now), you'll get 0305.0 which does not contain the change.

petebacondarwin:
Right - thanks @vicb - we'll need a manuel PR like https://github.com/cloudflare/workers-sdk/pull/12698 to pin to 20260301.1.

petebacondarwin:
> Right - thanks @vicb - we'll need a manuel PR like #12698 to pin to 20260301.1.

But I guess from a peer dependency point of view, if we need a future range, it will need to be `^20260305.1` since we have an older release from last week at `20260305.0` that would take precedence.

So that means we can't really land this PR until 5th March?

vicb:
> > Right - thanks @vicb - we'll need a manuel PR like #12698 to pin to 20260301.1.
> 
> But I guess from a peer dependency point of view, if we need a future range, it will need to be `^20260305.1` since we have an older release from last week at `20260305.0` that would take precedence.
> 
> So that means we can't really land this PR until 5th March?

Maybe you can exclude all the older release explicitly (the one in the future) in the range. I'm not sure if it works but worth trying.

petebacondarwin:
> > > Right - thanks @vicb - we'll need a manuel PR like #12698 to pin to 20260301.1.
> > 
> > 
> > But I guess from a peer dependency point of view, if we need a future range, it will need to be `^20260305.1` since we have an older release from last week at `20260305.0` that would take precedence.
> > So that means we can't really land this PR until 5th March?
> 
> Maybe you can exclude all the older release explicitly (the one in the future) in the range. I'm not sure if it works but worth trying.

That's a clever idea...

petebacondarwin:
## Note on the unusual `workerd` peer dependency range

The `workerd` peer dependency in `packages/unenv-preset/package.json` uses an unusual semver range:

```
>=1.20260301.1 <1.20260305.0 || >1.20260305.0 <2.0.0-0
```

This is intentional. There is a rogue version `1.20260305.0` of `workerd` that we must exclude. The range accepts all versions from `1.20260301.1` up to (but not including) `2.0.0`, while explicitly skipping `1.20260305.0`.

packages/unenv-preset/package.json
```diff
@@ -49,7 +49,7 @@
 	},
 	"peerDependencies": {
 		"unenv": "2.0.0-rc.24",
-		"workerd": "^1.20260218.0"
+		"workerd": ">=1.20260301.1 <1.20260305.0 || >1.20260305.0 <2.0.0-0"
```
vicb: I don't think this is correct?
i.e. 1.20260302.0 would match while I think it shouldn't

Probably something like

`1.20260301.1 || >= 1.20260302.1 < 1.20260303.0 || >= 1.20260303.1 < 1.20260304.0 || ... || >= 1.20260305.1`

I'll review the rest of the PR later tonight

petebacondarwin: Good point. I think maybe this will be satisfactory until we get to 5 March:

```
1.20260301.1 || 1.20260302.1 || 1.20260303.1 || 1.20260304.1 || >= 1.20260305.1 < 2.0.0-0
```

workers-devprod:
- ✅ @cloudflare/wrangler

</details>

packages/unenv-preset/package.json
vicb: Or

```
1.20260301.1 || ~1.20260302.1 || ~1.20260303.1 || ~1.20260304.1 || ^1.20260305.1
```

Or a mix of the above :)

I do like ` >= 1.20260305.1 < 2.0.0-0` more than caret but is it used less

packages/unenv-preset/src/preset.ts
```diff
@@ -54,126 +54,87 @@ const nativeModules = [
...
-		...childProcessOverrides.hybridModules,
-		...workerThreadsOverrides.hybridModules,
-		...readlineOverrides.hybridModules,
+	// The native workerd modules that are enabled for the current compat date and flags.
+	const nativeModules = [
+		...defaultNativeModules,
+		...overrides.flatMap((o) => o.nativeModules),
 	];
 
+	// The modules are are provided by unenv for the current compat date and flags.
```
vicb:
```suggestion
	// Hybrid modules are provided by unenv but use some native workerd APIs.
```

packages/unenv-preset/src/preset.ts
```diff
@@ -54,126 +54,87 @@ const nativeModules = [
...
 	return {
 		meta: {
 			name: "unenv:cloudflare",
 			version,
 			url: __filename,
 		},
 		alias: {
-			// `nodeCompatModules` are implemented in workerd.
-			// Create aliases to override polyfills defined in based environments.
+			// Alias each native module (and its `node:...` equivalent) to itself to ensure we override any polyfills from the base unenv presets.
```
vicb, line 104: Note: the newer modules are only available with the `node:` prefix, maybe create an issue to follow up on that. We have[ a list in this repo](https://github.com/cloudflare/workers-sdk/blob/8d0c8350f49638413d1a2359c07fef9b4745619d/packages/unenv-preset/src/index.ts#L13) - ditto for hybridModules below.

This should also probably be change in unenv if it exports soem of those modules

packages/unenv-preset/src/preset.ts
```diff
@@ -54,126 +54,87 @@ const nativeModules = [
...
-				dynamicHybridModules.flatMap((m) => [
+				hybridModules.flatMap((m) => [
 					[m, `@cloudflare/unenv-preset/node/${m}`],
 					[`node:${m}`, `@cloudflare/unenv-preset/node/${m}`],
 				])
 			),
 		},
 		inject: {
-			// Setting symbols implemented by workerd to `false` so that `inject`s defined in base presets are not used.
+			// Do not inject globals implemented natively by workerd, by setting the value to `false`.
```
vicb: note: I think "defined in base presets" was valuable to understand why this set them to `false`. There would be no need otherwise

packages/unenv-preset/src/preset.ts
```diff
@@ -54,126 +54,87 @@ const nativeModules = [
...
 			clearImmediate: false,
 			setImmediate: false,
-			...consoleOverrides.inject,
-			...processOverrides.inject,
+			// Inject globals provided by unenv for the current compat date and flags.
+			...injects,
 		},
-		polyfill: ["@cloudflare/unenv-preset/polyfill/performance"],
-		external: dynamicNativeModules.flatMap((p) => [p, `node:${p}`]),
+		polyfill: [...polyfills],
```
vicb:
```suggestion
		polyfill: polyfills,
```

packages/unenv-preset/package.json
petebacondarwin: Perfect! Updated

packages/unenv-preset/src/preset.ts
```diff
@@ -1096,3 +972,48 @@ function getReadlineOverrides({
...
+				polyfills: ["@cloudflare/unenv-preset/polyfill/performance"],
+			};
+}
+
+export type Compatibility = {
+	compatibilityDate: string;
+	compatibilityFlags: string[];
+};
+
+export type Override = {
```
vicb, line 1009: nit:: add the comments here rather than where the value are computed?

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -1062,6 +1062,82 @@ export const WorkerdTests: Record<string, () => void> = {
...
+		});
+
+		// Test performance.markResourceTiming() doesn't throw
+		// Note: Both workerd and unenv polyfill accept no arguments (stub implementations)
+		assert.doesNotThrow(() =>
+			(perfHooks.performance.markResourceTiming as unknown as () => void)()
+		);
+
+		// Test all 24 constants exist
+		assertTypeOfProperties(perfHooks.constants, {
```
vicb: Not sure if needed?
I we test constants we should probably test their values intead?
I don't think that test belong here (vs unenv or workerd)

vicb (dismissed):
LGTM, thanks Pete!

Added a few minor comments/suggestions

packages/unenv-preset/src/preset.ts
petebacondarwin: https://github.com/cloudflare/workers-sdk/issues/12736

packages/unenv-preset/src/preset.ts
petebacondarwin: Fair point. I'll update

petebacondarwin:
```suggestion
			// Do not inject globals implemented natively by workerd. Setting the value to `false` ensures that any base preset inject is not used.
```

packages/unenv-preset/src/preset.ts
petebacondarwin: Can you clarify which comments?

Do you mean things like:

```
// The native workerd modules that are enabled for the current compat date and flags.
```

petebacondarwin: [b038cb5](https://github.com/cloudflare/workers-sdk/pull/10618/commits/b038cb5cc60308fc45086755d28643e097e50e27)

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: Agreed workerd has rests for the values of these already.

petebacondarwin: ff90c5b594a198bef02aba059b6849efb74a6ccf

.changeset/loud-taxis-tie.md
```diff
@@ -0,0 +1,7 @@
+---
+"@cloudflare/unenv-preset": minor
+---
+
+Use the native workerd `node:perf_hooks` module and `Performance` global classes when available
```
vicb:
```suggestion
Use the native workerd `node:perf_hooks` module and the `Performance` global class when available
```

packages/unenv-preset/src/preset.ts
```diff
@@ -1,11 +1,11 @@
 import { version } from "../package.json";
 import type { Preset } from "unenv";
 
-// Built-in APIs provided by workerd.
+// Built-in APIs provided by workerd that do not need individual compatibility flags or dates.
```
vicb: You might want to use that.
If you do, would be good to mention v2 and 2024-09-23 - or simply link to the docs
```suggestion
// Built-in APIs provided by workerd when the `nodejs_compat` is used.
```

.changeset/loud-taxis-tie.md
petebacondarwin: There are multiple classes related to `Performance` that are enabled, so I would rather stick with the plural here. How about:

```suggestion
Use the native workerd `node:perf_hooks` module and `Performance` related global classes when available
```

packages/unenv-preset/src/preset.ts
petebacondarwin: Hmm. OK how about:

```suggestion
// Built-in APIs provided by workerd that are always enabled when `nodejs_compat` is enabled, and cannot be controlled individually by compatibility flags.
```

The point being that this whole preset is triggered by nodejs_compat flag, but some of the APIs can be individually switched on or off.

vicb: I like that version. 
Again nothing preventing the merge, already approved.

petebacondarwin:
> Oops, was not actually approved for some reason...

Just the new codeowners-plus action dismissing reviews - we have now turned that off

CMCDragonkai:
Will this finally allow the workers-sdk via wrangler to allow this code:

```ts
const { performance } = await import('node:perf_hooks');
```

Right now in wrangler 4.77 it's still `No such module "node:perf_hooks".`?

petebacondarwin:
@CMCDragonkai - have you set your compat date to after 17 March?

CMCDragonkai:
Yes, I just did and it worked.

PR #10621, opened 2025-09-11
Use native `node:vm` module when available

petebacondarwin:
The native code appears in workerd as of https://github.com/cloudflare/workerd/releases/tag/v1.20250910.0.
It was turned on by default (when `nodejs_compat` is on) on 2025-10-01.

The remote mode tests would fail if we didn't have a version of workerd in prod that supports this.

| Export                  | workerd       | unenv                            | diff                                  |
| ----------------------- | ------------- | -------------------------------- | ------------------------------------- |
| new Script                  | throws         | fake script object                                 | different                                  |
| Script.runInContext     | throws        | throws                           | same                                  |
| Script.runInNewContext  | throws        | throws                           | same                                  |
| Script.runInThisContext | throws        | throws                           | same                                  |
| Script.createCachedData | throws        | throws                           | same                                  |
| constants               | object        | object                           | same                                  |
| compileFunction         | throws        | throws                           | same                                  |
| createContext           | throws        | returns fake context object         | different                             |
| createScript            | throws        | returns new Script               | different                             |
| isContext               | returns false | returns true if fake context obj | n/a since no way to create in workerd |
| measureMemory           | throws        | returns fake promise             | different                             |
| runInContext            | throws        | throws                           | same                                  |
| runInThisContext        | throws        | throws                           | same                                  |
| runInNewContext         | throws        | throws                           | same                                  |

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  unenv is not backported

### DIFF
--- a/.changeset/upset-ducks-sin.md
+++ b/.changeset/upset-ducks-sin.md
@@ -1,0 +1,7 @@
+---
+"@cloudflare/unenv-preset": patch
+---
+
+Use native node:vm module when available
+
+It is enabled starting on 2025-10-01 or when the `enable_nodejs_vm_module` compatibility flag is set.
--- a/packages/wrangler/e2e/unenv-preset/worker/index.ts
+++ b/packages/wrangler/e2e/unenv-preset/worker/index.ts
@@ -698,6 +698,36 @@ export const WorkerdTests: Record<string, () => void> = {
 			/not implemented/
 		);
 	},
+
+	async testVm() {
+		const vm = await import("node:vm");
+
+		assertTypeOfProperties(vm, {
+			Script: "function",
+			constants: "object",
+			compileFunction: "function",
+			createContext: "function",
+			createScript: "function",
+			isContext: "function",
+			measureMemory: "function",
+			runInContext: "function",
+			runInThisContext: "function",
+			runInNewContext: "function",
+		});
+
+		assertTypeOfProperties(vm.default, {
+			Script: "function",
+			compileFunction: "function",
+			constants: "object",
+			createContext: "function",
+			isContext: "function",
+			measureMemory: "function",
+			runInContext: "function",
+			runInNewContext: "function",
+			runInThisContext: "function",
+			createScript: "function",
+		});
+	},
 };
 
 /**
--- a/packages/unenv-preset/src/preset.ts
+++ b/packages/unenv-preset/src/preset.ts
@@ -78,6 +78,7 @@ export function getCloudflarePreset({
 	const domainOverrides = getDomainOverrides(compat);
 	const wasiOverrides = getWasiOverrides(compat);
 	const consoleOverrides = getConsoleOverrides(compat);
+	const vmOverrides = getVmOverrides(compat);
 
 	// "dynamic" as they depend on the compatibility date and flags
 	const dynamicNativeModules = [
@@ -92,6 +93,7 @@ export function getCloudflarePreset({
 		...domainOverrides.nativeModules,
 		...wasiOverrides.nativeModules,
 		...consoleOverrides.nativeModules,
+		...vmOverrides.nativeModules,
 	];
 
 	// "dynamic" as they depend on the compatibility date and flags
@@ -107,6 +109,7 @@ export function getCloudflarePreset({
 		...domainOverrides.hybridModules,
 		...wasiOverrides.hybridModules,
 		...consoleOverrides.hybridModules,
+		...vmOverrides.hybridModules,
 	];
 
 	return {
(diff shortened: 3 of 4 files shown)


packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -505,4 +505,64 @@ export const WorkerdTests: Record<string, () => void> = {
...
+		assertType(
+			// @ts-expect-error undocumented API
+			vm.default.createScript,
+			"function",
+			"vm.default.createScript"
+		);
+	},
 };
+
+function assertType(value: unknown, type: string, name: string) {
```
vicb: +1 for adding this - it helps write more concise code and having the error message is quite useful while debugging.

We should refactor existing to use that in a follow up PR.
We could also add an helper taking a map/object of propName / type.

packages/wrangler/e2e/unenv-preset/worker/index.ts
```diff
@@ -505,4 +505,64 @@ export const WorkerdTests: Record<string, () => void> = {
 		assert.strictEqual(typeof http2.connect, "function");
 		assert.strictEqual(http2.constants.HTTP2_HEADER_STATUS, ":status");
 	},
+
+	async testVm() {
+		const vm = await import("node:vm");
+
+		assertType(vm.Script, "function", "vm.Script");
```
vicb: nit:

```ts
for (const target of [vm, vm.default]) {
  assertType(target.Script, "function", "Script");
  // ...
}
```

or a variation that adds the target name to the error message.

Can be done as a follow-up

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: Yes I agree. I was going to do this but thought I would get a sense from you as to whether that is useful or not first.
I prefer not to bundle them into a map/object. If we have them in a separate calls the callstack shows more clearly which test failed.
Also we would need to provide the receiver of the property to the helper too, so it doesn't really make things nicer.

packages/wrangler/e2e/unenv-preset/worker/index.ts
petebacondarwin: Thought about this but I find that there is a balancing act to have in tests.

The more logic you add to a test the more difficult it is to see what is going on ([DRY vs WET tests](https://dev.to/artshllaku/wet-vs-dry-testing-principles-you-should-know-4hob)).

Moreover, there are cases (not here) where the default does not have exactly the same properties as the import namespace (e.g. `node:console`). So then you get this situation where you have a bit of both.

packages/wrangler/e2e/unenv-preset/worker/index.ts
vicb: > If we have them in a separate calls the callstack shows more clearly which test failed. 

IMO the message is usually enough for that (I have been adding messages to debug failures before)

> Also we would need to provide the receiver of the property to the helper too, so it doesn't really make things nicer.

Again the message is probably good enough. We could also change `name: string` to `...messages: string[]`

But having the helper is already great. The map/obj is optional.

Thanks

packages/wrangler/e2e/unenv-preset/worker/index.ts
vicb: It's true that when I used it (i.e. `process`) the global and imported obj are the same.

My thinking was that it would be better for maintenance. If you have 30 LOC in the test, it's easy for someone to overlook that you should add 2 different assertions at 2 different places.

vicb: I usually refer DRY vs DAMP for tests.

How I read that is that the test case should contain everything needed to understand what's going on.

i.e. what the post your pointed to contain:

```ts
  test('should allow user to login with valid credentials', async () => {
    await login('user1', 'password1');
    await expect(page).toHaveURL('https://example.com/dashboard');
  });
```

If I need to scroll 100 lines up to figure out what the `login()` helper does then the test becomes hard to reason about. And if you happen to have to update the helper, it affect many tests.

IMO factoring code inside as a single test case doesn't go against WET/DAMP.

petebacondarwin: > It's true that when I used it (i.e. `process`) the global and imported obj are the same.
> 
> My thinking was that it would be better for maintenance. If you have 30 LOC in the test, it's easy for someone to overlook that you should add 2 different assertions at 2 different places.

This is not always the case. For example `console`:

```
> const c = await import("console");
> c._ignoreErrors
undefined
> c.default._ignoreErrors
true
> globalThis.console._ignoreErrors
true
```

Note that the import does not include the `_...` prefixed undocumented APIs.

.changeset/upset-ducks-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/unenv-preset": patch
+---
+
+Use native node:vm module when available
```
vicb, line 5: Could you please add details about the default enable compat date and the compat flag

vicb (dismissed):
This PR is a breaking change for compat date >= 2025-10-01 as some workerd function throws when unenv polyfill would not.

Could you please details all of those difference in the PR description.

IIRC Next.js uses some of those vm APIs so OpenNext might break after updating wrangler past this PR. We should at least run the Open Next e2e test suite with a compat date after October 1st.

petebacondarwin:
There are two main differences between the two implementations:

- workerd will throw when you call `createContext()` whereas unenv returns a fake context object
- workerd will throw if you create a new Script whereas unenv returns a fake script object

If these two are changed then the rest of the differences go away.

petebacondarwin:
Blocked on https://github.com/cloudflare/workerd/pull/5606

petebacondarwin:
The upstream PR has now landed, so this will be unblocked when it is released tomorrow.

vicb (approved):

vicb:
rebased + bump workerd to 1202 where the console implementation has been merged

PR #10622, opened 2025-09-11
feat: Update R2 Data Catalog compaction commands to work on catalog level (remove table/namespace)

nagraham:
Closes internal jira IGLOO-102.

Summary of changes
- Simplify experience of using R2 Data Catalog compaction enable/disable commands by applying to the whole Catalog (not individual tables)
- Require customers to set `--token` argument in `enable`
- Add ability for customers to configure the default target file size for compacted files
- Update the command status to "open-beta"

**Simplify Experience**
We want to modify the commands to simplify the R2 Data Catalog compaction experience for customers. Rather than enable/disable for individual tables (which is tedious), the commands will enable/disable for the whole Catalog. We will add back the table/namespace feature, but we want to wait to carefully consider its design in the user interface.

**Require `--token` arg to enable**
Our compaction system works asynchronously in the background, working to optimize customer's file sizes in their Iceberg catalogs. However, our system needs to access the customer R2 Bucket and R2 Data Catalog (similar to how a customer accesses it via a query engine). We encrypt and store their token, and only allow access via our compaction system. Because compaction wouldn't work without the token, we need to make it required when enabling for the whole catalog.

**Configure target file size**
Different workloads / tables have different requirements when it comes to the "optimal" size of Parquet files. Thus, we allow customers to set a default value that is applied to their tables.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/issues/25102
  - [] Documentation not necessary because:
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: This is a new product/feature

### DIFF
--- a/packages/wrangler/src/__tests__/r2.test.ts
+++ b/packages/wrangler/src/__tests__/r2.test.ts
@@ -1266,12 +1266,6 @@ For more details, refer to: https://developers.cloudflare.com/r2/api/s3/tokens/"
 				});
 
 				describe("enable", () => {
-					const mockToken = "test-token-123";
-
-					beforeEach(() => {
-						vi.stubEnv("CLOUDFLARE_API_TOKEN", mockToken);
-					});
-
 					it("should enable compaction for the catalog", async () => {
 						msw.use(
 							http.post(
@@ -1280,7 +1274,7 @@ For more details, refer to: https://developers.cloudflare.com/r2/api/s3/tokens/"
 									const body = await request.json();
 									expect(request.method).toEqual("POST");
 									expect(body).toEqual({
-										token: mockToken,
+										token: "fakecloudflaretoken",
 									});
 									return HttpResponse.json(
 										createFetchResult({ success: true }, true)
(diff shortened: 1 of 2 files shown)


nagraham:
Actually, after some internal discussion, we plan to switch to having the token in a required commandline arg (`--token`) rather than the `CLOUDFLARE_API_TOKEN` env variable. 

The reason for this is because in the future we hope to improve this command to use an OAuth flow that will allow users to grant permission for us to mint a new token with the exact permissions we need. The Pipelines team is working on this approach. We plan to migrate to it at a later time. Once we do, we can make the `--token` arg optional.

Additionally, our use case is slightly different than R2 SQL. They always require that token when executing queries. However, in our case, we only need it once when the customer enables their catalog. Further updates to the catalog's table maintenance settings (e.g. disabling/enabling, or updating default configurations) won't require setting that `CLOUDFLARE_API_TOKEN` again because we already have it saved.

PR #10623, opened 2025-09-11
containers: update push and build to resolve the correct uri

IRCody:
Before it was possible to get incorrect/confusing uris for images pushed with containers push and build.

Example:

containers push registry.cloudflare.com/image:tag would become registry.cloudflare.com/<accountid>/registry.cloudflare.com/image:tag.

This change modifies resolveImageName to behave like:

* image:tag -> prepend registry.cloudflare.com/accountid/ 
* registry.cloudflare.com/image:tag -> registry.cloudlfare.com/accountid/image:tag
* registry.cloudflare.com/accountid/image:tag -> no change
* anyother-registry.com/anything -> no change

Fixes #CC-6029

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Fixing behavior to be more correct not changing how the user interacts.
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because:  This feature is only in v4.

### DIFF
--- a/packages/wrangler/src/__tests__/containers/deploy.test.ts
+++ b/packages/wrangler/src/__tests__/containers/deploy.test.ts
@@ -1998,7 +1998,7 @@ function mockDockerImageInspectDigests(containerName: string, tag: string) {
 		expect(args).toEqual([
 			"image",
 			"inspect",
-			`${getCloudflareContainerRegistry()}/${containerName}:${tag}`,
+			`${containerName}:${tag}`,
 			"--format",
 			"{{ json .RepoDigests }} {{ .Id }}",
 		]);
@@ -2034,7 +2034,7 @@ function mockDockerImageInspectSize(containerName: string, tag: string) {
 		expect(args).toEqual([
 			"image",
 			"inspect",
-			`${getCloudflareContainerRegistry()}/${containerName}:${tag}`,
+			`${containerName}:${tag}`,
 			"--format",
 			"{{ .Size }} {{ len .RootFS.Layers }}",
 		]);
@@ -2113,7 +2113,7 @@ function mockDockerTag(from: string, to: string, tag: string) {
 		expect(cmd).toBe("/usr/bin/docker");
 		expect(args).toEqual([
 			"tag",
-			`${getCloudflareContainerRegistry()}/${from}:${tag}`,
+			`${from}:${tag}`,
 			`${getCloudflareContainerRegistry()}/${to}:${tag}`,
 		]);
 		return defaultChildProcess();


IRCody:
> this is much nicer, thanks :)
> 
> should we do anything to `wrangler containers build -p` for consistency?

Good call. I added that in the latest commit.

packages/containers-shared/src/images.ts
```diff
@@ -124,24 +125,29 @@ export async function prepareContainerImagesForDev(args: {
...
+ * anyother-registry.com/anything -> no change
  */
 export function resolveImageName(accountId: string, image: string): string {
 	let url: URL;
 	try {
 		url = new URL(`http://${image}`);
 	} catch {
-		return image;
+		// not a valid url so assume it is in the format image:tag and pre-pend the registry
+		return getCloudflareRegistryWithAccountNamespace(accountId, image);
```
gpanders, line 139: What if image is a file path, like `./Dockerfile`?

IRCody: It really should never be since this is only called before pushing an image.

I'm not sure if the function should be changed to deal with that or just change the name to be more reflective of what it is used for. WDYT?

gpanders: I think this is fine as is then, thanks for clarifying.

packages/wrangler/src/__tests__/cloudchamber/build.test.ts
```diff
@@ -52,6 +51,150 @@ describe("buildAndMaybePush", () => {
...
+		expect(dockerImageInspect).toHaveBeenNthCalledWith(2, "docker", {
+			imageTag: `${getCloudflareContainerRegistry()}/test-app:tag`,
+			formatString: "{{ .Size }} {{ len .RootFS.Layers }}",
+		});
+		expect(dockerLoginManagedRegistry).toHaveBeenCalledOnce();
+	});
+
+	it("should be able to build image and push with registry.cloudflare.com/test-app:tag", async () => {
+		await runWrangler(
+			"containers build ./container-context -t test-app:tag -p"
```
emily-shen: was this supposed to be "containers build ./container-context -t registry.cloudflare.com/test-app:tag -p"

same with the test below

IRCody: 😓 yea. Updating.

IRCody: I modified them.

I also changed build to use the users provided tag until we are going to push since that seems to make more sense to me?

emily-shen (approved):

PR #10628, opened 2025-09-11
Extend services bindings to support VPC Services.

efalcao:
Implements WVPC-42.

Extend services bindings to support VPC Services.

e.g. `"services": [{ "binding": "MYAPI", "service_id": "0199295b-b3ac-7760-8246-bca40877b3e9" }]`

I am pretty confident with the `deploy` and `types`, but less so with `wrangler dev` and its various modes - I will need to dig in more but wanted to get this initially reviewed. Also because `services` is now a union (either a `service` or a `service_id`) there is some added complexity here that probably needs improving.

Docs PR to follow
---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a v3 feature

### DIFF
--- a/packages/wrangler/src/api/startDevWorker/ConfigController.ts
+++ b/packages/wrangler/src/api/startDevWorker/ConfigController.ts
@@ -198,7 +198,29 @@ async function resolveBindings(
 				input.bindings
 			),
 			r2: extractBindingsOfType("r2_bucket", input.bindings),
-			services: extractBindingsOfType("service", input.bindings),
+			services: extractBindingsOfType("service", input.bindings).map(
+				(binding) => {
+					if (binding.service_id) {
+						return {
+							binding: binding.binding,
+							service_id: binding.service_id,
+						};
+					} else {
+						// eslint-disable-next-line @typescript-eslint/no-explicit-any
+						const serviceBinding: any = {
+							binding: binding.binding,
+							service: binding.service,
+						};
+						if ("environment" in binding && binding.environment) {
+							serviceBinding.environment = binding.environment;
+						}
+						if ("entrypoint" in binding && binding.entrypoint) {
+							serviceBinding.entrypoint = binding.entrypoint;
+						}
+						return serviceBinding;
+					}
+				}
+			),
 			d1Databases: extractBindingsOfType("d1", input.bindings),
 			ai: extractBindingsOfType("ai", input.bindings)?.[0],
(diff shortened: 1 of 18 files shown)


packages/wrangler/src/config/environment.ts
```diff
@@ -891,29 +891,46 @@ export interface EnvironmentNonInheritable {
 	 * @nonInheritable
 	 */
 	services:
```
penalosa, line 893: This is a pretty unusual way to add a new binding. Usually there'd be a new top level key `vpcs` which had these VPC services under it. Is there a particular reason this needs to be under `services`? I think it's probably going to cause a fair amount of user confusion, and will make the implementation more complex

packages/wrangler/src/api/startDevWorker/ConfigController.ts
```diff
@@ -198,7 +198,29 @@ async function resolveBindings(
...
-			services: extractBindingsOfType("service", input.bindings),
+			services: extractBindingsOfType("service", input.bindings).map(
+				(binding) => {
+					if (binding.service_id) {
+						return {
+							binding: binding.binding,
+							service_id: binding.service_id,
+						};
+					} else {
+						// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
penalosa, line 209:
```suggestion
```

packages/wrangler/src/deployment-bundle/create-worker-upload-form.ts
```diff
@@ -439,8 +444,16 @@ export function createWorkerUploadForm(worker: CfWorkerInit): FormData {
 		});
 	});
 
-	bindings.services?.forEach(
-		({ binding, service, environment, entrypoint, props }) => {
+	bindings.services?.forEach((serviceBinding) => {
+		if ("service_id" in serviceBinding && serviceBinding.service_id) {
+			metadataBindings.push({
+				name: serviceBinding.binding,
+				type: "connectivity_service_binding",
```
penalosa, line 451: This is a different binding type to `service`, and so should really be a different config property

penalosa: Also, it's a little odd to include a `_binding` suffix in the binding type—I don't think most others do it like that

PR #10634, opened 2025-09-12
fix: error if the container image uri has an account id that doesn't match the current account

emily-shen:
Fixes CC-5928

in your wrangler config, you can specify a container image uri with or without the account id and if you didn't, we would resolve it to include your account id when needed. 
so `registry.cloudflare.com/tag:name` resolves to `registry.cloudflare.com/account-id/tag:name` when deploying.

however, previously, if you had the wrong account id, we would still append the current id so you'd get
`registry.cloudflare.com/account-id/wrong-account-id/tag:name`.

technically, you could previously intentionally name an image with something that looks like an account id (e.g. aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/tag:name) and isn't, and now we would fail, but i think that is enough of an edge case that we should consider it a bug.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: bug
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: containers

### DIFF
--- a/packages/containers-shared/src/images.ts
+++ b/packages/containers-shared/src/images.ts
@@ -161,6 +161,5 @@ export function resolveImageName(accountId: string, image: string): string {
 	}
 
 	// is managed registry and doesn't have the account id,add it to the path
-	// is managed registry and doesn't have the account id,add it to the path
 	return `${url.hostname}/${accountId}${url.pathname}`;
 }


nikitassharma:
How do we handle the case of an image that looks like `wrong-account-id/image:tag` (or I guess `account-id/image:tag` as well), are we still prepending `registry.cloudflare.com/account-id` to both? Not sure if we should be explicitly handling that case.

emily-shen:
> How do we handle the case of an image that looks like `wrong-account-id/image:tag` (or I guess `account-id/image:tag` as well), are we still prepending `registry.cloudflare.com/account-id` to both? Not sure if we should be explicitly handling that case.

that's not allowed currently, users have to provide some sort of registry, so `registry.cloudflare.com/image:tag` is the minimum

packages/containers-shared/src/images.ts
```diff
@@ -138,10 +138,22 @@ export function resolveImageName(accountId: string, image: string): string {
...
 		return image;
 	}
+	// check if already looks like it has an account id (32 char hex string)
+	const accountIdPattern = /^\/([a-f0-9]{32})\//;
+	const match = accountIdPattern.exec(url.pathname);
+	if (match) {
+		const foundAccountId = match[1];
+		if (foundAccountId !== accountId) {
+			throw new Error(
+				`Image "${image}" does not belong to your account\nImage belongs to account: "${foundAccountId}"\nCurrent account: "${accountId}"`
```
emily-shen:
```suggestion
				`Image "${image}" does not belong to your account\nImage appears to belong to account: "${foundAccountId}"\nCurrent account: "${accountId}"`
```

packages/wrangler/src/__tests__/containers/deploy.test.ts
```diff
@@ -69,6 +69,25 @@ describe("wrangler deploy with containers", () => {
...
+					class_name: "ExampleDurableObject",
+					image:
+						"registry.cloudflare.com/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/hello:world",
+				},
+			],
+		});
+		await expect(runWrangler("deploy index.js")).rejects
+			.toThrowErrorMatchingInlineSnapshot(`
+			[Error: Image "registry.cloudflare.com/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/hello:world" does not belong to your account
+			Image belongs to account: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
emily-shen:
```suggestion
			Image appears to belong to account: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```

packages/containers-shared/src/images.ts
```diff
@@ -143,12 +143,24 @@ export function resolveImageName(accountId: string, image: string): string {
...
+		const foundAccountId = match[1];
+		if (foundAccountId !== accountId) {
+			throw new Error(
+				`Image "${image}" does not belong to your account\nImage appears to belong to account: "${foundAccountId}"\nCurrent account: "${accountId}"`
+			);
+		}
+		return image;
+	}
 
+	// is managed registry and doesn't have the account id,add it to the path
```
dario-piotrowicz:
```suggestion
```

You're dulicating the comment, I guess that's not intentional?

PR #10648, opened 2025-09-15
Build @cloudflare/pages-shared

GregBrimble:
Replaces https://github.com/cloudflare/workers-sdk/pull/10495

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: untestable
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not Wrangler change

### DIFF
--- a/.changeset/twelve-hands-throw.md
+++ b/.changeset/twelve-hands-throw.md
@@ -1,5 +1,5 @@
 ---
-"@cloudflare/pages-shared": minor
+"@cloudflare/pages-shared": patch
 ---
 
 Build `@cloudflare/pages-shared`. `@cloudflare/pages-shared/dist` now contains a build and bundled copy of `asset-server` and `metadata-generator`.
--- a/.changeset/twelve-hands-throw.md
+++ b/.changeset/twelve-hands-throw.md
@@ -1,0 +1,5 @@
+---
+"@cloudflare/pages-shared": minor
+---
+
+Build `@cloudflare/pages-shared`. `@cloudflare/pages-shared/dist` now contains a build and bundled copy of `asset-server` and `metadata-generator`.
--- a/packages/pages-shared/package.json
+++ b/packages/pages-shared/package.json
@@ -7,14 +7,17 @@
 		"directory": "packages/pages-shared"
 	},
 	"files": [
+		"./dist/",
 		"tsconfig.json",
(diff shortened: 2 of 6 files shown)


packages/pages-shared/package.json
```diff
@@ -7,14 +7,17 @@
...
 		"tsconfig.json",
 		"asset-server/**/*",
 		"environment-polyfills/**/*",
 		"metadata-generator/**/*"
 	],
 	"scripts": {
+		"build": "node -r esbuild-register ./scripts/build.ts",
 		"check:lint": "eslint . --max-warnings=0",
 		"check:type": "tsc",
+		"prepack": "npm run build",
```
penalosa, line 20: What's this for?

packages/pages-shared/tsconfig.build.json
penalosa, line 1: Why is this different to the regular tsconfig?

penalosa (approved):
Couple questions, but mostly LGTM

packages/pages-shared/package.json
GregBrimble: Habit of mine for distributed and built packages. It's a lifecycle hook for building it before packing up for distribution. [Looks like kv-asset-handler does it too](https://github.com/cloudflare/workers-sdk/blob/main/packages/kv-asset-handler/package.json#L35). Happy to remove (from both?) if you'd prefer to rely on the ordering of CI jobs though.

packages/pages-shared/tsconfig.build.json
GregBrimble: Sets the outdir, tells it to emit declarations, etc. I can turn these into CLI flags inside the build script when we're calling tsc, but figured you might prefer seeing that in an actual file rather than stringy CLI flags. Gotta go somewhere though.

penalosa: Fair enough, sounds good

.changeset/twelve-hands-throw.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/pages-shared": minor
```
penalosa:
```suggestion
"@cloudflare/pages-shared": patch
```

PR #10651, opened 2025-09-15
Add new attribute allowed_sender_addresses

edevil:
Fixes EMAIL-1228.

_Describe your change..._

Added new attribute called "allowed_sender_addresses" to send_email binding. This allows the binding to be restricted in the possible values for the sender of emails.

---

- Tests
  - [x] Tests included
- Public documentation
  - [X] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/25125
- Wrangler V3 Backport
  - [ ] Wrangler PR: 

### DIFF
--- a/.changeset/soft-flies-search.md
+++ b/.changeset/soft-flies-search.md
@@ -1,0 +1,6 @@
+---
+"miniflare": minor
+"wrangler": minor
+---
+
+Added new attribute "allowed_sender_addresses" to send email binding.
--- a/packages/miniflare/test/plugins/email/index.spec.ts
+++ b/packages/miniflare/test/plugins/email/index.spec.ts
@@ -263,6 +263,96 @@ This is a random email body.
 	t.is(res.status, 200);
 });
 
+test("Multiple allowed senders send_email binding works", async (t) => {
+	const mf = new Miniflare({
+		modules: true,
+		script: SEND_EMAIL_WORKER,
+		email: {
+			send_email: [
+				{
+					name: "SEND_EMAIL",
+					allowed_sender_addresses: [
+						"milchick@example.com",
+						"miss-huang@example.com",
+					],
+				},
+			],
+		},
+		compatibilityDate: "2025-03-17",
+	});
+
+	t.teardown(() => mf.dispose());
+
+	const res = await mf.dispatchFetch(
+		"http://localhost/?" +
+			new URLSearchParams({
+				to: "someone@example.com",
+				from: "milchick@example.com",
+			}).toString(),
+		{
+			body: `To: someone <someone@example.com>
+From: someone else <milchick@example.com>
+Message-ID: <im-a-random-message-id@example.com>
+MIME-Version: 1.0
+Content-Type: text/plain
+
+This is a random email body.
+`,
+			method: "POST",
+		}
+	);
+
+	t.is(await res.text(), "ok");
+	t.is(res.status, 200);
+});
+
+test("Sending email from a sender not in the allowed list does not work", async (t) => {
+	const mf = new Miniflare({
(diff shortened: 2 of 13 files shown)


.changeset/soft-flies-search.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": patch
```
dario-piotrowicz:
```suggestion
"miniflare": patch
"wrangler": patch
```

packages/miniflare/test/plugins/email/index.spec.ts
```diff
@@ -263,6 +263,49 @@ This is a random email body.
 	t.is(res.status, 200);
 });
 
+test("Multiple allowed senders send_email binding works", async (t) => {
```
dario-piotrowicz, line 266: We should also have a test for making sure that using a not allowed sender triggers the error

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -206,13 +206,22 @@ export function printBindings(
...
+					"allowed_sender_addresses" in emailBinding
+						? emailBinding.allowed_sender_addresses
+						: undefined;
+				let value =
+					destination_address ||
+					allowed_destination_addresses?.join(", ") ||
+					"unrestricted";
+
+				if (allowed_sender_addresses) {
+					value += ` - senders: ${allowed_sender_addresses.join(", ")}`;
```
dario-piotrowicz, line 219: I don't think that adding a ` - senders:` makes for a great UI:
Screenshot 2025-09-16 at 15 18 32

possibly we should have some other ways of showing this, and maybe even truncating or hiding the information 🤔 

I can also see that testing for this is missing here (it was also missing before this PR anyways)

I propose that we just go with the current solution here (as I don't want to block/hinder the PR) and then I can improve the UI and add tests as a followup task (unless you'd like to @edevil)

What do you think? 🙂

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -206,13 +206,22 @@ export function printBindings(
...
+				}
 				return {
 					name: emailBinding.name,
 					type: friendlyBindingNames.send_email,
-					value:
-						destination_address ||
-						allowed_destination_addresses?.join(", ") ||
-						"unrestricted",
+					value: value,
```
dario-piotrowicz:
```suggestion
					value,
```

.changeset/soft-flies-search.md
LuisDuarte1: adding a field on the binding should be a `minor` on both right? since it's a feature

packages/wrangler/src/utils/print-bindings.ts
edevil: Yes, I agree that the TUI could probably be improved. I mainly added the sender addresses because we were already adding the other restrictions.
But lets leave it for a follow up, when we figure out how to improve it. We'll probably look at how other bindings are being displayed.

.changeset/soft-flies-search.md
dario-piotrowicz: yeah good call! 🙂 

```suggestion
"miniflare": minor
"wrangler": minor
```

edevil: Done.

packages/miniflare/test/plugins/email/index.spec.ts
edevil: Added new test.

dario-piotrowicz (approved):
Looks great to me! thanks for the updates 😄 🫶

PR #10654, opened 2025-09-15
fix: update R2 SQL response format and env variable

laplab:
This PR changes env variable used for R2 SQL token from `CLOUDFLARE_API_TOKEN` to `WRANGLER_R2_SQL_AUTH_TOKEN`. It also updates the `query` command to use the latest version of response from R2 SQL service.

---

- Tests
  - [x] Tests included
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/issues/24988
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: It's a new service


packages/wrangler/src/r2/sql.ts
```diff
@@ -83,14 +81,22 @@ export const r2SqlQueryCommand = createCommand({
...
-		const token = getCloudflareAPITokenFromEnv();
+		let token = getWranglerR2SqlAuthToken();
 		if (!token) {
-			throw new UserError(
-				"Missing CLOUDFLARE_API_TOKEN environment variable. " +
-					"Please follow instructions in https://developers.cloudflare.com/r2/sql/platform/troubleshooting/ to create a token. " +
-					"Once done, you can prefix the command with the variable definition like so: `CLOUDFLARE_API_TOKEN=... wrangler r2 sql query ...`. " +
-					"There also other ways to provide the value of this variable, see https://developers.cloudflare.com/workers/wrangler/system-environment-variables/ for more details."
-			);
+			token = getCloudflareAPITokenFromEnv();
```
penalosa, line 86: What about `const token = getWranglerR2SqlAuthToken() ?? getCloudflareAPITokenFromEnv()`?

penalosa (approved):

packages/wrangler/src/r2/sql.ts
laplab: We could do that! The only thing is I want to keep the warning log about the fallback. I don't think that behaviour is super clear to the user and it would be great to have some way to notify them

PR #10665, opened 2025-09-16
feat(wrangler): Add Media binding support

austin-mc:
This PR introduces wrangler support for the upcoming the Media binding.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Docs to follow - not a public feature yet
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: New feature

### DIFF
--- a/packages/wrangler/e2e/dev-with-resources.test.ts
+++ b/packages/wrangler/e2e/dev-with-resources.test.ts
@@ -870,6 +870,36 @@ describe.sequential.each(RUNTIMES)("Bindings: $flags", ({ runtime, flags }) => {
 		});
 	});
 
+	it.skipIf(!CLOUDFLARE_ACCOUNT_ID)("exposes Media bindings", async () => {
+		await helper.seed({
+			"wrangler.toml": dedent`
+					name = "my-media-demo"
+					main = "src/index.ts"
+					compatibility_date = "2025-09-06"
+
+					[media]
+					binding = "MEDIA"
+					remote = true
+				`,
+			"src/index.ts": dedent`
+					export default {
+						async fetch(request, env, ctx) {
+							if (env.MEDIA === undefined) {
+								return new Response("env.MEDIA is undefined");
+							}
+
+							return new Response("env.MEDIA is available");
+						}
+					}
+				`,
+		});
+		const worker = helper.runLongLived(`wrangler dev`);
+		const { url } = await worker.waitForReady();
+		const res = await fetch(url);
+
+		await expect(res.text()).resolves.toBe("env.MEDIA is available");
+	});
+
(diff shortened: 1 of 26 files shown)


packages/wrangler/e2e/dev-with-resources.test.ts
penalosa, line 1: This should also have a test of behaviour in https://github.com/cloudflare/workers-sdk/tree/main/packages/wrangler/e2e/remote-binding/miniflare-remote-resources.test.ts

packages/wrangler/src/config/environment.ts
```diff
@@ -992,6 +992,23 @@ export interface EnvironmentNonInheritable {
...
+	 * NOTE: This field is not automatically inherited from the top level environment,
+	 * and so must be specified in every named environment.
+	 *
+	 * @default {}
+	 * @nonInheritable
+	 */
+	media:
+		| {
+				binding: string;
+				/** Whether the Media binding should be remote or not (only available under `--x-remote-bindings`) */
```
penalosa:
```suggestion
				/** Whether the Media binding should be remote or not */
```

packages/wrangler/src/utils/print-bindings.ts
```diff
@@ -790,7 +807,10 @@ export function warnOrError(
 			}
 		);
 	}
-	if (remote === false && supports === "remote") {
+	if (
+		remote === false &&
+		(supports === "remote" || supports === "always-remote")
+	) {
```
penalosa: This shouldn't include `always-remote`

packages/wrangler/src/config/environment.ts
austin-mc: Nice catch, fixed in force push!

packages/wrangler/src/utils/print-bindings.ts
austin-mc: Removed in force push!

packages/wrangler/e2e/dev-with-resources.test.ts
austin-mc: I have a test ready to go, but its blocked by ReadableStream support for remote bindings right now. The binding is currently gated to internal accounts only, are you ok with us merging this without that test and I can add it as soon as ReadableStreams are supported, or do you want to wait to merge?

vicb:
@austin-mc what is the status here? (also it looks like the PR needs a rebase)

austin-mc:
> @austin-mc what is the status here? (also it looks like the PR needs a rebase)

Still in progress, was blocked on ReadableStream support for remote bindings. That should be working now, just need to fix up an e2e test

PR #10668, opened 2025-09-16
wrangler: fix: support the deletion of secrets with complex names

danielrs:
URL encode the secret names before being sent to the API, should allow users to delete secrets with complex names. Change is still compatible with older Wrangler versions thanks to the optional query param flag `url_encoded`.

Tracking internally: https://jira.cfdata.org/browse/WC-3991

_Describe your change..._

URL encode all secret names in the `DELETE .../secrets/:secret` call, should allow the deletion of secrets with complex names.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: Updating the API docs, rather than developer docs.
- Wrangler V3 Backport
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/10760
  - [ ] Not necessary because: API still supports plain secret names.

---

Docs have new query param `url_encoded`:
https://developers.cloudflare.com/api/resources/workers/subresources/scripts/subresources/secrets/methods/delete/
image

### DIFF
--- a/.changeset/warm-clowns-visit.md
+++ b/.changeset/warm-clowns-visit.md
@@ -1,0 +1,5 @@
+---
+"wrangler": patch
+---
+
+Support the deletion of secrets with complex names
--- a/packages/wrangler/src/__tests__/secret.test.ts
+++ b/packages/wrangler/src/__tests__/secret.test.ts
@@ -563,16 +563,15 @@ describe("wrangler secret", () => {
 			msw.use(
 				http.delete(
 					`*/accounts/:accountId/workers/${servicesOrScripts}/:scriptName${environment}/secrets/:secretName`,
-					({ params }) => {
+					({ request, params }) => {
 						expect(params.accountId).toEqual("some-account-id");
 						expect(params.scriptName).toEqual(
 							legacyEnv && env ? `script-name-${env}` : "script-name"
 						);
-						if (!legacyEnv) {
-							if (env) {
-								expect(params.secretName).toEqual(input.secretName);
-							}
-						}
+						expect(params.secretName).toEqual(input.secretName);
+						expect(
+							new URL(request.url).searchParams.get("url_encoded")
+						).toEqual("true");
 						return HttpResponse.json(createFetchResult(null));
 					},
 					{ once: true }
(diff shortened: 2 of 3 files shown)


for-the-kidz:
should we do this for the other 3 secret endpoints as well?

for-the-kidz (approved):

danielrs:
- [x] Merge only after testing against API.

danielrs:
> should we do this for the other 3 secret endpoints as well?

Wrangler does not do `GET .../secrets/<secret>`. We also don't support secret manipulation on Workers for Platforms scripts from Wrangler.

packages/wrangler/src/__tests__/secret.test.ts
```diff
@@ -616,6 +615,23 @@ describe("wrangler secret", () => {
 			expect(std.err).toMatchInlineSnapshot(`""`);
 		});
 
+		it("should delete a secret: special characters", async () => {
```
dario-piotrowicz:
```suggestion
		it("should delete a secret which name includes special characters", async () => {
```

.changeset/warm-clowns-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"wrangler": minor
```
dario-piotrowicz:
```suggestion
"wrangler": patch
```

isn't this a patch? 🤔

.changeset/warm-clowns-visit.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Secret names are URL-encoded when deleting a secret.
```
dario-piotrowicz:
```suggestion
Support the deletion of secrets with complex names
```

dario-piotrowicz (approved):
lgtm 🙂 (with a couple of small suggestions)

petebacondarwin:
Interestingly the `wrangler versions secret delete` command already works with these complex keys...

petebacondarwin (approved):

petebacondarwin:
Since this is a patch (bug-fix) then it should indeed be backported to V3.

petebacondarwin:
The reason why the CI job to autocreate that backport failed is because this PR is from a fork.

PR #10652, opened 2025-09-15
Rename Hyperdrive env var with `CLOUDFLARE_` prefix

edmundhung:
Fixes [DEVX-2118](https://jira.cfdata.org/browse/DEVX-2118)

This PR renames Hyperdrive local connection string environment variable from `WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_<BINDING_NAME>` to `CLOUDFLARE_HYPERDRIVE_LOCAL_CONNECTION_STRING_<BINDING_NAME>`. The old variable name is still supported but will now show a deprecation warning.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: covered by existing tests
- Public documentation
  - [x] Cloudflare docs PR(s): https://github.com/cloudflare/cloudflare-docs/pull/25160
  - [ ] Documentation not necessary because:
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [x] Not necessary because: not a bugfix

PR #10677, opened 2025-09-16
fix(vite-plugin-cloudflare): populate build time env var from .env

edmundhung:

_Describe your change..._

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -112,6 +112,14 @@ export function resolvePluginConfig(
 		experimental: pluginConfig.experimental ?? {},
 	};
 	const root = userConfig.root ? path.resolve(userConfig.root) : process.cwd();
+	const buildTimeEnv = vite.loadEnv(viteEnv.mode, root, [
```
jamesopstad: Not sure `buildTimeEnv` is the right name for this. How about `prefixedEnv` or something similar?

packages/vite-plugin-cloudflare/src/plugin-config.ts
```diff
@@ -112,6 +112,14 @@ export function resolvePluginConfig(
...
+		"CLOUDFLARE_",
+		"WRANGLER_",
```
jamesopstad: Do we need to support `WRANGLER_` here as well? I think once https://github.com/cloudflare/workers-sdk/pull/10652 is merged, we could just support `CLOUDFLARE_` seeing it's a new feature.

.changeset/beige-results-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cloudflare/vite-plugin": patch
+---
+
+fix: populate build time env var from `.env`
```
jamesopstad:
```suggestion
fix: populate Hyperdrive local connection string from `.env`
```

jamesopstad: This should probably also include an example but I think we should land https://github.com/cloudflare/workers-sdk/pull/10652 first.

packages/vite-plugin-cloudflare/src/plugin-config.ts
edmundhung: I see why you want to skip `WRANGLER` here, but that means anyone setting `WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_<BINDING_NAME>` in a `.env` file will have it ignored without any warning message on the new name. How about adding a special case for it now, with a comment to remove it later?

edmundhung: e.g.
```ts
const cloudflareEnv = vite.loadEnv(viteEnv.mode, root, [
  "CLOUDFLARE_",
  "WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_",
]);
```

packages/vite-plugin-cloudflare/src/plugin-config.ts
edmundhung: Updated in https://github.com/cloudflare/workers-sdk/pull/10677/commits/d2ab32d64ed45f194dff81f999e683e2e6797e45

.changeset/beige-results-flow.md
edmundhung: Updated in https://github.com/cloudflare/workers-sdk/pull/10677/commits/d2ab32d64ed45f194dff81f999e683e2e6797e45

packages/vite-plugin-cloudflare/src/plugin-config.ts
edmundhung: Updated in https://github.com/cloudflare/workers-sdk/pull/10677/commits/d2ab32d64ed45f194dff81f999e683e2e6797e45 if you are fine with my suggestion

packages/vite-plugin-cloudflare/playground/bindings/.env
```diff
@@ -0,0 +1,2 @@
+# To test whether vite plugin could configure the hyperdrive binding properly
+WRANGLER_HYPERDRIVE_LOCAL_CONNECTION_STRING_HYPERDRIVE="postgres://user:password@example.com:3456/testdb"
```
jamesopstad: Should we test the `CLOUDFLARE_` prefixed version instead?

```suggestion
CLOUDFLARE_HYPERDRIVE_LOCAL_CONNECTION_STRING_HYPERDRIVE="postgres://user:password@example.com:3456/testdb"
```

michrome:
Hey @edmundhung is there an easy/quick way to bring this fix over to [`@astrojs/cloudflare`](https://github.com/withastro/astro/issues/14540#issuecomment-3468862480)?

Issue ref https://github.com/withastro/astro/issues/14540 🙏

PR #10678, opened 2025-09-16
Remove dummy auth & add trace header

penalosa:
Remove dummy auth from KV and add support for injecting trace headers

---

- Tests
  - [ ] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:

### DIFF
--- a/packages/wrangler/src/cfetch/internal.ts
+++ b/packages/wrangler/src/cfetch/internal.ts
@@ -2,7 +2,10 @@ import assert from "node:assert";
 import Cloudflare from "cloudflare";
 import { fetch, FormData, Headers, Request, Response } from "undici";
 import { version as wranglerVersion } from "../../package.json";
-import { getCloudflareApiBaseUrl } from "../environment-variables/misc-variables";
+import {
+	getCloudflareApiBaseUrl,
+	getTraceHeader,
+} from "../environment-variables/misc-variables";
 import { UserError } from "../errors";
 import { logger } from "../logger";
 import { APIError, parseJSON } from "../parse";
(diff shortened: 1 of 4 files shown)


packages/wrangler/src/cfetch/internal.ts
```diff
@@ -243,6 +243,12 @@ export function addUserAgent(headers: Headers): void {
 	headers.set("User-Agent", `wrangler/${wranglerVersion}`);
 }
 
+export function addTraceHeader(headers: Headers): void {
```
vicb: Maybe update the name to indicate that it might not add anything?
i.e. `addTraceHeaderWhenNeeded` or `maybeAddTraceHeader`

packages/wrangler/src/cfetch/internal.ts
```diff
@@ -243,6 +243,12 @@ export function addUserAgent(headers: Headers): void {
...
+	if (getTraceHeader()) {
+		headers.set("Cf-Trace-Id", getTraceHeader() as string);
+	}
```
vicb: nit: avoid the cast and calling the fn twice:

```suggestion
	const traceHeader = getTraceHeader();
	if (traceHeader) {
		headers.set("Cf-Trace-Id", traceHeader);
	}
```

packages/wrangler/src/environment-variables/factory.ts
```diff
@@ -100,7 +100,8 @@ type VariableNames =
 	/** Docker host configuration (handled separately from environment variable factory). */
 	| "WRANGLER_DOCKER_HOST"
 	/** Docker host configuration (handled separately from environment variable factory). */
-	| "DOCKER_HOST";
+	| "DOCKER_HOST"
+	| "WRANGLER_TRACE_ID";
```
vicb: Could you make it clearer that this is not in the `// ## Docker Configuration` section and add a comment.

Is this wrangler specific or should rather be `CLOUDFLARE_...`

vicb: Would `// ## Development & Local Testing` be a better section to add that to?

vicb (approved):
LGTM with minor comments/suggestions

PR #10683, opened 2025-09-17
fix(miniflare): api proxy preserve multiple Set-Cookie headers

yuripave:
On local service bindings, cross worker RPC will go through miniflare API proxy. Previous implementation of Headers serialization used `Object.fromEntries()` which works for most cases except when there are multiple Set-Cookie headers (since objects can't have duplicate keys). This PR fixes this by serializing Headers as an array of [key, value] pairs using `[...headers.entries()]` instead, which preserves all header values including multiple Set-Cookie headers.

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
- Wrangler V3 Backport
  - [ ] Wrangler PR: 
  - [ ] Not necessary because:


dario-piotrowicz (commented):
Thanks! This looks fantastic to me @yuripave 🚀 🫶 

The only thing missing, could you please add a changeset for this fix? 🙂 
(via `npx changeset` at the root of the repo)

yuripave:
> Thanks! This looks fantastic to me @yuripave 🚀 🫶
> 
> The only thing missing, could you please add a changeset for this fix? 🙂 (via `npx changeset` at the root of the repo)

done!

.changeset/hungry-crews-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": patch
+---
+
+fix(miniflare): api proxy preserve multiple Set-Cookie headers
```
dario-piotrowicz: Since this is a miniflare specific fix including "miniflare" in the changelog itself is not necessary :slightly_smiling_face: 

```suggestion
fix: api proxy preserve multiple Set-Cookie headers
```

dario-piotrowicz (approved):
@yuripave I'm really sorry for missing your earlier message, thanks for having pinged me again! :bow: 

Again, looks great to me! :smile:  with just one very minor final suggestion :slightly_smiling_face:

.changeset/hungry-crews-tap.md
yuripave: updated and rebased, thanks!

PR #10685, opened 2025-09-17
add `@cloudflare/` scope to all private packages and rename root package

dario-piotrowicz:
Fixes https://jira.cfdata.org/browse/DEVX-2202

This PR add the `@cloudflare/` scope to all private packages, it adds a check to validate that all relevant private packages follow this convention and additionally it renames the root package from `wrangler-root` to (`@cloudflare/`)`workers-sdk`

---

- Tests
  - [x] Tests included
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: change internal to the monorepo
- Wrangler V3 Backport
  - [ ] Wrangler PR: 


tools/deployments/validate-private-packages.ts
```diff
@@ -0,0 +1,71 @@
...
+if (require.main === module) {
+	console.log("::group::Checking package names");
+	checkPrivatePackageScopes()
+		.then((errors) => {
+			if (errors.length > 0) {
+				console.error(
+					"::error::Package names checks:" + errors.map((e) => `\n- ${e}`)
+				);
+			}
+			console.log("::endgroup::");
```
vicb, line 15: should that be in a `.finally(...)` instead?

vicb: not strictly required but would be nicer IMO.

Alternative is to move the `::group::` inside the then

tools/deployments/validate-private-packages.ts
```diff
@@ -0,0 +1,71 @@
...
+		})
+		.catch((error) => {
+			console.error("An unexpected error occurred", error);
+			process.exit(1);
+		});
+}
+
+async function getPrivatePackageJsons() {
+	// TODO: we only check package.jsons in packages and tools, in fixtures we use the `@fixture/` scope,
+	//       consider if we should use the `@cloudflare/` scope instead
```
vicb: nit: this TODO does not really belong here (vs /fixtures)

tools/deployments/validate-private-packages.ts
```diff
@@ -0,0 +1,71 @@
...
+	const packageJsonPaths = await glob("{packages,tools}/**/package.json", {
```
vicb: IIUC this depends on cwd being the root folder.
It would be nicer to be able to run this whatever the cwd?

tools/deployments/validate-private-packages.ts
dario-piotrowicz: I don't think using `finally` would work given the `process.exit()` calls, I've moved the `::group::` inside the then 👍

tools/deployments/validate-private-packages.ts
dario-piotrowicz: I don't think that's the case, have you seen the next line?
```ts
cwd: resolve(__dirname, "../../"),
```

tools/deployments/validate-private-packages.ts
dario-piotrowicz: Fixed, thanks 👍

tools/deployments/validate-private-packages.ts
dario-piotrowicz: Actually moving `::group::` inside the then doesn't work as intended:
Screenshot 2025-09-18 at 11 08 15

I'll just then also call `console.log("::endgroup::");` inside the catch

packages/vite-plugin-cloudflare/e2e/fixtures/invalid-worker-environment-options/package.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "@vite-plugin-cloudflare-e2e/invalid-worker-environment-options",
+	"name": "@cloudflare/invalid-worker-environment-options",
```
petebacondarwin: Can these be `@cloudflare/vite-plugin-e2e-` prefixed?

fixtures/import-npm/packages/import-example/package.json
```diff
@@ -1,5 +1,5 @@
 {
-	"name": "import-example",
+	"name": "@cloudflare/import-example",
```
petebacondarwin: Can these be `@cloudflare/npm-import-fixture-` prefixed?

petebacondarwin (approved):

fixtures/import-npm/packages/import-example/package.json
dario-piotrowicz: sorry, what do you beam by "these"?

packages/vite-plugin-cloudflare/e2e/fixtures/invalid-worker-environment-options/package.json
dario-piotrowicz: done 🙂

fixtures/import-npm/packages/import-example/package.json
dario-piotrowicz: Do you mean updating the name for all the fixtures so that they are prefixed with `@cloudflare/fixture-`?

petebacondarwin: These packages are only used in the `npm-import` fixture, right?
So it would be nice to make that clear by prefixing them.
E.g. `@cloudflare/npm-import-fixture-import-example`

dario-piotrowicz: updated 😄 

321c9c166fb7942cf31c798d74ff7a4dc5225876

tools/deployments/validate-private-packages.ts
```diff
@@ -0,0 +1,73 @@
...
+			// We are not interested in dependencies
+			"**/node_modules/**",
+			// The package.jsons in the fixtures directory use the `@fixture/` scope
+			// TODO: consider if we should use the `@cloudflare/` scope instead
+			"fixtures/**",
+			// C3 template package.jsons have <TBD> names that are populated by C3 during the app creation
+			"packages/create-cloudflare/templates/**",
+			// The package.jsons in the vite-plugin playground use the `@playground/` scope
+			// TODO: consider if we should use the `@cloudflare/` scope instead
+			"packages/vite-plugin-cloudflare/playground/**",
```
petebacondarwin, line 38: @dario-piotrowicz - we forgot to add `vendor/**` to this list

PR #10686, opened 2025-09-17
CFSQL-1371 reconcile dumpSql changes from D1 and document how to keep them in sync

alsuren:
Fixes CFSQL-1371

_Describe your change..._

This adds a comment at the top of dumpSql.ts so we don't let them get out of sync again. It also adds the d1 stats tracking machinery (which remains unused in miniflare) so that there now only whitespace differences between the two files.

---

- Tests
  - [ ] Tests included
  - [x] Tests not necessary because: stats tracking functionality is not used by miniflare (is extensively covered by D1's internal test suite)
- Public documentation
  - [ ] Cloudflare docs PR(s): 
  - [x] Documentation not necessary because: not public facing
- Wrangler V3 Backport
  - [x] Wrangler PR: https://github.com/cloudflare/workers-sdk/pull/10890
  - [ ] Not necessary because: no public-facing functional change

### DIFF
--- a/.changeset/plain-women-think.md
+++ b/.changeset/plain-women-think.md
@@ -1,0 +1,5 @@
+---
+"miniflare": patch
+---
+
+Document that dumpSql is shared with d1-worker, and incorporate some D1 internal stats gathering machinery (which is currently unused by miniflare)
--- a/packages/miniflare/src/workers/d1/dumpSql.ts
+++ b/packages/miniflare/src/workers/d1/dumpSql.ts
@@ -1,3 +1,6 @@
+// NOTE: this file is duplicated between miniflare and d1-workers, and should be kept in sync by hand.
+// Please tag someone from the D1 team as a reviewer and they will do this for you.
+//
 // NOTE: this function duplicates the logic inside SQLite's shell.c.in as close
 // as possible, with any deviations noted.
 import { SqlStorage } from "@cloudflare/workers-types/experimental";
@@ -8,8 +11,11 @@ export function* dumpSql(
 		noSchema?: boolean;
 		noData?: boolean;
 		tables?: string[];
-	}
+	},
+	/** Optional stats tracking. Will be mutated in place if provided */
+	stats?: { rows_read: number; rows_written: number }
 ) {
+	// WARNING: the caller in D1 assumes non-empty exports, so think carefully before removing this initial yield.
 	yield `PRAGMA defer_foreign_keys=TRUE;`;
 
 	// Empty set means include all tables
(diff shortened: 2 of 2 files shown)


.changeset/plain-women-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"miniflare": patch
+---
+
+document that dumpSql is shared with d1-workers
```
dario-piotrowicz: isn't this PR also adding the `stats` argument to miniflare? 🤔

dario-piotrowicz:
I think that this change should actually be backported to wrangler v3 🤔

> Wrangler V3 Backport
>   Not necessary because: no functional change

But there are functional changes here no? 🤔

dario-piotrowicz:
(PS: sorry for reviewing the PR already, I only now noticed that it's in draft 😅)

.changeset/plain-women-think.md
alsuren: Fair point. Updated.

It's worth noting that dumpSql is not part of miniflare's public api. It is used by database.worker.ts only if you execute `PRAGMA miniflare_d1_export()`. This is why I don't think it's worth backporting. I don't expect miniflare to ever use the stats argument to dumpSql()

It may be that we will want to backport this file wholesale when we next fix a bug in dumpSql.ts, but I don't think we need to do so until it causes conflicts.

> (PS: sorry for reviewing the PR already, I only now noticed that it's in draft 😅)

This was my fault. I made it a non-draft PR because I thought it was going to be a simple change, and then I realised that there was a bit more work to do because of how our internal copy has diverged over time.

packages/miniflare/src/workers/d1/dumpSql.ts
```diff
@@ -83,25 +93,45 @@ export function* dumpSql(
...
 		const rest_of_schema = db.exec(
-			`SELECT name, sql FROM sqlite_schema AS o WHERE (true) AND sql NOT NULL AND type IN ('index', 'trigger', 'view') ORDER BY type COLLATE NOCASE /* DESC */;`
+			[
+				"SELECT name, sql",
+				"FROM sqlite_schema AS o",
+				"WHERE (true) AND sql NOT NULL",
+				"  AND type IN ('index', 'trigger', 'view')",
+				// 'DESC' appears in the code linked above but the observed behaviour of SQLite appears otherwise
+				"ORDER BY type COLLATE NOCASE",
+			].join(" ")
```
alsuren, line 129: This is slightly sad, but D1 uses 2 spaces for indentation and workers-sdk uses tabs. This is the best way I could think of to make both autoformatters produce something that doesn't make your eyes bleed. There is no functional change here.

dario-piotrowicz: 🥹👍

dario-piotrowicz (approved):
LGTM 🙂 

(For posterity) as discussed internally we agreed with the decision not to include tests here 🙂 

The backport is also not super necessary but since (after I asked... sorry about that 🙇) you already got the backport PR ready we can also proceed with that 😄 

Thanks @alsuren 🫶

